PR #1267, opened 2018-11-19
Is building underground quest [ready for review]

### LINKED ISSUE #912: add location to buildings with negative layer from -1 to -5
### General
Affected tag(s) to be modified/added: [location](https://wiki.openstreetmap.org/wiki/Key:Location)
Question asked: **Is this building is existing and located underground?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose (it is used in StreetComplete itself, allows to clearly mark something as underground - some mappers use solely negative layer/level and expect that it is enough to describe something as located underground)
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one) location=underground/surface/overground/overhead should cover all cases, except complete mistaggings
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary: distinguishing location=surface/overground/overhead requires survey, checking whatever structure exists requires survey

### Ideas for implementation

**Element selection:** http://overpass-turbo.eu/s/wv2

**Metadata needed:** no

**Proposed GUI:**

matkoniecz:
fixes #912

this a bit unusual quest help to fix one of common tagging mistakes

main use of adding missing location=underground in case of SC is to skip this weird buildings
in other quests

And here is usage example: https://www.openstreetmap.org/changeset/64655345#map=19/48.20834/16.37226&layers=N

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
@@ -20,7 +20,7 @@ public class IsBuildingUnderground extends SimpleOverpassQuestType
 
 	@Override protected String getTagFilters()
 	{
-		return "ways, relations with building and !location and layer~-.*";
+		return "ways, relations with building and !location and layer~-[0-9]+";
 	}
 
 	public AbstractQuestAnswerFragment createForm()
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
@@ -37,7 +37,7 @@ public class IsBuildingUnderground extends SimpleOverpassQuestType
 		}
 	}
 
-	@Override public String getCommitMessage() { return "Add whatever building is fully underground"; }
+	@Override public String getCommitMessage() { return "Determine whatever building is fully underground"; }
 	@Override public int getIcon() { return R.drawable.ic_quest_building; }
 	@Override public int getTitle(@NonNull Map<String, String> tags)
 	{
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
@@ -1,0 +1,51 @@
+package de.westnordost.streetcomplete.quests.building_underground;
+
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+
+import java.util.Map;
+
+import javax.inject.Inject;
+
+import de.westnordost.streetcomplete.R;
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType;
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder;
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
```diff
@@ -0,0 +1,51 @@
...
+import de.westnordost.streetcomplete.quests.AbstractQuestAnswerFragment;
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment;
+
+public class IsBuildingUnderground extends SimpleOverpassQuestType
+{
+	@Inject public IsBuildingUnderground(OverpassMapDataDao overpassServer) { super(overpassServer); }
+
+	@Override protected String getTagFilters()
+	{
+		return "ways, relations with building and !location and layer~-.*";
```
westnordost: Did you test whether the tag filter syntax is properly translated to valid overpass? Because of `layer~-.*`, it would be better to test it once.

app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
```diff
@@ -0,0 +1,51 @@
...
+	public void applyAnswerTo(Bundle answer, StringMapChangesBuilder changes)
+	{
+		if(answer.getBoolean(YesNoQuestAnswerFragment.ANSWER)){
+			changes.add("location", "underground");
+		} else {
+			changes.add("location", "surface");
+		}
+	}
+
+	@Override public String getCommitMessage() { return "Add whatever building is fully underground"; }
```
westnordost: Or 
```suggestion
	@Override public String getCommitMessage() { return "Determine whatever building is fully underground"; }
```
?

app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
```diff
@@ -0,0 +1,51 @@
...
+	@Override public int getIcon() { return R.drawable.ic_quest_building; }
```
westnordost: I made a special icon for this already a while back. If it is not already exported, you will find it in the quest icons svg

app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
matkoniecz: It found expected buildings in Vienna, including one where I was able to make the edit ( https://www.openstreetmap.org/changeset/64655345#map=19/48.20834/16.37226&layers=N ).

I was not checking generated code of overpass turbo query.

app/src/main/java/de/westnordost/streetcomplete/quests/building_underground/IsBuildingUnderground.java
matkoniecz: It seems to not be exported. From looking at https://github.com/westnordost/StreetComplete/commit/429a92c932f95acb2356adce3965f66634760f21 it seems that exporting means saving .svg pattern (copied from larger file) to a new file with .xml extension.

westnordost:
added it

matkoniecz:
I really like this icon, really surprised me - and in a pleasant way.

westnordost:
thanks ^^

PR #1269, opened 2018-11-20
ask whatever traffic lights have button, fixes #574 [change icon background]

### LINKED ISSUE #574: New Quest: Do these traffic lights have a button?
Hi, this is a series of new possible Quests to give more details for pedestrian traffic signals. Since there are already quests for these crossings, it shouldn't be hard to implement these (I'm not a programmer, but I guess so).
The point is, that it's a good thing for disabled, especially blind people, to now, if the traffic light is button_operated, has a sound, etc. But these informations are very rare on OSM, because you don't notice them as a normal person.

Necessary Tags on the Node:
_highway=crossing_ + _crossing=traffic_signals_
Sould also work for:
_highway=traffic_signals_ + _crossing=traffic_signals_

### General
Affected key(s) (or tags) to be modified: _button_operated=yes/no_
Question asked: _Is this crossing button operated?_

Affected key(s) (or tags) to be modified: _traffic_signals:sound=yes/no_
Question asked: _Are there sound signals for visually impaired people?_

Affected key(s) (or tags) to be modified: _traffic_signals:arrow=yes/no_
Question asked: _Is there a tactile arrow available at the traffic light pole? (mostly unter the button)_

I think these are the most common supports for blind people at traffic signals, the other ones are too rare to make an own quest for them.

### Checklist
- [x ] 🚧 To be added tag is established and has a useful purpose
- [x ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x ] 🐿️ Easily answerable by everyone from the outside
- [x ] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [ x] 🕓 Applies to a reasonable number of elements (Worth the effort)

matkoniecz:


rugk:
Nitpick, but this is totally GitHub's awkwardness here: The "Fixes …" must not be used in the title; it does not work. I am sorry… :wink:

heinrich5991:
It suffices if it is in the commit message. If it *only* in the title, it doesn't work.

rugk:
Yeah, it works in some strange way by then closing the issue, when the commit message is merged into master, but that is not how it is supposed to work. Anyway, does not really matter…

app/src/main/res/values/strings.xml
```diff
@@ -624,5 +624,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_tracktype_grade5">Soft</string>
     <string name="quest_building_underground_name_title">Is %s completely underground?</string>
     <string name="quest_building_underground_title">Is this building completely underground?</string>
+    <string name="quest_traffic_signals_button_title">Do these traffic lights have a button?</string>
```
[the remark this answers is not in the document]
matkoniecz: "to request a green light" (I am not really happy with that but I have no better ideas). Or maybe add a photo with the button?

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
```diff
@@ -145,6 +146,7 @@
 				new AddPowerPolesMaterial(o),
 				new AddCarWashType(o),
 				new AddBenchBackrest(o),
+				new AddTrafficSignalsButton(o),
```
rugk: Should not this be prioritized higher (use cases [were given here](https://github.com/westnordost/StreetComplete/issues/574#issuecomment-440344393). I'd just do so that the quest "traffic_signals:sound" is asked _after_ this traffic_signals:button_operated, because it is usually easier to answer whether that thing has a button (you can answer that while standing and waiting at the traffic light), while for the sound you have to wait until you get green.

So IMHO the logical order to answer these quests is:
* button – when I press it
* [wait]
* sound – when I (do not) hear it when it is green

goldfndr:
Should perhaps some other elements be added?  [There are many other crossing=traffic_signals](https://overpass-turbo.eu/s/DQG):

* ways with highway=footway
* nodes with highway=traffic_signals

(Surely railway=traffic_signals would be excluded, as those wouldn't be button operated.)

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
goldfndr: How about ["push-button"](https://en.wikipedia.org/wiki/Push-button) instead of "button"?

"Do these traffic lights have a push-button that can increase pedestrian priority?"? (Admittedly nowhere near [up goer 5 friendly](http://splasho.com/upgoer5/).)

matkoniecz:
> nodes with highway=traffic_signals

It may be road crossing without pedestrian crossing.

Also, I think it is fine to wait with mapping this detail until pedestrian highway=crossing are marked.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
matkoniecz: "while for the sound you have to wait until you get green", wait, so cases described at https://wiki.openstreetmap.org/wiki/Key:traffic%20signals:sound?uselang=en " there is only the signal to find the pole" and "there is only the signal when walking is allowed" are actually sometimes happening in some regions?

matkoniecz: On the other hand noone actually uses this values, so it may be probably safely ignored as "someone wanted this tagging scheme and believed that OSM Wiki has magic powers" - see https://taginfo.openstreetmap.org/keys/traffic_signals:sound#values

rugk:
@matkoniecz Did you close this by accident?

matkoniecz:
Yes. I have no idea what happened.

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
westnordost: to request green light is good. I don't know what is the difference between a push-button and a button.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
rugk: Hmm, did not really thought about that… Maybe yes, I only noticed the "green 'beep'" more often.

rugk: But I am fairly sure that those traffic lights, where here is "only an accessibility button", i.e. the one that does not influence the traffic light/time to wait, but just enables the sound/accessibility; have no sound by default and only have it when you press that button.

PR #1302, opened 2018-12-28
Add an option to restore hidden quests (fix #302)

### LINKED ISSUE #302: Add an option to restore hidden quests
This would be useful if you have accidently hidden a quest or if you want to solve a quest that has become solvable.

ENT8R:
Just a small a small PR, but I thought it might be useful sometimes 😄 

Fixes #302


westnordost:
ok

PR #1309, opened 2018-12-31
Add ways to search query

MrKrisKrisu:
Some bus stops are tagged as way


westnordost:
Ok. I refactored the file, so I re-did your change.

PR #1315, opened 2019-01-06
Kotlin

westnordost:


app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
```diff
@@ -104,8 +104,8 @@ abstract class AbstractQuestAnswerFragment : AbstractBottomSheetFragment() {
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         val view = inflater.inflate(R.layout.fragment_quest_answer, container, false)
         content = view.findViewById(R.id.content)
-	    buttonPanel = view.findViewById(R.id.buttonPanel)
-	    otherAnswersButton = buttonPanel.findViewById(R.id.otherAnswersButton)
+        buttonPanel = view.findViewById(R.id.buttonPanel)
```
rugk: How did this issue happen? I thought Android Studio enforces these styling things?

However, `.kt` is not listed in the editorconfig file, [only java](https://github.com/westnordost/StreetComplete/blob/master/.editorconfig#L14), so maybe you want to edit your preferred styling there?

So this does not happen again... especially for contributors. 😜

westnordost: ok

PR #1318, opened 2019-01-07
Update orchard/crops data

rugk:
Fixes https://github.com/rugk/crops-parser/issues/57

For some reason, the file size doubled, but I have no idea why. Well… maybe more data, certainly won't hurt. :smile:


westnordost:
Have a look at the data you want to replace the previous data with and you will see why.

rugk:
I had a quick look and upps, yeah… too quick, sry. Next try… :laughing:

PR #1325, opened 2019-01-31
Add popular religions for DE

Discostu36:


westnordost:
It doesn't really matter whether Judaism is shown in the first or second row of icons in terms of convenience (default order is Christian, Muslim, Buddhist, Hindu, Judaism,...), but fair enough, why not

Discostu36:
Well, this quest is about places of worship, not about the religion of the population. And I've been to several synagogues but I don't remember seeing any Buddhist, Hindu or Yazidi places of worship.

rugk:
> I don't remember seeing any Buddhist, Hindu or Yazidi places of worship.

Propbably, because they are not visible from the outside (can be usual resistential buildings). That could then also be reflected in the OSM data, so I'd consider personal experience a bad "data source" here.

westnordost:
@Discostu36 Well, problem is, that the research I did for the `popularReligions.yml` for all the countries specified there was based on (wikipedia) data about the number of followers, not on how many places of worships can be found. But ok, let's leave it at that. No point in discussing this if the effect is going to be just if something turns up a little further up or down the list.

PR #1327, opened 2019-02-02
Edit metadata for South Korea

depth221:
I think this metadata will be useful for South Korea than previous metadata.
Thank you in advance!


depth221:
Then, how about editing the popular sports as "baseball, soccer, golf, basketball, and badminton" based on [this like](https://www.statista.com/statistics/829536/south-korea-most-popular-sports/)? this link are used in the Wikipedia article as [reference](https://en.wikipedia.org/wiki/Sport_in_South_Korea#cite_note-3).

depth221:
Unfortunately, I can no longer read the link because of paywall.

depth221:
I understood your intention, and think it is very logical. Then these edits aren't likely to be needed. Thanks!

PR #1329, opened 2019-02-06
Changesets refactor

westnordost:
5 steps:
- [x] refactor the quest upload to be more easily extendable and more easily testable. (I am slowly losing the overview at 500+ LOC)
- [x] in the future, treat (substantial) changes to the geometry as unsolvable conflict and drop quest answer. For example:
  - shop node has been moved hundreds of meters (shop moved or node was re-purposed by a mapper)
  - road end(s) have been extended or merged with another road-piece
  - ... (will need to think about further what changes on the geometry are still ok)
- [x] optionally release the changes done so far
- [x] implement the ability to split up a way at a certain position, including propagating the change to possible relations automatically
- [x] persisting split ways for offline usage (+undo?), interweave with quest system
- [x] Add UI for all or some way-based quests to cut a way into two pieces when choosing he other answer option "It differs along the way..."


westnordost:
> implement the ability to split up a way at a certain position, including propagating the change to possible relations automatically

[x] all that work done already
[ ] tests for updating relations
[ ] special logic for splitting up closed ways (require 2 split points and merge ends)
[ ] tests for that special logic

app/src/main/java/de/westnordost/streetcomplete/data/osm/split/SplitWayUpload.kt
```diff
@@ -0,0 +1,258 @@
...
+                }
+            }
+        }
+        return result
+    }
+
+    /** Returns whether it has been treated as a special relation type */
+    private fun updateSpecialRelation(relation: Relation, indexOfWayInRelation: Int, newWays: List<Way>): Boolean {
+        val relationType = relation.tags?.get("type") ?: ""
+        if (relationType == "restriction" || relationType == "destination_sign") {
```
matkoniecz: https://github.com/openstreetmap/iD/issues/6221 proposes to treat in this way all restrictions with `from` and `to` members. This seems to make sense.

westnordost: That sounds reasonable, but required also is a "via" node. And for `destination_sign`, the wiki says, that the role should actually be "intersection" (or "sign") and not "via".

So, depending on the relation type, it must still be decided which one is the via node/way. Not sure how iD handles this.

app/src/main/java/de/westnordost/streetcomplete/data/osm/split/SplitWayUpload.kt
```diff
@@ -0,0 +1,261 @@
...
+
+        val diffDelta = split1.delta - split2.delta
+        return if (diffDelta < 0) -1 else if (diffDelta > 0) 1 else 0
+    }
+}
+
+private fun createSplitPosition(firstPosition: LatLon, secondPosition: LatLon, delta: Double) =
+    SphericalEarthMath.createTranslated(
+        firstPosition.latitude + delta * (secondPosition.latitude - firstPosition.latitude),
+        firstPosition.longitude + delta * (secondPosition.longitude - firstPosition.longitude))
```
matkoniecz: It seems to be breaking across 180 line. Likely such objects are already discarded or impossible to appear or it should be rare but I would consider throwing an exception...

`firstPosition.longitude + delta * (secondPosition.longitude - firstPosition.longitude)`

```
firstPosition.longitude = 179
secondPosition.longitude = -179
delta = 0.5
```

Expected: 180 or -180

`179 + 0.5 * (-179 - 179)` = `179 + 0.5 * -179 * 2` = 0`

westnordost: Right, that is a very good point! Thank you!

app/src/main/java/de/westnordost/streetcomplete/data/osm/split/SplitWayAtPosition.kt
```diff
@@ -0,0 +1,28 @@
+package de.westnordost.streetcomplete.data.osm.split
+
+import de.westnordost.osmapi.map.data.Node
+import de.westnordost.osmapi.map.data.Way
+
+/** data class that carries the information for one split to perform on a random position on a way. */
+class SplitWayAtPosition(way: Way, val firstNode: Node, val secondNode: Node, val delta: Double) {
```
matkoniecz: Not sure where it should be handled - maybe here, maybe in upper levels of API that are not yet written.

But in general, especially in StreetComplete delta extremely close to 0 or 1 should be clamped to 0/1.

It makes no sense to split way and create a new node centimeters from already existing one - it is better to split way on an already existing node.

matkoniecz:
One thing that seems to be missing - on splitting some closed ways one needs to create a multipolygon (on splitting say `natural=water` closed way).

Maybe it can be assumed that it will not happen in StreetComplete and is not worth implementing but may be worth noting it in comments that it is supposed to not happen.

app/src/main/java/de/westnordost/streetcomplete/data/osm/split/SplitWayAtPosition.kt
westnordost: It should be handled in the UI as the clamp should be visible there as well

westnordost:
Multipolygon? Not sure what you mean. A closed way is just a closed way, if you split it up, you get two unclosed ways. That is the behavior of JOSM.

matkoniecz:
Yes, but 

* there is closed way with `natural=water`
* way is split in two `natural=water` ways

is incorrect (water area will disappear), and following is necessary:

* there is closed way with `natural=water`
* way is split in two `natural=water` ways
* multipolygon relation is created, with this two ways as members
* tags are removed from ways and added to relation

westnordost:
Probably not a multipolygon but two separated closed ways. But anyway, splitting a way is quite a different operation from splitting an area. The way is just 1-dimensional, so it can be split in a point. The area is 2-dimensional, so the split itself is a line, or a polyline.

In OSM, there exists no data structure for an area. I wonder, does JOSM have such a function, splitting a closed way (=area) along a (poly)line?

I can imagine use cases in StreetComplete where such a function would be helpful, for example if one wants to split up a building, a farmland or so. I'll keep this in mind, but for now, this won't be part of the implementation. A simple split of a way is already complicated enough, also UI wise. Imagine the user would need to draw a polyline (on a smartphone screen) to split an area with.

westnordost:
I'll probably be back from my "cave" this weekend, pushing a huge refactor of the whole upload system:
- persisting split way uploads, integrating it into the upload service
- pulling apart the different functionalities into separate classes that are also separately tested. For example, one class to upload a single quest answer change, one class to manage changesets, one class to update the diff-result of an upload, one class to coordinate the upload of all quest answers
- not letting the undo osm quests inherit functionality from osm quests as certain fields (such as quest status, last update time etc) are irrelevant for it

app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/persist/OsmQuestSplitWayDaoTest.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test fun putAndGet() {
+	    val id = 1L
+	    val input = createOsmQuestSplitWay(id)
+	    dao.put(input)
+	    val output = dao.get(id)!!
+
+	    assertEquals(input.questType, output.questType)
+	    assertEquals(input.wayId, output.wayId)
+	    assertEquals(input.id, output.id)
```
westnordost: maybe `questId` instead

app/src/main/java/de/westnordost/streetcomplete/ApplicationComponent.java
```diff
@@ -1,5 +1,7 @@
 package de.westnordost.streetcomplete;
 
+import org.jetbrains.annotations.NotNull;
```
westnordost: use `NonNull`instead

app/src/main/java/de/westnordost/streetcomplete/data/osm/persist/OsmQuestSplitWayDao.kt
```diff
@@ -0,0 +1,63 @@
...
+        dbHelper.writableDatabase.delete(NAME, "$QUEST_ID = $id", null)
+    }
+
+    private fun OsmQuestSplitWay.createContentValues() = ContentValues().also { v ->
+        v.put(QUEST_ID, id)
+        v.put(QUEST_TYPE, questType.javaClass.simpleName)
+        v.put(WAY_ID, wayId)
+        v.put(SOURCE, source)
+        v.put(SPLITS, serializer.toBytes(ArrayList(splits)))
+    }
```
westnordost: extension function Map<String,Any?> to ContentValues? Or factory function that creates a ContentValues from a vararg of Pair<String,Any?>?

app/src/main/java/de/westnordost/streetcomplete/ktx/LatLon.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import de.westnordost.osmapi.map.data.LatLon
+import kotlin.math.abs
+
+/** OSM has limited precision of 7 decimals */
+fun LatLon.equalsInOsm(other: LatLon) =
+    !latitude.isDifferent(other.latitude, 1e-7) &&
+    !longitude.isDifferent(other.longitude, 1e-7)
+
+private fun Double.isDifferent(other: Double, delta: Double) = abs(this - other) > delta
```
westnordost: add test

app/src/main/java/de/westnordost/streetcomplete/data/osm/persist/OsmQuestSplitWayDao.kt
westnordost: Implemented, but not used, because it is not type safe during compile time

app/src/main/java/de/westnordost/streetcomplete/data/osm/persist/UndoOsmQuestDao.kt
```diff
@@ -0,0 +1,63 @@
...
+        questTypeList.getByName(getString(QUEST_TYPE)) as OsmElementQuestType<*>,
+        Element.Type.valueOf(getString(ELEMENT_TYPE)),
+        getLong(ELEMENT_ID),
+        serializer.toObject(getBlob(TAG_CHANGES)),
+        getString(CHANGES_SOURCE),
+        ElementGeometryDao.createObjectFrom(serializer, this)
+    )
+}
+
+// TODO test for this!!
```
westnordost: dont forget

westnordost:
Now reviewing the implementation, testing, adding more unit tests

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+    fun upload(changesetId: Long, way: Way, splits: List<SplitPolylineAtPosition>): List<Way> {
+        val updatedWay = way.fetchUpdated()
+            ?: throw ElementDeletedException("Way #${way.id} has been deleted")
+        if(updatedWay.isClosed() && splits.size < 2)
+            throw ElementConflictException("Must specify at least two split positions for a closed way")
+        checkForConflicts(way, updatedWay)
+
+        val nodes = updatedWay.fetchNodes()
+        val positions = nodes.map { it.position }
+        val sortedSplits = splits.map { it.toSplitWay(positions) }.sorted()
```
westnordost: add comment why this is important

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        uploadElements.addAll(splitWayAtIndices(updatedWay, splitAtIndices))
+        val handler = UpdateElementsHandler(uploadElements)
+        try {
+            osmDao.uploadChanges(changesetId, uploadElements, handler)
+        } catch (e: OsmConflictException) {
+            throw ChangesetConflictException(e.message, e)
+        }
+        return handler.updatedElements.filterIsInstance<Way>()
```
westnordost: add comment why only ways

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+        /* Handle circular ways specially: If you split at a circular way at two nodes, you just
+           want to split it at these points, not also at the former endpoint. So if the last node is
+           the same first node, join the last and the first way chunk. (copied from JOSM) */
+        if (nodesChunks.size > 1 && nodesChunks.first().first() == nodesChunks.last().last()) {
+            val lastChunk = nodesChunks.removeAt(nodesChunks.lastIndex)
+            lastChunk.removeAt(lastChunk.lastIndex)
+            nodesChunks.first().addAll(0, lastChunk)
+        }
+
+        val indexOfChunkToKeep = nodesChunks.indexOfMaxBy { it.size }
```
westnordost: add comment about chunk to keep

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+        val tags = originalWay.tags?.toMap()
+        var newWayId = -1L
+        return nodesChunks.mapIndexed { index, nodes ->
+            if(index == indexOfChunkToKeep) {
+                val way = OsmWay(originalWay.id, originalWay.version, nodes, tags)
+                way.isModified = true
+                way
```
westnordost: .also?

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+            }
+        }
+        return result
+    }
+
+    /** Returns whether it has been treated as a special relation type */
+    private fun updateSpecialRelation(relation: Relation, indexOfWayInRelation: Int, newWays: List<Way>): Boolean {
+        val relationType = relation.tags?.get("type") ?: ""
+        val originalWayRole = relation.members[indexOfWayInRelation].role
+        if (originalWayRole == "from" || originalWayRole == "to") {
```
westnordost: explain in short what from-to-relations are about (why they need special handling)

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+            }
+        }
+    }
+
+    /** Returns the elements that have been changed */
+    private fun updateRelations(originalWay: Way, newWays: List<Way>) : Collection<Relation> {
+        val relations = originalWay.fetchParentRelations()
+        val result = mutableSetOf<Relation>()
+        for (relation in relations) {
+            for(i in relation.members.size - 1 downTo 0) {
```
westnordost: explain why iterating in reverse

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+        val newRelationMembers = newWays.map { way ->
+            OsmRelationMember(way.id, originalWayRole, WAY) }.toMutableList()
+        val isOrientedBackwards = originalWay.isOrientedForwardInOrderedRelation(relation, indexOfWayInRelation) == false
+        if (isOrientedBackwards) newRelationMembers.reverse()
+
+        relation.members.removeAt(indexOfWayInRelation)
+        relation.members.addAll(indexOfWayInRelation, newRelationMembers)
+    }
+
+    private fun Way.fetchNodes(): List<Node> = osmDao.getNodes(nodeIds)
```
westnordost: handle Exception?

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+    val pos1: LatLon, val index1: Int,
+    val pos2: LatLon, val index2: Int,
+    public override val delta: Double) : SplitWay() {
+    override val index get() = index1
+    override val pos: LatLon get() {
+        val line = listOf(pos1, pos2)
+        return SphericalEarthMath.pointOnPolylineFromStart(
+            line,
+            SphericalEarthMath.distance(line) * delta
+        )!!
```
westnordost: one-liner

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+ *  If the way changed significantly in the meantime, it will throw an ElementConflictException */
+private fun SplitPolylineAtPosition.toSplitWay(positions: List<LatLon>): SplitWay {
+    return when(this) {
+        is SplitAtPoint -> toSplitWay(positions)
+        is SplitAtLinePosition -> toSplitWay(positions)
+    }
+}
+
+private fun SplitAtPoint.toSplitWay(positions: List<LatLon>): SplitWayAtPoint {
+    // could be several indices, for example if the way has the shape of an 8.
```
westnordost: Maybe then the split should be done at *every* node at that position. Same for lines.

app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/persist/OsmQuestSplitWayDaoTest.kt
```diff
@@ -0,0 +1,83 @@
...
+		dao.put(createOsmQuestSplitWay(id, 1L))
+		dao.put(createOsmQuestSplitWay(id, 123L))
+		assertEquals(123L, dao.get(id)?.wayId)
+		assertEquals(1, dao.getAll().size)
+	}
+
+	@Test fun getAll() {
+		dao.put(createOsmQuestSplitWay(1L, 1L))
+		dao.put(createOsmQuestSplitWay(2L, 2L))
+		assertEquals(2, dao.getAll().size)
```
matkoniecz: Maybe add also assert about where split was done?

app/src/main/java/de/westnordost/streetcomplete/data/QuestController.java
```diff
@@ -178,16 +202,26 @@ public boolean createNote(long osmQuestId, String questTitle, String text, @Null
...
 	{
-		CreateNote createNote = new CreateNote();
-		createNote.position = position;
-		createNote.text = text;
-		createNote.imagePaths = imagePaths;
-		createNoteDB.add(createNote);
+		OsmQuest q = osmQuestDB.get(osmQuestId);
+		// race condition: another thread may have removed the element already (#288)
+		if(q == null || q.getStatus() != QuestStatus.NEW) return false;
+
```
matkoniecz: It seems to me that race condition may happen anyway - with another thread acting after this check and before next instructions.

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SingleOsmElementTagChangesUpload.kt
```diff
@@ -0,0 +1,129 @@
...
+       bigger than the usual GPS inaccuracy in the city. */
+    SphericalEarthMath.distance(node.position, newNode.position) > 20
+
+private fun isWayGeometrySubstantiallyDifferent(way: Way, newWay: Way) =
+    /* if the first or last node is different, it means that the way has either been extended or
+       shortened at one end, which is counted as being substantial:
+       If for example the surveyor has been asked to determine something for a certain way
+       and this way is now longer, his answer does not apply to the whole way anymore, so that
+       is an unsolvable conflict. */
+    way.nodeIds.firstOrNull() != newWay.nodeIds.firstOrNull() ||
```
matkoniecz: What about entire way being moved? I made/notices such edits where objects were dragged for a long distance rather than delete/recreate it.

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
```diff
@@ -0,0 +1,301 @@
...
+        if(updatedWay.isClosed() && splits.size < 2)
+            throw ElementConflictException("Must specify at least two split positions for a closed way")
+        checkForConflicts(way, updatedWay)
+
+        val nodes = updatedWay.fetchNodes()
+        val positions = nodes.map { it.position }
+        val sortedSplits = splits.map { it.toSplitWay(positions) }.sorted()
+
+        val uploadElements = mutableListOf<Element>()
+        var newNodeId = -1L
```
matkoniecz: Always start from -1? What happens with more than one way split in changeset (or is it guaranteed to not happen)?

westnordost: Yes, always start from -1. The "temporary ids" (negative ids) are, to my knowledge, scoped per POST call to `changeset/$id/upload`, not per changeset. This makes sense, if you think about that the newly added elements will be on the map immediately after that call, not only after the changeset is closed.

app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/persist/OsmQuestSplitWayDaoTest.kt
westnordost: Don't know what you mean

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SingleOsmElementTagChangesUpload.kt
westnordost: This kind of reusing element ids for entirely different things is frowned upon in the community, because it carries over the history of a node that is entirely unrelated to its current usage. Additionally, the moving of its nodes does not appear in the history of the way or relation, making it really hard to retrace the "real" history in a case like this. In short, just don't do it.
So, in the unlikely event that a mistake is introduced because (most likely) a node got this treatment, I don't think anyone would blame StreetComplete for it. 

If StreetComplete were to safeguard against this, it would mean that any even so slight adjustment of the position of the node and worse, any node within a way, would need to be treated as an unsolvable conflict. (Fuzzy "if X% of nodes moved more than Ym"-checks are out of the question) Slight adjustments of nodes are very common.

In short, I won't make this check more strict.

app/src/main/java/de/westnordost/streetcomplete/data/QuestController.java
westnordost: Theoretically yes, but what thread should that be? `splitWay` is called by user interaction, the user is not able to quickly answer the quest while this method is executed, because this method is executed on the main thread as well. It doesn't look like the fact that it is executed on the main thread creates a noticeable hiccup in the UI but if you are interested, you could measure how long it takes for this method to finish. 

Though, even if it wouldn't be on the main thread, the user would not have enough time to quickly answer the question after all while this method executes. If the upload thread removes the quest in the meantime because the element changed substantially, it would also be no problem, because the `OsmQuestSplitWay` is created and saved to be uploaded later, but on upload, it will simply fail with a conflict then.

app/src/main/java/de/westnordost/streetcomplete/util/SphericalEarthMath.java
```diff
@@ -104,12 +104,43 @@ public static double enclosedArea(BoundingBox bbox)
...
-				Math.toRadians(pos2.getLongitude()
-				));
+		return distance(
+			toRadians(pos1.getLatitude()),
+			toRadians(pos1.getLongitude()),
+			toRadians(pos2.getLatitude()),
+			toRadians(pos2.getLongitude())
+		);
+	}
+// TODO test these two functions!
```
westnordost: dont forget

app/src/main/java/de/westnordost/streetcomplete/quests/SplitWayFragment.kt
```diff
@@ -0,0 +1,196 @@
...
+import javax.inject.Inject
+
+class SplitWayFragment : Fragment(), IsCloseableBottomSheet {
+
+    interface Listener {
+        fun onAddSplit(point: LatLon)
+        fun onRemoveSplit(point: LatLon)
+        fun onSplittedWay(osmQuestId: Long, splits: List<SplitPolylineAtPosition>)
+    }
+    private val splits: MutableList<Pair<SplitPolylineAtPosition, LatLon>> = mutableListOf()
```
westnordost: save instance state?

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
westnordost: So, the question is how this should behave for ways that look for example like this:
Untitled

If the user clicks on the point where the way intersects itself to split it, should the app split the way at both nodes at that position or just the first?

westnordost:
Okay, looks like the feature is done. All features done, review done, all unit tests done and testing of the feature done. I tested one split [here](https://www.openstreetmap.org/changeset/73704022) successfully.

This feature is huge, so it needs more internal pre-beta testing, also testing the edge cases. Of course I wrote unit tests, but unit tests can only cover the edge cases I anticipated.

Is anyone willing to help out to test this? Best would is set up looking at the device logs later or at least check his OSM history after trying it out. Even just helping with only a few cases or simple UX testing would be a great help.

Since the PR includes a complete refactor of the whole upload system, as well as how changes are persisted locally, the following needs to be tested:
- [ ] closing forms - for each of any quest + leave note instead + split way + create note via click on map + click on other quest, click on cancel (if applicable), pressing back (testing each). 
- [ ] split way UI of course
  - [ ] animation works
  - [ ] undo within dialog works
  - [ ] quest counter and animation works
  - [ ] screen rotate works (**my smartphone's rotation sensor is broken, I can no longer test this myself!**)
  - [ ] splitting of closed ways
  - [ ] clicks snap to already existing nodes
- [ ] split way upload
  - [ ] quests on split up ways are unlocked
  - [ ] splitting a way that is part of a relation (i.e. bus route)
  - [ ] splitting a way that is part of a from-via-to relation (i.e. turn restriction)
  - [ ] automatic conflict handling: if nodes are inserted or removed on the way that are not near the nodes for which the split is made, or if only tags are changed on the way, the app should still be able to do the split. But if the nodes adjacent to the split(s) have been moved, removed, or nodes inserted in between, or the way as a whole has been extended or shortened, the conflict is not solveable automatically and the change is dropped.
- [ ] undo **an already uploaded** quest and upload the undo (=revert). 
- [ ] solve a quest and upload it.
- [ ] create a note and upload it.
- [ ] leave a note instead of answering and upload it.

HolgerJeromin:
> Is anyone willing to help out to test this? Best would be people who are set up to look at the device logs or at least check their OSM history after trying it out.

I have not much time to test but would like to give it a try. I have Syslog access to my Fairphone 2 and quite osm experience.

westnordost:
Cool! Do you have Android studio on your PC or do you need a build?

app/src/main/java/de/westnordost/streetcomplete/data/osm/upload/SplitSingleWayUpload.kt
westnordost: I decided that it should split the way in this example into three pieces. The implementation was rather simple, and that it what a user would expect.

westnordost:
I created a debug build here:
https://www.westnordost.de/misc/StreetComplete-split-ways.apk

westnordost:
splitway

HolgerJeromin:
I selected a quest (has this way a footway?) and splitted the 70 meter way on about 15 points :)
Some split could be set less than a meter from the end of the osm way. Perhaps we need a safety border around the ends?
After that I got one star on the upper left. (new install, no osm auth till now).
I expected that I can define the footway without uploading it. 
This would help not only for new users but for offline usage, too.

westnordost:
>After that I got one star on the upper left. (new install, no osm auth till now).
> I expected that I can define the footway without uploading it.

It doesn't upload it if you are not authorized. If you have auto-sync on, the app hides the information what is upload and what not from you, instead, it will try to upload each change immediately. The reason why it did not try to do that, is #1446 .

So, you can split the footway offline, no problem. But the quests for the splitted up way will only appear after that split was uploaded.

westnordost:
I found an edge case and fixed it. The app offers to split the way for areas, i.e. surface of a plaza.

HolgerJeromin:
> So, you can split the footway offline, no problem. But the quests for the splitted up way will only appear after that split was uploaded.

I understand the problems. But the question is not solved. Offline User is probably confused why the question is not raised again after split.

Second problem: I reopened the app and the star counter is zero again. (still not authorized)

goldfndr:
I looked at [the successful test split](https://www.openstreetmap.org/changeset/73704022#map=15/53.5759/9.9341&layers=N) (yay!) and observed that [the tags for the changeset](https://www.openstreetmap.org/api/0.6/changeset/73704022) (quest type, comment) don't match [the osmChange tags](https://www.openstreetmap.org/api/0.6/changeset/73704022/download). Maybe have a different comment if it won't be combined?

westnordost:
I'm not sure what you mean. That it doesn't say "split up way" in the commit message? That is deliberate. The split ways are sorted into the same changesets as the changes made for the quest types the split had been made for - same as for reverts of changes.
This is in order to provide context as to why the split was made, in case the split is not immediately followed up by an answer for the split up parts. (I.e. in offline mode, cause the new quests for the splits only show up after upload)

HolgerJeromin:
Sorry, was not able to read in in the Code. But have a question: 
If I split a small part from a long way, it would be the best if the longest (as in node count) part would get the osm way id and the stub gets a new id from the server.

westnordost:
> Second problem: I reopened the app and the star counter is zero again. (still not authorized)

**OK this sounds like a bug. Will investigate.** I also received [feedback in the German forum](https://forum.openstreetmap.org/viewtopic.php?id=67199), so probably a beta-Release needs to be postponed (it was initially planned to be this Wednesday).
> I understand the problems. But the question is not solved. Offline User is probably confused why the question is not raised again after split.

Yes, they might be. I was hoping that the dialog shown before the split-way screen explains this good enough. Do you or anyone else have a suggestion how to make the wording clearer so that the user is not surprised why he can't answer the question anymore?

>  If I split a small part from a long way, it would be the best if the longest (as in node count) part would get the osm way id and the stub gets a new id from the server.

Yes, this is the case.

@HolgerJeromin Are you interested to test the cases I wrote about above from A-Z(, or some of them)?

westnordost:
> Second problem: I reopened the app and the star counter is zero again. (still not authorized)

Ok fixed

app/src/main/res/values/strings.xml
```diff
@@ -672,4 +672,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="about_missing_stars">Currently, the star count is not shared across devices, but answers are added directly to the OpenStreetMap database.</string>
     <string name="how_to_get_stars">Choose one of the markers on the map and answer the question. This way you will improve OpenStreetMap data and get stars. If you see no markers you may need to zoom out the map.</string>
     <string name="quest_cyclewayPartSurface_title">What\'s the surface of the cycleway here?</string>
+    <string name="quest_generic_answer_differs_along_the_way">"Differs along the way…"</string>
+    <string name="quest_split_way_description">If it differs along the way, the first step is to split up the way. After that, the quest can be answered for each split part separately.\nNote that the quests for the split parts may appear mach later.\nSplit it now?</string>
```
matkoniecz:
```suggestion
    <string name="quest_split_way_description">If it differs along the way, the first step is to split up the way. After that, the quest can be answered for each split part separately.\nNote that the quests for the split parts may appear much later.\nSplit it now?</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -672,4 +672,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="quest_split_way_tutorial">Tap on the way to split it at the point(s) where the answer differs. Try to be as precise as possible, you can zoom in as usual.</string>
+    <string name="quest_split_way_too_imprecise">"Please zoom in further"</string>
```
matkoniecz: This appears also at the maximum possible zoom.

westnordost: Hm. The message appears if you tapped multiple nodes at once with your finger. If the nodes are absurdedly close to each other, then it wouldn't be possible to split up the way.

westnordost: I'd love to just set the maximum zoom further up. But The default maximum zoom in tangram-es is 20.5 and it cannot be turned up further (and only starting at the new tangram-es version that we do not use).

At zoom 20.5, the diameter of the tap of the finger is about 3 meters in our latitude and about 5 on the equator. So, that is about still 4 zoom levels away from achieving the OSM 7-decimals-precision.

So the best choice here is to instead of requiring to further zoom in at already that high of a zoom level, just choose the node that is closest to the center of the tap position.

Will work on it. Thank you!

westnordost: Implemented it. Would you mind testing it again at the position where you discovered it?

westnordost:
While testing, I discovered two more problems:
- [ ] on creation/commenting of each note, the app calls the upload-photos API, even if there are no photos to upload.
- [ ] the (split way) upload does not correctly update the local copy of the uploaded way with the inserted nodes: The inserted nodes are still referred to with negative ids after the upload. This destroys any subsequent attempt to upload new versions of the way (i.e. answering further questions)

So, the release will be postponed further.

app/src/main/res/values/strings.xml
westnordost: @matkoniecz

matkoniecz:
Is the second error manifesting as `de.westnordost.osmapi.common.errors.OsmBadUserInputException: Bad Request
(400) - Placeholder node not found for reference -1 in way 713795301` crashes? Or is it a potential separate issue?

westnordost:
Yes

HolgerJeromin:
> Yes, they might be. I was hoping that the dialog shown before the split-way screen explains this good enough. Do you or anyone else have a suggestion how to make the wording clearer so that the user is not surprised why he can't answer the question anymore (until next upload)?

The first dialog is forgotten. Another Toast "The quest you selected will be available after upload" would have helped me.

westnordost:
Solution still WIP...

westnordost:
Tests listed here complete. Further tests done:
- [ ] split way which is member of multiple different relations
- [ ] split way which is a member of an unknown relation
- [ ] unsolvable conflict during split way upload

PR #1355, opened 2019-03-15
AddProhibitedForPedestrians.kt: remove trailing OR

goldfndr:
I don't know if it makes a computational difference.


westnordost:
Me neither right now, but it is definitely wrong. Thank you!

Issue #1112, opened 2018-06-28
Up to date Quest


schlagi123:
Hello,

I have the idea to create a new quest for checking, if a [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity) is up to date (the amenity itself, opening_hours, etc). I am not sure, if it is possible, because the last update of the data must be check. Furthermore a new tag is needed for showing the up to date status and i do not know, if there exist such a tag. If the amenity is not up to date it should be possible to add another tag to show, that the amenity must be updated.

This is only an idea and should be discussed, because only with such a mechanism is it possible to hold the data up to date.

ENT8R:
I think this is a duplicate of #112, and with that also #314 and maybe also #534
The reason why this won't be implemented can be found here: https://github.com/westnordost/StreetComplete/issues/314#issuecomment-397865306

schlagi123:
you are right thanks

goldfndr:
Perhaps #112 should be reopened?

#314 dealt with a particular mechanism for changing+checking, and #112 was closed because @westnordost thought #314 was a preferred mechanism for changing+checking. But, #314 was closed because said mechanism wasn't preferable after all.

westnordost:
Setting a `check_date`-annotation is still the preferred mechanism (because there is no other), but not applicable for #314 because StreetComplete does currently not offer this kind of changing+checking quests.

This ticket is about creating such a class of quests. However, my comment to @schlagi123 is that it is problematic / does not fit into the scope of this app to let a user review *all* tags of a place, rather than just a specific one (opening hours, name, ...).
The way I imagine this up-to-date quest class to be is that the quests we have now will simply pop up again after some time and ask again for the information to be gathered through the quest. If the information supplied by the surveyor is the same, nothing will be changed except for the `check_date:*` tag. If it is different, the current tag (and `check_date`) will be replaced.

Let's open this ticket again to discuss.

matkoniecz:
> and ask again for the information to be gathered through the quest

At least I would prefer to have this quest in form "is following opening hours info correct?". Comparing this should take far less time than entering opening hours again.

ENT8R:
> At least I would prefer to have this quest in form "is following opening hours info correct?"

[...]

PR #1364, opened 2019-03-30
experimenting with resurveying API [idea in a pseudocode]

matkoniecz:
I was updating my fork to merge in Kotlin changes, and during reimplementing my "show `fixme` tags" quest I had a little idea for #1112 

Boolean quests are the simplest ones to generate resurvey quests in a generic way. There is no benefit from showing possible answer and asking it still up to date, like with for example `opening_hours` quest. Though in case of answer that would result in retagging asking "are you sure" would be useful.

So boolean yes/no quests should be easiest to implement in a generic form (rather than handcraft separate version for every single one).

But currently there is no sane way to generate resurvey quest from current classes.

The idea is to modify how boolean quests are specified, so both usual quest and resurvey quest can be generated from the same data.

This PR has pseudocode of how this info may be specified. It looks OK to me, though maybe because it competed with some truly atrocious ideas (like using reflection).

Main problem with current idea is that YesNoRepeatedQuestAnswerFragment would need access to tags of a given OSM element to ask "are you sure" question, and it may be not feasible to pass that information here.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bike_parking_cover/AddBikeParkingCover.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bike_parking_cover/AddBikeParkingCover.kt
@@ -9,9 +9,12 @@ import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
 class AddBikeParkingCover(o: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(o) {
 
     override val tagFilters = """
-        nodes, ways with amenity=bicycle_parking and access !~ private|no and !covered
+        nodes, ways with amenity=bicycle_parking and access !~ private|no
         and bicycle_parking !~ shed|lockers|building
     """
+    //maybe use something more generic, rather than assuming that single tag is always added?
+    override val addedKey = "covered"
+    override val addedValues = {true: "yes", false: "no"}
     override val commitMessage = "Add bicycle parkings cover"
     override val icon = R.drawable.ic_quest_bicycle_parking_cover
 
@@ -19,8 +22,5 @@ class AddBikeParkingCover(o: OverpassMapDataDao) : SimpleOverpassQuestType<Boole
         R.string.quest_bicycleParkingCoveredStatus_title
 
     override fun createForm() = YesNoQuestAnswerFragment()
-
-    override fun applyAnswerTo(answer: Boolean, changes: StringMapChangesBuilder) {
-        changes.add("covered", if (answer) "yes" else "no")
-    }
+    override fun createRepeatedSurveyForm() = YesNoRepeatedQuestAnswerFragment(addedKey, addedValues)
 }


matkoniecz:
In general, the idea is to change format of how quest is specified to something less readable (hopefully only a bit) but allowing to generate not just standard quest but also resurvey quests.

westnordost:
Can't really say anything concrete other than that I don't like focussing on a solution that can only be applied to boolean quests. I'd need to wrap my head around this further to come up with a (better) solution.

There are two options: First option is to create completely separate quests for the repeated part, and with using Kotlin now, it is also less code now. The second option is to somehow extend the quest interface to facilitate recurring quests in general.
Of course, the second option is much more desirable.

I agree that one part of the solution would be to separate the overpass query into a (to borrow from school mathematics) "necessary condition" and "sufficient condition". These two conditions could simply be concatenated with an `and` for `SimpleOverpassQuestType`s.

In case of the recurring quest the "necessary condition" would be concatenated with a `lastcheck`-query instead and the form to be opened would be a different one which just shows the current data and asks if it is still current, if not, shows the quests original fragment, if yes, just tags a new `lastcheck`.

matkoniecz:
> Can't really say anything concrete other than that I don't like focussing on a solution that can only be applied to boolean quests.

I though about a bit and for now I have following ideas

- [0] do not offer resurvey quests (current state)
- [1] just reask the same question if data is old (really sad in cases like opening hours where survey can be really simplified if old data is still correct)
- [2] automatically prefill data in the answer (applies only to quests with a more complicated form, requires specially crafted for some quests like opening hours, likely require specially crafted code for each separate quest), users answers as usually
- [3] reask question, ask for confirmation in case where answer indicated that data should be changed

[0], [1] is unsatisfactory. [2] applies only to quests where there is fillable form, [3] is useful only for quests where user clicks single button to answer.

Would it be OK to have solution for all "single click to answer" quests, not only nonboolean ones? Or is it necessary to do also [2] (or something better)?

I think that for example proper opening hours resurvey quest would be a big project on its own (requires parsing of opening hours), so I think that solving all at once is not viable.

Hopefully "all single click quests" would be class that would be acceptable to solve in isolation from other quest groups.

rugk:
I'd like [2] quite much. Actually, you can also select an element by default there for quests where you just select one. (and possibly just adjust the question title/header)

I know, the problem then is: One needs to ask a boolean question of whether the data is the same. Actually, the "resurvey quest" always begins with a boolean question, does not it? (It can be implemented kinda as one quest actually.) I.e. "Is this info still up-to-date?".

If no, then we could go to [2]. (At this stage, it is also obvious to the user why things are pre-filled/pre-selected or so.)

This could be the user flow for "non-boolean" quests. For quests that are already a boolean thing, obviously the "Is this info still up-to-date?" should toggle the tag value directly. (maybe after confirmation)

westnordost:
([1] and) [3] sounds like a good approach because it is a reasonably simple solution. OH is an extreme example, most quests are done with a few taps. I think for most quests, not only the single-tap ones, [3] is a good solution. 
But by the way, what quests should be asked recurringly anyway, and in which intervals?

rugk:
IMHO [3] is not really nice, depending on how fast you are resurveying. First a) we already had users (that think) SC showed them an already solved quest again (so this may be dissatisfying for the user). b) resurveying should not take too much effort, especially if you consider the "waste" of time that may go into it when users just select/enter the same answer again – even for the street name quest this would be too much. (Okay, although I'd admit the street name quest may not need resurveying that often or even at all…)

But I admit from the technical site, [3] is obviously easy to implement. :smiley: 

So hmm, … maybe the effort to put into the UX of this depends on how fast you plan to re-survey.

matkoniecz:
>  But by the way, what quests should be asked recurringly anyway, and in which intervals?

I would expect interval measured in years - from 1 for opening hours to 10 for bench backrest quest or building level. My initial guess would be 5 years for maxspeed and maxheight. Maybe even more for a roof shape quest. And support nearly all of them - I would skip building type, underground building, housenumber, parking type, religion and not yet active name quest. Also construction, because it is already implemented as recurring one.

> I think for most quests, not only the single-tap ones, [3] is a good solution.

Looking at the list again - yeah, opening hours quests is actually unique. Bikeway and sidewalk quest are with significant input burden but in others most of time would be needed for survey itself.

matkoniecz:
OK, so this approach may actually work for nearly all quests - resurvey, compare answer with expected and if there is data change ask user whatever they are sure.

Though my solution relies on passing data of OSM element into YesNoRepeatedQuestAnswerFragment (or equivalent for nonboolean quests), so that form can know whatever user should be asked for confirmation.

Is it OK from viewpoint of how data flows in app? I just want to check is it something fundamentally and obviously flawed in such architecture.

westnordost:
Well, a solution would be that the quest form remains oblivious of the fact that this is a recurring question, the logic for the confirmation is handled *after* the user made his answer.

PR #1365, opened 2019-04-01
add office=religion to OH and wheelchair quests

matkoniecz:


westnordost:
I don't even know this.

PR #1366, opened 2019-04-02
Do not ask about railway crossings on private roads [requires retesting]

### LINKED ISSUE #1321: Do not ask "How is this railway crossing protected?" on railway crossings on private roads
It is both hard to survey for (in some cases may be doable only for people working there) and value of this information is really low.

It typically happens within railway yards.

See say https://www.openstreetmap.org/node/1674572452#map=19/50.02815/20.00323 on https://www.openstreetmap.org/way/28013394#map=19/50.02815/20.00323 

I can do that, unfortunately it will make query more complicated.

matkoniecz:
fixes #1321

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
@@ -32,7 +32,6 @@ class AddRailwayCrossingBarrier(private val overpassMapDataDao: OverpassMapDataD
         val bboxFilter = OverpassQLUtil.getGlobalOverpassBBox(bbox)
         return bboxFilter + """
    way["highway"]["access"~"^private|no$"];
-   (._;>;);
    node(w) -> .private;
    node["railway"="level_crossing"][!"crossing:barrier"]->.crossings;
    (.crossings; - .private;)->.public;
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
@@ -1,13 +1,16 @@
 package de.westnordost.streetcomplete.quests.railway_crossing
 
+import de.westnordost.osmapi.map.data.BoundingBox
+import de.westnordost.osmapi.map.data.Element
 import de.westnordost.streetcomplete.R
-import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.OsmElementQuestType
 import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.MapDataWithGeometryHandler
 import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+import de.westnordost.streetcomplete.data.osm.tql.OverpassQLUtil
+import de.westnordost.streetcomplete.quests.AbstractQuestAnswerFragment
 
-class AddRailwayCrossingBarrier(o: OverpassMapDataDao) : SimpleOverpassQuestType<String>(o) {
-
-    override val tagFilters = "nodes with railway=level_crossing and !crossing:barrier"
+class AddRailwayCrossingBarrier(private val overpassMapDataDao: OverpassMapDataDao) : OsmElementQuestType<String> {
     override val commitMessage = "Add type of barrier for railway crossing"
     override val icon = R.drawable.ic_quest_railway
 
@@ -15,7 +18,24 @@ class AddRailwayCrossingBarrier(o: OverpassMapDataDao) : SimpleOverpassQuestType
 
     override fun createForm() = AddRailwayCrossingBarrierForm()
 
+    override fun download(bbox: BoundingBox, handler: MapDataWithGeometryHandler): Boolean {
+        return overpassMapDataDao.getAndHandleQuota(getOverpassQuery(bbox), handler);
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? = null
+
     override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
         changes.add("crossing:barrier", answer)
+    }
+
+    private fun getOverpassQuery(bbox: BoundingBox): String {
+        val bboxFilter = OverpassQLUtil.getGlobalOverpassBBox(bbox)
+        return bboxFilter + """
+   way["highway"]["access"~"^private|no$"];
+   (._;>;);
+   node(w) -> .private;
+   node["railway"="level_crossing"][!"crossing:barrier"]->.crossings;
+   (.crossings; - .private;)->.public;
+   .public out;"""
     }
 }


matkoniecz:
```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':app:connectedDebugAndroidTest'.
> There were failing tests. See the report at: file:///home/travis/build/westnordost/StreetComplete/app/build/reports/androidTests/connected/index.html
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
BUILD FAILED in 3m 58s
```

Well, my local build worked well so it seems that it is not a problem introduced by PR.

Also, master branch is failing anyway.

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrierForm.kt
```diff
@@ -4,18 +4,21 @@ import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.quests.AImageListQuestAnswerFragment
 import de.westnordost.streetcomplete.view.Item
 
-class AddRailwayCrossingBarrierForm : AImageListQuestAnswerFragment<String,String>() {
+class AddRailwayCrossingBarrierForm : AImageListQuestAnswerFragment<AddRailwayCrossingBarrierForm.BarrierOption,RailwayCrossingBarrierAnswer>() {
+    override fun onClickOk(selectedItems: List<BarrierOption>) {
+        if(selectedItems.lastIndex == 0) {
+            applyAnswer(RailwayCrossingBarrierAnswer(selectedItems[0]))
+        }
```
westnordost: That is part of what `List.single()` does in Kotlin. It 1. asserts that there is only one item in the list, and 2. returns it.

westnordost:
I find Travis CI really annoying, perhaps I will disable it. It always fails for *some* reason and only there. I stopped paying attention to this.

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
```diff
@@ -1,21 +1,50 @@
...
-        changes.add("crossing:barrier", answer)
+    private fun getOverpassQuery(bbox: BoundingBox): String? {
+        val bboxFilter = OverpassQLUtil.getGlobalOverpassBBox(bbox)
+        return bboxFilter + "way[\"highway\"][\"access\"~\"^private|no\$\"];\n" +
+                "(._;>;);\n" +
+                "node(w) -> .private;\n" +
+                "node[\"railway\"=\"level_crossing\"][!\"crossing:barrier\"]->.crossings;\n" +
+                "(.crossings; - .private;)->.public;\n" +
+                ".public;\n" +
+                "out;"
```
westnordost: Kotlin tip:
Kotlin supports multi line strings with `"""`, so you could write

```kotlin
private fun getOverpassQuery(bbox: BoundingBox) =
   OverpassQLUtil.getGlobalOverpassBBox(bbox) + """
   way["highway"]["access"~"^private|no$"];
   (._;>;);
   node(w) -> .private;
   node["railway"="level_crossing"][!"crossing:barrier"]->.crossings;
   (.crossings; - .private;)->.public;
   .public;
   out;
"""
```
Makes it better readable because you spare all the `"`, `\n`, `+` and esacping `\"`.

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
```diff
@@ -1,21 +1,50 @@
...
+        when {
+            answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.MISSING -> changes.add("crossing:barrier", "no")
+            answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.HALF -> changes.add("crossing:barrier", "half")
+            answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.DOUBLE_HALF -> changes.add("crossing:barrier", "double_half")
+            answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.FULL -> changes.add("crossing:barrier", "full")
+        }
+    }
 
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.add("crossing:barrier", answer)
```
westnordost: Actually, there is no reason to change anything here from the current behavior. 

But anyway, here is another Kotlin tip:
`if` and `when` statements are itself also statements, so you can write
`val x = if(a) 1 else if(c) 2 else 3`

So, here you could refactor this method to look like this:

```kotlin
override fun applyAnswerTo(answer: RailwayCrossingBarrierAnswer, changes: StringMapChangesBuilder) {
    changes.add("crossing:barrier", when {
        answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.MISSING -> "no"
        answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.HALF -> "half"
        answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.DOUBLE_HALF -> "double_half"
        answer.barrierType == AddRailwayCrossingBarrierForm.BarrierOption.FULL -> "full"
    })
}
```
And since `when` is just a super-powered Java `case`, you can further simplify to
```kotlin
override fun applyAnswerTo(answer: RailwayCrossingBarrierAnswer, changes: StringMapChangesBuilder) {
    changes.add("crossing:barrier", when(answer.barrierType) {
        AddRailwayCrossingBarrierForm.BarrierOption.MISSING -> "no"
        AddRailwayCrossingBarrierForm.BarrierOption.HALF -> "half"
        AddRailwayCrossingBarrierForm.BarrierOption.DOUBLE_HALF -> "double_half"
        AddRailwayCrossingBarrierForm.BarrierOption.FULL -> "full"
    })
}
```

But I am just telling you this because I assume that you may be interested in learning more Kotlin. In case not, ignore everything of this comment except my first sentence ;-)

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
westnordost: Also, couldn't you write out .public; or something like this? I thought this is possible

matkoniecz:
Thanks for a review, especially nice Kotlin-related things!

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrierForm.kt
```diff
@@ -4,18 +4,21 @@ import de.westnordost.streetcomplete.R
...
+        Item(BarrierOption.DOUBLE_HALF, R.drawable.ic_railway_crossing_double_half),
+        Item(BarrierOption.FULL, if (countryInfo.isLeftHandTraffic) R.drawable.ic_railway_crossing_full_l else R.drawable.ic_railway_crossing_full)
     )
 
     override val itemsPerRow = 4
 
-    override fun onClickOk(selectedItems: List<String>) {
-        applyAnswer(selectedItems.single())
-    }
+    enum class BarrierOption { MISSING, HALF, DOUBLE_HALF, FULL }
```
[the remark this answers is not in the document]
matkoniecz: That is result of me not understanding AImageListQuestAnswerFragment combined with a bug in query (I was not excluding entries where barrier type was tagged and was confused why solving quest is not resulting in editing OSM).

So I ended copying car wash structure at then fixing the real problem.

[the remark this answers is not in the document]
matkoniecz: And now that mess is gone!

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
matkoniecz: `.public out;` works - http://overpass-turbo.eu/s/HDo

Changed.

app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt
```diff
@@ -1,21 +1,41 @@
...
+        return overpassMapDataDao.getAndHandleQuota(getOverpassQuery(bbox), handler);
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? = null
+
     override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
         changes.add("crossing:barrier", answer)
     }
+
+    private fun getOverpassQuery(bbox: BoundingBox): String? {
```
westnordost: Here's a little nitpicking, no need to change it if you dont' want to: The method declares itself to return "maybe" a string (= a `String` or `null`) but looking at the code, it definitely always returns a `String`. So, you can remove the `?`.

This does not matter at the moment because `OverpassMapDataDao` is a Java class which thus doesn't know the difference between a `String` and `String?`.

matkoniecz: fixed!

westnordost:
I want to release the v11.0 in an hour or so. If you can guarantee that the overpass query is correct and working, I can merge it right away.

westnordost:
So did you check?

matkoniecz:
Yes, I just completed checking and it quest query works fine.

What happens during quest solving was not tested.

matkoniecz:
Though given that it is a minor detail it can wait without trouble for the next release.

westnordost:
Well the quest-solving did not change, so I can merge it

dhrisha:
www.dollar13.com
How to make money online
www.railwayjobss.in
All government vacancy

Issue #1354, opened 2019-03-14
Adding "Ghosts" (or missing shop names) for OSMUK Quarterly Project


atomoil:
### General

Affected tag(s) to be modified/added: [Name](https://wiki.openstreetmap.org/wiki/Key:name) on [Shops](https://wiki.openstreetmap.org/wiki/Shops)
Question asked: **What is the shop name?**
(ideally it would be **Is the shop name correct?** but I don't think that's possible)

I am one of the Directors of OpenStreetMap UK and a mobile app developer and we would love to allow mappers to use Street Complete for the [Quarterly Projects](https://osmuk.org/quarterly-projects/) that we organise. 

The next topic is "Ghosts" aka Closed or Renamed Shops and Amenities and so this quest suggestion is to enable that (and start adding quests to StreetComplete 😎). Robert Whittaker has a [great tool for showing Ghosts](https://osm.mathmos.net/ghosts/), but I wonder if we could make it broader for StreetComplete, and capture shop names as I don't _think_ there is a way to find only shop names that haven't been updated for a period of time.

I have StreetComplete downloaded and compiling in Android Studio, so can do the work to add it, although I might need some pointers as my speciality is iOS so I usually use Xcode (or React Native or Xamarin for cross platform (aka iOS & Android)), but I am very adaptable.

### Checklist
Checklist for quest suggestions (see guidelines):
- [✅] 🚧 To be added tag is established and has a useful purpose
- [✅ ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [✅] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [✅] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [✅] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

**Element selection:**
Based on looking at the tagFilters property, I _think_ it would be: 
`nodes, ways with ((amenity ~ post_office|pharmacy|bank|atm|bureau_de_change) or (shop)) and (!name and noname != yes)`
but any/all guidance or suggestions are more than welcome.

**Proposed GUI:**

****
```
  ^
[ What is the shop name? ]

          Shop Name:
 ________________________
|________________________|

```

ENT8R:
This is actually already implemented but not shown at all because there is no tag-to-name dictionary at the moment. See also #309

rugk:
But it is interesting to know there is such a project about it.

[...]

PR #1376, opened 2019-04-14
Add place name quest (fixes #309, unblocks #1354)

### LINKED ISSUE #309: Quest: POI without name - example with shops
hi,

can we please have places without names? I think many types of POI could benefit such quests, but depending on the type of POI, the name can be in a relation (group of buildings & stuff).

Here is an "easy" example with [shops](http://wiki.openstreetmap.org/wiki/Key:shop):
```
[out:json][timeout:25];
(
  node["shop"][!"name"]({{bbox}});
  way["shop"][!"name"]({{bbox}});
);
out skel;
```

I assume it should also be the same for [amenities](http://wiki.openstreetmap.org/wiki/Key:amenity), but it's a more difficult request that should be discussed. But if most pharmacy have a name (at least in France), a "bicycle_parking" won't.

thanks for this great application, you brought me back to cartography :)

westnordost:
Still todo:
- see if it is possible to somehow replace those escaped unicode characters in the localizations with normal characters (as the JSONs are already UTF-8)
- update readme etc. to link the osmnames repository
- update readme etc. to mention where the translation data is from (and how to contribute / report bugs)


rugk:
Hint: GitHub does not link/care about issue/PR numbers linked in the title of PRs (or issues, for what it's worth).

HolgerJeromin:
You are right :-) fixes #309, unblocks #1354

rugk:
Sorry for doing this off-topic discussion, but yeah, not even closing works, actually. It only works if/because the same is mentioned in the commit. However, in this case here (where the commit message is different) even that does not work.

PR #1379, opened 2019-04-18
New quest/shop names

atomoil:
so, I've managed to hack together a first start on the shop names quest. This is my first time writing Kotlin, and what's here is from looking at the StreetComplete source and deducing how to put a quest  together, so it's probably full of rookie mistakes and I am totally open to any and all feedback on how to improve this.

Some questions have arisen:
- is there anything I've missed?
- what's the best way to look at this quest within the app?
- where do the StreetComplete icons come from? They look rather awesome, and so I am inspired to match the quality, but I have absolutely no clue about generating Android vector drawables. Is there an existing editable source for the icons I can start from for example? (Note: for the sake of getting something up and running I have copied an existing icon)
- how do I integrate this with the osmnames / osmfeatures library (actually, I forked the repo a while back and have just rebased the changes onto my branch, so I can see there is some fresh work underway to integrate this, so I might be able to answer this myself, but any pointers are gratefully received).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
@@ -50,6 +50,7 @@ import de.westnordost.streetcomplete.quests.religion.AddReligionToWaysideShrine;
 import de.westnordost.streetcomplete.quests.localized_name.data.PutRoadNameSuggestionsHandler;
 import de.westnordost.streetcomplete.quests.localized_name.data.RoadNameSuggestionsDao;
 import de.westnordost.streetcomplete.quests.segregated.AddCyclewaySegregation;
+import de.westnordost.streetcomplete.quests.shop_name.AddShopName;
 import de.westnordost.streetcomplete.quests.sidewalk.AddSidewalk;
 import de.westnordost.streetcomplete.quests.surface.AddPathSurface;
 import de.westnordost.streetcomplete.quests.tactile_paving.AddTactilePavingBusStop;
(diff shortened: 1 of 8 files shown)


matkoniecz:
https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt may be useful

> what's the best way to look at this quest within the app?

Go to settings, go to to quest selection (first row in "advanced" section), move it to a top in quest list, possibly also disable other quests.

> where do the StreetComplete icons come from? They look rather awesome

Nearly all were made by @westnordost And yes, these icons are really great :)

There is already `R.drawable.ic_quest_label` in https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt

atomoil:
Actually, given that place_name is unblocked and moving forwards, I wonder if there is still a place for shop_name - @westnordost what do you think?

westnordost:
Oh, yes, sorry if there was a misunderstanding. 

The AddPlaceName quest actually includes also shops already, see @matkoniecz comment. When I wrote that  #1354 is unblocked, I meant that implementing this "ghosts" quest is now unblocked. So:
1. get list of names that closed/renamed from Robert Whittaker's website or a TBD api, search for them via Overpass
2. ask users what is the name of this ${place}, like in the `AddPlaceNameQuest`. The form at least could be reused from the `AddPlaceNameQuest `- please build your work on top of the branch `add-place-name`.

I am probably not available too much on the Easter weekend, just so so you know.

PR #1396, opened 2019-05-13
Add option to open location in another app

### LINKED ISSUE #190: Using geo intends (Open other map app on location)
Sometimes I find quests which are not solvable with this app. 
It would be helpful if I could open this location with the full featured editor vespucci. 
The easiest way is opening a geo intend which can open osmand and vespucci. 

The other way around would be to open street complete from osmand or osm homepage via geo intend.

ENT8R:
This PR adds to things:
- The option (located in the upper right menu) to open the current visible location in another application installed on the phone (e.g. Vespucci, Google Maps, Maps.Me, or any other android application that can handle a [Geo URI](https://en.wikipedia.org/wiki/Geo_URI_scheme))
- An intent handler for a Geo URI intent by another app (so you can open a specific location in StreetComplete using a Geo URI)

It fixes #190


app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -416,6 +451,21 @@ private void confirmUndo(final OsmQuest quest)
...
+
+				Uri uri = Uri.parse(String.format("geo:%f,%f?z=%f",
+					position.getLatitude(),
+					position.getLongitude(),
+					zoom
+				));
+
+				intent = new Intent(Intent.ACTION_VIEW, uri);
+				if (intent.resolveActivity(getPackageManager()) != null) {
+					startActivity(intent);
```
westnordost: What actually happens here? Will an activity selector be shown, or will the app previously selected as "primary geolocation app" be started?

Also, if there is no activity that takes a geo intent, and the Android system does not already display the lack of it in some way (showing an empty app selection screen?), perhaps a toast message should be shown instead.

app/src/main/java/de/westnordost/streetcomplete/tangram/MapFragment.java
```diff
@@ -654,6 +654,16 @@ public void setMapOrientation(float rotation, float tilt)
 		onMapOrientation(rotation, tilt);
 	}
 
+	public void setPosition(LngLat position)
```
westnordost: `getPosition` returns a `LatLon` (from osmapi) while this method accepts a `LngLat` (from tangram-es).

The `MapFragment` should use only one position data type on its interface. I tend to say that `LngLat` is a slightly better choice because it has more to do with tangram-stuff than OSM Api data classes.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -247,6 +252,36 @@ public void onServiceDisconnected(ComponentName className)
...
+			if ("geo".equals(data.getScheme()))
+			{
+				double latitude = 0;
+				double longitude = 0;
+				float zoom = -1;
+
+				String geoUriRegex = "^geo:(-?[0-9]*\\.?[0-9]+),(-?[0-9]*\\.?[0-9]+).*?(?:\\?z=([0-9]*\\.?[0-9]+))?";
+				Pattern pattern = Pattern.compile(geoUriRegex);
+				Matcher matcher = pattern.matcher(data.toString());
+				while (matcher.find())
```
westnordost: Why while?

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -247,6 +252,36 @@ public void onServiceDisconnected(ComponentName className)
...
+				{
+					latitude = Double.parseDouble(matcher.group(1));
+					longitude = Double.parseDouble(matcher.group(2));
+					if (matcher.group(3) != null) {
+						zoom = Float.valueOf(matcher.group(3));
+					}
```
westnordost: A faulty URL should be handled gracefully here. I'd wrap the parsing into a try-catch

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -247,6 +252,36 @@ public void onServiceDisconnected(ComponentName className)
...
+				}
+
+				mapFragment.setPosition(new LngLat(longitude,  latitude));
+				if (zoom != -1)
+				{
+					mapFragment.setZoom(zoom);
```
westnordost: Not sure what `zoom=.0` would do. In the worst case, some `IllegalArgumentException` (probably not). Still, I think checking for `zoom > 0` would be safer here.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -247,6 +252,36 @@ public void onServiceDisconnected(ComponentName className)
...
+				while (matcher.find())
+				{
+					latitude = Double.parseDouble(matcher.group(1));
+					longitude = Double.parseDouble(matcher.group(2));
+					if (matcher.group(3) != null) {
+						zoom = Float.valueOf(matcher.group(3));
+					}
+				}
+
+				mapFragment.setPosition(new LngLat(longitude,  latitude));
```
westnordost: The URL could contain invalid coordinates like 99999.9999. This should also be handled gracefully before feeding it to the map fragment. At least `LatLon` will throw an exception when trying to initialize it with invalid values. Not sure if `LngLat` also does this.

westnordost: Also, if the geo intent is not in the form of the regex at all, this method will just use LngLat(0,0) with zoom=1. Not sure if this is desired or an error message should be shown.

app/src/main/java/de/westnordost/streetcomplete/tangram/MapFragment.java
westnordost: Also, the setter and the getter should be in the same class. Probably in `MapFragment` because this is nothing specific to quests.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: > Will an activity selector be shown, or will the app previously selected as "primary geolocation app" be started?

Both. If the user selected a default app to handle a Geo URI before, this app opens without the activity selector. Otherwise you can select an application from the list.

> Also, if there is no activity that takes a geo intent

Right now, nothing will happen at all... Is it better to display a toast message or remove the item from the options menu if the intent can't be handled by any other activity?

westnordost: I think it is better to display a short toast message in that case than hide the option, e.g. "No map application installed" or similar.

app/src/main/java/de/westnordost/streetcomplete/tangram/MapFragment.java
westnordost: Are you sure? I checked yesterday and I think I remember it only being used in two instances. A LngLat can easily be converted to a LatLon, there should be an utility class for that.

ENT8R: Oh... true. It is only used by the `MainActivity` and the `QuestController`. I thought it would be more 🤷‍♂

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: By using [`find()`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#find--) it is also possible to extract the result from a given substring. I agree, that this doesn't make that much sense in our use case, so the following is enough:

```suggestion
				if (matcher.matches())
				{
				        latitude = Double.parseDouble(matcher.group(1));
					longitude = Double.parseDouble(matcher.group(2));
					if (matcher.group(3) != null) {
						zoom = Float.valueOf(matcher.group(3));
					}
				}
```

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
rugk: Maybe "No _other_ map application installed" (would be self-contradictory otherwise, as SC is such a mapping app).

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: > Not sure if `LngLat` also does this.

No, `LngLat` doesn't do this... Maybe `LatLon` is the better alternative? 😄 I can also revert https://github.com/westnordost/StreetComplete/pull/1396/commits/107defc64245f8419e659b7c7ec22dcbce98aea7

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: No other map application installed is now the text of the toast

westnordost: Wait, right, so if StreetComplete can interpret a geo URI, then the case that no application exists that will take the intent does not exist, or is the source activity automatically excluded there?

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
westnordost: You can just do the validation in-place, it's a simple >= && <= check anyway.

It is just important to properly validate the incoming "user data" here, also for security reasons, as it may in fact be *any* string.

app/src/main/java/de/westnordost/streetcomplete/tangram/MapFragment.java
```diff
@@ -654,16 +654,44 @@ public void setMapOrientation(float rotation, float tilt)
...
-	public float getRotation()
+	public LngLat getPositionAt(PointF pointF)
 	{
-		return controller != null ? controller.getRotation() : 0;
+		LngLat pos = controller.screenPositionToLngLat(pointF);
+		if(pos == null) return null;
+		return pos;
+	}
+
+	public PointF getPointOf(LatLon pos)
```
westnordost: Well I guess then for consistency, this should also accept a LngLat

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -247,6 +253,34 @@ public void onServiceDisconnected(ComponentName className)
...
 			questController.deleteOld();
 		}
+
+		Intent intent = getIntent();
+		Uri data = intent.getData();
+		if (Intent.ACTION_VIEW.equals(intent.getAction()) && data != null)
+		{
+			if ("geo".equals(data.getScheme()))
+			{
+				String geoUriRegex = "geo:(-?[0-9]*\\.?[0-9]+),(-?[0-9]*\\.?[0-9]+).*?(?:\\?z=([0-9]*\\.?[0-9]+))?";
```
westnordost:
```suggestion
				String geoUriRegex = "^geo:(-?[0-9]*\\.?[0-9]+),(-?[0-9]*\\.?[0-9]+).*?(?:\\?z=([0-9]*\\.?[0-9]+))?";
```
Ensures that `matcher.find()` will only ever find one match because `^` matches to the start of the string.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: It is excluded. But if you have both versions installed (debug and release) then you can of course switch between both... :smile:

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: In https://github.com/westnordost/StreetComplete/pull/1396/commits/149eec72d4ba631b1ae24edefcff39a5e4e8f420 I planned to replace `matcher.find()` with `matcher.matches()` because this makes more sense for our use case. But I only removed the `^` at the beginning of the regex (which is not needed when using [`matcher.matches()`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#matches--)) and did not replace the method call...

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: Mmhh... But with a wrong URL, the regex match will fail too, so if [`matcher.matches()`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#matches--) returns false, nothing else is done...

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
ENT8R: I added the extra validation step now but I would say that it can not be simply **any** string. The regex only searches for digits and no words/strings/characters: https://regex101.com/r/ZVp8Zf/1

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
westnordost: Okay, but, for your information:

`matches` matches the whole string against the RegEx, while `find` only searches if a part of the input string matches the RegEx.

The Geo URI scheme is standardized in RFC 5870, see [wikipedia](https://en.wikipedia.org/wiki/Geo_URI_scheme), a URI could look like this:
`geo:37.786971,-122.399677;u=35`.

The `?z=` zoom is not part of the standard so I can't say where it should be expected within the Geo URI. Your RegEx expects it to be at the utter end of the Geo URI, so `geo:37.786971,-122.399677?z=15;u=35` wouldn't match your RegEx.

But I think this is fine because all the examples on the Internet assume those URL query-type parameters like zoom and query to be at the end.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
westnordost: Right, I thought maybe it would be possible to smuggle something like `0.1.234` in there, but the regex looks sound.

westnordost:
Merged, thank you!

cyclingcat:
Could it be that the translation for thie new menu item is missing? My Android 5.1 phone shows this item in English:

Screenshot_2019-06-15-11-19-13

The cycling cat

ENT8R:
The string was translated for the first time one week ago (according to POEditor), so it was not possible for the translation to be added to version 12.1 which was released more than two weeks ago...
BTW: if you see something like that in the future and you wonder whether a string has been translated or not, you might as well also look by yourself into the translation editor: https://poeditor.com/join/project/IE4GC127Ki

PR #1432, opened 2019-06-19
add type of a ferry quest [ready to merge]

### LINKED ISSUE #39: Add Type of Ferry quest
The app should probably filter the number of markers it shows in the map view and prioritize quests for which the added value is high or the frequency in osm is low. Conceptually similar to [tf-idf](https://en.wikipedia.org/wiki/Tf%E2%80%93idf).

---

To be more specific here: consider the case of ferries. Ferries are tagged `route=ferry` and a `duration` tag describes how long the route takes. Without a `duration` tag you can only guess the duration.

The ferry tag is rare and the duration tags adds high value - a marker for a missing ferry duration tag should probably be prioritized among many opening hour, building level and roof shape markers.

http://wiki.openstreetmap.org/wiki/Tag:route%3Dferry

matkoniecz:
fixes #39

Note that long ferries are excluded, like any long ways.

In this case quest will collect information that sometimes can be already derived from other data. But even if it collects data that changes routing only in 1% of solved quests it is still worth tagging it explicitly as incorrect routing caused by a poor ferry data may cause massive detours.

I looked at other access values and most of them are mistakes, foot=designated where foot=yes works equally well or extremely rare situations that can be handled with a note.

I am considering more complicated motor vehicle quest to handle cases where only some types of vehicles are allowed (say, only motorcycles). For now I trust user to switch to creating notes in such cases.

-----


app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt
```diff
@@ -0,0 +1,31 @@
...
+    override val icon = R.drawable.ic_quest_ferry
+    override val hasMarkersAtEnds = true
+
+    override fun getTitle(tags: Map<String, String>): Int {
+        val hasName = tags.containsKey("name")
+        return if (hasName)
+            R.string.quest_ferry_motor_vehicle_name_title
+        else
+            R.string.quest_ferry_motor_vehicle_title
+    }
```
westnordost: FYI you can also write the above as this in Kotlin. I don't mind either way though
```kotlin
        if (tags.containsKey("name"))   R.string.quest_ferry_motor_vehicle_name_title
        else                            R.string.quest_ferry_motor_vehicle_title
```

app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt
```diff
@@ -0,0 +1,31 @@
...
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+
+
+class AddFerryAccessMotorVehicle(o: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(o) {
+
+    override val tagFilters = "ways, relations with route=ferry and !motor_vehicle"
+    override val commitMessage = "Specify ferry access"
```
westnordost:
```suggestion
    override val commitMessage = "Specify ferry access for motor vehicles"
```

app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessPedestrian.kt
```diff
@@ -0,0 +1,29 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+class AddFerryAccessPedestrian(o: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(o) {
+
+    override val tagFilters = "ways, relations with route=ferry and !foot"
+    override val commitMessage = "Specify ferry access"
```
westnordost:
```suggestion
    override val commitMessage = "Specify ferry access for pedestrians"
```

app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessPedestrian.kt
```diff
@@ -0,0 +1,29 @@
...
+    override val icon = R.drawable.ic_quest_ferry
```
westnordost: A different icon (same icon but with pedestrian-blue background) would be nice.

app/src/main/res/values/strings.xml
```diff
@@ -655,4 +655,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="label_housenumber">house number</string>
     <string name="label_blocknumber">block number</string>
     <string name="quest_busStopShelter_covered">Whole stop is covered</string>
+    <string name="quest_ferry_pedestrian_name_title">Is %s accessible to pedestrians?</string>
+    <string name="quest_ferry_pedestrian_title">Is this ferry accessible to pedestrians?</string>
+    <string name="quest_ferry_motor_vehicle_name_title">Is %s accessible to motor vehicles?</string>
+    <string name="quest_ferry_motor_vehicle_title">Is this ferry accessible to motor vehicles?</string>
```
westnordost: Not sure if "accessible" is the right word for it. Rather "does X transport...?" or "is X for ...?" or something like that?

matkoniecz: I like "does X transport...?" (it will fail for private ferries without public access but that should be ridiculously rare and I expect original mapper to tag it explicitly)

app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt
matkoniecz: I do not feel that it is more readable. Rubocop linter keeps trying to use similar structure in Ruby, and I am also not fan of it.

But thanks for an info!

app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessPedestrian.kt
matkoniecz: Should I copy `StreetComplete/app/src/main/res/drawable/ic_quest_ferry.xml` and modify background color or should I modify some generator of svg icons?

I see that in some commits also `StreetComplete/res/quest_icons.svg` was changed but it is unclear to me what is original source and how generation is triggered.

I found `export_svgs.py` script but it has note 

> # This script does not work, currently.

westnordost: Yes, you can simply copy ic_quest_ferry

app/src/main/res/values/strings.xml
goldfndr: I suggest "Are __ permitted to ride on %s?"

app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessPedestrian.kt
matkoniecz: Done!

app/src/main/res/drawable/ic_quest_ferry.xml
```diff
@@ -69,7 +69,7 @@
         android:fillColor="#ccd6dd"/>
     <path
         android:pathData="m1.19,23.38c1.36,3.39 3.82,6.36 7.23,8.33 2.05,1.19 4.25,1.89 6.47,2.15l3.99,0c5.12,-0.61 9.9,-3.54 12.68,-8.35 0.4,-0.69 0.74,-1.41 1.04,-2.13 -2.65,-0.52 -5.22,-0.42 -7.85,0 -2.68,0.43 -5.72,0.57 -7.85,0 -4.76,-1.28 -6.06,0.51 -7.85,0C7.55,22.95 5.18,20.92 4.51,23.27Z"
-        android:fillColor="#529add"/>
+        android:fillColor="#3e74c9"/>
```
westnordost: The car one should have the yellow background, like for the maxspeed quest etc.
The pedestrian one should be the same but have the blue background like for the pedestrian crossing quest etc

matkoniecz: I am pretty sure that this is a change of water color to keep it consistent - see https://github.com/westnordost/StreetComplete/pull/1432/commits/04d1d0b2370194e816c512cefcb5eeebd6ec0c78

> to make water distinguishable from blue background, and keep water color the same on both
> I made water color a bit darker 

I copied blue background from other pedestrian quest for pedestrian variant and it turned out to be the same shade as water.

So I made water darker on both icons.

westnordost: Ah okay, the diff looked as if you changed that background color. Then, all is fine, thank you!

matkoniecz:
I am still thinking about

> I suggest "Are __ permitted to ride on %s?"

proposed by @goldfndr

> "Are motor vehicles permitted to ride on %s?"

seems a bit weird to me, but maybe it is actually better

westnordost:
Sounds a little weird. One does not "ride" on a ferry.

goldfndr:
Perhaps it's just [en-US](https://www.sunset.com/travel/northwest/the-northwests-five-best-ferry-rides)? Or [elsewhere](https://www.cntraveler.com/galleries/2013-08-25/photos-istanbul-bosphorus-ferries)? It's definitely "ride" in US:WA. ([WSDOT has a fairly extensive ferry system locally](https://www.wsdot.com/ferries/schedule/Default.aspx); note that a paragraph starts with "Riders".)

westnordost:
Ok then it looks like I am wrong. So one can ride a ferry, like a train, but does a car also ride that ferry then? Likewise, a train that also transports cars, do these cars ride on the train with you?

Or is this wording limited to people?

matkoniecz:
For me 

> "Are pedestrians permitted to ride on %s?"

seems perfectly fine. But 

> "Are motor vehicles permitted to ride on %s?"

also seems weird to me. @goldfndr - are you sure that this is OK?

rugk:
Remember that @goldfndr is the only native English speaker in this issue here. :smiley: With everything else, the translators then have to deal with these kind of issues… :wink:

matkoniecz:
> Remember that @goldfndr is the only native English speaker in this issue here

I know, but I wanted to ask explicitly about 

> "Are motor vehicles permitted to ride on %s?"

variant.

goldfndr:
I'm thinking that, rather than the awkward usage of "riding" for passenger vehicles (which, upon reflection, sounds a bit odd to me too), @matkoniecz's current wording is fine. (I was thinking [boarding](https://88844ferry.com/default.aspx) but that makes the question much longer.) But "this ferry" needs to be added to the strings that don't include "%s", as the questions are incomplete otherwise.

Oh, lest I forget again, StreetComplete tends to use the word "allowed" rather than "permitted". My apologies.

matkoniecz:
> My apologies.

Nothing wrong happened! Review of wording and confirmation that it is OK is very useful. More than enough to outweigh slight confusion by proposing alternative wording and withdrawing it on a second thought.

app/src/main/res/values/strings.xml
matkoniecz: Looking at POEditor (I was translating latest strings to Polish) I noticed that while

> Does %FERRY_NAME transport pedestrians?

is OK, but variants for nameless ferries

> Does transport pedestrians?

and

> Does transport motor vehicles?

seem broken. It should be probably

> Does ferry transport pedestrians?

and

> Does ferry transport motor vehicles?

westnordost: *this ferry route

matkoniecz: I updated Engrish-fixing PR to use this form.

app/src/main/res/values/strings.xml
```diff
@@ -653,4 +653,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="label_housenumber">house number</string>
     <string name="label_blocknumber">block number</string>
     <string name="quest_busStopShelter_covered">Whole stop is covered</string>
+    <string name="quest_ferry_pedestrian_name_title">Does %s transport pedestrians?</string>
+    <string name="quest_ferry_pedestrian_title">Does transport pedestrians?</string>
```
rugk: Is not there a "…" or so missing? Grammatically the subject is missing here.

rugk: Or possibly "Does this transport…"?

app/src/main/res/drawable/ic_quest_ferry_pedestrian.xml
```diff
@@ -0,0 +1,77 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
rugk: As per [this](https://wiki.openstreetmap.org/w/index.php?title=StreetComplete%2FQuests&type=revision&diff=1882295&oldid=1874863) we also need the SVG version of this icon. (`ic_quest_ferry_pedestrian`)

Respectively which color needs to be changed? To which color code? (AFAIK the background color.)

/cc @goldfndr

rugk: Generally, the SVG version is outdated. Please always also change the SVG versions, not only the Android scalables.

rugk: Okay, guessed and tinkled a little and I hope this is the correct icon: https://github.com/rugk/streetcomplete-quest-svgs/commit/91f65b7ef49b571b3c0620df4dd94d39c53dfd3d

app/src/main/res/values/strings.xml
westnordost: You are late to the party. This is long solved.

app/src/main/res/drawable/ic_quest_ferry_pedestrian.xml
westnordost: Yes, you are right. Thank you!

PR #1436, opened 2019-06-21
Improve issue templates

rugk:
* [x] automatically assigns "bug" and "enhancement" labels to issues chosen as such
* [x] wrapped some instructions in comments, so they won't appear in issues

Both done in separate commits, so you can revert one, if you do not like it. (or cherry-pick)

Also, I could obviously also auto-assign the "new quest", but it seems you are using this label to as an "this issue is approved to be implemented as a new quest".
So as an idea, you could, however:
* introduce a tag like "proposed-quest" that gets automatically assigned, so you can then unlabel it and re-label it as "new quest" if you approve the idea

Just an example idea…


rugk:
Uhm, sorry for these strange merge commits, my fork's git history is likely a little screwed. :innocent:

westnordost:
Looks good, thanks!

rugk:
What about the last idea?

> introduce a tag like "proposed-quest" that gets automatically assigned, so you can then unlabel it and re-label it as "new quest" if you approve the idea

westnordost:
I'd rather not, other categories of tickets must be discussed first as well before I label them.

rugk:
Well… then… you did notice what change you'e merged? (it's always good to read PR descriptions :stuck_out_tongue_winking_eye:)

> automatically assigns "bug" and "enhancement" labels to issues chosen as such

Anyway, you can of course always re-assign labels, so IMHO if there is for some time a wrong label, I would not mind.

westnordost:
Right, I overread this one. Thank you

PR #1437, opened 2019-06-22
increase bus stop name importance

matkoniecz:
also used by many data consumers

-----

BTW, I plan to propose some other changes in quest order. One of repeated misleading impressions of StreetComplete is "it asks only about street surface what is boring" and I plan on proposing some changes in default order to reduce this issue.

But I spotted surprisingly low position of bus stop name and decided to submit it separately.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
@@ -94,6 +94,7 @@ public class QuestModule
 				new AddRoadName(o, roadNameSuggestionsDao, putRoadNameSuggestionsHandler),
 				new AddPlaceName(o, featureDictionaryFuture),
 				new AddOneway(o, trafficFlowSegmentsDao, trafficFlowDao),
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
```diff
@@ -93,6 +93,7 @@
 				// ↓ 2. important data that is used by many data consumers
 				new AddRoadName(o, roadNameSuggestionsDao, putRoadNameSuggestionsHandler),
 				new AddPlaceName(o, featureDictionaryFuture),
+				new AddBusStopName(o),
```
westnordost: I find AddOneway more important than bus stop name, but otherwise it is fine

matkoniecz: Changed as requested. Given rarity of oneway quest it makes little difference, though I am curious what makes bus stop name less important than AddPlaceName.

westnordost:
OKay, thanks!

PR #1441, opened 2019-06-25
promote some quests before surface quest 

matkoniecz:
according to my attempt at UX testing, some spotted reviews and interviewing some users
large number of surface quests may overhelm users, confuse them and result in failure
to discover other available quests

These two quests are more interesting to a mapper, except rare situation
will not be as repetitive as surface quest and except rare situation their number
of occurences will be far more limited.

This is intended to provide more interesting mapping for users.

Main negative is that in some places no quests of this type will be available, so
app will waste two Overpass requests and do not get quests

-----


matkoniecz:
I unfortunately failed to save links to this comments, but I may try to find again some of them.

westnordost:
'Tis reasonable 😃

PR #1444, opened 2019-06-25
Quest prority - reordering the list

matkoniecz:
This is final part of proposed priority changes. I found two quests that are both more important and more interesting than surface quest for roads and increased their importance.

Also, noticed that some quests with actually used data are below quests that add data that is currently unused.

-------

This should be a final PR in series, main goal was to bump quests both more interesting and more important than surface quest a bit higher. Now I see nothing fitting.

-----


goldfndr:
Way Lit is also used by OsmAnd if "Street lighting" is enabled. (Configure map, Map rendering, Details)

L101 should be `new AddReligionToPlaceOfWorship(o), // icons on maps are different - OSM Carto, mapy.cz, OsmAnd, Sputnik etc` (plural icons, OsmAnd capitalization)

matkoniecz:
Both changed, thanks!

westnordost:
All is reasonable, thank you for adding the comments for verifiability! Looks good!

PR #1445, opened 2019-06-27
Add hints for opening hours differing by day of week [WIP]

matkoniecz:
I made a small UX test. Three people attempted opening hours hours quest for object where opening hours differed by day. All three required hint to proceed, they were either confused from start or selected all days when shop was open and failed to answer it.

In comparison one person who had trouble with building level quest was happy with explanation in "differ per building part" hint that she found by herself.

This hint is not solving the entire problem. There are still people stuck after selecting days but before realizing what should be done, that never tried using "other answers...". But it should help at least some of people.

-----

I also added `showHint` to `AbstractQuestAnswerFragment.kt` to reduce code duplication.

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
@@ -15,6 +15,7 @@ import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import android.widget.Button
 import android.widget.LinearLayout
 import android.widget.PopupMenu
+import androidx.annotation.StringRes
 
 import java.lang.ref.WeakReference
 import java.util.Locale
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -51,7 +51,7 @@
     <string name="quest_openingHours_chooseMonthsTitle">"Select months"</string>
     <string name="quest_openingHours_chooseWeekdaysTitle">"Select group of weekdays with the same opening hours"</string>
     <string name="quest_openingHours_24_7_confirmation">"In other words, is it open any time, every day?"</string>
-    <string name="quest_openingHours_differs_by_day_of_week">You can use the \"ADD OPENING HOURS\" button multiple times, adding separate groups of days with the same opening hours.</string>
+    <string name="quest_openingHours_differs_by_day_of_week">You can use the \"ADD OPENING HOURS\" button multiple times, adding separate groups of days with the same opening hours every time.</string>
     <string name="quest_openingHours_answer_differs_by_day_of_week">Differs by day of week…</string>
     <string name="quest_noteDiscussion_title">"Can you contribute anything to this note?"</string>
     <string name="quest_noteDiscussion_no">"No, hide"</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -51,6 +51,8 @@
     <string name="quest_openingHours_chooseMonthsTitle">"Select months"</string>
     <string name="quest_openingHours_chooseWeekdaysTitle">"Select weekdays"</string>
     <string name="quest_openingHours_24_7_confirmation">"In other words, is it open any time, every day?"</string>
+    <string name="quest_openingHours_differs_by_day_of_week">Use the \"ADD OPENING HOURS\" button.\nSelect group of days with the same opening hours.\nFill opening hours.\nUse again \"ADD OPENING HOURS\"\nFill opening hours for the next group of days.\n\nRepeat if needed\n\nOnce completed use \"OK\" button.</string>
+    <string name="quest_openingHours_answer_differs_by_day_of_week">Differs by day of week…</string>
     <string name="quest_noteDiscussion_title">"Can you contribute anything to this note?"</string>
     <string name="quest_noteDiscussion_no">"No, hide"</string>
     <string name="quest_noteDiscussion_comment2">"— %1$s, %2$s"</string>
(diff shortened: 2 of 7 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
```diff
@@ -275,6 +276,15 @@ abstract class AbstractQuestAnswerFragment<T> : AbstractBottomSheetFragment() {
...
     }
+
+
+    fun showHint(@StringRes messageId: Int){
+        activity?.let { AlertDialog.Builder(it)
+                .setMessage(messageId)
+                .setPositiveButton(android.R.string.ok, null)
+                .show()
+        }
+    }
```
westnordost: Good idea to reuse this snippet. However, showing a hint is nothing specific to AbstractQuestAnswerFragment.

Better create an extension function on `Context` and put it somewhere in the ktx package directory.

```
fun Context.showHint(@StringRes messageId: Int){
   AlertDialog.Builder(this)
        .setMessage(messageId)
        .setPositiveButton(android.R.string.ok, null)
        .show()
}
```
And use like `activity?.showHint(...)`

app/src/main/res/values/strings.xml
```diff
@@ -51,6 +51,8 @@
     <string name="quest_openingHours_chooseMonthsTitle">"Select months"</string>
     <string name="quest_openingHours_chooseWeekdaysTitle">"Select weekdays"</string>
     <string name="quest_openingHours_24_7_confirmation">"In other words, is it open any time, every day?"</string>
+    <string name="quest_openingHours_differs_by_day_of_week">Use the \"ADD OPENING HOURS\" button.\nSelect group of days with the same opening hours.\nFill opening hours.\nUse again \"ADD OPENING HOURS\"\nFill opening hours for the next group of days.\n\nRepeat if needed\n\nOnce completed use \"OK\" button.</string>
```
westnordost: Sounds a bit dry / unfriendly / complex.

matkoniecz: Yeah, I am not happy about it. I hope that someone will propose a better version.

matkoniecz: Hmm. Maybe

goldfndr: How about changing "Select weekdays" to "Select weekdays with identical hours"?

matkoniecz: @goldfndr Great find, I completely missed it. I went with `Select group of weekdays with the same opening hours` - is it actually better, or is `Select weekdays with identical hours` preferable? I also reduced hint from dry and unfriendly to `You can use the "ADD OPENING HOURS" button multiple times, adding separate groups of days with the same opening hours every time.` what seems to be better.

goldfndr: Your more verbose title is better. My only concern was whether the dialog will still fit on the screen, especially after translations, but if that's nothing to worry about…

In your hint, I'd change the portion after the comma to `adding a separate group of days with the same opening hours each time.` to make it clear that once or a few times might be adequate, to not overwhelm.

Is it possible to place an icon near (far right of) the title (e.g. a circled question mark) that would bring up the hint? That could reduce the need for verbosity.

matkoniecz: >  My only concern was whether the dialog will still fit on the screen, especially after translations

app/src/main/res/values/strings.xml
```diff
@@ -49,8 +49,10 @@
     <string name="quest_openingHours_add_hours">"Add hours"</string>
     <string name="quest_openingHours_add_weekdays">"Add weekdays"</string>
     <string name="quest_openingHours_chooseMonthsTitle">"Select months"</string>
-    <string name="quest_openingHours_chooseWeekdaysTitle">"Select weekdays"</string>
+    <string name="quest_openingHours_chooseWeekdaysTitle">"Select group of weekdays with the same opening hours"</string>
```
westnordost: Does this button still fit on the screen on small (4.3") screens?

westnordost: If it does not fit or potentially does not fit in other languages, and the text does not wrap but is cut off, you should be able to change the style of the title bar by modifying `DialogWindowTitle.Bubble` in `styles.xml` (`android:maxLines`)

matkoniecz: Not sure. I tried one handed mode to simulate this, but it is scaling screen.

It is anyway in WIP for related reason as translations of just 120% length are getting cutoff on 5.0" screen.

westnordost: You can use an emulator or just set a longer test and see if it wraps. If it wraps, it is okay.

matkoniecz: It gets cutoff with "...".

matkoniecz: > you should be able to change the style of the title bar by modifying DialogWindowTitle.Bubble in styles.xml (android:maxLines)

Unfortunately, with

```
    <style name="DialogWindowTitle.Bubble" parent="android:Widget">
        <item name="android:textAppearance">@style/TextAppearance.Title</item>
        <item name="android:paddingBottom">8dp</item>
        <item name="android:maxLines">3</item>
    </style>
```

it is cutoff at 5.0" screen as follows:

`Select group of weekdays with the same opening hoursSelect group of ...`

I thought about switching from PopupMenu to full screen menu but first I want to try https://stackoverflow.com/a/48161007/4130619

westnordost: What does the PopupMenu have to do with this? Isn't this about the title text of the weekday picker dialog?

westnordost:
Obsolete now, the UI changed so that one can also specify "Add hours" to an existing day.

Issue #178, opened 2017-05-04
Swipe/encouraging/information screens


rugk:
Similar to openstreetmap.org's iD editor tutorial, but here it could just let the user answer some types of example quests, which will not be submitted to OpenStreetMap, of course…
This way users can get to know all (or the majority) of the quests types and see how (easy) it is to solve them as sometimes users may be in areas where there are literally only house (floor) quests they could be disappointed at first.

westnordost:
Then they will be disappointed after the tutorial when they realize there are only house level quests :-) 

I think this is the wrong angle.

Don't forget this is an early version, there will be more different quests in the future, and also more important ones.



On 4 May 2017 09:14:15 CEST, rugk <notifications@github.com> wrote:
>Similar to openstreetmap.org's iD editor tutorial, but here it could
>just let the user answer some types of example quests, which will not
>be submitted to OpenStreetMap, of course…
>This way users can get to know all (or the majority) of the quests
>types and see how (easy) it is to solve them as sometimes users may be
>in areas where there are literally only house (floor) quests they could
>be disappointed at first.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/westnordost/StreetComplete/issues/178

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

rugk:
Mhh, maybe, but at least they know there are other quests.
Anyway a tutorial could just be a general introduction for users.

callis:
Right now the tasks itself are straight forward do not require explanation

Nevertheless, a short introduction might be beneficial and could introduce the most relevant aspects of the code of conduct. https://wiki.openstreetmap.org/wiki/Good_practice

Topics like: Map what's on the ground, Don't remove "tags/edit things" that you don't understand, if in doubt leave it out, ...

westnordost:
You can't remove anything with this app. What do you suppose should be the content of the code of conduct as it may be presented here?

On 4 May 2017 16:38:44 CEST, callis <notifications@github.com> wrote:
>Right now the tasks itself are straight forward do not require
>explanation
>
>Nevertheless, a short introduction might be beneficial and could
>introduce the most relevant aspects of the code of conduct.
>https://wiki.openstreetmap.org/wiki/Good_practice
>
>Topics like: Map what's on the ground, Don't remove "tags/edit things"
>that you don't understand, if in doubt leave it out, ...
>
[...]

PR #1446, opened 2019-06-27
do not attack new user with request to register/login so early

matkoniecz:
During UX testing it turned out that starting is really confusing for an user actually new to OSM

She needed to (everything listed in full, as it is intended to solve a part of a larger problem)

- install application
- grant permissions (works great)
- wait for quests to load (this is not obvious that it is necessary, #178 exists with a good idea for solving this)
- figure out app interface
- figure out interface of the first quest (for bonus points - first one turned out to be opening hours quest, that maybe should be hidden for green users with <10 stars)
- solved quest
- immediately login window appeared
- it was necessary to register, with entire form filling + ToS + mail confirmation drudgery
- after registration app was stuck waiting for login, it was necessary to cancel this and request login again, login spinner initially refused to act (not sure is it user/phone error or an app bug)
- now it was necessary to login

She told me that if she would be trying app on her own there would be no chance at all to pass through all this hoops at the very start.

All of that to solve first quest. I am pretty sure that delaying nag screen about registration would be a good idea as it would not harm more experienced users, they will make more than 7 edits and login a bit later. It would mean that edits by people who made six edits and less are lost, but they would probably be lost on going through registration process.

And it would help some newbie users that would be too confused by a current process, but would be OK with more gradual ramping up.

-----

Implementation itself is more suspect as current AnswersCounter is about interface only, in fact it is giving answer reduced by one as it is increased only after animation completes (so login window is triggered on 7th edit).

It can be solved by having separate `unsynced` and `displayedUnsynced`, ignoring disrepancy or not mixing displayed statistics and statistics infuencing UI behavior (I have some additional ideas for special handling for newbie mappers).

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -496,6 +496,7 @@ public class MainActivity extends AppCompatActivity implements
 	@UiThread private void requestOAuthorized()
 	{
 		if(dontShowRequestAuthorizationAgain) return;
+		if(answersCounter.waitingForUpload() <= 5) return;
 
 		View inner = LayoutInflater.from(this).inflate(
 				R.layout.dialog_authorize_now, null, false);
--- a/app/src/main/java/de/westnordost/streetcomplete/statistics/AnswersCounter.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/statistics/AnswersCounter.java
@@ -86,6 +86,10 @@ public class AnswersCounter
 		updateTexts();
 	}
 
+	public Integer waitingForUpload(){
+		return unsynced;
+	}
+
 	@SuppressLint("StaticFieldLeak") public void update()
 	{
 		new AsyncTask<Void, Void, Void>()


rugk:
If you do so, you should possibly re-work the UI a little to make it obvious to the user **that everything is lost** when they don't login?

matkoniecz:
It would change outcome only for people making less than 7 edits and it is already covered by

> It would mean that edits by people who made six edits and less are lost, but they would probably be lost on going through registration process.

westnordost:
This is reasonable. 

> after registration app was stuck waiting for login, it was necessary to cancel this and request login again, login spinner initially refused to act (not sure is it user/phone error or an app bug)

...is worth investigating. If it is stuck, it means that the app is not called-back from the browser. What browser was she using?

matkoniecz:
> What browser was she using?

Chrome, but she went through a bit chaotic switching between StreetComplete, Gmail, Chrome, went to desktop once or twice between that...

During registration OSM Website was also quite broken so I suspect that there may be something wrong with her phone/browser.

I plan on retesting path for user who goes to register from browser launched during authorization, including wonky OSM website.

PR #1448, opened 2019-06-28
rework settings layout

matkoniecz:
I reworked setting layout based on what people actually want to change. My testing has a small sample size. But for now basically all people were interested in quest list - at least viewing it. Some also in changing. Some even started given unprompted advice to make it far more prominent.

So it is moved higher while some settings much less likely/useful to change were downgraded into advanced settings.

in UX testing multiple people that reached settings menu declared that quest selection should not be hidden in advanced settings

I also shown quest list to some people after UX testing session and reactions where mix of
- why I was not aware that X is disabled? I am very interested in this topic.
- it is interesting (about specific quest)
- it is amazing (about quest variety)
- I really like icon X (often for the underground buildings)
- I was not aware that there are some many question types
- it should be displayed on first startup (I am pretty sure that it would have more negatives than positives).

Also some people using SC declared that it should be less hidden.

So it is promoted higher, out of advanced settings, visible without scrolling settings list.

I also downgraded cache size into advanced, in normal operation there is no need to change it.

I removed separate single element theme and cache group as unnecessary.

-----

Why I keep writing books in commit messages? Is full story behind such change is useful. And more useful than something like "reworked settings moving ones more likely to be used to the top, quest list was especially wanted according to the UX testing"? I somehow made XXL commit message again, I am planning to strongly avoid doing it with the next one, but maybe it is useful.

-----

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -142,7 +142,6 @@
     <string name="quest_sport_ice_hockey">"Ice hockey"</string>
     <string name="quest_sport_netball">"Netball"</string>
     <string name="quest_sport_rugby">"Rugby"</string>
-    <string name="pref_category_cache">"Cache"</string>
     <string name="pref_title_sync">"Auto-sync"</string>
     <string name="map_btn_zoom_out">"Zoom out"</string>
     <string name="map_btn_zoom_in">"Zoom in"</string>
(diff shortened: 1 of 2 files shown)


rugk:
For visual changes, a screenshot in the PR description would be helpful though. :smile:

westnordost:
> Why I keep writing books in commit messages? Is full story behind such change is useful. And more useful than something like "reworked settings moving ones more likely to be used to the top, quest list was especially wanted according to the UX testing"? I somehow made XXL commit message again, I am planning to strongly avoid doing it with the next one, but maybe it is useful.

I see no reason why a git commit message must be short. For documentation, it is a very good idea even.

westnordost:
The changes make sense, thank you!

PR #1450, opened 2019-06-28
 make questions signs more visible in the sidewalk and cycleway quest [ready for review]

matkoniecz:
only one tester was solving this quest type (as the first ever in her case)
and it went like this

- select this quest
- see road with no sidewalk
- select no sidewalk
- tap outside quest

More visible question signs may help to avoid this mistake as it will be more obvious that both sides should be filled to solve this. Current ones are not clearly visible for someone with poor eyesight or in a glaring sun.

-----


matkoniecz:
Note that this quest, unlike some other, is not providing feedback on closing after providing input. I am currently debugging it.

westnordost:
I had the question mark only as an outline to emphasize it is only a placeholder. If you change the question mark for the sidewalk, you should also change it for all other places where such a "street selection puzzle" is shown.

Though, generally, if it is potentially unclear for users how to use the UI, a short hint in the form of one sentence (something like: "Tap the question marks(s) to answer for each side") at the side somewhere could be helpful as well. And/or a pressed-feedback with a ripple effect (starting with v21).

matkoniecz:
> a short hint in the form of one sentence (something like: "Tap the question marks(s) to answer for each side") at the side somewhere could be helpful as well.

I also thought about it but hoped that this change will be enough to make it accessible.

> I had the question mark only as an outline to emphasize it is only a placeholder.

In this case I think that meaning of question mark is quite strong already and it will be not diluted by making it clearly noticeable.

In strong sun it was nearly invisible and easy to miss. I suspect, but not tested, that it is likely to be problem even with a good lighting for someone with poor eyesight.

westnordost:
Ok. It would be good then however to also change it in the source files (the SVGs)

callis:
Another idea would be to change the color of the missing road side. Like it is done in webforms when I missed to fill my phone number.

matkoniecz:
> Ok. It would be good then however to also change it in the source files (the SVGs)

Where this source files are located? I tried `find *| grep .svg` and found svgs in `res` folder, where none matched this produced image.

goldfndr:
https://github.com/westnordost/StreetComplete/blob/master/res/cycleways.svg has the outlined question marks at the top.

matkoniecz:
Thanks! I missed it completely and I looked at this file.

westnordost:
Merged, thank you!

PR #1451, opened 2019-06-29
warn on abandoning partially filled quest

matkoniecz:
gives second line of feedback in case of a confused user described in e5f30dc88b84b72d79817422e25e110d0023cb30

affects sidewalk and cycleway quest

-----


westnordost:
Right, I consider that it was missing as a bug. Thank you!

PR #1452, opened 2019-06-29
explain that photos are useful [ready for review]

matkoniecz:
During testing it turned out that people were strongly underestimating usefulness of photos in notes for other mappers. This attempts to counteract this.

Note that text field is poor - I want it at once to use as much as horizontal space as possible and prefer to use vertical spa and make it work independent of a text length (translations). I am unsure how to solve this and created PR to check whatever solution is obvious.

### DIFF
--- a/app/src/main/res/layout/fragment_attach_photo.xml
+++ b/app/src/main/res/layout/fragment_attach_photo.xml
@@ -21,7 +21,7 @@
 
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/gridView"
-        android:layout_toEndOf="@id/photosAreUsefulExplanation"
+        android:layout_toEndOf="@id/takePhotoButton"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_centerVertical="true"/>
--- a/app/src/main/res/layout/fragment_attach_photo.xml
+++ b/app/src/main/res/layout/fragment_attach_photo.xml
@@ -11,9 +11,17 @@
         android:contentDescription="@string/quest_leave_new_note_photo"
         android:layout_centerVertical="true"/>
 
+    <TextView
+        android:id="@+id/photosAreUsefulExplanation"
+        android:layout_toEndOf="@id/takePhotoButton"
+        android:text="@string/quest_leave_new_note_photos_are_useful"
+        android:layout_width="120dp"
+        android:layout_height="64dp"
+        android:layout_centerVertical="true"/>
+
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/gridView"
-        android:layout_toEndOf="@id/takePhotoButton"
+        android:layout_toEndOf="@id/photosAreUsefulExplanation"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_centerVertical="true"/>
--- a/app/src/main/res/layout/fragment_attach_photo.xml
+++ b/app/src/main/res/layout/fragment_attach_photo.xml
@@ -15,8 +15,8 @@
         android:id="@+id/photosAreUsefulExplanation"
         android:layout_toEndOf="@id/takePhotoButton"
         android:text="@string/quest_leave_new_note_photos_are_useful"
-        android:layout_width="120dp"
-        android:layout_height="64dp"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
         android:layout_centerVertical="true"/>
 
     <androidx.recyclerview.widget.RecyclerView
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -35,6 +35,7 @@
     <string name="quest_leave_new_note_no">"No, just hide"</string>
     <string name="quest_leave_new_note_photo">Attach photo</string>
     <string name="quest_leave_new_note_photo_delete_title">Do you want to delete this picture?</string>
+    <string name="quest_leave_new_note_photos_are_useful">Adding a photo will make the report much more useful.</string>
     <string name="quest_streetName_title">"What is the name of this road?"</string>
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -655,4 +655,5 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="label_housenumber">house number</string>
     <string name="label_blocknumber">block number</string>
     <string name="quest_busStopShelter_covered">Whole stop is covered</string>
+    <string name="quest_leave_new_note_photos_are_useful">Adding photo will make the report much more useful.</string>
```
rugk: Either singular:
```suggestion
    <string name="quest_leave_new_note_photos_are_useful">Adding an photo will make the report much more useful.</string>
```

…or (IMHO better) plural:
```suggestion
    <string name="quest_leave_new_note_photos_are_useful">Adding photos will make the report much more useful.</string>
```

matkoniecz: Right, but it is "a photo", right?

rugk: yep, edited

matkoniecz: Thanks for reminding me that a/an/the exists. I went with singular as single photo in every note would be a massive improvement, requesting multiple ones may be too much.

westnordost:
Does the wording "report" appear at some other place? I'd expect "note".

matkoniecz:
I now used @rugk tip also here, text is now a placeholder of recycler view what removes most problems. It allows to mention a bit more (address plaques part is based on my experience with remote editing based on SC notes), but maybe just short hint that notes are useful should be added...

app/src/main/res/layout/fragment_attach_photo.xml
```diff
@@ -11,9 +11,17 @@
...
+        android:layout_toEndOf="@id/takePhotoButton"
+        android:text="@string/quest_leave_new_note_photos_are_useful"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_centerVertical="true"/>
+
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/gridView"
-        android:layout_toEndOf="@id/takePhotoButton"
+        android:layout_toEndOf="@id/photosAreUsefulExplanation"
```
westnordost: It is problematic to place a view in relation to an element that is `View.GONE`. Not sure why it worked for you, usually is shouldn't. If the view are aligned next to each other, maybe consider a horizontal `LinearLayout` instead to solve this?

app/src/main/res/values/strings.xml
```diff
@@ -35,6 +35,7 @@
     <string name="quest_leave_new_note_no">"No, just hide"</string>
     <string name="quest_leave_new_note_photo">Attach photo</string>
     <string name="quest_leave_new_note_photo_delete_title">Do you want to delete this picture?</string>
+    <string name="quest_leave_new_note_photos_are_useful">Adding a photo will make the note much more useful. Identifiers of location and direction like address plaques are very valuable.</string>
```
westnordost: Not sure what you mean with "identifiers". Also, it is quite long. What exactly is the main point you are trying to make with this note?

app/src/main/res/layout/fragment_attach_photo.xml
matkoniecz: That got carried over from the old version. What about both of views getting ` android:layout_toEndOf="@id/takePhotoButton"`? Always only one is displayed so I would expect it to work.

app/src/main/res/values/strings.xml
matkoniecz: That photo with address plaque is much more likely to be useful for a remote editor. Photo of a shop entrance is sometimes not good enough to actually edit without survey.

For example I would make edit based on https://westnordost.de/p/4162.jpg in https://www.openstreetmap.org/note/1836721 if it would have also an address plaque. Though maybe I should assume that it is in the same place as now nonexisting POI?

But I went trough several notes, and based on that I have seen: overall it is probably not worth making this message more complicated.

matkoniecz:
BTW, is there some reason why you merge rather than squashing commits? (I am a bit curious why because I usually merge & squash with intention of making more readable history - and I wonder is there some reason to do it other way)

matkoniecz:
And thanks for reviewing pull request so well and quickly! I was prepared to wait a long time for review and I think that once again no pull request is ready for a review!

westnordost:
> BTW, is there some reason why you merge rather than squashing commits?

I do not like squash because it rewrites the history. Merge keeps it as it happened. As your commits are pretty good quality (also compared to mine... ;-) ) in that they are quite atomic and descriptive, I find it makes the history actually more readable.

westnordost:
In that regard, I like the approach of Fossil SCM better than Git:

> Git puts a lot of emphasis on maintaining a "clean" check-in history. Extraneous and experimental branches by individual developers often never make it into the main repository. And branches are often rebased before being pushed, to make it appear as if development had been linear. Git strives to record what the development of a project should have looked like had there been no mistakes.

> Fossil, in contrast, puts more emphasis on recording exactly what happened, including all of the messy errors, dead-ends, experimental branches, and so forth. One might argue that this makes the history of a Fossil project "messy". But another point of view is that this makes the history "accurate". In actual practice, the superior reporting tools available in Fossil mean that the added "mess" is not a factor.

> One commentator has mused that Git records history according to the victors, whereas Fossil records history as it actually happened. 

https://www.fossil-scm.org/home/doc/trunk/www/fossil-v-git.wiki

matkoniecz:
> As your commits are pretty good quality

Nice to hear it :)

> One commentator has mused that Git records history according to the victors, whereas Fossil records history as it actually happened.

That is a good summary. Now I think that I often treat squashing as "now I am victorious, and I overcome all that mistakes, dead ends and typos" ritual.

PR #1453, opened 2019-06-29
fix broken sentence in ferry quest

matkoniecz:
I am pretty sure that old version is wrong and new one is correct, but I may be wrong again.

Pinging @goldfndr


westnordost:
*this ferry route

I'd say

matkoniecz:
Updated.

PR #1454, opened 2019-06-29
rename variables to make translating easier

matkoniecz:
as Japanese addressing system is unusual it is quite unclear what one is actually translating
I discovered what is happening by grepping source code and running blame, but it is not something that typical translator can do


westnordost:
The string ids right now are fine like this. It is block and it is house number, translators needn't worry about that it is in Japan.

PR #1455, opened 2019-06-30
reorganize settings again - degrade autosync

matkoniecz:
autosync is not an important option, and enabling it may be problematic - for example one of the complaints in English reviews on Google Play is from person confused by interface that used to be default and now appears only with autosync disabled

autosync also requires people to manually sync what is increasing chance of unsynced data, or data synced very late

In addition advanced setting were moved around - moving to top ones that are more likely to be usefully used

-----

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -7,7 +7,7 @@
     <string name="oauth_cancelled">"Authorization canceled"</string>
     <string name="oauth_failed_permissions">"Authorization failed: All listed permissions need to be granted"</string>
     <string name="title_activity_settings">"Settings"</string>
-    <string name="pref_category_communication">"Communication"</string>
+    <string name="pref_category_account">Account</string>
     <string name="pref_category_display">"Display"</string>
     <string name="pref_title_authorization">"Authorize OSM access"</string>
     <string name="pref_title_authorized_username_summary">"Authorized account: %s"</string>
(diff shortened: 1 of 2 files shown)


matkoniecz:
This PR was triggered by reading through reviews in English and Polish on Google Play. 

For all reading this - Is there some review of StreetComplete on Google Play in your language (that is not Polish nor English) that directly or indirectly has some legitimate complaint or useful suggestion? Without any issue covering this? In that case it may be useful to create issue to let know people about it, assuming that it is possible to create an useful one.

westnordost:
I have no data on this, but I believe that there are not few people who use this setting. Especially seasoned mappers who are used to other editors will prefer to upload and download manually (me included).

Anyway, there are two (more) legitimate reasons why this setting is important and not to be hidden in an advanced section, in my opinion:
1. people with no mobile data flatrates
2. very privacy-aware people since theoretically one could create movement profiles from your contributions in realtime

I too read all the reviews (in all languages) and I never identified autosync-off as a problem.

matkoniecz:
I have no strong feelings here and yes, this two uses are valid ones.

For reference, review that I mentioned was 

>  Aplikacja ciekawa ale jak zaloguje się na swoje konto to nie można wysłać odpowiedzi. Po udzieleniu odpowiedzi ikona służąca do wysyłania wskazuje zero. Zwiększa się tylko liczba przy gwiazdce.

from https://play.google.com/store/apps/details?id=de.westnordost.streetcomplete&hl=pl&showAllReviews=true

it is complaining that upload is broken because count next to an upload icon is not increasing

Though on second thought it was fixed by interface change, as with autosync disabled it would not behave this way

PR #1460, opened 2019-07-01
increase height of download bar

matkoniecz:
according to my UX testing people consistently miss presence of a download bar

this is an attempt to be more clear that app is working

-----


westnordost:
Ok. I plan to redesign the download bar by the way when the user stats are implemented. It should be a circle that shows in the middle the type of quest that is currently downloading plus a (indeterminate) progress circle. For autosync-off, the download-button turns into this progress display when clicked.

Issue #83, opened 2017-04-02
Quest: support trees


PanderMusubi:
Please add support for individual trees, see https://wiki.openstreetmap.org/wiki/Tag:natural%3Dtree

Make sure only to offer one-of-many choices such as `leaf_type`, `genus`, `species` and/or (when applicable for the species) `sex`. Note that for these four values, many can be derived when the most specific was chosen in the app. Best to consult the decision tree with a tree expert / botanist.

Other values such as `taxon`, `name`, `circumference`, `height`, `diameter_crown` and `start_date` are to be derived (taxon) or difficult to impossible to simply observe by looking and should not be offered.

picsi:
A common tag is also ``species:foo`` where "foo" is a two-digit language code. This is used for common names. I think it would be useful to have this as more regular people (non-wizards) know the common names rather than the scientific names.

So maybe let's have a 'be sure of what you're doing' before the user inputs a scientific name in the ``species`` field, or worse: the user adds a comon name into the ``species`` field.

PanderMusubi:
Let them pass a simple English or localises test first.

Jotam:
Mmh, maybe not the most urgent quest to implement.

PanderMusubi:
Agreed, nevertheless a valuable one. Perhaps only start on this when offering a compulsory user test to check if they know what they are doing is easily available in street complete.

matkoniecz:
I used to have `leaf_cycle` and `leaf_type` quests in my fork but I removed them because

- individual trees are impossible to precisely locate during survey and I was never sure whatever I am answering correctly. Maybe it would be improved by map showing locations of trees - but it is rejected by guidelines for SC map style
- quests turned out to be extremely boring, even for somebody who really loves nature
- I had constant feeling that solving this quests is a complete waste of time

Overall, I would not recommend adding this quest.

rugk:
Hmm… do they map forests as single trees where you use StreetComplete? :laughing: :wink: 
Because otherwise you may only find these in parks or some single ones at streets or so. I think they are both always the same (when in a row or so) and less often different when standing near each other,
And in parks one may not have too many trees at once to answer. I think there can hardly be a quest, which is even more boring than the street surface one in cities… :wink: That's why I just move it down from the priority.
So I still consider this not soo bad, maybe limit it to tree rows or so… (then you only answer once for a row, similar to how you do it for a street)

[...]

PR #1465, opened 2019-07-03
add leaf_type quest for forests, fixes #366 [ready to merge]

### LINKED ISSUE #366: New quest: leaf_type for landuse:forest
The [`leaf_type`](https://wiki.openstreetmap.org/wiki/Key:leaf_type) quest for forests should be easy to answer and always stays the same… And one quest can cover a big area… :smiley:

matkoniecz:
This PR is intended as preview of proposed image for needles and broadleaved options (mixed image is obviously not ready), as previous PR failed due to lack of suitable images.

missing things: crediting authors of images, drawables in multiple versions, <s>mixed image composed from both photos</s>, <s>excluding unreasonably large areas (see http://overpass-turbo.eu/s/KqM )</s> (done), <s>can't say option for tagging cacti forests</s> (is `leaf_type=leafless` forest actually a thing? I looked at data and every single one caused me to open an issue as it was questionable at best, given that I am not planning to add this option for now).

I used https://commons.wikimedia.org/wiki/File:Whitebark_pine_Pinus_albicaulis_needle_clusters.jpg https://commons.wikimedia.org/wiki/File:Populus_tremuloides_(5002308731).jpg

Screenshot: 
https://user-images.githubusercontent.com/899988/60584390-6ca1bf80-9d8d-11e9-911d-d50ef5491827.png

-----


matkoniecz:
And thanks @westnordost for upgrades to generic quest classes! It is implemented in basically the same way as earlier PR and all issues that were present there (except image selection) are now fixed without any additional work done in PR itself.

goldfndr:
Would it be workable to also fix #83 by including `natural=tree` nodes and changing the `mixed` value to `leafless` for nodes?

Also, the text `Does this trees` should be changed to `Do these trees` for the queried elements.

matkoniecz:
> Would it be workable to also fix #83

From technical side - extremely simple. 4 minutes of coding, 15 of testing, 1 to make PR. But personally I would decline this quest proposal.

> I see that you dropped it due to finding it boring

Also, it was next to impossible to identify specific tree that generated quest. It turns out that at least in my region `natural=tree` nodes are either about major named trees where no survey is necessary or accuracy is lower than distance between trees.

Frequently even count of mapped trees wildly mismatched reality.

westnordost:
**The quest needs to be limited to very small patches of wood that the user can easily overlook.**

Regarding the pictures, what would work just as well (or better?) would be to use icons. I see you had problems finding a fitting image for "both" (=mixed). With icons it would be no problem, just put both icons next to each other:
https://emojipedia.org/evergreen-tree/
https://emojipedia.org/deciduous-tree/
I already drew an icon for a normal tree for the "detached home" icon that could be reused.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
```diff
@@ -127,6 +128,7 @@
 				new AddMotorcycleParkingCapacity(o),
 				new AddPathSurface(o),
 				new AddTracktype(o),
+				new AddLeafType(o), // used by OSM Carto
```
westnordost: maybe rename to `AddWoodLeafType` to leave room for `AddTreeLeafType`

matkoniecz:
> Regarding the pictures, what would work just as well (or better?) would be to use icons.

I had trouble find finding icons that depict leaf (broad vs needle) rather than tree shape.

But maybe I should not worry about it, with text labels even more tricky cases will be tagged corectly. I thought about tree covered areas of [red cedars](https://en.wikipedia.org/wiki/Juniperus_virginiana) (should be tagged as needles)

180px-2014-05-13_08_32_55_Eastern_Red_Cedar_at_South_Riding_Golf_Club_in_South_Riding,_Virginia

or [black alders](https://en.wikipedia.org/wiki/Alnus_glutinosa) (broadleaf)

220px-20120904Schwarzerle_Reilingen01

and other trees breaking typical conifer/ball shape.

> I see you had problems finding a fitting image for "both" (=mixed).

I plan on making one where upper-left part would from needle photo and lower-right part would be from broadleaf photo

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
matkoniecz: Renamed (though to `AddForestLeafType`)

matkoniecz:
> The quest needs to be limited to very small patches of wood that the user can easily overlook.

Yes, it is planned. I even have a working overpass query though threshold will require a bit of testing.

rugk:
Backlink to issue: https://github.com/westnordost/StreetComplete/issues/366 (was not linked anywhere in this PR)

matkoniecz:
See the title.

rugk:
I said _linked_ not mentioned. It's only linked in the commit, but I could not find it so fast. I don't need to explain [the right way again](https://help.github.com/articles/closing-issues-via-commit-messages/), I did it enough times. No need to discuss this here.

westnordost:
> I had trouble find finding icons that depict leaf (broad vs needle) rather than tree shape.

But what would be the problem with the quest asking for the trees itself, not the leaves? (conifers and deciduous trees)

matkoniecz:
Because `leaf_type` is not about tree shape, it is about type of leaves
 
> leaf_type=* describes the type of leaves, for instance broadleaved or needleleaved. 

I am not sue is there even is tag to mark tree shapes.

Note that while red cedars may have shape typical to broadleaved trees this is still a confifer.

Similarly, shape of black alder is often like a typical conifer tree - it is still deciduous tree.

westnordost:
The definition of conifers and deciduous trees are not about tree shapes, but indeed the leaves, as far as I know. My remark was about how to phrase the question.

matkoniecz:
In that case

> But what would be the problem with the quest asking for the trees itself, not the leaves? (conifers and deciduous trees)

I would expect "needles" and "broadleaved" to be words easier to understood than "conifers" and "deciduous".

In case of going with tree-shape icons it is even more important as it will help to avoid tagging based on tree shape.

matkoniecz:
And for area limits - it can be done in SC (requires adding support for relations) or directly in an Overpass query (requires Overpass v0.7.55, not deployed on French server what would block adding it as an alternative server).

To be more exact - both SC and Overpass have length limits, not area limits. But I consider it preferable as it
 - gives additional penalty to complex multipolygons with holes
 - excludes narrow strips of small area but not solvable in one go (for example tree rows mapped as areas)

http://overpass-turbo.eu/s/KwF limits to 700 meters length and seems to give good results (linked query finds all tree-covered areas below this edge length)

http://overpass-turbo.eu/s/KwH may be good for testing - it shows largest qualifying tree-covered areas

See https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Length for Overpass docs

matkoniecz:
> it can be done in SC (requires adding support for relations

Done in https://github.com/westnordost/StreetComplete/pull/1465/commits/65bb9752696e595dbfa27d48b7f7ba633a2934c6 . Note that it applies to all quests, current limit is 500 m. May be worth opening a separate PR for it.

matkoniecz:
Note that limit directly in query is likely wanted instead/also, to avoid downloading extremely large areas and eating through mobile data limit and later through phone resources in processing it.

Downloading forests may fetch unreasonably large areas and it is not unusual to fetch 10 MB with one query. And there are horror stories with imports of forest data done as a single multipolygon.

goldfndr:
For the polyline/polygon length, would it be more sensible to do the distance comparison within each loop, so that it returns early instead of always calculating the entire list?

matkoniecz:
Proper filtering by size (in query) and image for mixed leaves is now added.

There is some minor work on drawables (adding multiple sizes, crediting authors) but code is ready for review.

Screenshot_2019-07-21-08-08-25-827_de westnordost streetcomplete debug

app/src/main/res/values/strings.xml
```diff
@@ -659,4 +659,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_ferry_pedestrian_title">Does this ferry route transport pedestrians?</string>
     <string name="quest_ferry_motor_vehicle_name_title">Does %s transport motor vehicles?</string>
     <string name="quest_ferry_motor_vehicle_title">Does this ferry route transport motor vehicles?</string>
+    <string name="quest_leafType_title">Does this trees have needles or leaves?</string>
```
```suggestion
    <string name="quest_leafType_title">Do the trees here have needles or leaves?</string>
```

matkoniecz:
I ended writing script to generate drawable files (as promised to myself after doing it manually last time). I am posting it in case that someone is interested in using it or reviewing it for stupid things.

Quality 85 was used based on https://stackoverflow.com/a/44208640/4130619 advise, it seems to work well.

```
# requires imagemagick
# on Ubuntu installable with
# sudo apt-get install imagemagick

def execute_command(command)
  puts command
  `#{command}`
end

def rescale_square(input_filepath, output_folder_path, size)
  output_filename = File.basename(input_filepath, ".*") + ".jpeg"
  output_filepath = File.join(output_folder_path, output_filename)
  # recommended parameters from https://stackoverflow.com/a/44208640/4130619
  convert_command = "-resize #{size}x#{size} -sampling-factor 4:2:0 -strip -quality 85 -interlace JPEG -colorspace RGB"
  execute_command("convert \"#{input_filepath}\" #{convert_command} \"#{output_filepath}\"")
end

def generate_drawable(input_filepath, output_path_root)
  drawables = [
    { name: "drawable-mdpi", size: 128 },
    { name: "drawable-hdpi", size: 192 },
    { name: "drawable-xhdpi", size: 256 },
    { name: "drawable-xxhdpi", size: 384 },
  ]
  drawables.each do |drawable|
    output_folder_path = output_path_root + drawable[:name]
    rescale_square(input_filepath, output_folder_path, drawable[:size])
  end
end

def main
  output_path_root = "/home/mateusz/Documents/StreetComplete/app/src/main/res/"
  Dir["*.{png,jpg,jpeg}"].each do |file|
    generate_drawable(file, output_path_root)
  end
end

main
```

Issue #361, opened 2017-06-30
Quest: maxweight for bridges


krzyk:
This quest is very similar to the maxspeed one. Add a quest to set the `maxweight` tag (together with `source:maxweight=sign`) to highways.

It might be beneficial to have the tonnes set using images/text (so selecting the most popular maxweights would be easier) + a way to enter a number if none of the images/text have the given limit.

This should be applicable to all the highways that maxspeed + service roads also (there are restrictions on those for maxweight already in OSM).

I'm not sure how to deal with the situation when there is not max weight set for given road, so now I'm wondering, is this quest doable in StreetComplete?

westnordost:
Lets do it only for bridges.

On 30 June 2017 11:04:57 CEST, "Krzysztof Krasoń" <notifications@github.com> wrote:
>This quest is very similar to the maxspeed one. Add a quest to set the
>`maxweight` tag (together with `source:maxweight=sign`) to highways.
>
>It might be beneficial to have the tonnes set using images/text (so
>selecting the most popular maxweights would be easier) + a way to enter
>a number if none of the images/text have the given limit.
>
>This should be applicable to all the highways that maxspeed + service
>roads also (there are restrictions on those for maxweight already in
>OSM).
>
>I'm not sure how to deal with the situation when there is not max
>weight set for given road, so now I'm wondering, is this quest doable
>in StreetComplete?
>
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/westnordost/StreetComplete/issues/361

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

matkoniecz:
https://wiki.openstreetmap.org/wiki/Key:maxweight is not mentioning how to tag unsigned cases so I created https://wiki.openstreetmap.org/wiki/Talk:Key:maxweight#Support_for_.22no_weight_limit.22

rugk:
It looks bad for this quest:

> The vast majority of roads have no explicit weight limit posted, though, so this is clearly the default situation. When there's such a clear default, it's not common to explicitly tag the absence of a special case, even if such a value exists.

https://wiki.openstreetmap.org/wiki/Talk:Key:maxweight#Support_for_.22no_weight_limit.22

@matkoniecz Maybe ask them about bridges, only?

westnordost:
The question is: How to tag **bridges** that have no weight limit posted?

[...]

PR #1467, opened 2019-07-04
Maxweight - single sign version [ready for review]

matkoniecz:
This is the first step toward fixing #361

What is included
- support for places where there is OK to assume that one standard maxweight sign covers very large part of uses (excludes UK, USA, Australia)
- `maxweight:signed` - seems to be OK (on tagging mailing list there was complaint from person unaware about its purpose and hopefully resolved procedural complaint and discussion of alternative tagging schemes, but no opposition to marking that survey found no max weight sign at all), hopefully it will not be problematic

Open issues
- code review - I looked through it many times and stopped noticing new problems, but many issues are still there

Planned for the next PR
- <s>retest whatever short ton input works (US is disabled now)</s> (done)
- handle US units - currently it is correctly set to imperial in https://github.com/westnordost/StreetComplete/blob/master/res/country_metadata/measurementSystem.yml#L38 but to enable both short ton and pound units it would need to be `[imperial, metric]` (incorrect for height quest) or have separate metadata file or have special hndling somewhere
- is it a good idea to ask on `highway=trunk`? `highway=primary`? Once mappers start to use it, there may be a chance to tweak filter to reduce number of places where it is asked. I tried to do it already, but failed as I was unable to distinguish between "not surveyed", "surveyed, there is no max weight sign"
- ability to handle signs more complicated than plain max weight sign - requires coding sign switcher and even more research. I expect that notes created by mappers will be useful to find common alternative tags and to check is it worth implementing (for US https://wiki.openstreetmap.org/wiki/File:MUTCD_R12-5.svg was reported as more common than plain max weight sign, based on name https://wiki.openstreetmap.org/wiki/File:UK_traffic_sign_626_2AV2_(2011).svg is likely to be often used for bridges in UK)
- recheck talk-au, talk-us, slack-us, slack-ca, Georgian section on official forum and process mailing lists posts in my mail account

Maybe TODOs
- https://en.wikipedia.org/wiki/Road_signs_in_Botswana#/media/File:SADC_road_sign_TR202.svg - not only nordics have yellow traffic signs
- https://en.wikipedia.org/wiki/Template:Traffic_signs - maybe it may be worth reviewing also other countries

Not actually needed but I really wanted to include it:
Bi_state_vietnam_memorial_bridge_maxweight_signs

-----


app/src/main/res/values/strings.xml
```diff
@@ -617,7 +617,11 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_railway_crossing_barrier_none">It\'s not</string>
     <string name="action_deselect_all">Deselect all</string>
     <string name="quest_buildingType_retail_description">including buildings for restaurants, cafés, …</string>
-
+    <string name="quest_maxweight_title">"What is the weight limit here?"</string>
+    <string name="quest_maxweight_answer_noSign">There is no sign…</string>
+    <string name="quest_maxweight_unusualInput_confirmation_description">This weight looks implausible. Are you sure that it is correct?</string>
+    <string name="quest_maxweight_answer_other_sign">There is an other weight limit sign</string>
```
goldfndr: "an other" is likely to be misread as "another" instead of an *other*. Some suggested alternatives for answer_other_sign

* There is a complex weight limit sign
* There is a non-simple weight limit sign
* The sign has more than one weight limit
* The signage has more than one weight limit
* The signs have more than one weight limit

For the leave note text, I suggest the following change:

`-` Would you like to leave a note? Note with a photo will allow other mappers to handle it. Common signs will be also added to StreetComplete, allowing a direct input.
`+` Would you like to leave a note? A note with a photo will enable other mappers to handle it. Common signs will also be added to StreetComplete, allowing a direct input.

Also, I'm guessing the `…` should be on answer_other_sign, not on answer_noSign.

matkoniecz: Again, thanks for review of English!

> "an other" is likely to be misread as "another" instead of an other. Some suggested alternatives for answer_other_sign

As the next stage I am planning to add support for common complex signs, like R12-5 ( https://wiki.openstreetmap.org/wiki/File:MUTCD_R12-5.svg https://wiki.openstreetmap.org/wiki/Key:maxweight#United_States ).

So plan is that also some complex signs will be answerable directly within app.

So it would be nice to have something that would work both now and also after this planned change.

matkoniecz: > For the leave note text, I suggest the following change: 

Changed.

> Also, I'm guessing the … should be on answer_other_sign, not on answer_noSign.

Changed.

goldfndr: In anticipation of complex signs, I suggest changing `an other` to [`some other`](https://www.theatlantic.com/politics/archive/2016/08/the-rise-of-the-others/497690/). It might look odd and ill-fitting and imprecise, but that's english. :)

`some other type of` would spell it out more clearly, if there's adequate room for the text. `some other kind of` could also work, but `kind` is more often used for broad ranges of animals/plants/materials, not as often for specialized subsets.

matkoniecz: > if there's adequate room for the text.

There is no more space, hopefully "some other" will fit.

goldfndr: ([The hyperlink](https://www.theatlantic.com/politics/archive/2016/08/the-rise-of-the-others/497690/) is only intended to demonstrate that `some other` is common/understood enough to appear on a US census form that a hundred million people will see.)

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
```diff
@@ -0,0 +1,175 @@
...
+import de.westnordost.streetcomplete.quests.max_weight.Measurement.*
+
+private enum class Measurement { METRIC, IMPERIAL }
+
+class AddMaxWeightForm : AbstractQuestFormAnswerFragment<MaxWeightAnswer>() {
+
+    override val otherAnswers = listOf(
+            OtherAnswer(R.string.quest_maxweight_answer_other_sign) { onUnsupportedSign() },
+            OtherAnswer(R.string.quest_maxweight_answer_sign_with_exceptions) { onUnsupportedSign() },
+            OtherAnswer(R.string.quest_maxweight_answer_noSign) { applyAnswer(NoMaxWeightSign("dummy")) }
```
westnordost: Usually answers given via this menu should be confirmed first. See for example the postbox collection times, the opening hours, max height quests etc.

app/src/main/res/values/strings.xml
```diff
@@ -617,7 +617,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_railway_crossing_barrier_none">It\'s not</string>
     <string name="action_deselect_all">Deselect all</string>
     <string name="quest_buildingType_retail_description">including buildings for restaurants, cafés, …</string>
-
+    <string name="quest_maxweight_title">"What is the weight limit here?"</string>
+    <string name="quest_maxweight_answer_noSign">There is no sign</string>
+    <string name="quest_maxweight_unusualInput_confirmation_description">This weight looks implausible. Are you sure that it is correct?</string>
+    <string name="quest_maxweight_answer_other_sign">There is some other weight limit sign…</string>
+    <string name="quest_maxweight_answer_sign_with_exceptions">There are exceptions…</string>
+    <string name="quest_maxweight_unsupported_sign_request_photo">Would you like to leave a note? A note with a photo will enable other mappers to handle it. Common signs will also be added to StreetComplete, allowing a direct input.</string>
```
```suggestion
    <string name="quest_maxweight_unsupported_sign_request_photo">Would you like to leave a note? A note with a photo will enable other mappers to take care of it.</string>
```

I find this note about that StreetComplete will improve in the future problematic. This would also be translated to all the other languages and then later re-translated when the text is changed.

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/Measure.kt
```diff
@@ -0,0 +1,32 @@
+package de.westnordost.streetcomplete.quests.max_weight
+
+//if the minimum required API would be 24, we could maybe use https://developer.android.com/reference/android/icu/util/Measure
```
westnordost: Now this is incorrect for weight.
```suggestion

```

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/Measure.kt
```diff
@@ -0,0 +1,32 @@
...
+}
+
+data class MetricMeasure(val tons: Double) : Measure() {
+    override fun toTons() = tons
+    override fun toString() =
+        if (tons % 1 == 0.0) {
+            tons.toInt().toString()
+        } else {
+            tons.toString()
+        }
```
westnordost: Why the if-else?

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/Measure.kt
```diff
@@ -0,0 +1,32 @@
+package de.westnordost.streetcomplete.quests.max_weight
+
+//if the minimum required API would be 24, we could maybe use https://developer.android.com/reference/android/icu/util/Measure
+abstract class Measure {
+    abstract fun toTons(): Double
+    abstract override fun toString(): String
+}
```
westnordost: Maybe rename to `WeightMeasure`, as there is already another Measure-class?

Also, this is a good example of a candidate for a `sealed` class.

app/src/main/res/layout/quest_maxweight_us.xml
```diff
@@ -0,0 +1,63 @@
...
+            android:layout_height="wrap_content"
+            android:text="WEIGHT LIMIT"
+            android:textStyle="bold"
+            android:textSize="18dp"
+            android:gravity="center"
+            android:textColor="@color/traffic_black"
+            tools:ignore="HardcodedText"/>
+
+        <EditText
+            android:id="@+id/maxSpeedInput"
```
westnordost: ??

app/src/main/res/layout/quest_maxweight_us.xml
```diff
@@ -0,0 +1,63 @@
...
+            android:gravity="center"
+            android:textStyle="bold"
+            android:textColor="@color/traffic_black"
+            android:maxLength="2"
+            android:textSize="42dp"
+            android:layout_marginTop="-12dp"
+            tools:text="80"/>
+
+        <Spinner
+            android:id="@+id/speedUnitSelect"
```
westnordost: !!

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    private fun onUnsupportedSign() {
+        AlertDialog.Builder(activity!!)
+                .setMessage(R.string.quest_maxweight_unsupported_sign_request_photo)
+                .setPositiveButton(R.string.quest_generic_hasFeature_yes) { _, _ -> onClickCantSay() }
```
westnordost: Rather use the "ok" string. This string might be translated in certain languages as something like "Yes it has"

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
```diff
@@ -0,0 +1,175 @@
...
+        if (isMetric) tonInput?.requestFocus()
+        if (isImperial) poundInput?.requestFocus()
+    }
+
+    private fun getSpinnerItems(units: List<String>) = units.mapNotNull {
+        when(it) {
+            "metric" -> "t"
+            "imperial" -> "lbs"
+            else -> null
+        }
```
westnordost: Confused here: What about the US sign that should have the short tons and pounds in the spinner?

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    override fun onClickOk() {
+        if (userSelectedUnrealisticHeight()) {
+            confirmUnusualInput { applyMaxWeightFormAnswer() }
+        } else {
+            applyMaxWeightFormAnswer()
+        }
+    }
+
+    private fun userSelectedUnrealisticHeight(): Boolean {
```
westnordost: *weight

Best to Ctrl+f through this class if any "Height" is left from the copypasta.

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
```diff
@@ -0,0 +1,175 @@
...
+            }
+        } catch (e: NumberFormatException) {
+            return null;
+        }
+        return null
+    }
+
+    private fun isMetric() =
+        weightUnitSelect?.let { it.selectedItem == "t" }
+            ?: (countryInfo.measurementSystem[0] == "metric")
```
westnordost: What about short tons?

app/src/main/res/values/strings.xml
```diff
@@ -617,7 +617,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_railway_crossing_barrier_none">It\'s not</string>
     <string name="action_deselect_all">Deselect all</string>
     <string name="quest_buildingType_retail_description">including buildings for restaurants, cafés, …</string>
-
+    <string name="quest_maxweight_title">"What is the weight limit here?"</string>
+    <string name="quest_maxweight_answer_noSign">There is no sign</string>
+    <string name="quest_maxweight_unusualInput_confirmation_description">This weight looks implausible. Are you sure that it is correct?</string>
+    <string name="quest_maxweight_answer_other_sign">There is some other weight limit sign…</string>
+    <string name="quest_maxweight_answer_sign_with_exceptions">There are exceptions…</string>
```
westnordost: Or "Depends on vehicle..."?

app/src/main/res/values/strings.xml
```diff
@@ -617,7 +617,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_railway_crossing_barrier_none">It\'s not</string>
     <string name="action_deselect_all">Deselect all</string>
     <string name="quest_buildingType_retail_description">including buildings for restaurants, cafés, …</string>
-
+    <string name="quest_maxweight_title">"What is the weight limit here?"</string>
+    <string name="quest_maxweight_answer_noSign">There is no sign</string>
+    <string name="quest_maxweight_unusualInput_confirmation_description">This weight looks implausible. Are you sure that it is correct?</string>
+    <string name="quest_maxweight_answer_other_sign">There is some other weight limit sign…</string>
```
westnordost: Bit too long on smalls screens. "Other max weight sign..."?

matkoniecz:
> I don't think it will. I have the impression that there is broad consensus that sometimes it is necessary to tag that something is not there. However, maxweight:signed=no is so inconsistent with the others like opening_hours:unsigned=yes etc. Why not keep consistent? Can you link to the discussion where maxweight:signed=no was established as the winner?

I thought more about acceptance of the entire idea that tagging "maxweight sign is not present there".

I was not thinking about a specific tag, partially because I do not care at all about exact key/value used for tagging this. I would be equally OK with `maxweight:unsigned=yes` or `unsigned=maxweight` (this one would require in some cases `unsigned=maxweight;maxspeed` and this sounds like an expensive Overpass query so I am a bit dubious about it).

I am not sure what would be the best

- `maxweight:unsigned=yes`
- `maxweight:signed=no`
- `unsigned=*`

and what should be considered as minimal acceptance - a succesful proposal, lack of protests, some support on ml, doing the same as already popular tag?

app/src/main/res/values/strings.xml
matkoniecz: Common ones are "Except inhabitants", "Except deliveries".

Maybe "It depends"?

westnordost: Oh, that is what you meant. Yeah, it depends is good, it covers all.

matkoniecz: Modified in cf4e9534

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
matkoniecz: Fixed

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/MaxWeightAnswer.kt
```diff
@@ -0,0 +1,6 @@
+package de.westnordost.streetcomplete.quests.max_weight
+
+sealed class MaxWeightAnswer
+
+data class MaxWeight(val value: Measure) : MaxWeightAnswer()
+data class NoMaxWeightSign(val dummy: String) : MaxWeightAnswer()
```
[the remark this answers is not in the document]
matkoniecz: Simply removing it gives `Data class must have at least one primary constructor parameter`

It is possible to convert it to simply `class` (and Android Studio suggests further conversion to `object`).

I am not entirely sure about consequences - from what I see code is currently not depending on it being a data class and also conversion to object should be OK.

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/Measure.kt
matkoniecz: I changed also Measure in max height quest.

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/MaxWeightAnswer.kt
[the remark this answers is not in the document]
westnordost: Yes, `object` is the way. See also `OpeningHoursAnswer.kt`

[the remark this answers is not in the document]
matkoniecz: Turned into `object`

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeight.kt
```diff
@@ -0,0 +1,51 @@
...
+            "CA", // requires special icons
+            "US", // requires special icons - single weight sign is available, but it is a rare variant
+            "AU"  // requires special icons - single weight sign is available, but it is a rare variant
+    )
+
+    override val tagFilters = """
+        ways with highway ~ trunk|trunk_link|primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential|living_street|service
+         and !maxweight and maxweight:signed!=no
+         and !maxaxleload and !maxbogieweight
+         and !maxweight:hgv and !maxweight:bus and !maxweight:hgv_articulated and !maxweight:tourist_bus
```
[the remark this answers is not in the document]
matkoniecz:
```
way[~"weight"~"."]({{bbox}});
(._;>;);
out body;
```

appears to work http://overpass-turbo.eu/s/KMm

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
matkoniecz: US sign have short tons described as "t" so it is also displayed this way and later converted to `st` for US locations.

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
matkoniecz: US sign have short tons described as "t" so it is also displayed this way and later converted to st for US locations.

app/src/main/res/values/strings.xml
matkoniecz: see also https://github.com/westnordost/StreetComplete/pull/1467#discussion_r300831197

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeight.kt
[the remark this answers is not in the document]
westnordost: (But this is not required for this PR)

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
westnordost: Then it should not be called "metric" there, otherwise the code is confusing to read.

westnordost:
>I am not sure what would be the best
>    maxweight:unsigned=yes
>    maxweight:signed=no
>    unsigned=*
> and what should be considered as minimal acceptance - a succesful proposal, lack of protests, some support on ml, doing the same as already popular tag?

Well, `unsigned` is obviously bollocks, for the reason you already mentioned. The usage of `*:unsigned=yes` or `*:signed=no` schema should simply be consistent.

Until just now, I thought I remembered that we settled on `*:unsigned=yes` for the opening hours and collection times but having looked another time, I now see that they already use `*:signed=no`, so, sorry for the noise, `maxweight:signed=no` is completely ok.

matkoniecz:
>>> Confused here: What about the US sign that should have the short tons and pounds in the spinner?

>> US sign have short tons described as "t" so it is also displayed this way and later converted to st for US locations.


> Then it should not be called "metric" there, otherwise the code is confusing to read.

and

> handle US units - currently it is correctly set to imperial in https://github.com/westnordost/StreetComplete/blob/master/res/country_metadata/measurementSystem.yml#L38 but to enable both short ton and pound units it would need to be [imperial, metric] (incorrect for height quest) or have separate metadata file or have special handling somewhere

Maybe https://github.com/westnordost/StreetComplete/blob/master/res/country_metadata/measurementSystem.yml#L38 should have 

> US: [imperial, short_ton]

or

> US: [imperial, US]

? Or is special hardcoding in quest (and overriding country_metadata for US) better?

Either way US is a special case, not sure what is better - explicit declaring it in the metada file or not, either way special case for US needs to be implemented in some places.

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
matkoniecz: > handle US units - currently it is correctly set to imperial in https://github.com/westnordost/StreetComplete/blob/master/res/country_metadata/measurementSystem.yml#L38 but to enable both short ton and pound units it would need to be [imperial, metric] (incorrect for height quest) or have separate metadata file or have special handling somewhere

Maybe https://github.com/westnordost/StreetComplete/blob/master/res/country_metadata/measurementSystem.yml#L38 should have 

> US: [imperial, short_ton]

or

> US: [imperial, US]

? Or is special hardcoding in quest (and overriding country_metadata for US) better?

Either way US is a special case, not sure what is better - explicit declaring it in the metada file or not, either way special case for US needs to be implemented in some places.

westnordost: You mean, only in the US, there is pounds AND short tons, and in other countries that use imperial measure, there is just short tons? If yes, what makes you arrive at that conclusion? To me, it would be most logical to both allow pounds (lbs) and short tons as input for countries that use the imperial measure.

matkoniecz: It probably would make sense, though for now US is only place that has units that is not ton. Other places using imperial units appear to be using tons for max weight (or are so small that have no bridges).

westnordost: What is the source of this information?

matkoniecz: > What is the source of this information?

For countries using imperial units I looked through official sites/Wikipedia lists of traffic signs/asked local mappers via mailing lists/slack.

westnordost: Also, this is probably a problem. Doesn't GB use metric tonnes for measurement now?

https://github.com/westnordost/StreetComplete/blob/master/res/country_metadata/measurementSystem.yml#L19

Probably weight and lengths need to be defined separately.

matkoniecz: > Doesn't GB use metric tonnes for measurement now?

Yes, for example GB switched (except some crown dependencies that are tiny islands where data is not triggering this quest, and in case of building new one it can be solved with a note).

westnordost: Okay, so the measurement system for lengths and for weights need to be defined separately. Then one option would be to call these files `weightUnits` and `lengthUnits` instead, and use values like `st`, `lbs`, `meter` and so on instead of the more abstract `imperial` and `metric`

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/Measure.kt
matkoniecz: I admit that it was copied from max height quest, without testing is it actually needed - https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/max_height/Measure.kt#L12-L16

matkoniecz:
Metadata file would be something like

`measurementSystemForWeightLimits.yml`

```
default: [ton]
US: [short_ton_formatted_as_ton, pounds]
```

I am assuming that my research resulted in a correct answers, and everyone except US is really using tons as units.

I am also assuming that change from ton to a new unit is not going to happen.

But metadata file hopefully will add little overhead and in case that my assumptions are wrong it will be easier to amend metadata file over hardcoded handling (though in case of new unit it will be necessary to add its support).

(I started to write this comment to describe hardcoding of US handling as superior but I managed to convince myself that opposite is true).

matkoniecz:
It will be a bit funny to properly implement and test support for values from this metadata file I would need to add countries with fake data – as most cases are currently unused. I am tempted to skip it and support just two combinations currently present. 

ton + short_ton_formatted_as_ton makes no sense at all and there is no good way to support it
short_ton_formatted_as_ton + pounds -> situation in the USA
ton -> the situation everywhere else

but following requires fake data for testing
only short_ton_formatted_as_ton
only pounds
ton + pounds
pounds + ton
pounds + short_ton_formatted_as_ton

westnordost:
By the way, every ton there is is formatted as ton of it is the unit that is used in that country

westnordost:
It is not necessary that the implementation is tested for every theoretical case, but that it is tested for every case that can occur

app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeightForm.kt
matkoniecz: Should be fixed now or at least improved.

matkoniecz:
I am now more happy about the code.

Main problems are

- there are quite fragile conversions between 
 * `TON` (as unit type) <-> `"t" (as unit) <-> `"ton"` (in metadata)
 * `TON` (as unit type) <-> `"t" (as unit) <-> short ton in metadata
 * `POUND` (as unit type) <-> `"lbs" (as unit) <-> `"pound"` in metadata
It probably should be fixed

- no multiple layouts (and handling of multiple layouts) necessary to handle US, Canada and Australia - this requires extensive work so I want to do in the second pass, after going through with a basic version acceptable elsewhere (there is certain small risk that it turns out that this entire quest will be not accepted)
- existing layouts for US is allowing only two digits, so only up to 99 pounds what is not sufficient for typical valued (but US is disabled anyway)

westnordost:
Given your previous comment, I don't understand why this is for review? So, US , Canada and Australia are not working, but you implemented something for it already? So what exactly should I review?

First thing that catches my eye are these two lines:
```
measurementSystem: [imperial]
measurementSystemForWeightLimits: [short_ton_formatted_as_ton, pound]
```
I prefer it to be
```
heightMeasurementSystem: [imperial]  // or heightUnits: [feet and inch]
speedMeasurementSystem: [imperial]  // or speedUnits: [mph]
weightUnits: [short ton, pound]
```

- no reason for the underscores
- wherever short ton, or long ton, is used, it is of course formatted as ton, no? So I don't know what good that information is there. Or is there a country in the world where both different tons are used, so the one ton is written differently?
- your `measurementSystemForWeightLimits` does not specify the measurement system, but units, so also let's call it like that
- with a new list for weight units in place the key  `measurementSystem` is not appropriate anymore, so it should be renamed. What does it cover? Currently, length measurement and speed measurement. But we know already that in the UK, always mph are used for speed and always miles (and yards) for distances. But for height, both meters and feet/inch are used. So, there needs to be a distinction anyway.

matkoniecz:
> Given your previous comment, I don't understand why this is for review? So, US , Canada and Australia are not working, but you implemented something for it already? So what exactly should I review?

Whatever code is OK, especially related to loading metada files and populating dropdown units that received most changes since last review. It is tested (I was temporarily commenting out the line disabling quest in the USA) but more testing is always welcomed.

Maybe also rereview of other code that was already reviewed - it had plenty of issues that I fixed so original problems should be now gone - but maybe I missed something or other problems are more noticeable.

> wherever short ton, or long ton, is used, it is of course formatted as ton, no?

In theory multiple tons can be used at once, in such case there are usually qualifiers used. For example "metric ton" that was (maybe still is) used, mostly in UK where exact ton is not clear.

But main reason for using this name was to make clear that formatting it as "t" for user and sharing input with normal ton is intentional.

I may change if just `short ton` would be better.

> your measurementSystemForWeightLimits does not specify the measurement system, but units, so also let's call it like that

Yes, I will change it soon.

> with a new list for weight units in place the key measurementSystem is not appropriate anymore because it does not cover all measures, so it should be renamed

I will look at it and I will try to find a better name.

westnordost:
Do you mind if I take over to do the rest (review + change the things I'd want to have changed myself)?

matkoniecz:
No, though I feel bad about that I have not completed it :(

I planned to complete it and now I am unsure what would be the next step.

westnordost:
superceded by #1622

PR #1468, opened 2019-07-04
make explicit link to handle confusing POEditor interface

matkoniecz:
Maybe it is just me, but POEditor navigation is not obvious. I think that explicit link may be helpful (though maybe I am sole confused person).


westnordost:
I don't know, sure why not

PR #1471, opened 2019-07-05
explain preferred language for notes [ready for review]

matkoniecz:
In UX test all people who tried to create note were unsure which language for note text was preferable.
thanks to @rugk for the idea where to fit even more text

### DIFF
--- a/app/src/main/res/layout/form_leave_note.xml
+++ b/app/src/main/res/layout/form_leave_note.xml
@@ -20,6 +20,7 @@
         android:inputType="textMultiLine"
         android:gravity="top|start"
         android:minLines="3"
+        android:hint="@string/create_new_note_hint"
         />
 
     <FrameLayout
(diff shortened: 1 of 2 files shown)


westnordost:
It is quite a lot of text. IMO the first sentence makes no sense. Shouldn't people know what an input text is?

Regarding the language, what about this text: "Best write the note in the locally spoken language or otherwise in English". It makes clear that it is just a suggestion (so people can still write in any language they want) but emphasizes that the "fallback" should be English.

matkoniecz:
> Best write the note in the locally spoken language or otherwise in English

Hopefully "but you can use also your own" is obviously implicit and there is no need to write it explicitly.

goldfndr:
Grammatically, I'd favor "It's best to write" rather than "Best write".

As is, the noun is implied, so it isn't a complete sentence. It's very informal, suitable if one is already engaged in a conversation, which StreetComplete might qualify for.

Issue #1466, opened 2019-07-03
After update to LOS 16, no quests are loaded via WiFi, only sometimes it works via mobile data


Moini:
* Start app in flight mode, with WiFi on
* Select to search for quests 
* See no red bar, but get notification that it looks for new quests
* After some time, get a notification that quests couldn't be loaded

It sometimes works via mobile data, but of course takes a lot longer. it often stops after about 20% of the red bar.

The maps are updated normally via WiFi.

**Versions affected**
Lineage OS 16 (build from July 1st), StreetComplete 12.1 from fdroid

**Not affected**
Lineage OS 15 (it worked before I upgraded the OS on my phone)

matkoniecz:
Is it possible that you or someone else is using Overpass API on the same Wifi network?

I noticed exactly the same behavior with multiple phones running scans for quests as the same time - it is caused by rate limiting of Overpass allowing only one query at time from a single IP.

It may explain why mobile data scan works (as it is using a separate IP).

What is displayed at http://overpass-api.de/api/status during failing scans (use the same WIFI network as phone)?

Moini:
Hi, and thanks for replying!

The very first time I loaded that web page after starting the search in StreetComplete, it showed 1 currently running query (and a growing red bar). Then it stopped growing, and told me that slots were available in 47 and 57 seconds (or so). 

I'm not in Russia, and there's no other device in my network that looks at openstreetmap, or anything related (as far as I know...). The only running app that is related to OSM is this one (I hope).

matkoniecz:
I admit that I have no idea what may be going wrong here, most of things that I considered would block quest downloads only sometimes or completely block overpass.

Maybe @westnordost will have an idea what is going wrong.

If you would be interested in doing that - capturing logs is very likely to help.

https://wiki.lineageos.org/logcat.html has an explanation how it can be done

"With a computer" method can be also done by running Android Studio and looking at logcat tab.

This may be quite time-consuming to setup this things if you have neither `adb` nor Android Studio, nor a rooted phone. But in case that you have one of this things then it is quite likely that logs will help to reveal what is going wrong.

Moini:
No problem, I have adb installed - I need it for backups, file transfer and for installing the OS. 

What I did:

* Wake up phone
* StreetComplete was still open
* center on current location
* select to search for quests here
* after 2 sec, get info that it didn't work
* retry
* nothing happens
* retry
* nothing happens

[...]

PR #1472, opened 2019-07-05
stop timeouting after 45 seconds [ready for review]

### LINKED ISSUE #1010: integration tests are prone to failing with java.net.SocketTimeoutException
It applies to both add_housenumber and construction group of integration tests.

This tests sometimes fail with `de.westnordost.osmapi.common.errors.OsmConnectionException: java.net.SocketTimeoutException: failed to connect to overpass-api.de/178.63.11.215 (port 443) after 45000ms: isConnected failed: ETIMEDOUT (Connection timed out)`

It is happening despite overpass turbo operating without any issues.

It was happening also before I started making any changes to integration tests and from what I experienced starting to use attic data failed to change frequency of this issue.

Failures are intermittent, typically disappearing once I decide to debug this issue :)

I think that implementing https://github.com/westnordost/osmapi/issues/16 would be a good first step toward figuring what is causing this problem - can you confirm that it would be a viable PR?

matkoniecz:
default overpass timeout is 180 seconds
should fix #1010, may help with #1466

Construction tests started failing in a new way, but quest download already works nicer

> AddBusStopName: Added 0 new and removed 0 already resolved quests. (Total: 0) in 82s

> AddPlaceName: Added 41 new and removed 0 already resolved quests. (Total: 41) in 95s

logged in my testing would be a timeout errors before this change (BTW, I wonder is AddBusStopName unreasonably expensive or is it just an unlucky call)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/OsmModule.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/OsmModule.java
@@ -64,18 +64,16 @@ public class OsmModule
 		return new OsmMapDataFactory();
 	}
 
-	public static int defaultOverpassTimeoutInMiliseconds(){
-		// see https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#timeout:
-		// default value is 180 seconds
-		// give additional 4 seconds to get and process refusal from Overpass
-		// or maybe a bit late response rather than trigger timeout exception
-		return (180 + 4) * 1000;
-	}
+	// see https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#timeout:
+	// default value is 180 seconds
+	// give additional 4 seconds to get and process refusal from Overpass
+	// or maybe a bit late response rather than trigger timeout exception
+	private static int OVERPASS_QUERY_TIMEOUT_IN_MILISECONDS = (180 + 4) * 1000;
 
 	@Provides public static OverpassMapDataDao overpassMapDataDao(
 			Provider<OverpassMapDataParser> parserProvider, SharedPreferences prefs)
 	{
-		Integer timeout = defaultOverpassTimeoutInMiliseconds();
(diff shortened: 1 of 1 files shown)


rugk:
If it fixes #1010 maybe we can re-enable integration tests for CI? (AFAIK they were disabled, because of that issue.)

matkoniecz:
Nope. See log of CI tests, startup of emulator was failing.

app/src/main/java/de/westnordost/streetcomplete/data/OsmModule.java
```diff
@@ -64,19 +64,29 @@ public static OsmConnection osmConnection(OAuthConsumer consumer)
...
 	}
 
+	public static int defaultOverpassTimeoutInMiliseconds(){
+		// see https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#timeout:
+		// default value is 180 seconds
+		// give additional 4 seconds to get and process refusal from Overpass
+		// or maybe a bit late response rather than trigger timeout exception
+		return (180 + 4) * 1000;
+	}
+
```
westnordost: `private static int DEFAULT_OVERPASS_TIMEOUT_IN_SECONDS`?
The link to the Overpass Api documentation  note about the 4 seconds is probably enough here

westnordost:
AddBusStopName is quite expensive, because of the `or`s.

westnordost:
@PR: Makes sense, if the default Overpass timeout is that long!

app/src/main/java/de/westnordost/streetcomplete/data/OsmModule.java
matkoniecz: I moved to `OVERPASS_QUERY_TIMEOUT_IN_MILISECONDS` name as 184 seconds (including extra buffer) is not actually equal to overpass timeout. 

I used miliseconds to avoid later `* 1000`.

Hopefully it is OK.

PR #1475, opened 2019-07-08
request an early login on explicit upload [ready for review]

matkoniecz:
fixes #1446 - upload button was not working for unlogged users with low star count
affected people not logged in, with low star count and disabled autosync

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -494,7 +494,6 @@ public class MainActivity extends AppCompatActivity implements
 	@UiThread private void requestOAuthorized()
 	{
 		if(dontShowRequestAuthorizationAgain) return;
-		if(answersCounter.waitingForUpload() <= 5) return;
 
 		View inner = LayoutInflater.from(this).inflate(
 				R.layout.dialog_authorize_now, null, false);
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/statistics/AnswersCounter.java
```diff
@@ -30,7 +30,7 @@
 	private View unsyncedContainer;
 
 	private boolean isFirstUpdateDone;
-	private boolean isAutosync;
+	public boolean isAutosync;
```
westnordost: Instead please put `isAutosync` from `onStart()` in `MainActivity` into own function and use that.

```java
private Prefs.Autosync isAutosync()
{
  return Prefs.Autosync.valueOf(prefs.getString(Prefs.AUTOSYNC,"ON")) == Prefs.Autosync.ON
}
```

matkoniecz:
>  because there is `Prefs.Autosync.WIFI`

Argh, I even remembered this time about it and tested, but only to ensure that manual upload still works.

matkoniecz:
Updated again and I again think that it is correct (due to redoing everything in much simpler I went with `reset --hard`, but I may submit version with full history if that is preferable)

westnordost:
Ok, looks good!

matkoniecz:
Thanks for a re-re-re-review! I am so happy that it is now correct (hopefully it is really correct).

westnordost:
maybe you should test it ;-) (I didn't, got a big heap of uncommitted changes in a spiraling-out-of-control refactor of the whole upload logic)

matkoniecz:
I tested it. Unfortunately earlier attempts were also tested, so it is hard to convince myself that this time everything was caught...

> refactor of the whole upload logic

Sounds scary, especially if that is together with way splitting.

PR #1476, opened 2019-07-09
fix missing |

matkoniecz:
I also thought about building regexpes from arrays, but for now went with this.

before: mudwater matched, mud nor water not. After change both mud and water matches as intended.

-----


goldfndr:
The parentheses around all but area=yes look unnecessary, is there a strategy or aesthetic value?

westnordost:
Oh, thanks!

matkoniecz:
> The parentheses around all but area=yes look unnecessary, is there a strategy or aesthetic value?

I admit that I am not sure why it is there, but I worry that changing it will break something subtle (and this is not some code that is often changed, so I prefer to do other refactoring).

PR #1477, opened 2019-07-09
 ask about name of amenity=ferry_terminal + improvements to a comments

matkoniecz:
I have no strong feelings about names of `amenity=ferry_terminal` - I am equally fine with deciding that mandatory `name`/`noname` for all of them is pointless and request change (maybe PR) in the Vespucci.

http://overpass-turbo.eu/s/KB8 (ones without name)
https://taginfo.openstreetmap.org/tags/amenity=ferry_terminal#combinations (60% have name tagged)
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dferry_terminal

---

Note: not a grant-based PR.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -29,13 +29,13 @@ class AddPlaceName(
                 "townhall", "prison", "courthouse", "embassy", "police", "fire_station", "ranger_station",                              // civic
                 "bank", "bureau_de_change", "money_transfer", "post_office", "library", "marketplace", "internet_cafe",                 // commercial
                 "community_centre", "social_facility", "nursing_home", "childcare", "retirement_home", "social_centre", "youth_centre", // social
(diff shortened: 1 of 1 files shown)


matkoniecz:
Vespucci was mentioned, as this PR was triggered by Vespucci complaining that `amenity=ferry_terminal` has neither `name` nor `noname` tag.

Complaint seemed to make sense, though this specific "ferry terminal" was `noname=yes` and my experience with such objects is very limited.

It is possible that Vespucci presets should be amended.

westnordost:
Sounds good, merged!

> Note: not a grant-based PR.

How do you differentiate?

matkoniecz:
> How do you differentiate?

Not included in the grant plan (though the plan has quite strong flexibility, and I now see a way to claim that it is included - but it has no real impact anyway, at least in this case).

BTW, if you are interested in what I ended proposing as a project plan (and got accepted) - I can send it to you via mail.

westnordost:
Sure, I am interested!

rugk:
You won't/can't make it/don't want to make it public? (I would totally accept it, you don't have to your/your choice, but just curious…)

PR #1478, opened 2019-07-10
make star icon clickable

matkoniecz:
2 out of 7 users during UX tests tried to click it, as their very first action
it is also quite obvious place to click
this adds some minimal info there, in future it may be place to show multiple badges/achievements

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -1122,6 +1122,20 @@ public class MainActivity extends AppCompatActivity implements
 
 	// ---------------------------------------------------------------------------------------------
 
+	/** Menu raised by clicking on the star icon */
+	public void starInfoMenu(View view) {
+		String message = getString(R.string.about_contributing) + "\n\n" + getString(R.string.about_missing_stars);
+		if(answersCounter.waitingForUpload() + answersCounter.uploaded() == 0){
+			message = getString(R.string.how_to_get_stars) + "\n\n" + getString(R.string.about_contributing);
+		}
+		new AlertDialog.Builder(MainActivity.this)
+			.setMessage(message)
+			.show();
+	}
+
+
+	// ---------------------------------------------------------------------------------------------
+
 	/** Does not necessarily mean that the user has internet. But if he is not connected, he will
 	 *  not have internet */
 	private boolean isConnected()
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -44,6 +44,7 @@ import android.view.inputmethod.InputMethodManager;
 import android.widget.CheckBox;
 import android.widget.ImageView;
 import android.widget.ProgressBar;
+import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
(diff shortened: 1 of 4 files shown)


rugk:
So what happens when you click on it?

matkoniecz:
Have you tried looking at the patch (see "Files changed" tab at top, linked to https://github.com/westnordost/StreetComplete/pull/1478/files )?

rugk:
Just wanted to say this is a good info in a PR.

matkoniecz:
> this adds some minimal info there

goldfndr:
```xml
- <string name="about_missing_stars">Currently star count is not shared across devices, but answers are added directly to the OpenStreetMap database.</string>
- <string name="how_to_get_stars">Select one of markers on the map and answer the question. This way you will improve OpenStreetMap data and get stars. If you see no markers you may need to zoom out map.</string>
+ <string name="about_missing_stars">Currently, the star count is not shared across devices, but answers are added directly to the OpenStreetMap database.</string>
+ <string name="how_to_get_stars">Choose one of the markers on the map and answer the question. This way you will improve OpenStreetMap data and get stars. If you see no markers you may need to zoom in or out the map.</string>
```

I changed to "zoom in or out" because when one zooms out very far the markers disappear. I changed "Select" to "Choose", as the latter is a more fitting common term in the UI (UX?) when one doesn't have simultaneous multiple choice ability and the element interaction has an immediate consequence.

Normally, I prefer a "then" in such a long "if", but dropping it in informal conversation is common.

rugk:
@goldfndr If you want your diff to be directly apply-able (on GitHub) you need to incooperate that in a comment on a _single line_ in a GitHub [review](https://help.github.com/en/articles/reviewing-changes-in-pull-requests).

westnordost:
> on a single line

Which is not possible for a multi-line diff, or is it?

westnordost:
Anyway, regarding this PR, I am not sure about this. I have other plans for the star button, namely, to lead to the stats and achievements screen later on. So, this feels like an intermediate solution to me, and I generally am hesitant to push intermediate solutions to be translated.

On the other hand, if the strings are written in a way that they fit later for maybe the "achievements" (= "did you know"-screens with nice pictures, triggered when reaching certain goals), it would not hurt.

goldfndr:
I created a comment, using [gitlab syntax](https://github.com/isaacs/github/issues/1452#issuecomment-495892763), and now realize the syntax is GitLab's not GitHub's. :blush:

matkoniecz:
> I have other plans for the star button, namely, to lead to the stats and achievements screen later on.

Yes, it was discussed a bit in #178 - I like this plan, this is intended as the first step in this direction.

> On the other hand, if the strings are written in a way that they fit later for maybe the "achievements" (= "did you know"-screens with nice pictures, triggered when reaching certain goals), it would not hurt.

My plan is to use both

> Select one of markers on the map and answer the question. This way you will improve OpenStreetMap data and get stars. If you see no markers you may need to zoom out map.

and

> Your answers are directly improving OpenStreetMap. Contributed data is open, everyone can use it. Many great projects are already using it.

in one of rewards/badges in a proper star menu.

> Currently star count is not shared across devices, but answers are added directly to the OpenStreetMap database.

is intentionally separate and merged manually into one string as hopefully in the future it will be no longer true (given results of UX testing I will try to work on stars/rewards/synchronization).

matkoniecz:
@goldfndr 

Thanks for language review! I applied most of it

> I changed to "zoom in or out" because when one zooms out very far the markers disappear.

I decided to not use this part. For heavy zoom out there either is (or should be) some warning about this situation and warning that quest are hidden. And among new mappers is not something that seems to happen (though my sample size is low) and "zoom in or out" is making this advice really, really unclear.

matkoniecz:
@westnordost At some point it got removed - is it a regression or an intentional change?

westnordost:
Intentional. The tutorial replaced ths.

FloEdelmann:
I still regularly try to click on the star to get to the profile

PR #1481, opened 2019-07-11
make res/country_metada less confusing

matkoniecz:
I again forgot that running app/generateCountryMetadata.py is necessary

-----

My work on this pull request was sponsored by a NGI Zero Discovery [grant](https://www.openstreetmap.org/user/Mateusz%20Konieczny/diary/368849) (discovered and fixed as part of making #1467)

### DIFF
--- a/res/country_metadata/additionalStreetsignLanguages.yml
+++ b/res/country_metadata/additionalStreetsignLanguages.yml
@@ -1,3 +1,4 @@
+# use app/generateCountryMetadata.py to generate files used by StreetComplete
 # A mapping of additional languages that one can expect to find on a street sign that are not official languages
 # Sources:
 # https://wiki.openstreetmap.org/wiki/Multilingual_names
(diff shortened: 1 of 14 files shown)


westnordost:
To be honest, I don't find this PR particularly useful.

What would be useful, would be to run the generation script automatically on each build. The reason why this is not done yet is because I do not want to have every one who wants to build the app be required to have Python installed.

But there is another solution. 

The script could be converted into a Groovy script. A Groovy script could be executed directly as part of the gradle build script (`*.gradle` files are actually Groovy scripts). I already did this recently with the `downloadPresets` task.
(The gradle build script can actually also be written in Kotlin, see [this documentation](https://kotlinlang.org/docs/reference/using-gradle.html). I didn't have time to dig my nose in it yet though.)

matkoniecz:
> To be honest, I don't find this PR particularly useful.

Because it will not be noticed by people editing this files? In my case I edited file, compiled app, noticed problem, looked at file again (including comments at the top) and started looking for bug in my code.

Later I remembered that this file were not used directly but are merely source of data.

westnordost:
Well I guess this PR does not hurt either. My last comment was not meant as a requirement, just as a suggestion what I would find more useful than these comments.

PR #1482, opened 2019-07-11
Reproducible builds

matkoniecz:
keep newlines the same across systems to avoid spurious changes

currently running `generateCountryMetadata.py` on Linux will change all newlines to `\n` making much larger diff than necessary
this will use Windows style newlines also when writing to file on Linux (the same as currently committed in repo)

-----

My work on this pull request was sponsored by a NGI Zero Discovery [grant](https://www.openstreetmap.org/user/Mateusz%20Konieczny/diary/368849) (discovered and fixed as part of making #1467)

### DIFF
--- a/app/generateCountryMetadata.py
+++ b/app/generateCountryMetadata.py
@@ -16,7 +16,7 @@ for filename in sorted(os.listdir(sourceDir)):
 	if filename.endswith(".yml"):
 		basename = os.path.splitext(filename)[0]
 		with open(sourceDir + filename, 'r', encoding='utf8') as file:
-			data = yaml.load(file.read());
+			data = yaml.load(file.read())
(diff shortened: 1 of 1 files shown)


westnordost:
What is your git setting `core.autocrlf` set to (for this repository)?

See https://help.github.com/en/articles/configuring-git-to-handle-line-endings

matkoniecz:
> What is your git setting core.autocrlf set to (for this repository)?

It is unset for me.

> See help.github.com/en/articles/configuring-git-to-handle-line-endings

Looking at this it seems that .gitattributes file may be a preferable solution.

westnordost:
> Looking at this it seems that .gitattributes file may be a preferable solution.

Yes

PR #1484, opened 2019-07-13
 handle operator as name for postboxes (alternative PR version)

### LINKED ISSUE #1473: display operator for collection times of this postbox quest

matkoniecz:
also, allow each quest to define its own title replacement(s)
(for now only the first one will be used, as there is no
quest at this moment that needs to use multiple ones)

fixes #1473, closes #1474 

there is also change of name of a popular function - as a separate commit, so it should be easy to roll back if more verbosity here is unwanted

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
@@ -8,7 +8,7 @@ interface QuestType<T> {
     val icon: Int
 
     /** returns title resource for when the element has the specified [tags]. The tags are unmodifiable */
-    fun getTitle(tags: Map<String, String>): Int //TODO: rename to getTitleTemplate
+    fun getTitle(tags: Map<String, String>): Int
 
     /** returns on or more replacements, to be used in the title template. The tags are unmodifiable. */
     fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>): Array<String?>
--- a/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
@@ -7,8 +7,11 @@ interface QuestType<T> {
     /** the icon resource id used to display this quest type on the map */
     val icon: Int
 
-    /** the title resource id used to display the quest's question */
-    val title: Int
+    /** returns title resource for when the element has the specified [tags]. The tags are unmodifiable */
+    fun getTitle(tags: Map<String, String>): Int //TODO: rename to getTitleTemplate
+
+    /** returns on or more replacements, to be used in the title template. The tags are unmodifiable. */
+    fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>): Array<String?>
 
     /** returns the string resource id that explains why this quest is disabled by default or zero
      * if it is not disabled by default */
--- a/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
@@ -7,8 +7,8 @@ interface QuestType<T> {
     /** the icon resource id used to display this quest type on the map */
     val icon: Int
 
-    /** returns title resource for when the element has the specified [tags]. The tags are unmodifiable */
-    fun getTitle(tags: Map<String, String>): Int
+    /** the title resource id used to display the quest's question */
+    val title: Int
 
     /** returns on or more replacements, to be used in the title template. The tags are unmodifiable. */
     fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>): Array<String?>
--- a/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
@@ -10,9 +10,6 @@ interface QuestType<T> {
     /** the title resource id used to display the quest's question */
     val title: Int
 
-    /** returns on or more replacements, to be used in the title template. The tags are unmodifiable. */
-    fun getTitleReplacements(tags: Map<String, String>, featureName: Lazy<String?>): Array<String?>
-
     /** returns the string resource id that explains why this quest is disabled by default or zero
      * if it is not disabled by default */
     val defaultDisabledMessage: Int get() = 0
--- a/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
@@ -11,7 +11,7 @@ interface QuestType<T> {
     val title: Int
 
     /** returns on or more replacements, to be used in the title template. The tags are unmodifiable. */
-    fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>): Array<String?>
+    fun getTitleReplacements(tags: Map<String, String>, featureName: Lazy<String?>): Array<String?>
 
     /** returns the string resource id that explains why this quest is disabled by default or zero
      * if it is not disabled by default */
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/OsmElementQuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/OsmElementQuestType.kt
@@ -8,7 +8,7 @@ import de.westnordost.streetcomplete.data.osm.download.MapDataWithGeometryHandle
 
 interface OsmElementQuestType<T> : QuestType<T> {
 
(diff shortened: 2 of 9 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
```diff
@@ -7,8 +7,11 @@ interface QuestType<T> {
     /** the icon resource id used to display this quest type on the map */
     val icon: Int
 
-    /** the title resource id used to display the quest's question */
-    val title: Int
+    /** returns title resource for when the element has the specified [tags]. The tags are unmodifiable */
+    fun getTitleBaseString(tags: Map<String, String>): Int
```
westnordost: Requesting these changes:

1. Please leave `val title: Int` in QuestType as is. Only quests based on OSM elements (`OsmElementQuestType`) should have a `getTitle(Map<String, String>)` because only OSM elements have tags. For example, quests based on notes do not have tags.
2. Please leave the name `getTitle(Map<String, String>)` as is. It is shorter, and descriptive enough.

Not entirely necessary, just suggestions:
1. `getTitleReplacements` could be renamed to `getTitleArgs`. It is shorter and the meaning is still carried over, what do you think? Naming is consistent with for example [here](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#format(java.util.Locale,%20java.lang.String,%20java.lang.Object...)), [here ](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase.html#query(boolean,%20java.lang.String,%20java.lang.String[],%20java.lang.String,%20java.lang.String[],%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)) or [here](https://developer.android.com/reference/android/content/res/Resources.html#getString(int,%20java.lang.Object...))
2. in `getTitleReplacements`, the parameter name for the feature name is `typeName`. "Type" could mean different things, especially in programming context. I think `featureName` is better because the dictionary is also called `FeatureNameDictionary` (consistency) and I chose that name for that class not by accident. See [Features](https://wiki.openstreetmap.org/wiki/Features).

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -57,7 +57,10 @@ class AddPlaceName(
     override val commitMessage = "Determine place names"
     override val icon = R.drawable.ic_quest_label
 
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_placeName_title_name
+    override fun getTitleBaseString(tags: Map<String, String>) = R.string.quest_placeName_title_name
+    override fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>): Array<String?> {
+        return arrayOf(typeName.value)
+    }
```
westnordost: You can use the short-hand syntax here, so it is a one-liner.
```kotlin
override fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>) = arrayOf(typeName.value)
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/OsmElementQuestType.kt
```diff
@@ -8,6 +8,18 @@ import de.westnordost.streetcomplete.data.osm.download.MapDataWithGeometryHandle
 
 interface OsmElementQuestType<T> : QuestType<T> {
 
+    val nameGivingTags: List<String> get() = listOf("name", "brand")
```
westnordost: Generally it is problematic (and thus not good style) to have different functions on an interface influence each other. For example, if I override `getTitleReplacements`, `nameGivingTags` is suddenly defunct. Also, the interface should always be as small as possible. You introduce basically two levels here - `nameGivingTags` being the slightly easier-to-use and `getTitleReplacements` being the slightly more powerful interface. I think this is more confusing than helpful.

I'd very much prefer to have *only* `getTitleReplacements`. Code-wise, it is pretty straightforward, not much more complicated to write than the `nameGivingTags`:

```kotlin
override fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>) =
    arrayOf(tags["name"] ?: tags["brand"])
```

app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
```diff
@@ -12,23 +12,24 @@ import java.util.*
...
     return Html.fromHtml(getString(getQuestTitleResId(questType, element), spanName))
 }
 
-private fun getElementName(element: Element?, locale: Locale, featureDictionaryFuture: FutureTask<FeatureDictionary>?) =
+private fun getElementName(questType: QuestType<*>, element: Element?, locale: Locale, featureDictionaryFuture: FutureTask<FeatureDictionary>?): String? {
+    val tags = element?.tags
+    val typeName = lazy {featureDictionaryFuture?.get()?.let { it.byTags(tags).forLocale(locale).find()?.firstOrNull()?.name }}
+    val nameCandidates = questType.getTitleReplacements(tags ?: emptyMap(), typeName);
+    if(nameCandidates.isEmpty()) return null
+    return nameCandidates.first()
```
westnordost: Instead of the three last lines, you can write
```kotlin
return questType.getTitleReplacements(tags ?: emptyMap(), typeName).firstOrNull();
```

app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
matkoniecz: > Please leave the name getTitle(Map<String, String>) as is. It is shorter, and descriptive enough.

Done.

app/src/main/java/de/westnordost/streetcomplete/data/QuestType.kt
```diff
@@ -10,6 +10,9 @@ interface QuestType<T> {
     /** the title resource id used to display the quest's question */
     val title: Int
 
+    /** returns on or more replacements, to be used in the title template. The tags are unmodifiable. */
```
westnordost: > The tags are unmodifiable

By the way, as opposed to Java, in Kotlin, the collection interfaces `Map`, `List`, etc. are all immutable. The mutable counterparts are `MutableMap`, `MutableList`, etc.

westnordost: Also, `getTitleReplacements`should go into `OsmElementQuestType` as well, of course

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
```diff
@@ -35,7 +35,21 @@ class AddPostboxCollectionTimes(o: OverpassMapDataDao) : SimpleOverpassQuestType
...
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_postboxCollectionTimes_title
+    override fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>): Array<String?> {
+        for(nameTag in listOf("name", "brand", "operator")) {
+            if (tags.containsKey(nameTag)) {
+                val value = tags.getValue(nameTag)
+                return arrayOf(value)
+            }
+        }
+        return arrayOf()
+    }
```
westnordost: I suppose it is a matter of taste, but you really find this better to read than
```kotlin
override fun getTitleReplacements(tags: Map<String, String>, typeName: Lazy<String?>) = 
    arrayOf(tags["name"] ?: tags["brand"] ?: tags["operator"])
```
?

matkoniecz: What about case where neither `name` nor `brand` nor `operator` tag is present? Is it OK to return `arrayOf(null)`? From looking at the code it should work (not tested), but it seems a bit ugly to me.

Though this function is really verbose and it would be nice to shorten it.

westnordost: > Is it OK to return `arrayOf(null)`?

Yes

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
matkoniecz: Is it OK to have long lines? In this case 130 characters.

westnordost: If it is still readable, it is okay.

Though, by the way, you could also leave away the `: Array<String?>`, because it is inferred

matkoniecz:
Thanks for a review!

westnordost:
Thanks for the PR!

PR #1486, opened 2019-07-19
Require minimum variety of quest types [add AQuestDao test]

### LINKED ISSUE #1463: Require minimum variety of quest types
In some places after initial download user see quests of only one type.

It would be useful to ensure at least some variety. It is not obvious for people how many different quests are available and this protests from displaying just single type of quests.

Also, often there is a big cluster of some quests blocking further downloads due to high quest count but user is uninterested on solving them (at least at this specific moment).

I am opening an issue as it looks like more than finding relevant part of code and tweaking line or two, so I prefer to check is it going to be accepted.

So I proposed to add also min quest type count in a download strategy

To modify /app/src/main/java/de/westnordost/streetcomplete/data/download/QuestDownload.java (downloadQuestTypes)

Code doing now related thing: /app/src/main/java/de/westnordost/streetcomplete/data/osm/download/OsmQuestDownload.java

Place to write an SQL query: /app/src/main/java/de/westnordost/streetcomplete/data/osm/persist/OsmQuestDao.java

matkoniecz:
fixes #1463

I based it on `getCount` function. Possibly there is a better way to do that.

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/download/MobileDataAutoDownloadStrategy.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/download/MobileDataAutoDownloadStrategy.java
@@ -37,4 +37,9 @@ public class MobileDataAutoDownloadStrategy extends AActiveRadiusStrategy
 	{
 		return 600;
 	}
+
+	@Override
+	protected int getMinQuestsTypesInActiveRadius() {
+		return 2;
+	}
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/AQuestDao.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/AQuestDao.java
@@ -11,6 +11,7 @@ import java.util.Collection;
 import java.util.List;
 
 import de.westnordost.osmapi.map.data.BoundingBox;
+import de.westnordost.streetcomplete.data.osm.persist.OsmQuestTable;
 
 public abstract class AQuestDao<T extends Quest>
 {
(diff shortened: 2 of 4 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/download/MobileDataAutoDownloadStrategy.java
```diff
@@ -37,4 +37,9 @@
 	{
 		return 600;
 	}
+
+	@Override
+	protected int getMinQuestsTypesInActiveRadius() {
+		return 2;
+	}
```
westnordost: I think 2 will not make much of a difference to the current behavior. The auto downloader currently downloads 4 quest types until it checks again if there are now enough quests.

Question is, why is the current behavior not already enough to ensure a minimum variation in quest types? To find the answer to this question is important to know what should be changed to better the behavior.

matkoniecz: > Question is, why is the current behavior not already enough to ensure a minimum variation in quest types? To find the answer to this question is important to know what should be changed in order to better the behavior.

I encountered case where multiple downloaded quests generated 0 results and road name quest generated big cluster of markers.

It was bad in two ways - (1) cluster was close geographically but not in place where someone wanted to walk (2) tester believed that app will continue to ask solely about street names.

westnordost: ok

matkoniecz:
> You tested it, right?

Yes, as usual I checked whatever new feature actually works (in this case I temporarily bumped number of required quests to 10 to skip looking for place where with specific data pattern)

matkoniecz:
Now I see that while there are claims in log like `"Not downloading quests because there are enough quests in " + radius + "m radius"` it is actually checking in `enclosingBoundingBox(pos, radius)`.

I missed it earlier and noticed it now, but I have no good idea how to formulate it better.

westnordost:
That's fine though

westnordost:
I'll close this, because this is made obsolete by the Multi download

PR #1488, opened 2019-07-21
Introduce end-of-line normalization

matkoniecz:
based on https://www.git-scm.com/docs/gitattributes
this way repository will have LF endings,
text files after checkout will have system specific line endings

this should end problem of newlines changing during editing/regenerating files and being present as spurious part of diffs

Replaces and closes #1482

I encountered claims that it may be better to manually specify manually behavior for each extension rather than relying on `* text=auto` (for example in https://stackoverflow.com/a/38017715/4130619). I am not entirely sure is it something that actually is worth doing, it would require some fiddling and it seems that autodetection works fine.


westnordost:
Very good, thank you for taking care of this!

Issue #1489, opened 2019-07-23
Ask about cycleway/footway surface


matkoniecz:
### General
Affected tag(s) to be modified/added: [cycleway:surface](https://wiki.openstreetmap.org/wiki/Key%3Acycleway%3Asurface), [footway:surface](https://wiki.openstreetmap.org/wiki/Key%3Afootway%3Asurface)

Question asked: **What is the surface of bicycle part of this way?** + **What is the surface of pedestrian part of this way?**

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam) (though is some regions it may be asphalt over and over again for cycleway part, but it should not be worse that standard `surface` quest)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort) - is 8000 elements worldwide enough? http://overpass-turbo.eu/s/KYI

### Ideas for implementation

**Element selection:** http://overpass-turbo.eu/s/KYK

**Metadata needed:** no

**Proposed GUI:** reuse existing surface GUI

I consider it as one of steps toward #279 ("add quest to make surface=paved/unpaved more detailed")

westnordost:
Yes, makes sense, but the element selection is odd. Why `segregated=yes` instead of `cycleway=track` (+opposite_track etc)

westnordost:
Oh, I get it, you proposed this to be about `highway=path|footway|cycleway`, not about `highway=residential|... + cycleway=track|...`.

Then, this is kind of a special case of `AddPathSurface`

matkoniecz:
> Oh, I get it, you proposed this to be about highway=path|footway|cycleway

yes, though

> highway=residential|... + cycleway=track|...?

also may make sense - but should the question be asked also only for `surface=paved`? Or is it useful to ask for all elements where cycleway surface was not set explicitly?

westnordost:
I think it makes sense to ask for any. If there is already an explicitly tagged cycleway, it will have some surface

matkoniecz:
[...]

PR #1493, opened 2019-07-25
ask about surface for cycleway part [ready to merge]

matkoniecz:
implements part of #1489 (footway version is a very similar, so I plan to go through code review with this one).

Confirming that question and query makes sense for `cycleway` track would be great - this type of tagging is not used in my region (dry run was made as usual).

-----


app/src/main/res/values/strings.xml
```diff
@@ -665,4 +665,5 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="about_contributing">Your answers are directly improving OpenStreetMap. Contributed data is open, everyone can use it. Many great projects are already using it.</string>
     <string name="about_missing_stars">Currently, the star count is not shared across devices, but answers are added directly to the OpenStreetMap database.</string>
     <string name="how_to_get_stars">Choose one of the markers on the map and answer the question. This way you will improve OpenStreetMap data and get stars. If you see no markers you may need to zoom out the map.</string>
+    <string name="quest_cyclewayPartSurface_title">What is the surface of cycleway here?</string>
```
```suggestion
    <string name="quest_cyclewayPartSurface_title">What's the surface of the cycleway here?</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -0,0 +1,29 @@
...
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+
+class AddCyclewayPartSurface(o: OverpassMapDataDao) : SimpleOverpassQuestType<String>(o) {
+
+    override val tagFilters = """
+        ways with (
+        (bicycle ~ designated|yes and segregated=yes and highway ~ path|footway|cycleway|bridleway and surface = paved)
+        or
+        (cycleway=track and highway ~ ${OsmTaggings.ALL_ROADS.joinToString("|")})
```
westnordost: This won't work. For cycleways tagged on roads, it would need to be a selection to separately tag the left and the right side.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -0,0 +1,29 @@
...
+import de.westnordost.streetcomplete.data.meta.OsmTaggings
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+
+class AddCyclewayPartSurface(o: OverpassMapDataDao) : SimpleOverpassQuestType<String>(o) {
+
+    override val tagFilters = """
+        ways with (
+        (bicycle ~ designated|yes and segregated=yes and highway ~ path|footway|cycleway|bridleway and surface = paved)
```
westnordost: Only for `surface = paved`? It should rather be surface = anything paved. I think there is a constant for that in `OsmTaggings`.
Though, maybe drop this altogether, because if it is already tagged as segregated, there is not the danger that the app asks for each and every path in a forest or something.

matkoniecz: I would expect thing tagged for example `highway=path surface=paving_stones segregated=yes bicycle=designated foot=designated` to have paving stones both for cycleway and footway part.

(of course, some may be tagged incorrectly but...)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: It seems that with such handling it would be possible to ask about `cycleway:left=track`, `cycleway:right=track`, `cycleway:both=track`, but not about `cycleway=track`.

I assumed that `cycleway=track` may have cycleway on one or both sides, but in case of cycleway on both sides it almost always will be the same surface.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
westnordost: Why would you expect that?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
westnordost: But sometimes it is not.

Also, especially for broad roads, it is difficult to ask for both sides at the same time implicitly because from one side one may not clearly see the other. That will make lazy users tend to answer just simply the surface on their side of the road, hoping that it will probably the same on the other, or that it is not that important.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: Because using `surface=paving_stones` for `highway=path` representing something like

300px-Witosa_bike_2

would be incorrect - half of surface is an asphalt.

Assuming correct tagging `surface=paving_stones` would be used for cases where both pedestrian and cyclist part has paving stones surface.

(at least that is how I always tagged such objects and never expected that others would tag it differently - why `surface=paving_stones` would be used in case where half of surface is asphalt?)

westnordost: Hmm, I understand your point. Though, what about `!surface or surface=paved` then?

matkoniecz: And in turn `surface=paved` is correct where both sides are paved - but exact surface is different.

matkoniecz: > Though, what about !surface or surface=paved then?

Ones without surface should be caught by generic path surface quest. I hope that people will answer correctly with `surface=paved` in cases like this, and specific surface where it is the same for both.

westnordost: Ok

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: It seems to me that it requires two separate quests. Or maybe a single quest with two separate forms would be better - otherwise quest list in preferences would have an apparent duplicate.

westnordost: I'd say two different quests

matkoniecz: I dropped `cycleway=track`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -0,0 +1,24 @@
...
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+
+class AddCyclewayPartSurface(o: OverpassMapDataDao) : SimpleOverpassQuestType<String>(o) {
+
+    override val tagFilters = """
+        ways with highway ~ path|footway|cycleway|bridleway and surface = paved and bicycle ~ designated|yes and segregated=yes
+        and !cycleway:surface and !surface:cycleway
+    """
+    override val commitMessage = "Add path surfaces"
+    override val icon = R.drawable.ic_quest_way_surface
```
westnordost:
```suggestion
    override val icon = R.drawable.ic_quest_bicycleway_surface
```

matkoniecz:
Thanks for the icon!

I will be making also footway equivalent of this quest - `ic_quest_footway_surface.xml` icon would be great.

matkoniecz:
For the technical side of footway variant - I will open a new PR that will include both this one and a new work, and after passing review I will close this PR and keep new one open.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
rugk: > I hope that people will answer correctly with surface=paved in cases like this

I doubt that though.

I was about to write this in a new comment, when I discovered this here:

---

Only if  surface = paved? Two issues:
* What if it is a more specific value like `paving_stones` or `sett`? (a user could "accidentially" set this for segregated ways only loooking at the footway)
* What if it is not set? (If I get the geneal "paved" quest and I have a segregated way, I would not select anything as a user in a segregated way, because it is segregated. Also SC image for "paved" may bias the user's selection in the general case, because IIRC it shows an asphalt way - maybe it could explicitly show such a segregated way to prevent this?)

IMHO this query is a little too specific.

**Proposed solution:** Why not do it the other way around: exclude such segregated ways in the generic surface quest and _instead_ show this (much more useful/correct/specific) one?

westnordost: This means though that StreetComplete tags the surface of segregated cycle- and sidewalks even if it turns out to be the exact same surface. Is this acceptable?

rugk: At least it very explicitly confirms it is the same surface. I know this would be a totally new UI, but if the two quests were asked together (and you only have a "left side"/"right side" selection for "is paved"/"is not paved") or so, then you could provide an answer "both have the same surface"...

matkoniecz: > This means though that StreetComplete tags the surface of segregated cycle- and sidewalks even if it turns out to be the exact same surface. Is this acceptable?

What worse, it leaves `surface` tag unfilled. I think that it is preferable to have `surface=paved cycleway:surface=asphalt footway:surface=paving_stones` over `cycleway:surface=asphalt footway:surface=paving_stones`.

westnordost: It probably is preferable. But the surface-quest is still asked for segregated ways, so the `surface` key will be added sooner or later anyway right?

matkoniecz: Yes, it is asked. But

> exclude such segregated ways in the generic surface quest and instead show this (much more useful/correct/specific) one?

proposed in this thread would change it, that is why I am skeptical.

westnordost: If it is changed like I proposed earlier - do not ask for the surface of a segregated way - then it would be quite incompatible with the quest filter here.

matkoniecz: So there are following ways to handle it

(1)
standard path quests asks about all paths, cycleway/footway quest only for segregated with surface=paved and with pedestrian/cyclist access
what may go wrong: user may incorrectly answer with overly specific surface rather than surface=paved

(2)
standard path quests asks about unsegregated paths, cycleway/footway quest for all segregated with pedestrian/cyclist access
what will go wrong: surface tag will be never set, cycleway:surface and footway:surface will be sometimes set to the same values

(3)
standard path quests asks about all ways, cycleway/footway quest for all segregated with pedestrian/cyclist access
what will go wrong: cycleway:surface and footway:surface will be sometimes set to the same values

I still think that (1) is the preferable solution.

westnordost: It is considered okay to tag building:levels on the building outline even if the building contains building:parts. For data consumers which do not process such detail. I find it reasonable to act the same for surfaces

PR #1498, opened 2019-07-27
use collection that preserves insertion order

matkoniecz:
HashMap was causing elements in relation to not be ordered in the original order
fixes (on my computer) issue discussed in https://github.com/westnordost/StreetComplete/commit/e877c0cce9e5cbbf849576f37b1b248221288cc7

I still worry that also sets may be unordered - are used for example in https://github.com/westnordost/StreetComplete/blob/3244b26320754907a44bfa622338293d39318db9/app/src/main/java/de/westnordost/streetcomplete/data/osm/download/NodeWayMap.java#L41


westnordost:
Oh, you already created one!

Good job on finding that bug!

PR #1500, opened 2019-07-27
Call updateQuestsInRect onStart of QuestsMapFragment (fixes #1415) [WIP; Test why it does not solve 1415]

### LINKED ISSUE #1415: lost request off line
Sometimes I lost the request when I'm off line. I have to found a new conexion and the points of request reappear.

**How to Reproduce**
- charging requests for an area,
- walk through this area
- and after a few minutes, the points dissappear 

**Versions affected**
Android version 9
StreetComplete version 12.1

westnordost:
Based on your observations, @matkoniecz , I would make these changes to attempt to fix the issue. I did not test it yet, wanted to hear your opinion first on whether you think this would solve it.


app/src/main/java/de/westnordost/streetcomplete/tangram/QuestsMapFragment.java
```diff
@@ -298,6 +305,11 @@ protected void updateView()
 		if(lastDisplayedRect != null && lastDisplayedRect.equals(tilesRect)) return;
 		lastDisplayedRect = tilesRect;
 
+		updateQuestsInRect(tilesRect);
+	}
+
+	private void updateQuestsInRect(Rect tilesRect)
+	{
 		// area to big -> skip ( see https://github.com/tangrams/tangram-es/issues/1492 )
```
matkoniecz:
```suggestion
		// area too big -> skip ( see https://github.com/tangrams/tangram-es/issues/1492 )
```

matkoniecz:
From looking at code and my notes - yes, it should fix it.

matkoniecz:
I now tested it. Unfortunately bug is still present :(

matkoniecz:
According to new testing round - now code is prematurely returning in https://github.com/westnordost/StreetComplete/pull/1500/files#diff-6a1df81cf0b361b95285c0d152382eadR323 on `if(minRect == null) return;` check.

westnordost:
Turns out, I cannot reproduce this problem at all on my phone, so I can not really test if my solution works. I'll look at your hint and see if I find a problem there.

matkoniecz:
Let me know if nothing clear will be found, I will try to either debug it or at least describe exactly what happens.

westnordost:
Yes you could help. If you set a breakpoint in line 322, what is `tilesRect` and what is `retrievedTiles`? My assumption would be that `retrievedTiles` is empty because it was cleared on `onStop`. `tilesRect`'s size should be at least 1.
But one of my assumptions is wrong.

matkoniecz:
```
tilesRect = {Rect@6336} "Rect(9098, 5550 - 9099, 5551)"
tiles = {ArrayList@6337}  size = 0
retrievedTiles = {HashSet@5650}  size = 4
 0 = {Point@6353} "Point(9099, 5550)"
 1 = {Point@6354} "Point(9099, 5551)"
 2 = {Point@6355} "Point(9098, 5550)"
 3 = {Point@6356} "Point(9098, 5551)"
```

westnordost:
Ok, so `retrievedTiles` is not empty...
This means that either `onStop` is not called - unlikely, or `updateQuestsInRect` is called by someone else after `onStop` but before the call triggered by `onStart`.

matkoniecz:
> or updateQuestsInRect is called by someone else after onStop but before the call triggered by onStart

It is called by onGlobalLayout listener.

```
updateQuestsInRect:320, QuestsMapFragment (de.westnordost.streetcomplete.tangram)
updateView:315, QuestsMapFragment (de.westnordost.streetcomplete.tangram)
run:-1, -$$Lambda$2_Jtz9DeceGHtlZxeQNJMnj0Clc (de.westnordost.streetcomplete.tangram)
onGlobalLayout:23, ViewUtils$1 (de.westnordost.streetcomplete.util)
dispatchOnGlobalLayout:912, ViewTreeObserver (android.view)
performTraversals:1996, ViewRootImpl (android.view)
doTraversal:1134, ViewRootImpl (android.view)
run:6045, ViewRootImpl$TraversalRunnable (android.view)
run:860, Choreographer$CallbackRecord (android.view)
doCallbacks:672, Choreographer (android.view)
doFrame:608, Choreographer (android.view)
run:846, Choreographer$FrameDisplayEventReceiver (android.view)
handleCallback:739, Handler (android.os)
dispatchMessage:95, Handler (android.os)
loop:148, Looper (android.os)
main:5441, ActivityThread (android.app)
invoke:-1, Method (java.lang.reflect)
run:738, ZygoteInit$MethodAndArgsCaller (com.android.internal.os)
main:628, ZygoteInit (com.android.internal.os)
```

westnordost:
This is really weird - how can that listener still be registered? Or is it re-registered somewhere between onStop and onStart?

matkoniecz:
```
					try {
						vto.removeOnGlobalLayoutListener(this);
					} catch (IllegalStateException e) {
						Log.wtf(TAG, "removeOnGlobalLayoutListener failed");
						// ignore: checking isAlive before is not sufficient (race condition)
					}
```

debug reveals that `vto.removeOnGlobalLayoutListener(this);` command is often failing. Is it normal or is it something that is supposed to not happen?

westnordost:
No, that should not usually happen. It should only happen if the view it was contained in is already gone (e.g. Activity destroyed or sent in background perhaps).

That's what I though at least.

westnordost:
By the way, you can also set a breakpoint, rightclick on that breakpoint and make it not pause the execution and instead let it f.e. log something. This way, you do not need to recompile to check how often something happens etc.

westnordost:
(I accidentally branched the fix for the undo button off this PR, nevermind that.)

@matkoniecz I depend on you debugging this further, because I cannot reproduce this problem at all.

matkoniecz:
> No, that should not usually happen. It should only happen if the view it was contained in is already gone (e.g. Activity destroyed or sent in background perhaps).

In this case it is happening as the view is still gone. App is woken up, but maybe view is caosidered still gone - user see lock screen so this is explainable.

But `vto.removeOnGlobalLayoutListener(this);` fails also as app is running and I rotate phone! I have no idea why it is happening.

In addition I modified code to set breakpoint in case when listener is succesfuly remove - it never, ever happens!

I have no idea what is happening here.

```
				@Override public void onGlobalLayout()
				{
					boolean removed = true;
					try {
						vto.removeOnGlobalLayoutListener(this);
					} catch (IllegalStateException e) {
						Log.wtf(TAG, this.toString());
						Log.wtf(TAG, "removeOnGlobalLayoutListener failed");
						// ignore: checking isAlive before is not sufficient (race condition)
						removed = false;
					}
					if(removed) {
						Log.wtf(TAG, "removeOnGlobalLayoutListener is success!");
					}
					runnable.run();
				}
```


> By the way, you can also set a breakpoint, rightclick on that breakpoint and make it not pause the execution and instead let it f.e. log something. This way, you do not need to recompile to check how often something happens etc.

Thanks! This is a very useful info!

matkoniecz:
So problem is that 

* listeners in some cases are not uregistering itself and as result fire at improper times - it is unclear what is causing this and how it may be fixed
* `retrievedTiles` tiles may be updated despite that `onQuestsCreated` was never actually called and retrieval was unsuccesful. But checking status of call from return value is not easy as chain from `updateQuestsInRect` leads to asynchronous call (`listener.onFirstInView(bbox)` calls `questController.retrieve(bbox)` with `workerHandler.post` asynchronous call that calls `relay.onQuestsCreated` which may fail to relay). Checking value of listener of relay is also not correct, it may change between our check and call of `relay.onQuestsCreated`.

So what can be done? Give up and retrieve all quests from database on each unlock just because some phones are borked?

Maybe it is possible to avoid use of `ViewUtils.postOnLayout`? Especially `if(v != null) ViewUtils.postOnLayout(v, this::updateView)` - there is also second use but it seems to be not causing problems here.

So what `updateView` is doing?

`updateView` calls `super.updateView();` and later runs list of checks before potential call to `updateQuestsInRect`.

This PR add new call to `updateQuestsInRect` that will be triggered on start at correct moment, so only what is called in `super.updateView();` is potentially wanted. Doing this is fixing problem (on my phone).

Maybe even what is called in `super.updateView();` can be skipped or called from a different place and `if(v != null) ViewUtils.postOnLayout(v, this::updateView)` eliminated?

westnordost:
Maybe you could try some minimal setup in a new app project to debug why the postOnLayout does not work - and find a way how it will. Perhaps, the lambda is the problem?

matkoniecz:
Good idea - I will do that, but I will be gone for about week.

westnordost:
@matkoniecz I noticed that in `androidx.core:core-ktx` (`View.kt` in package `androidx.core.view`), there are already the extension functions `View.doOnLayout` and `View.doOnPreDraw` defined, so the ones I wrote myself could actally be removed. Perhaps using these functions in place of mine will already solve the problem as I made a mistake when implementing these?

matkoniecz:
Is `androidx.core.view` supposed to be unavailable from Java? It seems to me that it requires converting affected file to Kotlin (but I am not 100% sure).

Automatic conversion to Kotlin for some reason is not fully working. There is one part with invalid code that seems to be easily fixed, but Dagger starts to complain anyway and crashes compilation.

To be more specific
without converting to Kotlin

`import androidx.core.view.*;` was not changing that `view.doOnLayout` is triggering "Cannot resolve symbol". From what I see `adroidx.core.view` is used solely in Kotlin code (both in StreetComplete codebase and in what I found on the internet).

Kotlin conversion ( https://github.com/matkoniecz/Zazolc/commits/kotlin-conversion-failed )
Kotlin conversion produces one syntax error that appears to be fixable. But anyway compilation starts to crash, with
```
/home/mateusz/Documents/StreetComplete/app/build/tmp/kapt3/stubs/debug/de/westnordost/streetcomplete/tangram/MapControlsFragment.java:17: error: Dagger does not support injection into private fields
    private android.content.SharedPreferences prefs;
```


```
/home/mateusz/Documents/StreetComplete/app/src/main/java/de/westnordost/streetcomplete/ApplicationComponent.java:64: error: de.westnordost.streetcomplete.tangram.MapControlsFragment cannot be provided without an @Inject constructor or from an @Provides-annotated method.
	void inject(MapControlsFragment mapControlsFragment);
	     ^
      de.westnordost.streetcomplete.tangram.MapControlsFragment is injected at
          de.westnordost.streetcomplete.ApplicationComponent.inject(mapControlsFragment)
```
It is a bit confusing because Dagger is complaining about incorrect code in Dagger generated code.

I plan on working through it further and posting it in case that fix is obvious and to note that this bug is not abandoned.

westnordost:
Of course it doesn't work like that, View.doOnLayout is an extension function. Such a concept does not exist in Java. In Java, extension functions need to be called like a static function, see https://stackoverflow.com/questions/28294509/accessing-kotlin-extension-functions-from-java

Let's **not** convert QuestsMapFragment to Kotlin just yet.

matkoniecz:
@westnordost Implemented, tested and YES! #1415 is slain!

westnordost:
obsolete through #1571

Issue #1433, opened 2019-06-20
rapid undo failed


matkoniecz:
During testing #1432 I made a test edit in a live OSM database that I immediately reverted[1]. To my surprise revert was not actually applied - see history of https://www.openstreetmap.org/way/22766359/history

For example https://www.openstreetmap.org/changeset/71430814 contains test edit adding access tag, revert edit was not applied. Selecting "undo" again was showing toast that nothing more can be undone.

It may be caused by competing edits and undos applied to the same some object in a short time, I am not considering likely that #1432 code is buggy.

[1] not the optimal way of testing but I have no idea how to do it better and fully test it, usually I solve quests on a real survey - in this case the nearest quest was quite far away from me

westnordost:
Could you try to reproduce it, looking at the log? Iirc it is logged if a change is not applied due to a conflict

mmd-osm:
Does SC fetch some data from the API via /map right after the first upload and before triggering the revert? 

There's typically a very small replication delay, which could cause some object (versions) not being part of the /map call, if you send the query too quickly after the upload.

JOSM handles this by processing the diffResult, iD has a delay of a few seconds right after the upload and before loading new data.

westnordost:
@mmd-osm 
On upload of each change*, SC gets the new version of the uploaded element from the diffResult and saves the updated version of the element to local storage, yes.
If there is a conflict while uploading an element (server version is higher than local version), only then SC does a `GET /[node|way|relation]/$id` (etc.) to update the local version of the element and tries again.

Are you saying that one cannot rely on that this GET call will return always the current version of the element?

* a revert is the same mechanism

mmd-osm:
CGImap *can* point to different databases for read/write and read-only operations. During normal operations, both point to the same primary database server *karm* in Amsterdam. 

It might happen during a fail over, that a read-only mirror *katla* is used for /map and other GET operations, in which case you might experience a few seconds delay (see https://munin.openstreetmap.org/openstreetmap.org/katla.openstreetmap.org/postgres_replication_9_5_main.html).

To detect a replication delay, one option could be to store the previous diffResult response and use the object/object versions to validate subsequent GET calls.

Here's the respective iD issue: https://github.com/openstreetmap/iD/issues/1646 with has some further pointers. Some of the discussion may not be relevant anymore, I don't recall exactly, how the replication was set up before the move to AMS.

matkoniecz:
> Could you try to reproduce it, looking at the log?

[...]

PR #1507, opened 2019-07-29
Undo button visibility

### LINKED ISSUE #1480: undo stacking during startup
Step 1: solve two different quest

Step 2: Switch applications, use system killer of apps

Step 3: Start StreetComplete, as it starts keep clicking "Undo" button

As result multiple undo windows will appear, each allowing to keep undoing. It is possible to undo more times than solved quests and get negative star count.

westnordost:
Fixes #1480, #1433

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -142,6 +142,7 @@ public class MainActivity extends AppCompatActivity implements
 	private ProgressBar uploadProgressBar;
 
 	private View unsyncedChangesContainer;
+	private MenuItem btnUndo;
 
 	private float mapRotation, mapTilt;
 	private boolean isFollowingPosition;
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
```diff
@@ -849,6 +869,7 @@ private void showMarkerSolvedAnimation(@DrawableRes int iconResId, PointF startS
 		flingQuestMarkerTo(img, answersCounter.getAnswerTarget(), () -> {
 			root.removeView(img);
 			answersCounter.addOneUnsynced(source);
+			btnUndo.setVisible(true);
```
westnordost: Not all quests can be undone. I.e. note quests can't. Maybe need to call `updateUndoButtonVisibility();` instead, or find out first if the solved quest is something undoable.

matkoniecz:
One problem that I spotted is that in case where the first solved quest is a note one undo button still appears.

westnordost:
Yes exactly, but it is fixed by my last commit, is it not?

matkoniecz:
I retested it now and it still happens. Tested by uninstalling dev version of SC, installing dev version compiled from 184005a270f65ec056934a710de8844fe019e38b selecting single note quest, typing gibberish and selecting OK.


Curiously, switching apps and killing SC then entering it is NOT showing undo button but showing one more note quest is showing it again.

Solving normal quest, switching apps and killing SC then entering it is showing undo button.

westnordost:
Ok I will have a look at it tomorrow

PR #1511, opened 2019-07-31
replace fair use images by ones on a compatible licence

matkoniecz:
Replaces fair use images - with questionable legal status, by images that are worse but not terrible and on a suitable licences.

After this PR all used assets are on an open licence.

-----


westnordost:
OK

PR #1516, opened 2019-08-05
Overpass performance

### LINKED ISSUE #1514: Performance improvements: Faster Overpass queries
Some Overpass queries are very slow and clog up the download process by pitching the app in a long queue, waiting to replenish its Overpass download quota. If the queries were executed faster, the app would reach the quota later.

**So:**

1. Find out which quests take long and why. First step here is to create a little program in the `tests/` directory that iterates through the quest type list and outputs the compiled overpass query for each for some bbox to console. See https://github.com/westnordost/StreetComplete/issues/1457#issuecomment-509031473. The output of this program would also be interesting for @rugk who is maintaining the quests wiki page and links to the used overpass queries for each quest type.
Executing each query on some location should give an idea which quest types are problematic.

2. Implement optimizations based on these observations. In the predecessor ticket to this one, a few suggestions have already been made:
    1. make possible the usage of regex in keys in the TQL syntax and use it where applicable
    2. express nested `or`s in OQL with `if: <condition>` blocks if they don't contain regex
    3. Far the biggest impact though has some fiddling of evaluation strategies for Overpass Queries. At the moment, it is unknown what the best strategy would be here. See https://github.com/westnordost/StreetComplete/issues/1457#issuecomment-513545692 and /cc @mmd-osm

westnordost:
Fixes #1514 

Additionally to the code, the list here needs to be reviewed to find any potential mistakes in the overpass query generation algorithm:

https://github.com/westnordost/StreetComplete/issues/1514#issuecomment-518362266


matkoniecz:
I was looking at https://github.com/westnordost/StreetComplete/issues/1514#issuecomment-518362266 (I made no attempt to verify code or to check that queries listed there are actually what SC uses).

Is it intentional that _link roads will no longer gave max speed quest? In my experience link roads often have limits, even ones on motorways (it is caused by appearance of `^` `$` delimiters).

The same for AddWayLit.

```
[bbox:{{bbox}}];
node[barrier = height_restrictor] -> .n1;
node[amenity = parking_entrance][parking ~ '^underground|multi-storey$'][!maxheight][!'maxheight:physical'] -> .n2;
(.n1; .n2;);

out meta geom;
```

seems to be missing `[!maxheight][!'maxheight:physical']` part for `[barrier = height_restrictor]` or it is spurious for `[amenity = parking_entrance]`.


----

I found several things to improve during looking at queries, so I will open some PRs with minor fixes before reviewing code.

westnordost:
Thank you for the review.

Regarding `AddMaxHeight`, this is indeed a bug! Phew, good that you found it! I'll have a look!

Regarding `AddWayLit` and `AddMaxSpeed`, nothing changed. It was always be like this.

matkoniecz:
> Regarding AddWayLit and AddMaxSpeed, nothing changed. It was always be like this.

In that case I am going to reclassify it as "things to improve found during looking at queries" and open PR for that.

> I'll have a look!

And once it is fixed I will merge this PR into version of app that I use on survey, maybe I will spot something additional.

westnordost:
### AddMaxHeight
```
[bbox:{{bbox}}];
node[barrier = height_restrictor] -> .n2;
node[amenity = parking_entrance][parking ~ '^underground|multi-storey$'] -> .n3;
(.n2; .n3;) -> .n1;
node.n1[!maxheight][!'maxheight:physical'];

out meta geom;
```
```
[bbox:{{bbox}}];
way[highway ~ '^motorway|motorway_link|trunk|trunk_link|primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential|living_street|track|road$'] -> .w2;
way[highway = service][access !~ '^private|no$'][vehicle !~ '^private|no$'] -> .w3;
(.w2; .w3;) -> .w1;
way.w1[covered = yes] -> .w4;
way.w1[tunnel ~ '^yes|building_passage|avalanche_protector$'] -> .w5;
(.w4; .w5;) -> .w1;
way.w1[!maxheight][!'maxheight:physical'];

out meta geom;
```

matkoniecz:
BTW, it is really great that performance of queries will be so increased! Both for SC users and Overpass servers.

westnordost:
Hmm, it is still wrong, now for other queries.

westnordost:
Ok, now it should be fine. (Updated the comment in #1514 with all the filters)

PR #1518, opened 2019-08-13
further small readme tweaks

matkoniecz:
based on proofreading by a friend, he is not a native speaker but his English is better than mine

pinging @goldfndr for review

He also proposed changing 

> It asks simple questions, with answers directly used to edit and improve OpenStreetMap data. 

to

> It asks simple questions, answers to which are directly used to edit and improve OpenStreetMap data. 

but I am dubious about it

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 StreetComplete is an easy to use editor of OpenStreetMap data available for Android. It can be used without any OpenStreetMap-specific knowledge. It asks simple questions, with answers directly used to edit and improve OpenStreetMap data. The app is aimed at users who do not know anything about OSM tagging schemes but still want to contribute to OpenStreetMap.
 
-StreetComplete automatically looks for nearby places where a survey is needed and shows them as quest markers on its map. Each of these quests can then be solved on site by answering a simple question. For example, tapping on a marker may show a question "What is the name of this road?", with a text field to answer it.
+StreetComplete automatically looks for nearby places where a survey is needed and shows them as quest markers on its map. Each of these quests can then be solved on site by answering a simple question. For example, tapping on a marker may show the question "What is the name of this road?", with a text field to answer it.
 More examples are shown in the screenshots below.
 
 The user's answer is automatically processed and uploaded directly into the OSM database. Edits are done in meaningful changesets using the user's OSM account.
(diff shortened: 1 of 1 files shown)


westnordost:
Yes, me too. And I do not really see the improvement in this PR.

matkoniecz:
"the question" in 

> StreetComplete automatically looks for nearby places where a survey is needed and shows those as quest markers on its map. Each of these quests can then be solved on site by answering a simple question. For example, tapping on a marker may show the question "What is the name of this road?", with a text field to answer it.

AFAIK is incorrect as some specific question is not discussed earlier, it should be "a question".

README.md
```diff
@@ -2,14 +2,14 @@
 
 StreetComplete is an easy to use editor of OpenStreetMap data available for Android. It can be used without any OpenStreetMap-specific knowledge. It asks simple questions, with answers directly used to edit and improve OpenStreetMap data. The app is aimed at users who do not know anything about OSM tagging schemes but still want to contribute to OpenStreetMap.
 
-StreetComplete automatically looks for nearby places where a survey is needed and shows those as quest markers on its map. Each of these quests can then be solved on site by answering a simple question. For example, tapping on a marker may show the question "What is the name of this road?", with a text field to answer it.
+StreetComplete automatically looks for nearby places where a survey is needed and shows them as quest markers on its map. Each of these quests can then be solved on site by answering a simple question. For example, tapping on a marker may show a question "What is the name of this road?", with a text field to answer it.
```
westnordost: Revert this and I'll merge it

matkoniecz: Including the -> a change?

goldfndr: The those -> them change is good. The the -> a change is bad, as it refers to one specific wording and not a more vague example or multiple specific examples.

goldfndr:
The definite article "the" always[1] refers to the subject/object (noun or pronoun or name or whatever) that it precedes, but I'd be interested in seeing a specific example (if you have one) in which something earlier could influence the definite vs indefinite decision.

[1] (English generally has numerous exceptions but none come to mind. Even *Yoda speak* would conform.)

README.md
westnordost: K has goldfndr said

PR #1519, opened 2019-08-14
ask about name of tourism=chalet

matkoniecz:


westnordost:
I can think of examples where individual chalets do not have names: In accomodations that are essentially chalet complexes, the individual chalets may just have numbers, like room numbers. I'd expect those places usually be tagged as `tourism=resort`, but I am not sure if the definition of resort is so sound that it is always understood and used this way.

matkoniecz:
>  In accomodations that are essentially chalet complexes, the individual chalets may just have numbers, like room numbers.

In that case I would treat tagging individual buildings with `tourism=*` tags as incorrect in the same way as mapping each hotel room as `tourism=apartment` or each gas pump at petrol station as `amenity=fuel`.

westnordost:
You have a point. However, I am not even sure if `tourism=resort` is unmistakeably defined in the wiki (and used in practice) the way I interpret it to be, as a "collection of chalets".
So, this requires a bit of research.

westnordost:
Actually, I just found that "resort" is not only not what I thought it is, it is also deprecated. In South East Asia, there are many places calling themselves "resort" that are this kind of collection of chalets and/or apartments.
A few years back, I mapped this island:
https://www.openstreetmap.org/#map=17/11.81594/102.46438
and it turns out, I already used `tourism=chalet` for this kind anyway.

westnordost:
So, this PR is fine.

PR #1520, opened 2019-08-14
start asking about max speed and lit status on link roads

matkoniecz:
It makes sense to ask about max speed on link roads (even on motorways some of them have speed restrictions, sometimes quite steep so useful for routing software to warn about it).

Link roads also are not special as far as lit status goes, it makes sense to ask about it (with the same restrictions as other roads).


westnordost:
Ok

PR #1521, opened 2019-08-14
stop asking about bicycle parking type where there is no access

matkoniecz:
access=private was handled already, acces=no not

during testing I discovered that this tagging is fairly rare but still happens (for example https://www.openstreetmap.org/way/354177980 used during testing)


westnordost:
👍

PR #1522, opened 2019-08-14
stop duplicating office lists, use full list in each quest [ready for review]

matkoniecz:


app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.java
```diff
@@ -22,4 +22,9 @@
 	};
 
 	public static final String SURVEY_MARK_KEY = "check_date";
+
+	public static final String[] OFFICES_VISITED_BY_GENERAL_PUBLIC = {
+		"insurance", "government", "lawyer", "estate_agent", "political_party", "travel_agent",
+		"tax_advisor", "therapist", "religion",
+	};
```
westnordost: # Opening hours

lawyers are not really walk-in for customers, as much as an architect or other company providing a service (such as: building a table, building a smartphone app) aren't. They do work for clients, but the office hours are certainly not opening hours.

Example: [White & Case](https://www.whitecase.com/locations/emea/hamburg) (big law firm) lists no opening hours, only a phone number.

Same with political parties, it is not really public in any sense. You can't simply walk in there, it is rather the point where the political party meets. Maybe they organize some public Q&A hours, maybe the don't.

Also, from my experience, therapists do not have a reception, thus, it's not walk-in and appointments are made via phone or internet.

Regarding tax advisors, I am not sure. The offices I know *look like* one could just walk in there from the outside, but I am not sure if they really have opening hours.

# Name

✔️ 

# Wheelchair

Well, nothing changed here. Some are not walk-in, but so is f.e. a concert hall or theatre, and `wheelchair` is and should be tagged on these as well.

matkoniecz:
> lawyers are not really walk-in for customers, as much as an architect or other company providing a service (such as: building a table, building a smartphone app) aren't. They do work for clients, but the office hours are certainly not opening hours.

> Example: White & Case (big law firm) lists no opening hours, only a phone number.

I have (fortunately) little experience here

> Same with political parties, it is not really public in any sense. You can't simply walk in there, it is rather the point where the political party meets. Maybe they organize some public Q&A hours, maybe the don't.

It may be confusion of lumping together places with public outreach where one may meet representative of a politician or sometimes actually one of the politicians. Random example (top result of "biuro poselskie"): http://www.aleksandermrowczynski.org/biuro-poselskie/

But probably there are some handling internal stuff not open to public.

Not sure how this two types are distinguished in tagging (I posted question on tagging mailing list, I think that I would mark it with `access` tag).

> Also, from my experience, therapists do not have a reception, thus, it's not walk-in and appointments are made via phone or internet.

I have also fortunately limited experience here but there are therapist offices with reception (but bigger ones like that are far more noticeable).

matkoniecz:
I will probably end with `OFFICES_VISITED_BY_GENERAL_PUBLIC` and `OFFICES_WITH_WALK_IN_BY_GENERAL_PUBLIC` to keep both office lists in the same place.

westnordost:
Yeah but then putting it in an application wide constant is almost for naught

matkoniecz:
After thinking over it again - single use application-wide constant is weird, but worth keeping two very similar tag lists together.

app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.java
```diff
@@ -22,4 +22,13 @@
...
 
 	public static final String SURVEY_MARK_KEY = "check_date";
+
+	public static final String[] OFFICES_VISITED_BY_GENERAL_PUBLIC = {
+		"insurance", "government", "travel_agent", "tax_advisor", "therapist", "religion",
+		"lawyer", "estate_agent", "political_party"
+	};
+
+	public static final String[] OFFICES_WITH_WALK_IN_BY_GENERAL_PUBLIC = {
+		"insurance", "government", "travel_agent", "tax_advisor", "therapist", "religion",
```
westnordost:
```suggestion
		"insurance", "government", "travel_agent", "tax_advisor", "religion"
```

PR #1526, opened 2019-08-19
support popular craft values and more office values

### LINKED ISSUE #1525: support craft=tailor and other craft=* for OH, wheelchair and name quests
I will do it after #1522 (assuming that supporting craft=* will not be rejected)

Noticed thanks to Vespucci validator.

matkoniecz:
fixes #1525

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -49,17 +49,24 @@ class AddPlaceName(
                 "amusement_arcade", "adult_gaming_centre", "tanning_salon", "horse_riding"
             ),
             "office" to arrayOf(
+                // also listed for AddOpeningHours and AddWheelchair quests
                 "insurance", "government", "travel_agent", "tax_advisor", "religion", "employment_agency",
+                // also listed for AddWheelchair quest
                 "lawyer", "estate_agent", "political_party", "therapist",
+                // ones below not listed for other quests
                 "company", "lawyer", "educational_institution", "telecommunication",
                 "administrative", "ngo", "association", "it", "accountant", "research", "architect",
                 "financial", "newspaper", "advertising_agency"
             ),
             "craft" to arrayOf(
-                "carpenter", "shoemaker", "tailor", "winery", "photographer", "electrician",
-                "brewery", "metal_construction", "plumber", "confectionery", "caterer", "hvac",
-                "window_construction", "dressmaker", "electronics_repair", "handicraft", "gardener",
-                "stonemason", "painter", "glaziery", "key_cutter", "beekeper", "blacksmith",
+                // also listed for AddOpeningHours and AddWheelchair quests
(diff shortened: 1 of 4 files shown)


westnordost:
Why would painters, window construction workshops, caterers, hvac, blacksmiths, upholsterer etc. have no name? Summarizing:
- if the place is open for walk-in by general public -> ask for opening hours
- if the place is either the above or at least public in the frame of public events -> ask for wheelchair accessibility
- if the place like this usually or at least sometimes has a name -> ask for a name

I don't think your categorization with the OsmTaggings works here. Probably, almost any "craft" thing will have a name. In the case of crafts, the list for opening hours and wheelchair accessibility should almost be the same.

westnordost:
Actually, it is the question whether not much more if not almost all `office` keys would usually have a name

matkoniecz:
I pushed some commits, but are untested (and tag list would benefit from fresh look)

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -48,7 +48,27 @@ class AddPlaceName(
...
+                // also listed for AddOpeningHours and AddWheelchair quests
+                "carpenter", "shoemaker", "tailor", "photographer", "dressmaker",
+                "electronics_repair", "key_cutter", "stonemason",
+                // also listed for AddPlaceName quest
+                "winery", "electrician", "plumber",
+                // ones below not listed for other quests
+                "metal_construction", "confectionery", "caterer", "hvac", "window_construction",
+                "handicraft", "gardener", "painter", "glaziery", "beekeper", "blacksmith",
+                "roofer", "upholsterer", "pottery"
+            )
```
westnordost: or maybe any `craft`? The same is also done for `shop` further up.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -33,7 +33,19 @@ class AddWheelchairAccessBusiness(o: OverpassMapDataDao) : SimpleOverpassQuestTy
...
+                "insurance", "government", "travel_agent", "tax_advisor", "religion", "employment_agency",
+                // also listed for AddPlaceName quest
+                "lawyer", "estate_agent", "political_party", "therapist"
+            ),
+            "craft" to arrayOf(
+                // also listed for AddOpeningHours and AddPlaceName quests
+                "carpenter", "shoemaker", "tailor", "photographer", "dressmaker",
+                "electronics_repair", "key_cutter", "stonemason",
+                // also listed for AddPlaceName quest
+                "winery", "electrician", "plumber"
```
westnordost: Not sure why you include plumber and electrician. Wouldn't you then also need to include hvac (etc etc)?

matkoniecz:
I am thinking about excluding shop/craft=vacant/no

This problem is basically not appearing for craft and quite rare for office so I decided to not include this checks.

westnordost:
Ok, looks good!

PR #1533, opened 2019-08-23
do not ask about names of private roads

### LINKED ISSUE #1529: Stop asking about names of private roads
For extreme example scan for quests near Vatican - StreetComplete finds several nearly unsurveyable quests in Papal Gardens, like asking for name of https://www.openstreetmap.org/way/604750625#map=19/41.90407/12.44958

matkoniecz:
yes, some cases of useful quests will disappear but reducing rate of unsolvable quests is more important
fixes #1529

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
@@ -48,7 +48,7 @@ class AddRoadName(
         OverpassQLUtil.getQuestPrintStatement()
 
     /** return overpass query string to get roads with names near roads that don't have names
-     *  private roads ae not filtered out here, partially to reduce complexity but also
+     *  private roads are not filtered out here, partially to reduce complexity but also
      *  because the road may have a private segment that is named already or is close to a road
      *  with a useful name
      * */
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
@@ -42,8 +42,8 @@ class AddRoadName(
         OverpassQLUtil.getGlobalOverpassBBox(bbox) +
         ROADS_WITHOUT_NAMES + "->.unnamed;\n" +
         "(\n" +
-        "  way.unnamed['access'!~'private|no'];\n" +
-        "  way.unnamed['foot']['foot'!~'private|no'];\n" +
+        "  way.unnamed['access' !~ '^private|no$'];\n" +
+        "  way.unnamed['foot']['foot' !~ '^private|no$'];\n" +
         "); " +
         OverpassQLUtil.getQuestPrintStatement()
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
```diff
@@ -40,10 +40,18 @@ class AddRoadName(
...
+        "(\n" +
+        "  way.unnamed['access'!~'private|no'];\n" +
+        "  way.unnamed['foot']['foot'!~'private|no'];\n" +
+        "); " +
         OverpassQLUtil.getQuestPrintStatement()
 
-    /** return overpass query string to get roads with names near roads that don't have names */
+    /** return overpass query string to get roads with names near roads that don't have names
+     *  private roads ae not filtered out here, partially to reduce complexity but also
+     *  because road may have private segment that is named already or be close to road
```
westnordost:
```suggestion
     *  because the road may have a private segment that is named already or is close to a road
```

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
```diff
@@ -40,10 +40,18 @@ class AddRoadName(
...
+     *  with an useful name
```
westnordost:
```suggestion
     *  with a useful name
```

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
```diff
@@ -40,10 +40,18 @@ class AddRoadName(
     /** returns overpass query string for creating the quests */
     private fun getOverpassQuery(bbox: BoundingBox) =
         OverpassQLUtil.getGlobalOverpassBBox(bbox) +
-        ROADS_WITHOUT_NAMES + "; " +
+        ROADS_WITHOUT_NAMES + "->.unnamed;\n" +
+        "(\n" +
+        "  way.unnamed['access'!~'private|no'];\n" +
+        "  way.unnamed['foot']['foot'!~'private|no'];\n" +
```
westnordost: FYI `'` are only necessary if the key/value contains characters that are not a-z.

Also, use '^private|no$' for the regular expression, the regex operator in OQL ist a "find in", not a "matches whole string".

matkoniecz: I though about making it match solely `no` and `private` but I decided to make it match also stuff like `access=yes;no`.

I probably should look deeper into what is generated by quests with similar restrictions and follow this.

matkoniecz: And I added `'` because in past was burned by Overpass API deciding to start crashing on previously accepted but illegal syntax, so I prefer to be more careful (but I may also change it).

matkoniecz:
> You really need to proofread what you write in English, regarding the articles.

Sorry :(

> Also, did you test it - on the Vatican?

Yes. I now also fixed incorrect data there that was causing some roads to still appear as quests there.

Issue #1534, opened 2019-08-23
Deauthorisation when tapping "authorise OSM access" while being authorised


LimaKilo:
When the device is authorised and you tap on the "authorise OSM access" menu entry, the device gets deauthorised and OSM opens for a new authorisation. It should either ask for confirmation to do that or have a description that it will do this. I tapped on it because I thought it might show my profile.

**Versions affected**
Android 6, Streetcomplete Version 13.0

westnordost:
And what's the problem with logging in again?

LimaKilo:
> 
> And what's the problem with logging in again?

It's annoying. It's just a minor thing, but a button that logs me out should say so.

westnordost:
Well ok, I don't mind if someone wants to implement this.

typebrook:
@westnordost 
A single alert dialog should be fine, check #1535

PR #1535, opened 2019-08-25
Add deauthorize comfirmation dialog

typebrook:
#1534

When user click on OAuth preference option and OSM account is authorized,
ask if user want to deauthorize it. If not, just do OAuth.

Screenshot_20190825-203737


westnordost:
Cool, thank you! I have a few comments, but I will just do this myself after the merge. Particularly, I want to change:

- translation is done in POEditor and then always pulled from there. Will push the new strings to POEdtior and then edit in your strings.
- first time I see `takeIf` in action. Though, in this case, I have the opinion that it does not increase readability (rather the contrary), I'll rather use a simple if-else here.
- I will remove the string `oauth_confirm_deauthroize_username`. Since every string needs to be translated into 20+ languages, I weigh for every string if it is necessary
- I want to improve the English wording

typebrook:
> weigh for every string if it is necessary
> readability

Good points! I'll take care of these rules next time.
Have a nice day!

westnordost:
You too!

PR #1536, opened 2019-08-25
Fix some text issues while proof-reading the architecture file

rugk:
Follow-up of https://github.com/westnordost/StreetComplete/pull/1528

The architecture file had some typos and similar issues.

For the dashes change, refer to https://en.wikipedia.org/wiki/Dash. En dash is the correct one here.

/cc @matkoniecz


westnordost:
Ok

PR #1543, opened 2019-08-29
exclude ruined buildings from quests

### LINKED ISSUE #1541: Exclude ruins=yes - from address and building quest
I propose to start filtering out object with ruins=yes for quests about buildings (building type, address quest, roof shape maybe also building levels)

Such buildings are often not really answerable and value of answers will be low. Large number of questions will end in failure to reply.

Some ruins were built was reasons that are not clear now, a roof may be caved in making the roof shape question unanswerable (and not really fixable by tagging it), building levels are weird at the best (is 0 correct in case of completely missing roof) etc.

Though I admit that the real reason is that at least in some places many object could have historic=* tag, are missing it but ruins=yes is present.

images

matkoniecz:
Such buildings are often not really answerable and value of answers will be low. Large number of questions will end in failure to reply.

Some ruins were built was reasons that are not clear now, a roof may be caved in making the roof shape question unanswerable (and not really fixable by tagging it), building levels are weird at the best (is 0 correct in case of completely missing roof) etc.

Though I admit that the real reason is that at least in some places many object could have historic=* tag (that is already filtering out objects in building type quest), but historic key is untagged. Many of such objects have ruins=yes.

fixes #1541

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
@@ -17,7 +17,7 @@ class AddBuildingLevels(o: OverpassMapDataDao) : SimpleOverpassQuestType<Buildin
             "commercial","office","warehouse","industrial","manufacture","parking","farm",
             "farm_auxiliary","barn","cabin").joinToString("|") + "\n" +
         " and !building:levels and !height and !building:height\n " +
-        " and !man_made and location!=underground\n "
+        " and !man_made and location!=underground and ruins!=yes\n "
     override val commitMessage = "Add building and roof levels"
     override val icon = R.drawable.ic_quest_building_levels
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
@@ -11,7 +11,7 @@ class AddBuildingType (o: OverpassMapDataDao) : SimpleOverpassQuestType<String>(
     // information about the purpose of the building, so no need to force asking it
     override val tagFilters = """
         ways, relations with building = yes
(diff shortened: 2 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
```diff
@@ -17,7 +17,7 @@ class AddBuildingLevels(o: OverpassMapDataDao) : SimpleOverpassQuestType<Buildin
             "commercial","office","warehouse","industrial","manufacture","parking","farm",
             "farm_auxiliary","barn","cabin").joinToString("|") + "\n" +
         " and !building:levels and !height and !building:height\n " +
-        " and !man_made and location!=underground\n "
+        " and !man_made and location!=underground and ruins!=yes\n "
```
westnordost: Why not this?

matkoniecz: https://www.openstreetmap.org/way/131358967 is an extreme example and one of things that prompted my patch eliminating ruins=yes - just wall part remaining from something that was ancient marketplace/mall and SC will now happily create quest about it. 

And what worse there is no tagging that will get rid of it, in this case even adding historic=* will not help and given no access to it is not viable to tag with `height` (maybe there is some official data/3D model on a suitable licence?).

But even without such ridiculous cases - there are often building in advanced decay where floor count is far from obvious and wasting time on tagging them is likely to be pointless.

westnordost: I'd say that floor count is the **current** floor count, not one that might have been one time.

[But looking at the site on google maps 3d](https://www.google.de/maps/@41.891174,12.4867434,51a,35y,82.29h,61.68t/data=!3m1!1e3), I understand the problem.

Basically, I'd say the tag is wrong - it is not a building. It is a piece of wall. But I guess there are no correct tags for this maybe? Anyway, I rest my case. Let's exclude ruins from these quests.

PR #1544, opened 2019-08-29
ask about footway part of segregated way

### LINKED ISSUE #1489: Ask about cycleway/footway surface

matkoniecz:
together with #1493 fixes #1489

-----


matkoniecz:
Note - it has no pretty special icon. I may try adding one (but it is unlikely to be great, even if made from already available parts).

westnordost:
Ok I made you an icon

matkoniecz:
Thanks!

matkoniecz:
version with icon available

PR #1560, opened 2019-09-14
No max height for private parkings

HolgerJeromin:
I am not sure if this is ok, but I got a quest for this one
https://www.openstreetmap.org/node/4844325991


MegaArthur:
Why not?

matkoniecz:
See https://github.com/westnordost/StreetComplete/pull/960#issuecomment-372016986

PR #1571, opened 2019-09-18
Final fix of missing quest

matkoniecz:
TODO: what 

```
		{
			if(!isShowingControls)
			{
				hideAll(leftSide, -1);
				hideAll(rightSide, +1);
			}
		});
```

code is actually doing? I packaged it into a separate function but it is quite poorly named as I am not sure what it is supposed to do from reading the code. It was added in https://github.com/westnordost/StreetComplete/commit/7f2f1399e8994a8a80383a511a897bf03be8dbc0

I can try testing what will happen on removing it but maybe you remember why it was added.

-----

This is #1500 with a tiny improvements in https://github.com/westnordost/StreetComplete/commit/74cf89c5347df5dcaf02c53bd570d855a1a5439d

Syntax is a bit flustrating but it can be eliminated on migrating to Kotlin.

Thank you for pointing me in this direction, hopefully with this changes fix is really fixing it in all cases and not just on my phone.


westnordost:
👍 

How about `hideControls` for the function name?

matkoniecz:
Already used for

```
	public void hideControls()
	{
		isShowingControls = false;
		animateAll(rightSide, +1, false, 120, 200);
		animateAll(leftSide, -1, false, 120, 200);
	}
```

westnordost:
ah right, well, could you perhaps rename hideControls to animateHideControls then?

westnordost:
Ah nevermind, I will change namings etc anyway once I convert it to kotlin

Issue #1385, opened 2019-04-28
Add whether shop=laundry is a self-service laundry


westnordost:
### General
Affected tag(s) to be modified/added: [self_service](https://wiki.openstreetmap.org/wiki/Key:self_service)
Question asked: **Is this a self-service laundry?**

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary: Coin-operated washing-machines can usually be seen from the outside
- [?] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort): Either taginfo is broken, or almost no laundry shops have self_service tagged right now

### Implementation
A yes-no quest or like the car-wash quest if there should be more options.

majkaz:
This seems to be a very location specific quest. 

Locally, the test "Not an overwhelming percentage of elements have the same answer (No spam)" will fail - except in very touristy areas, where you might find the one or other odd self-service laundry, any other such shop operates more or less like a dry cleaning - you drop your laundry there, and get a clean (and pressed) one back later.

matkoniecz:
> except in very touristy areas

And in some cities where washing machine is not present in all rooms/apartments.

westnordost:
Well in my city usually every apartment has one, still, there are about as many self-service laundries scattered around town as laundry shops.

majkaz:
Well, in my country, every apartment has one... and there are self-service laundries in the capital (touristy area and expats). In two additional towns (with big universities) there are few self-service laundries. In the rest of the country, there might be the odd one... 

In my region (which has about 1/10 of the whole country area), there isn't a single one, except the one I have found in a hostel, just for the guests... This is what I mean with a "location-specific" quest.

But never mind the issue, there are not many laundries either - neither on the ground, nor mapped - just dry cleaners. The quest won't matter here. "Our" equivalent of shop=laundry is traditionally just the ironing / pressing part. Not much changed here.

PR #1583, opened 2019-09-29
Reorder some crosswalk related quests

peternewman:
So they appear in a logical order when completing them.

I was tempted to move AddTrafficSignalsButton up too, above the others, but I appreciate it's less useful.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
@@ -152,12 +152,12 @@ public class QuestModule
 				new AddToiletsFee(o), // used by OsmAnd in the object description
 				new AddBabyChangingTable(o), // used by OsmAnd in the object description
 				new AddBikeParkingCover(o), // used by OsmAnd in the object description
-				new AddTrafficSignalsSound(o),
+				new AddTactilePavingCrosswalk(o), // Paving can be completed while waiting to cross
+				new AddTrafficSignalsSound(o), // Sound needs to be done as or after you're crossing
 				new AddRoofShape(o),
 				new AddWheelchairAccessPublicTransport(o),
 				new AddWheelchairAccessOutside(o),
 				new AddTactilePavingBusStop(o),
-				new AddTactilePavingCrosswalk(o),
 				new AddBridgeStructure(o),
 				new AddReligionToWaysideShrine(o),
 				new AddCyclewaySegregation(o),


westnordost:
Ok, makes sense to group them together

peternewman:
Thanks for merging.

matkoniecz:
And thanks for making a PR

matkoniecz:
@peternewman Are you maybe interested in contributing more?

https://github.com/westnordost/StreetComplete/issues/1385 is an idea for a new quest - not the most exciting one, but relatively easy to implement.

If you are interested in adding it and have time, but you are unsure how something like that can be done, let me know and I will make documentation/tutorial about making new quests.

peternewman:
> @peternewman Are you maybe interested in contributing more?
> 
> #1385 is an idea for a new quest - not the most exciting one, but relatively easy to implement.
> 
> If you are interested in adding it and have time, but you are unsure how something like that can be done, let me know and I will make documentation/tutorial about making new quests.

Hi @matkoniecz ,

Thanks for the invite/suggestion. I think I might stick to just being a user/translator (to British English, so also not very exciting) of StreetComplete, I'm already involved in quite a few other open source projects which I could do with spending more time on, without adding more to the list. I'd probably also prefer to avoid having to deal with yet another development environment to compile and test stuff (this change was just made in the GitHub web based editor as you might have guessed). I couldn't immediately see any compilation instructions either, or is it fairly trivial if you've made another Android app in the past?

I wouldn't mind beta-testing new versions of the app if there is a way to get them as an APK or via the App Store?

westnordost:
> is it fairly trivial 

Yes, you just need to [download and install Android Studio](https://developer.android.com/studio/) which comes bundled with all tools needed, checkout and open the project in this application and click on the green play button on the top.

> I wouldn't mind beta-testing new versions of the app if there is a way to get them as an APK or via the App Store?

I publish beta versions on Google Play, you'd just need to click a button on the store page to enter the open beta. (Though, I am thinking about changing this model to an invite-via-lonk betatest model because there are over 5000 betatesters right now)

peternewman:
> > is it fairly trivial
> 
> Yes, you just need to [download and install Android Studio](https://developer.android.com/studio/) which comes bundled with all tools needed, checkout and open the project in this application and click on the green play button on the top.

Thanks, I've added that in https://github.com/westnordost/StreetComplete/pull/1603 .

I'll probably pass on trying it out for now thanks, for the reasons mentioned above.

> > I wouldn't mind beta-testing new versions of the app if there is a way to get them as an APK or via the App Store?
> 
> I publish beta versions on Google Play, you'd just need to click a button on the store page to enter the open beta. (Though, I am thinking about changing this model to an invite-via-lonk betatest model because there are over 5000 betatesters right now)

Ah thanks, that seems to only appear on the device, not on the Play Store website. I'm now signed up as a beta tester anyway.

PR #1596, opened 2019-10-07
Fixed capitalisation in strings.xml

MegaArthur:
All other strings that appear when the user presses "Other answers" are capitalised, this one wasn't.

### DIFF
--- a/app/src/main/res/values-en-rGB/strings.xml
+++ b/app/src/main/res/values-en-rGB/strings.xml
@@ -369,7 +369,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_buildingLevels_roofLevelsLabel2">"levels in the roof"</string>
     <string name="quest_buildingLevels_levelsLabel2">"regular levels (not counting the roof)"</string>
     <string name="pref_title_authorized_summary">"Authorised"</string>
-    <string name="quest_buildingLevels_answer_multipleLevels">"differs per building part"</string>
+    <string name="quest_buildingLevels_answer_multipleLevels">"Differs per building part"</string>
     <string name="quest_buildingLevels_answer_description">"In this case, simply input the values of the highest building part."</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -369,7 +369,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_buildingLevels_roofLevelsLabel2">"levels in the roof"</string>
     <string name="quest_buildingLevels_levelsLabel2">"regular levels (not counting the roof)"</string>
     <string name="pref_title_authorized_summary">"Authorised"</string>
-    <string name="quest_buildingLevels_answer_multipleLevels">"differs per building part"</string>
+    <string name="quest_buildingLevels_answer_multipleLevels">"Differs per building part"</string>
```
HolgerJeromin: I think these change should go through the normal localization pipeline

app/src/main/res/values/strings.xml
```diff
@@ -409,7 +409,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_cycleway_value_lane_soft">cycle lane (dashed markings)</string>
     <string name="quest_cycleway_value_none">none</string>
     <string name="quest_cycleway_value_shared">lane shared explicitly with other traffic</string>
-    <string name="quest_buildingLevels_answer_multipleLevels">differs per building part</string>
+    <string name="quest_buildingLevels_answer_multipleLevels">Differs per building part</string>
```
rugk: well this change is okay AFAIK (you cannot change the origin strings in poeditor)

But as for all translations, see https://github.com/westnordost/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app

MegaArthur:
Is it really worth it going through the whole translation process just to change one letter in values-en-rGB though?

westnordost:
Okay, thanks!

PR #1600, opened 2019-10-08
add general fee quest

### LINKED ISSUE #873: General Fee Quest
### General
_Currently, there are quests to complete whether there is a fee for toilets or a car parks.
But there are many other OSM objects where the `fee` tag could be used._

Affected key(s) (or tags) to be modified: [fee](https://wiki.openstreetmap.org/wiki/Key:fee)
Question asked: **Does it cost something to use this [object type]?**

### Checklist

### Ideas for implementation
**Element selection:** (see list of objects below) with `[!fee]`

**Metadata needed:** Not necessarily, but maybe some objects could be disabled in some countries if there is a strong tendency for things to be (not) free so that it is obsolete to ask it

**Proposed GUI:** Simply a yes/no form as it already exists for the 2 current fee quests.

### (Incomplete) List of POIs possibly with fee:
- Museums: `tourism=museum`
- Attractions: `tourism=attraction`
- Theme parks: `tourism=theme_park` (could be "no" if it is to be paid at the particular attractions)
- Galleries: `tourism=gallery`
- Viewpoints: `tourism=viewpoint`
- Public gardens: `leisure=garden`
- Beach resorts: `leisure=beach_resort`
- Charging stations: `amenity=charging_station`
- Newspaper vending "machines": `[amenity=vending_machine][vending=newspapers]`
- Charging stations (for mobile phones etc.): `amenity=device_charging_station`
- and of course `amenity=toilets` and `amenity=parking`

matkoniecz:
fixes #873

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
@@ -1,0 +1,24 @@
+package de.westnordost.streetcomplete.quests.general_fee
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+class AddGeneralFee(o: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(o) {
+
+    override val tagFilters = "nodes, ways, relations with " +
+            "(tourism=museum or leisure=beach_resort or tourism=gallery) " +
+            "and access !~ private|no and !fee and name"
+    override val commitMessage = "Add fee info"
+    override val icon = R.drawable.ic_quest_money
+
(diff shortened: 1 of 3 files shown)


HolgerJeromin:
Correct issue number is #873

app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+class AddGeneralFee(o: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(o) {
+
+    override val tagFilters = "nodes, ways, relations with (tourism=museum or leisure=beach_resort or tourism=gallery) and access !~ private|no and !fee and name"
```
HolgerJeromin: IIRC other quests have line breaks in it to get clean diffs in the future.

matkoniecz:
Should be fixed now, thanks for a review!

westnordost:
Looks good, thank you!

Issue #1606, opened 2019-10-18
Marker re-initialization with since update


typebrook:
https://github.com/westnordost/StreetComplete/blob/d72c22303fd8edb3652c6ed68f3eb74a923ca200/app/src/main/java/de/westnordost/streetcomplete/tangram/MapFragment.java#L454

When `MainActivity` is focused, it continuously calls this method. According to the name `onRotationChanged`, it should only be triggered when map is rotating. 

Because in this method we set style for Marker, it could potentially causes APP crash when doing [`MapController.updateSceneAsync()`](https://tangrams.readthedocs.io/en/latest/API-Reference/android-sdk/0.10.0/com/mapzen/tangram/MapController.html#updateSceneAsync-java.util.List-) in MapFragement (though we don't do it now) with error message below:

```
java.lang.RuntimeException: Tried to perform an operation on an invalid id! 
This means you may have used an object that has been disposed and is no longer valid.
```
**How to Reproduce**
v14.1 and just log this method

**Versions affected**
v14.1

westnordost:
The method is called by the compass component. It actually means, that the **smartphone** was rotated/tilted, not the map. The callback for when the map is rotated, is `onMapOrientation`.
If this is confusing, I'd suggest to refactor the class to not implement all those listener interfaces itself, but use private anonymous SAM classes / lambdas for this, i.e. in line 158:

```
compass.setListener((rotation, tilt) -> {
    //...
});
```

I am kind of hold of on any (refactor) work on `MapFragment` etc. until an upgrade of the tangrams-es dependency 0.10.x becomes necessary.

---

Anyway, the naming of the mentioned method is not the main issue you address. It is a pity that tangram-es just invalidates (=removes) all markers on each scene update.
I think the best solution here would be to never call `MapController.updateSceneAsync()` directly but instead wrap it in a function like this:
```
protected int updateSceneAsync(List<SceneUpdate> sceneUpdates) {
    // TODO: set all markers to null
    controller.updateSceneAsync(sceneUpdates);
    // TODO: reinitialize all markers, as applicable
}
```
Because `QuestsMapFragment` also uses markers, the implementation to reinitialize all the markers is not trivial. Probably it will be necessary to have a wrapper object for every marker that will not be invalidated through a `updateSceneAsync` call and is able to reinitialize its marker if asked to.

typebrook:
Thanks for the explanation! looks like I misunderstood this method.

[...]

Issue #1589, opened 2019-09-30
3D buildings should be disabled when creating notes


ltog:
If one wants to place a note in an area where OSM data contains building heights and the buildings in this area are tall, it's hard to properly place the marker for creating a note, especially on the northern side of buildings.

I therefore propose (if it's possible) to disable the rendering of 3D buildings while one places the marker for notes.

**How to Reproduce**
see above

**Versions affected**
all (?)

ltog:
@MegaArthur : Have you actually tried it?

HolgerJeromin:
@MegaArthur if this is so easy: 
Where will be the note? On the building corner or in the building?
![rps20191001_075634_087](https://user-images.githubusercontent.com/2410353/65937727-17735980-e421-11e9-812f-f025a6fd2223.jpg)

westnordost:
To understand better the use case, can you tell me  what this note would be about?

ltog:
It might be a note about a missing (complicated) opening hours (including a photo) or about a missing POI (sometimes including a photo) in general. When strolling around with a tired/bored young child it's often easier (or the only possibility) to take a photo rather than mapping a feature immediately. Disabling the 3D view would allow to place a note more precisely since in 3D view one cannot be sure where a note exactly will be placed (as elegantly shown by @HolgerJeromin in https://github.com/westnordost/StreetComplete/issues/1589#issuecomment-536879430 ).

MegaArthur:
Oh, sorry, I didn't know that higher buildings 

> @MegaArthur if this is so easy:
> Where will be the note? On the building corner or in the building?
> ![rps20191001_075634_087](https://user-images.githubusercontent.com/2410353/65937727-17735980-e421-11e9-812f-f025a6fd2223.jpg)

Well, from experience in sattelite mapping tall buildings it should be near the corner.
I forgot that SC is meant to be used without requiring any prior mapping experience/knowledge, so I might agree that flattening buildings could be useful.

HolgerJeromin:
> Well, from experience in sattelite mapping tall buildings it should be near the corner.

I am 99% sure it will be inside the building :-)

rugk:
> flattening buildings could be useful

If this is done with some fancy animation of buildings disappearing into the ground (when you open the "notes" view/add a note), then this would even be kinda cool. :sunglasses: :wink:

typebrook:
@rugk
[...]

PR #1607, opened 2019-10-18
Disable 3D buildings when creating notes

typebrook:
To solve #1589, this PR use [`MapController.updateSceneAsync()`](https://tangrams.readthedocs.io/en/latest/API-Reference/android-sdk/0.10.0/com/mapzen/tangram/MapController.html#updateSceneAsync-java.util.List-) to disable 3D buildings at runtime, includes the following changes in order:

1. 2ec0366
Add method `onSceneUpdate()` to clean added Markers before doing scene update, or APP just crash.
This is another issue explained in #1606.

1. 342af32
Add method `toggle3DBuildings(boolean)` into `MapFragment`. So it could be used to disable/enable 3D building draws when `CreateNoteFragment` is in/out. Because the use case in #1589 is to make sure where to put a new note, so buildings draws should not be invisible, but turn off extrude to show its shape.
output2

1. 81bb301
Add animation when toggling buildings draws. Because [transition](https://tangrams.readthedocs.io/en/latest/Syntax-Reference/draw/#transition) cannot be applied to polygon. To make animation, the way is to update `shaders` value to shrink building height. Also 3 lines were added into scene file, so map-style repo need to be updated.
output

---
- Be aware these code changes redraw the map when clicking create-note button, this may affects user experience
- I am not sure use animation and change scene file is a good idea. So if making 3D buildings disabled is good enough, just merge 342af32
- If animation needs enhancement, take a look at tangram doc about [shaders](https://tangrams.readthedocs.io/en/latest/Syntax-Reference/shaders/) and [WebGL API](https://www.khronos.org/files/webgl/webgl-reference-card-1_0.pdf)

### DIFF
--- a/app/src/main/assets/map_theme/layers/land.yaml
+++ b/app/src/main/assets/map_theme/layers/land.yaml
@@ -110,9 +110,6 @@ styles:
     buildings-style:
         base: polygons
         blend: opaque
-        shaders:
-            blocks:
-                position:
     buildings-outline-style:
         base: lines
         blend: inlay
--- a/app/src/main/assets/map_theme/layers/land.yaml
+++ b/app/src/main/assets/map_theme/layers/land.yaml
@@ -110,6 +110,9 @@ styles:
     buildings-style:
         base: polygons
         blend: opaque
+        shaders:
+            blocks:
+                position:
     buildings-outline-style:
         base: lines
         blend: inlay
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -948,7 +948,7 @@ public class MainActivity extends AppCompatActivity implements
 	{
 		freezeMap();
 		showInBottomSheet(new CreateNoteFragment());
-		mapFragment.toggle3DBuildings(false);
+		mapFragment.setShow3DBuildings(false);
 	}
 
 	/* ---------------------------------- VisibleQuestListener ---------------------------------- */
@@ -998,7 +998,7 @@ public class MainActivity extends AppCompatActivity implements
 			((IsCloseableBottomSheet)f).onClickClose(() ->
 			{
 				getSupportFragmentManager().popBackStackImmediate(BOTTOM_SHEET, FragmentManager.POP_BACK_STACK_INCLUSIVE);
-				mapFragment.toggle3DBuildings(true);
+				mapFragment.setShow3DBuildings(true);
 				unfreezeMap();
 			});
 		}
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -948,6 +948,7 @@ public class MainActivity extends AppCompatActivity implements
 	{
 		freezeMap();
 		showInBottomSheet(new CreateNoteFragment());
+		mapFragment.toggle3DBuildings(false);
 	}
 
 	/* ---------------------------------- VisibleQuestListener ---------------------------------- */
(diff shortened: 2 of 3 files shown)


westnordost:
Wow, that is great! I wouldn't have thought, that an implementation for this could be so straightforward! Thank you for this contribution! @ltog : Did you see this PR?

I have three notes about the PR:

1. If it does not cause an inconvenience, I'd prefer if you renamed `toggle3DBuildings` to `setShow3DBuildings`, just to be more consistent with the naming in the rest of the code. But this is not really important, I'd merge your PR in any case
2. Regarding the animation, I'd very much prefer if there was no animation because:
    - I'd like to keep the code in `MapFragment` as short as possible to keep the class clearly arranged, partly because it already is quite big, one of the largest classes in the project IIRC.
    - The animation would (probably) only look good if its duration would be about the same length and starts at the same time as the "note pin falls down" animation. But the scene update already takes longer then the fall down animation. So, looking at your GIF, there is a visible switch when clicking the button, but then also later through the buildings sinking into the ground. This looks confusing because due to the delay there is less of a cause-and-effect relation between clicking the button and that animation.
3. On the GIF, it looks as if tangram-es needs to download the tiles anew on a scene update. Is this true? (You can test this by cutting the internet connection just before pressing the create-note button) If it is, this might be an issue (and IMO a bug in tangram-es) because clicking that button should not have that disruptive consequences. (Downloading the tiles is sometimes really slow)

westnordost:
I realized that #1606 must be solved in the frame of this PR. As it is now, the markers for the view direction, position and accuracy will get removed on clicking on the create note button and will be gone after that.

ltog:
@typebrook : Wow! This is awesome! I'm looking forward to see this in action on my phone! Thank you for your effort!

typebrook:
@westnordost 
Thanks for the suggestion, I think latest commits meet your requirement and is ready to be merged.

---
>  I'd very much prefer if there was no animation  because I'd like to keep the code in MapFragment as short as possible.
Well, both **disable building extrude** and **add animation to shrink height** only costs one line to add a new `SceneUpdate`. So there is no difference.

But I agree with other points:
- Make `MapFragment` as short as possible:
Object about map easily grows dramatically with new features time by time, and comes with poor readability and huge complexity of code change. (My story is take over a project with 4K-lines Mapbox MapFragment , just a disaster...) Maybe later we can move some logic to another file by using Kotlin [extension function](https://kotlinlang.org/docs/reference/extensions.html), or one more child class between `MapFragment` and `QuestsMapFragment`?

- Remove animation:
Since map redraw, a short time with blank screen is not preventable, that's why I delay the animation for 0.7s. If animation cannot goes with "note pin falls down", keep it may confuse user.
Sorry @ltog, no cool animation:smile:, but at least disable 3D buildings is fine now.

Also
> On the GIF, it looks as if tangram-es needs to download the tiles
Don't worry, tile cache works fine with Scene update. Map redraw works on every tiles one by one, so it is normal to take some time.

westnordost:
> Object about map easily grows dramatically with new features time by time, and comes with poor readability and huge complexity of code change. (My story is take over a project with 4K-lines Mapbox MapFragment , just a disaster...) Maybe later we can move some logic to another file by using Kotlin extension function, or one more child class between MapFragment and QuestsMapFragment?

Absolutely, there has to be a way to split the functionality further. But as said, this would need to be done after upgrade to tangram-es 10.x because that upgrade itself necessitates some more wrapper code. Things we can do after that:

1. convert to Kotlin -> a little bit shorter
2. add a subclass, so we have 1. (convenience) wrapper class around MapView/MapController, 2. subclass with location marker + compass, 3. QuestsMapFragment. The first will still be the largest class.
3. use more delegation. For example, after the 10.x upgrade, there should probably be a class that manages animations on the map

westnordost:
I already did review it, but put the comment into the wrong issue: https://github.com/westnordost/StreetComplete/issues/1606#issuecomment-544267802

PR #1608, opened 2019-10-19
ask whatever laundry is self service

### LINKED ISSUE #1385: Add whether shop=laundry is a self-service laundry

matkoniecz:
fixes #1385

@westnordost Are you maybe planning icon for this quest? I can try finding/making one but your is going to be much better :)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.java
@@ -55,6 +55,7 @@ import de.westnordost.streetcomplete.quests.religion.AddReligionToWaysideShrine;
 import de.westnordost.streetcomplete.quests.localized_name.data.PutRoadNameSuggestionsHandler;
 import de.westnordost.streetcomplete.quests.localized_name.data.RoadNameSuggestionsDao;
 import de.westnordost.streetcomplete.quests.segregated.AddCyclewaySegregation;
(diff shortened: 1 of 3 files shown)


westnordost:
I don't expect that there are two laundries right next to each other, so I'd prefer if there was only one string to translate. (The one without the name.)

westnordost:
Added a laundry icon

matkoniecz:
> I don't expect that there are two laundries right next to each other

I was thinking about a bit different use case.

In case of place that is tagged as `shop=laundry` and is now say a restaurant, user may answer to

> Is this a self-service laundry?

"no", as in "this is not a laundry".

But maybe it is still not worth translating two strings?

westnordost:
The wording could be "is this laundry a self service laundry"

matkoniecz:
Thanks for the icon! I also applied suggested phrasing.

westnordost:
Looks good, thanks!

PR #1611, opened 2019-10-20
fix filtering out weird footways

matkoniecz:


westnordost:
Dang, I'm such an idiot.

westnordost:
Though, I will solve this differently. See:
- https://taginfo.openstreetmap.org/keys/conveying#values
- https://taginfo.openstreetmap.org/keys/indoor#values

So I will write ` and (!conveying or conveying = no) and (!indoor or indoor = no) `

PR #1613, opened 2019-10-20
do not require surface=paved for footway/cycleway part quest (fixes #1587)

### LINKED ISSUE #1587: Improve AddPathSurface/AddCyclewayPartSurface/AddFootwayPartSurface interplay for segregated cycle ways
**Use case**
When completing AddPathSurface, I've found some segregated cycle ways which don't have the tags StreetComplete is expecting, it offers me the option to split along it's _length_, but not it's _width_.

For example:
https://www.openstreetmap.org/way/576119011
highway=cycleway
segregated=yes

**Proposed Solution**
I think the fix might be two-part:

- Tighten AddPathSurface to exclude segregated=yes
- Loosen AddCyclewayPartSurface to not require surface=paved (which is surely incorrect if the two sides are different). Also remove the requirement for bicycle ~ designated|yes; if it's segregated it implies both bicycle and pedestrian according to the wiki. Perhaps the solution is to check for `bicycle!=no` or even `!bicycle or bicycle ~ designated|yes`?
- Likewise make the equivalent changes in AddFootwayPartSurface

If we should be tagging bicycle/foot designated, it could also add those as part of it's change?

Otherwise reading the code it seems I should AddPathSurface and select the generic paved option, but then I can't see any way to set the bicycle/foot tags on the way.

westnordost:


westnordost:
Which wiki definition do you mean?

peternewman:
> Which wiki definition do you mean?

https://wiki.openstreetmap.org/wiki/Key:segregated

westnordost:
But where does it state that
> no explicit tags for usage of the path are added
?

peternewman:
True, but in the Greater London area at least, 50% of them just have segregated and don't have specific bicycle/foot tags.

I'd only really read the "This key is used for combined cycle- and footways." bit, I guess the tag examples imply you should also tag the usage too.

westnordost:
Well I could add it like this:
```
highway = cycleway 
or (highway ~ path|footway and bicycle != no)
or (highway = bridleway and bicycle ~ designated|yes)
```
meaning, path/footway kind of implies the `segregated` to reference what is being segregated is a cycleway unless explicitly stated otherwise. But, even if correct close to 100% of the times, I'd rather not make these kinds of assumptions based on incomplete data. As far as I know, it is also not done in other places in StreetComplete.

Instead I think it would be better to add a (auto)fix rule to iD, osmose task or similar to fill in this data.

westnordost:
I did a little bit of research with overpass-turbo. Of ways tagged with `highway=cycleway` and `segregated=yes`, these are the percentages where `foot=*` is missing:

Berlin: 5%
London: 49%
Amsterdam: 25%
Paris: 23%
Brussels: 32%
Warsaw: 30%
Stockholm: 2%

London seems to be a bit of an outlier there. For `highway=footway` and `segregated=yes`, the count where `bicycle=*` is missing is about one half to one third of the percentages shown above. So, for cycleways, more people assume that `segregated` implies that it's segregated with pedestrians.

westnordost:
Actually, I changed my mind. The question clearly states "What is the surface of this footway" (or cycleway, for the other quest). So, if the user does not see a foot/cycleway, he will leave a note.

peternewman:
> I did a little bit of research with overpass-turbo. Of ways tagged with `highway=cycleway` and `segregated=yes`, these are the percentages where `foot=*` is missing:
> 
> Berlin: 5%
> London: 49%
> Amsterdam: 25%
> Paris: 23%
> Brussels: 32%
> Warsaw: 30%
> Stockholm: 2%

Interesting that your result is about the same as my more generic `highway=path` and `segregated=yes` and none of foot/bicycle/horse=* for London.

> London seems to be a bit of an outlier there.
Agreed, but apart from Berlin and Stockholm, people are just a bit vague/lazy about tagging what can go on a segregated way.

> For `highway=footway` and `segregated=yes`, the count where `bicycle=*` is missing is about one half to one third of the percentages shown above. So, for cycleways, more people assume that `segregated` implies that it's segregated with pedestrians.

In the UK at least it's illegal to cycle on the pavement, so `highway=footway` and `segregated=yes`, the count where `bicycle=yes` would seem a bit odd and I'd probably tag it as `highway=path` and `segregated=yes`.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -8,7 +8,13 @@ import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
 class AddCyclewayPartSurface(o: OverpassMapDataDao) : SimpleOverpassQuestType<String>(o) {
 
     override val tagFilters = """
-        ways with highway ~ path|footway|cycleway|bridleway and surface = paved and bicycle ~ designated|yes and segregated=yes
+        ways with
+        (
+          highway = cycleway 
+          or (highway ~ path|footway and bicycle != no)
```
peternewman: I assume this will also match `highway=path` with no bicycle key whatsoever?

westnordost: Yes

Issue #1604, opened 2019-10-14
Ignore escalators for the path surface quest


ferdinand0101:
Escalator are usually directly next to steps which leads to false tagging

Ignore steps with the convey tag

matkoniecz:
What is wrong with answering that surface is metal?

ferdinand0101:
1. Yes it's pointless and due to escalors and stairs being usually close together you can not see which is which and that will lead to incorrect tagging

sdicke:
A solution would be not to ask which surface the path has, but to ask something like „Which surface has this escalator?“ and for lifts „Which surface has this lift?“.

ferdinand0101:
Well as matkoniez already mentioned escalators only have surface metal
But I agree on the elevator thing

westnordost:
Good suggestion, I shall remove paths with the convey-tag from the overpass query.

peternewman:
> Well as matkoniez already mentioned escalators only have surface metal

Pedant (although this particular one has gone):
https://londonist.com/2013/09/londons-undergrounds-last-wooden-escalator-to-be-removed

There are also Google hits for Macy's, I don't know if that one has gone too.

PR #1616, opened 2019-10-23
Add handrail quest: ask if steps have a handrail or not

### LINKED ISSUE #1390: quest for steps on handrails
This quest type is useful for accessibility purposes and used into algorythm to estimate accessibility of an itinerary for blind/elderly. 
I know that it was not seen as a priority in issue #503 , but I guess it was waiting for someone to create a dedicated issue on handrails especially (issue 503 was about different quests on staircases, including handrails).
Please tell me what do you think about it.. 
cc @smortex

>> QUEST PROPOSAL (UPDATE)

### General
Affected tag(s) to be modified/added: [handrail](https://wiki.openstreetmap.org/wiki/Key:handrail)
Question asked: **Is there (an) handrail(s) along these steps?**
Answer help : Select the location of the handrail (left, right, middle).

Use case : 
1 I am an elderly, connected, I love to travel, I am still active but have moving issue, I want to know if I can go to this supermarket where there is a few steps, with no concerns..
2 I am a travel agency, I want to show accessible path to my clients. 
3 I am a city planner, I want to make my city more welcoming to people.
### Checklist

### Ideas for implementation

**Proposed GUI:**
Use the same method as for sidewalks ("is there a pavement along this street), user choose the side of steps by clicking on an image/minimap and answer yes or no.

xuiqzy:
Please review my decision and explanation for only including `node`s and excluding only the `handrail` tag.
I tried to add a bit of documentation in the code for others to understand the rationale of how this quest is implemented (what is included/excluded etc), I hope it is done okay.

I wasn't sure about the ordering of the quest in relation to the others, maybe you have some ideas for that.

fixes #1390

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
@@ -1,0 +1,48 @@
+package de.westnordost.streetcomplete.quests.handrail
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataDao
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+/**
+ * Only ask about whether there exists a handrail, not about the location (left, right and/or
+ * middle), because the primary use case here is accessibility of the path (not 3d rendering, etc),
+ * which does not depend on the location of the handrail.
+ * The details would only slow down the answering in a survey by requiring more thought and a more
+ * complex GUI (interpreting the direction of the OSM way for example).
+ *
+ * Github Issue for this quest:
+ * https://github.com/westnordost/StreetComplete/issues/1390
+ */
+class AddHandrail(overpassServer: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(overpassServer) {
+    // Do not include nodes and relations, even though these exist with the right tags, because
+    // according to the wiki page for `highway=steps` [2] it can only be applied to ways. It also
+    // does not make much sense for other types of elements and it is more likely to be a
+    // Do not exclude elements tagged with `handrail:left`, `handrail:center` or `handrail:right`
+    // because according to the wiki page for `handrail=*` [1], the `handrail` tag should
+    // always have just `yes` or
+    // `no` as a value to indicate the existence of a handrail somewhere. If there is a mismatch of
+    // the tags afterwards, that can be recognized using one of the various
+    // Quality Assurance tools [3].
+    // [1] https://wiki.openstreetmap.org/wiki/Key:handrail
+    // [2] https://wiki.openstreetmap.org/wiki/Tag:highway%3Dsteps
+    // [3] https://wiki.openstreetmap.org/wiki/Quality_assurance
+    override val tagFilters = "nodes with highway = steps and !handrail"
+
+    // changeset comment to be used for uploading changes made through this quest
+    override val commitMessage = "Add whether steps have a handrail"
+
+    // icon for displaying quest on the map
+    override val icon = R.drawable.ic_quest_handrail
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_handrail_title
+
+    // use simple yes/no user interface without additional pictures for prompting user for answer
+    override fun createForm() = YesNoQuestAnswerFragment()
+
+    override fun applyAnswerTo(answerHandrailExists: Boolean, changes: StringMapChangesBuilder) {
+        changes.add("handrail", if (answerHandrailExists) "yes" else "no")
+    }
+}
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
```diff
@@ -0,0 +1,48 @@
...
+ * The details would only slow down the answering in a survey by requiring more thought and a more
+ * complex GUI (interpreting the direction of the OSM way for example).
+ *
+ * Github Issue for this quest:
+ * https://github.com/westnordost/StreetComplete/issues/1390
+ */
+class AddHandrail(overpassServer: OverpassMapDataDao) : SimpleOverpassQuestType<Boolean>(overpassServer) {
+    // Do not include nodes and relations, even though these exist with the right tags, because
+    // according to the wiki page for `highway=steps` [2] it can only be applied to ways. It also
+    // does not make much sense for other types of elements and it is more likely to be a
```
HolgerJeromin: Here is something missing...

app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
```diff
@@ -0,0 +1,48 @@
...
+    // Do not exclude elements tagged with `handrail:left`, `handrail:center` or `handrail:right`
+    // because according to the wiki page for `handrail=*` [1], the `handrail` tag should
+    // always have just `yes` or
+    // `no` as a value to indicate the existence of a handrail somewhere. If there is a mismatch of
+    // the tags afterwards, that can be recognized using one of the various
```
HolgerJeromin: In other situations we skipped the quest in such a situation. So I would exclude `handrail:*` here as well. 
Surveyor time is valuable...

xuiqzy: So you would assume that if only `handrail:*` is tagged, the survey was valid and should not be repeated by Streetcomplete, do I understand that correctly?

app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
xuiqzy: oops, I wanted to say it is likely to be a tagging mistake. Does that make sense?

app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
westnordost: Yes

app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
```diff
@@ -0,0 +1,49 @@
...
+    // Quality Assurance tools [3].
+    // [1] https://wiki.openstreetmap.org/wiki/Key:handrail
+    // [2] https://wiki.openstreetmap.org/wiki/Tag:highway%3Dsteps
+    // [3] https://wiki.openstreetmap.org/wiki/Quality_assurance
+    override val tagFilters = "nodes with highway = steps and !handrail"
```
westnordost: I think you meant to write `ways with ...`

xuiqzy: yes, thanks

xuiqzy:
Thanks for the comments @westnordost !
I cut down on the comments now. But isn't it useful to have a shortcut to the corresponding issue to the quest instead of needing to search it in the git commit history? I just thought it would make sense, since there is such a direct 1:1 mapping from implemented quest in the file to github issue here?
But I don't feel strongly about it, so if that doesn't make sense for you, I'll just remove it :)

westnordost:
Well, depends on your tools. But any good IDE will have an annotate / history function built-in into the editor and so has Android Studio. In the editor, right-click on the bar that shows the line numbers on the left, press _annotate_. Or, to get the history, rightclick in the editor and select Git -> Show History. That is all there is.

xuiqzy:
Okay, that is pretty easy indeed! I changed it.

westnordost:
Looks good, thank you!

xuiqzy:
Thanks for merging! :)
Just in case you didn't see the review request above: Was the ordering of the quests, as well as only asking it for ways okay?

westnordost:
Yes

versat:
It seems that it also asks for handrails if it is an escalator. I am not sure if it makes sense in this case. Just wanted to let you know. If this is by intention then everything is fine i guess.

matkoniecz:
Yes, escalators are not excluded, thanks for reporting! (#1604 was very similar). I wonder whatever indoor stairs are also worth excluding.

xuiqzy:
I also don't know any escalators without a handrail, but there are indoor steps with and without handrail. Why would you exclude indoor steps @matkoniecz ?

MegaArthur:
I have definitely seen indoor steps without a handrail, @matkoniecz

matkoniecz:
@versat I opened https://github.com/westnordost/StreetComplete/pull/1646 stopping the question for conveying steps, still asked for indoor steps.

PR #1617, opened 2019-10-24
Add test documentation to architecture

rugk:
See/As per/Fixes https://github.com/westnordost/StreetComplete/issues/1605#issuecomment-546048230


westnordost:
See https://github.com/westnordost/StreetComplete/issues/1605#issuecomment-546066286

PR #1618, opened 2019-10-24
Add landuse to name Quest

wvanderp:
This pullreqest adds landuse to the name quest.

i've selected "cemetery", "allotments", "recreation_ground",  "military", "quarry" for this quest.

i think these are the few landuses which always have an name.

open to feedback and revision request

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -54,9 +54,10 @@ class AddPlaceName(
                 "amusement_arcade", "adult_gaming_centre", "tanning_salon", "horse_riding"
             ),
             "landuse" to arrayOf(
-                    "cemetery",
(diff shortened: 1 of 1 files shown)


westnordost:
That's a good idea! Maybe also add public parks? 

**However**, the deciding factor is not if that landuse will usually have a name, but if it can be expected that (at the entrance to) this landuse there will be a sign denoting the name. Otherwise, surveyors will tag it incorrectly as `noname=yes` just because they do not see a sign.

Would you please revise the list to include only those landuses where one would expect a sign (at the entrance)?

wvanderp:
Then I would change it as follows:

Things with a name:
- cemetery

Things which might have a name:
- allotments
- some military

things which might be difficult
- quarry
- recreation_ground

So, what I propose is to add cemetery and allotments and add military under the military tags

westnordost:
Sounds good!

westnordost:
Good, thank you!

matkoniecz:
Note #2186 requesting to never ask for names of cemeteries and allotments

PR #1619, opened 2019-10-25
Fix attribution

matkoniecz:
Also I found some cases that I was unable to solve:

https://github.com/westnordost/StreetComplete/blob/master/app/src/main/res/drawable-xxhdpi/surface_grass.jpg
https://github.com/westnordost/StreetComplete/blob/master/app/src/main/res/drawable-xxhdpi/surface_gravel.jpg
https://github.com/westnordost/StreetComplete/blob/master/app/src/main/res/drawable-xxhdpi/surface_fine_gravel.jpg
https://github.com/westnordost/StreetComplete/blob/master/app/src/main/res/drawable-xxhdpi/surface_paving_stones.jpg

Added in https://github.com/westnordost/StreetComplete/commit/11a9cb5d76efeeb7b3b936feef0cc5e15ad4e0b1#diff-436f37988eb05deaf4d70d465178dbd1 without adding credit - is it not listed because it is your image?

-----


westnordost:
Honestly, I don't remember any more. Perhaps they were public domain.

PR #1620, opened 2019-10-29
add extra filters to internet access quest

wvanderp:
This PR will add more entities that are interesting for this quest

these are:

**amenity** 
- restaurant
- fast_food
- cafe
- pub
- bar
- nightclub
- biergarten
- theatre
- bus_station
- community_centre

**tourism**
- camp_site
- museum

**leisure**
- marina
- sports_centre
- fitness_centre

**shop**
- hairdresser
- car_repair
- department_store
- supermarket
- beauty
- mall

**public_transport**
- station


westnordost:
The list of places where it should be asked is deliberately very small. Remember, the quest is shown worldwide. Maybe in the Netherlands, the chance is reasonably high that at these places, there will be wifi, but then again in other countries, the chance may be close to 0%.

The list includes a few places where one stays abroad because in any country, the chance is higher there, plus it is also a more valuable information for OSM surveyors themselves :-).

Looking through your list, most would not be eligible. Maybe camp sites or other `tourism`-places where one can stay.

westnordost:
I added all accomodations as well as community centre and youth centre to the query

PR #1621, opened 2019-10-29
Make master list and sync them across the quests

wvanderp:
This PR makes master list for the name, opening hour and wheelchair quest.
all the unique tags stay separate in the quest and the list can be shared across quest.

I also shuffled some only present in some quest to the common list if it was appropriate for all three quest.

This PR also adds the office and craft tags to the name quest

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -73,21 +73,6 @@ class AddPlaceName(
                 //name only
                 "nature_reserve", "sports_centre", "dance", "golf_course",
                 "stadium", "marina"
-            ),
-            "office" to arrayOf(
-                // common
-                "insurance", "government", "travel_agent", "tax_advisor", "religion", "employment_agency",
-
-                // name and wheelchair
-                "lawyer", "estate_agent", "political_party", "therapist"
-            ),
-            "craft" to arrayOf(
-                // common
-                "carpenter", "shoemaker", "tailor", "photographer", "dressmaker",
-                "electronics_repair", "key_cutter", "stonemason",
-
-                // name and wheelchair
-                "winery"
             )
         ).map { it.key + " ~ " + it.value.joinToString("|") }.joinToString("\n  or ") + "\n" + """
         )
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -27,31 +27,67 @@ class AddPlaceName(
           or office
           or tourism = information and information = office 
           or """.trimIndent() +
+
+        // The common list is shared by the name quest, the opening hours quest and the wheelchair quest.
+        // So when adding other tags to the common list keep in mind that they need to be appropriate for all those quests.
+        // Independent tags can by added in the "name only" tab.
+
         mapOf(
             "amenity" to arrayOf(
-                "restaurant", "cafe", "ice_cream", "fast_food", "bar", "pub", "biergarten", "food_court", "nightclub",                  // eat & drink
-                "cinema", "theatre", "planetarium", "arts_centre", "studio",                                                            // culture
-                "events_venue", "conference_centre", "exhibition_centre", "music_venue",                                                // events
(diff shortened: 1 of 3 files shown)


westnordost:
It would have been better to separate the re-sorting of the mentioned tags and adding of more tags, so now it is more difficult to review for me.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -16,36 +16,58 @@ class AddWheelchairAccessBusiness(o: OverpassMapDataDao) : SimpleOverpassQuestTy
...
+                "place_of_worship",
+                "hospital"
             ),
             "tourism" to arrayOf(
-                "zoo", "aquarium", "theme_park", "gallery", "attraction", "viewpoint",
+                // common
+                "zoo", "aquarium", "theme_park", "gallery", "museum",
+                // wheelchair only
+                "attraction", "viewpoint",
                 "museum", "hotel", "guest_house", "hostel", "motel", "apartment", "chalet"
```
westnordost: museum is mentioned twice here

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -16,36 +16,58 @@ class AddWheelchairAccessBusiness(o: OverpassMapDataDao) : SimpleOverpassQuestTy
...
+                // common
+                "restaurant", "cafe", "ice_cream", "fast_food", "bar", "pub", "biergarten", "food_court", "nightclub", // eat & drink
+                "cinema", "planetarium", "casino", "arts_centre",                                                      // amenities
+                "townhall", "courthouse", "embassy", "community_centre", "youth_centre", "library",                    // civic
+                "bank", /*atm,*/ "bureau_de_change", "money_transfer", "post_office", "marketplace", "internet_cafe",  // commercial
+                "car_wash", "car_rental", "boat_rental", "fuel",                                                       // car stuff
+                "dentist", "doctors", "clinic", "pharmacy", "veterinary",                                              // health
+
+                // wheelchair only
+                "theatre", "arts_centre", "conference_centre",
```
westnordost: arts_centre is mentioned twice

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -27,31 +27,67 @@ class AddPlaceName(
...
+                // name and wheelchair
+                "lawyer", "estate_agent", "political_party", "therapist"
+            ),
+            "craft" to arrayOf(
+                // common
+                "carpenter", "shoemaker", "tailor", "photographer", "dressmaker",
+                "electronics_repair", "key_cutter", "stonemason",
+
+                // name and wheelchair
+                "winery"
```
westnordost: The name is already asked for **any** craft, see line 26. This makes no sense here.

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -27,31 +27,67 @@ class AddPlaceName(
...
+                "nature_reserve", "sports_centre", "dance", "golf_course",
+                "stadium", "marina"
+            ),
+            "office" to arrayOf(
+                // common
+                "insurance", "government", "travel_agent", "tax_advisor", "religion", "employment_agency",
+
+                // name and wheelchair
+                "lawyer", "estate_agent", "political_party", "therapist"
+            ),
```
westnordost: The name is already asked for **any** office, see line 27. This makes no sense here.

westnordost:
Thanks! There is another thing left to do, see https://github.com/westnordost/StreetComplete/pull/1621#pullrequestreview-309153568

westnordost:
Ok, but I also wrote

> From the wheelchair access, we can keep planetarium and ferry_terminal The rest is either not really open to the general public (without appointment / without being enrolled) or not reasonable to tag this (boat rental, dive centre)

westnordost:
I went through the list again and cleaned it up (found some more duplicates, mistakes and mis-sortings and merged it (from a fork of your branch)

PR #1622, opened 2019-10-31
Maxweight (supersedes #1467)

westnordost:
Wasn't able to push on @matkoniecz branch, so I forked it


app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt
```diff
@@ -87,80 +86,66 @@ class AddMaxSpeedForm : AbstractQuestFormAnswerFragment<MaxSpeedAnswer>() {
...
         else -> null
     }
 
     private fun userSelectedUnusualSpeed(): Boolean {
-        val speed = speed.toInt()
-        val speedUnit = speedUnitSelect?.selectedItem as String?
-        val speedInKmh = if (speedUnit == "mph") mphToKmh(speed.toDouble()) else speed.toDouble()
-        return speedInKmh > 140 || speed > 20 && speed % 5 != 0
+        val kmh = getSpeedFromInput()?.toKmh() ?: return false
+        return kmh > 140 || kmh > 20 && kmh % 5 != 0
```
peternewman: @westnordost This throws an "implausible" warning for 20 mph in the UK, and would do the same for 30. The % 5 bit needs to be done on the raw speed value rather than the converted kmh value.

I'm not quite sure how I didn't hit this before, or is this only triggered in some cases (this was a zone not a sign).

peternewman: https://github.com/westnordost/StreetComplete/pull/1632

westnordost: Oh, thank you for reporting it and thank you even more for offering a fix in a PR!

app/src/main/res/values/strings.xml
```diff
@@ -623,7 +623,11 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_railway_crossing_barrier_none">It\'s not</string>
     <string name="action_deselect_all">Deselect all</string>
     <string name="quest_buildingType_retail_description">including buildings for restaurants, cafés, …</string>
-
+    <string name="quest_maxweight_title">"What is the weight limit here?"</string>
```
rugk: Hmm, here is a quotation character too much, is not it? This should not be quoted…

PR #1627, opened 2019-11-06
Recycling quest

### LINKED ISSUE #223: amenity=recycling without recycling:*=yes tag
`amenity=recycling` without a knowledge which waste i can dump is not useful.

[Wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Drecycling) has a list. [Taginfo](https://taginfo.openstreetmap.org/tags/recycling_type=container#combinations) has a list of most used values:

recycling:glass=* (but some should be glass_bottles)
recycling:paper=*
recycling:clothes=*
recycling:glass_bottles=*
recycling:cans=*
recycling:plastic=*

related to #222 (this here could be asked after adding the type)

westnordost:
fixes #223


westnordost:
@matkoniecz

app/src/main/java/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterials.kt
```diff
@@ -0,0 +1,52 @@
...
+    override val icon = R.drawable.ic_quest_recycling_materials
+
+    override fun download(bbox: BoundingBox, handler: MapDataWithGeometryHandler): Boolean {
+        return overpassServer.getAndHandleQuota(getOverpassQuery(bbox), handler)
+    }
+
+    private fun getOverpassQuery(bbox: BoundingBox) = """
+        ${bbox.toGlobalOverpassBBox()}
+        node[amenity = recycling][recycling_type = container][~"^recycling:.*$"~".*"] -> .known;
+        node[amenity = recycling][recycling_type = container] -> .all;
```
goldfndr: I don't know how much it would save, but is it possible to swap these, filtering the .all set for the .known set? Something like ```node.all[~"^recycling:.*$"~".*"] -> .known;``` ?

westnordost: Yes, thanks!

matkoniecz:
It is potentially confusing for user that selecting "scrap metal" is not selecting "cans".

superminor: bottom half of the battery icon in night mode merges with background

matkoniecz:
Also, maybe "Glass bottles" can be "Glass bottles and jars"? Or "bottles and jars"?

Or maybe jar in icon is sufficient?

Overall: great!

app/src/main/res/values/strings.xml
```diff
@@ -703,4 +703,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_recycling_type_plastic_bottles">Plastic bottles only</string>
     <string name="quest_recycling_type_plastic_packaging">Plastic packaging only</string>
     <string name="quest_recycling_type_scrap_metal">Scrap metal</string>
+    <string name="quest_recycling_type_any_glass">Any glass</string>
+    <string name="quest_recycling_glass_title">Can only glass bottles be recycled here, or any type of glass?</string>
```
matkoniecz: I would explicitly include jars in question and specifically describe weird types of glass - window glass, tempered glass, pyrex, light bulb glass, glassware and so on.

Without that I would happily answer "yes, all kinds of glass" and be sure that I am correct.

I even opened https://github.com/osm-fr/osmose-backend/issues/549 thinking that validator is wrong.

Maybe ask that in "are you sure dialog"? Or have additional text below like wheelchair quest has?

matkoniecz: I can make additional PR adding this.

westnordost: The question should not get too long and the button "Any glass" must also stay very short.  So I think an additional text like in the wheelchair quest would be best. I'd gladly accept a PR

matkoniecz:
one potential change: It is potentially confusing for user that selecting "scrap metal" is not selecting "cans".

Maybe switching to interface like plastic would be a good solution?

westnordost:
The problem with that solution is that either one of the two would be initially hidden. 

If only "scrap metal" is displayed, which would be technically the correct thing to display (as cans is a type of scrap metal), people may miss out on the cans, which at least in Germany should be recyclable at any place where one can recycling plastic packaging.

westnordost:
**TODO**: test how the popup-menu works together with the memorize-last-picked-values feature.

matkoniecz:
> memorize-last-picked-values feature

I see no effects of it in my testing - order of items seems to not change based on an earlier, not uploaded answers and plastic is always initially displayed as "plastic" not packaging/bottles.

westnordost:
> I see no effects of it in my testing - order of items seems to not change based on an earlier, not uploaded answers 

Well, that is a bug then. The order should *always* change, except if the total number of selectable items fits in one row.

westnordost:
Turns out the favourites-feature was bugged since I converted all the quests to Kotlin.

PR #1628, opened 2019-11-07
Add postbox ref

### LINKED ISSUE #1556: Add missing ref=* on amenity=post_box
### General
Affected tag(s) to be modified/added: [ref=*](https://wiki.openstreetmap.org/wiki/Key:ref) on [amenity=post_box](https://wiki.openstreetmap.org/wiki/Tag:amenity=post_box)
Question asked: **What is the reference number of this post box?**

### Checklist
- [ x ] 🚧 To be added tag is established and has a useful purpose
- [ x ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [ x ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [ x ] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [ x ] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

**Element selection:** 

amenity=post_box nodes in countries where post boxes usually have numbers (certainly includes UK and France), which have no ref=* tag, and ref:signed=* (if present) is not equal to "no".

**Proposed GUI:**

Q: What is the reference number of this post box?

* The reference number is [ ________________ ]
* This box should have a number but it is not shown or is unreadable.
* This box does not have a number (perhaps it is a private post box).

In the first case ref=* is set to the value given. In the second case ref:signed=no is set. In the third case ref=no is set.

(There has been some discussion about ref=* more generally in #81.)

westnordost:
fixes #1556


matkoniecz:
Is it open as PR because in lands in the next version, not this one or because some feedback is welcomed?

Not sure whatever it is better to test this one or work on some other issues.

westnordost:
Mostly because it will get into the next version. You are always invited to look through it, but I think your time is invested somewhere else better.

PR #1632, opened 2019-11-09
Ensure we're checking if the speed value is unusual

peternewman:
Not the converted value in kmh, otherwise every mph value will be odd.

As I stuck in a comment in here:
https://github.com/westnordost/StreetComplete/pull/1622/files#r344422067
> This throws an "implausible" warning for 20 mph in the UK, and would do the same for 30. The % 5 bit needs to be done on the raw speed value rather than the converted kmh value.
> 
> I'm not quite sure how I didn't hit this before, or is this only triggered in some cases (this was a zone not a sign).

This code is entirely untested, but seemed trivial enough it was worth just opening a PR rather than an issue.


westnordost:
Thanks!

PR #1636, opened 2019-11-10
keep idea, fix code of #1632

matkoniecz:
fixes compilation error


westnordost:
Weird, I got not compilation error when I just created the release. But now that I try it again, I get it. Thank you for noticing, I'll hotfix 1501 with your fix

peternewman:
Sorry. I'm sort of curious, what was the error? I can't obviously see anything that would have stopped my code from compiling.

matkoniecz:
> I can't obviously see anything that would have stopped my code from compiling.

Nullability check.

`Operator call corresponds to a dot-qualified call 'getSpeedFromInput()?.toValue().rem(5)' which is not allowed on a nullable receiver 'getSpeedFromInput()?.toValue()'.`

> Sorry.

No problem! It was helpful and it would be nicer to have it working without modifications, but it was much better than nothing at all.

peternewman:
Ah I see, it's not clever enough to realise the earlier false will protect it, although I guess if it was multi-threaded it could be an issue.

westnordost:
Exactly. Kotlin this way forces you to write threadsafe code.

PR #1639, opened 2019-11-11
make situation with jars and weird glass types clear

matkoniecz:
this PR is made onto #1627 branch

in `recycling:glass` quest:
adds bonus info about weird glass types, replaces "glass bottles" by "bottles and jars" (glass should be made clear by context)

in `recycling:*` quest:
explicitly mentions jars (only glass bottles were mentioned in text, jars were on icon but excluded by text)

-----

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -706,6 +706,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_recycling_type_scrap_metal">Scrap metal</string>
     <string name="quest_recycling_type_any_glass">Any glass</string>
     <string name="quest_recycling_glass_title">Can only glass bottles and jars be recycled here, or any type of glass?</string>
-    <string name="quest_determineRecyclingGlass_description_any_glass">Examples of glass that is often not accepted is such glass as from window panes, glassware, light bulbs or mirrors.</string>
+    <string name="quest_determineRecyclingGlass_description_any_glass">Examples of glass that is often not accepted: tempered glass, window panes, glassware, light bulbs or mirrors.</string>
 
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -706,6 +706,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_recycling_type_scrap_metal">Scrap metal</string>
     <string name="quest_recycling_type_any_glass">Any glass</string>
     <string name="quest_recycling_glass_title">Can only glass bottles and jars be recycled here, or any type of glass?</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -704,6 +705,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_recycling_type_plastic_packaging">Plastic packaging only</string>
     <string name="quest_recycling_type_scrap_metal">Scrap metal</string>
     <string name="quest_recycling_type_any_glass">Any glass</string>
-    <string name="quest_recycling_glass_title">Can only glass bottles be recycled here, or any type of glass?</string>
+    <string name="quest_recycling_glass_title">Can only glass bottles and jars be recycled here, or any type of glass?</string>
+    <string name="quest_determineRecyclingGlass_description_any_glass">\"any glass\" includes window glass, glassware, tempered glass, optical glass, light bulb glass, mirrors and other that are often not accepted.</string>
```
westnordost: Sounds a bit odd to enumerate what "any glass" encompasses. Any glass is any glass. How about giving only a few well known examples of glass that are definitely not glass bottles or similar.

```suggestion
    <string name="quest_determineRecyclingGlass_description_any_glass">Examples of glass that is often not accepted is such glass as from window panes, glassware, light bulbs or mirrors.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -704,6 +705,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_recycling_type_plastic_packaging">Plastic packaging only</string>
     <string name="quest_recycling_type_scrap_metal">Scrap metal</string>
     <string name="quest_recycling_type_any_glass">Any glass</string>
-    <string name="quest_recycling_glass_title">Can only glass bottles be recycled here, or any type of glass?</string>
+    <string name="quest_recycling_glass_title">Can only glass bottles and jars be recycled here, or any type of glass?</string>
+    <string name="quest_determineRecyclingGlass_description_any_glass">Examples of glass that is often not accepted: tempered glass, window panes, glassware, light bulbs or mirrors.</string>
```
goldfndr: "Examples" is plural, so change "that is often" to "that are often".

matkoniecz: Fixed. I thought that sentence construction would be

    (examples) of (glass that is often not accepted)

not

    (example of glass) (that are often not accepted)

goldfndr: Your alternative sentence construction (with "glass" as a collective noun or singular noun) is, at first glance, valid. But the terms after the colon are majority plural, so a substitution (e.g. "light bulbs is not accepted") breaks it, so to speak. :)

PR #1641, opened 2019-11-12
 replace files with unknown source by ones with a known one 

matkoniecz:
@westnordost Do you have any idea what is the source of plop sounds added in https://github.com/westnordost/StreetComplete/commit/881a4cccb23348a71805c97b2e26adbc686c9e38 ?

-----

In case that someone would be curious why: I need to list all files with its authors and licenses. Part of it (annotations at top of every file) makes sense in projects with multiple licences or some other convoluted situation, so I will not make PRs for that.

But this part I think is at least a bit useful so I will make PRs for that.


westnordost:
> @westnordost Do you have any idea what is the source of plop sounds added in 881a4cc ?

Either I made them myself, or I took public-domain-licensed ones from freesound.org.

Same really for all the other pictures. I never fetched any image with an incompatible license.

> In case that someone would be curious why: I need to list all files with its authors and licenses

I see that, but why?

matkoniecz:
> Same really for all the other pictures. I never fetched any image with an incompatible license.

Yeah, I know. It is "I need source for file XYZ", not "I think that this project is full of copyvios".

> I see that, but why?

NGI0 grant is relatively free from bureaucracy but I encountered part where I am dubious about usefulness of an obligatory solution. They are apparently fans of https://reuse.software/tutorial/

Note that I am not obligated to pull it into StreetComplete - it is sufficient that I will make a fork for parts that are unwanted by you. For parts that seem useful anyway I will open PRs.

Though in case that you want I can make PR will full blown REUSE-compliant copyright specifications at header of every single source code file.

westnordost:
> Though in case that you want I can make PR will full blown REUSE-compliant copyright specifications at header of every single source code file.

Uff, I generally see that as unnecessary bloat, as this information is already available in `COPYING`. The argumentation of those that say it should be included at the top of every file is that people may reuse single files of a GPL project and taken out of the source project, the license information is gone. However, with that argument, it would also not be enough to put the license and author information into a central authors.txt as you are doing now, but put it into the meta-data of each (image) file.

I think it is the duty of anyone who takes a file out of a GPL (or otherwise licensed) project to maintain the license information.

matkoniecz:
> However, with that argument, it would also not be enough to put the license and author information into a central authors.txt as you are doing now, but put it into the meta-data of each (image) file.

Well, they went into direction of single metadata file for every single image. At least authors.txt is machine readable and it is possible to automate this.

westnordost:
I meant metadata of the image. As far as I know, this is possible. Who is "they"?

matkoniecz:
Authors of https://reuse.software

> You also want to license your image files under GPL-3.0-or-later. Unfortunately, images and other binary files do not have comment headers that one can easily edit. Other examples include automatically generated files and certain data and configuration files for which comments are non-trivial.

> There is a simple trick to circumvent this. Create the files cat.jpg.license and dog.jpg.license, each containing the same information about license and copyright holder as above.

goldfndr:
https://github.com/fsfe/reuse-docs/blob/master/faq.md mentions a "bulk-license" option, that would drastically reduce a compliance commit.

PR #1644, opened 2019-11-14
make debug screen listing more clear and provide support for more quests

matkoniecz:
previously name was "null" was could be unclear especially for non-programmers

changing geometry to way enables support for more quests that previously crashed

now just note quest and oneway quest are not working

-----


westnordost:
I am surprised not more quests crash. Some forms (rightfully) assume that certain tags are set on the element.

matkoniecz:
Some were set already in https://github.com/matkoniecz/Zazolc/commit/f35e56d28e7694744e79d620eaf8b58fb7c66265#diff-7d0aa2bc1556fe5648d749e29ca5c8ae

PR #1648, opened 2019-11-15
allow multiple substitutions in the quest title [requested changes implemented]

matkoniecz:
currently without influence on how app works, but I thought about

> Is \<info> accurate for \<name>?

(for this one I am currently testing a prototype)

or specify more info for opening hours quest

> What are the opening hours of \<name\> (\<object type>)

as currently I relatively often need to use Vespucci to check whatever what type of object is without OH. It is often necessary to locate it or confirm that it is actually gone.

-----

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -69,7 +69,7 @@ class AddPlaceName(
     override val icon = R.drawable.ic_quest_label
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_placeName_title_name
-    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>) = if (featureName.value == null) arrayOf() else arrayOf(featureName.value!!)
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>) = featureName.value?.let { arrayOf(it) } ?: arrayOf()
 
     override fun download(bbox: BoundingBox, handler: MapDataWithGeometryHandler): Boolean {
         val overpassQuery = bbox.toGlobalOverpassBBox() + "\n" + filter.toOverpassQLString() + getQuestPrintStatement()
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -69,7 +69,7 @@ class AddPlaceName(
     override val icon = R.drawable.ic_quest_label
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_placeName_title_name
-    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>) = arrayOf(featureName.value)
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>) = if (featureName.value == null) arrayOf() else arrayOf(featureName.value!!)
 
     override fun download(bbox: BoundingBox, handler: MapDataWithGeometryHandler): Boolean {
         val overpassQuery = bbox.toGlobalOverpassBBox() + "\n" + filter.toOverpassQLString() + getQuestPrintStatement()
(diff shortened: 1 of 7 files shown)


matkoniecz:
I would be also fine with rejecting it for now as it is NOP, and merge it only with quest that would actually use it, but it seems to be not causing introduction of some new architecture or noticeable complexity.

matkoniecz:
Closing for now due to a discovered bug (quest list is not working once quest with multiple substitutions is added)

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsUtil.kt
```diff
@@ -0,0 +1,10 @@
+package de.westnordost.streetcomplete.settings
+
+import android.view.View
+import de.westnordost.streetcomplete.data.QuestType
+
+fun genericQuestTitle(resourceProvider: View, type: QuestType<*>): String {
+    val questTitleTemplate = resourceProvider.resources.getString(type.title)
+    val parameterCount = questTitleTemplate.split("%s").size - 1
+    return resourceProvider.resources.getString(type.title, *Array(parameterCount){"…"})
+}
```
westnordost: 1. Better as first parameter `resources: Resources`
2. You can't assume that every placeholder is a `%s`. And I don't mean `%d` etc, but the `%4$2s` syntax which might rightfully occur sometimes. I looked at the [documentation for Formatter](https://developer.android.com/reference/java/util/Formatter.html) (which `Resources::getString()` uses internally) and sadly, it doesn't expose any functions with which one can find out how many placeholders are defined. Luckily, as far as I know, the Formatter is lenient in regards to if more args are supplied than there are placeholders.

So my suggestion is to just supply an array of 10 or so "…" and that's it.

The other possible solution would be to continue to allow optional strings for the arg array and let all the implementations of `getTitleArgs` behave like this...
```
    fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String?> {
        return arrayOf(name) // <- name may be null
    }
```
... and use the `getTitleArgs` function for the display in the quest list as well, only replace every `null` with`...`. This may even be the cleanest solution.

matkoniecz: > So my suggestion is to just supply an array of 10 or so "…" and that's it.

That was my initial idea, but it seemed like an extremely ugly hack (though I would be deeply surprised to see it ever failing). But after thinking it again, assuming that all placeholders will be always `%s` is even uglier.

> ... and use the getTitleArgs function for the display in the quest list as well, only replace every null with"…". This may even be the cleanest solution.

OK, I will switch to this version.

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -69,7 +69,7 @@ class AddPlaceName(
     override val icon = R.drawable.ic_quest_label
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_placeName_title_name
-    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>) = arrayOf(featureName.value)
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>) = if (featureName.value == null) arrayOf() else arrayOf(featureName.value!!)
```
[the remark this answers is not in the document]
matkoniecz: Not for me, mostly because I am still not used to `.let` syntax. Though it avoid duplication of `featureName.value`, so lets go with it.

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsUtil.kt
matkoniecz: Looking at it now: only `OsmElementQuestType` has `getTitleArgs`, not `QuestType`, reuires passing `featureName: Lazy<String?>` and so on.

I think I will go with repeating dots 10 times.

westnordost: So, what is the problem? 99% of quests are OsmElementQuestTypes and featureName is just a Lazy<String?>, note the `?`, so it can also be a lazy that returns null

westnordost:
Anyway, looks good!

PR #1650, opened 2019-11-15
add helpful comments

matkoniecz:
or at least would be helpful for me about 10 hours ago


goldfndr:
The added description for the function `getOffsetDateString` in DateUtil.kt says "returns date in yyyy-mm-ddThh:mm:ssZ" but its usage in AMarkCompletedConstruction.kt has "T00:00:00Z" appended.

matkoniecz:
Fixed, lets pretend that I was not a victim of the same bug for the fourth time today.

PR #1653, opened 2019-11-17
Minor tweak to the language of some messages

peternewman:
Make it less casual

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -611,7 +611,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_segregated_title">How are the footway and cycleway laid out here?</string>
     <string name="dialog_tutorial_upload">To upload your changes manually, use the button in the toolbar that looks like this.</string>
     <string name="quest_segregated_separated">Segregated from one another</string>
-    <string name="quest_segregated_mixed">Cyclists and pedestrians share same space</string>
+    <string name="quest_segregated_mixed">Cyclists and pedestrians share the same space</string>
     <string name="quest_maxheight_title">"What is the height limit here?"</string>
     <string name="quest_maxheight_parking_entrance_title">What is the height limit of this parking entrance?</string>
     <string name="quest_maxheight_height_restrictor_title">What is the height limit of this height restrictor?</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -611,7 +611,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_segregated_title">How are the footway and cycleway laid out here?</string>
     <string name="dialog_tutorial_upload">To upload your changes manually, use the button in the toolbar that looks like this.</string>
     <string name="quest_segregated_separated">Segregated from one another</string>
-    <string name="quest_segregated_mixed">Cyclists and pedestrians share same space</string>
+    <string name="quest_segregated_mixed">Cyclists and pedestrians share the same space</string>
```
peternewman: I assume this will still fit okay?

goldfndr:
I agree with the second commit (insert "the") but I can't help feeling that the first one was intentionally worded for its double/triple meaning (["stretch of road"](https://dictionary.cambridge.org/us/example/english/stretch-of-road)).

peternewman:
> I can't help feeling that the first one was intentionally worded for its double/triple meaning (["stretch of road"](https://dictionary.cambridge.org/us/example/english/stretch-of-road)).

Possibly @goldfndr . I could just imagine non-native speakers either not understanding/not getting the reference or spending a long time translating it into other languages where it just ends up a bit nonsensical.

app/src/main/res/values/strings.xml
westnordost: Yes

westnordost:
Yes I meant it the way @goldfndr said. "You have to go down quite a stretch of road"

westnordost:
@peternewman ping

peternewman:
Sorry I've been rather busy with work.

> Yes I meant it the way @goldfndr said. "You have to go down quite a stretch of road". Not aware of any other meaning of it.

So as in this "piece of road", not "asking a lot of work of the person" or "very long distance"?

So the end is currently this:

> This can sometimes be quite the stretch just to solve a single quest.

Assuming you mean the former, this probably reads better.

> This can sometimes be quite a stretch of road just to solve a single quest.

I realise it's rather clunky but I can't personally think of another way to word it that would likely translate as intended for others.

westnordost:
> So as in this "piece of road", not "asking a lot of work of the person" or "very long distance"?

I meant to say, it may be a very long piece of road to go down, resulting in a lot of work to just solve one quest.

peternewman:
So it looks like as @goldfndr said you were going for the double meaning then?

Do you want me to revert my change there? My only comment is it's a bit colloquial and may be challenging for others to translate with the intended meaning.

westnordost:
Now I understand what you meant with the double meaning.

Yes, I'd prefer if the double meaning was preserved. The language used in StreetComplete is rather colloquial anyway, f.e. the German translation uses the colloquial variant of "you" to address the user (="thou" if it was still in use in modern day English, by the way).

peternewman:
Okay text change reverted and a comment made in POEditor so hopefully others will see it.

westnordost:
Nice, thanks!

PR #1665, opened 2019-11-21
icon for night mode for gymnastics [change original rather than making special dark mode version]

matkoniecz:
I have some doubts is it OK, but at least it is now not merging into background.

Modified icon is in the first column, the second row.

Screenshot_2019-11-21-13-47-46-326_de westnordost streetcomplete debug

-----


westnordost:
I prefer to have only one icon for any configuration

matkoniecz:
@westnordost What kind of software can open this files? Inkscape is not opening this kind of svgs. Opening them as text files is viable but not very efficient, as I need to recompile entire app to see the new version of the icon.

westnordost:
This is not a SVG, it's an Android vector drawable. You can view them in Android Studio (Preview tab), there is no editor other than the text editor.
These vector drawables are imported from SVGs by choosing New ->Vector Asset when rightclicking on the res-folder

matkoniecz:
Thanks for info! I will close this and open a new PR, as no code will be reused and different file will be edited.

PR #1671, opened 2019-11-28
give hint how to use street side puzzle

matkoniecz:
turns out that some users need tutorial/explanation how to use it
lets have it directly in the app

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -688,5 +688,5 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generalFee_title">Does it cost a fee to enter %s?</string>
     <string name="quest_laundrySelfService_title">Is this laundry a self service laundry?</string>
     <string name="quest_handrail_title">Do these steps have a handrail?</string>
-    <string name="quest_street_side_puzzle_tutorial">Tap question marks and select an answer for given side of the road.</string>
+    <string name="quest_street_side_puzzle_tutorial">Tap question marks and select an answer for the given side of the road.</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -688,5 +688,5 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generalFee_title">Does it cost a fee to enter %s?</string>
     <string name="quest_laundrySelfService_title">Is this laundry a self service laundry?</string>
     <string name="quest_handrail_title">Do these steps have a handrail?</string>
-    <string name="quest_street_side_puzzle_tutorial">Tap question marks and select an answer for the given side of the road.</string>
+    <string name="quest_street_side_puzzle_tutorial">Tap the question marks and select an answer for the given side of the road.</string>
 </resources>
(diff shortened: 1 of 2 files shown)


westnordost:
Can you attach a screenshot?

The form that contains the puzzle has no padding, just adding a text field (without padding) might look odd.

Otherwise I agree, the form could use some short explanation.

westnordost:
Also, you forgot the "the" again in the text.

matkoniecz:
> The form that contains the puzzle has no padding, just adding a text field (without padding) might look odd.

Yeah, padding would make it better. It is not so horrible that I would notice the problem without pointing it out but now I see this.

> Can you attach a screenshot?

https://user-images.githubusercontent.com/899988/69833928-464a5800-1237-11ea-83b2-067d07c34ec2.png
https://user-images.githubusercontent.com/899988/69833930-46e2ee80-1237-11ea-8a3a-abbe2812b58a.png

matkoniecz:
Margin added, though it margin value was selected a bit randomly. I considered also margin from the bottom side, but it seems not necessary to me (I am not really trusting my opinion here so feedback is welcomed).

https://user-images.githubusercontent.com/899988/69858876-42e4ba00-1293-11ea-8a69-7a880d936053.png

westnordost:
Couldn't you use the same @dimen/ values as for the quest form with paddings?

Also, for all other quest form descriptions, the description is at the top. Is there any good reason to deviate from this rule? 

Also, I think it would better to separate the "part of the form that has padding" with the part of the form that has no padding with a horizontal grey line the same way the button bar is separated from the rest of the form. You could reuse the same drawable.

app/src/main/res/values/strings.xml
```diff
@@ -688,4 +688,5 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generalFee_title">Does it cost a fee to enter %s?</string>
     <string name="quest_laundrySelfService_title">Is this laundry a self service laundry?</string>
     <string name="quest_handrail_title">Do these steps have a handrail?</string>
+    <string name="quest_street_side_puzzle_tutorial">Tap question marks and select an answer for the given side of the road.</string>
```
```suggestion
    <string name="quest_street_side_puzzle_tutorial">Tap the question marks and select an answer for the given side of the road.</string>
```
Sorry, you didn't catch all the thes. :-P

matkoniecz:
> Couldn't you use the same @dimen/ values as for the quest form with paddings?

I missed it, thanks! It now looks better after fixing that as itb matches margins for buttons.

> For all other quest form descriptions, the description is at the top. Is there any good reason to deviate from this rule?

I thought about is an optional and thought that placing it there will be less flow breaking. But then it will be anyway between the main interface and buttons.

I was basing it also on wheelchair quest, where hint is above buttons. But wheelchair quest has no quest interface like lane quest. 

Looking at road name, bus name, religion, recycling, sport and probably other quests I now see that I am breaking the standard UI. I will fix this. 

> Also, I think it would better to separate the "part of the form that has padding" with the part of the form that has no padding with a horizontal grey line the same way the button bar is separated from the rest of the form. You could reuse the same drawable.

I added it, but I think it was better without it.

https://user-images.githubusercontent.com/899988/69915059-1b385200-144b-11ea-83ba-d9e7cb6cca8b.png
https://user-images.githubusercontent.com/899988/69915060-1b385200-144b-11ea-98ea-6204ca7a9147.png

matkoniecz:
Ops, I forgot to move interface info to the top. Done, now divider looks OK.

Issue #213, opened 2017-05-10
quest for missing addr:street


matkoniecz:
Current state: https://github.com/westnordost/StreetComplete/issues/213#issuecomment-406603714

----------

As discussed in #205 at least for some regions (for example Poland) it makes sense to have quest asking about missing addr:street.

To make quests less complicated it should not be done as part of existing "add house number" quest.

Note that it would be preferable to ask question about addr:street immediately after solving housenumber quest, so overpass query finding quests target should not be in form "addr:housenumber without add:street", but rather the same as one finding missing house numbers.

Given that this quest apparently makes sense only is some regions (where addr:street is both likely to not be obvious and available to surveyor) it is preferable to add

> offline reverse country geocoder (GPS position -> country+region you are in, necessary for localized quests

mentioned in https://github.com/westnordost/StreetComplete/issues/188#issuecomment-299728054 before adding this quest (or at least quest filter, with this one disabled by default).

DaubnerF:
I would like to mention that this quest would make sense for my current mapping efforts in Austria as well. So if you implement it, please don't restrict it only to Poland :)
At least where I live, the street name associated with a house is readily available to the surveyor (most of the time).

Dostoi:
Same in France :)

westnordost:
The quest will not be limited to certain countries but limited to house number 1,2(,3,4).

DaubnerF:
Hmm, I guess this is better than restricting it to Poland, I still wish I could turn it on independent of housenumbers.
Maybe add a settings dialog, where certain quest types can be disabled altogether, with this type disabled by default?

rugk:
> Maybe add a settings dialog, where certain quest types can be disabled altogether, with this type disabled by default?

https://github.com/westnordost/StreetComplete/issues/35

DaubnerF:
Ah, very nice, thanks rugk.
So maybe we could have this quest-type independent of area and housenumber? :)

skfd:
Also note that some mapping communities use relations instead of addr:street. Thanks.

britiger:
In same cases the you can interpolate the addr:street using the street next to the housenumber. At  housenumbers on intersections it is difficult to automate it. Streetcomplete should also ask for the addr:street with a short list of the named roads next to building, including option "none of them" for creating a note.

matkoniecz:
Looking at it again: my plan is to reuse add road name quest interface, it even has (a bit hidden, I noticed it only now) option to select nearby road name.

[...]

Issue #26, opened 2017-03-25
Quest for mobility access


Monduiz:
Could a mobility access quest be added? This would be great to add this info on buildings.

westnordost:
Buildings in general would be quite a lot, plus the normal surveyor won't be able to find that out for the most buildings, since he cannot simple go in there. I do not want to have quests that are not solveable by the user.

I would suggest only for shops (and what else?)

Monduiz:
Well, shops are often nodes, not buildings and most often than not, they are nodes attached to a way(buildings), just like amenities.  Yes, I can see how buildings would be a lot. 

Civic places and transport though would make sense.

CloCkWeRX:
I'd put up there amenity=toilet
Answers: https://wiki.openstreetmap.org/wiki/Key:wheelchair#Common_values

CloCkWeRX:
And of course; visualisation/consumed by https://wheelmap.org/map

PanierAvide:
+1 for transports, in particular (but not only) bus stops (highway=bus_stop or public_transport=platform).

rmikke:
I would suggest this quest for shops and amenities.

wojtekmaj:
I vote for this! This is one of the things that is always omitted, easy to check and makes a huge difference, adds a huge value for OSM making it much more useful for people truly in need.

westnordost:
Oh, that has been implemented already by @Teester (and being refined as we speak)

PR #1676, opened 2019-12-01
resurveying opening hours

### LINKED ISSUE #1651: resurveying opening hours
based on more general #1112 #1364

In context of resurveying things `opening_hours` tag is a bit special:

- it goes out of date relatively quickly
- entering this data takes long time, so "reask question every X years" is not feasible unlike most of other quests
- has quite complex syntax

To handle it properly and to not end with something horrible for start it would be nice to have something along lines: "Is following opening hours still correct for $NAME?" and show opening hours data.

On answering yes, it would add `check_date:opening_hours`.

On answering no it would delete `opening_hours` tag and `check_date:opening_hours` if present, what should trigger opening hours quest allowing to enter correct data.

-------

Most of the fun is in "show opening hours data". Raw tag value is used in test version, but it is obviously not a proper solution, parsing opening hours data into some readable format is necessary, with unparsable opening hours fields ignored.

For parsing I see two solutions - (1) follow JOSM and embed js code from https://github.com/opening-hours/opening_hours.js and use it for parsing or (2) write a new parser from scratch, starting from one handling a small subset of `opening_hours` syntax.

@westnordost For me both versions sounds a bit frustrating and I have no strong preference - do you have any preference toward either one? I predict that running js code within SC may have some issues with getting more dependencies (js library itself, runner of js code that would be probably Rhino or an other Java library) while new parser will take more time to recreate something already written.

-----

Working on this was triggered by going to a shop and discovering there that it is closed because OSM data for `opening_hours` was old and wrong.

matkoniecz:
see https://github.com/westnordost/StreetComplete/pull/1676#issuecomment-563352766 for things requiring fixes

-----

fixes #1651, fixes part of #1112

It is a quite big PR, pulling in a new parsing dependency, quite complex checker of a parsed object whatever it fits SC requirements and the first resurvey quest.

Icon is the same as for AddOpeningHours, at least for now. I though whatever and how to distinguish icons for resurvey quests from normal quests. I think that exactly the same icon for the normal quest and a resurvey quest should be OK. I am unsure how to handle disabling/enabling resurvey quests.

This one is added like any other quest, I would leave potential changing this for a future. Enable/disable silently following normal quest, making impossible to disable resurvey quests? Have them on the list of quest types as a separate entry like this one? Have a global setting „disable resurvey quests?”)

Note that opening hours may be reformatted, without changing structure. I think that it is OK, example changes are documented as tests. Reformatting would apply on user solving quest.
For example `Tu-Su   10:30    -   18:00` to `Tu-Su 10:30-18:00`. Or `Tu-Su 10:30-28:00` to `Tu-Su 10:30-04:00`. Or `Mo 9-19` to `Mo 9:00-19:00`.

I think that it would be a good idea to start the code review from looking through tests added in this PR. Some decisions, like handling of cosmetic changes, are kind of subjective and maybe a different decision should be taken. 

There are some opening hours displayable with SC (like ones with an open end) that are not handled for now to reduce complexity of this PR. It should be fairly easy to add them in the future as a separate PRs.

-----


matkoniecz:
Yes, my plan is to finish other open PRs. In the worst case - finish work on open branches with some substantial investment like `addr:street`

I will try to avoid http://www.commitstrip.com/en/2014/11/25/west-side-project-story/ - it seems that I started to fall into it

Ironically this PR is attempt to get to the end of one of open projects, at least that was I telling myself.

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
```diff
@@ -0,0 +1,397 @@
+package de.westnordost.streetcomplete.quests.opening_hours
+
+import de.westnordost.streetcomplete.quests.opening_hours.adapter.OpeningMonthsRow
+import de.westnordost.streetcomplete.quests.opening_hours.adapter.toOpeningMonthsList
+import de.westnordost.streetcomplete.quests.opening_hours.model.Weekdays
+import org.junit.Assert
+import org.junit.Test
```
westnordost: import org.junit.Assert.* ?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/WorkingAssert.kt
```diff
@@ -0,0 +1,9 @@
+package de.westnordost.streetcomplete.quests.opening_hours
+
+object Assert {
+    fun assert(conditionAssertedToBeTrue: Boolean) {
+        if(!conditionAssertedToBeTrue) {
+            throw AssertionError()
+        }
+    }
+}
```
westnordost: What's this? Are you looking for [require](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html) or [check](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check.html)?
(See Preconditions.kt in the standard library)

westnordost:
Hui wow, this is a lot. I will do a proper review soon!

westnordost:
Why is addr:street a substantial effort?

matkoniecz:
> Why is addr:street a substantial effort?

Part of it was researching addresses - for example https://wiki.openstreetmap.org/w/index.php?title=Key:addr:conscriptionnumber&action=history and research necessary to make this edits was part of this. I did it because I was researching what kind of handling is necessary for conscription numbers in context of #213. Hilariously answer turned out to be "no handling is necessary, this tag may be simply ignored".

Name suggestion pools for `addr:street`, `addr:place` is also something that I never used before. And for now `addr:street` suggestion pool is based on hacky use of localized name quest - but with all localization stripped away. And now I want to add support for `addr:place` suggestion pool.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/WorkingAssert.kt
matkoniecz: I initially used `assert()` then I discovered that by default it is disabled.

And later I started wondering how to handle this in production code - just abort by returning null? Abort by throwing an exception and catching it? Abort by throwing an exception, catch it and log. Refuse to catch exception and crash only in the debug mode?

All of asserts are for testing whatever assumptions made by subsequent code are true, during development asserts were already useful and will be probably useful in case of future expansions of

matkoniecz:
> What's this? Are you looking for require or check?
> (See Preconditions.kt in the standard library)

Great, I will switch to https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html / https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check.html

I initially used `assert()` then I discovered that by default it is disabled.

And later I started wondering how to handle this in production code - just abort by returning null? Abort by throwing an exception and catching it? Abort by throwing an exception, catch it and log that. Refuse to catch exception and crash only in the debug mode?

All of this asserts are for testing whatever assumptions made by a subsequent code are true. During development asserts were already useful and will be probably useful in case of future expansions to handle more classes of opening hours.

westnordost:
I see, well, if you are stuck, I can finish it if you want 

Am 2. Dezember 2019 06:57:03 MEZ schrieb Mateusz Konieczny <notifications@github.com>:
>> Why is addr:street a substantial effort?
>
>Part of it was researching addresses - for example
>https://wiki.openstreetmap.org/w/index.php?title=Key:addr:conscriptionnumber&action=history
>and research necessary to make this edits was part of this. I did it
>because I was researching what kind of handling is necessary for
>conscription numbers in context of #213. Hilariously answer turned out
>to be "no handling is necessary, this tag may be simply ignored".
>
>Name suggestion pools for `addr:street`, `addr:place` is also something
>that I never used before. And for now `addr:street` suggestion pool is
>based on hacky use of localized name quest - but with all localization
>stripped away. And now I want to add support for `addr:place`
>suggestion pool.

matkoniecz:
> I see, well, if you are stuck, I can finish it if you want

Not, I am not stuck - for now I know what and how should be done. In case of being stuck and running out of ideas I will let you know.

app/src/main/res/values/strings.xml
```diff
@@ -688,5 +688,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generalFee_title">Does it cost a fee to enter %s?</string>
     <string name="quest_laundrySelfService_title">Is this laundry a self service laundry?</string>
     <string name="quest_handrail_title">Do these steps have a handrail?</string>
+    <string name="resurvey_opening_hours_title">"What are the opening hours of %s? Correct them if necessary."</string>
```
westnordost: UX wise, it is a bit odd to have an "OK" button for this question that is visible from the start on even so the user did not input anything yet. It could be misunderstood that for example pressing the "OK" button will lead to a form where one correct them or something.

More clear would be if there was a question like _"Are these opening hours for %s still correct?"_ and if the user clicks on "No", he is prompted to correct them. Technically it should be easily possible, make the opening hours unclickable first and on pressing no, exchange the yes-no button bar layout and make the opening hours form clickable.

This approach also solves another problem and that is how to treat opening hours that can not be displayed in the form: `24/7` at least. So, in the initial form, "24/7" (or "always open") can be printed in big letters in place of the opening hours input and only when the user clicks no, the normal opening hours form is displayed

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+import de.westnordost.streetcomplete.data.osm.tql.toGlobalOverpassBBox
+import de.westnordost.streetcomplete.quests.DateUtil
+
+class ResurveyOpeningHours(private val overpassServer: OverpassMapDataDao) : OsmElementQuestType<OpeningHoursAnswer> {
+    override val commitMessage = "resurvey opening hours"
+    override val icon = R.drawable.ic_quest_opening_hours
+    override fun getTitle(tags: Map<String, String>) = R.string.resurvey_opening_hours_title
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        return arrayOf(tags["name"]!!)
+    }
```
westnordost: Isn't that the default?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+    //
+    // objects where tag was surveyed and confirmed to be correct by adding check_date tags
+    // are with larger delay, as it means this info on this specific object is relatively stable
+    protected fun getQueryPart(objectType: String, nameOfGeneratedGroup: String, reviewIntervalInDays: Int) =
+            "$objectType[name][opening_hours]['opening_hours:signed'!='no'](if:!is_date(t['check_date:opening_hours']) || date(t['check_date:opening_hours']) < " +
+                    "date('${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z')) -> .old_opening_hours_tag;\n" +
+                    "$objectType[name][opening_hours](newer: '${DateUtil.getOffsetDateString(-reviewIntervalInDays)}T00:00:00Z') -> .recently_edited_objects_with_opening_hours;\n" +
+                    "(.old_opening_hours_tag; - .recently_edited_objects_with_opening_hours;)-> $nameOfGeneratedGroup;\n"
+
+    override fun isApplicableTo(element: Element) = null
```
westnordost: Whether it is applicable or not is easily determinable by looking at the element's properties, shouldn't it?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+    // will cause OSM elements to become ineligible for this quest for reviewIntervalInDays days.
+    // It allows supporting check_date and any other survey markers without parsing of any tags.
+    //
+    // objects where tag was surveyed and confirmed to be correct by adding check_date tags
+    // are with larger delay, as it means this info on this specific object is relatively stable
+    protected fun getQueryPart(objectType: String, nameOfGeneratedGroup: String, reviewIntervalInDays: Int) =
+            "$objectType[name][opening_hours]['opening_hours:signed'!='no'](if:!is_date(t['check_date:opening_hours']) || date(t['check_date:opening_hours']) < " +
+                    "date('${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z')) -> .old_opening_hours_tag;\n" +
+                    "$objectType[name][opening_hours](newer: '${DateUtil.getOffsetDateString(-reviewIntervalInDays)}T00:00:00Z') -> .recently_edited_objects_with_opening_hours;\n" +
+                    "(.old_opening_hours_tag; - .recently_edited_objects_with_opening_hours;)-> $nameOfGeneratedGroup;\n"
```
westnordost: This would be slightly easier readable if you used a raw string (=multiline string).

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+                getQueryPart("way", wayGroupName, reviewIntervalInDays) +
+                getQueryPart("relation", relationGroupName, reviewIntervalInDays) +
+                "($nodeGroupName; $wayGroupName; $relationGroupName;);\n" +
+                getQuestPrintStatement()
+    }
+
+    // Note that newer segment will ensure that any edit,
+    // including adding or updating review marker like check_date or survey:date tags
+    // will cause OSM elements to become ineligible for this quest for reviewIntervalInDays days.
+    // It allows supporting check_date and any other survey markers without parsing of any tags.
```
westnordost: I don't understand this comment. What segment?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+    private fun getOverpassQuery(bbox: BoundingBox): String {
+        val groupName = ".old_opening_hours"
+        val nodeGroupName = groupName + "_nodes"
+        val wayGroupName = groupName + "_ways"
+        val relationGroupName = groupName + "_relations"
+        val reviewIntervalInDays = 380
+        return bbox.toGlobalOverpassBBox() + "\n" +
+                getQueryPart("node", nodeGroupName, reviewIntervalInDays) +
+                getQueryPart("way", wayGroupName, reviewIntervalInDays) +
+                getQueryPart("relation", relationGroupName, reviewIntervalInDays) +
```
westnordost: Couldn't you use `nwr`?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    override fun download(bbox: BoundingBox, handler: MapDataWithGeometryHandler): Boolean {
+        return overpassServer.getAndHandleQuota(getOverpassQuery(bbox)) { element, geometry ->
+            if (element.tags != null) {
+                // require opening hours that are supported
+                if (OpeningHoursTagParser.parse(element.tags["opening_hours"]!!) != null) {
+                    handler.handle(element, geometry)
+                } else {
+                    Log.wtf("AAAA", "opening_hours=" + element.tags["opening_hours"] + " was rejected as not representable in SC")
```
westnordost: This should go before this is released

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
```diff
@@ -0,0 +1,463 @@
...
+import de.westnordost.streetcomplete.quests.opening_hours.model.OpeningMonths
+import de.westnordost.streetcomplete.quests.opening_hours.model.TimeRange
+import de.westnordost.streetcomplete.quests.opening_hours.model.Weekdays
+import java.io.ByteArrayInputStream
+
+object OpeningHoursTagParser {
+    // returns null for values that are invalid or not representable in
+    // StreetComplete opening hours edit widget
+    // otherwise returns data structure that can be directly used to
+    // initialize this editing widget
```
westnordost: Documentation comments should be written like this `/** */`

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
```diff
@@ -0,0 +1,463 @@
...
+import ch.poole.openinghoursparser.*
+import de.westnordost.streetcomplete.quests.opening_hours.adapter.OpeningMonthsRow
+import de.westnordost.streetcomplete.quests.opening_hours.adapter.OpeningWeekdaysRow
+import de.westnordost.streetcomplete.quests.opening_hours.model.CircularSection
+import de.westnordost.streetcomplete.quests.opening_hours.model.OpeningMonths
+import de.westnordost.streetcomplete.quests.opening_hours.model.TimeRange
+import de.westnordost.streetcomplete.quests.opening_hours.model.Weekdays
+import java.io.ByteArrayInputStream
+
+object OpeningHoursTagParser {
```
westnordost: This should be a normal class like any other, injected the same like other dependencies of classes because otherwise it is an implicit dependency, makes unit testing of classes that use this class harder and obscures the real dependencies of a class.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
```diff
@@ -0,0 +1,463 @@
...
+        // https://github.com/simonpoole/OpeningHoursParser/pull/24
+        val input = ByteArrayInputStream("".toByteArray())
+        val parser = OpeningHoursParser(input)
+        try {
+            val rules = parser.rules(true)
+            return rules[0]
+        } catch (e: ParseException) {
+            e.printStackTrace()
+            throw RuntimeException()
+        }
```
westnordost: Debugging code leftovers

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    // Note that newer segment will ensure that any edit,
+    // including adding or updating review marker like check_date or survey:date tags
+    // will cause OSM elements to become ineligible for this quest for reviewIntervalInDays days.
+    // It allows supporting check_date and any other survey markers without parsing of any tags.
+    //
+    // objects where tag was surveyed and confirmed to be correct by adding check_date tags
+    // are with larger delay, as it means this info on this specific object is relatively stable
+    protected fun getQueryPart(objectType: String, nameOfGeneratedGroup: String, reviewIntervalInDays: Int) =
+            "$objectType[name][opening_hours]['opening_hours:signed'!='no'](if:!is_date(t['check_date:opening_hours']) || date(t['check_date:opening_hours']) < " +
```
westnordost: The query should also take into account the other alternatives of check_date, if there are (still) any around.

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
```diff
@@ -0,0 +1,397 @@
...
+
+    @Test
+    fun `reject unimplemented for now groups of days`() {
+        Assert.assertEquals(OpeningHoursTagParser.parse("Mo,Tu,Sa 09:00-20:00"), null)
+    }
+
+    @Test
+    fun `reject multiple lists of days in a single rule, as it would change form on processing, possibly with unexpected and unwanted results`() {
+        Assert.assertEquals(OpeningHoursTagParser.parse("Mo-Fr 7:30-18:00, Sa-Su 9:00-18:00"), null)
+    }
```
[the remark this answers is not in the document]
matkoniecz: As two separate rules. `Mo-Fr 7:30-18:00; Sa-Su 9:00-18:00`

In theory it would be possible to convert `Mo-Fr 7:30-18:00, Sa-Su 9:00-18:00` into `Mo-Fr 7:30-18:00; Sa-Su 9:00-18:00` but I went with rejecting this instead (at least for now) as it

- would get quite complex
- this syntax is fairly rare
- would be a structural, not a cosmetic change and I decided to go with "cosmetic changes only"

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: It is checking just `check_date:opening_hours` ( https://taginfo.openstreetmap.org/keys/check_date%3Aopening_hours#overview )

For other variants see https://taginfo.openstreetmap.org//search?q=opening_hours

There is `opening_hours:check_date` with just 329 uses (on the other hand, it is 1/3 of `check_date:opening_hours`). But I think it is OK to not support this.

Later things with 15 uses and less - see say https://taginfo.openstreetmap.org/keys/survey%3Adate%3Aopening_hours

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: `Log.wtf("AAAA"` should certainly go away and be fixed, but I think that some way of logging "some oh quests were discarded as syntax was unsupported" would be useful.

Though I have no strong opinion here.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
matkoniecz: Not exactly. It catches ParseException and rethrows RuntimeException.

ParseException must be caught at some level and RuntimeException is allowed to not be caught.

And given that this will not happen except situation of library making some incompatible changes I think that `RuntimeException` is OK here.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: I was not considering it and it seems possible to implement it.

But would it be ever usefully triggered?

Checking `element.timestamp` ensures that it will never appear as result of other edit. For example adding `name` tag may allow to start passing "has name check" but checking new recent timestamp would fail it anyway.

After undo timestamp would cause it to immediately fail check anyway.

It will only make difference in this element getting removed is some obscure cases. Like `shop=whatever` + `name=not edited since 2006` + `opening_hours=Mo 9:00-9:02` and user solving first "add `wheelchair` tag" quest, what would update timestamp and cause this quest to disappear. What would not be useful anyway.

And not checking timestamp would lead to this quest generated after user adds `opening_hours` tag what also would be not OK.

So I would probably move this explanation to comment explaining why it is not implemented.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: I thought that `nwr` is unavailable, but I turned out to be mistaken. Thanks, this should make query significantly more readable.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
matkoniecz: Thanks, should be now fixed here and elsewhere.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: Reformatted to

```
    /** Note that 'newer:' query part will ensure that any edit,
     * including adding or updating review marker like check_date or survey:date tags
     * will cause OSM elements to become ineligible for this quest for reviewIntervalInDays days.
     * It allows supporting check_date and any other survey markers without parsing of any tags.
     * In addition, objects where tag was surveyed and confirmed to be correct by adding check_date:opening_hours
     * are with larger delay, as it means this info on this specific object is relatively stable
     */
```

is it now more clear? It is above

```
    private fun getOverpassQuery(bbox: BoundingBox): String {
        val reviewIntervalInDays = 380
        val date = "'${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z'"
        val dateTrippledTime = "'${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z'"
        return bbox.toGlobalOverpassBBox() + """
            nwr[name][opening_hours]['opening_hours:signed'!='no']
                (if:
                    !is_date(t['check_date:opening_hours']) ||
                    date(t['check_date:opening_hours']) < date($dateTrippledTime)
                ) -> .old_opening_hours_tag;

            nwr[name][opening_hours](newer: $date) -> .recently_edited_objects_with_opening_hours;

            (.old_opening_hours_tag; - .recently_edited_objects_with_opening_hours;);
            """.trimIndent() + getQuestPrintStatement()
}
```

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
```diff
@@ -0,0 +1,397 @@
...
+    fun `accept simple valid opening hours`() {
+        Assert.assertNotEquals(OpeningHoursTagParser.parse("Mo 09:00-20:00"), null)
+    }
+
+    @Test
+    fun `test full structure of a simple opening hour`() {
+        val returned = OpeningHoursTagParser.parse("Mo 09:00-20:00")
+        Assert.assertNotEquals(returned, null)
+        Assert.assertEquals(returned!!.size, 1)
+        Assert.assertEquals(returned[0].months, OpeningMonthsRow().months)
```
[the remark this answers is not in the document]
matkoniecz: I am not sure that `require(dates.size == 1)` -> `dates.single()` is increasing readability, the second one is not a clear assert to me (though I can change it anyway).

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: Good idea, changed.

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
```diff
@@ -0,0 +1,397 @@
...
+    @Test
+    fun `reject intervals`() {
+        Assert.assertEquals(OpeningHoursTagParser.parse("Mo 10:00-16:00/90"), null)
+    }
+
+    @Test
+    fun `issue #26 in OpeningHoursParser is not crashing parser`() {
+        Assert.assertEquals(OpeningHoursTagParser.parse("8095652114"), null)
+    }
+}
```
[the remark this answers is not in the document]
matkoniecz: Test should be now far more readable, thanks for a good ideas!

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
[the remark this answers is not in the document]
westnordost: Ok just wanted to let you know of this function

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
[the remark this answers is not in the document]
westnordost: > As two separate rules. Mo-Fr 7:30-18:00; Sa-Su 9:00-18:00

Ah, that's what you mean. Well, are you actually sure about this? I think the current state might be that StreetComplete uses the `;` unless the weekdays overlap.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
westnordost: You are right. The only situation where an implementation here would be meaningful if there would be another download mode in the future that downloads all the quests at once by doing a normal bbox query. (I was thinking about that some time ago because the more quests we have, the longer the time to download them all in a bbox while downloading *everything* within a bbox is of constant time.)

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
```diff
@@ -0,0 +1,82 @@
...
+     * including adding or updating review marker like check_date or survey:date tags
+     * will cause OSM elements to become ineligible for this quest for reviewIntervalInDays days.
+     * It allows supporting check_date and any other survey markers without parsing of any tags.
+     * In addition, objects where tag was surveyed and confirmed to be correct by adding check_date:opening_hours
+     * are with larger delay, as it means this info on this specific object is relatively stable
+     */
+    private fun getOverpassQuery(bbox: BoundingBox): String {
+        val reviewIntervalInDays = 380
+        val date = "'${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z'"
+        val dateTrippledTime = "'${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z'"
```
westnordost: Copy-paste mistake?

matkoniecz: yes :(

westnordost: Also, better define two constants, one for openingHoursReviewInterval and one for elementReviewInterval (or something) than this triple-stuff.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
westnordost: Jup

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
westnordost: Only us developers are going to see this log.

If we want reliable statistics, better would be a (global) analysis of how much % of opening hours use which syntax feature. Should be interesting. One could use the OpeningHoursParser library in a small java application that downloads all the opening hours (in the world) and counts features. Maybe Simon already did that.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
westnordost: In Kotlin, there are no checked exceptions. You sure you need to catch it?

Anyway: Ok

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
westnordost: There is `opening_hours:lastcheck` with 1.6k uses.
I agree `check_date` is the more widely adapted tag and is also documented as opposed to `lastcheck`, but spread of `lastcheck` will not vanish over night, so that should IMO be supported.

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
```diff
@@ -0,0 +1,174 @@
...
+        assertNotNull(OpeningHoursTagParser.parse("Mar-Oct Tu-Su 10:30-18:00; Mar-Oct Mo 10:30-14:00; Nov-Dec Tu-Su 11:00-17:00; Nov-Dec Mo 11:00-14:00"))
+        assertNotNull(OpeningHoursTagParser.parse("Th 17:30-19:30 open"))
+        assertNotNull(OpeningHoursTagParser.parse("Mo-Su 09:00-12:00, 13:00-14:00")) // allow time gap
+        assertNotNull(OpeningHoursTagParser.parse("Su-Mo 09:00-12:00")) // try to trigger bugs with range going end of week
+        assertNotNull(OpeningHoursTagParser.parse("Fr-Mo 09:00-12:00")) // try to trigger bugs with range going end of week
+        assertNotNull(OpeningHoursTagParser.parse("Mo 17:30-19:30; Th 17:00-19:00")) // multiple rules, no overrides
+
+        // over weekend rules, without triggering collision detector
+        assertNotNull(OpeningHoursTagParser.parse("Su-Mo 09:00-12:00; Tu-Sa 10:10-10:11"))
+        assertNotNull(OpeningHoursTagParser.parse("Fr-Mo 09:00-12:00; Tu-Th 10:10-10:11"))
```
westnordost: Maybe for all the rules that are accepted it would make sense to also check if parsing and saving again results in the same syntax?

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
```diff
@@ -0,0 +1,174 @@
...
+        assertEquals(returned[0].months, OpeningMonthsRow().months)
+        assertEquals(returned[0].weekdaysList.size, 2)
+        assertEquals(returned[0].weekdaysList[0].timeRange.start, 9 * 60)
+        assertEquals(returned[0].weekdaysList[0].timeRange.end, 12 * 60)
+        assertEquals(returned[0].weekdaysList[0].timeRange.isOpenEnded, false)
+        assertEquals(returned[0].weekdaysList[1].timeRange.start, 13 * 60)
+        assertEquals(returned[0].weekdaysList[1].timeRange.end, 14 * 60)
+        assertEquals(returned[0].weekdaysList[1].timeRange.isOpenEnded, false)
+        val weekData = BooleanArray(7) { true }
+        assertEquals(returned[0].weekdaysList[0].weekdays, Weekdays(weekData))
```
westnordost: Actually, the first parameter in `assertEquals` is what is expected and the second is the actual string. Not that big of a problem, just saying.

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
matkoniecz: And parsing and saving again will check anyway whatever parsing is possible, so `assertNotNull` tests may be dropped.

Thanks to more compact test form such check is now reasonable.

matkoniecz: `Jun Mo 10:00-11:00;Jul Mo 11:00-12:00` is now formatted as `Jun Mo 10:00-11:00; Jul Mo 11:00-12:00` what is both more readable and preferred by https://openingh.ypid.de/evaluation_tool/

Thanks for the idea

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: I added `changes.deleteIfExists("opening_hours:lastcheck")` on solving.

In query I think it may be skipped. Edit adding `opening_hours:lastcheck` will update timestamp (like any edit), so for 380 days resurvey quest will not be created.

But 1080 days of resurvey delay done for `check_date:opening_hours` should be OK to skip as it is below 2000 uses worldwide and increase to query complexity seems not worth it .

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: I was thinking rather about making debugging "why this quest is not generated" easier. More than once I though that  found some problem only to discover on looking into logs that it is 2000 object limit, or note on the element or something similar.

matkoniecz: > One could use the OpeningHoursParser library in a small java application that downloads all the opening hours (in the world) and counts features. Maybe Simon already did that.

Done already, but some time ago :) See https://github.com/simonpoole/OpeningHoursParser#openinghoursparser

> It parses 146'825 (91%) of 161'268 unique test strings in non-strict mode. The remaining 14'443 are likely valid errors, spot checking shows that they have obvious issues. In strict mode further 22'440 fail (total 36'883).

I also run a similar code, but I was looking only for crashes to ensure that this code will not crash on weird tags. And even found something, see https://github.com/simonpoole/OpeningHoursParser/issues/26 . It was done as a monstrous hack - see https://github.com/matkoniecz/Zazolc/commit/7245406f9538c9a880a0b8805df348f347f21baa

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz:
```
val resurveyCutoff = "'${DateUtil.getOffsetDateString(-reviewIntervalInDays)}T00:00:00Z'"
val repeatedResurveyCutoff = "'${DateUtil.getOffsetDateString(-reviewIntervalInDays * 3)}T00:00:00Z'"
```

app/src/main/res/values/strings.xml
matkoniecz: I mostly upgraded UX to this version. For now remaining things is `24/7` support and:

> make the opening hours unclickable first

Is it about avoiding accidental clicks resulting in confusion? It is not obvious that entries are clickable, but I think that it is OK to let user edit them also at this screen.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
matkoniecz: injection dependency is part where my understanding is weakest of all what is happening in this app (and I probably should change that), but `OpeningHoursTagParser` is used a bag of functions, it is not used to create objects. Like `DateUtil`.

I have no idea how dependency injection would work here. Drop static functions and start creating and passing `OpeningHoursTagParser` object?

Is there any other class that is handled in a similar way?

app/src/main/res/values/strings.xml
westnordost: Well, yes, the user would not expect the form to be editable at this point.
Easiest way to implement this is to put a transparent clickable view on top of the recycler view. Not sure if setting the recycler view to disabled would do anything.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
westnordost: Oh, so #26 was never fixed and you had to workaround it?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
westnordost: Just replace `object` with `class` in line 12, then at every place where you use  `OpeningHoursTagParser`, declare it as a constructor parameter. In the test, of course, you create it yourself.

See for example `AddPlaceName` which depends on (a Future of) the `FeatureDictionary`

westnordost: There is no magic in dependency injection. It basically just means that every class that uses another class should get that classed passed in and not create it itself.

```
class Car(val motor: Motor)
```
^- Car depends on a motor, this dependency is injected by whoever instantiated the car.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: #26 seems unrelated - it is about a wheelchair access quest

westnordost: I meant https://github.com/simonpoole/OpeningHoursParser/issues/26

app/src/main/res/values/strings.xml
matkoniecz: > Easiest way to implement this is to put a transparent clickable view on top of the recycler view.

Seems that recycler adapter needs to be edited anyway to get rid of delete buttons.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/ResurveyOpeningHours.kt
matkoniecz: Was fixed in 17.1 - see https://github.com/simonpoole/OpeningHoursParser/releases

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
matkoniecz: I modified the class in https://github.com/westnordost/StreetComplete/pull/1676/commits/3ddd5b4bf22b2460b33ce96d58f5d4388724444f, it should now use DI and it is also still working.

> There is no magic in dependency injection. It basically just means that every class that uses another class should get that class passed in and not create it itself.

Oh, basic idea for dependency injection is clear for me. It is so simple that I independently noticed it as a good idea during refactoring one of my projects (long time ago, before I even learned about terms like "dependency injection" or "refactoring" and before I was using git).

But in practice Dagger is for my highly confusing. Whenever I learn about @Module vs @Component difference I will forget which one was doing what within 20 minutes.

For more specific example - why I added https://github.com/westnordost/StreetComplete/pull/1676/files#diff-4d3f82a0d1eeea6b01ea730c337174d7 ?

ApplicationComponent is a `@Component` so it will be (somehow) used by Dagger to generate code used by ApplicationModule. 

But I admit that adding `void inject(ResurveyOpeningHoursForm resurveyOpeningHoursForm);` in this PR was basically a magic incantation to exercise compilation error without understanding why I am doing this.

I again run went through a bit of official Dagger docs, code handling Dagger, code generated by Dagger, and some tutorial found somewhere. It seems that I understand a bit more now, but I am not entirely sure about this.

matkoniecz:
So it seems that three things are remaining:

- remove delete button before user marks opening hour as incorrect
- remove ability to delete parts of oh data before selecting "no" button
- potentially adding support for multiple nonintersecting day ranges within a single rule `Mo-Fr 7:30-18:00, Sa-Su 9:00-18:00` (this one may be done later as a separate PR, seems to not be a blocker)

Is there anything else that I missed?

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParserTest.kt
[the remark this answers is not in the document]
matkoniecz: `Mo,Tu,Sa 09:00-20:00`, `Mo-Th,Sa 09:00-20:00`, `Mo-Th,PH 09:00-20:00` - nonintersecting ranges, shared hours are supported by SC and could be supported also by this quest.

`Mo-Fr 7:30-18:00, Sa-Su 9:00-18:00` makes sense but is not supported by SC, due to differing hours it is represented as two separate rules. Splitting such rules into multiple ones would work (as long as ranges are not intersecting etc) would work but for me it is changing structure what I would prefer to avoid, at least in the initial PR.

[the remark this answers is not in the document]
matkoniecz: I updated tests to describe this two situations separately - see https://github.com/westnordost/StreetComplete/pull/1676/commits/8096d23452d887da1ae3357ace53ba3842862639

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/OpeningHoursTagParser.kt
westnordost: To be honest, I never got so deep into Dagger myself, I only worked myself into this as far as I needed to to have everything working as I want it to be. It could be that the module / component code in this project is not best practice.

The `void inject(ResurveyOpeningHoursForm resurveyOpeningHoursForm);` in the interface of the ApplicationComponent is just to tell Dagger that it should also generate a method (this method) which should inject dependencies into an already created object when invoked. This is necessary, because in Android development, we can not simply create new activities and new fragment objects ourselves using the constructor. The Android system creates these objects for us, so the "normal" way we'd go for dependency injection is not available. The workaround is to have the class inject its dependencies after construction itself, by calling the method of the application component itself.
In my opinion, this is only a workaround for this situation and is less than ideal. For example, this makes unit testing of that fragment more difficult again because the dependencies are indeed not really injected but the fragment injects them itself by telling Dagger to inject them.

westnordost:
Yes, the duplicate(?) data model, see my last non-code-comment.

matkoniecz:
I am not entirely happy about my solution with special layouts, adapter and object for read only opening hours but at least it works.

I was unable to figure out better working way to get rid of delete buttons.

So what remains now:

- duplicate model
- retesting whatever I managed to break something during making fixes (especially the model thing)

Additional potential bonus features, that may be done also as a separate PRs and are not blocking merge (at least I think so):
- adding support for multiple nonintersecting day ranges within a single rule, with share hours `Mo-Th,Sa 09:00-20:00`
- adding support for `24/7`

matkoniecz:
one new thing to fix (except merge conflicts) is to move yes/no button in the initial answer ("is this OH still correct") into the same row as "can't say"

matkoniecz:
OK, now I am officially stuck and out of ideas. Help with dealing with one issue would be welcome because I have no idea what happened.

`.joinToString("; ")` returns null - but only when in function and when tested. It works otherwise.

Problem is with 3ddd5b4bf22b2460b33ce96d58f5d4388724444f (use dependency injection for OpeningHoursTagParser),
and it affects solely AddOpeningHoursTest ("apply opening hours answer" one), that is failing with null pointer exception.

I tested it and real quest actually works. The same code coped into fubnction (not called) works.

I suspect that it may be related to mocks, but I have not found even a way to test whatever given object is mocked - and variable listing in debug is not indictaing that parser is a mock in this case.

I also suspect that mocks are dead-end as debugger shows me stepping into internalIntoTag function! (though in a bit weird state, and is unable to show state of variables).

But I have no idea what else may be happening.

See https://github.com/matkoniecz/Zazolc/commit/7f9f1336126b92392e362f97749f8567bd7a99ec
and https://github.com/matkoniecz/Zazolc/commit/24aa74c4d72dcbbeade20794b8351c16b2127064
at https://github.com/matkoniecz/Zazolc/commits/survey_failing_test

I will look into more but hopefully I am missing something obvious.

westnordost:
I found one thing in  3ddd5b4, maybe that is the cause of the problem you have.

westnordost:
Ok I see the error I found has nothing to do with the failing tests. I'll check it out and have a look.

westnordost:
I had a quick look. Likely the problem is that even though `AddOpeningHours` does not need to parse a string with the opening hours, it still depends on `OpeningHoursTagParser` (I didn't check why yet) to correctly convert the user input into the opening hours tag.
So if you just mock `OpeningHoursTagParser` in the tests, you need to at least also mock its expected response for whatever function `AddOpeningHours` uses of it.

westnordost:
I created a PR on your PR: https://github.com/matkoniecz/Zazolc/pull/1/files

matkoniecz:
Also, I need to process notes from https://github.com/westnordost/StreetComplete/commit/3ddd5b4bf22b2460b33ce96d58f5d4388724444f

matkoniecz:
@westnordost 

I suspect that it is not intentional, but I want to check.

Why `toOpeningWeekdays` is merging only rules that directly next to each other?

For example rules about Monday, Monday, Tuesday will be merged by this function. Rules about Monday, Tuesday, Monday will not be merged by this function and may be partially merged in the next step (if time is not overlapping).

https://github.com/westnordost/StreetComplete/blob/1177bf20f4f55449b3628bf1cb577a8aead6aad2/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/adapter/OpeningHoursModelCreator.kt#L13

see https://github.com/matkoniecz/Zazolc/commit/d6f6151d20df0986727325956b3d410ad2a5ebec for some tests documenting this usage

I plan to switch it to always merging rule in this step, if `weekdays` is the same.

------------------------

I finally had time today to do some SC coding and nearly eliminated `OpeningMonths`.

I started going through and removing also `OpeningWeekdays` but it ended with functions of signatures such as `private fun List<List<OpeningWeekdaysRow>>.toWeekdaysClusters(): List<List<List<OpeningWeekdaysRow>>>` what I do not see as an improvement (unfinished wip is [here](https://github.com/matkoniecz/Zazolc/blob/a17bfd98892447ddaf51be4438231583fcfdc62d/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/adapter/AddOpeningHoursAdapter.kt)).

I plan to keep `OpeningWeekdays` without removing them, used solely in generating tag value of `opening_hours`.

It would allow to keep one depth of function within an object, ending with `private fun List<OpeningWeekdays>.toWeekdaysClusters(): List<List<OpeningWeekdays>>`.

westnordost:
> Why toOpeningWeekdays is merging only rules that directly next to each other?
> For example rules about Monday, Monday, Tuesday will be merged by this function. Rules about Monday, Tuesday, Monday will not be merged by this function and may be partially merged in the next step (if time is not overlapping).

Because the opening hours should be WYSISWG. If the user explicitly did not add them below each other, he will have had a reason for that and thus probably not want that the rules are optimized in the background. The reason could be that this is how the opening hours are written on the sign. Especially looking at resurveying these opening hours, it is easier if the opening hours are organized in the app like on the sign.

westnordost:
> I plan to keep OpeningWeekdays without removing them, used solely in generating tag value of opening_hours.

I don't understand this decision right now, but maybe later when I see the big picture.

Marc-marc-marc:
could you document:
- what is the logic (timestamp of the metadata ? no change for this key ?) behind triggering a resurvey ?
- what is the key used to store the information that it was checked when nothing has changed ?
- in general, shouldn't we first ask if the object still exists? at the time when I was reading a lot of notes, there was a lot of "impossible to x, the POI has been closed for X years"

Helium314:
> Because the opening hours should be WYSISWG. If the user explicitly did not add them below each other, he will have had a reason for that and thus probably not want that the rules are optimized in the background. The reason could be that this is how the opening hours are written on the sign. Especially looking at resurveying these opening hours, it is easier if the opening hours are organized in the app like on the sign.

Just some comment on this:
a) this is not mentioned in the app. I had previously expected that opening hours are optimized in the background.
b) I am not willing to enter some (typically bank) opening hours like
Mon 9:00 AM – 12:30 PM 1:30 PM – 4:00 PM
Tue 9:00 AM – 12:30 PM 1:30 PM – 4:30 PM
Wed 9:00 AM – 12:30 PM 1:30 PM – 4:00 PM
Thu 9:00 AM – 12:30 PM 1:30 PM – 5:00 PM
Fri 9:00 AM – 12:30 PM 1:30 PM – 3:00 PM
day by day. Instead I just set Mon - Fri 9:00 AM – 12:30 PM and then do the rest (which admittedly looks ugly in the opening_hours tag, but takes long enough already)
c) if optimization should be implemented, some `don't optimize` button could be added

westnordost:
So I will take this over in frame of the microgrant

westnordost:
This is obsolete now, new PR #2045

Issue #1575, opened 2019-09-19
On failure to upload photo try again later


westnordost:
When the upload of a note (comment) with attached photo fails due to a non-permanent error such as a socket timeout or connection error, keep the to-be-created-note in local storage and try again on next upload.

Alternative would be to let the whole upload be cancelled if the photo-upload-service is not reachable. Same as is done for any issue when communicating with the OSM API for uploading notes and map changes.

Follow-up of #1569

westnordost:
Requires the branch kotlin-data to be merged first, to avoid merge conflicts.

PR #1683, opened 2019-12-11
Kotlin data

westnordost:
Objective of this refactor (moving to Kotlin):
- modernizing core code that is now 2-3 years old (try to 
- making code more concise (reducing code by a third) and (hopefully) easier to understand
- unfortunately I mixed in a feature: #1575


app/src/main/java/de/westnordost/streetcomplete/data/osm/Countries.kt
```diff
@@ -0,0 +1,35 @@
...
+    constructor(vararg exceptions: String) : this(exceptions.toList())
+}
+
+/** Whether these countries contain the given position.
+ *  Needs a CountryBoundaries to calculate that. */
+fun Countries.containsPosition(pos: LatLon, boundaries: CountryBoundaries) = when(this) {
+    is AllCountries -> true
+    is AllCountriesExcept -> !boundaries.isInAny(pos, exceptions)
+    is NoCountriesExcept -> boundaries.isInAny(pos, exceptions)
+}
```
westnordost: Hm, or 
```kotlin
/** whether the given position is in set of the given countries */
CountryBoundaries.isInAny(pos: LatLon, countries: Countries)
``` ?

"Needs a CountryBoundaries to calculate that" is maybe a sign that the function is a little counterintuitive.

PR #1694, opened 2020-01-03
Expand leaf type quest to also include tree rows

### LINKED ISSUE #1688: Expand leaf type quest to also include tree rows
**Use case**
Same as leaf type for Forests/woods

**Proposed Solution**
Since there was already discussion on doing single trees which was rejected
This should be acceptable since tree rows are usually relatively big

MegaArthur:
Closes  #1688


westnordost:
For tree rows, it would probably make sense to enable the possibility to split ways right? Could you add this and test that it does not affect the area-types?

MegaArthur:
> For tree rows, it would probably make sense to enable the possibility to split ways right? Could you add this and test that it does not affect the area-types?

Is just the override required to enable way splitting?

matkoniecz:
To enable way split? Yes. But note "test that it does not affect the area-types".

MegaArthur:
Okay, going to test it tomorrow or the day after

MegaArthur:
@westnordost Tested, the way split appears on tree rows but not forest areas, all is good.

westnordost:
Nice, good! Thanks for the PR!

PR #1696, opened 2020-01-08
Show AddTactilePavingBusStop quests in Luxembourg

dwaxweiler:
As a local in Luxembourg, I can confirm that more and more bus stops get tactile paving. Therefore, I have separated the whitelists for AddTactilePavingCrosswalk and for AddTactilePavingBusStop and added Luxembourg to the latter.


westnordost:
Nice, thank you for your local insight!

westnordost:
And also for posting a PR yourself about that

dwaxweiler:
You are welcome! Thanks for your quick response!

PR #1697, opened 2020-01-11
Update AddVegan.kt for ice cream shops

ferdinand0101:
add vegan tags to ice cream stores


westnordost:
Looks like I overlooked this one. Looks good and reasonable! I will optimize the query after the merge, thank you!

Issue #1716, opened 2020-02-02
Clarity in building type quest (house/garage)


smichel17:
### How do I tag a multi-car garage?

- "Single garage" shows a single garage door and single ridge along the top.
- "Multiple garages" shows three garage doors and three ridges along the top.

It's not clear how I should tag a 2-car garage, with two garage doors (or one very wide one) under a single roof (one ridge).

### Barn hybrids

There are a couple buildings in my area that are old barns, that are no longer on a farm (it has been converted to residential area) and are currently being used as garages or sheds. And one that I believe is a newly-constructed garage, made to look like a barn from the outside. Where is the line for whether I should mark these as a shed/garage or barn?

### Housing

- It's not clear to me exactly when the plain "House" building type should be used. It seems like a catch-all for residential buildings that don't fit any of the more specific labels, but you can already just select "Residential" without being more specific, for that case.
    - The picture looks like it is not attached to any other houses (but is close by). What distinguishes a "house" from a "detached house"?
    - Edit: I've found #25 which seems to have more information, and am reading it now. Maybe I will have wording suggestions after I finish.
- The "Row house" picture shows each of the houses being identical. Should it still be marked as a row house if the format is A-B-B-A (in my case, A's are mostly on the second story while B's are first-story). Or should it be an apartment building? (Or just "House"?)
- How should I mark a building that fits the semi-detached house description, but has 3 homes? The 3rd is larger and sits on the second floor, above the other two, which are rental units — it belongs to a friend of mine.

matkoniecz:
What is your language in SC? Maybe problem is in the translation?

Answering questions (but is there way to improve SC to avoid this question):
> It's not clear to me exactly when the plain "House" building type should be used. It seems like a catch-all for residential buildings that don't fit any of the more specific labels, but you can already just select "Residential" without being more specific, for that case. 

`house` ( https://wiki.openstreetmap.org/wiki/Tag:building%3Dhouse ) is  for buildings constructed for a single family.

`residential` (https://wiki.openstreetmap.org/wiki/Tag:building%3Dresidential) is a general term

> The picture looks like it is not attached to any other houses (but is close by). What distinguishes a "house" from a "detached house"?

House may or may not be attached to other houses. Detached house is not.

matkoniecz:
[...]

PR #1720, opened 2020-02-03
Add description for multiple garages

smichel17:
For #1716

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypes.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypes.kt
@@ -45,7 +45,7 @@ enum class BuildingType(val item:Item<String>) {
 
     CARPORT       (Item("carport",    R.drawable.ic_building_carport,    R.string.quest_buildingType_carport, R.string.quest_buildingType_carport_description)),
     GARAGE        (Item("garage",     R.drawable.ic_building_garage,     R.string.quest_buildingType_garage)),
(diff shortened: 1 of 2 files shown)


westnordost:
Hm actually it was deliberate to not act on your suggestion, because I don't know how this additional information would be helpful to remove doubt.

smichel17:
My problem was: "The garage in front of me has two doors and one roof. What should I mark it as?"

- It did not obviously fit the single garage picture because the single garage picture only has one entrance
- It did not obviously fit the multiple garages picture because the multiple garages picture has multiple roofs.

westnordost:
But it is good that it does match neither, because in the case described, the user ought to use his judgement (or in doubt, open a note) to either sort this garage as a garage or as a garages: Some garages with 2 doors might correctly be tagged with garage, others with garages.

smichel17:
> the user ought to use his judgement

That's reasonable, but the judgement should be about whether it fits the criteria for single or multiple, not what those criteria are (this is the job of the more-involved OSM community, or in cases where there is not consensus but you still want to offer the quest, yours).

In this case, it seems like the main criteria are having different owners/users/purposes, but those criteria are not mentioned in the quest, so users have to figure that out based on the name and icon.

> (or in doubt, open a note)

Maybe this is a personal problem, but I tend to skip quests where in doubt, rather than add a note.
Perhaps the UI could push me towards notes more, but that's a separate issue.

westnordost:
Hmm, okay. After all, your addition is 1:1 what is in the wiki, so it should provide the necessary information. Will merge for the next major because it contains new strings.

PR #1721, opened 2020-02-03
Move "house" up in list

smichel17:
This makes it more aligned with others of a similar type (attached),
giving an additional hint as to the type of house. (Down below, it's
surrounded by other detached houses, farm and hut).

For #1716


westnordost:
Ok

PR #1722, opened 2020-02-03
Add tourism information quest

### LINKED ISSUE #1115: Tourist information type (`information` key for `tourism=information`)
### General
Affected tag(s) to be added: [information](https://wiki.openstreetmap.org/wiki/Key:information) for `tourism=information` which doesn't have it.

Question asked: **What type of tourist information is this?** *(This question needs improvement because, in English, "tourist information" means a tourist information office.)*

[`tourism=information`](https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dinformation) is used for tourist information offices, and guideposts and maps etc. The [`information` key](https://wiki.openstreetmap.org/wiki/Key:information) is used to specify what type it is (`information=office` for a tourist information office). Currently 15% of `tourism=information` do not have a `information` tag ([cf. taginfo](http://taginfo.openstreetmap.ie/tags/tourism=information#combinations)). There is  big difference between a tourist information office and a guidepost! OSM data can be improved by adding this.

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose - Yes, 500k instances of `information`.
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam) Taginfo says 49% are `information=guidepost`, 23% `board`, 10% `map`, 3% `office`.
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort). There are [~150 examples in London](https://overpass-turbo.eu/s/zYd), and [50 in Berlin](https://overpass-turbo.eu/s/zYe).

### Ideas for implementation
`[tourism=information][!information]` overpass filter to select it. [https://overpass-turbo.eu/s/zYb Overpass Turbo link].

There are ~40K `information=add` which could be a note for "add this information type"? Perhaps those objects could be changed? Further research needed.

For implementation, multiple choice based on common values on the [https://wiki.openstreetmap.org/wiki/Key:information OSM wiki for `information`].

I'm a programmer and willing to look at implementing this myself, if it's acceptable.

Dosenpfand:
fixes #1115

Especially the drawables probably need some more work (e.g. suitable ones chosen, properly cropped, etc.), so I am looking forward to feedback.


westnordost:
Flawless implementation, good ordering, thank you!

Three things though:

1. There is now a new directory ` new_img/`, you probably committed that by accident

2. I see you took those tags that are defined in the wiki and ordered it by usage count. That's a good start. The most important part though is that we need to present a reasonable list of choices from the point of view of the surveyor. What I want to say is this:

taginfo gives the total count of current values for the `information=*` key. Those places with a `tourist=information` tag that do not have the subtag set yet will most likely not have the same distribution as those that do. For example, I expect that it is much more likely that people who want to tag route markers somewhere in the forest, will add the subtag `information=route_marker` directly than those that just want to tag a "tourist information" (office), thinking `tourism=information` is enough. Same with other keys that do not translate to something similar. 
So, I expect a much much more offices, boards and terminals among those elements with `tourism=information` that lack the subtag so far than anything else. Certain things I do not expect to appear at a reasonable amount at all.

I suggest the following order:
1. office
2. board
3. terminal
4. guidepost
5. map
6. maybe tactile model & tactile map (same concern as with route marker though)

And drop these:
- audio guide: Never seen a public audio guide somewhere, could you steal the headphones? How is it going to work? If it is a board that contains a QR code to download an app or similar, it would still be a board, I'd say. But anyway, especially with this picture, it's rather confusing what this is supposed to be.
- route marker: As said, I think it is very unlikely that someone who wanted to tag a route marker forgot about adding the tag that makes this element a route marker
- maybe tactile model & tactile map

And if it does not fit, there is always the possibility to leave a note. I expect that this will happen quite often anyway because a lot of things might not fit quite clearly into those main categories because something has been tagged as `tourism=information` but no subtag because the one who added it was not sure how to subtag it.

3. As the images themselves are not too helpful, I think the text must be clearer and at least use the full name of the place:

- Office -> Tourist Information Office
- "Board" is not really clear, maybe needs a description. At least "Information Board" or something?
- "Information Terminal" is perhaps better

Dosenpfand:
> Flawless implementation, good ordering, thank you!

Thanks!

>     1. There is now a new directory ` new_img/`, you probably committed that by accident

You are right, I removed it.

> I suggest the following order:
> 
>     1. office
> 
>     2. board
> 
>     3. terminal
> 
>     4. guidepost
> 
>     5. map
> 
>     6. maybe tactile model & tactile map (same concern as with route marker though)

I changed it to this order and dropped the tactile options.


> And drop these:
> 
>     * audio guide: Never seen a public audio guide somewhere, could you steal the headphones? How is it going to work? If it is a board that contains a QR code to download an app or similar, it would still be a board, I'd say. But anyway, especially with this picture, it's rather confusing what this is supposed to be.

I interpreted this as some kind of stationary audio guide, similar to a terminal, where you can press a button and and the audio starts playing. But they are definetiely not common, and probably in combination with a regular terminal most of the time. I removed it.

>     * route marker: As said, I think it is very unlikely that someone who wanted to tag a route marker forgot about adding the tag that makes this element a route marker

I agree, but does the same then not also apply to "guidepost" to a lesser degree?

>     * maybe tactile model & tactile map

Additionally, they are a bit hard to distinguish, since all the tactile models that i found are also tactile maps, (but not necessarily the other way around). Furthermore, all tactile maps could just be simply tagged as map.

>     1. As the images themselves are not too helpful, I think the text must be clearer and at least use the full name of the place:
> 
> 
>     * Office -> Tourist Information Office
> 
>     * "Board" is not really clear, maybe needs a description. At least "Information Board" or something?
> 
>     * "Information Terminal" is perhaps better

I thought that it is already strongly implied by the question, Q: "What type of **tourist information** is this?"  A: [Tourist information] office.
But I changed it anyways, to make it clearer.

westnordost:
> I agree, but does the same then not also apply to "guidepost" to a lesser degree?

Yes (to a lesser degree). Also a map. Though, you often have those maps with more (tourist) information to the left and right of that map. There are also those guideposts that show the way to the nearest sights or amenities (rather than for hiking) - this is closer to something that can be read "tourist information" than the type of guidepost you have the photo for. 

But still you have a point. Maybe move it to the very last position?

Since it is only 5 now, and the first 3 are the most likely ones, perhaps remove the "show more" button and show them all right away.

> Furthermore, all tactile maps could just be simply tagged as map.

I was going to write that as well, but alas, this is just how the tagging organically grew and is documented, and there is no real alternative.

> I thought that it is already strongly implied by the question

Yes probably, maybe the translators will have problems seeing this in context. I'd expect that not in any language a "tourist information office" has "office" in its name and pulling the proper name for the thing apart might be confusing. Same with board: For example, we are both German, and I wouldn't regard a board showing public transport arrival/departure times as tourist information. Actually, most [board types described](https://wiki.openstreetmap.org/wiki/Key:board_type) are something I wouldn't denote as tourist information in German.

Dosenpfand:
> But still you have a point. Maybe move it to the very last position?

Done.

> Since it is only 5 now, and the first 3 are the most likely ones, perhaps remove the "show more" button and show them all right away.

Done.

> Yes probably, maybe the translators will have problems seeing this in context. 

Right, I was only thinking about English.

HolgerJeromin:
° Additionally, they are a bit hard to distinguish, since all the tactile models that i found are also tactile maps

Aachen Dom has a tactile model of the building in front of it. Definitive no map.

https://www.openstreetmap.org/node/1272915553

rory:
Thanks for doing this. 🙂

I do wonder about "board" vs "map" though. Many boards will also have a map (the [image for a board](https://commons.wikimedia.org/wiki/File:Avon_Valley_Nature_Reserve_Information_Board.jpg) includes a small map). Nearly all maps are physically _on_ a board. Do you think there is a danger that people will mix them up?

westnordost:
Surveyors will need to use their judgement whether something is rather a map or rather a board. I think the amount of places where there is doubt will be rather low, but anyway, if there is doubt, the surveyor can always make a photo and have it discussed in a note.

Issue #1724, opened 2020-02-07
Allow entering 3d mode without first locking on current location


smichel17:
I just discovered 3d mode. It's really cool, and learning about it has made me more interested in completing "number of stories" type quests. However, it was very hard to find. I only learned about it because I saw screenshots on some issues here on the repo, and even after that it took me around 15 minutes of combined reading (to see if anyone said how to do it) and trial and error to figure it out. For anyone else like me: tap the compass twice while your current location is locked on.

A big reason for this is that I don't lock on to my location very much. I do *use* my location, but I prefer to drag around the map to help orient myself and make sure I'm tagging the right building/road/etc. And I use the compass even less, because I prefer to keep the view oriented along the road I'm walking on. Ironically, 3d mode helps with both of those problems, if only I'd known about it.

Anyway, it's possible to unpin the location and scroll around once you've entered 3d mode, so it's unclear to me why you have to be location locked to enter it in the first place. I *have* tapped the compass multiple times (to no effect) when not locked on location, so I believe this would help with discovery.

smichel17:
Ah, I see. It's because compassMode is part of LocationAwareMapFragment and we use location lock to implicitly assume the type of map fragment :smiling_imp:

westnordost:
You can also use the two-finger gesture (shove with two fingers) to tilt the map

westnordost:
So, I am closing this because it is already possible to enter 3D mode by tilting the map, no compass or location required.

smichel17:
Technically yes, but this was more about the discoverability of the feature than the actual functionality.

smichel17:
I was looking into this this morning, but ran out of time; I'll get back to it in the evening.

PR #1725, opened 2020-02-07
On compass press, tilt map if not tilted, when not tracking location

smichel17:
This implementation does not currently work because even when the tilt has been set to 0f, it usually reports being very slightly tilted (eg, tilt=5.317779E-7) and thus, isFlat is almost always false.

If you spam click the compass button, it will occasionally report `tilt=0f` and work properly. I'm not sure exactly why this is.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -54,6 +54,8 @@ import de.westnordost.streetcomplete.quests.*
 import kotlinx.android.synthetic.main.fragment_map_with_controls.*
 import javax.inject.Inject
 import kotlin.math.max
+import kotlin.math.absoluteValue
+import kotlin.math.PI
 
 /** Contains the quests map and the controls for it. */
 class MainFragment : Fragment(R.layout.fragment_map_with_controls),
@@ -397,7 +399,16 @@ class MainFragment : Fragment(R.layout.fragment_map_with_controls),
 
     private fun onClickCompassButton() {
         val mapFragment = mapFragment ?: return
-        val isNorthUp = mapFragment.cameraPosition?.rotation == 0f
+        // Allow a small margin of error around north/flat. This both matches
+        // UX expectations ("it looks straight..") and works around a bug where
+        // the rotation/tilt are not set to perfectly 0 during animation
+        val margin = 0.025f // About 4%
+        // 2PI radians = full circle of rotation = also north
+        val isNorthUp = mapFragment.cameraPosition?.rotation?.let {
+            it <= margin || 2f*PI.toFloat()-it <= margin
+        } ?: false
+        // Camera cannot rotate upside down => full circle check not needed
+        val isFlat = mapFragment.cameraPosition?.tilt?.let { it <= margin } ?: false
         if (!isNorthUp) {
             mapFragment.updateCameraPosition(300) {
                 rotation = 0f
@@ -406,6 +417,12 @@ class MainFragment : Fragment(R.layout.fragment_map_with_controls),
         }
         if (mapFragment.isFollowingPosition) {
             setIsCompassMode(!mapFragment.isCompassMode)
+        } else {
+            if (isNorthUp) {
+                mapFragment.updateCameraPosition(300) {
+                    tilt = if (isFlat) PI.toFloat() / 5f else 0f
+                }
+            }
         }
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.graphics.PointF
 import android.hardware.SensorManager
 import android.location.Location
+import android.view.animation.DecelerateInterpolator
 import android.view.WindowManager
 import androidx.core.content.edit
 import com.mapzen.android.lost.api.LocationRequest
@@ -55,11 +56,14 @@ open class LocationAwareMapFragment : MapFragment() {
     private var zoomedYet = false
 
     /** Whether the view should automatically rotate with the compass (like during navigation) */
+    // Since the with-compass rotation happens with no animation, it's better to start the tilt
+    // animation abruptly and slide out, rather than sliding in and out (the default interpolator)
+    private val interpolator = DecelerateInterpolator()
     var isCompassMode: Boolean = false
         set(value) {
             field = value
             if (value) {
-                controller?.updateCameraPosition { tilt = PI.toFloat() / 5f }
+                controller?.updateCameraPosition(300, interpolator) { tilt = PI.toFloat() / 5f }
             }
         }
 


smichel17:
title is terrible but the context is in #1724 so you know what I'm going for.

westnordost:
You are quite hands-on. Okay, let's try out how users will accept this feature. I'll merge it once the problem with the "almost 0f" is resolved. :-)

smichel17:
> You are quite hands-on.

:D

I've done enough Android development to be comfortable source diving a bit. If I needed to do larger changes I'd leave it to you, but for a change like this, that doesn't require me to look at many files, it's less than an hour of work… that's reasonable.

> almost 0f

Pushed a commit to work around it, most of the time, just by ignoring very small values. I do not consider this a hack, because at rotation/tilt values that small it would look to the user like the map is flat.

There is still one issue: sometimes the `rotation` gets "stuck" and incorrectly reports a high value (eg, 6.0), even though the map is north-aligned. It does not appear to happen to the tilt value. I am not sure exactly what is causing it, but I can reliably reproduce it by scrolling the map while location is locked on (thereby breaking the lock). Repeatedly tapping the compass will get it back to normal after 4-5 taps.

westnordost:
The rotation and tilt is in radians, if I remember correctly. So 6.28... would again be north I guess.

westnordost:
> I am not sure exactly what is causing it, but I can reliably reproduce it by scrolling the map while location is locked on (thereby breaking the lock)

I can't reproduce it or I do not understand the repro.

smichel17:
Your hint about radians was good. I think it getting stuck at ~6.28 was also due to the "almost 0f" bug — it did not quite make it all the way around to 2pi. I added a check for being close to 2pi, also, and now it seems to work pretty well. Do you want me to squash the commits?

smichel17:
> I can't reproduce it or I do not understand the repro.

It's kind of irrelevant now that it's fixed, but the steps were:

- Lock on
- Switch to compass mode
- Scroll the map away
- Tap the compass button to bring the view back to north
- Try tapping the compass button again ⇒ view does not tilt

smichel17:
Ah! I dove a bit deeper and the animator is in SC, not tangram. I think I'm going to let you fix this one. The fix is super easy but understanding all the related code and figuring out the right spot for the fix will take longer than I want to spend on this.

Here's the deal: Over in `CameraManager.animateCameraUpdate`, you've got [an ObjectAnimator](https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/map/tangram/CameraManager.kt#L116). You add an update listener to push the new coordinates to tangram. The problem is, *ObjectAnimator does not guarantee that the final `onAnimationUpdate` call will have the exact final value you passed to it.* That is, it often over- or under-shoots by just a little bit. So, you need to push the final values to tangram inside the onEnd callback, that currently only unassigns the animator.

smichel17:
Never mind, this was not actually hard to fix.

smichel17:
Okay, rebased so history is nice. I also used kotlin's `run` function to make it less verbose; I can revert that or split it into a separate commit if you don't like that style.

app/src/main/java/de/westnordost/streetcomplete/map/tangram/CameraManager.kt
```diff
@@ -141,7 +141,10 @@ class CameraManager(private val c: MapController, private val contentResolver: C
         animator.setValues(*propValues.toTypedArray())
         animator.duration = duration
         animator.interpolator = interpolator
-        animator.addListener(onEnd = this::unassignAnimator)
+        animator.addListener(onEnd = {
+            animate(it as ValueAnimator)
+            unassignAnimator(it)
+        })
```
westnordost: Actually, if you wonder why this is done seemingly so complicated with the "lastAnimator": This is because there should only be one call to `pushCameraPositionToController` (per animation frame), even if multiple animations of different properties are running at the same time.
Initial implementation was to just push the camera position to the controller in this listener, but this resulted in the animated values overwriting each other. So, the solution was that only one animator will push the position, and that one is the animator that will run the longest.

So, the additional call to `animate` on end of a single object animator should rather go to the "lastAnimator".

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -396,16 +397,24 @@ class MainFragment : Fragment(R.layout.fragment_map_with_controls),
...
+                if (isNorthUp) {
+                    updateCameraPosition(300) {
+                        tilt = if (isFlat) PI.toFloat() / 5f else 0f
+                    }
+                }
             }
-        }
-        if (mapFragment.isFollowingPosition) {
-            setIsCompassMode(!mapFragment.isCompassMode)
         }
```
westnordost: Ah, didn't know `run`. So, that's another option to avoid the pesky "smart cast to X not possible because it is a field and might changed in the meantime":

1. `val x = x ?: return`
2. `x?.let { x -> ... }`
3. `val x = x; if (x != null) { ... }`
4. `x?.run { ... }`

I'd rather not use it, because
- consistency: not used anywhere else in the app
- I (personally) like "flat" functions, i.e. that don't have such a deep `{ { { ... } } }` hierarchy. I find it easier to read. 
- I tend to avoid functions that change what `this` is in a block is. Had some nasty bugs in the past that were not obvious when reading the code. (Shadowing properties of `this` of a containing block)

smichel17: Reverted to the old style. Nicer diff, anyway :P

app/src/main/java/de/westnordost/streetcomplete/map/tangram/CameraManager.kt
smichel17: Hm. So, I tried doing this in the latest.

1. Can you check that what I did is what you intended?
2. Unfortunately, this didn't work.

westnordost: Hm yes, this is what I intended.

So, if that means that every object animator will stop just before reaching the final value (sounds like a bug in the android animation system to me), then what ought to fix it is additionally to how the code is now, `onEnd` of *every* object animator, set the animated property to the final value, but do not push that to the camera controller because the "lastAnimator" will do that.

smichel17: I'm not sure which would be more work, but the way Red Moon handles this is to use a single animator for everything, and if there's an animation currently running when a new one starts (eg, filter is turned off during fade in animation), we just set the values to start there.

If you actually want to be running multiple animators at once, I think there's an AnimatorCoordinator or something like that.

smichel17:
Additional newly-discovered benefit of this: Makes it easier to use 3d with touchscreen gloves, which work fine for taps but aren't great for gestures.

app/src/main/java/de/westnordost/streetcomplete/map/tangram/CameraManager.kt
westnordost: True, but these animations are be started at different points in time as well as have different interpolators and durations.

smichel17: The `lastAnimator` bit is easy enough for me to understand. The part I'm actually having trouble with is figuring out how updated CameraPosition values actually get pushed to `_tangramCamera` from the animators. Obviously it happens in `applyCameraUpdate`, but that only gets called in `updateCamera`, which can be called from several places, none of which seem to be `onAnimationUpdate`. However, understanding this is starting to take too much time, so _I'm going to black box it and just accept that the values are getting where they need to go._ I want to fix the "almost `0f`" bug, not re-architect your tangram wrappers :rofl:

Anyway, back to `pushCameraPositionToController`, the flow of which I do understand. Here's what I'm thinking about:

- Remove all `pushCameraPositionToController` calls from the current animators, leaving them responsible only for updating the camera position.
- Remove the `lastAnimator` concept.
- Add a new [TimeAnimator](https://developer.android.com/reference/android/animation/TimeAnimator) whose only purpose is calling `pushCameraPositionToController`
- Instead of checking `lastAnimatorEndTime < endTime`,  compare the `timeAnimatorEndTime`. If the new end time is later than the TimeAnimator's end time extend the TimeAnimator (ie, keep pushing the camera position, as long as any animations are currently running).

westnordost: > The part I'm actually having trouble with is figuring out how updated CameraPosition values actually get pushed to _tangramCamera from the animators

ObjectAnimators do this automatically. ObjectAnimators directly manipulate the properties of the animated object (the `target`), in this case _tangramCamera. `applyCameraUpdate` is just to apply an update once and push it to tangram right away.

westnordost: > Add a new TimeAnimator that exclusively calls pushCameraPositionToController
> Instead of checking lastAnimatorEndTime < endTime, compare the timeAnimatorEndTime. If the new end time is later than the TimeAnimator's end time, extend the TimeAnimator (ie, keep pushing the camera position, as long as any animations are currently running).

I didn't know about `TimeAnimator`, this sounds like the right thing to use. Using it will not really lead to less code, but probably a bit cleaner code.

> Remove all pushCameraPositionToController calls from the current animators, leaving them responsible only for updating the camera position.

The current animators do not call `pushCameraPositionToController`

smichel17: > ObjectAnimators directly manipulate the properties of the animated object (the target)

d'oh :man_facepalming:. I do vaguely remember reading this now, and wonder why I had Red Moon manually applying the animatedValue instead :man_shrugging: 

> The current animators do not call `pushCameraPositionToController`

They call `animate`, which calls `pushCameraPositionToController` :P

westnordost: > They call animate (or rather, the lastAnimator does), which calls pushCameraPositionToController :P

Only the "lastAnimator" does this. If you change the implementation so that a TimeAnimator will do this, then none of the ObjectAnimators will do this anymore.

smichel17: I think we're on the same page here, just talking past each other.

westnordost: ok

smichel17: Okay, here's what I'm confused about now: `adb logcat -s 'SC:*'` with the code I just pushed.

The correct values are pushed to `_tangramCamera`, but then some time *after* `onEnd`, incorrect values overwrite them. `onEnd` should be happening after the animation is finished, so where are the off-zero values coming from?

smichel17: Out of time this morning but I'm narrowing in on the culprit with 5f27b54.

I'm leaving the commit history intact for the moment so you can checkout and see for yourself (will squash later), but the diff between master and 5f27b54 is *only* adding the `onUpdateListener`.

As far as I can tell, the only difference between setting the values manually and automatically with the animator target is that I'm setting them conditionally. So, maybe running multiple animators with the same target is the cause of the bug?

smichel17: One thing I noticed is that the `_tangramCamera` as a target of the animators is one of the few times that accessing those values is not `synchronized(_tangramCamera)`. I tried adding that into the `onAnimationUpdate` call and it caused crashes, though.

westnordost: I lost track here a little. You add line 144-153 because you want to push the animator end value(s) to the `tangramCamera`. But it doesn't work - it is overwritten.
Let's not waste any more time with this and instead use the fix you did from the very beginning, perhaps with a little comment.

To be honest, l don't believe that the animator system in Android is so crooked that the animator will stop right before reaching the exact end value. But even if it did, if the workaround for this turns out to add even more complexity to it (and the workaround also doesn't work and needs to be work-arounded), then it is not worth it to fix.

From a UX perspective, the user would expect that the "to north" button will treat very small (=barely visible) deviations from North as "it is north already" anyway.

smichel17: > I lost track here a little.

Commits need squashing and logging removed, but the current implementation works.

> You add line 144-153 because you want to push the animator end value(s) to the `tangramCamera`

I was trying to understand what's going on. Does it reach the final value and then overshoot? Does it get overridden by a different animator? etc.

Originally I only had some logging. Then I started adding random stuff (well, educated guesses) to see how it would respond. This one happened to work. I still don't quite understand why.

> l don't believe that the animator system in Android is so crooked that the animator will stop right before reaching the exact end value.

It's possible I'm misremembering, and it's actually the time frame that it overshoots slightly (eg, animate for 320ms even though you said 300).

> From a UX perspective, the user would expect that the "to north" button will treat very small (=barely visible) deviations from North as "it is north already" anyway.

Seems reasonable, regardless of whether the bug is fixed. I was using `0.02f`. That caught ~90% of cases, but sometimes the "almost `0f` bug still hit. Should I bump it up to `0.03f`?

> it is not worth it to fix

The current CameraManager code seems to work (end values are correct). It's +8 lines that don't seem like they should be necessary. Should I keep it or leave it?

westnordost: > The current CameraManager code seems to work (end values are correct). It's +8 lines that don't seem like they should be necessary. Should I keep it or leave it?

Rather if you removed them again. It is a workaround that "happens to work" for a bug perhaps exists (in Android). Not working around this possible bug has no real downsides, so less code here is better.

> 0.03f

I'd say whatever feels natural as a user. 0.03f would be around 5°, might be a bit too much.

smichel17: I went with `val margin = 0.025f`. While I am able to start at true north, rotate slightly, and then tap the compass and have it tilt instead of moving back to true north, I could not do it reliably (maybe once every 5 or 10 tries), and this was with me *trying* to do that, so in actual operation I think it should happen basically never. Also pulling it out into a val makes it really easy to change in the future.

smichel17:
Rebased; I think this is ready to merge.

westnordost:
Looks great! Sorry for the long forth and back for such a small change

Issue #1557, opened 2019-09-11
Documentation: do I need an account to contribute?


Mistress-Anna:
I've filled a few blanks on a map in the streetcomplete app, but it's not clear in the app whether I need an account on OSM to have my contributions effective or if it's immediate. Maybe this should be part of the FAQ?

matkoniecz:
Just to explain: app will ask to login/register after less than 10 edits.

I will add it to the FAQ (assuming that no one will do it earlier).

rugk:
> after less than 10 edits.

Maybe also consider decreasing this limit? 3 or 5 should be enough. (given that this question appears)

Or, showing a text below/above the authorize/login button saying:
> **Note:** You did not yet login to OpenStreetMap, your contributions will not be submitted.

Because I have the impression that the OP @Mistress-Anna didsaw this button in the settings, but did not know whether logging in is needed or not. Is that correct, @Mistress-Anna?

matkoniecz:
To document my assumptions:

There are many different users that may use StreetComplete

- new to both SC and OSM - they will not be expecting login/registration.
Main benefit of delayed login is that they will be less overwhelmed by initial contact and less likely to escape. Main drawbacks are that that may fail betrayed by late login call and that initial contributions may be lost.

Additional testing may be useful but I think that for this group of users current SC behavior is ok and earlier showing registration form may discourage them (but still, as long as it is not shown on first or second solved quest it should be ok).

- new to SC - they are far more likely to be surprised by editing without login. Some, especially ones just exploring SC without editing (or making just one or two test edits) are most likely to be surprised by ability to apparently edit without login or by edits not appearing in the database. I would expect that this group is also likely to actually read FAQ or read more carefully settings menu.

westnordost:
My experience is that users are incredibly good in clicking away dialogs ("later") that prompt them to do something (me included): Prompt to rate an app on Google Play, reading license agreements, offers to download an app instead of using the mobile website, cookie banners, GDPR banners, ...

I think users won't have a problem of clicking away the authorization prompt if they don't feel like registering now (or start registering and find out that they do not want to do it right now after all). Simply important is to tell the user that none of his answers will be published as long as he doesn't login and this is the information the dialog transports.

rugk:
Okay, so if we do not want a dialog the user can click away easily, maybe better use a **toast message** (only after solving the 1st quest, or after 1-10 quests, and afterwards then nag the user with prompts, as it is done currently)?
[...]

PR #1742, opened 2020-02-21
Disable auto-sync preference when not logged in

smichel17:
Also moved auto-sync preference below authorization preference, to reflect the hierarchy.

When not logged in, it's not possible to upload quest answers at all, so the auto-sync setting should have no effect. Thus, we should disable it in those situations.

If you don't agree, feel free to discuss and/or close this. I only opened it as a PR because it was barely more effort than opening an issue.

From what I read in #1557, it looks like the auto-sync preference *may* impact how often we bug users to log in (I have not looked at the code to confirm whether this is currently the case). I'd argue that auto-upload *should* not have any effect on how often we prompt for login. If it does, I'll add a follow up commit to address that (but that's more work, so I want to make sure you'd be interested in merging, first).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -94,11 +94,11 @@ class SettingsFragment : PreferenceFragmentCompat(),
 
     override fun onStart() {
         super.onStart()
-        updateOsmAuthSummary()
+        updateOsmAuthDependentPrefs()
         activity?.setTitle(R.string.action_settings)
     }
 
-    private fun updateOsmAuthSummary() {
+    private fun updateOsmAuthDependentPrefs() {
         val oauth = preferenceScreen?.findPreference<Preference>("oauth")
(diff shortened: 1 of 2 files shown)


HolgerJeromin:
As a first time user of an app I often check the prefs.
After three quests the App would be logged in and a pref would be active which I have not set to my need.

matkoniecz:
> When not logged in, it's not possible to upload quest answers at all, so the auto-sync setting should have no effect. Thus, we should disable it in those situations.

Are you sure? I thought that autosync is also about scanning for new quests.

smichel17:
> Are you sure? I thought that autosync is also about scanning for new quests.

Not sure at all! And in that case this PR is baloney :P

I'll double check shortly.

smichel17:
Okay, it also applies to for quests. I think I probably got that impression because the popup when you choose a setting other than automatic only talks about uploading. So I think we could make it clearer by either

- changing the preference title to something like "Auto-scan and upload"  
or
- Changing the dialog to also mention manually scanning for new quests. ("There is a menu item to manually scan for new quests."). A picture would not be necessary here since they need to go through that menu to get to the settings in the first place.

Issue #1573, opened 2019-09-19
[New quest] What payment methods are accepted here?


Atrate:
### Affected tag(s) to be modified/added:
 payment:method=yes/no/interval

### Question asked:
What payment methods are accepted here?

### Example keys:

payment:coins=yes/no/interval

payment:cash=yes/no/interval

payment:notes=yes/no/interval

payment:visa=yes/no/interval

payment:mastercard=yes/no/interval

And the explicit

payment:debit_cards=no

payment:credit_cards=no

### Checklist for quest suggestions (see guidelines):

- [x]  🚧 To be added tag is established and has a useful purpose
- [x]  🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x]  🐿️ Easily answerable by everyone from the outside but a survey is necessary
- [x]  💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x]  🕓 Applies to a reasonable number of elements (Worth the effort)

ferdinand0101:
So where would you want to run this ?
Any shop and restaurant ?

matkoniecz:
Is there any current user of this data?

Are there places where answer would not be basically the same for every shop? For example in Poland basically any shop accepts the same payments, with very rare case of cash-only shops and very rare cases of fully automated payment locations not accepting cash.

ferdinand0101:
Well in a lot of restaurants especially imbisses it's a 50 50 thing if they'll accept cards
So I can definitely see some use here

rugk:
In any case, this would be an indoor-only quest, i.e. disabled by default.

MegaArthur:
Yes, it would be disabled by default. As matkoniecz has mentioned, most shops (in Poland) accept the same methods of payment, but it can differ substantially for amenities like restaurants or beauty shops etc.

westnordost:
That's a lot of data to add, I am not sure about the usefulness of this.

ferdinand0101:
I believe to make it more useful only EC and credit cards are relevant
I know of no store which doesn't accept cash

matkoniecz:
>  I know of no store which doesn't accept cash

Happens in some regions, Sweden is quite famous for that.

HolgerJeromin:
> I believe to make it more useful only EC and credit cards are relevant

[...]

PR #1743, opened 2020-02-23
Add quest: accepts cash yes/no

quite:
Discussed in issue #1573


quite:
Really WIP!

- filter is rough, or things are just tough? it catches:
  - this statue https://www.openstreetmap.org/node/6063606630
  - this way https://www.openstreetmap.org/way/161771522
  - this way, a named quarter/residential area: https://www.openstreetmap.org/way/193174595
- needs proper icon. unicode money bag? but that usually has dollar sign on it... currency sign ¤ is not well known right
- ...

westnordost:
> filter is rough, or things are just tough? it catches:
> this statue https://www.openstreetmap.org/node/6063606630
> this way https://www.openstreetmap.org/way/161771522
> this way, a named quarter/residential area: https://www.openstreetmap.org/way/193174595

Can't be!

quite:
> Though, please think for each place, is it really probable that this place may not accept cash?

You have not been in Sweden I hear :-)

> Can't be!

But it seems to be... I tried playing around with overpass turbo quickly, and don't have that problem. Maybe I constructed the filters wrongly...

app/src/main/res/values-sv/strings.xml
```diff
@@ -765,4 +765,5 @@ Dela den nu?"</string>
     <string name="about_title_donate">"Donera"</string>
     <string name="about_summary_donate">"Visa din uppskattning! ❤️"</string>
     <string name="about_description_donate">"Tack för att du överväger att stöda den här appen! För att se aktuella donationer och donera själv, tryck på respektive plattform:"</string>
-</resources>
\ No newline at end of file
+    <string name="quest_accepts_cash_title">"Tar %s emot kontanter?"</string>
```
westnordost: This will be overwritten by pulling the translations from POEditor

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.NoCountriesExcept
+import de.westnordost.streetcomplete.data.osm.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.download.OverpassMapDataAndGeometryDao
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+class AddAcceptsCash(o: OverpassMapDataAndGeometryDao) : SimpleOverpassQuestType<Boolean>(o) {
+//nodes, ways with amenity = toilets and access !~ private|customers and !fee
```
westnordost:
```suggestion

```

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
```diff
@@ -0,0 +1,55 @@
...
+        nodes, ways, relations with
+        (
+          (shop and shop !~ no|vacant|mall)
+          or amenity ~ ${amenity.joinToString("|")}
+          or leisure ~ ${leisure.joinToString("|")}
+          or craft ~ ${craft.joinToString("|")}
+          or tourism ~ ${tourism.joinToString("|")}
+        )
+        and name and !payment:cash and !payment:coins and !payment:notes
+    """
```
westnordost: I think it would be better readable if you put the values directly into the tag filter string like you did for the shops. But this is not blocking a merge

app/src/main/res/values-sv/strings.xml
quite: Removed.

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
quite: Removed.

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
quite: I avoided long lines with foo|bar|baz|… I'll leave it like so.

westnordost:
Will merge for next major release because it contains new strings

quite:
Any plan for when this release will happen?  When can translation of the strings be done?

westnordost:
No plans yet, usually its every 2 weeks up to 1 month after the last bugfix release.

Translations are done in poeditor

quite:
> Translations are done in poeditor 

Ah, and you'd inform when there are new strings to translate in advance of
that release then, I s'pose :)

matkoniecz:
> when there are new strings to translate in advance of that release then, I s'pose :)

The best way to do that is to select watch option that includes at least releases (top of the page) and once new beta appears - look at poeditor. Sadly, poeditor has no option to get notifications about missing strings.

peternewman:
> The best way to do that is to select watch option that includes at least releases (top of the page) and once new beta appears - look at poeditor. Sadly, poeditor has no option to get notifications about missing strings.

Is there not a catch 22 here? Don't you ideally want to push the new strings up to poeditor now, so there's a chance they'll be translated in time for the release (and ideally the beta too)?

matkoniecz:
I am not doing this part, so it is a guess but:

- sending strings to poeditor may require a manual action by @westnordost 
- it avoids risk of translators translating something and then strings changing again (some of my English in my PRs was Engrish and was fixed before the release)

peternewman:
> I am not doing this part, so it is a guess but:
> 
>     * sending strings to poeditor may require a manual action by @westnordost
> 
>     * it avoids risk of translators translating something and then strings changing again (some of my English in my PRs was Engrish and was fixed before the release)

Hi @matkoniecz ,

Yeah it wasn't really aimed at you particularly, but it seemed the most appropriate comment to reply to.

@westnordost may feel it's too much effort, but you could for example have a develop branch, which splits from master and has commits like this in, from the fact it's been approved and @westnordost 's comments, it sounds like it's ready to go in, but is just awaiting a future release given the new strings. If poeditor's strings always tracked develop, they would be ready for future releases. Although this does assume that none of the system gets upset with translations for phrases that aren't used in the current codebase.

quite:
In any case I've translated some new strings to Swedish now (including the one from this quest). It seems @westnordost pushed them to poeditor before making any beta release.

westnordost:
Correct. Usually 0-2 days before a beta release, I push the new strings to POEditor, then during the beta phase (usually >=2 days), the strings can be translated before a full release

Issue #1692, opened 2020-01-03
roof shape quest with "recent used order" confusing


HolgerJeromin:
I am pretty sure in the past the roof shape quest had constant sort order. 
In one of the latest update it changed to "recent used order". 
The old, stable sorting was better IMO since I knew the first one was `gabled` and could select it without double checking.

**Versions affected**
Lastest play store beta

westnordost:
The stable order was a bug, this now is the intended behaviour. It bug used to occur for *all* quests.
What is your reasoning for why for this quest, recently used roof shapes abound not be moved to the front?

Am 3. Januar 2020 08:08:25 MEZ schrieb Holger Jeromin <notifications@github.com>:
>I am pretty sure in the past the roof shape quest had constant sort
>order. 
>In one of the latest update it changed to "recent used order". 
>The old, stable sorting was better IMO since I knew the first one was
>`gabled` and could select it without double checking.
>
>**Versions affected**
>Lastest play store beta

HolgerJeromin:
The mru order is faster if the selecting needs time. For example surface quest has a very long list with two layer menu (paved, unpaved...) or building levels have select, enter number, select, enter. 
But with this quest all (?) possible answers are shown at once. 
I knew: first is gabled. When I needed another one it was found quickly. If the next roof is the same, i remember the position (third from the first row) an am very fast.

Now this knowledge is not usable, as I have to scan many icons every time.

pkoby:
I'd also say that in the surface quest, it's very quick to see the difference between different options. The roof shapes require a closer look to compare the icons (consider hipped vs. half-hipped, for instance). I haven't tried this new version, so I can't speak to how it works in practice, but I do agree that knowing where the icons were was pretty useful.

matkoniecz:
>  But with this quest all (?) possible answers are shown at once.

I think that `tracktype` one is in a similar situation. I use it very rarely in my region, but I remember being confused by seemingly random order of `grade2`, `grade1`, `grade5` and so on.

westnordost:
Okay, makes sense that the tracktype should not be mixed up. Though for the roof shapes, if you are in an area where half-hipped or hipped or whatever is really common, you would want to have them appear in the front. If the icon is not well distinguishable from others on first sight, then rather the icons should be improved. (Though, in my opinion they are already as clear as they can get. Any suggestions about making them clearer?)

Let's collect all the quests where the items should not be mixed (plus the reasons for that).

[...]

PR #1754, opened 2020-03-03
remove "show more" button from image selector form (#1692)

westnordost:
see https://github.com/westnordost/StreetComplete/issues/1692#issuecomment-585310314

The original reason to introduce this button was when the UI looked like this:

image

The answer button bar always floated at the bottom of the screen. With the current UI, only the OK button always floats at the bottom while the other answers are at the end of the form. The problem with the old UI was that there was no clear indicator if the currently visible choices where the only choices or if there are more. To indicate that there is more, there was a slight fade-to-white at the bottom as can be seen here.

image

But the more clear indicator that there is more was simply a button that is visible right when opening the form that says "show more".

The current UI does not have this problem because the answer button bar ("Can't say",...) is always at the bottom of of all choices and the end of the form is indicated by rounded corners at the bottom of the bubble dialog.

This show more button is used in situations where answer options were included that - looking at taginfo - were very unlikely to be chosen but were included for completeness sake. To not make the form unnecessarily long for 99% of answers and make users able to give a quick answer without scanning through the whole list, those unlikely answers are cut off.

The following quests have these cut-offs currently:
- **roof quest**: Roof shapes that make up less than 0.1% of tagged roof shapes are cut-off
- for the others, I don't have the numbers, but I think they are somewhat along these lines as well

Having read why they were introduced, what changed in the meantime and what purpose they (still) serve, do you think this PR should be merged, @smichel17 , @matkoniecz ?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AImageListQuestAnswerFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AImageListQuestAnswerFragment.kt
@@ -20,7 +20,7 @@ import kotlinx.android.synthetic.main.quest_generic_list.*
  * I is the type of each item in the image list (a simple model object). In MVC, this would
  * be the view model.
  *
- * T is the type of the answer object (also a simple model object) created by the quest 
+ * T is the type of the answer object (also a simple model object) created by the quest
  * form and consumed by the quest type. In MVC, this would be the model.
  */
 abstract class AImageListQuestAnswerFragment<I,T> : AbstractQuestFormAnswerFragment<T>() {
@@ -34,8 +34,6 @@ abstract class AImageListQuestAnswerFragment<I,T> : AbstractQuestFormAnswerFragm
     protected open val itemsPerRow = 4
     /** return -1 for any number. Default: 1  */
     protected open val maxSelectableItems = 1
-    /** return -1 for showing all items at once. Default: -1  */
-    protected open val maxNumberOfInitiallyShownItems = -1
     /** return true to move last picked items to the front. On by default. Only respected if the
      *  items do not all fit into one line */
     protected open val moveFavoritesToFront = true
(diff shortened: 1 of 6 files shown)


matkoniecz:
Two sides: 

other roof values may be tagged rarely, but often roof is not fitting perfectly (for example deviations from idealized rood shape due to windows on side of the building), so it is likely that someone will relatively often look at all possible shapes far more often than use them.

Similarly, for multi-use pitches I sometimes look through all sports to be sure that have not missed something (nearly always pointlessly anyway)

Current "see more" fadeout is not perfectly clear and it probably can be missed by some.

Overall I have no strong preference, but I think that I would support removal of "show more", mostly to reduce interface complexity - replace drag & click & drag by just dragging.

But I have no strong opinion here and I think that I would be fine with both versions.

westnordost:
> Current "see more" fadeout is not perfectly clear and it probably can be missed by some.

This was the old UI.

matkoniecz:
> This was the old UI.

Sorry, I somehow completely missed new one. Despite mention in the issue. I guess that my "argument" for keeping button is fundamentally faulty.

smichel17:
I played around with it a little. The only changed quest that I was able to easily find was a roof type quest. For it, I stand by my initial opinion think this way is strictly superior.

matkoniecz:
> The only modified form

At least sport quest (for pitches without sport) shouid be also affected. Maybe also religion quest.

smichel17:
Yes, and fire hydrant & produce. But I'm not sure of coordinates where I could find those quests :)

matkoniecz:
Are you making app for testing with Android Studio?

Go to settings, scroll down to "debug" section - unlocked on manual builds (default when building with AS) and see quest form without hunting for location.

Or use overpass turbo to locate matching elements using wizard query builder ( http://overpass-turbo.eu/s/RgX )

smichel17:
> Are you making app for testing with Android Studio?

I am building it myself. I seldom use Android Studio, though (I mostly use the command line tools).

> Go to settings, scroll down to "debug" section - unlocked on manual builds (default when building with AS) and see quest form without hunting for location.

D'oh. I knew about this, it just completely slipped my mind. …They all look good to me.

(Aside: I'm still planning on working on [# 1735](https://github.com/westnordost/StreetComplete/issues/1735), just had something else come up that weekend and haven't gotten back to it yet. Maybe this weekend, maybe next.)

Issue #1756, opened 2020-03-04
[New Quest] Defibrillator location (access/opening hours)


Sequynth:
### General
Even though a lot of defibrillators have been deployed recently in public spaces, usage remains [fairly low](https://www.sueddeutsche.de/gesundheit/gesundheit-defibrillatoren-in-deutschland-werden-selten-genutzt-dpa.urn-newsml-dpa-com-20090101-190426-99-971052) (situation in germany, german). One reason for such few uses is that people dont know where they are located or are even unaware of their existence. This quest could serve two purposes: First, it further specifies the location of a defibrillator, potentially saving time spent on looking for a specific device. Second, it actually raises awareness among SC users that such devices are available in many cities these days and some of those are already mapped.
Providing also access= and opening_hours= information, this could prevent people from going for a nearby but inaccessible defibrillator, while a second one further away might be readily available.

Affected tag(s) to be modified/added: [defibrillator:location=](https://wiki.openstreetmap.org/wiki/DE:Tag:emergency%3Ddefibrillator)
Question asked: **Where exactly is this defbrillator?**

Two separate quests could be
(access=yes/private **Is this a public defibrillator?**
opening_hours= **When is this defibrillator available?**)
access and opening_hours could potentially be inferred from the values of the building they are located in, but im not sure how this is generally handled/mapped in OSM.

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
The tag was approved back in [2013](https://wiki.openstreetmap.org/wiki/Proposed_features/automated_external_defibrillator) and there are currently [~25k](https://taginfo.openstreetmap.org/tags/emergency=defibrillator) nodes worldwide, of which [~5k](https://taginfo.openstreetmap.org/keys/defibrillator:location#overview) have a value for defibrillator:location, leaving ~20k nodes without further information about the location.
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
Once a location= tag is provided, the quest does not appear to other users (do I understand this check-element correctly?)
- [ ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary
Many defibrillators are located in public buildings, i.e. entering is possible, but survey from outside is usually not sufficient. Some devices might be on private property, here a second quest could ask for access=
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
[...]

PR #1758, opened 2020-03-06
icons for defibrillator quests added

Sequynth:
#1756 
Except for the heart, all shapes were copied from existing ones. No new color was introduced.


westnordost:
I'd probably modify them a bit to look more like the actual defibrillator sign, but that I can do when the quest is actually implemented.

PR #1782, opened 2020-04-17
WIP: AddAddressStreet

dbdean:
Thanks to @matkoniecz for his great work on starting this. I've gotten it up to date with StreetComplete master and had a go at moving the implementation to AAddLocalizedNameForm.

Once this is completed it should resolve https://github.com/westnordost/StreetComplete/issues/213.

I've also
* Added the street-number to the prompt to make it easier to tell which house is being referred to (and often as an additional clue to the street it belongs to)
* Upped the distance for close streets to 250 metres as houses in the middle of long, straight streets were getting no suggestions
* Made it remember the last street you entered, for easier surveying along streets.

There's still plenty to do though, so this is just a WIP at the moment. I'm very open to suggestions, and it's very possible that some of this is a bit hacked together, as I'm not super familiar with this codebase - I'm happy to change anything!


westnordost:
Thank you for continuing this! I will do a rough review of the code now.

app/src/main/res/values/strings.xml
```diff
@@ -696,6 +696,11 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generalFee_title">Does it cost a fee to enter %s?</string>
     <string name="quest_laundrySelfService_title">Is this laundry a self service laundry?</string>
     <string name="quest_handrail_title">Do these steps have a handrail?</string>
+    <string name="quest_address_street_title">What street is this number %s of?</string>
+    <string name="quest_address_street_no_named_streets">This address has no named street</string>
+    <string name="quest_address_street_noStreet_confirmation_title">Are you sure that named street is not a part of this address?</string>
+    <string name="quest_address_street_noStreet_confirmation_positive">Without a street part</string>
+    <string name="quest_address_street_place_name_label">Place forming part of the address:</string>
```
westnordost, line 703: The wording is a bit awry. Maybe "To what street does the house number X belong"? Etc..

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
```diff
@@ -19,7 +19,7 @@ import de.westnordost.streetcomplete.util.isInMultipolygon
 
 class AddHousenumber(private val overpass: OverpassMapDataAndGeometryDao) : OsmElementQuestType<HousenumberAnswer> {
 
-    override val commitMessage = "Add housenumbers"
+    override val commitMessage = "Add address"
```
westnordost, line 22:
```suggestion
    override val commitMessage = "Add housenumbers"
```

I think it should keep describing exactly what had been added.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,108 @@
...
+        val housenumber = tags["addr:housenumber"]
+        return if (housenumber != null) arrayOf(housenumber) else arrayOf()
+    }
+
+    override fun createForm() = AddAddressStreetForm()
+
+    override fun isApplicableTo(element: Element): Boolean? = null
+
+    override fun download(bbox: BoundingBox, handler: (element: Element, geometry: ElementGeometry?) -> Unit): Boolean {
+        Log.wtf("aaa", getOverpassQuery(bbox))
```
westnordost, line 37: stray wtf log

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,108 @@
...
+                    getQuestPrintStatement()
+
+    // TODO: probably either this should be modified to match AddRoadName with custom set of roads  with config array migrated to OsmTaggings
+    // TODO: or AddRoadName can use something like that
+    /** return overpass query string to get roads with names
+     * */
+    private fun getStreetNameSuggestionsOverpassQuery(bbox: BoundingBox) =
+            bbox.toGlobalOverpassBBox() + "\n" + """
+        way[highway ~ "^(${OsmTaggings.ALL_ROADS.joinToString("|")})$"][name];
+        out body geom;""".trimIndent()
```
westnordost, line 62: `AddRoadName` only needs the names of roads that are near unnamed roads.
`AddAddressStreet` needs the names of roads that are near housenumbers with missing `address:street`.

The suggestions are put into the same table, but it makes sense if the queries query a different set of roads to only put into the road suggestions db what is really necessary. (Because currently that db table is *never* cleared or cleaned up.)

Anyway, actions:
1. Compare `AddRoadName.ROADS` and `OsmTaggings.ALL_ROADS` and use `OsmTaggings.ALL_ROADS` also in `AddRoadName` in place of `ROADS` if it is the same
2. kind of optional, but it would be better: analogous to `AddRoadName`, `AddAddressStreet` should only query the names of roads that are near housenumber nodes that are missing a road name.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,108 @@
...
+    val namePattern = Pattern.compile("name(:(.*))?")
+    for ((key, value) in this) {
+        val m = namePattern.matcher(key)
+        if (m.matches()) {
+            val languageCode = m.group(2) ?: ""
+            result[languageCode] = value
+        }
+    }
+    return if (result.isEmpty()) null else result
+}
```
westnordost, line 108: Copypasta from `AddRoadName`.
Suggestion: Create an extension function in a new file `app/src/main/java/de/westnordost/streetcomplete/quests/address/RoadNameSuggestionsDao.kt` called
`fun RoadNameSuggestionsDao.putOsmWayAsRoad(way: Way, wayGeometry: ElementPolylinesGeometry)`
or similar which is used by both classes.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,108 @@
...
+            bbox.toGlobalOverpassBBox() + "\n" + """
+        way[highway ~ "^(${OsmTaggings.ALL_ROADS.joinToString("|")})$"][name];
+        out body geom;""".trimIndent()
+
+    override fun applyAnswerTo(answer: AddressStreetAnswer, changes: StringMapChangesBuilder) {
+        var key = ""
+        when(answer){
+            is StreetName -> {key = "addr:street"}
+            is PlaceName -> {key = "addr:place"}
+        }
```
westnordost, line 69: You can shorten that to

```suggestion
        val key = when(answer) {
            is StreetName -> "addr:street"
            is PlaceName -> "addr:place"
        }
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,108 @@
...
+        //TODO: can we actually get addr:street/place:langcode here, and if so,
+        // should we polute OSM with this?
```
westnordost, line 71: In my opinion, if we are in a country where street names in several languages are common and thus it is common to put several `name:XX` on the street, then it is reasonable to also put this on the address elements. Usually, it is not more than two.

For example: https://www.openstreetmap.org/way/439628149

app/src/main/java/de/westnordost/streetcomplete/util/KryoSerializer.java
```diff
@@ -62,7 +63,8 @@
 			WeekdaysTimesRow.class,
 			OsmLatLon.class,
 			SplitAtPoint.class,
-			SplitAtLinePosition.class
+			SplitAtLinePosition.class,
+			Name.class,
```
westnordost, line 67: What is this?

app/src/main/res/layout/quest_localizedname_row.xml
```diff
@@ -23,7 +23,7 @@
         android:layout_height="wrap_content"
         android:minHeight="48dp"
         android:layout_toEndOf="@id/languageButton"
-        android:layout_toStartOf="@+id/deleteButton"
+        android:layout_toStartOf="@+id/nameSuggestionsButton"
```
westnordost, line 26: Isn't this XML used by several quests? Doesn't just renaming this button break the form for the other quests that use it?

dbdean:
Thanks for the comments, @westnordost. I'll find some time later and go through them as best as I can.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
dbdean: Have removed.

app/src/main/java/de/westnordost/streetcomplete/util/KryoSerializer.java
dbdean: I think this is left-over from when the AddAddressStreetForm had it's own serializer, before I moved it back to using the AAddLocalizedName Form. I've removed if from here now.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
dbdean: Fair point. I've removed the TODO comment, and left the code as-is.

app/src/main/res/values/strings.xml
dbdean: I've had a go at improving them. Some of these may need to change as I attempt to get the place names working better too.

app/src/main/res/layout/quest_localizedname_row.xml
dbdean: There's no button-renaming going on here. I'm just changing the RHS end of this layout element to be the nameSuggestionsButton instead of the deleteButton. This allows the drop down to show without occlusion at the same time as text in the field. I have tested this with other quests (like NameRoad) and with and without text in this quest, and it still works correctly. It even gives the text more room when the nameSuggestionsButton is not visible, which I wasn't really expecting, but is nice to see.

dbdean:
> For the adapter, I wonder why it is necessary - why couldn't `AddLocalizedNameAdapter` be reused?

You are correct, it is no longer required with my refactoring, and I have removed it.

dbdean:
@westnordost, I've noticed at the moment that it doesn't immediately show this quest when I put a streetnumber on a building. Should it be able to? If so, what changes to I need to make to make it happen?

westnordost:
`isApplicableTo` would need to return true.
`true `= it is now applicable to the element
`false `= it is not applicable to the element (anymore)
`null `= I don't know! Must download the quest via overpass to find out!

Have a look at `AddRoadName`.

app/src/main/res/layout/quest_localizedname_row.xml
westnordost: Ok!

westnordost:
You will tell me when I should have another look at this PR right?

dbdean:
@westnordost, sure. I'll ping you when I think it's ready for another look over.

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
```diff
@@ -113,15 +114,19 @@ class AddRoadName(
...
+                "way[highway ~ \"^($NAMEABLE_ROADS)$\"][!name][!ref][noname != yes][!junction][area != yes]"
         // this must be the same as above but in tag filter expression syntax
         private val ROADS_WITHOUT_NAMES_TFE by lazy { FiltersParser().parse(
-            "ways with highway ~ $ROADS and !name and !ref and noname != yes and !junction and area != yes"
+                "ways with highway ~ $NAMEABLE_ROADS and !name and !ref and noname != yes and !junction and area != yes"
         )}
+
+        // but we can find name suggestions on any type of already-named road
+        private val ROADS_WITH_NAMES =
+                "way[highway ~ \"^(${ALL_ROADS.joinToString("|")})$\"][name]"
```
dbdean, line 129: I've updated this so that there is a difference between roads-that-we-want-to-ask-names for (only `primary|secondary|tertiary|unclassified|residential|living_street|pedestrian`), and roads that could have names and we are looking at for name suggestions (here and in AddAddressStreet), which is all names in `ALL_ROADS`. So we won't ask for a name for a motorway, for example, but if there is a motorway around with a name it will appear as a suggestion.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    /** return overpass query string to get roads with names around addresses without streets
+     * */
+    private fun getStreetNameSuggestionsOverpassQuery(bbox: BoundingBox) =
+            bbox.toGlobalOverpassBBox() + "\n" + """
+        $ADDRESSES_WITHOUT_STREETS -> .address_missing_street;
+        $ROADS_WITH_NAMES -> .named_roads;
+        way.named_roads(
+            around.address_missing_street: $MAX_DIST_FOR_ROAD_NAME_SUGGESTION);
+        out body geom;""".trimIndent()
```
dbdean, line 64: name suggestions now only downloads streets that are around addresses without streets.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
dbdean: Ok, I've made changes to consolidate the way that name suggestions work here and in AddRoadName to meet (1). I've also changes the query to only look at roads around addresses without streets to meet (2). I've added some comments below to point out where I've done this.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
dbdean: Done in https://github.com/westnordost/StreetComplete/pull/1782/commits/5591b1a4e4c4a91d3591fdefcb8b74543899ef84

dbdean:
I think I've fixed all your comments, @westnordost, but I still have some work to do on getting addr:place working better. I'm happy if you want to hold off reviewing while I work on that - or you can have another go over now - your choice.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    companion object {
+        //TODO: this only appears to pick up nodes, so street-name suggestions don't appear
+        // in the middle of a long straight-stretch. There might need to be a better way.
+        const val MAX_DIST_FOR_ROAD_NAME_SUGGESTION = 250.0
+
+        private const val ADDRESSES_WITHOUT_STREETS =
+                """nwr["addr:housenumber"][!"addr:street"][!"addr:place"]"""
```
westnordost, line 86:
```suggestion
        private const val ADDRESSES_WITHOUT_STREETS = """
                (
                    nwr["addr:housenumber"][!"addr:street"][!"addr:place"][!"addr:block_number"];
                    nwr["addr:streetnumber"][!"addr:street"];
                );""".trimIndent()
```

In Japan, few streets have names. Addresses are recorded with housenumber+blocknumber.

For countries where there are conscription numbers, also look for `addr:streetnumber`.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,95 @@
...
+        // this must be the same as above but in tag filter expression syntax
+        private val ADDRESSES_WITHOUT_STREETS_TFE by lazy { FiltersParser().parse(
+                "nodes, ways, relations with addr:housenumber and !addr:street and !addr:place"
+        )}
```
westnordost, line 90:
```suggestion
        private val ADDRESSES_WITHOUT_STREETS_TFE by lazy { FiltersParser().parse("""
                nodes, ways, relations with 
                  addr:housenumber and !addr:street and !addr:place and !addr:block_number
                  or addr:streetnumber and !addr:street
        """)}
```
see above comment

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/data/RoadNameSuggestionsDao.kt
```diff
@@ -74,4 +78,27 @@ class RoadNameSuggestionsDao @Inject constructor(
...
+        for ((key, value) in this) {
+            val m = namePattern.matcher(key)
+            if (m.matches()) {
+                val languageCode = m.group(2) ?: ""
+                result[languageCode] = value
+            }
+        }
+        return if (result.isEmpty()) null else result
+    }
+
```
westnordost, line 103: Could this be extension functions in a kotlin file in the same directory as the address quests instead? No need to blow up the interface.

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadName.kt
westnordost: I think this makes sense 👍

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+import de.westnordost.streetcomplete.util.TextChangedWatcher
+import java.util.*
+import javax.inject.Inject
+
+class AddAddressStreetForm : AAddLocalizedNameForm<AddressStreetAnswer>() {
+    private var textField: EditText? = null
+    private var isPlaceName = false
+    private var defaultName = ""
+
+    override fun onClickOk(names: List<LocalizedName>) {
```
westnordost, line 31: functions should go below all variable definitions and constructor

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+            showNameSuggestionsMenu(v, nameSuggestionsList) { selected -> ;
+                textField!!.setText(selected)
+            }
+        }
+        textField!!.addTextChangedListener(TextChangedWatcher { checkIsFormComplete() })
+    }
+
+    companion object {
+        private const val NAMES_DATA = "names_data"
+    }
```
westnordost, line 156: the companion object should be the very last thing in a class definition

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+        // if the form is complete, we will reject close, unless it is still has only the default name
+        if (isFormComplete()) {
+            if (adapter.localizedNames.size == 1 && adapter.localizedNames.first().name == defaultName) {
+                return false
+            } else {
+                return true
+            }
+        } else {
+            return false
+        }
```
westnordost, line 117: You could shorten it to

```suggestion
        return isFormComplete() && !(adapter.localizedNames.size == 1 && adapter.localizedNames.first().name == defaultName
```

westnordost: Also, not sure if the wording of that function is off, but what if the form is not complete but also not empty, should it not also reject closing?

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private fun switchToPlaceName() {
+        AlertDialog.Builder(activity!!)
+                .setTitle(R.string.quest_address_street_noStreet_confirmation_title)
+                .setPositiveButton(R.string.quest_address_street_noStreet_confirmation_positive) {_, _ -> switchToPlaceNameLayout() } //
+                .setNegativeButton(R.string.quest_generic_confirmation_no, null)
+                .show()
+    }
```
westnordost, line 126: Confirmation dialogs are only necessary if okay'ing it would lead to the quest being answered. In the housenumber quest, there is no confirmation for "no house number but a name", because it just switches the layout.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    private fun switchToPlaceNameLayout() {
+        isPlaceName = true
+        setLayout(R.layout.quest_streetname_place)
+    }
```
westnordost, line 131: You should persist this layout change. See `AddHousenumberForm.isHousename` for how it is done

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    private fun setLayout(layoutResourceId: Int) {
+        val view = setContentView(layoutResourceId)
+        val buttonNameSuggestions : View = view.findViewById(R.id.nameSuggestionsButton)
+        textField = view.findViewById(R.id.name)
+
+        //TODO - use actual place names
```
westnordost, line 138: suggestions for place names could always be added later - as long as the form does not behave wrongly (i.e. show road names as suggestions for place names), that's fine.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
```diff
@@ -0,0 +1,179 @@
...
+            val locale = if(languageCode.isEmpty()) countryInfo.locale else Locale(languageCode)
+            val abbr = abbreviationsByLocale.get(locale)
+            val containsAbbreviations = abbr?.containsAbbreviations(name) == true
+
+            if (name.contains(".") || containsAbbreviations) {
+                possibleAbbreviations.add(name)
+            }
+        }
+
+        // TODO, not sure if I should be saving all the names (is there every going to be more than one?)
```
westnordost, line 43: Saving the "main" name (`addr:street`) is enough, because the localizations if any can be added back via the road name suggestions database.

westnordost:
Additionally, I originally said in #213 that the quest should only be asked for housenumbers at the beginning of the road (house numbers 1, 2, 3, 4, ...) to avoid spamming the user with quests where he answers with always the same answer and because it is more efficient to enter all the `addr:street` for a street in a full editor like JOSM via multiselect. Only for places where it is not clear to which road a building belongs, it is necessary to survey it. These would be the beginning of roads and maybe intersections. What do you think of this?

matkoniecz:
Unclear situation may also happen for cases where house is far away from both streets.

I thought that drop-down menu with street name suggestion would be good enough.

matkoniecz:
And sorry for not completing this one,  thanks to @dbdean for continuing it.

If you want I may help with it (or complete it), but if you prefer to complete it I can also let you do this.

dbdean:
> OK went through what you got so far. Not sure if it is possible to enter free text for the street name

It is possible to enter free text.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,95 @@
...
+                changes.addOrModify(key, name)
+            } else {
+                changes.addOrModify("$key:$languageCode", name)
+            }
+        }
+    }
+
+    companion object {
+        //TODO: this only appears to pick up nodes, so street-name suggestions don't appear
+        // in the middle of a long straight-stretch. There might need to be a better way.
```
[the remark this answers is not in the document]
dbdean: I updated the comment to say:

        // This only picks up nodes, so street-name suggestions don't appear
        // in the middle of a long straight-stretch (longer than 0.5km). This happens rarely
        // enough that I don't think it is a concern.

app/src/main/res/values/strings.xml
```diff
@@ -707,6 +707,11 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generalFee_title">Does it cost a fee to enter %s?</string>
     <string name="quest_laundrySelfService_title">Is this laundry a self service laundry?</string>
     <string name="quest_handrail_title">Do these steps have a handrail?</string>
+    <string name="quest_address_street_title">What street does address %s belong to?</string>
```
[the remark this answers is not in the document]
dbdean: I changed this to `address %s` rather than `house number %s` because `addr:housenumber` is frequently used on buildings (and landuses) that are not houses. I would prefer to keep it this way, as house number is not always correct.

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/data/RoadNameSuggestionsDao.kt
dbdean: This functionality is used by both AddRoadNameForm and AddAddressStreetForm and RoadNameSuggestionDao was already in AddAddressStreet, so I added the functionality here for both forms.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: The name of the function confuses me a little here too, but I think that in this case the isFormComplete is inherited from AAddLocalizedNameForm which basically says the form is not empty:

```
    // all added name rows are not empty
    override fun isFormComplete() = adapter.localizedNames.isNotEmpty()
            && adapter.localizedNames.all { it.name.trim().isNotEmpty() }
```

So as long as the form is not empty, then it is complete. But I could be wrong, I'm open to improvements.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        val housenumber = tags["addr:housenumber"]
+        return if (housenumber != null) arrayOf(housenumber) else arrayOf()
+    }
+
+    override fun createForm() = AddAddressStreetForm()
+
+    override fun isApplicableTo(element: Element): Boolean? =
+            ADDRESSES_WITHOUT_STREETS_TFE.matches(element)
```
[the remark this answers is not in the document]
dbdean: removed the need to be near a named street in https://github.com/westnordost/StreetComplete/pull/1782/commits/b1f906a0b2bbafadf9d5d46d1c92f1b9827de4d8

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: Fixed in https://github.com/westnordost/StreetComplete/pull/1782/commits/2522e90835c224c6d59a2feea053f266ed6dda89

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: Fixed in https://github.com/westnordost/StreetComplete/pull/1782/commits/d7f9933ddf9577973fb30a3665ef2e220999b692

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: Fixed in https://github.com/westnordost/StreetComplete/pull/1782/commits/d7f9933ddf9577973fb30a3665ef2e220999b692

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: ok, I removed the comment suggesting that it might need to do so in https://github.com/westnordost/StreetComplete/pull/1782/commits/7bbf0c31c7d0954980226831282b2dd40980eef8

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: Ok, I have made it go straight to the layout change in https://github.com/westnordost/StreetComplete/pull/1782/commits/5fd5672363dcde8ead5cff8496882807deec68ac

dbdean: Removed in https://github.com/westnordost/StreetComplete/pull/1782/commits/5fd5672363dcde8ead5cff8496882807deec68ac

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: Removed TODO comment in https://github.com/westnordost/StreetComplete/pull/1782/commits/7bbf0c31c7d0954980226831282b2dd40980eef8

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: It now shows no suggestions when you change to PlaceName mode.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetForm.kt
dbdean: I think I've done this in https://github.com/westnordost/StreetComplete/pull/1782/commits/9f1f245b92728da2fbda3330df1ce2e3f3a2c375, but I'm not sure how to test it.

dbdean:
> Additionally, I originally said in #213 that the quest should only be asked for housenumbers at the beginning of the road (house numbers 1, 2, 3, 4, ...) to avoid spamming the user with quests where he answers with always the same answer and because it is more efficient to enter all the `addr:street` for a street in a full editor like JOSM via multiselect. Only for places where it is not clear to which road a building belongs, it is necessary to survey it. These would be the beginning of roads and maybe intersections. What do you think of this?

I would prefer that the quest shows for all street numbers without streets/places/etc. There are a lot of places where street numbers are ambiguous even not in the low numbers (at intersections generally). I think given that (mappers using) StreetComplete created all of these numbers-without-streets (at least in my area), we should be responsible for cleaning them up. It's no more spammy than the addr:housenumber quest.

dbdean:
We could do something like (missing street-names within 50m of an intersection of two differently named roads) to catch intersections. I'm not sure how an equivalent design for addr:place would work.

Even so, could we make an option to show all for those of us who want to see them? Or maybe two quests: intersection street-names and non-intersection street names?

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
westnordost: ok

westnordost:
> I would prefer that the quest shows for all street numbers without streets/places/etc. There are a lot of places where street numbers are ambiguous even not in the low numbers (at intersections generally). I think given that (mappers using) StreetComplete created all of these numbers-without-streets (at least in my area), we should be responsible for cleaning them up. It's no more spammy than the addr:housenumber quest.

One argument for not showing the quest not near intersections was that the user might not always know in which he is currently is because he does not have a good overview.
But okay, let's go with showing it for all and see how it is accepted.

westnordost:
I am currently doing some smaller fixes and adaptions on it.

One good news: I think your assumption that `around` only works on nodes is wrong, at least since 0.7.55. See https://overpass-turbo.eu/s/U6a

westnordost:
The problem was the RoadNameSuggestionsDao, which did the checks only based on points. I am right now changing this.

westnordost:
Hmm, after I rummaged about in the code, I wonder if an interface like this wouldn't be nicer? (It's also easier to implement):

**What street does the house number 1 belong to?**
_Tap on the map to select which street it does belong to_

westnordost:
Well maybe this could be added later as an additional method how to input the road name

westnordost:
Well maybe this could be added later as an additional method how to input the road name. One simply has to override `onClickMapAt(position: LatLon, clickAreaSizeInMeters: Double)` and return true.

dbdean:
Can we use the Recycler view to show the last N chosen street names, and a blank space, then the user has to explicity select the last street name before they can click OK?

dbdean:
Might even be useful for the AddRoadName quest too.

dbdean:
Making that change is a little above my level of ability here at the moment though.

dbdean:
If the limiting it to the top-4 closes roads works, maybe we can just show them all at the same time in the form, and a button to put in a different road if it isn't one of them. Then the user could easily click on the last road, then click ok in only two touches.

westnordost:
> Making that change is a little above my level of ability here at the moment though.

Actually, it is easier than you think. The AALocalizedNameForm is a beast of complexity, that's why you think it is difficult. I will just do it quickly.

RubenKelevra:
> `What street does the (house) number %s belong to`

I think the title could be a bit more descriptive of what the user is expected to do. I was used to the 'select one from the list' with an arrow on the right of the street sign and was a bit surprised that I'm supposed to tap on a street on the map.

Maybe something like this?

`Tap on the street which this house number %s belongs to.`

westnordost:
Pretty much that text is visible there

Issue #1784, opened 2020-04-21
Hail & Ride/Hail and Ride Bus Stop Quests


peternewman:
StreetComplete incorrectly offers quests of shelter and tactile paving for hail and ride stops which don't really exist.

https://en.wikipedia.org/wiki/Hail_and_ride

**How to Reproduce**
Search for a Hail and Ride:
https://www.openstreetmap.org/search?query=Hail and Ride

See the quests exist, whereas it should be hidden in those cases.

From what I can tell officially these stops shouldn't exist and there's some way to tag the way relationship into the route instead, although the documentation for that is pretty non-existent, but others advise it's required for "It exists for routing and timetable purposes only."
https://www.openstreetmap.org/note/2165209

**Versions affected**
Android 9, SC v18.0

westnordost:
We can do nothing without a proper tagging for hail and ride bus stops. Also, why have bus stops for hail and ride at all? If it is hail and ride, there should not be bus stops at all, right?

peternewman:
> We can do nothing without a proper tagging for hail and ride bus stops.

Can you not at least filter out the irrelevant stops, or can you not do wildcard filtering on the name field?

> Also, why have bus stops for hail and ride at all? If it is hail and ride, there should not be bus stops at all, right?

Yes, that was why I initially added a note in OSM and then struggled to find any clear notes on how it should be tagged instead.

This may be relevant:
https://wiki.openstreetmap.org/wiki/NaPTAN/Surveying_and_Merging_NaPTAN_and_OSM_data#physically_present
Or
http://overpass-turbo.eu/s/oNM
Or
https://wiki.openstreetmap.org/wiki/NaPTAN/Tag_mappings#Bus.2FCoach

westnordost:
I find it a bit weird to have somethiing on a map that is not physically present at all.

But anyway, StreetComplete would not tag this stuff but just interpret it, so adding `physically_present != no` to the query is fine by me.

peternewman:
> I find it a bit weird to have somethiing on a map that is not physically present at all.

Agreed, I can't personally see the correct way to tag it in the interim and I'm loath to remove it until we can represent the data in another way.

> But anyway, StreetComplete would not tag this stuff but just interpret it, so adding `physically_present != no` to the query is fine by me.

Thanks, I think also `naptan:BusStopType != HAR` as some seem to have that but not physically present.

[...]

PR #1793, opened 2020-04-26
Don't show the bus stop quests for NAPTAN hail and ride stop either

peternewman:
As mentioned in https://github.com/westnordost/StreetComplete/issues/1784#issuecomment-617146743
> Thanks, I think also `naptan:BusStopType != HAR` as some seem to have that but not physically present = no.

More work on #1784 .

I haven't compiled and tested this, but given it's a pretty trivial test...

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_shelter/AddBusStopShelter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_shelter/AddBusStopShelter.kt
@@ -15,7 +15,7 @@ class AddBusStopShelter(o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestT
           or
           (highway = bus_stop and public_transport != stop_position)
         )
-        and !shelter and !covered and physically_present != no
+        and !shelter and !covered and physically_present != no and naptan:BusStopType != HAR
     """
     override val commitMessage = "Add bus stop shelter"
     override val wikiLink = "Key:shelter"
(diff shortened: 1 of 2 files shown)


peternewman:
Test cases: https://overpass-turbo.eu/s/TkJ

westnordost:
Hm, okay

peternewman:
Thanks, probably a bit of an edge case, but avoids someone just opening the same thing for slightly different tagging in future.

matkoniecz:
What is the meaning of this tag? 

I found https://wiki.openstreetmap.org/wiki/NaPTAN/Tag_mappings with "Hail and ride section of route," but it seems unrelated to bus stops. Or is it a small scale tagging for renderer? ( https://taginfo.openstreetmap.org/tags/naptan%3ABusStopType=HAR has just 100 of them )

peternewman:
> What is the meaning of this tag?

https://en.wikipedia.org/wiki/Hail_and_ride

> I found https://wiki.openstreetmap.org/wiki/NaPTAN/Tag_mappings with "Hail and ride section of route," but it seems unrelated to bus stops. Or is it a small scale tagging for renderer and tagging of nonexisting bus stops? ( https://taginfo.openstreetmap.org/tags/naptan%3ABusStopType=HAR has just 100 of them )

I don't know the specific reason they ended up in OSM, I guess from the NaPTAN import, and I think I read something somewhere that they are to help the bus route take the correct route effectively (I guess for the renderer). When I looked it seemed it would be possible to do it with attributes for the relationship, but that isn't how they're currently tagged.

Anyway they exist on the map, but not in the real world, so we need to filter them out so SC doesn't spam questions that can't be answered.

PR #1846, opened 2020-05-18
Rename bicyclist to cyclist

peternewman:
I haven't updated the string names or icons, I'm unclear of the challenges of that.

In English the longer word is old fashioned:
https://www.collinsdictionary.com/dictionary/english/bicyclist

Google has 72 million results for the shorter one and only 3 million for the longer one.


westnordost:
Ok

peternewman:
Thanks. I assume the icons and localisation variable names aren't worth the effort?

westnordost:
Yeah, this is not anything that is exposed to the user

PR #1847, opened 2020-05-19
AddAddressStreet: Treat a click on a non-street as as an attempt to discard form

dbdean:


dbdean:
I'm also chasing down a bug where some clicks on blank spaces are coming up with a street name which is far away.

dbdean:
In this map, when I select a AddAddressStreetQuest, and touch about where the red dot is:
image
I get "Galloway Street" even though I'm not clicking on any street, and Galloway Street is certainly not the closest to that touch - that would be Perrott Street.

When I put a breakpoint on `OnClickMapAt` the list only has Galloway Street in it, but I haven't dug any deeper to work out why `roadNameSuggestionsDao.getNames` would return that there.

Any ideas, @westnordost?

(location of data in OSM: https://www.openstreetmap.org/#map=19/-30.52346/151.64858)

dbdean:
Looking at the logic in `getNames`, Is it working on the bounding box of the road, and because Galloway Street is fairly diagonal, it is coming up over a fairly wide area?
Screenshot from 2020-05-19 19-52-26
I think I'll leave this to you @westnordost, I'm sure you understand what is happening better than me.

westnordost:
No, no idea

westnordost:
But regarding this PR: rather not. The split way form also does not allow closing the form by clicking on the map, consistency.
To have this dialog pop up only because one clicks 1mm too far away is I think not a good UX.

There is always the back button.

westnordost:
> Looking at the logic in getNames, Is it working on the bounding box of the road, and because Galloway Street is fairly diagonal, it is coming up over a fairly wide area?

Yes, Galloway Street should for this reason be in the result set from the database. But the result set from the database is then filtered further programmatically.

westnordost:
Preliminary result of my own debugging is that `LatLon.crossTrackDistanceTo` must be bugged

westnordost:
Suspicion: Does `LatLon.crossTrackDistanceTo` really calculate the distance between a position and the line segment spanned between points `start` and `end` or does it not really calculate the distance between a position and the **line** spanned between points  `start` and `end`?

westnordost:
Cause the extension of this segment in the turn certainly is closer to the tapped point than the other road:
image

dbdean:
I suspect you are right, I think the code in SphericalEarthMath is calculating the distance to the great circle, but not checking that the closest point is between the two points of the arc. Maybe we need a different formula to calculate the closest point in the great circle, and only then the distance if it is between start and end. If the closest point is outside, the distance is the direct distance to the closest of start, end?

dbdean:
You can probably take a short cut that might not need to fix crossTrackDistanceTo:

1. First find the two closest consecutive points on the street to the query point, and remember those distances
2. Calculate the cross-track distance from the query point to the great circle spanning the two closest consecutive points. Remember that distance.
3. The closest distance to the street is the minimum of the three distances calculated in (1) and (2).

dbdean:
No, that won't work actually.It wouldn't work for this example:
image
You would have to calculate the two point distances and the cross-track distance for each pair of consecutive points, and keep the minimum for all segments. I think that would work.

dbdean:
I'm going to stop now, as I clearly haven't thought this all the way through :)
That last suggestion wouldn't work in this case:
image

dbdean:
Would you like me to close this PR, and open a separate issue on the incorrect-street problem?

westnordost:
As you like, I won't forget about this bug in either case, already wrote it down

westnordost:
Fixed

PR #1868, opened 2020-05-30
exclude foot=no crossings

matkoniecz:
it is confusing for user and at least in some cases they create OSM notes
tactile paving quest already skips ones with foot=no


westnordost:
Makes sense, thanks

PR #1869, opened 2020-05-30
better log

matkoniecz:
element log is formatted as a link to OSM website allowing quicker access
dropped quest includes now also identifier of an affected OSM element

it is intended to make degugging why quest was not created easier
(in my specific case - because length limit is too aggressive for mapping on a bicycle in rural area

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmQuestDownloader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmQuestDownloader.kt
@@ -121,6 +121,6 @@ private fun LatLon.truncateTo5Decimals() = OsmLatLon(latitude.truncateTo5Decimal
 
 private fun Double.truncateTo5Decimals() = (this * 1e5).toInt().toDouble() / 1e5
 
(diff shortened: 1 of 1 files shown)


westnordost:
> element log is formatted as a link to OSM website allowing quicker access

Maybe, if the IDE one is using has this feature, but it also adds more clutter to the log. I'd rather go without this.

Also, the changes on line 80 don't seem to make sense.

matkoniecz:
> Maybe, if the IDE one is using has this feature

Even without that you can copy paste. And I assumed that Android Studio or similar IDE is standard for Android development. And at least on Lubuntu standard command line has the same feature anyway (and I use it in my scripts running on pure command line, without any IDE whatsoever).

Though if that is not convincing I will obviously remove it.

> Also, the changes on line 80 don't seem to make sense. 

Why? OSM element may exist but have a broken geometry. Or is it about a missing "for" that I will push soon?

westnordost:
Yes, it is. But still I'd not like to litter the log just because of a convenience feature of the IDE everybody is using.

westnordost:
> Why? OSM element may exist but have a broken geometry. Or is it about a missing "for" that I will push soon?

No, because you log two times the same thing

matkoniecz:
Misfeatures are now removed.

westnordost:
👍

Issue #1872, opened 2020-05-31
Act on Accessibility Report (produced as part of NGI 0)


westnordost:
Here is the report: [Report_NGI_StreetComplete.pdf](https://github.com/westnordost/StreetComplete/files/4707633/Report_NGI_StreetComplete.pdf)

Thank you, Yulia Valencia from https://accessibility.nl for the report and Mateusz for forwarding it to me. I have carefully read the report and determined the following action items:

~**Enable users who are blind to use this app:**~
You need vision to contribute with this app at all. So the below are not applicable.
- [x] ~1.3.1 (a) Connect checkboxes in "Enabled" column in quest selection screen programmatically to the header column titled "Enabled".~ will not fix, because this will only help text-to-speech systems for blind people
- [x] ~1.3.1 (b) Add "Quest completed" message when quest is completed~ will not fix, there is already audio feedback as well as an animation feedback
- [x] ~1.3.1 (c) Add a textual indication what he loading indicator is doing, f.e. "Loading..."~ Well, I'd say this is will not fix as well, because mostly aimed at blind users plus it should be pretty clear that it is a loading indicator for other users due to the rotating circle being widely known as such.
- [x] ~4.1.2 Add labels/state labels to buttons for text-to-speech~

**Enable users who have problems with their eyes (but are not blind):**
- [x] 1.4.3 Make text have more contrast, in particular, text buttons using the orange accent color. Maybe make the orange a little darker.
- [x] 1.4.4 (a) The "N" on the compass should always have the same size, independent of the system font size
- [x] 1.4.4 (b) Text on the map is not resized if the system font size is changed. The texts shown on the map should also get bigger if the system font size is bigger.
- [x] related: #1769

**Enable users who cannot use the touch screen to use this app:**
Hmm, TalkBack navigation is likely a bigger topic. There is probably more to do than what is listed in the quick accessibility report. So, it is not will-not-fix but I will probably not do this. Help is welcome, though.
- [ ] 2.1.1 TalkBack navigation for selecting the quests on the map
- [x] ~2.1.2~ this point is obsolete, report was based on old version of app
- [x] ~2.5.1 There should be buttons to control the map gestures like rotate and tilt~ Tilting and rotating the map are not vital features for using this app properly, so there is little reason to clutter the UI for such unimportant functionality

**Miscellaneous:**
[...]

PR #1877, opened 2020-06-01
Resize text on the map

matkoniecz:
This PR makes two things - text now gets larger as one zooms in, it is now also respecting device text size settings.

Both are inspired by #1872 

I admit that text now gets much larger on high zoom and I would not be surprised if it something disliked by @westnordost - in such case I may either retweaked to be subtler or entire "zoom in increases text" may be dropped (this mechanic is not a very noticeable but it actually appears in MAPS.ME, Google Maps, mapy.cz etc, so it is not something really weird).

--------

Sadly, I found no way to just pass scaling factor into map style (as say `global.text_size_scaling`) and do everything else there.

I tried things like that in map style itself and all of them failed - maybe I am missing something but it seems that it needs to be done outside map style file.

```
    text_size: [[15, global.text_size_scaling], [16, global.text_size_scaling * 2]]
    text_size: function() { return [[15, '12px'], [16, '50px']];}
    text_size: [[15, 12 * global.text_size_scaling], [16, 13 * global.text_size_scaling], [17, 15 * global.text_size_scaling], [18, 17 * global.text_size_scaling], [19, 24 * global.text_size_scaling]]
```

In case of this passing I will open PR against map style repo to add note about `text_size` and `language` getting overwritten by SC.


app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
```diff
@@ -193,7 +193,13 @@ open class MapFragment : Fragment(),
     }
 
     protected open suspend fun getSceneUpdates(): List<SceneUpdate> {
-        return listOf(SceneUpdate("global.language", Locale.getDefault().language))
+        val textScale = getResources().getConfiguration().fontScale;
```
westnordost, line 196: Don't you get suggestions/soft warnings here in Android Studio?

matkoniecz: No, without any protests. Is there anything wrong/imperfect here?

westnordost:
Not wrong, but usually Java getters are written like fields in Kotlin: 
resources.configuration.fontScale

Am 2. Juni 2020 03:25:20 MESZ schrieb Mateusz Konieczny <notifications@github.com>:
>@matkoniecz commented on this pull request.
>
>
>
>> @@ -193,7 +193,13 @@ open class MapFragment : Fragment(),
>     }
> 
>     protected open suspend fun getSceneUpdates(): List<SceneUpdate> {
>-        return listOf(SceneUpdate("global.language",
>Locale.getDefault().language))
>+        val textScale = getResources().getConfiguration().fontScale;
>
>No, without any protests. Is there anything wrong/imperfect here?

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
matkoniecz: I reset Android Studio suggestion settings (I disabled some due to completely spurious reports) and it appeared as expected.

matkoniecz:
https://github.com/ENT8R/streetcomplete-mapstyle/commit/5dcfd3226faffc68656daf26b4a0fe747bcb7cca#commitcomment-39604906 found a better solution - so it will be likely done by improving mapstyle (support for text scaling and scaling text) and once new version of mapstyle is merged then text scaling parameter will become available.

westnordost:
OK, waiting for @ENT8R to push to streetcomplete-mapstyle

westnordost:
So, only update of the map style is missing, right?

matkoniecz:
Yes, once map style is pulled in it should work (I can retest it again once that happens to avoid things like `//` as yaml comments).

westnordost:
Ok merged it to master

matkoniecz:
Retested, works, as far as I can see it can be merged.

westnordost:
Good, thanks!

PR #1885, opened 2020-06-04
workaround canvas drawing bug in Android 6 (fixes #1879)

### LINKED ISSUE #1879: Pointer pin changes resolution along left/right edge
When rotating the view while the pointer pin is at the edge of the screen, the resolution of the pointer pin gradually gets very low, when the pointer pin approaches the center of the left screen edge. Also happens at the center of the right edge of the screen. Does not happen at top or bottom edge

**How to Reproduce**
Move view away from actual position to get pointer pin; rotate the screen 

**Versions affected**
SC 20.0
Android 6.0

westnordost:
Could you please test this, @matkoniecz ?


matkoniecz:
It works :)

westnordost:
Nice, thanks

PR #1887, opened 2020-06-07
typos fixed in russian translate

IgorPolyakov:


matkoniecz:
Thanks for changes!

Would it be possible to make this edit via poeditor? See https://github.com/westnordost/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app

Changes made via pr would be overwritten on the next update of poeditor translation data (done before the update).

IgorPolyakov:
ok) i will transfer the changes to poeditor

smichel17:
> Changes made via pr would be overwritten on the next update of poeditor translation data (done before the update).

I will just put out there — https://weblate.org does continuous translation and can handle "manual" updates like this as well as those done through the web interface. The hosted version is gratis for [FLO](https://wiki.snowdrift.coop/about/free-libre-open) soft.are and the software is itself FLO, so it's possible to self-host. I haven't used POEditor so I'm not sure how it compares in terms of features and interface.

PR #1893, opened 2020-06-12
Extend building quests to nodes

Atrate:
**Rationale:** There are over 1 million (1 166 729) buildings marked as nodes instead of ways according to taginfo. This can be due to bad satellite imagery or trees obscuring the view in forests. I say the building-related quests should be extended to nodes.

**Solution:** Change all occurrences of "ways, relations" in building-related quests to "nodes, ways, relations".

Signed-off-by: Atrate <Atrate@protonmail.com>


Atrate:
Forgot resolving conflicts in my own repo will cause this PR to update, give me a while, going to force push

westnordost:
I remember deliberately deciding against this when I first implemented the quest. I don't remember the exact reason for that then, but surely they included:

It is kind of deprecated to map buildings as nodes. That they are mapped as buildings at all has I'd say historic reasons from times where the OpenStreetMap was very... rough. JOSM treats node-buildings as warnings, iD I think does not allow it at all. StreetComplete should swim with the flow here. 
For such "roughly" mapped things, the probability that the node is at a very wrong position or that it is completely bogus is also much higher than if an outline is available.

Additionally, it is much harder for the user to identify which building is meant if the outline is not shown. The user might otherwise tag the wrong building.

Atrate:
Alright, going to close this then

westnordost:
Thank you though for the effort!

PR #1899, opened 2020-06-15
increase contrast for orange text

matkoniecz:
related to #1872

> 1.4.3 Make text have more contrast, in particular, text buttons using the orange accent color. Maybe make the orange a little darker.

Current `#ff6600` orange vs `#ffffff` background has 2.94:1 contrast. Suggested contrast is 4.5:1, what would require making it noticeably darker (for example `#D14000`).

But it will in turn break contrast requirement in the dark mode where current one is sufficient.

So I changed contrast in light mode and copied current one to night mode.

I admit that despite claimed contract increase I see no change in visibility. I am not sure whatever I somehow fooled myself or maybe my eyes/phone are special.

Or tool/definitions of contrast are faulty here. Or maybe it actually increases visibility needed by people with worse eyesight.

### DIFF
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+    <color name="accent">#ff6600</color>
+    <color name="accent_variant">#F44336</color>
+
     <color name="attribution_text">#ccc</color>
 
     <color name="hint_text">#999</color>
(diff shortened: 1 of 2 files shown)


matkoniecz:
Change itself is not interesting, most of the work went into failed attempt to do it better or to find proof that it is not needed.

westnordost:
Thanks! Maybe you could do a screenshot? I think especially the "OK" button should look a bit different now

matkoniecz:
Checkmark is from automatic accessibility scanner that I used to confirm that it now passes this requirement (it is quite lame so I do not really recommend it, but it complained earlier about insufficient contrast like the report).

Screenshot_2020-06-15-23-06-56-660_de westnordost streetcomplete debug

westnordost:
Ok I'd say that it also looks good

PR #1908, opened 2020-06-19
prefer to show quests rather than labels

### LINKED ISSUE #1905: Named roads sometimes don't show quests, just dots
Named roads sometimes don't show quests, just dots

**How to Reproduce**
Find a short named road with a quest on it (e.g. surface type), zoom in, due to the name changing size, and depending on the length of the road and the length of the name, you may only ever get a dot, or sometimes get a dot and sometimes a quest. It doesn't happen on unnamed ones as there is nothing to clash with.

Sometimes the quest appears and disappears as you zoom in, sometimes it never appears.

**Versions affected**
v21.0-beta1 Android 9

matkoniecz:
fixes #1905

And, yes apparently #1462 was pure placebo.

I still see some cases where quest icon should appear but is blocked by something, but this PR removes at least one bug and reduces how many spurious collisions are happening.

Thanks to @peternewman for report, thanks to @m-rey for a testing location.

### DIFF
--- a/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
+++ b/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
@@ -58,7 +58,8 @@ layers:
             draw:
                 quest-icons:
                     interactive: true
-                    priority: function() { return feature.order }
+                    priority: 1
+                    order: function() { return feature.order }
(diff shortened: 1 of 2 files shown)


westnordost:
Nice, thank you!

PR #327, opened 2017-06-14
[wip] ask about detailing surface=paved/unpaved

matkoniecz:
fixes #279

Issue #279, opened 2017-06-01
add quest to make surface=paved/unpaved more detailed


matkoniecz:
Currently surface quest for detailed surface info, but there is no quest to make general surface values more clear.

And there is significant difference between surface=cobblestone and surface=asphalt, surface=grass_paver or surface=compacted (especially for cases like people on wheelchairs or cyclists).

I propose to add similar quests, asked for surface=paved and surface=unpaved and asking to make these data more detailed - it will allow both to make existing data more detailed and stop discouraging answering general answer.

For example I just cycled several streets, all paved. I would solve these quests as I am now waiting in queue - but as I am not remembering which part was surface=paving_stones and answering surface=paved would hide theses quests I decided to not do anything.

----

query: exclude ways with `cycleway:surface`, `surface:cycleway`, `footway:surface`, `surface:footway`, `segregated=yes`

CloCkWeRX:
Maybe a quest for https://wiki.openstreetmap.org/wiki/Key:smoothness on surface=unpaved might better target the "how well can I travel on this with a bike/wheelchair/etc"

matkoniecz:
> quest for https://wiki.openstreetmap.org/wiki/Key:smoothness on surface=unpaved might better target

smoothness is useful (for example - horrible asphalt) but in general it is very subjective. On the other hand difference between surface=grass_paver and surface=compacted is clear unambiguous and useful.

matkoniecz:
smoothness key is tricky and a separate issue so any discussion should happen in a separate issu/ticket/feature request.

matkoniecz:
Response to @westnordost in #322 (here, as it is relevant to this proposal)
  
> But what to do in the case the user can (rightfully) really not say what kind of public building it is? (ie its some weird mixture?) This is the same reason why surface=unpaved etc exists.

I never encountered case where it was impossible to use something more detailed that surface=unpaved. Sometimes road splitting was necessary and some generalization may be necessary, but so far I never encountered situation where it is impossible.

for example for http://www.openstreetmap.org/way/297445670#map=18/49.41306/21.98244 I ended using surface=compacted though I was tempted to use something like surface=asphalt;earth;sand;gravel;grass or surface=very_old_asphalt_mostly_completely_crushed_with_grass_and_earth_peaking_through_with_some_potholes_filled_with_earth.

westnordost:
I do not agree.

[...]

PR #1915, opened 2020-06-24
Detailed road surface [merge for v22]

matkoniecz:
road part of #279, revival of #327 code that was in limbo for last 3(!) years

also some related refactoring and some sort of related minor documentation


app/src/main/res/values/strings.xml
```diff
@@ -850,4 +850,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="user_profile_days_active">Days\nactive</string>
     <string name="user_profile_achievement_levels">Achievement\nlevels</string>
 
+    <string name="ic_quest_surface_detailed_answer_impossible">Multiple surfaces at once…</string>
+    <string name="ic_quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify surface? Note \"Differs along the way\" option allowing to handle different surface sections. Please use \"Can\'t say\" if there is a single surface but it is not available as an answer.</string>
```
westnordost, line 854:
```suggestion
    <string name="ic_quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify the surface? Note the \"Differs along the way\" answer option that allows to handle different surface sections. Please use \"Can\'t say\" if there is a single surface but it is not available as an answer.</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/surface/Surface.kt
```diff
@@ -23,3 +23,24 @@ enum class Surface(val item:Item<String>) {
...
 
 fun List<Surface>.toItems() = this.map { it.item }
+
+fun pavedSurfaces(): List<Surface> {
+    return listOf(
+            Surface.ASPHALT, Surface.CONCRETE, Surface.PAVING_STONES,
+            Surface.SETT, Surface.UNHEWN_COBBLESTONE, Surface.GRASS_PAVER,
+            Surface.WOOD, Surface.METAL
+    )
+}
```
westnordost, line 33:
```suggestion
val PAVED_SURFACES = listOf(
            ASPHALT, CONCRETE, PAVING_STONES,
            SETT, UNHEWN_COBBLESTONE, GRASS_PAVER,
            WOOD, METAL
    )
```

Better as constant. Also, you can import `Surface.*`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurface.kt
```diff
@@ -0,0 +1,101 @@
...
+        element.tags.forEach {
+            if(it.key.contains("surface:")) {
+                return false;
+            }
+            if(it.key.contains(":surface")) {
+                return false;
+            }
+        }
+        return true;
+    }
```
westnordost, line 56: You could write

```suggestion
    override fun isApplicableTo(element: Element) =
        REQUIRED_MINIMAL_MATCH_TFE.matches(element) && 
        element.tags.keys.none { it.contains("surface:") || it.contains(":surface") }
```

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
```diff
@@ -0,0 +1,105 @@
...
+            listOf(ASPHALT, CONCRETE, SETT, PAVING_STONES, WOOD, GRASS_PAVER).toItems()
+        else
+            listOf(DIRT, GRASS, PEBBLES, FINE_GRAVEL, SAND, COMPACTED).toItems()
+
+    // note that for unspecific groups null is used as a value, it makes them unselecteable
+    override val allItems = listOf(
+            Item(null, R.drawable.panorama_surface_paved, R.string.quest_surface_value_paved, null, pavedSurfaces().toItems()),
+            Item(null, R.drawable.panorama_surface_unpaved, R.string.quest_surface_value_unpaved, null, unpavedSurfaces().toItems()),
+            Item(null, R.drawable.panorama_surface_ground, R.string.quest_surface_value_ground, null, groundSurfaces().toItems())
+    )
```
westnordost, line 29: Wait, why are in the `allItems` all the categories, even though we are not interested at all in the categories?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
```diff
@@ -0,0 +1,105 @@
...
+import de.westnordost.streetcomplete.quests.OtherAnswer
+import de.westnordost.streetcomplete.util.TextChangedWatcher
+import de.westnordost.streetcomplete.view.Item
+import de.westnordost.streetcomplete.quests.surface.Surface.*
+
+class DetailRoadSurfaceForm  : AGroupedImageListQuestAnswerFragment<String, DetailSurfaceAnswer>() {
+
+    override val topItems get() =
+        if (osmElement!!.tags["surface"] == "paved")
+            listOf(ASPHALT, CONCRETE, SETT, PAVING_STONES, WOOD, GRASS_PAVER).toItems()
```
westnordost, line 20: Why not use `pavedSurfaces().toItems()` etc.?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
westnordost: You should probably check out the class `AImageListQuestAnswerFragment`. We do not actually want any grouping in the `DetailRoadSurfaceForm`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurface.kt
```diff
@@ -0,0 +1,101 @@
...
+    private fun getOverpassQuery(bbox: BoundingBox) =
+        bbox.toGlobalOverpassBBox() + "\n" + """
+
+          way[surface~"^(${UNDETAILED_SURFACE_TAG_MATCH})${'$'}"][segregated!="yes"][highway ~ "^${ HIGHWAY_TAG_MATCH }${'$'}"] -> .surface_without_detail;
+          // https://taginfo.openstreetmap.org//search?q=%3Asurface
+          // https://taginfo.openstreetmap.org//search?q=surface:
+          way[surface~"^(${UNDETAILED_SURFACE_TAG_MATCH})${'$'}"][~"(:surface|surface:)" ~ "."] -> .extra_tags;
+          way.surface_without_detail[surface ~ "^(${UNDETAILED_SURFACE_TAG_MATCH})${'$'}"][access !~ "^(private|no)${'$'}"] -> .not_private;
+          way.surface_without_detail[surface ~ "^(${UNDETAILED_SURFACE_TAG_MATCH})${'$'}"][foot][foot !~ "^(private|no)${'$'}"] -> .foot_access;
+          ((.not_private; .foot_access;); - .extra_tags;);
```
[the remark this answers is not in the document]
matkoniecz: All this `(:surface|surface:)` tags are really rare, so skipping them altogether is not a significant impact. And handling them is tricky (for example what should be done with `source:surface` while `surface` is edited? Removed? But what if it applies to `cycleway:right:surface`). So I decided to not think about handling them and just skip rare cases where such tags are present.

It also skips problem of people already specifying detailed surface using various tagging schemes (from relatively common to absurdly rare ones that I used in https://www.openstreetmap.org/way/141471022 ) - and I wanted to avoid "pointless" and pointless questions. It is also deliberately done to avoid irritating people tagging it somehow. Maybe I am too careful here.

matkoniecz:
`override val commitMessage = "More detailed surfaces"` - detailing surfaces of path is the next one planned, is it OK for them to share the message?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
matkoniecz: Because it is 6 surfaces (fitting nicely in two rows), while `pavedSurfaces()` have currently 8 of them. Currently `UNHEWN_COBBLESTONE` and `METAL` are not listed.

I would be also perfectly fine with listing all available paved surfaces here.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/Surface.kt
matkoniecz: Somehow I never tested importing file into itself. It works beautifully.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurface.kt
```diff
@@ -0,0 +1,101 @@
...
+        """.trimIndent() + "\n" +
+        getQuestPrintStatement()
+
+    private val HIGHWAY_TAG_MATCH = ROADS_WITH_SURFACES_BROADLY_DEFINED.joinToString("|")
+    private val UNDETAILED_SURFACE_TAG_MATCH = "paved|unpaved"
+    private val REQUIRED_MINIMAL_MATCH_TFE by lazy { FiltersParser().parse(
+            "ways with surface ~ ${UNDETAILED_SURFACE_TAG_MATCH} and segregated!=yes and highway ~ ${HIGHWAY_TAG_MATCH} and (access !~ private|no or (foot and foot !~ private|no))"
+    )}
```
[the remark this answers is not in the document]
matkoniecz: Good points, partially AS was too careful, partially leftover from previous version.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
matkoniecz: > Wait, why are in the allItems all the categories, even though we are not interested at all in the categories?

to keep interface as close as possible to other surface quests.

> We do not actually want any grouping in the DetailRoadSurfaceForm

Grouping will disappear soon.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurface.kt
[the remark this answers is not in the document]
westnordost: You think `surface:left` and `right` is absurd? Sounds fine to me, actually.

Anyway, I think at least after the sidewalk- and cycleway-surface quest are implemented, `cycleway:left/right/both:surface` etc. will be quite common.

I think:
- `source:surface` could be either deleted or ignored (better deleted)
- if `note:surface` (or `description:surface`.. only 95 usages) is defined, do not ask the question
- and the rest ignored

westnordost:
It is okay, but it would be more informative if you explicitly mentioned that this is road surfaces.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
westnordost: Let's list all of them. Some months ago, we removed most/all of the "Show more" stuff.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
westnordost: What do you mean, it will disappear soon? Anyway, it would be better if you used `AImageListQuestAnswerFragment`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
matkoniecz: switched to one big uniform list

matkoniecz:
> It is okay, but it would be more informative if you explicitly mentioned that this is road surfaces. 

Also done.

matkoniecz:
> Good idea to solve the "can't specify it any more detailed" with surface:note!

It was from TODO notes mentioning that @westnordost requested this feature :)

matkoniecz:
Retested again, looks good to me.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurface.kt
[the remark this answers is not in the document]
westnordost: By the way, maybe it is less code and easier to maintain if you used the tag filter expression to download but do the additional check in overriding `download`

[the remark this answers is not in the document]
matkoniecz: Looking at https://taginfo.openstreetmap.org/keys/description%3Asurface#values it seems that including them should not be a problem and this key is really rare anyway.

Thinking about it again - adding `surface:note` with explanation to cases known to me would not be a problem anyway.

I will switch exclusion to just `surface:note`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/Surface.kt
```diff
@@ -23,3 +24,24 @@ enum class Surface(val item:Item<String>) {
...
+            COMPACTED, FINE_GRAVEL, GRAVEL,
+            PEBBLES
+    )
+}
+
+fun groundSurfaces(): List<Surface> {
+    return listOf(
+            DIRT, GRASS, SAND
+    )
+}
```
westnordost, line 47: Why do you think these should be functions and not constants?

app/src/main/res/values/strings.xml
```diff
@@ -850,4 +850,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
 
+    <string name="ic_quest_surface_detailed_answer_impossible">Multiple surfaces at once…</string>
+    <string name="ic_quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify the surface? Note the \"Differs along the way\" answer option that allows to handle different surface sections. Please use \"Can\'t say\" if there is a single surface but it is not available as an answer.</string>
+    <string name="ic_quest_surface_detailed_answer_impossible_description">Why specifying single surface is impossible? Text length is limited to 255 characters.</string>
+    <string name="quest_surface_detailed_name_title">"What surface does the road %s have here? Please, select a specific one."</string>
+    <string name="quest_surface_detailed_square_name_title">"What surface does the square %s have here? Please, select a specific one."</string>
+    <string name="quest_surface_detailed_title">"What surface does this piece of road have? Please, select a specific one."</string>
+    <string name="quest_surface_detailed_square_title">"What surface does the square %s have here? Please, select a specific one."</string>
```
westnordost, line 859: Shouldn't have a placeholder

app/src/main/res/values/strings.xml
```diff
@@ -850,4 +850,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="user_profile_days_active">Days\nactive</string>
     <string name="user_profile_achievement_levels">Achievement\nlevels</string>
 
+    <string name="ic_quest_surface_detailed_answer_impossible">Multiple surfaces at once…</string>
+    <string name="ic_quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify the surface? Note the \"Differs along the way\" answer option that allows to handle different surface sections. Please use \"Can\'t say\" if there is a single surface but it is not available as an answer.</string>
+    <string name="ic_quest_surface_detailed_answer_impossible_description">Why specifying single surface is impossible? Text length is limited to 255 characters.</string>
```
westnordost, line 855: `ic_...`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/Surface.kt
matkoniecz: I failed to consider that, moved to constants.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
```diff
@@ -0,0 +1,100 @@
...
+import de.westnordost.streetcomplete.util.TextChangedWatcher
+import de.westnordost.streetcomplete.view.Item
+
+class DetailRoadSurfaceForm  : AImageListQuestAnswerFragment<String, DetailSurfaceAnswer>() {
+
+    override val items: List<Item<String>>
+        get() = if (osmElement!!.tags["surface"] == "paved")
+            (PAVED_SURFACES + UNPAVED_SURFACES + GROUND_SURFACES).toItems()
+        else
+            (UNPAVED_SURFACES + GROUND_SURFACES + PAVED_SURFACES).toItems()
```
westnordost, line 21: Ah! Now I understand why there was this solution with topItems. If it was paved, you first wanted to hide the non-paved items and only on another button click, show them. Sorry for the previous remark then, then it was already fine.

But it is fine like this as well.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    override val itemsPerRow = 3
+
+    override fun onClickOk(selectedItems: List<String>) {
+        // must not happen in isInExplanationMode
+        // must contain exactly 1 item (maxSelectableItems is set to 1 and isFormComplete requires selection)
+        applyAnswer(SurfaceAnswer(selectedItems[0]))
```
westnordost, line 28:
```suggestion
        applyAnswer(SurfaceAnswer(selectedItems.single()))
```
For cases like these, there is already something in the Kotlin standard library: `fun Collection.single()`

app/src/main/res/layout/quest_surface_detailed_answer_impossible.xml
```diff
@@ -0,0 +1,27 @@
...
+        android:id="@+id/explanationInput"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:inputType="textMultiLine"
+        android:gravity="top|start"
+        android:minLines="3"
+        tools:text="Surface is a mosaic of paving stones, wood and metal."
+        android:maxLength="255"
+        android:scrollHorizontally="false"
+        android:autofillHints="Tap here to write."
```
westnordost, line 25: `autofillHints` does something else than you think it does. I think you meant `hint`. But `hint` should be localized, too. A reasonable hint would be to give an example what should be written here.
On the other hand, `quest_surface_detailed_answer_impossible_description` already serves this purpose, so I think this can be removed.

```suggestion
```

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DetailRoadSurfaceForm.kt
matkoniecz: I think I like unified list more, I plan to switch to it also be normal surface quest in my fork. Less clicking to answer the question.

matkoniecz:
Two last commits fine from phone, so not tested for now

app/src/main/res/values/strings.xml
```diff
@@ -850,4 +850,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="user_profile_days_active">Days\nactive</string>
     <string name="user_profile_achievement_levels">Achievement\nlevels</string>
 
+    <string name="quest_surface_detailed_answer_impossible">Multiple surfaces at once…</string>
+    <string name="quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify the surface? Note the \"Differs along the way\" answer option that allows to handle different surface sections. Please use \"Can\'t say\" if there is a single surface but it is not available as an answer.</string>
+    <string name="quest_surface_detailed_answer_impossible_description">Why specifying single surface is impossible? Text length is limited to 255 characters.</string>
```
westnordost, line 855:
```suggestion
    <string name="quest_surface_detailed_answer_impossible_description">Please briefly specify the nature of the surface here, for example "sandy with patches of cobblestone". The text length is limited to 255 characters.</string>
```
Added missing noun markers and rephrased to include an example.

app/src/main/res/values/strings.xml
```diff
@@ -850,4 +850,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify the surface? Note the \"Differs along the way\" answer option that allows to handle different surface sections. Please use \"Can\'t say\" if there is a single surface but it is not available as an answer.</string>
+    <string name="quest_surface_detailed_answer_impossible_description">Please briefly specify the nature of the surface here, for example "sandy with patches of cobblestone". The text length is limited to 255 characters.</string>
+    <string name="quest_surface_detailed_name_title">"What surface does the road %s have here? Please, select a specific one."</string>
+    <string name="quest_surface_detailed_square_name_title">"What surface does the square %s have here? Please, select a specific one."</string>
+    <string name="quest_surface_detailed_title">"What surface does this piece of road have? Please, select a specific one."</string>
+    <string name="quest_surface_detailed_square_title">"What surface does this square have here? Please, select a specific one."</string>
```
westnordost, line 859: Thinking about this, does it really have any advantages to mention "Please select a specific one"?

Otherwise, the strings from the normal surface quest could be reused to safe the translators some work.

app/src/main/res/values/strings.xml
```diff
@@ -850,4 +850,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="quest_generic_cancel">Cancel</string>
```
westnordost, line 860: There is actually `android.R.string.cancel` (code) / `@android:string/cancel` (xml)

app/src/main/res/values/strings.xml
matkoniecz: Less confusing in quest list in settings, less confusing why paved/unpaved is gone.

westnordost: Fair enough. Maybe rephrase to "What specific surface does the road %s have here?" though then? It's more compact.

matkoniecz:
I am also thinking about path version (for `highway=path/footway/bridleway/cycleway`).

I see two solutions:

(a) duplicate code, edit necessary parts
(b) move shared part of a code to the abstract class

In this situation both seem to be acceptable - there is some shared code, but not too much of it, extremely unlikely that any new child classes would appear in the future. Do you have preference for either one?

matkoniecz:
> Nice work, thank you!

:)

westnordost:
Don't use an abstract class, there is very little code to share and from a software architecture point of view, inheritance should not be used for code-sharing anyway. (That's what delegation is for.)

PR #1916, opened 2020-06-24
The "N" on the compass have the same size, independent of the system font size

matkoniecz:
fixes 1.4.4 (a) of #1872 (NGI 0 accessibility report)

Screenshot_20200624-162906
Screenshot_20200624-162816


westnordost:
👍

PR #1917, opened 2020-06-25
Add a GitHub action to compile and lint

peternewman:
Still a WIP


westnordost:
Hm! Not sure what it does, but I salute you for taking the initiative! Reading the CI file, it looks like all of lint is executed on each push and pull request. Lint has hundreds of checks, many of which are not really relevant in my opinion. And even on my local machine, it takes longer than 10 minutes. I wouldn't fathom how long it takes on Github CI. For example, whenever I run lint, I get thousands of warnings. To be honest, I wouldn't want to have this report litter github here. 

So what would really be helpful would be to extend the `build.gradle` (Android build configuration) to include **certain** lint checks before a release is made. There is already the gradle task `lintVitalRelease` which apparently does those lint checks that classify as "fatal" errors. What would help is to find out any of:
- will a release build actually fail if `lintVitalRelease` fails?
- Is `lintVitalRelease` actually executed if the build is done through `Build -> Generate Signed Bundle/APK` from Android Studio?
- can `lintVitalRelease` be customized somehow, i.e. certain lint checks be added to those?
- (if not the above, how) can a gradle task be added that performs **certain** lint checks and fails if any check fails?
- How to make the signed release build depend on `lintVitalRelease` if it is not automatically and on the custom task with the additional lint checks? (I almost know the answer to this one, but not sure if this applies for `Build -> Generate Signed Bundle/APK` as well)

I usually run a lint inspection before each release anyway locally. Apparently I missed it one time, but this is the reason why I won't invest much time myself to look into answering these questions.

daveol:
I've tried this myself, my result after several experiments: [daveol/StreetComplete@075a8c4](/daveol/StreetComplete/commit/075a8c4f22b1e209af5b4fdca9e96b6d1611cc88)

I'll explain a bit of the things that currently happen in my version and I experimented with:
- `on.{push,pull_request}.paths-ignore`: This probably needs some tweaking, if a commit or pull request only adjusts files which match `*.md` the CI tasks won't run. This saves GitHub some needless checking or compiling by skipping the tests.
- `on.push.branches`: This is no suitable for the final version, and should probably be 'master' or something, but it prevents the test being run on branches where it's not needed, this prevents pull_requests from a branch from the same repository being ran twice.
- `jobs.{lint,test}.name`: Gives the check or build a better name.
- `actions/cache@v2`: I currently only use this to cache the gradle wrapper and dependencies downloads. I have used it on the build directories before but this makes it possible for the test to not run because gradle relies on the cached result. Caching the gradle wrapper and dependencies does speed up the build a bit tough
- `android-actions/setup-android@v2`: This is a replacement for the problem checkers used in this pull request, it also checks if `ANDROID_SDK_ROOT` is set and downloads the tools if that's needed. It is still very rough around the edges, it needs some more error handing and documentation.
- `android-actions/annotate-lint-results@v1.0.15`: The specific versioning was because of testing, but this converts the results from `app/build/reports/lint-results.xml` to Github annotations ([example](https://github.com/daveol/StreetComplete/runs/1047941440?check_suite_focus=true)) it is also still very rough around the edges, it also needs some more error handing and documentation.

In my testing it takes about 10 minutes ([more data](https://github.com/daveol/StreetComplete/actions)). However releases and accepting pull requests will not be blocked by a running or even failed CI runs, it is very much some additional information about them. If you want to wait for the results you can, but if you already ran the tests locally it's not that important since you already have the information you need. 

As for the `lint` task, maybe [disabling `abortOnError`](https://github.com/westnordost/StreetComplete/blob/master/app/build.gradle#L57) is the core of the issue?

westnordost:
> As for the lint task, maybe disabling abortOnError is the core of the issue?

There are lots of `MissingTranslation`s because only a few translations are really 100% complete. Some are 90-something% complete, some less, some more.

daveol:
> There are lots of MissingTranslations because only a few translations are really 100% complete. Some are 90-something% complete, some less, some more.

Those already get disabled on [the line above](https://github.com/westnordost/StreetComplete/blob/master/app/build.gradle#L56) though

westnordost:
I'm closing this as there hasn't been activity on this for a very long time and I am not convinced that this is necessary. I do a complete lint check before each release to weed out possible problems with the pulled translations.

PR #1920, opened 2020-06-26
Board type [merge for v22]

### LINKED ISSUE #1226: New quest: board_type=* (for information=board)
### General
Affected tag(s) to be modified/added: [board_type](https://wiki.openstreetmap.org/wiki/Key:board_type)
Question asked: **What is this information board about?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose – it's quite useful to know, what you can see on a board
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one) – an information board always has to have a topic/must be about "something"
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary – you can best see what is written on it, if you stand in front of it
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort) – `information=board` is used 126 813 times, `board_type` only 72 308 times.

### Ideas for implementation

**Element selection:** I guess that is obvious, there are no special restrictions I can see

**Metadata needed:** no

**Proposed GUI:**
image picker?
However I think an image picker might not be best due to the fact that these boards can look very different. So maybe a simple text selection of the most common values (possibly with a small explanation) would be more suitable. You could also just grab the X most-used values from taginfo and  offer these for input. (possibly in translated versions via usual POEditor translations)

---

BTW, this would nicely complement https://github.com/westnordost/StreetComplete/issues/1115, as a possible follow-up quest.

matkoniecz:
fixes #1226

Note that I am unsure about the icon.

- I can try making a new one (I am not expecting great results but I may try) 
- I can recolor one of existing icons (street quest? one that is currently added and is used for related quest?)
- or use current one (`R.drawable.ic_quest_information`), though it is used for related `AddInformationToTourism` already
- or maybe @westnordost wants to make one

BTW, I now see in diff that Android Studio had apparently decided to reformat imports - is it OK to include that or should I revert it?


app/src/main/java/de/westnordost/streetcomplete/quests/board_type/AddBoardType.kt
```diff
@@ -0,0 +1,21 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.mapdata.OverpassMapDataAndGeometryApi
+import de.westnordost.streetcomplete.data.osm.osmquest.SimpleOverpassQuestType
+
+class AddBoardType(o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestType<String>(o) {
+
+    override val tagFilters = "nodes with information=board and access !~ private|no and !board_type"
+    override val commitMessage = "Add board type"
+    override val icon = R.drawable.ic_quest_information // TODO: reusing icon - is it OK? Or maybe I should change at least colors? Or maybe @westnordost would make an icon even for such quest?
```
westnordost, line 12:
```suggestion
    override val icon = R.drawable.ic_quest_board_type
```
I added an icon

westnordost:
Looking at taginfo, it is clear that this is somewhat of a free-form tag, i.e. there can be many different values.

I think at least `map` and `public_transport` should be included. Maybe also `sight` since it is almost 1%. In Germany, many statues etc have some board with explanations, what tag would that be?

When selecting map, it should tag (instead/additionally?) tag `information=map`. For public transport, not sure if there is another tag in the public transportation scheme that should be used?

matkoniecz:
> I think at least map and public_transport should be included. 

`public_transport` is already listed.

> When selecting map, it should tag (instead/additionally?) tag information=map

instead, `information=map` cannot really exist with `information=board`

> In Germany, many statues etc have some board with explanations, what tag would that be?

I think that usually it would be `history` (based on similar ones in Poland).

matkoniecz:
> Maybe also sight since it is almost 1%

I am not sure how to shorten

> General information about a local sight, e.g. the city. This may be used if a more specific value is not applicable. 

and keep it more or less clear (for me even full form without looking at https://wiki.openstreetmap.org/wiki/File:Broomhill_Thames_View_Interpretive_Panel_8982.JPG was not really clear)

matkoniecz:
I have also a bit of trouble hot to phrase map entry without making it confusing for `board_type=history` including maps.

"Only a map"? "Just a map"? "Map for navigation"? "Can't say" -> "It is a map" -> explanation that history/nature may include a map?

westnordost:
Ah, history/nature may include a map? I didn't know that, then this is certainly something that should be mentioned.

I'd say, there are two possibilities how to solve this
1. "hide" the map option in the other answers menu and explain in the confirmation dialog that the other options may also include a map and ask to confirm whether this board really is just a map and only a map.
2. Add a hint text above the choices that hints at that all the options may also include a map. Name the map option appropriately like "Just a map and only a map"

I think option 1, the one you suggested is slightly better because it litters less of the UI and only shows the information when it is necessary.

> I am not sure how to shorten
>> General information about a local sight, e.g. the city. This may be used if a more specific value is not applicable.

"A sight that is not any of the above"?
"A sight that is about none of the aforemented topics"?
Or maybe don't show this option at all, since it is a "generic" choice but rather let the users leave a note then so that (someone else) can fill in the specific value.

matkoniecz:
> Ah, history/nature may include a map? I didn't know that, then this is certainly something that should be mentioned.


For example

> Here, during [Partitions of Poland ](https://en.wikipedia.org/wiki/Partitions_of_Poland) in years 1815-1914 Kluczwoda stream was a border between Austria and Russia

that I [tagged](https://www.openstreetmap.org/node/2960020793) yesterday as part of testing of this quest seems to be a clear case of `board_type=history` despite including a big map

granica-zaborow-w-bialym-kosciele-217056

matkoniecz:
>  "A sight that is about none of the aforemented topics"?

Example on wiki has mix of `nature` and `history`

matkoniecz:
There is one part that is the most likely to be questionable: I am not entirely sure is treating "map" as a bit of a magic value is OK.

matkoniecz:
And thanks for the icon!

app/src/main/res/values/strings.xml
```diff
@@ -849,5 +857,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="user_profile_local_rank">Rank in\n%s</string>
     <string name="user_profile_days_active">Days\nactive</string>
     <string name="user_profile_achievement_levels">Achievement\nlevels</string>
-
+    <string name="quest_board_type_map">It is a map</string>
+    <string name="quest_board_type_map_title">It is only a map?</string>
+    <string name="quest_board_type_map_description">Boards with some specific topic may include also maps. For example a public transport board may have a bus route map. Boards with topics not covered by any available answer should be solved by creating a note with \"Can\'t say\" . Is it just a map and only a map?</string>
```
westnordost, line 862:
```suggestion
    <string name="quest_board_type_map_title">It is just and only a map?</string>
    <string name="quest_board_type_map_description">If the board is about a specific topic, please specify it, regardless whether it also includes a map. For example a public transport board may have a bus route map. If the topic is not listed among the available answers, please consider leaving a note instead.</string>
```
Improved the wording a bit (in my opinion). Do you agree?

westnordost:
> There is one part that is the most likely to be questionable: I am not entirely sure is treating "map" as a bit of a magic value is OK.

That's completely okay.

peternewman:
> > I am not sure how to shorten
> > > General information about a local sight, e.g. the city. This may be used if a more specific value is not applicable.
> 
> "A sight that is not any of the above"?
> "A sight that is about none of the aforemented topics"?
> Or maybe don't show this option at all, since it is a "generic" choice but rather let the users leave a note then so that (someone else) can fill in the specific value.

Would including something about viewpoints help? That's normally where that sort of board is located. If we're meaning one like this:
https://londonist.com/2015/05/is-this-londons-most-outdated-information-board

app/src/main/res/values/strings.xml
```diff
@@ -849,5 +857,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="user_profile_local_rank">Rank in\n%s</string>
     <string name="user_profile_days_active">Days\nactive</string>
     <string name="user_profile_achievement_levels">Achievement\nlevels</string>
-
+    <string name="quest_board_type_map">It is a map</string>
+    <string name="quest_board_type_map_title">Is it just a map and only a map?</string>
+    <string name="quest_board_type_map_description">If the board is about a specific topic, please specify it, regardless whether it also includes a map. For example a public transport board may have a bus route map. If the topic is not listed among the available answers, please consider leaving a note instead.</string>
```
peternewman, line 862: How about:
```suggestion
    <string name="quest_board_type_map_description">If the board is about a particular topic, please specify it, regardless of whether it also includes a map. For example a public transport board may have a bus route map. If the topic is not listed among the available answers, please consider leaving a note instead.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -700,6 +700,14 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="create_new_note_hint">Best write the note in the locally spoken language or otherwise in English.</string>
     <string name="quest_cyclewayPartSurface_title">What\'s the surface of the cycleway here?</string>
     <string name="quest_footwayPartSurface_title">What\'s the surface of the footway here?</string>
+    <string name="quest_board_type_history">History</string>
+    <string name="quest_board_type_geology">Geology</string>
+    <string name="quest_board_type_plants">Plants</string>
+    <string name="quest_board_type_wildlife">Wildlife</string>
+    <string name="quest_board_type_nature">Nature (multiple topics)</string>
+    <string name="quest_board_type_notice_board">It is a notice board</string>
+    <string name="quest_board_type_public_transport">Public transport</string>
```
peternewman, line 709: What about adding sight as "Sight (viewpoint or panorama)" to better explain it. There are quite a few of them around.

westnordost: We'll leave out this option for now because "sight" is a generic option that often might be specified in more detail. If the user is forced to leave a note in case of a sight that does not fit into these categories, it will help to establish other specific values.

app/src/main/java/de/westnordost/streetcomplete/quests/board_type/AddBoardType.kt
```diff
@@ -0,0 +1,25 @@
+package de.westnordost.streetcomplete.quests.board_type
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.mapdata.OverpassMapDataAndGeometryApi
+import de.westnordost.streetcomplete.data.osm.osmquest.SimpleOverpassQuestType
+
+class AddBoardType(o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestType<String>(o) {
+
+    override val tagFilters = "nodes with information=board and access !~ private|no and !board_type"
```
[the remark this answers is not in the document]
westnordost: > Won't this miss board_type=board to retag?

What do you mean with this?

> Also we should cover board_type=yes for retagging too.

The quest currently applies to about 81000 elements. To change `!board_type` to `(!board_type or board_type = yes)` adds another 600 elements. So, fair enough, but it doesn't make a big difference.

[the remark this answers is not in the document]
westnordost: Then, also the line `changes.add("board_type", answer)` needs to be changed to `changes.addOrModify("board_type", answer)`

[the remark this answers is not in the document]
peternewman: Actually looking at the wiki, information is also listed as deprecated for similar reasons.

app/src/main/res/values/strings.xml
peternewman: Okay, it seems to be the preferred option compared to panorama for example. I just find notes a hassle to have to then come back and manually tag them on the OSM website.

Issue #1913, opened 2020-06-23
Add clarifications / details for quests


Helium314:
There are many situations where I am/was not sure how to answer.
Often it was possible to find an answer here on github or in the OSM wiki, sometimes not.
Since I see I am not the only one who sometimes doesn't know how to answer (example: #1889), I think it would be beneficial to have answers for relatively common situations provided directly in the app, behind some sort of "details" button (or maybe a link to the corresponding page in the OSM wiki, but often this is not particularly helpful).
Some of the questions/problems might not have a clear answer, but  for those that have this "details" button could help.

I gathered a bunch of examples where I was not sure and looked up the answer, or where I still don't know how to handle it correctly.
For some of them the answer seems rather obvious to me now, but initially it wasn't.
* lit (mostly when light is coming from nearby streets)
	* how bright does it need to be to be count as lit: see other people? be able to spot dog droppings on the ground? read a plan?
	* does it have to be lit full width?
	* when does a sports field neet to be lit to answer `yes`?
* surfaces
	* ~difference between paving stones and sett~ (better with the updated images)
	* what surfaces are tiles / natural stones (#3189, #616)
	* how precise should I be when splitting a path?
		* e.g. 2 m sett driveway on asphalt pavement?
		* e.g. 30 cm concrete speed bump on asphalt road?
		* what about crossing streets/paths with different surfaces?
	* different surfaces across the road
		* ~e.g. one lane concrete, one lane asphalt~
		* ~e.g. asphalt with paving stone patterns~
		* should just be _paved_, which forces adding a note (which is rather tedious for long roads with many segments)
* building type
	* detached house: simply defined by no contact to other houses?
	* semi-detached: should the halves be similar? What about top/bottom separation ([#2685](https://github.com/streetcomplete/StreetComplete/issues/2685#issuecomment-809011489))
	* ~apartment building: what is the difference to a house?~ (_single family_ in description should be sufficient)
	* "changed" buildings
		* e.g. old apartment building containing only offices
		* e.g. old industrial building containing apartments
		* there were issues about that, but it's not mentioned in the app
	* what kind of building is
		* container, driveway to underground parking, staircase, elevator, "cage with roof and walls" (often containing garbage/recycling containers or bicycle parking)
		* bus stop shelter: public transport or roof? (probably should be amenity=shelter)
		* subway/metro: public transport or train station?
* building levels
	* what is the lowest level (unclear semi-undergroud levels)
		* windows above ground, but no entrance
[...]

PR #1922, opened 2020-06-26
give hint about repurposed buildings

matkoniecz:
implements part of #1913

"Can't say" menu would  have a new entry:

> Purpose changed over time</string>

with 

>Select the purpose defining the building. The question is about the current building form. It is not about its historical form, it is not about how building is used.
>
> Church building used as a warehouse is still a church building.
> Train station building that is now abandoned or housing a museum should be still marked as a train station.
> But rebuilding may erase original form, for example factory may be rebuilt so extensively that the current form is now a residential building.

shown

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingTypeForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingTypeForm.kt
@@ -11,6 +11,7 @@ import de.westnordost.streetcomplete.quests.building_type.BuildingType.*
 class AddBuildingTypeForm : AGroupedImageListQuestAnswerFragment<String,String>() {
 
     override val otherAnswers = listOf(
+        OtherAnswer(R.string.quest_buildingType_answer_types_change_over_time) { showChangedPurposeHint() },
         OtherAnswer(R.string.quest_buildingType_answer_multiple_types) { showMultipleTypesHint() },
         OtherAnswer(R.string.quest_buildingType_answer_construction_site) { applyAnswer("construction") }
     )
@@ -40,6 +41,14 @@ class AddBuildingTypeForm : AGroupedImageListQuestAnswerFragment<String,String>(
             .setMessage(R.string.quest_buildingType_answer_multiple_types_description)
             .setPositiveButton(android.R.string.ok, null)
             .show()
+        }
+    }
+
+    private fun showChangedPurposeHint() {
+        activity?.let { AlertDialog.Builder(it)
+                .setMessage(R.string.quest_buildingType_answer_types_change_over_time_description)
+                .setPositiveButton(android.R.string.ok, null)
+                .show()
         }
     }
 }
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -542,6 +542,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_generic_item_invalid_value">Please select a more specific value.</string>
     <string name="quest_buildingType_answer_multiple_types">It has multiple purposes</string>
     <string name="quest_buildingType_answer_multiple_types_description">Simply select the main purpose of this building. E.g. if there is a shop on the ground floor with apartments above it, it is still mainly an apartment building.</string>
+    <string name="quest_buildingType_answer_types_change_over_time">Purpose changed over time</string>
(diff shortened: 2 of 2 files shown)


matkoniecz:
ping @Helium314

westnordost:
But `building` is tagged, so wouldn't that hint be wrong?

matkoniecz:
> wrong?

Which part?

westnordost:
> The question is about the current building form. It is not about its historical form

Right, I only read the first paragraph. The explanation is a bit long.

Anyway, I think I stated this before but I think the current uncertainty (in, that it is not explained) is better:
1. If the user has no doubt that a church used as an office building should be tagged as a church building, that's fine
2. If the user has no doubt that a church used as an office building should be tagged as an office building, that's still sort of fine because `building=office` is still an improvement from `building=yes`
3. If the user has doubt how to tag it, he can leave a note and explain it. This is actually the preferred way in such a case, because it allows that the building is tagged both for `building` and `building:use` correctly after discussion.

Offloading such a decision onto the user makes it more complex to him and at the same time does it not improve the situation, because he then might consider to tag it correctly as a church but the information that it is now used as an office is lost.

So this is why this hint should not be implemented.

matkoniecz:
> building=office is still an improvement from building=yes

I think that it is not an improvement, as it is both incorrect and harder to spot and fix.

I was thinking also about case of confused user that wants more info, become more confused and quits instead of leaving notes. Not sure how common this is, but so far everything I have seen indicates that casual/newbie users are quite confused and unsure  

(warning about a small biased sample size, also it is possible that such people would be unlikely to become OSM mappers anyway - AFAIK noone continued mapping after app testing - both from initial user testing and also from later similar and even smaller tests).

matkoniecz:
> So this is why this hint should not be implemented.

Can you review #1913? I liked at least some parts (for example about surface), but it would be probably more efficient to review ideas from it without making PRs for them.

westnordost:
> Can you review #1913

I'll do that sooner or later. For now it's too much and there are enough other open things to do while at the same time I do not have much time for StreetComplete at the moment.

matkoniecz:
> I'll do that sooner or later. 

Thanks! I will wait for that before making any new PRs based on ideas from it.

> I do not have much time for StreetComplete at the moment

So even bigger thanks for review of my PRs.

peternewman:
> > The question is about the current building form. It is not about its historical form

Erm that's exactly what the tag is about @matkoniecz, from the top of the wiki page https://wiki.openstreetmap.org/wiki/Key:building :

> Note that it may be not the same as the building's current use (tagged using building:use=*). For example, a hospital building that is abandoned or repurposed to be a marketplace is still a building=hospital, and to mark active hospitals amenity=hospital is used.

> 
> Anyway, I think I stated this before but I think the current uncertainty (in, that it is not explained) is better:
> 
>     1. If the user has no doubt that a church used as an office building should be tagged as a church building, that's fine
> 
>     2. If the user has no doubt that a church used as an office building should be tagged as an office building, that's still sort of fine because `building=office` is still an improvement from `building=yes`

Surely that means we're getting it tagged with the wrong data and that makes it much harder for a more experienced mapper to be able to spot it's building=church and building:use=office, whereas if it's just yes they know it needs a survey.

>     3. If the user has doubt how to tag it, he can leave a note and explain it. This is actually the preferred way in such a case, because it allows that the building is tagged both for `building` and `building:use` correctly after discussion.
> 
> Offloading such a decision onto the user makes it more complex to him and at the same time does it not improve the situation, because he then might consider to tag it correctly as a church but the information that it is now used as an office is lost.
> 
> So this is why this hint should not be implemented.

:facepalm: Just realised you're saying the same thing as me.

Controversial suggestion, given building/building:use is a bit of a mess, would it be worth considering either de-prioritising the quest, or splitting it into two so it first asks you is this building still being used for the purpose it was made for, and then either tags building if yes, or asks you twice and tags both if not? While possibly adding the direct quest as a disabled by default option in the quest selection for experienced operators so they can avoid the additional question (and just manually tag/note the new usage somehow).

westnordost:
The building quest is relatively high up because the housenumber quest depends on it. Splitting it in two quests would make the flow much longer for 99% of cases (while also not being possible anyway because there is no tag in OSM that signifies "building use is the same as building type").

So, I will close this PR as will not fix/merge. I agree that leaving a note always breaks the flow somewhat but I see no good way to handle this edge case without breaking the user's flow as per my last comment.

PR #1923, opened 2020-06-26
add alias to can't say about nonexisting objects [PR adding strings, not mergeable before v22]

### LINKED ISSUE #1097: Add option 'has been removed'
As, in the Netherlands, post boxes (to send mail) are disappearing everywhere. When a quest asks for collection times, an option 'has been removed' is a viable one.

Perhaps a query per county could indicate which kind of SC supported objects are disappearing with a certain high rate in order to add that 'has been removed' option is a practical addition to a query in that country or region.

Phone boots and post offices have disappeared all together here, but that might be an ongoing process in another country, so will more of certain objects that get outdated.

### LINKED ISSUE #1673: Place does not exist anymore answer option
I've just noted an inexperienced user changing opening hours to "closed down", they successfully closed my note, but the relevant POI was not updated with disused:amenity. I don't want to raise this directly with the user, and am not familiar with the StreetComplete interface, so cannot absolutely be sure this is a bug.

The relevant changeset was [77693359](https://www.openstreetmap.org/changeset/77693359) with the add opening hours challenge. Note I've now updated amenity, name & shop tags.

matkoniecz:
It turns out that many users were especially confused about questions referring to nonexisting objects
and were unaware that the proper solution is to create a note (though many of them figured this out)
this change will guide them toward note feature as applicable in case of nonexisting objects

main negative is increased complexity and more work for translators but as this confusion
is quite widespread (mentioned both in talks with more experienced mappers and revealed
in testing with newbies) it seems to be a worth doing

fixes #1673 fixes #1097 using a different method

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -43,7 +43,7 @@
     <string name="quest_leave_new_note_title">"Leave a note instead?"</string>
     <string name="quest_leave_new_note_description">"You can leave a public note at this location which other mappers can comment on and hopefully resolve or alternatively just hide the quest for yourself only"</string>
     <string name="quest_leave_new_note_request">"Please, leave a note"</string>
-    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that no longer exist. It happens when OpenStreetMap contains incorrect or outdated info. You can leave a public note at this location with an explanation of what is wrong or alternatively just hide the quest for yourself only."</string>
+    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that are not existing. It happens when OpenStreetMap contains incorrect or outdated info. You can leave a public note at this location with an explanation of what is wrong or alternatively just hide the quest for yourself only."</string>
     <string name="quest_leave_new_note_yes">"OK"</string>
     <string name="quest_leave_new_note_no">"No, just hide"</string>
     <string name="quest_leave_new_note_photo">Attach photo</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -43,7 +43,7 @@
     <string name="quest_leave_new_note_title">"Leave a note instead?"</string>
     <string name="quest_leave_new_note_description">"You can leave a public note at this location which other mappers can comment on and hopefully resolve or alternatively just hide the quest for yourself only"</string>
     <string name="quest_leave_new_note_request">"Please, leave a note"</string>
-    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that are not existing. It happens when OpenStreetMap contains wrong or outdated info. You can leave a public note at this location with explanation what is wrong or alternatively just hide the quest for yourself only."</string>
+    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that no longer exist. It happens when OpenStreetMap contains incorrect or outdated info. You can leave a public note at this location with an explanation of what is wrong or alternatively just hide the quest for yourself only."</string>
     <string name="quest_leave_new_note_yes">"OK"</string>
     <string name="quest_leave_new_note_no">"No, just hide"</string>
     <string name="quest_leave_new_note_photo">Attach photo</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -35,12 +35,15 @@
     <string name="map_attribution_osm">"© OpenStreetMap contributors"</string>
     <string name="quest_generic_otherAnswers">"Other answers…"</string>
     <string name="quest_generic_answer_notApplicable">"Can’t say…"</string>
+    <string name="quest_generic_answer_notExisting">"Does not exist…"</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -35,12 +35,15 @@
...
+    <string name="quest_generic_answer_notExisting">"Does not exist…"</string>
     <string name="quest_generic_confirmation_title">"Are you sure?"</string>
     <string name="quest_generic_confirmation_yes">"Yes, I am sure"</string>
     <string name="quest_generic_confirmation_no">"I will check"</string>
     <string name="quest_generic_error_field_empty">"You left this field empty"</string>
     <string name="quest_leave_new_note_title">"Leave a note instead?"</string>
     <string name="quest_leave_new_note_description">"You can leave a public note at this location which other mappers can comment on and hopefully resolve or alternatively just hide the quest for yourself only"</string>
+    <string name="quest_leave_new_note_request">"Please, leave a note"</string>
+    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that are not existing. It happens when OpenStreetMap contains wrong or outdated info. You can leave a public note at this location with explanation what is wrong or alternatively just hide the quest for yourself only."</string>
```
peternewman, line 46: The English could probably be improved by saying something like:
```suggestion
    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that no longer exist. It happens when OpenStreetMap contains incorrect or outdated info. You can leave a public note at this location with an explanation of what is wrong or alternatively just hide the quest for yourself only."</string>
```

Let me know if it needs to be a bit briefer and we could probably reword it slightly to shorten it.

notExisting for the variable isn't great, but given you're trying to say "doesn't exist" or "no longer existing" both could be a bit hard to camel case. Actually non-existent might be better (perhaps as nonExistent)?

app/src/main/res/values/strings.xml
```diff
@@ -35,12 +35,15 @@
...
+    <string name="quest_generic_answer_notExisting">"Does not exist…"</string>
     <string name="quest_generic_confirmation_title">"Are you sure?"</string>
     <string name="quest_generic_confirmation_yes">"Yes, I am sure"</string>
     <string name="quest_generic_confirmation_no">"I will check"</string>
     <string name="quest_generic_error_field_empty">"You left this field empty"</string>
     <string name="quest_leave_new_note_title">"Leave a note instead?"</string>
     <string name="quest_leave_new_note_description">"You can leave a public note at this location which other mappers can comment on and hopefully resolve or alternatively just hide the quest for yourself only"</string>
+    <string name="quest_leave_new_note_request">"Please, leave a note"</string>
+    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that are not existing. It happens when OpenStreetMap contains incorrect or outdated info. You can leave a public note at this location with an explanation of what is wrong or alternatively just hide the quest for yourself only."</string>
```
peternewman, line 46: Try this instead then @matkoniecz:

```suggestion
    <string name="quest_leave_new_note_about_nonexisting_object_description">"Questions may be asked about objects that do not exist. It happens when OpenStreetMap contains incorrect or outdated info. You can leave a public note at this location with an explanation of what is wrong or alternatively just hide the quest for yourself only."</string>
```

westnordost:
A problem I see with this is that it always adds this option, for any quest. Even quests where it is not likely at all that it is gone, for example a street, a sidewalk, a railway crossing etc etc.

Furthermore, quests that have otherwise no other answers than "can't say" have the "can't say" option right at the lower left corner of the answer bubble. Adding always this option will hide the "can't say" option in the other answers menu. So, this might actually be counter-productive as your intention is to hint users towards the "can't say" button but what you actually do is to hide it in the first place.

The very first thing and almost the only thing that is explained at all _in the tutorial shown to every first user_ is that if he can't answer a quest for one reason or another, he shall answer "can't say" to leave a note. I think the placement of this information in the tutorial is a much better place than in every "other answers" menu for every quest.

matkoniecz:
> Even quests where it is not likely at all that it is gone, for example a street, a sidewalk, a railway crossing etc etc.

Many reports about gone features were exactly about streets, not only in areas with bad imports (TIGER) but also in locations where area was poorly mapped for some other reasons or situation on the ground changed. Railway crossing becoming gone is also normal in regions dismanling railway infrastructure. For example in Poland many spurs, especially leading to former industrial areas in city centers are getting removed.

> Adding always this option will hide the "can't say" option in the other answers menu. 

And it will become "Other answers...". I admit that I am unsure whatever it will have negative or neutral (positive is unlikely) effect.

> The very first thing and almost the only thing that is explained at all in the tutorial shown to every first user is that if he can't answer a quest for one reason or another, he shall answer "can't say" to leave a note. I think the placement of this information in the tutorial is a much better place than in every "other answers" menu for every quest.

Yes, all this observations that resulted in implementing this predate intro and I have never tested impact of intro for users. Though I assumed that like any intro/docs/tutorial it will be skipped/ignored by large part of userbase.

matkoniecz:
Overall, I think that this change would be improvement but I am not 100% sure and it is partially based on outdated feedback.

westnordost:
Hm I would rather not have this in the app due to the reasons given in my last comment.

PR #1924, opened 2020-06-26
WIP: Position gps button to left of zoom controls

smichel17:
Context in commit message (where it belongs), copied here (for ease of access):

> When gps is enabled and the button disappears, it's awkward to have an
empty space below the zoom buttons. This moves the gps button to the
side and shifts the zoom buttons down, forming a .: shape. This way,
when the gps button disappears, the zoom buttons sit nicely at the
bottom.
> 
> This is a proof of concept implementation; it has an awkward interaction
with the "off screen" location pin, since the wrapping LinearLayout is
bigger than the visible buttons, so the pin is shown prematurely.

Looks like this:

Screenshot from 2020-06-26 11-17-46

I'll fix the pin interaction if you're interested in merging.

…and, uh, ignore the branch name :smile: I was originally going to propose an orthogonal set of changes but then decided to do this first.

### DIFF
--- a/app/src/main/res/layout/fragment_main.xml
+++ b/app/src/main/res/layout/fragment_main.xml
@@ -103,37 +103,47 @@
             android:layout_alignParentBottom="true"
             android:layout_alignParentRight="true"
             android:layout_marginBottom="22dp"
-            android:orientation="vertical"
             android:clipToPadding="false"
             tools:ignore="RtlHardcoded,RtlSymmetry">
-
-            <ImageButton
-                android:id="@+id/zoomInButton"
-                android:layout_width="@dimen/map_button_size"
-                android:layout_height="@dimen/map_button_size"
-                android:scaleType="center"
-                style="@style/RoundWhiteButton"
-                android:src="@drawable/ic_zoom_plus_black_24dp"
-                android:contentDescription="@string/map_btn_zoom_in"/>
-
-            <ImageButton
-                android:id="@+id/zoomOutButton"
-                android:layout_width="@dimen/map_button_size"
-                android:layout_height="@dimen/map_button_size"
-                android:scaleType="center"
-                style="@style/RoundWhiteButton"
-                android:src="@drawable/ic_zoom_minus_black_24dp"
-                android:contentDescription="@string/map_btn_zoom_out"/>
 
             <de.westnordost.streetcomplete.location.LocationStateButton
                 android:id="@+id/gpsTrackingButton"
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
What about keeping the linear layout but the optional gps on top?

smichel17:
That's also an option. Personally I prefer this. I think it looks better, and it mirrors the structure above, with appearing/disappearing buttons for notifications and manual uploads.

If you're making the suggestion because you think the implementation is easier that way, it's only slightly so. The bug is actually present in the current version if you position the location indicator in the empty space where the gps button was, it just doesn't feel quite as jarring due to the location.

smichel17:
I don't believe the bug will be be hard to fix; I just wanted to verify interest in merging before I spent the time.

westnordost:
Hm, not sure why you think it is awkward to have an empty space there. It is more convenient to reach the buttons with the thumb if the buttons are not on the lower edge.

Also, I was thinking about removing those two buttons and replace them with a slider similar to what can be seen in the app Mapy.cz

smichel17:
Mainly inefficient use of space -- moving the buttons down gives a larger main area.

Personally I pinch more than I tap the buttons, although I do use the buttons -- mostly when I am wearing gloves and can't easily make precise motions.

westnordost:
Did you try out Mapy.cz? You should. They also have this pointer pin. Generally, their UI is quite excellent. Mapy.cz superseded Maps.ME as my favorite end user OSM based app.

smichel17:
I haven't. I'll consider it. I have only 4 proprietary apps installed on my phone right now (excepting the crap that came preinstalled and cannot be removed :angry:; I've settled for hidding those via my launcher) and I'd like to stay at that number (or lower). I'll probably try it in an emulator, but getting an apk off of the play store is annoying (or I can set up an emulator with Google services, I suppose), so no guarantee of when I feel like doing that :P

westnordost:
Anyway, I'd like to keep the UI as it is right now. Note that modern smartphones are much longer than wide, by the way. The Nexus 4 is ancient.

smichel17:
Regardless of the size of the phone, empty space below the +/- buttons is wasted; it's very hard to show anything useful in the scribbled area because not much is small enough to fit there; the same space above the buttons allows you to view something bigger in the main area (or zoom in more) before it overlaps the zoom buttons. Screenshot from my current phone, a Galaxy S6

Galaxy S6 screenshot

I don't feel strongly enough to continue arguing about this right now. I can bring it up again at whatever point in the future you have more time.

PR #1928, opened 2020-06-28
fix #1831 adaptive launcher icon

### LINKED ISSUE #1831: New Icon
The old icon is old, boring and does not conform to the current app icon design guidelines anymore anyway.
Does someone have ideas or maybe even mockups for a new icon?

pstorch:
I still like the clean simple style of the existing launcher icon. This is just an adaptive version of it.

street_complete_adaptive_icon


westnordost:
Nice, thank you very much!

PR #1930, opened 2020-06-29
show icons where possible, fix #1927

### LINKED ISSUE #1927: In some cases dot may be shown instead the quest icon (for example way with both path surface quest and track quest may display neither, even with free space available)
As I mentioned in  #1908 there are still some quests not displayed when icon should appear. During testing I confirmed it to be quite irritating, as it makes some quest unsolvable.

It seems that I finally found  way to reproduce this, though it is still quite fragile.

**How to Reproduce**
Reinstall app

Set GPS location to https://www.openstreetmap.org/#map=19/50.06947/19.94163 and zoom into that point

Start downloading quests (triggering it manually), at some point path surface quest icons will appear. 

Keep downloading quests, once `lit` quest will be downloaded quest icons on footways are disappearing making both `surface` and `lit` quests unsolvable with dots shown instead.

EDIT: do not disable any quest types it may interfer with repreoduction of this bug. In my case disabling quests except `surface` for paths and `lit` before downloads blocks reproduction, though disabling other quests after downloading and reproduction is not changing situation...

I will try to debug and fix this, especially as it is related to  #1908 #1462 #1877 and may be caused by my PR.

**Versions affected**

Latest master, noticed both in Android 6 and Android 10.

matkoniecz:
https://github.com/westnordost/StreetComplete/pull/1930#issuecomment-651145134

<s>having priority 2 for icons and 1 for dots causes icons to be blocked
priority 10 for icons and 1 for dots works, at least on Xiaomi hardware

I have no idea what is happening here and I sytongly suspect that
my testing may be somehow wrong, independent confirmation would be welcomed
see #1927 report with info how this may be triggered

I suspect that this hack may be a bit fragile, real fix would be nice

Yes, setting `order` for dots/quest-dots sounds like a proper fix,
but it is not working (dots still block icons)

I also tested replacing `function() { return feature.order }` by some
constant (1, 10, 100, 1000, 10000) - it failed to change behavior,
so bad order value assigned there appears to not be the reason.</s>

### DIFF
--- a/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
+++ b/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
@@ -58,7 +58,7 @@ layers:
             draw:
                 quest-icons:
                     interactive: true
-                    priority: 10 #hack, but having priority 2 here and 1 for dots causes icons to be blocked by dots displayed for the same object (despite "collide: false")
+                    priority: 0
                     order: function() { return feature.order }
                     size: 66px
                     sprite: function() { return feature.kind }
@@ -69,6 +69,7 @@ layers:
             draw:
                 quest-dots:
                     priority: 1
+                    order: 1
                     size: 16px
                     collide: false
                     offset: [-2.5px, -22px]
--- a/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
+++ b/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
@@ -58,7 +58,7 @@ layers:
             draw:
                 quest-icons:
                     interactive: true
-                    priority: 1
+                    priority: 10 #hack, but having priority 2 here and 1 for dots causes icons to be blocked by dots displayed for the same object (despite "collide: false")
                     order: function() { return feature.order }
                     size: 66px
                     sprite: function() { return feature.kind }
@@ -68,6 +68,7 @@ layers:
             filter: { $zoom: { min: 16 } }
             draw:
                 quest-dots:
+                    priority: 1
                     size: 16px
                     collide: false
                     offset: [-2.5px, -22px]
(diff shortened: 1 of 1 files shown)


westnordost:
Since I cannot reproduce it but you can, maybe you could try to do the following:

Do not show the dots at all. To see if this problem is at all related to the dots.

Also, dots should always be behind the quest pins. The lower the priority, the more in front it is. Setting a higher priority for dots than for icons seems wrong.

matkoniecz:
For `order` higher values appear to win - see https://github.com/tangrams/tangram-docs/blob/1f5bf1c6b0568b42aab4ad9cb9c5abfc8e55bc1e/docs/Syntax-Reference/draw.md#order ("higher-ordered layers will be drawn over lower-ordered layers")

For `priority` that is about which style will be selected it is not specified - https://github.com/tangrams/tangram-docs/blob/610d78b966d603bfdd409580d32e8a793d3d29cd/docs/Syntax-Reference/layers.md#priority

matkoniecz:
> Do not show the dots at all. To see if this problem is at all related to the dots.

I did it during solving #1905 and I did it again. I tried removing both dot layers (dots not shown, icons shown), I tried keeping `dots_no_collide` (dots are blocking icons despite `collide: false`) and I tried keeping  `dots` (high zoom dots removed) - and this had interesting effect of icon missing completely at high zoom!

So I set `priority` of icons to 0, what now works as expected.

I have no idea why `priority: 10` worked. I opened https://github.com/tangrams/tangram-docs/issues/269 as at least for me it was confusing.

matkoniecz:
Thanks for pointing me to idea of retesting again without dots, this time I did all three variants of this test what finally pointed me in the right direction.

Testing of this would be welcomed, at this point the icon/dots/labels ordering is slowly getting ridiculous.

westnordost:
Nice, thank you very much for all this testing and fixing it!

peternewman:
Thanks @matkoniecz . I can confirm it's working for me too, including on a road which was problematic on the v21.0 beta.

PR #1943, opened 2020-07-09
Don't enter addresses in Luxembourg

grischard:
Please disable the adding of addresses in Luxembourg for now. The Luxembourg community is [working on a conflation](https://github.com/osmlu/addresses-tools) of the [official addresses database](https://github.com/osmlu/csventrifuge), and having new addresses with just the house number popping up is making our manual work (even) harder.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
@@ -28,6 +28,7 @@ class AddHousenumber(private val overpass: OverpassMapDataAndGeometryApi) : OsmE
 
     // See overview here: https://ent8r.github.io/blacklistr/?streetcomplete=housenumber/AddHousenumber.kt
     override val enabledInCountries = AllCountriesExcept(
+            "LU", // placeholder
             "NL", // https://forum.openstreetmap.org/viewtopic.php?id=60356
             "DK", // https://lists.openstreetmap.org/pipermail/talk-dk/2017-November/004898.html
             "NO", // https://forum.openstreetmap.org/viewtopic.php?id=60357
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
@@ -28,7 +28,7 @@ class AddHousenumber(private val overpass: OverpassMapDataAndGeometryApi) : OsmE
 
     // See overview here: https://ent8r.github.io/blacklistr/?streetcomplete=housenumber/AddHousenumber.kt
     override val enabledInCountries = AllCountriesExcept(
-            "LU", // placeholder
+            "LU", // https://github.com/westnordost/StreetComplete/pull/1943
             "NL", // https://forum.openstreetmap.org/viewtopic.php?id=60356
             "DK", // https://lists.openstreetmap.org/pipermail/talk-dk/2017-November/004898.html
             "NO", // https://forum.openstreetmap.org/viewtopic.php?id=60357


westnordost:
Could you add a link to the relevant discussion of the Luxembourg community for documentation? Linking this PR has no value because the PR exists in the Git History anyway.

So in the future, there should not be any survey-based recording of housenumbers in Luxembourg because everything is imported?

Also, if this PR just serves to make the import easier, I fear it is not an effective measure: If StreetComplete users are stopped to be asked for housenumbers in Luxembourg, this doesn't make the current places where only the housenumber but not the street has been recorded go away. So it seems like in any case, the import would need to made intelligent enough that it can deal with incomplete addresses in one way or another.

matkoniecz:
Note also that SC is now asking also for `addr:street ` (for technical reason there is some delay, but such quests should also appear).

Is Luxembourg community supporting stopping also that quest?

DavidMoraisFerreira:
Hello,
Not to sound pompous, but Guillaume and I are the only ones in the luxembourgish community that handle imports and large-scale QA. Therefore our discussions occur on [#osmlu](https://riot.im/app/#/room/#osmlu:matrix.org), which unfortunately does not leave a permanent record. Once we identify the import requirements and process, the usual import documentation process will be followed.

While I greatly appreciate any contribution to OSM, we have had the issue that StreetComplete users added large ranges of housenumbers (e.g. 70-88) onto one large block. This is caused by old and imprecise building outlines. Additions like these make it difficult to handle the import process. Furthermore, by only adding `addr:housenumber`, we have to run "enrichment" scripts over and over again.

I don't think the goal is to disable StreetComplete addresses in LU forever, it's just a temporary measure until we get the 90% imported.

westnordost:
Ok, then maybe this PR is the best documentation then after all.

From what you say, it sounds like a general problem, not one exclusive to Luxembourg. So, are you sure that this measure would actually be effective? Note that:
1. as mentioned, current incomplete addresses don't go away by this change
2. StreetComplete is not a web-app. Even if I release a new update within this week, it takes time to roll out, especially for F-Droid users. Then of course, people will not update right away (again, especially F-Droid users). I can see on the Google Play Publisher backend that there are still users who use ancient versions of the app
3. Luxembourg is not big. There are only a handful users that ever contributed to it and only two or three that are currently active. If this is a temporary measure to make it easier to import, then maybe it would be more effective to just write these few users to temporarily stop contributing housenumbers. I know the ids of the StreetComplete users that have been active in Luxembourg in the last few months

westnordost:
But whatever you think is most effective

grischard:
The combination of 1 and 3 mean that David and I have been fixing these by hand :).

westnordost:
Ok, so my remarks considered, do you still want to not have the housenumber quest show up for Luxembourg for now?

grischard:
Well, ideally StreetComplete wouldn't let the users submit an address without a street and a city. I realise it's a blunt instrument, but it seems to be the best possible option for now.

westnordost:
ok

aqwzs:
Hey, I'm sorry if it's inappropriate but is there any way the housenumber quest could be re-enabled, maybe if possible the user could get a warning when entering a house number without only numbers. Plus, the query used to find those wrong tags returns no result anymore so I don't see why we should be limiting the user.
Thank you :)

aqwzs:
So is there any update for now? yes, no, maybe, no idea, wait?

westnordost:
You need to discuss this with @grischard and @DavidMoraisFerreira

westnordost:
Or actually, in general with the Luxembourg community, if there are more

grischard:
Moien @aqwzs ! The queries don't return anything at the moment because David and I are still fixing addresses by hand all the time. If you want to help out with the conflation of the open data addresses, you're very welcome to join us in the Luxembourg community chat at https://matrix.to/#/#osmlu:matrix.org

Issue #1950, opened 2020-07-12
Allow to specify that split ways qualify for quest


matkoniecz:
**Use case**
I am testing parking lane quest (that is not generated for ways near `amenity=parking` and therefore not triggerable without overpass query).

Currently on splitting ways with parking lane quest none will get parking lane quest.

And one needs to wait until rescan happens or flush cache.

I want to add special property "afterSplitAllWaysQualifyForThisQuest".

**Proposed Solution**
No idea, but I will look for it if it is not rejected.

Not sure whatever any of current quest would benefit from that, probably not.

westnordost:
How should that work, technically? I don't think this is (easily) possible.

After all, not every split way should receive this quest, right?

matkoniecz:
In this case, every way that had parking lane quest and splitting was used as an answer will produce 2 or more ways.

All such ways should receive parking lane quest.

Splitting used as solution of other quests should not result in creation of parking lane quests.

No idea whatever it is implementable in a sane way.

westnordost:
Hmm, actually I think it might. `SplitWaysUploader:uploadSingle` uploads a single split way and knows the quest type. It could be hacked in there somewhere...

westnordost:
I'll do it

matkoniecz:
Done by #1951

PR #1951, opened 2020-07-12
Recreate quests on split way upload

westnordost:
implements #1950

for review and test @matkoniecz


matkoniecz:
I am not really able to judge the code or whatever database is actually needed and so on, but I tested it and it works beautifully.

In theory it may need disabling for some quests but it is purely theoretical for StreetComplete. And would be relatively easy to implement once actually needed.

matkoniecz:
Thank you for this, it makes solving roadside parking quest significantly less flustrating!

PR #1958, opened 2020-07-15
allow sidewalk-separate answer

### LINKED ISSUE #1925: Allow answering "sidewalk mapped separately" (sidewalk=separate) in the sidewalk quest
Via can't say menu

**Use case**
https://www.openstreetmap.org/note/2244771

**Proposed Solution**
See title (I will make pr for this after the signal that it would be accepted)

matkoniecz:
fixes #1925

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/SidewalkAnswer.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/SidewalkAnswer.kt
@@ -1,5 +1,5 @@
 package de.westnordost.streetcomplete.quests.sidewalk
 
 sealed class SidewalkAnswer
-open class SidewalkSides(val left:Boolean, val right:Boolean) : SidewalkAnswer()
+data class SidewalkSides(val left:Boolean, val right:Boolean) : SidewalkAnswer()
 object SeparatelyMapped: SidewalkAnswer()
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/SidewalkAnswer.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/SidewalkAnswer.kt
@@ -1,3 +1,5 @@
 package de.westnordost.streetcomplete.quests.sidewalk
 
-data class SidewalkAnswer(val left:Boolean, val right:Boolean)
+abstract class SidewalkAnswer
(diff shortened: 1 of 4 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/SidewalkAnswer.kt
```diff
@@ -1,3 +1,5 @@
 package de.westnordost.streetcomplete.quests.sidewalk
 
-data class SidewalkAnswer(val left:Boolean, val right:Boolean)
+abstract class SidewalkAnswer
+open class SidewalkSides(val left:Boolean, val right:Boolean) : SidewalkAnswer()
+object SeparatelyMapped: SidewalkAnswer()
```
westnordost, line 5: Please use sealed classes. Then you can delete the 
```
            else -> throw IllegalArgumentException("it should never happen")
```
in that other file

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/SidewalkAnswer.kt
```diff
@@ -1,3 +1,5 @@
 package de.westnordost.streetcomplete.quests.sidewalk
 
-data class SidewalkAnswer(val left:Boolean, val right:Boolean)
+sealed class SidewalkAnswer
+open class SidewalkSides(val left:Boolean, val right:Boolean) : SidewalkAnswer()
```
westnordost, line 4:
```suggestion
data class SidewalkSides(val left:Boolean, val right:Boolean) : SidewalkAnswer()
```

app/src/main/res/values/strings.xml
```diff
@@ -849,5 +849,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="user_profile_local_rank">Rank in\n%s</string>
     <string name="user_profile_days_active">Days\nactive</string>
     <string name="user_profile_achievement_levels">Achievement\nlevels</string>
+    <string name="quest_sidewalk_separately_mapped">Displayed separately on the map</string>
```
westnordost, line 852:
```suggestion
    <string name="quest_sidewalk_separately_mapped">Yes, displayed separately on map</string>
```

"Displayed separetely [...]" doesn't sound like an answer to "Does this street have a sidewalk?"

matkoniecz:
I retested it and new text still fits.

PR #1993, opened 2020-08-02
Add icon for "Kind to Animals" achievement

FloEdelmann:
I have only created the SVG with Inkscape, I hope it can be converted to an Android XML drawable without much hassle.

Used (and edited) images:

* Apple, leaf, carrot: from StreetComplete quest icons
* Squash: https://openclipart.org/detail/212733/vegetables
* Daikon: https://openclipart.org/detail/312402/simple-daikon
* Aubergine: https://openclipart.org/detail/229813/aubergine
* Bananas, pear, avocado: https://openclipart.org/detail/280304/fruit-icons-1-package

Preview with achievement frame:

ic_achievement_veg

PR #1997, opened 2020-08-03
Add graphic for "Sixth Sense" achievement

FloEdelmann:
Follows #1993.

Used (and edited) images:

* Dog: https://openclipart.org/detail/284912/police-dog
* person, streetlights, zebra crossing, trees: StreetComplete achievement graphics by Sanja Dimitrijevic

Preview with achievement frame:

ic_achievement_blind


westnordost:
Sanja wrote: "I love it! Looks great!". She also said that she wants to start again contributing more graphics within a month or two.

I have a few comments regarding this graphic:
- the graphic is a little empty in the center. All the detail is on the sides. Not exactly sure what to do about it. Maybe one solution would be to make the person with his dog bigger so that he sticks out to the upper right and move the traffic lights to the right, maybe also enlarging it a bit.
- when I last researched it, the the yellow band with three black dots is actually mostly or only used in Germany, the international symbol for blind people is basically a pedestrian with a blindman's stick. Maybe it could be made more clear that he is blind, because his blindman's stick is also not that well visible by giving him sunglasses... or something

matkoniecz:
> yellow band with three black dots

On the first sign it seemed to be a radioactivity symbol.

HolgerJeromin:
The stick should probably be white

westnordost:
White stick on white zebra strip?

smichel17:
Maybe this is just because I'm not familiar with the badge on the person's arm, but at that size it looks like :radioactive: to me

peternewman:
> The stick should probably be white

> White stick on white zebra strip?

I assume your downvote was due to the lack of contrast @westnordost ? They're generally white:
https://en.wikipedia.org/wiki/White_cane
https://www.rnib.org.uk/cane-explained

The zebra crossing should have equal ratio of black and white, so the cane would be more obvious over a black bit of crossing, or perhaps it could have an outline, or one or the other be off white. See arguably the most famous zebra crossing in the world:
https://en.wikipedia.org/wiki/Abbey_Road,_London

peternewman:
>     * the graphic is a little empty in the center. All the detail is on the sides. Not exactly sure what to do about it. Maybe one solution would be to make the person with his dog bigger so that he sticks out to the upper right and move the traffic lights to the right, maybe also enlarging it a bit.

If he walked the other way, the man could be much bigger in the middle of the frame, with the dog either within it or escaping the frame.

michaelblyons:
> If he walked the other way, the man could be much bigger in the middle of the frame, with the dog either within it or escaping the frame.

Dropping the leash in favor of the special harness for guide dogs would bring the dog and person closer together, and they could both fit in the middle. (Also less ambiguity in the scene, since only blind people use those harnesses as far as I know.)

peternewman:
> since only blind people use those harnesses as far as I know.)

It seems other types of service dog can too, but probably not many/any pets.

westnordost:
A friend of mine will drew an icon for this, so this can be closed.

Issue #1912, opened 2020-06-23
Show type of place when asking for details


Helium314:
Some places are rather badly named, e.g. stores without name / with only generic name that are named after the owner in OSM.
In such cases it is often completely unclear which of possibly several nearby shops is meant in the quest.
Another somewhat similar situation is a place that has different opening hours for different parts, e.g a shop with attached Café where both have the same name in OSM, and it is unclear which part is meant.

In such cases it would help to know the type of shop, so one can better identify the place where details should be added (wheelchair access, opening hours,...).

I am aware that these cases are quite rare, and always adding the type can increase the amount of text considerably. So maybe the type could be added to the "can't say" menu.

matkoniecz:
I was sometimes confused and unable to locate place, while "oh, I am looking shoe shop" or "right, is is a restaurant" type of hint allowed me to locate it. Or made my more certain that it is actually gone.

westnordost:
For which quests should this be done in your opinion?

matkoniecz:
opening hours, wheelchair and probably other business ones (I solve primarily opening hours one).

Helium314:
This, and possibly the cash payment quest.
Maybe the WiFi access quest... but I would assume this is only asked for restaurants and cafés (never tried it).
Vegetarian/vegan quest is only for restaurants and cafés, so adding the type would not help.

Helium314:
I don't want to be as annoying as I feel I am right now, but almost every time I'm answering opening hours and wheelchair quests in Czechia I have problems, because (compared to more "western" countries):
* Shops often have no name (and even many that have, have it written in a way that you need to search for it)
* Many people apparently don't/didn't care about putting the actual name (if it exists) into the name tag, often the name tag seems to contain some random word written in large letters on the store front
* There seems to be rather high fluctuation, meaning many shops only live for a short time
* Shop locations are often only accurate within ~20 m, which is not enough when there are several shops nearby

All this means I have to really search for the place the quest is about in roughly 30% of the cases, and even if I don't find it often I am not certain that it doesn't exist.
So I would really appreciate ANYTHING that helps me identify a shop...

I would even try to add this enhancement myself, as it seems to be rather simple. But after looking at the code I concluded my non-existent coding skills are not enough to even follow how anything makes its way into the strings (replacing %s).

westnordost:
[...]

Issue #2001, opened 2020-08-06
Consistently start changeset comments with uppercase letters 


cyclingcat:
I just noticed that the comments StreetComplete put into its changesets have the case of the first letter (often: "add") used inconsistently. See this screenshot from my OSM profile:

![Change_sets](https://user-images.githubusercontent.com/32601012/89538885-66390f00-d7fb-11ea-91d8-0315b3fc4874.png)

As changeset comments have the nature of headlines I'd suggest to convert the (probably few) comments with a lowercase first letter to uppercase (e.g. "**A**dd whether traffic signals have a button for pedestrians").

(SC 20.2 and 21.2 from F-Droid.)

The cycling cat

westnordost:
I don't mind either way. If you create a PR for this, I will accept it.

westnordost:
So will you do it? For me, Uppercase consistency in changeset comments has absolutely no importance for me.

Atrate:
If no one does, I'll spin up a PR but only in about two weeks' time

cyclingcat:
If you don't mind, please leave this ticket open. I have never created a PR so far and have always wondered how this works, so I want to take this opportunity to learn how to do it, assuming it's an almost trivial case here, ideal for beginners. (The coming hot days prevent me from cycling anyway :-))

The cycling cat

westnordost:
Basically you fork this repository, do your changes, commit (better on a branch) and push them to your forked repository. Then, on github, you tap on the "New Pull Request" button in this repository and select to pull from your repository ("compare across forks" link).
The Github desktop client might make the workflow easier, I don't know.

matkoniecz:
> I have never created a PR so far and have always wondered how this works, so I want to take this opportunity to learn how to do it, assuming it's an almost trivial case here, ideal for beginners.

If you are stuck with something - ask here.

And yes, it is nearly simplest possible task (except that part that you need to check and change now than one text string).

Atrate:
To make it even simpler, you don't even need to download the repo to your computer, such simple changes can be done in Github's integrated editor — fork, edit, save by comitting, create PR.

PR #2012, opened 2020-08-11
Add type of place

Helium314:
add type of place to wheelchair access and opening hours quests as discussed in #1912 
due to my limited experience I'm not 100% sure everything is correct, but it worked flawlessly in the tests I did.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -100,7 +100,7 @@ class AddOpeningHours (
             R.string.quest_openingHours_no_name_title
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
-        val name = tags["name"] ?: tags["brand"] ?: featureName.value
+        val name = tags["name"] ?: tags["brand"]
         return if (name != null) arrayOf(name,featureName.value.toString()) else arrayOf()
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -92,7 +92,10 @@ class AddOpeningHours (
 
     override fun getTitle(tags: Map<String, String>) =
         if (hasProperName(tags))
-            R.string.quest_openingHours_name_title
+            if (hasFeatureName(tags) && !tags.containsKey("brand"))
+                R.string.quest_openingHours_name_type_title
+            else
+                R.string.quest_openingHours_name_title
         else
             R.string.quest_openingHours_no_name_title
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -98,7 +98,7 @@ class AddOpeningHours (
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
         val name = tags["name"] ?: tags["brand"] ?: featureName.value
-        return if (name != null) arrayOf(name) else arrayOf()
+        return if (name != null) arrayOf(name,featureName.value.toString()) else arrayOf()
     }
 
     override fun download(bbox: BoundingBox, handler: (element: Element, geometry: ElementGeometry?) -> Unit): Boolean {
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
@@ -95,7 +95,7 @@ class AddWheelchairAccessBusiness(
             R.string.quest_wheelchairAccess_name_title
     
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
-        val name = tags["name"] ?: tags["brand"] ?: featureName.value
+        val name = tags["name"] ?: tags["brand"]
         return if (name != null) arrayOf(name,featureName.value.toString()) else arrayOf()
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
@@ -85,6 +85,11 @@ class AddWheelchairAccessBusiness(o: OverpassMapDataAndGeometryApi) : SimpleOver
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_wheelchairAccess_name_title
 
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        val name = tags["name"] ?: tags["brand"]
+        return if (name != null) arrayOf(name,featureName.value.toString()) else arrayOf()
+    }
+
     override fun createForm() = AddWheelchairAccessBusinessForm()
 
     override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
@@ -89,7 +89,7 @@ class AddWheelchairAccessBusiness(
     override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
 
     override fun getTitle(tags: Map<String, String>) = 
-        if (hasProperName(tags) && !tags.containsKey("brand"))
+        if (hasFeatureName(tags) && !tags.containsKey("brand"))
             R.string.quest_wheelchairAccess_name_type_title
         else
             R.string.quest_wheelchairAccess_name_title
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
@@ -1,11 +1,16 @@
 package de.westnordost.streetcomplete.quests.wheelchair_access
 
+import de.westnordost.osmfeatures.FeatureDictionary
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.osm.osmquest.SimpleOverpassQuestType
 import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
 import de.westnordost.streetcomplete.data.osm.mapdata.OverpassMapDataAndGeometryApi
+import java.util.concurrent.FutureTask
 
-class AddWheelchairAccessBusiness(o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestType<String>(o)
+class AddWheelchairAccessBusiness(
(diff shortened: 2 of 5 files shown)


westnordost:
Implementation looks fine.

Are you sure there are no other quests where a name is displayed but the feature name should also be displayed? This pull request requires some work directly after merge from me because I need to manually change all the translations for the modified strings. I'd rather if I only do this one time.

matkoniecz:
In theory - "Is this place wheelchair accessible?", but it has extremely limited use, so I am unsure about whatever it is worth doing.

Helium314:
For wheelchair acces of businesses it is already in the PR.
For "Is the restroom at %s wheelchair accessible?" it might be ok, but here I think it's similar to the diet quest, this is probably asked mostly at food places: And maybe a few other places, like public buildings.
For public transport and outside I don't see any use.

I will add type of place for the Cash quest, here it could be useful as this quest is for many kinds of places.

Helium314:
I noticed that there is something wrong: If the tag exists, brand names are shown instead of type.
So I'll try to fix this...

Helium314:
And for this node: https://www.openstreetmap.org/node/4462863502 the type is shown as "null", possibly because SC does not recognize the (rare) shop=printing

Helium314:
For now I set it to use the new string (with type) only if `hasFeatureName` (in opening hours) is true and there is no brand (`!tags.containsKey("brand")`).

If the conditions are not satisfied, there is a fallback to the old string (without type).
This works, but I think it would be useful to show the type also for places with brands. Not so much for McDonald's, but more for less known brands.
So... how can I get the type also if there is a brand?

matkoniecz:
> For now I set it

Is it some new commit? Have you pushed it to github? Or is it describing current state?

Helium314:
This is just current state on my PC.
All I know about github, pull request & stuff comes from #2001, so I am new and slow and (for now) don't know how to edit the code on github other than using the online editor...

I will update on github once I manage to get the type even if there is a brand. Or after I give up on this.

Helium314:
So I gave up on trying to get the type if there is a brand, and added the `hasFeatureName` function to cash & wheelchair quests to avoid `null` as type of place.
Cash quest is added, opening hours & wheelchair acces updated.

I did not yet test this on every quest, will do that later today.

Helium314:
So far it looks good, only two (unrelated!) observations: (@westnordost)
* I was asked the accepts cash quest for a marketplace. Does that really make sense? I would expect this can be different depending on the trader.
* The wheelchair quest is not asked if the place has no name. Why opening hours, but not wheelchair?

Helium314:
No, but I moved the map to some random Swedish city (ended up in Norrköping) to test the cash acceptance quest

westnordost:
Mh, I guess a marketplace makes no sense

> The wheelchair quest is not asked if the place has no name. Why opening hours, but not wheelchair?

Maybe if it has no name, it is unlikely that it is big enough that it is worth asking. I think many of the quests have this requirement so first the name has to be specified, then it is easier to refer to it in later quests.

Helium314:
Hmm yes, many on the no-name places I found can't even be entered. Maybe that makes sense.

westnordost:
Also, it would mean another string to translate for all those quests

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -83,11 +88,23 @@ class AddWheelchairAccessBusiness(o: OverpassMapDataAndGeometryApi) : SimpleOver
     override val icon = R.drawable.ic_quest_wheelchair_shop
     override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
 
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_wheelchairAccess_name_title
+    override fun getTitle(tags: Map<String, String>) = 
+        if (hasFeatureName(tags) && !tags.containsKey("brand"))
```
westnordost, line 92: Why the check for brand? Don't you want _"What are the opening hours of McDonalds (fast food restaurant)?"_ Why not?

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -83,11 +88,23 @@ class AddWheelchairAccessBusiness(o: OverpassMapDataAndGeometryApi) : SimpleOver
...
+            R.string.quest_wheelchairAccess_name_type_title
+        else
+            R.string.quest_wheelchairAccess_name_title
+    
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        val name = tags["name"] ?: tags["brand"] ?: featureName.value
```
westnordost, line 98: Why the `?: featureName.value`?

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
Helium314: see https://github.com/westnordost/StreetComplete/pull/2012#issuecomment-675608159
What I actually get is "What are the opening hours of McDonalds (`brand name in chinese`)?"
If there is a brand, then it is shown instead of type. Usually in English, sometimes in other languages.
And I have no clue where this comes from...

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
Helium314: Honestly I just copied `getTitleArgs` from AddOpeningHours.kt.
I wasn't sure about it since it is not in in the default `getTitleArgs`, but decided to copy everything from opening hours.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
westnordost: Okay, I also know why, but why not? It is in Chinese if you are in China, right?

Helium314: No, I had it in Austria and Czechia, with the language set to UK English... and since I feel completely unable to solve this, I decided to take the easy way out :/

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
westnordost: The line means
`assign to variable "name" what's the value of the tag "name", if it is not set, what's the value of the tag "brand" and if that is not set, the generic feature name`

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
westnordost: Okay, that must be a bug iin the Feature Dictionary

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
Helium314: Then acually the `featureName.value` is useless... (and could give strange questions when occurring twice)
Since the Wheelchair Quest requires a name anyway (in the Overpass query), `tags["name"]` must be set and ` val name = tags["name"]` should be enough.

westnordost: Well, `val name = tags["name"] ?: tags["brand"]` would be enough

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
westnordost: Ok I released a new version of the feature dictionary. 

https://github.com/westnordost/osmfeatures/blob/master/CHANGELOG.md

Would you try it out (update dependency in the build.gradle - 1.2)

westnordost: The code you should use is

```kotlin
// if the feature can be named (generally), use the "Is NiceHaircut (barber shop) wheelchair accessible?" string 
        if (hasFeatureName(tags))
            R.string.quest_accepts_cash_type_title
// otherwise "Is NiceHaircut wheelchair accessible?" string 
        else
            R.string.quest_accepts_cash_title


    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
// same here -> can it be named generally? Then supply both name and generic feature name
        if (featureName.value != null)
            arrayOf(name, featureName.value.toString())
// otherwise just supply the name
        else
            arrayOf(name)
    }

```
The comments are just for you to understand, please don't also commit them

westnordost: (Nevermind my deleted comment, it is fine the way it is)

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
westnordost, line 98:
```suggestion
        val name = tags["name"] ?: tags["brand"]
```

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -92,13 +92,16 @@ class AddOpeningHours (
...
-            R.string.quest_openingHours_name_title
+            if (hasFeatureName(tags) && !tags.containsKey("brand"))
+                R.string.quest_openingHours_name_type_title
+            else
+                R.string.quest_openingHours_name_title
         else
             R.string.quest_openingHours_no_name_title
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
         val name = tags["name"] ?: tags["brand"] ?: featureName.value
```
westnordost, line 103:
```suggestion
        val name = tags["name"] ?: tags["brand"]
```

westnordost (changes requested):
I realized that your solution is fine this way. It's not so clean, but I see no other way without changing more.
The featureDictionary contains also brand names but in this case, we explicitly don't want brand names. The only things that should be changed is marked as suggestions, you can just click to apply them.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
Helium314: Works nicely, now it shows the English brand name.
However, I still think using brand instead of type can look strange (e.g. *"What are the opening hours of McDonald's (McDonald's)?"* ). So I would like to keep excluding brands from being shown.

westnordost: Ok, I would have thought that your `hasFeatureName(tags) && !tags.containsKey("brand")` check led to that in this case, the brand is not shown?

Helium314: With `if (hasFeatureName(tags) && !tags.containsKey("brand"))` it works like before, showing *"What are the opening hours of McDonald's* without brand or type.
But here the updated feature dictionary does not matter, which is why I initially tested with `if (hasFeatureName(tags))` (without the exclusion of brands). Then brands are still shown instead of type, but at least in English.

westnordost (approved):

westnordost:
Turned out to be more complicated after all, so thank you for your persistance.

PR #2018, opened 2020-08-14
add versionNameSuffix '-DEBUG' to debug version

Strubbl:
to see in the about screen, where the version is displayed, that the debug version is currently used

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -32,7 +32,9 @@ android {
         minSdkVersion 17
         targetSdkVersion 29
         versionCode 2203
-        versionName "22.2"
+        def version = "22.2"
+        versionName version
+        buildConfigField "String", "VERSION", "\"$version\""
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
     buildTypes {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -48,6 +48,7 @@ android {
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
             applicationIdSuffix ".debug"
+            versionNameSuffix '-DEBUG'
         }
     }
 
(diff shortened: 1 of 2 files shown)


westnordost:
For the debug version, already the icon is yellow 

Am 14. August 2020 21:33:30 MESZ schrieb Strubbl <notifications@github.com>:
>to see in the about screen, where the version is displayed, that the
>debug version is currently used
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/westnordost/StreetComplete/pull/2018
>
>-- Commit Summary --
>
>  * add versionNameSuffix &#39;-DEBUG&#39; to debug version
>
>-- File Changes --
>
>    M app/build.gradle (1)
>
>-- Patch Links --
>
>https://github.com/westnordost/StreetComplete/pull/2018.patch
>https://github.com/westnordost/StreetComplete/pull/2018.diff

Strubbl:
Does that mean, you do not want to add the suffix to the version string? If yes, please close this PR.

Strubbl:
I just thought about the about screen and the crash report mails, where the version name is in. This way you can see that it's the debug version which is installed. 

But of course if that change is not need, feel free to close the PR. It's just a proposal.

westnordost:
Hmm, that makes sense. No, it was just a comment.

westnordost:
Before this is merged, it just needs to be made sure that it has no unwanted sideeffects. For example, will the statistics generation still recognize the debug version as "StreetComplete"?

westnordost:
I checked the statistics generation scripts. That should be no problem.

I checked the source code of the app. There is one problem: See usage of `BuildConfig.VERSION_NAME` in `NotificationsSource.kt` which is subsequently put into the data class `NewVersionNotification` which is passed on to `WhatsNewDialog`: The `WhatsNewDialog` expects the version strings to not have any suffix.
I think easiest and cleanest solution would be to strip the suffix from BuildConfig.VERSION_NAME in `NotificationsSource`, f.e. `private val currentVersion: String get() = BuildConfig.VERSION_NAME.stripTheSuffixFromItIfThereIsAny()`

Strubbl:
I have no knowledge about kotlin and therefore i cannot implement the `.stripTheSuffixFromItIfThereIsAny()` part. If somebody could help here, that would be nice.

westnordost:
Well, you could always implement it Java-like, like
```
private fun getCurrentVersion(): String {
    // strip the suffix from BuildConfig.VERSION_NAME if it is there and return
}
```

Kotlin comes with a ton of useful string manipulation functions (more than Java), so I'd have to look myself. I think [String.substringBefore](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/substring-before.html) should do.

Strubbl:
I solved it with the introduction of a build config variable in the build.gradle

westnordost:
Oh, good!

PR #2020, opened 2020-08-14
update leakcanary to 2.4 from 1.5.4

Strubbl:
with the old version included in the debug version of the app, i got a strange behaviour where it is asking me for storage permission, but this permission was already given.

i upgraded leakcanary according to their documentation: https://square.github.io/leakcanary/upgrading-to-leakcanary-2.0/


westnordost:
Nice, also less boilerplate code! Thanks!

PR #2022, opened 2020-08-14
Library updates, lint cleanup

TacoTheDank:
Updated gradle wrapper to 6.6 using git bash.


westnordost:
Thank you @TacoTheDank , though:

My experience is that one cannot safely apply all those lint fixes offered by Android studio without good knowledge of the codebase itself and the rammifications thereof. This is why I tend to reject this PR unless it is well tested.
Especially I find many of the lint checks useless or not improving readability. I often run the linter myself and deliberately choose not to "correct" certain things, many of which I see now in your PR.

Some of the things you changed seem useful, such as the update of gradle or the upgrade of some dependencies, though I would like to ask you to do each of these in singular PRs so that these can also be tested individually if they break anything.

TacoTheDank:
That's true, especially with as large a codebase as this one. I'll undo the lint cleanup.

TacoTheDank:
I went through all the library updates to make sure that they wouldn't break by comparing the .jar source releases of each release up to where they are now, so they're probably good.

westnordost:
Okay, thank you!

PR #2023, opened 2020-08-15
Use ContextCompat methods.

Isira-Seneviratne:
* Make use of `ContextCompat` methods such as `getSystemService()`.
* Use the `getSystemService()` extension function in the Kotlin code.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestAutoSyncer.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestAutoSyncer.kt
@@ -1,28 +1,26 @@
 package de.westnordost.streetcomplete.data.quest
 
 import android.annotation.SuppressLint
-import android.content.BroadcastReceiver
-import android.content.Context
-import android.content.Intent
-import android.content.IntentFilter
-import android.content.SharedPreferences
+import android.content.*
 import android.location.LocationManager
 import android.net.ConnectivityManager
 import android.util.Log
+import androidx.core.content.getSystemService
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
 import androidx.lifecycle.OnLifecycleEvent
-
-import javax.inject.Inject
-
 import de.westnordost.osmapi.map.data.LatLon
 import de.westnordost.osmapi.map.data.OsmLatLon
 import de.westnordost.streetcomplete.Prefs
 import de.westnordost.streetcomplete.data.download.*
 import de.westnordost.streetcomplete.data.upload.UploadController
-import de.westnordost.streetcomplete.location.FineLocationManager
 import de.westnordost.streetcomplete.data.user.UserController
-import kotlinx.coroutines.*
+import de.westnordost.streetcomplete.location.FineLocationManager
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import javax.inject.Inject
 import javax.inject.Singleton
 
 /** Automatically downloads new quests around the user's location and uploads quests.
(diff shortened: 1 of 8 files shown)


Atrate:
What's the rationale behind that? Does it improve performance or stability? From what I can see it does not seem to improve readability...

Isira-Seneviratne:
Some of the functions, such as `Resources.getDrawable()`, have been deprecated.

westnordost:
I don't see much value in this actually. Yes, they are deprecated, but as long as they are not removed from the API(, and I don't think they ever will be,) it's fine.
In some months and up to a year from now, compatibility with Android <5 will be dropped and then all the current deprecated usages of `getDrawable` etc. can be changed to the non-deprecated form anyway, right?

TacoTheDank:
@westnordost Well in that case, I say why not just get it done now, if it's going to be done anyway?

matkoniecz:
> Well in that case, I say why not just get it done now, if it's going to be done anyway?

Because it will keep support for Android <5 phones for longer, giving owners time for an upgrade (I am betting that number of active users with such phones is decreasing and will hit acceptable numbers for EOL in such time).

Atrate:
> Also, using ContextCompat 's getSystemService() function allows services to be retrieved in a backward-compatible manner

Doesn't this mean A5 compat would be kept?

Isira-Seneviratne:
> 
> 
> > Also, using ContextCompat 's getSystemService() function allows services to be retrieved in a backward-compatible manner
> 
> Doesn't this mean A5 compat would be kept?

Yeah, it calls `Context`'s `getSystemService(Class)` on Marshmallow and higher, and `getSystemService(String)` on lower versions.

app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestAutoSyncer.kt
```diff
@@ -55,8 +52,7 @@ import javax.inject.Singleton
     }
 
     // new location is known -> check if downloading makes sense now
-    private val locationManager = FineLocationManager(
-        context.getSystemService(Context.LOCATION_SERVICE) as LocationManager) { location ->
+    private val locationManager = FineLocationManager(context.getSystemService()!!) { location ->
```
westnordost, line 55: Even though it is not required, it would enhance readability to specify which system service is queried here

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
```diff
@@ -83,15 +84,11 @@ open class LocationAwareMapFragment : MapFragment() {
...
-            (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager).defaultDisplay,
+        compass = Compass(context.getSystemService()!!,
+            (context.getSystemService<WindowManager>()!!).defaultDisplay,
             this::onCompassRotationChanged
         )
-        locationManager = FineLocationManager(
-            context.getSystemService(Context.LOCATION_SERVICE) as LocationManager,
-            this::onLocationChanged
-        )
+        locationManager = FineLocationManager(context.getSystemService()!!, this::onLocationChanged)
```
westnordost, line 91: same here

app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestAutoSyncer.kt
Isira-Seneviratne: Okay.

Isira-Seneviratne:
> 
> 
> So as mentioned, the `ContextCompat.getDrawable` is something I wouldn't want in the app right now because it does not increase readability and when upgrading the minSdk to Lollipop, it might be missed to revert back to using `resources.getDrawable(...)` because it is not shown as a warning in the linter anymore. What you could do is to start a "Lollipop" branch that shall include all the changes to be done when the minSdk is set to lollipop, if you like. I could then later take on this branch when it is time to remove support for 4.x.
> 
> Right, but the system service part makes sense because it increases readability and will not be obsolete once the minSdk is set to Lollipop.

Okay, I'll revert the `ContextCompat.getDrawable()` commit.

westnordost:
Looks good, thank you!

PR #2024, opened 2020-08-15
Update a few libraries

TacoTheDank:
From https://github.com/westnordost/StreetComplete/pull/2022

- Use Kotlin JDK8
- Update RTL-Viewpager 1.0.2 -> 2.0.0 ([changelog](https://github.com/duolingo/rtl-viewpager/blob/master/CHANGELOG.md))
- Update YAMLBeans 1.13 -> 1.15 ([changelog](https://github.com/EsotericSoftware/yamlbeans/releases))

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -123,7 +123,7 @@ dependencies {
     implementation 'androidx.exifinterface:exifinterface:1.2.0'
 
     // Kotlin
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.8'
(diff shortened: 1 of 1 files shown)


Isira-Seneviratne:
AFAIK, `kotlin-stdlib-jdk8` only introduces extensions for Java 8-specific APIs, such as the Streams and Date/Time APIs.

westnordost:
Ok! Since the RTL viewpager update is a major one, did you test it? It is used in the opening hours quest (when selecting start and end time).

westnordost:
I made the mistake before to trust that minor updates don't break anything, so I always test before updating dependencies, especially if it is just one of many open source libraries not distributed by a big entity like Google.

TacoTheDank:
@westnordost I see. I'll get to testing it when I have the time.

TacoTheDank:
@westnordost Built a debug APK and tested. It works perfectly fine as far as I can tell. Here's the APK in case you'd like to test as well. https://mega.nz/file/8CQmCQwb#aFWdAB5A72qzmvx0X_yrKMtnTibOAHfVsz1QlUlE358

westnordost:
Then it can be merged, thank you!

PR #2025, opened 2020-08-16
Use Java 8 APIs.

Isira-Seneviratne:
* Enable [Java 8 API desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring).
* Make use of the `ZonedDateTime` and `Instant` classes.
* Make use of `Map.computeIfAbsent()`.
* Due to enabling API desugaring, [bug fixes](https://issuetracker.google.com/issues/37042460) for `ConcurrentHashMap` on Android Lollipop are included in the app.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,6 +5,8 @@ apply plugin: 'kotlin-kapt'
 
 android {
     compileOptions {
+        coreLibraryDesugaringEnabled true
+
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
@@ -36,6 +38,7 @@ android {
         versionName version
         buildConfigField "String", "VERSION", "\"$version\""
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        multiDexEnabled true
     }
     buildTypes {
 
@@ -106,6 +109,10 @@ dependencies {
     // dependency injection
     implementation 'com.google.dagger:dagger:2.14.1'
     kapt 'com.google.dagger:dagger-compiler:2.14.1'
+
+    // Java 8 API access
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.0.10'
+    implementation 'androidx.multidex:multidex:2.0.1'
 
     // Android stuff
     implementation 'com.google.android.material:material:1.2.0'
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/elementgeometry/NodeWayMap.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/elementgeometry/NodeWayMap.kt
@@ -3,21 +3,15 @@ package de.westnordost.streetcomplete.data.osm.elementgeometry
 
 /** Knows which vertices connect which ways. T is the identifier of a vertex  */
 class NodeWayMap<T>(ways: List<List<T>>) {
-    private val wayEndpoints = LinkedHashMap<T, MutableList<List<T>>>()
+    private val wayEndpoints: MutableMap<T, MutableList<List<T>>> = LinkedHashMap()
 
     init {
         for (way in ways) {
             val firstNode = way.first()
             val lastNode = way.last()
 
-            if (!wayEndpoints.containsKey(firstNode)) {
-                wayEndpoints[firstNode] = ArrayList()
-            }
-            if (!wayEndpoints.containsKey(lastNode)) {
-                wayEndpoints[lastNode] = ArrayList()
-            }
-            wayEndpoints[firstNode]!!.add(way)
-            wayEndpoints[lastNode]!!.add(way)
+            wayEndpoints.computeIfAbsent(firstNode, { ArrayList() }).add(way)
+            wayEndpoints.computeIfAbsent(lastNode, { ArrayList() }).add(way)
         }
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.java
@@ -1,19 +1,20 @@
 package de.westnordost.streetcomplete;
 
-import android.app.Application;
 import android.content.SharedPreferences;
 import android.util.Log;
+
+import androidx.appcompat.app.AppCompatDelegate;
+import androidx.multidex.MultiDexApplication;
 
 import java.util.concurrent.FutureTask;
 
 import javax.inject.Inject;
 
-import androidx.appcompat.app.AppCompatDelegate;
 import de.westnordost.countryboundaries.CountryBoundaries;
 import de.westnordost.osmfeatures.FeatureDictionary;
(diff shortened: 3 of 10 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/osm/elementgeometry/NodeWayMap.kt
```diff
@@ -3,21 +3,15 @@ package de.westnordost.streetcomplete.data.osm.elementgeometry
...
-            if (!wayEndpoints.containsKey(firstNode)) {
-                wayEndpoints[firstNode] = ArrayList()
-            }
-            if (!wayEndpoints.containsKey(lastNode)) {
-                wayEndpoints[lastNode] = ArrayList()
-            }
-            wayEndpoints[firstNode]!!.add(way)
-            wayEndpoints[lastNode]!!.add(way)
+            wayEndpoints.computeIfAbsent(firstNode, { ArrayList() }).add(way)
+            wayEndpoints.computeIfAbsent(lastNode, { ArrayList() }).add(way)
```
westnordost, line 14: That's nice! Though when I paste this line into my source code locally, Android Studio warns
`Call requires API level 24 (current min is 17): java.util.HashMap#computeIfAbsent`

app/build.gradle
```diff
@@ -36,6 +38,7 @@ android {
         versionName version
         buildConfigField "String", "VERSION", "\"$version\""
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        multiDexEnabled true
```
westnordost, line 41: Why is multidex necessary? minify is enabled

Isira-Seneviratne: The desugaring process generates a secondary DEX file with the new APIs.

app/src/main/java/de/westnordost/streetcomplete/data/osm/elementgeometry/NodeWayMap.kt
Isira-Seneviratne: Yeah, that's because the implementation of `computeIfAbsent()` used in the desugaring process is included in the `Map` interface as a default method:

```
default V computeIfAbsent(K key,
            Function<? super K, ? extends V> mappingFunction) {
        Objects.requireNonNull(mappingFunction);
        V v;
        if ((v = get(key)) == null) {
            V newValue;
            if ((newValue = mappingFunction.apply(key)) != null) {
                put(key, newValue);
                return newValue;
            }
        }

        return v;
    }
```

Isira-Seneviratne:
> 
> 
> Regarding `Instant`, as `Date` and `Calendar` are not deprecated in Java 8, how is it better than those instead of simply a different API? I see it is a little shorter, but that alone is not worth replacing all the `Date`s in the code with `Instant`.

Hope [this explanation](https://www.baeldung.com/java-8-date-time-intro#issues) helps.

Isira-Seneviratne:
On another note, [warnings](https://issuetracker.google.com/issues/155423440) are displayed during the desugaring process. This has been fixed in AGP 4.1.

westnordost:
Thank you for the explanations. Actually, I'd rather refrain from it then, for now. I want to avoid using multidex. If the minSdkVersion is at one point Android Lollipop, the support library stuff for multidex isn't necessary anymore and maybe then, but from my experience, the multidex support library makes certain things more complicated. To be honest, I don't remember exactly what, I only remember that it was always worth trying to avoid multidex if possible. (Not sure if this also applies to ART because I don't know how the multidex is taken care of in ART).
After all, Proguard/R8 itself is such a big potential for spurious problems (**right now the current release on F-Droid is broken and I am struggling to find the issue in my proguard conf**) and pretty much the only reason to use code minimization through ProGuard is to avoid multidex (well, plus code obfuscation), because who cares about a few hundred KB of unused code?

**BUT** since 99% of the app is Kotlin anyway, instead of `Map.computeIfAbsent`, [`MutableMap.getOrPut`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-or-put.html) could be used.

Even though this is then not relevant anymore, I will read the article you linked in a moment, thanks!

Isira-Seneviratne:
> 
> 
> Thank you for the explanations. Actually, I'd rather refrain from it then, for now. I want to avoid using multidex. If the minSdkVersion is at one point Android Lollipop, the support library stuff for multidex isn't necessary anymore and maybe then, but from my experience, the multidex support library makes certain things more complicated. To be honest, I don't remember exactly what, I only remember that it was always worth trying to avoid multidex if possible. (Not sure if this also applies to ART because I don't know how the multidex is taken care of in ART).

ART natively supports multiple DEX files.

> After all, Proguard/R8 itself is such a big potential for spurious problems (**right now the current release on F-Droid is broken and I am struggling to find the issue in my proguard conf**) and pretty much the only reason to use code minimization through ProGuard is to avoid multidex (well, plus code obfuscation), because who cares about a few hundred KB of unused code?
> 
> **BUT** since 99% of the app is Kotlin anyway, instead of `Map.computeIfAbsent`, [`MutableMap.getOrPut`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-or-put.html) could be used. Would you like to create a PR for that?

Sure :)

> 
> Even though this is then not relevant anymore, I will read the article you linked in a moment, thanks!

Happy to help :)

westnordost:
The Java8 Date/Time API indeed looks good. In my current develop branch `resurvey-new`, I am using even more Date and Calendar to calculate differences in days, years, months etc, so after this is merged and after minSdk is Lollipop, thus only ART to worry about, using the Java8 Date/Time API could be put on the to do list fo migrating to minSdk Lollipop.

> ART natively supports multiple DEX files.

Ok, I was hoping that ART would have lifted the 64k limit instead of "just" supporting multiple DEX files.

westnordost:
You'll probably create a new PR for MutableMap.getOrPut, right? So I can close this one.

Isira-Seneviratne:
> 
> 
> The Java8 Date/Time API indeed looks good. In my current develop branch `resurvey-new`, I am using even more Date and Calendar to calculate differences in days, years, months etc, so after this is merged and after minSdk is Lollipop, thus only ART to worry about, using the Java8 Date/Time API could be put on the to do list fo migrating to minSdk Lollipop.
> 
> > ART natively supports multiple DEX files.
> 
> Ok, I was hoping that ART would have lifted the 64k limit instead of "just" supporting multiple DEX files.

Yeah, it compiles all the DEX files into a single `.oat` file.

westnordost:
By the way, how did you learn of the app StreetComplete? You are the first contributor from Sri Lanka, there is not even a translation of the app into siṁhala.

Isira-Seneviratne:
> 
> 
> By the way, how did you learn of the app StreetComplete? You are the first contributor from Sri Lanka, there is not even a translation of the app into siṁhala.

I saw it mentioned on Wikipedia's [article](https://en.wikipedia.org/wiki/OpenStreetMap) for OpenStreetMap.

westnordost:
Ah, okay. I was wondering whether it is a coincidence that over this weekend particularly, there were like 3-4 new contributors.

PR #2026, opened 2020-08-16
Use NotificationCompat.Builder instead of Notification.Builder.

Isira-Seneviratne:


Isira-Seneviratne:
Sure.

westnordost:
Nice, thank you!

PR #2029, opened 2020-08-17
Use KTX functions.

Isira-Seneviratne:
* Remove the AndroidX Core library, as its contents are included in the Core KTX library.
* Switch to the Fragment KTX and Preference KTX libraries.
* Make use of several [KTX extensions](https://developer.android.com/kotlin/ktx/extensions-list).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt
@@ -1,25 +1,25 @@
 package de.westnordost.streetcomplete.quests.max_speed
 
 import android.os.Bundle
-import androidx.annotation.IdRes
-import androidx.appcompat.app.AlertDialog
 import android.view.View
 import android.view.inputmethod.EditorInfo
 import android.widget.ArrayAdapter
 import android.widget.EditText
 import android.widget.ImageView
 import android.widget.Spinner
+import androidx.annotation.IdRes
+import androidx.appcompat.app.AlertDialog
 import androidx.core.view.children
-
+import androidx.core.view.isGone
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.ktx.numberOrNull
 import de.westnordost.streetcomplete.quests.AbstractQuestFormAnswerFragment
 import de.westnordost.streetcomplete.quests.OtherAnswer
-import de.westnordost.streetcomplete.util.TextChangedWatcher
+import de.westnordost.streetcomplete.quests.max_speed.SpeedMeasurementUnit.KILOMETERS_PER_HOUR
+import de.westnordost.streetcomplete.quests.max_speed.SpeedMeasurementUnit.MILES_PER_HOUR
 import de.westnordost.streetcomplete.quests.max_speed.SpeedType.*
-import de.westnordost.streetcomplete.quests.max_speed.SpeedMeasurementUnit.*
+import de.westnordost.streetcomplete.util.TextChangedWatcher
 import kotlinx.android.synthetic.main.quest_maxspeed.*
-import java.lang.IllegalStateException
 
 
 class AddMaxSpeedForm : AbstractQuestFormAnswerFragment<MaxSpeedAnswer>() {
@@ -49,10 +49,9 @@ class AddMaxSpeedForm : AbstractQuestFormAnswerFragment<MaxSpeedAnswer>() {
         super.onViewCreated(view, savedInstanceState)
 
         val couldBeSlowZone = countryInfo.isSlowZoneKnown && POSSIBLY_SLOWZONE_ROADS.contains(osmElement!!.tags["highway"]!!)
-        zone.visibility = if (couldBeSlowZone) View.VISIBLE else View.GONE
+        zone.isGone = !couldBeSlowZone
 
         speedTypeSelect.setOnCheckedChangeListener { _, checkedId -> setSpeedType(getSpeedType(checkedId)) }
-
     }
 
     override fun onClickOk() {
@@ -87,7 +86,7 @@ class AddMaxSpeedForm : AbstractQuestFormAnswerFragment<MaxSpeedAnswer>() {
         speedInput?.addTextChangedListener(TextChangedWatcher { checkIsFormComplete() })
 
         speedUnitSelect = rightSideContainer.findViewById(R.id.speedUnitSelect)
-        speedUnitSelect?.visibility = if (speedUnits.size == 1) View.GONE else View.VISIBLE
+        speedUnitSelect?.isGone = speedUnits.size == 1
         speedUnitSelect?.adapter = ArrayAdapter(requireContext(), R.layout.spinner_item_centered, speedUnits)
         speedUnitSelect?.setSelection(0)
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/NoteDiscussionForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/NoteDiscussionForm.kt
@@ -8,6 +8,7 @@ import android.text.format.DateUtils
 import android.text.format.DateUtils.MINUTE_IN_MILLIS
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.view.isGone
 import androidx.fragment.app.add
 import androidx.fragment.app.commit
 import androidx.recyclerview.widget.LinearLayoutManager
@@ -140,22 +141,20 @@ class NoteDiscussionForm : AbstractQuestAnswerFragment<NoteAnswer>() {
             val userName = if (comment.user != null) comment.user.displayName else getString(R.string.quest_noteDiscussion_anonymous)
 
             val commentActionResourceId = comment.action.actionResourceId
-            if (commentActionResourceId != 0) {
-                itemView.commentStatusText.visibility = View.VISIBLE
+            val hasNoteAction = commentActionResourceId != 0
+            itemView.commentStatusText.isGone = !hasNoteAction
+            if (hasNoteAction) {
                 itemView.commentStatusText.text = getString(commentActionResourceId, userName, dateDescription)
-            } else {
-                itemView.commentStatusText.visibility = View.GONE
             }
 
-            if (!comment.text.isNullOrEmpty()) {
-                itemView.commentView.visibility = View.VISIBLE
+            val hasComment = !comment.text.isNullOrEmpty()
+            itemView.commentView.isGone = !hasComment
+            if (hasComment) {
                 itemView.commentText.text = comment.text
                 itemView.commentInfoText.text = getString(R.string.quest_noteDiscussion_comment2, userName, dateDescription)
 
                 val bitmap = comment.user?.avatar ?: anonAvatar
                 itemView.commentAvatarImage.setImageBitmap(bitmap)
-            } else {
-                itemView.commentView.visibility = View.GONE
             }
         }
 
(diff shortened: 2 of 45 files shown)


westnordost:
Hui, that's a lot. This may take some time to review, I think I will go through it bit by bit.

The project used to be 100% Java, so some things may look quite Java-ish. My take on this is that I am quite indifferent if it looks Java-ish or Kotlin-ish as long as increases readability. That usually means I use Kotlin-like constructs because it is simply less verbose, but in some instances it might be too much because one has to consider that most devs are rather accostumed to Java than Kotlin, so if a construct looks too extranous I might have deliberately decided to do it the Java-way.

Anyway, will go through it, thank you very much for the effort!

app/src/main/java/de/westnordost/streetcomplete/view/ItemViewHolder.kt
```diff
@@ -41,7 +42,7 @@ class ItemViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
             textView?.text = null
 
         val descriptionId = item.descriptionId
-        descriptionView?.visibility = if (descriptionId != null) View.VISIBLE else View.GONE
+        descriptionView?.isVisible = descriptionId != null
```
westnordost, line 45: I am not sure if I am a fan of that extension function because it hides away the information between which states (visible / gone vs visible / invisible) it switches.

I had to look in the [documentation](https://developer.android.com/reference/kotlin/androidx/core/view/package-summary#isvisible) to see what it does exactly.  I think exclusively using `isInvisible` and `isGone` would be better because it is explicit between which states it alternates

app/src/main/java/de/westnordost/streetcomplete/settings/OAuthFragment.kt
```diff
@@ -164,7 +164,7 @@ class OAuthFragment : Fragment(R.layout.fragment_oauth),
 
         override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
             if (url == null) return false
-            val uri = Uri.parse(url) ?: return false
+            val uri = url.toUri()
```
westnordost, line 167: Hm right, this probably never worked. My intention here was that if the string is not a valid uri, it would just return false. Is there `String.toUriOrNull` or something?

Isira-Seneviratne: Nope.

app/src/main/java/de/westnordost/streetcomplete/view/ItemViewHolder.kt
Isira-Seneviratne: > I had to look in the [documentation](https://developer.android.com/reference/kotlin/androidx/core/view/package-summary#isvisible) to see what it does exactly. I think exclusively using `isInvisible` and `isGone` would be better because it is explicit between which states it alternates

Okay, I'll remove usages of `isVisible`.

app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/NoteDiscussionForm.kt
```diff
@@ -65,9 +67,7 @@ class NoteDiscussionForm : AbstractQuestAnswerFragment<NoteAnswer>() {
         inflateNoteDiscussion(osmNoteQuestController.get(questId)!!.note.comments)
 
         if (savedInstanceState == null) {
-            childFragmentManager.beginTransaction()
-                .add(R.id.attachPhotoFragment, AttachPhotoFragment())
-                .commit()
+            childFragmentManager.commit { add<AttachPhotoFragment>(R.id.attachPhotoFragment) }
```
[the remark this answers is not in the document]
Isira-Seneviratne: It actually calls one of the existing methods in `FragmentTransaction`.

Arguments can be passed via a `Bundle`.

Isira-Seneviratne:
I only made use of `isVisible` to retrieve whether or not a view is visible.

app/src/main/java/de/westnordost/streetcomplete/settings/OAuthFragment.kt
smichel17: You should be able to replace 166 and 167 with
```kotlin
val uri = url?.takeIf(::isValidUrl)?.toUri() ?: return false
```

https://developer.android.com/reference/android/webkit/URLUtil#isValidUrl(java.lang.String)

smichel17: (Of course, then you need to import `android.webkit.URLUtil`)

app/src/main/java/de/westnordost/streetcomplete/user/AchievementInfoFragment.kt
```diff
@@ -117,27 +118,25 @@ class AchievementInfoFragment : Fragment(R.layout.fragment_achievement_info) {
...
         } else {
-            achievementDescriptionText.visibility = View.GONE
             achievementDescriptionText.text = ""
         }
 
         val unlockedLinks = achievement.unlockedLinks[level].orEmpty()
-        if (unlockedLinks.isEmpty() || !showLinks) {
-            unlockedLinkTitleText.visibility = View.GONE
-            unlockedLinksList.visibility = View.GONE
+        val isGone = unlockedLinks.isEmpty() || !showLinks
```
westnordost, line 130: better `hasNoUnlockedLinks`

app/src/main/java/de/westnordost/streetcomplete/map/QuestsMapFragment.kt
```diff
@@ -184,8 +188,9 @@ class QuestsMapFragment : LocationAwareMapFragment() {
             else -> listOf(listOf(g.center))
         }.flatten().all {
             val isContained = controller.latLonToScreenPosition(it, p, false)
-            isContained && p.x >= offset.left && p.x <= mapView.width - offset.right
-              && p.y >= offset.top  && p.y <= mapView.height - offset.bottom
+            val (left, top, right, bottom) = offset
+            val (x, y) = p
+            isContained && x >= left && x <= mapView.width - right && y >= top && y <= mapView.height - bottom
```
westnordost, line 193: I think this is not an improvement. It's neither shorter nor easier to read.

Also, it introduces a potential for a bug: Who tells you that the first component is `left` etc? -> Only the documentation. So, this important detail is hidden away from the code, if this code wasn't correct, I wouldn't be able to see it in this review. This `component` stuff in kotlin is not very well thought through in my opinion, it's like using arrays rather than objects to store data, always a potential for bugs.

app/src/main/java/de/westnordost/streetcomplete/user/AchievementInfoFragment.kt
Isira-Seneviratne: Thanks!

app/src/main/java/de/westnordost/streetcomplete/map/QuestsMapFragment.kt
Isira-Seneviratne: I'll remove that particular commit, then.

app/src/main/java/de/westnordost/streetcomplete/settings/OAuthFragment.kt
Isira-Seneviratne: Thanks for the suggestion. I'll update the code.

smichel17: Actually, you can do the same with the lines after, which is another good readability increase, I think
```kotlin
val uri = url.takeIf(::isValidUrl)?.toUri().takeIf {
    it.isHierarchical && it.scheme == callbackScheme && it.host == callbackHost
} ?: return false
```
Although, personally control flow statements without brackets and newlines\* make me nervous, so I'd spend the vertical space for
```kotlin
if (uri == null) {
    return false
}
```
\*exception: one-liners like `return if (a) b else c`

smichel17: I suppose you could do
```kotlin
if (uri != null) {
    // verifier stuff
}
return uri == null
```
or if you don't give a damn about readability and only care about vertical space :rofl: 
```kotlin
return url.takeIf(::isValidUrl)?.toUri().takeIf {
    it.isHierarchical && it.scheme == callbackScheme && it.host == callbackHost
}?.also { uri -> continuation?.let { c ->
    uri.getQueryParameter(OAUTH_VERIFIER)?.let { verifier ->
        c.resume(verifier)
    } ?: c.resumeWithException(
        OAuthExpectationFailedException("$OAUTH_VERIFIER parameter not set by provider")
    )
}} != null
```

app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt
```diff
@@ -48,11 +48,9 @@ class AddMaxSpeedForm : AbstractQuestFormAnswerFragment<MaxSpeedAnswer>() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        val couldBeSlowZone = countryInfo.isSlowZoneKnown && POSSIBLY_SLOWZONE_ROADS.contains(osmElement!!.tags["highway"]!!)
-        zone.visibility = if (couldBeSlowZone) View.VISIBLE else View.GONE
+        zone.isGone = !countryInfo.isSlowZoneKnown || osmElement!!.tags["highway"]!! !in POSSIBLY_SLOWZONE_ROADS
```
westnordost, line 51: Mh, I find it better to have the `couldBeSlowZone` variable because it explains what the checks are made for. When assigning directly to `isGone`, it's a double negation and more difficult to understand.

The zone choice should only be shown if the street in question could be a slow zone.

app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/NoteDiscussionForm.kt
```diff
@@ -140,22 +141,20 @@ class NoteDiscussionForm : AbstractQuestAnswerFragment<NoteAnswer>() {
             val userName = if (comment.user != null) comment.user.displayName else getString(R.string.quest_noteDiscussion_anonymous)
 
             val commentActionResourceId = comment.action.actionResourceId
-            if (commentActionResourceId != 0) {
-                itemView.commentStatusText.visibility = View.VISIBLE
+            val isStatusGone = commentActionResourceId == 0
```
westnordost, line 144: actually, if `commentActionResourceId == 0`, it means that it is a normal note comment without any additional action like "closed note", "reopened note" or "hidden note". I suggest renaming it to !`hasNoteAction`

app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/NoteDiscussionForm.kt
```diff
@@ -140,22 +141,20 @@ class NoteDiscussionForm : AbstractQuestAnswerFragment<NoteAnswer>() {
...
+            itemView.commentStatusText.isGone = isStatusGone
+            if (!isStatusGone) {
                 itemView.commentStatusText.text = getString(commentActionResourceId, userName, dateDescription)
-            } else {
-                itemView.commentStatusText.visibility = View.GONE
             }
 
-            if (!comment.text.isNullOrEmpty()) {
-                itemView.commentView.visibility = View.VISIBLE
+            val isCommentGone = comment.text.isNullOrEmpty()
```
westnordost, line 150: actually, it means that there is no comment. For example, if you just resolve an OSM note without providing an additional comment. So I suggest to rename it !`hasComment`

westnordost:
Ok, all files reviewed, no more comments!

Isira-Seneviratne:
> 
> 
> Ok, all files reviewed, no more comments!

I hope my new changes are satisfactory :)

westnordost:
Yes! Looks very good, I think this can be merged now!

PR #2030, opened 2020-08-17
[WIP] Pedestrian Crossing Island Quest

### LINKED ISSUE #1961: Quest: Does this crossing have a pedestrian island?
### General

Affected tag(s) to be modified/added: [crossing:island](https://wiki.openstreetmap.org/wiki/Key:crossing:island)
Question asked: **Does this crossing have a pedestrian island?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose ([Taginfo](https://taginfo.openstreetmap.org/keys/?key=crossing%3Aisland) lists 99k items)
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam): I'm not sure about this one. I think most crossings won't have a pedestrian island, but quite a few do. That might vary by region though.

### Ideas for implementation

**Element selection:**
`nodes with crossing and crossing != "island" and !crossing:island`

**Proposed GUI:**
Simple Yes / No

### Related

* https://github.com/westnordost/StreetComplete/issues/495 (same quest suggestion, but with the older `crossing=island` tag, which doesn't prevent "no" as an answer to stop the quest from reappearing to other users)
* https://github.com/westnordost/StreetComplete/issues/1637 / https://github.com/westnordost/StreetComplete/commit/a609da8c7f0a51c24f6ca5fdd06b0b361b7cf8ec (include `crossing=island` in crossing quest, and set `crossing:island` to `yes` accordingly)

kmpoppe:
Resolves #1961.
This is work in progress. Reason: The filter is so generic, that the number of quests feel spam-y (picture follows).
I'd need help in improving this.
Thanks!
K


kmpoppe:
Screenshot_20200817-180249
This is how a larger traffic crossing looks with the current filter.

matkoniecz:
I would look at real data, I feel that it is happening far more often on bigger roads than on crossings through `highway=residential` or `highway=service`.

This picture reveals one more issue - how it should be tagged if island is marked explicitly and there are two separate `highway=crossing`?

westnordost:
Nice graphic! Don't forget to update the SVG to include the icon as well as add yourself as an author of that icon in the authors.txt

kmpoppe:
> Nice graphic! Don't forget to update the SVG to include the icon as well as add yourself as an author of that icon in the authors.txt

Thanks, just took the existing crossing one, made it SVG with ShapeShifter, did some editing with Gravit Designer online, and ShapeShifted it back to Android... Not much authoring there that's to be credited, is there?

Which SVG?

kmpoppe:
> I would look at real data, I feel that it is happening far more often on bigger roads than on crossings through `highway=residential` or `highway=service`.

I mean that IS real data, it's a crossing right outside my quarter...

> Maybe excluding also unmarked crossings?

I also thought excluding `unmarked` and `uncrontrolled` could be an any, but then again, quite rightly, the Wiki says that even those could have an island. I think there we'd have to decide what users should do (check for islands on zebras or marked ones) and leave the rest to be it?

> This picture reveals one more issue - how it should be tagged if island is marked explicitly and there are two separate `highway=crossing`?

Interesting issue 🤔

matkoniecz:
> I mean that IS real data, it's a crossing right outside my quarter...

I meant data overall, to avoid asking this quest in places where island is very unlikely. This specific crossing may be just very bad and unavoidable case.

westnordost:
> SVG

`res/quest_icons.svg`, that's the source file for all those quest icons.

> quest spam

This is indeed a problem, especially cases where the island is a separate way, thus the question itself is invalid. Question: If it is mapped in detail like this, should `island` not be tagged, or `island=no` or something else? If `island` should not be tagged for these cases, any idea how to filter these out? Will such crossings mapped in more detail have any other tags?

matkoniecz:
> Will such crossings mapped in more detail have any other tags?

Maybe skipping crossing on oneway roads would be way to get rid of them (and also of some valid quests)?

kmpoppe:
> Maybe skipping crossings on oneway roads would be way to get rid of them (and also of some valid quests)?

The good Markus Göllnitz provided me with an example from Overpass that'll give you the way a node is on: https://overpass-turbo.eu/s/X8E
I'm guessing that that's not implemented in SC as of yet, so some more work would be needed there I think

matkoniecz:
>  I'm guessing that that's not implemented in SC as of yet, so some more work would be needed there I think

See level crossing barriers quest - https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/railway_crossing/AddRailwayCrossingBarrier.kt (though maybe there is a better way than skipping all oneways)

westnordost:
I have another idea that would be more exclusive of such situations where the intersection is mapped in such detail that `island = yes/no` is irrelevant/unwanted:
Only show quests for `highway = crossing` nodes on roads that do not cross with another way (are an intersection of footway with road), because that is a strong indicator if not a surefire way of finding out that the road+footways have not been mapped the way as depicted in the screenshot.

westnordost:
`SimpleOverpassQuestType` is a shortcut (abstract class) that makes it easy so that the query can be written in the element filter syntax. But the actual interface you need to implement is OsmElementQuestType, which gives you complete freedom in what you do in the `download` method. Most just do an overpass query in overpass syntax.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
```diff
@@ -0,0 +1,29 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquest.SimpleOverpassQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.mapdata.OverpassMapDataAndGeometryApi
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+class AddCrossingIsland(o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestType<Boolean>(o) {
+
+    override val tagFilters = """
+        nodes, ways with
+          crossing
```
westnordost, line 13: Should be `highway = crossing`?

Atrate: Do islands exists when railway crossings are considered? If not, then yeah, highway=crossing would be more appropriate.

(I personally have not seen islanda on railway crossings, but maybe that's just me)

kmpoppe: Yeah ... I just took what was in the Issue - taking a closer look, `highway=crossing` is the root tag.

kmpoppe:
> > SVG
> 
> `res/quest_icons.svg`, that's the source file for all those quest icons.

I might be overlooking something. What's a working way to incorporate the SVGs in one another?

kmpoppe:
The resident Overpass 🧙‍♂️ made this: https://overpass-turbo.eu/s/X8K - could you'll check that in your area if the result makes sense? (It does in my homezone).

matkoniecz:
What is the reason for `way.crossedWays["highway"~"^(path|footway|pedestrian)$"];` ?

At least in my region where footways are mapped separately it basically eliminates this quest (Kraków, Poland).

camelCaseNick:
> What is the reason for `way.crossedWays["highway"~"^(path|footway|pedestrian)$"];` ?
> 
> At least in my region where footways are mapped separately it basically eliminates this quest (Kraków, Poland).

Didn't you answer your question yourself? It filters out those crossing where such a `highway~^(path|footway|pedestrian)$` exists.

kmpoppe:
> What is the reason for `way.crossedWays["highway"~"^(path|footway|pedestrian)$"];` ?

Markus and I believe that's what Tobias said: "Only show quests for `highway = crossing` nodes on roads **that do not cross with another way (are an intersection of footway with road)**, because that is a strong indicator if not a surefire way of finding out that the road+footways have not been mapped the way as depicted in the screenshot."

camelCaseNick:
> The resident Overpass mage_man made this: https://overpass-turbo.eu/s/X8K - could you'll check that in your area if the result makes sense? (It does in my homezone).
> Edit: https://overpass-turbo.eu/s/X8O

Note that the former also gets rid of crossing at junctions, the latter doesn't.

westnordost:
You mean at interjunctions of roads? Do you expect any `highway=crossing` there?

camelCaseNick:
> You mean at interjunctions of roads? Do you expect any `highway=crossing` there?

No, but as I had no reason to expect to _not_ find them there and have not checked whether it is common enough to get annoying, I thought I'd rather inform you.

kmpoppe:
image
Junction with crossing ... don't know whether it's really tagged that way but it COULD exist

westnordost:
So, then this is fine

```
node["highway"="crossing"]["crossing"]["crossing"!="island"][!"crossing:island"] -> .crossings;
.crossings < -> .crossedWays;

(
  way.crossedWays["highway"]["access"~"^(private|no)$"];
  way.crossedWays["highway"]["oneway"!="no"];
  way.crossedWays["highway"~"^(path|footway|cycleway|pedestrian)$"];
) -> .excludedWays;

( .crossings; - node(w.excludedWays); );
```

kmpoppe:
Changed the Quest to `OsmElementQuestType` - now we're talking. Non spam, but still results that seem to make sense.
Something I still need to do: Get the new icon into `res/quest_icons.svg`. I have no idea on how to do that correctly.

westnordost:
I usually draw the icon with inkscape in quest_icons.svg, save it, then copy only the new icon into a new file and import it using Android Studio.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
```diff
@@ -0,0 +1,58 @@
...
+    override fun isApplicableTo(element: Element): Boolean? = null
+
+    override fun download(bbox: BoundingBox, handler: (element: Element, geometry: ElementGeometry?) -> Unit): Boolean {
+        return overpass.query(getOverpassQuery(bbox), handler)
+    }
+
+    private fun getOverpassQuery(bbox: BoundingBox): String {
+        val accessFilter = "\"access\"~\"^(private|no)\$\""
+        val onewayFilter = "\"oneway\"!=\"no\""
+        val pathFilter = "\"highway\"~\"^(path|footway|cycleway|pedestrian)\$\""
```
westnordost, line 33: I don't think this increases readability, especially since you have to escape all the `"` there as well.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
```diff
@@ -0,0 +1,58 @@
...
+
+        return bbox.toGlobalOverpassBBox() + """
+            node
+              ["highway" = "crossing"]
+              ["crossing"]
+              ["crossing" != "island"]
+              [!"crossing:island"] -> .crossings;
```
westnordost, line 40: Those that are used to reading OQL statements are used to reading these filters in one line if they fit. But it's fine if you want to leave it that way

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
westnordost: Also, why did you escape the dollar? I think this is a mistake

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
westnordost: Also, it is allowed in OQL to leave out the `"` if the key/value only consists of letters

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
kmpoppe: Well. It works with escaping and it threw an exception (due to an error vom overpass) without them.
I'll just reintrage them in the OQL string.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
kmpoppe: Again, will revert.

westnordost:
Still working on it?

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
xuiqzy: I also don't think the variables in the current form improve readability, but you can at least use triple quotes (for a string literal in kotlin) there too (like in the return) to get rid of the need to escape the quotes.

westnordost:
Do you mind if I finish this? It's just a tiny bit of work to finish it, I don't like leaving around almost finished work

kmpoppe:
Certainly, feel free. I'm currently buried in work stuff (sadly, none of that pays me for doing OSM Work) so now I can't spend so much time. Thanks for putting the work into it (reads: if you want to finish the memorial quest, that's also fine!)

westnordost:
Ok, I'll finish only this PR for now. Yeah, I am super happy to have gotten this grant :-)

westnordost:
Ok, done. There was a little mistake in the OQL that I didn't notice in a previous review.

PR #2031, opened 2020-08-18
Make changes for when the minSdk is set to 21.

Isira-Seneviratne:


westnordost:
Note that I will probably not review this for some time. Here is some more things that could be changed:
- remove resources only made for versions below 21
- replace certain Compat-functions and only used for compatibility with below 21, for example `Locale.toBcp47LanguageTag`. Some may be marked with "LOLLIPOP" somewhere in the comment
- replace/remove compat libraries used only for versions below 21, for example `androidx.work` could probably be replaced

app/src/main/java/de/westnordost/streetcomplete/controls/QuestDownloadProgressFragment.kt
```diff
@@ -98,12 +100,13 @@ class QuestDownloadProgressFragment : Fragment(R.layout.fragment_quest_download_
             .start()
     }
 
+    @SuppressLint("UseCompatLoadingForDrawables")
     private fun updateDownloadProgress() {
         if (downloadProgressSource.isDownloadInProgress) {
             showProgressView()
             val questType = downloadProgressSource.currentDownloadingQuestType
             if (questType != null) {
-                progressView.setIcon(resources.getDrawable(questType.icon))
```
westnordost, line 106: What's wrong with `resources`?

Isira-Seneviratne: `Resources.getDrawable()` has been deprecated in favor of `Context.getDrawable()`.

westnordost: That's really weird, isn't the Resources class the thing to access any resources? Are drawables not resources anymore?

Isira-Seneviratne: My bad, it's deprecated in favor of `Resources.getDrawable(int, Theme)`.

app/src/main/java/de/westnordost/streetcomplete/ktx/Locale.kt
```diff
@@ -11,10 +11,6 @@ import java.util.Locale
  * @return String: The BCP 47 language tag for the current locale
  */
 fun Locale.toBcp47LanguageTag(): String? {
-    //if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-    //    return toLanguageTag()
-    //}
-
     var language = language
```
westnordost, line 14: Heh, you might want to look at that again...

app/src/main/java/de/westnordost/streetcomplete/user/QuestStatisticsByQuestTypeFragment.kt
```diff
@@ -62,11 +63,9 @@ class QuestStatisticsByQuestTypeFragment : Fragment(R.layout.fragment_quest_stat
         val clickableContainer = FrameLayout(ctx)
         clickableContainer.layoutParams = ViewGroup.LayoutParams(256,256)
         // foreground attribute only exists on FrameLayout up until KITKAT
```
westnordost, line 65: missed to remove comment. Suggest to Ctrl+F for Kitkat, Ice Cream Sandwich and so on

app/src/main/java/de/westnordost/streetcomplete/user/QuestStatisticsByQuestTypeFragment.kt
```diff
@@ -51,6 +51,7 @@ class QuestStatisticsByQuestTypeFragment : Fragment(R.layout.fragment_quest_stat
         })
     }
 
+    @SuppressLint("UseCompatLoadingForDrawables")
```
westnordost, line 54: Why?

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadNameForm.kt
```diff
@@ -56,7 +51,6 @@ class AddRoadNameForm : AAddLocalizedNameForm<RoadNameAnswer>() {
     override fun onClickOk(names: List<LocalizedName>) {
         val possibleAbbreviations = LinkedList<String>()
         for ((languageTag, name) in adapter.localizedNames) {
-            // if Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP was, could use Locale.forLanguageTag
             val locale = if (languageTag.isEmpty()) countryInfo.locale else Locale(languageTag)
```
westnordost, line 54: What about this?

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddLocalizedNameAdapter.kt
```diff
@@ -137,7 +137,6 @@ class AddLocalizedNameAdapter(
         if (languageTag == "international") return context.getString(R.string.quest_streetName_menuItem_international)
         val languageCode = languageTag.substringBefore('-')
         val isRomanization = languageTag.endsWith("Latn")
-        // if Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP was, could use Locale.forLanguageTag
         val locale = Locale(languageCode)
```
westnordost, line 140: What about this?

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.java
```diff
@@ -46,15 +34,9 @@ public static boolean hasLocationPermission(Context context)
 
 	public static IntentFilter createLocationAvailabilityIntentFilter()
 	{
-		String action = LocationUtil.isNewLocationApi() ? MODE_CHANGED : PROVIDERS_CHANGED_ACTION;
-		return new IntentFilter(action);
+		return new IntentFilter(MODE_CHANGED);
 	}
 
 	// because LocationManager.MODE_CHANGED is not defined before KitKat
 	private static final String MODE_CHANGED = "android.location.MODE_CHANGED";
```
westnordost, line 41: I see more that can be removed here.

Much of `LocationUtil` is a compatibility wrapper

app/src/main/java/de/westnordost/streetcomplete/user/QuestStatisticsByQuestTypeFragment.kt
Isira-Seneviratne: A warning that `ResourcesCompat.getDrawable()` should be used is displayed.

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.java
Isira-Seneviratne: What could be removed?

app/src/main/java/de/westnordost/streetcomplete/ktx/Locale.kt
Isira-Seneviratne: My bad.

app/src/main/java/de/westnordost/streetcomplete/user/QuestStatisticsByQuestTypeFragment.kt
Isira-Seneviratne: Thanks!

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.java
Isira-Seneviratne: I created another PR that removes some of the methods in `LocationUtil`.

app/src/main/java/de/westnordost/streetcomplete/user/QuestStatisticsByQuestTypeFragment.kt
westnordost: And why silence the warning?

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.java
westnordost: I think that PR was wrong

app/src/main/java/de/westnordost/streetcomplete/controls/QuestDownloadProgressFragment.kt
westnordost: And why not use `Resources.getDrawable(in, Theme)` then?

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddLocalizedNameAdapter.kt
```diff
@@ -137,8 +137,7 @@ class AddLocalizedNameAdapter(
         if (languageTag == "international") return context.getString(R.string.quest_streetName_menuItem_international)
         val languageCode = languageTag.substringBefore('-')
         val isRomanization = languageTag.endsWith("Latn")
-        // if Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP was, could use Locale.forLanguageTag
-        val locale = Locale(languageCode)
+        val locale = Locale.forLanguageTag(languageCode)
```
westnordost, line 141: I think this is wrong, or at least, incomplete. Look at 
https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-

It takes a language tag like for example `si-Latn`. The information whether something is a romanization (or another transliteration into another script) can be accessed from `locale.script` and `locale.getDisplayScript()`. These functions are only available since LOLLIPOP.
Furthermore, with LOLLIPOP, the huge if-else can be simplified, because we no longer need `R.string.quest_streetName_menuItem_romanized` etc but can just use `locale.getDisplayScript()`

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddRoadNameForm.kt
```diff
@@ -56,8 +51,7 @@ class AddRoadNameForm : AAddLocalizedNameForm<RoadNameAnswer>() {
     override fun onClickOk(names: List<LocalizedName>) {
         val possibleAbbreviations = LinkedList<String>()
         for ((languageTag, name) in adapter.localizedNames) {
-            // if Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP was, could use Locale.forLanguageTag
-            val locale = if (languageTag.isEmpty()) countryInfo.locale else Locale(languageTag)
+            val locale = Locale.forLanguageTag(languageTag)
             val abbr = abbreviationsByLocale.get(locale)
```
westnordost, line 55: What if `languageTag `is empty? -> Error

app/src/main/java/de/westnordost/streetcomplete/user/QuestStatisticsByQuestTypeFragment.kt
westnordost: Or asked another way: Why is there a warning for using `ResourcesCompat.getDrawable()`? There is also a warning for using `Resources.getDrawable`, why not silence this instead, or just leave it as it is?

app/src/main/java/de/westnordost/streetcomplete/quests/localized_name/AddLocalizedNameAdapter.kt
Isira-Seneviratne: Thanks, I'm not very familiar with the `Locale` class.

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.java
Isira-Seneviratne: I added a comment there.

westnordost:
Another thing that comes to my mind would be that in pre-Lollipop, referencing colors in vector drawables is not (really) possible. For many icons, there is a day and a night variant because of this. If the min-sdk is lollipop, instead the drawable can reference a color instead, so the drawables don't need to be duplicated. Example `ic_building_toilets.xml`

Isira-Seneviratne:
> Another thing that comes to my mind would be that in pre-Lollipop, referencing colors in vector drawables is not (really) possible. For many icons, there is a day and a night variant because of this. If the min-sdk is lollipop, instead the drawable can reference a color instead, so the drawables don't need to be duplicated. Example `ic_building_toilets.xml`

Sorry for the late reply. I'll make the change, thanks!

westnordost:
I think a color value does not exist for it yet, as this is just one of many colors used in the vector drawables.

No issue with the late reply. After all, this PR is going to be around for a while, maybe 2 more months. And even if you abandoned this PR now, it still is valuable for the project because then later I can finish it when finally the minsdk is set to 21.

westnordost:
Sorry, my original plan was to up the min sdk version end last year because I intended to couple that with the change to add "measurement with AR"-feature. That feature wasn't developed so far after all (and thus requiring camera permission), so that's why there was no good reason to up the min sdk version yet.

So it looks like this is going to lie around a while longer, but I will get back to it eventually. No need to bother fixing the conflicts, I will do that when the time comes.

westnordost:
Finally, one year late, I decided to finally drop support for Android 4. Thank you for your work so far on it, Isira. Quite a few things have changed in the app since you last brought this PR up to master, so there are some conflicts to solve.

I will mark this PR as "help wanted" in case other contributors to this app would like to finish up this PR (that is: solve merge conflict, check if everything still works) unless you are interested in bringing it to completion (again).

If noone else will do it, I will finish up this PR, so your work on it will merged anyway and you'll be credited for your work of course.

westnordost:
Alright, this is brought up to current master plus all kinds of optimizations done that are only available since v21. Mostly, referencing color values in vector drawables makes it possible to delete all those resources for night.

westnordost:
The resulting APK's size is 45 MB, so about 30 MB lighter than now.

PR #2033, opened 2020-08-19
Use Compat classes.

Isira-Seneviratne:
Make use of more of Jetpack's `Compat` classes, such as [`HtmlCompat`](https://developer.android.com/reference/androidx/core/text/HtmlCompat) and [`ConfigurationCompat`](https://developer.android.com/reference/kotlin/androidx/core/os/ConfigurationCompat), to avoid deprecation warnings with `Html`, `Configuration` and `TextView`.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
@@ -3,11 +3,11 @@ package de.westnordost.streetcomplete.quests
 import android.content.res.Resources
 import android.text.Html
 import android.text.Spanned
-
+import androidx.core.text.HtmlCompat
 import de.westnordost.osmapi.map.data.Element
 import de.westnordost.osmfeatures.FeatureDictionary
-import de.westnordost.streetcomplete.data.quest.QuestType
 import de.westnordost.streetcomplete.data.osm.osmquest.OsmElementQuestType
+import de.westnordost.streetcomplete.data.quest.QuestType
 import java.util.*
 import java.util.concurrent.FutureTask
 
@@ -19,7 +19,8 @@ fun Resources.getQuestTitle(questType: QuestType<*>, element: Element?, featureD
 fun Resources.getHtmlQuestTitle(questType: QuestType<*>, element: Element?, featureDictionaryFuture: FutureTask<FeatureDictionary>?): Spanned {
     val arguments = getTemplateArguments(questType, element, configuration.locale, featureDictionaryFuture)
     val spannedArguments = arguments.map {"<i>" + Html.escapeHtml(it) + "</i>"}.toTypedArray()
-    return Html.fromHtml(getString(getQuestTitleResId(questType, element), *spannedArguments))
+    return HtmlCompat.fromHtml(getString(getQuestTitleResId(questType, element), *spannedArguments),
+        HtmlCompat.FROM_HTML_MODE_LEGACY)
 }
 
 private fun getTemplateArguments(
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
@@ -4,6 +4,7 @@ import android.content.res.Resources
 import android.text.Html
 import android.text.Spanned
 import androidx.core.os.ConfigurationCompat
+import androidx.core.os.LocaleListCompat
 import androidx.core.text.HtmlCompat
 import de.westnordost.osmapi.map.data.Element
 import de.westnordost.osmfeatures.FeatureDictionary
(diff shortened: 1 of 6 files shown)


app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
```diff
@@ -100,13 +101,13 @@ class LocationStateButton @JvmOverloads constructor(
...
             state = LocationState.valueOf(parcel.readString()!!)
-            activated = parcel.readInt() == 1
+            activated = ParcelCompat.readBoolean(parcel)
         }
 
         override fun writeToParcel(out: Parcel, flags: Int) {
             super.writeToParcel(out, flags)
             out.writeString(state.name)
-            out.writeInt(if (activated) 1 else 0)
+            ParcelCompat.writeBoolean(out, activated)
```
westnordost, line 110: I find this rather unnecessary

app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
```diff
@@ -3,23 +3,27 @@ package de.westnordost.streetcomplete.quests
...
-import de.westnordost.streetcomplete.data.quest.QuestType
 import de.westnordost.streetcomplete.data.osm.osmquest.OsmElementQuestType
+import de.westnordost.streetcomplete.data.quest.QuestType
 import java.util.*
 import java.util.concurrent.FutureTask
 
 fun Resources.getQuestTitle(questType: QuestType<*>, element: Element?, featureDictionaryFuture: FutureTask<FeatureDictionary>?): String {
-    val arguments = getTemplateArguments(questType, element, configuration.locale, featureDictionaryFuture)
+    val locale = ConfigurationCompat.getLocales(configuration)[0]
+    val arguments = getTemplateArguments(questType, element, locale, featureDictionaryFuture)
```
westnordost, line 17: Well actually, the deprecation of `configuration.locale` is there because one is supposed to take into account that the user may have a chain of locales to apply - if a resource is not found for the first locale, then the second locale should be used, and so forth.

So this change only silences the warning but doesn't solve the problem, `getTemplateArguments` would need to accept a list of locales and go through all the given locales until one name has been found.

westnordost:
All else is fine, thank you!

app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
Isira-Seneviratne: In that case, `getTemplateArguments` could be changed to accept `LocaleListCompat` instead.

westnordost: Yes, and do some for loop or something within the lazy until we reached the end of the LocaleList or a name was found

westnordost: (maybe outsource that logic from the lazy into an own function, might get a little hard to read otherwise)

app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
```diff
@@ -29,16 +30,35 @@ fun Resources.getHtmlQuestTitle(questType: QuestType<*>, element: Element?, feat
...
+        }
+    }
+    return null
+}
+
+private inline fun LocaleListCompat.forEach(action: (Locale) -> Unit) {
+    for (i in 0 until size()) {
+        action(this[i])
+    }
+}
```
westnordost, line 61: Wow sheesh, Google devs did not even implement the `Iterable` interface?

westnordost: Anyway, looks good!

Maybe since this is actually a new implementation, you could test this once? UnitTest would of course be supreme, but not required, there wasn't any before anyway.

westnordost: The method is called for the "What is the name of this place?" quest.

Isira-Seneviratne: > 
> 
> Wow sheesh, Google devs did not even implement the `Iterable` interface?

Yeah, it seems to be the case with the actual [`LocaleList`](https://developer.android.com/reference/android/os/LocaleList.html) as well.

Isira-Seneviratne: > 
> 
> Anyway, looks good!
> 
> Maybe since this is actually a new implementation, you could test this once? UnitTest would of course be supreme, but not required, there wasn't any before anyway.

Sure.

Isira-Seneviratne: There isn't any visible difference, as far as I can tell.

Screenshot_20200819-184829226

Isira-Seneviratne: On another note, you might find this important.

Screenshot_20200819-185237261

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
smichel17: I found it to be an increase in readability (it's nice to have things typed correctly). If you wanted to keep a the same form as the others, you could write extension functions `Parcel.readBooleanCompat()` and `Parcel.writeBooleanCompat()` (if you do this, don't use `readBoolean` and `writeBoolean` as names because I believe extension functions will not override already existing functions — or you could add `Boolean.toInt()` and `Int.toBoolean()` functions instead).

app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt
westnordost: Hm, don't know. Looks like `FineLocationManager.removeUpdates()` is not called, but it is (on onPause and onStop)

Perhaps the `@OnLifecycleEvent` annotations don't work as they are supposed to(?)

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
westnordost: Thing is, you don't know (without consulting the documentation) what `readBoolean` will do - there are no boolean types in a parcel - so will it write a string "true" /"false", or an int, short or whatever else? Or will it understand fuzzily everything that is "truthy" (like JS, YAML, PHP,... etc)? The `Boolean.toInt()` approach would be more clear here, but it is such a short statement that I think it is not really worth it.

westnordost (approved):

westnordost:
So, looks good to me

PR #2038, opened 2020-08-21
Use more KTX extensions.

Isira-Seneviratne:
* Use the [`String.parseAsHtml()`](https://developer.android.com/reference/kotlin/androidx/core/text/package-summary#(kotlin.String).parseAsHtml(kotlin.Int,%20android.text.Html.ImageGetter,%20android.text.Html.TagHandler)) extension function instead of `HtmlCompat.fromHtml()`.
* Use the [`lifecycleScope`](https://developer.android.com/reference/kotlin/androidx/lifecycle/package-summary#(androidx.lifecycle.LifecycleOwner).lifecycleScope:androidx.lifecycle.LifecycleCoroutineScope) extension property.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
@@ -16,8 +16,8 @@ import android.widget.PopupMenu
 import androidx.annotation.AnyThread
 import androidx.appcompat.app.AlertDialog
 import androidx.core.os.bundleOf
+import androidx.core.text.parseAsHtml
 import androidx.core.view.isGone
-import androidx.core.text.HtmlCompat
 import com.google.android.flexbox.FlexboxLayout
 import de.westnordost.osmapi.map.data.OsmElement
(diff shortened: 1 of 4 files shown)


westnordost:
Then I don't understand how `lifecycleScope` works. How is the scope attached to the lifecycle of the lifecycle owner? Could it be that the `lifecycleScope` needs to be attached to the lifecycle owner's lifecycle manually in the constructor? If yes, I don't see much of a advantage here, that's one line of boilerplate, exactly the same as in the current solution. If not, that sounds like black magic.

westnordost:
Usage of parseAsHtml though is fine

Isira-Seneviratne:
> Then I don't understand how `lifecycleScope` works. How is the scope attached to the lifecycle of the lifecycle owner? Could it be that the `lifecycleScope` needs to be attached to the lifecycle owner's lifecycle manually in the constructor? If yes, I don't see much of a advantage here, that's one line of boilerplate, exactly the same as in the current solution. If not, that sounds like black magic.

In that case, shall I revert that commit?

westnordost:
Yes please

westnordost:
Top, thank you!

PR #2049, opened 2020-08-25
Use LocationManagerCompat.isLocationEnabled().

Isira-Seneviratne:
Make use of [`LocationManagerCompat.isLocationEnabled()`](https://github.com/androidx/androidx/blob/abb30e5fd0f6b6905e290087020a979655824050/core/core/src/main/java/androidx/core/location/LocationManagerCompat.java#L58).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -1,25 +1,18 @@
 package de.westnordost.streetcomplete;
 
-import android.content.res.Configuration;
-import android.graphics.Point;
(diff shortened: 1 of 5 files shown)


westnordost:
Sorry, this PR is wrong.

`LocationUtil.isLocationOn` does something different than `LocationManagerCompat.isLocationEnabled`, just look at the implementation. It is completely different.

Isira-Seneviratne:
`Settings.Source.LOCATION_MODE` is deprecated starting with API level 28, and it recommends to use `LocationManager.isEnabled()` instead: https://developer.android.com/reference/android/provider/Settings.Secure#LOCATION_MODE

PR #2055, opened 2020-08-28
Use the vector drawable support library.

Isira-Seneviratne:
* Make use of the [vector drawable support library](https://developer.android.com/guide/topics/graphics/vector-drawable-resources#vector-drawables-backward-solution) to significantly reduce app size (a full release build before this change was ~60 MB and ~30 MB after the change).
* This also allows backported improvements for vector drawables to be used on API levels 21-23.


westnordost:
> Make use of the vector drawable support library to significantly reduce app size (a full release build before this change was ~60 MB and ~30 MB after the change).

Why though? Starting API 21, most android vector drawables should not be converted to pixel graphics by the build tools anymore.

> This also allows backported improvements for vector drawables to be used on API levels 21-23.

What exactly? Maybe it is not relevant for this app.

westnordost:
Oh, this is to be merged not for after upgrading to API 21 but now. Well, I don't think it is worth it then since upgrading to 21 is on the menu anyway.

Isira-Seneviratne:
> > Make use of the vector drawable support library to significantly reduce app size (a full release build before this change was ~60 MB and ~30 MB after the change).
> 
> Why though? Starting API 21, most android vector drawables should not be converted to pixel graphics by the build tools anymore.
> 
> > This also allows backported improvements for vector drawables to be used on API levels 21-23.
> 
> What exactly? Maybe it is not relevant for this app.

Google [recommends](https://youtu.be/fgbl34me3kk?t=665) using the support library even on Lollipop and higher.

westnordost:
Mh interesting, this would actually take the biggest issue I have with Android < LOLLIPOP support away.

I though I heard something about that there is a problem with app compat support for vector drawables and thus, having the build system convert all the vector drawables to raster graphics for API < LOLLIPOP would be the better/safer solution. Though, this information is a few years old. Could it be that this app compat library did not already exist a few years back?

Isira-Seneviratne:
This should be helpful.

westnordost:
Hmm, but we use vector drawables in things like `drawableLeft` etc

Isira-Seneviratne:
> Hmm, but we use vector drawables in things like `drawableLeft` etc

There are tags like `drawableLeftCompat` that can be used instead.

westnordost:
Ok, so the graph mentions a post by Chris Banes on Medium for details. I didn't go trhough it right now, but it is from 2016. Maybe the info on it is outdated?

Isira-Seneviratne:
> Ok, so the graph mentions a post by Chris Banes on Medium for details. I didn't go trhough it right now, but it is from 2016. Maybe the info on it is outdated?

I think only some info is outdated. For instance, [`VectorDrawableCompat`](https://developer.android.com/reference/androidx/vectordrawable/graphics/drawable/VectorDrawableCompat) delegates to the framework `VectorDrawable` on Android Nougat and higher.

westnordost:
So this is will-not-fix / obsolete once the v21 branch is merged.

PR #2061, opened 2020-09-01
Use variables for multi-dependencies

daveol:
Some dependencies like kotlinx, mockito and dagger have multiple
dependencies defined which should be the same version. This ensures they
are kept the same by using a variable to define the version for these
dependencies.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -84,6 +84,12 @@ configurations {
     compile.exclude group: 'org.intellij', module:'annotations'
 }
 
+ext {
+    mockito_version = '2.28.2'
+    kotlinx_version = '1.3.8'
+    dagger_version = '2.14.1'
+}
+
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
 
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -92,18 +98,18 @@ dependencies {
...
     // tests
     testImplementation 'junit:junit:4.12'
-    testImplementation 'org.mockito:mockito-core:2.28.2'
-    testImplementation 'org.mockito:mockito-inline:2.28.2'
+    testImplementation "org.mockito:mockito-core:$mockito_version"
+    testImplementation "org.mockito:mockito-inline:$mockito_version"
     testImplementation 'org.assertj:assertj-core:2.8.0'
 
     androidTestImplementation 'androidx.test:runner:1.2.0'
     androidTestImplementation 'androidx.test:rules:1.2.0'
```
daveol, line 106: Should we also define `$androidx_test_version` to catch these?

westnordost:
The Android Studio IDE shows updatable libraries as "warnings", thus will also show libraries with the same version number that they can be updated now.

If variables are introduced, IIRC this detection does not work anymore.

daveol:
looking at both the file and project structure yields a message about updating the dependency in Android Studio.
image
image

westnordost:
Oh, okay then, then this can be merged.

Issue #561, opened 2017-08-28
New quest: Peak features


ServusWorld:
# Where it shows up
`natural=peak` where either the `summit:cross=*` or the `summit:register=*` tag is missing (or both).

# Form
![form_peak](https://user-images.githubusercontent.com/31391499/29782252-266322ae-8c1c-11e7-8a51-e28e452f5097.png)

Two checkboxes with icons:
- [Summit register](https://www.flaticon.com/free-icon/open-book_174230)
- [Summit cross](https://www.flaticon.com/free-icon/crucifixion_263283)

(Actually, the form must be initialised so that the corresponding checkbox is already checked if one of the quested tags already exists, but it's no problem if it ignores it, then the user keeps the information latest :))

# What will be added
If the user checked "Summit cross", `summit:cross=yes` is added, else `summit:cross=no`
If the user checked "Summit register", `summit:register=yes` is added, else `summit:register=no`

# Other answers
_(no ideas so far ...)_

matkoniecz:
Why with checkboxes rather than using style used for marking matching sports for leisure=pitch quest?

ServusWorld:
Well, that's true.

matkoniecz:
Main problem here is that vast majority of peaks have neither first nor second feature.

Is there way to select (by country?) peaks that are likely to have either of this features?

rugk:
@ServusWorld Could you please use the checkboxes as indicated in the issue template? (You can still edit your issue.)

ServusWorld:
Yes, I will do so.
(Not now, but not later than tomorrow.)

westnordost:
Would there reasonably be a summit register and/or cross to be expected on all these peaks here?:

http://www.openstreetmap.org/#map=12/50.9042/14.2863

Are there not more properties with which this could be contained?

ServusWorld:
At least, the quest fulfils "Applies to a reasonable number of elements (Worth the effort)" :)

Crosses maybe many of them, summit registers rather not ...

westnordost:
1. I think it should probably be limited to certain countries where summit crosses (what about non christian countries, do they have something else?) and registers are commonplace
2. Probably it makes sense to limit this to peaks that have a name (and/or recorded height?)

westnordost:
By the way, your icons for the mockups are nice. But generally, images or icons are only **necessary** when the text alone is not enough to convey what is being asked for. So, I think in this case (and the organic quest), nothing speaks against having simply two radio buttons.

[...]

PR #2065, opened 2020-09-03
Summit register (#561) [ready for review, icon added]

matkoniecz:


westnordost:
LGTM, only the icon is missing

matkoniecz:
Note that has "fixes #561" in commit but implements only summit register, as I think that summit crosses should be tagged as `man_made=cross`

Issue #1654, opened 2019-11-18
Add quest presets


Helium314:
In issue #1301 @westnordost suggested to open a new issue regarding quest profiles, i.e. different quest selections and priorities to select depending on the current situation.
Sorry if this issue is already somewhere around and I just didn't find it.

These profiles can be useful for switching between e.g. a "normal" and a "train" profile, where the train profile has no quests like road name or opening hours you can't solve from a train anyway. Another example would be "day" and "night" profiles, where the "is ... lit?" quest is disabled or low priority in the day profile and high priority at night.

In a [comment](https://github.com/westnordost/StreetComplete/issues/1301#issuecomment-451788128) to #1301 @westnordost asked for UI proposals of this feature.
So my idea is that profile selection could be in a menu that is invoked from an (to be added) point in the "..." menu. Or simply in the settings, either below quest selection, or part of it (profile selection dropdown on top, including a "new profile" point).

westnordost:
I think the current profile should always be visible for traceability, especially if the profile is selected automatically somehow.

Helium314:
I didn't mean the profile should be selected automatically. Using "day" and "night" was probably a bad example...

If the current profile should always be visible, I see no way around adding an additional line below the star and upload counts. In that case it should perhaps be possible to disable profiles to have the UI as it is now.

Helium314:
Some thoughts on how this could look:
(I keep calling this "profile", but in the end it should maybe have a different name to avoid confusion with the OSM profile)

In settings -> quest selection there is possibility to add a new profile. Either in the "..." menu or some extra button.
Adding new profiles asks which profile to copy from (or create a new one, which is like after resetting quests), and allows to enter a name (some "profile \<number\>" could be default)
Renaming and deleting profiles should be possible in the menu.

Where to put profile selection:
* add in "3 lines" menu, there is enough space
	* con: not always visible
* add extra button somewhere, containing some profile number or first letter
	* con: one more always displayed button (but it can be hidden if there is only 1 profile)
* add in (quest) settings
	* con: not always visible, switching is slow
When tapping the profile selection element a menu with all available profile opens and the user chooses one. Maybe a link to quest selection of the selected profile by long pressing a profile could be done.

waldyrious:
> (I keep calling this "profile", but in the end it should maybe have a different name to avoid confusion with the OSM profile)

FWIW, "preset" is a common term for this sort of thing.

rugk:
[...]

Issue #2054, opened 2020-08-27
Ability to create profiles


Norwi:
**Use case**
I use Streetcomplete in a lot of different situations. But if I am in the passenger seat of a car, I need different quests (and a different prioritisation) than when I am strolling through the streets or using public transport. And sometimes I just want to do a specific set of quests.
Therefore, it is a bit annoying to always search for the quests in the settings and sort it according to the current situation or mode of transport. A simple click but be a lot easier and faster.

**Proposed Solution**
It would be really nice to have different profiles (I don't know if you can use that word in English the same way as in German but I can't think of a better term) or settings that you can simply click and that changes the quest filter/order. Each profile has the same interface as the normal quest sorting that is now in place. And of course naming of the profiles should be available.

**Cheap Mockup**
![github_profiles](https://user-images.githubusercontent.com/18899671/91500945-d69bf300-e8c4-11ea-9c4d-0aaf21dc95c9.png)

matkoniecz:
#1654?

westnordost:
Yes, this is a duplicate. Thank you for that nice mockup though!

Issue #893, opened 2018-02-21
Rearranging of quests is broken.


Natenom:
Version: 4.0

How to reproduce:
  * Go to settings -> Advanced -> Quest selection
  * Click RESET.
  * Move "Does this bench have a backrest" (last position on the list) to the second position.
  * Move "Is this bicycle parking covered pro tected from rain)?" to the third position.
  * Now scroll down in the list until you can't see the bench quest anymore and then up again. The bench quest was movedto  a random position. Position 2 is now used by "building number quest" and position 3 by "bicyling parking covered quest".

I can always reproduce this, also with other quests.

Natenom:
It works if you avoid moving quests to the first few positions.

westnordost:
Hmm, I can't. With second position, you mean after the note quest? After moving the bicycle parking to position 3, do you exit the quest selection screen and reenter it?

Natenom:
Pos1 is the note quest. No, just scroll down and up again.

Natenom:
Okay, this is wired; I did that ~10 times before reporting this bug. But now it doesn't happen anymore...

RubenKelevra:
I can confirm this bug, as soon as you scroll down and up again the bench quest is gone and on the second place is "Which is the name of the street?" (or similar - I use the German locale).

But actually, the bug start even earlier:

-Reset
-Scroll down, and drag the Bench task to the second place
-everything looks fine.
-scroll down & up again and the street name task is gone.

westnordost:
I'll try with your reproduction.

westnordost:
Hm, can't reproduce.

RubenKelevra:
@westnordost Maybe it's just an issue on a specific device type or on a specific version of Android?

So on Android 8.0 on a OnePlus 5T it's definitely 100% reproducible.

westnordost:
I will give you a bit an insight about how it works:

1. When you open that dialog, the list of quests is **once** sorted by the custom order you specified and then displayed. Rotating the smartphone counts as reopening the dialog.
2. When you click RESET, the custom order you specified is cleared, then the list of quests is **once** sorted by the custom order again (which is empty, so nothing is sorted in effect)
3. When you move items around in the list, the custom order is updated and the display is updated **independently from another**. In other words, if the display is wrong, it does not mean that the order in the internal model is wrong.

Would you please after you have experienced the bug in the display rotate your device? How is the order then?

matkoniecz:
I just tested

[...]

PR #2069, opened 2020-09-03
[WIP] add quest presets

Helium314:
Since I really want to have quest presets (#1654), I tried adding them myself...

This PR adds working presets (each has its own quest order and visibility), but there is still a lot to be done (help is appreciated):
* by default, no quest is visible (not even notes). A manual reset of the quest list is necessary
* currently there are 3 presets, and no way to add, remove or rename anything
* the preset names are directly in the arrays.xml instead of using stings (I was lazy, easy to fix)
* I basically have no experience with coding other than pushing around huge amounts of numbers, so it's probably a good idea to be suspicious about *everything* in this PR

How it works:
* Quest order: there is now one orderList for each preset. The orderLists are stored in the same *QUEST_ORDER* string as previously, and use `+` as a separator. The list is reloaded when the preset is changed.
* Quest visibility: the database is the same, but in the cache *int* instead of *boolean* is used, because visibilies for all presets are stored in one *int* (one visibility per bit, and the preset determines which bit to use)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/Prefs.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/Prefs.java
@@ -31,6 +31,7 @@ public class Prefs
 	// not shown anywhere directly
 	public static final String
 			QUEST_ORDER = "quests.order",
+			QUEST_PRESET = "quests.preset",
 			QUEST_INVALIDATION = "quests.invalidation",
 			LAST_SOLVED_QUEST_TIME = "changesets.lastQuestSolvedTime",
 			MAP_LATITUDE = "map.latitude",
--- a/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderList.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderList.kt
@@ -18,7 +18,7 @@ import javax.inject.Singleton
     /* Is a singleton because it has a in-memory cache that is synchronized with changes made on
        the DB */
 
-    private val orderLists: MutableList<MutableList<String>> by lazy { load() }
+    private var orderLists: MutableList<MutableList<String>> = load()
 
     private val questTypeOrderLists: List<List<QuestType<*>>> get() =
         orderLists.mapNotNull { orderList ->
@@ -51,14 +51,25 @@ import javax.inject.Singleton
         save(orderLists)
     }
 
+    @Synchronized fun reload() {
+        orderLists = load()
+    }
+
     private fun load(): MutableList<MutableList<String>> {
-        val order = prefs.getString(Prefs.QUEST_ORDER, null)
+        val fullorder = prefs.getString(Prefs.QUEST_ORDER, null)?.split("+")
+        val order = fullorder?.getOrNull(prefs.getString(Prefs.QUEST_PRESET,null)?.toIntOrNull() ?: 0)
         return order?.split(DELIM1)?.map { it.split(DELIM2).toMutableList() }?.toMutableList() ?: mutableListOf()
     }
 
     private fun save(lists: List<List<String>>) {
         val joined = lists.joinToString(DELIM1) { it.joinToString(DELIM2) }
-        prefs.edit { putString(Prefs.QUEST_ORDER, if (joined.isNotEmpty()) joined else null) }
+        val fullorder = prefs.getString(Prefs.QUEST_ORDER, null).orEmpty().split("+").toMutableList()
+        val index = prefs.getString(Prefs.QUEST_PRESET,null)?.toIntOrNull() ?: 0
+        while (fullorder.size <= index) {
+            fullorder.add("")
+        }
+        fullorder[index] = joined
+        prefs.edit { putString(Prefs.QUEST_ORDER, fullorder.joinToString("+")) } // in worst case i have sth like "+++", but should work
     }
 
     private fun applyOrderItem(before: QuestType<*>, after: QuestType<*>) {
--- a/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/VisibleQuestTypeDao.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/VisibleQuestTypeDao.kt
@@ -5,6 +5,9 @@ import androidx.core.content.contentValuesOf
 
 import javax.inject.Inject
 
+import android.content.SharedPreferences
+import de.westnordost.streetcomplete.Prefs
+import de.westnordost.streetcomplete.data.quest.QuestTypeRegistry
 import de.westnordost.streetcomplete.data.quest.QuestType
 import de.westnordost.streetcomplete.data.visiblequests.QuestVisibilityTable.Columns.QUEST_TYPE
 import de.westnordost.streetcomplete.data.visiblequests.QuestVisibilityTable.Columns.VISIBILITY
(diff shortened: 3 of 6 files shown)


westnordost:
Let's call them profiles.

Okay, so you are planning to add such profiles and they should be user defined, right? And each profile will have a custom quest order plus a custom quest enablement? Why would you choose this approach over fixed profile presets, one for car/bus passengers, one for bicyclists, one for pedestrians? What would be the use cases for free profile presets?
Also, there was the idea that they will automatically be selected depending on the detected movement pattern (speed probably). Did you drop that idea and if yes, why? If no, how would it fit in your concept?

Additionally, the profile (presets) are closely related to the quest selection and order, so I think it should be in the same screen as the quest selection, maybe as tabs, maybe as a dropdown at the top. We should discuss this first before you implement it. For example, the advantage of tabs would be that it could look like in that mockup in #2054  with more information, an icon, an interface to add new presets etc.

> I basically have no experience with coding other than pushing around huge amounts of numbers, so it's probably a good idea to be suspicious about everything in this PR

Mhh, maybe not the best idea then to start with such a medium-sized feature then. You have to understand that it takes a lot of time for me as well to steward it, too.

smichel17:
I will have more free time in ~2 weeks from now. If you're willing to wait that long to work on this, I can help a little with code review.

Helium314:
> Okay, so you are planning to add such profiles and they should be user defined, right?

yes

> And each profile will have a custom quest order plus a custom quest enablement?

yes

> Why would you choose this approach over fixed profile presets, one for car/bus passengers, one for bicyclists, one for pedestrians?

With fixed presets, do you mean really fixed and no possiblity to change? Or some pre-selected lists where users can still change visibility and order? The latter seems pretty useful to me.

> What would be the use cases for free profile presets?

Mostly flexibility. I have the impression that users have very different ideas how to use profiles.
From #2054: "And sometimes I just want to do a specific set of quests.". This can only be done with free profiles. 
For me the time it takes to scan for quests is often an issue, so I try to limit the selection to quests that I expect to be available in the area want to complete. Maybe this can be done with fixed presets, but then there would have to be many presets (e.g. car passenger in urban or rural areas)
Free profiles allow loading quests without changing the "normal" profile (switch profile, enable quests you want, scan, switch back). This is often useful in areas where the more "popular" quests are already solved, and is more convenient than re-arraning quests (and safer: #893). For this specific case a "clone profile" option would be useful...

> Also, there was the idea that they will automatically be selected depending on the detected movement pattern (speed probably). Did you drop that idea and if yes, why? If no, how would it fit in your concept?

I didn't even think about that... I think this was your interpretation about my idea of having day/night profiles in #1654, but this was not intended to be switched automatically.
Automatic switching would probably be quite annoying if it doesn't work REALLY well, so I currently don't plan to include it.

> Additionally, the profile (presets) are closely related to the quest selection and order, so I think it should be in the same screen as the quest selection, maybe as tabs, maybe as a dropdown at the top. We should discuss this first before you implement it.

The current implementation is mainly to have the profiles working. I will need to learn a bit more before I can implement it like in the mockup. The mockup seems pretty nice, although I'd rather have it as a dropdown menu in the quest selecton screen.

Ideas I would like to include:
* (re)name profiles
* add and delete profiles (ask for name when creating a new one, maybe pre-filled with something generic like "Profile 2")
* clone profile (new profile pre-filled with order & visibility from an existing profile)
* (maybe) pre-set profiles (pedestrian, train, bus,...)

westnordost:
> With fixed presets, do you mean really fixed and no possiblity to change? Or some pre-selected lists where users can still change visibility and order? The latter seems pretty useful to me.

I had a mind, really fixed, so one cannot edit that preset. For example if there is a specific preset "as a passenger in a bus", it's pretty clear that noone will be able to answer opening hours and it also does not make sense to allow users to enable it (or any other that is not enabled), does it?


OK I reckon there might be two use cases for profiles: 1. "I am on a X, I **can** only solve certain things", 2. "I want to only focus on Y right now".
Though, use case number 2 seems a little odd for me. Why not change the order / default profile then? Also, I didn't announce it yet, but a large feature is planned where I will work on something like an "aspect/focus view" where one can focus on solving a certain group of quests, such as "only building-things", "only cycleways" etc.

> The current implementation is mainly to have the profiles working. I will need to learn a bit more before I can implement it like in the mockup. The mockup seems pretty nice, although I'd rather have it as a dropdown menu in the quest selecton screen.

But where to put the "add profile" button then? The mockup has the advantage that more info can be shown, so a profile is not just named "public transport" but "solvable on the bus". This is important because "public transport" could also mean "all quests related to public transport" rather than "solvable on the bus". If it is just a dropdown without additional information/description provided, it'll be hard to make that clear.

smichel17:
> Let's call them profiles.

I think they need an extra word to disambiguate with your OSM profile -- how about quest profiles?

>> use cases for free profile presets?
> the time it takes to scan for quests is often an issue
>  (and safer: #893)

I don't have anything against freely added profiles, but I do think they should be added for the right reasons. Let's not add workaround features until we've tried fixing the bug(s) and improving download speed directly, first.

> This is often useful in areas where the more "popular" quests are already solved

I don't understand why the auto-download mechanism wouldn't work here. If the popular quests ate already solved, auto-download should proceed down the list until there are enough quests showing. That can be slow, but #1901 should fix this.

> Automatic switching 
> Ideas I would like to include

To keep the size of this feature reasonable, I think most of these can be added in future PRs.

- To start, I would implement a small, fixed number of preset profiles: pedestrian, train, bus/car (passenger).
    - I haven't looked at the code yet but I think this is what you already did.
- The UI for choosing the active profile can start as a simple preference, with a plan to move it to the main menu later.
- The quest selection UI can stay identical; just make it affect whichever profile is active.

This way you get to avoid building very much UI, which is often a pain on Android. I think this would cut down the size of this feature to something more reasonable

smichel17:
Ack, posted without refreshing :/

>  "aspect/focus view" where one can focus on solving a certain group of quests, such as "only building-things", "only cycleways"

This seems.. functionally identical to quest profiles?

westnordost:
Well, what's planned is that the map view also changes. For example, the building type for each building is shown directly on the map, possibly color-coded cycleways/no-cycleways etc.

smichel17:
I could imagine the map view changing between pedestrian and vehicle profiles. And even if it didn't, they might still share an interface, depending on the implementation and how much flexibility is needed. For example, which map view is shown could be one aspect of a quest profile (err, maybe needs a different word than "quest" now. "Mapping profile"?). In any case, I think we should probably continue this discussion in an issue, when you're ready to announce it more widely.

peternewman:
> > With fixed presets, do you mean really fixed and no possiblity to change? Or some pre-selected lists where users can still change visibility and order? The latter seems pretty useful to me.
> 
> I had a mind, really fixed, so one cannot edit that preset. For example if there is a specific preset "as a passenger in a bus", it's pretty clear that noone will be able to answer opening hours and it also does not make sense to allow users to enable it (or any other that is not enabled), does it?

We might still want to change the order or disable some of the chosen quests though @westnordost . For example I find the bike/motorbike parking capacity quest too ambiguous to answer in most cases, and most cycle lanes aren't marked in the UK so I prefer to just answer the ones where it's explicitly present. I'll often hide or de-prioritise those quests.

More generally, for debugging when raising an issue it would be good to be able to make a new preset with everything disabled apart from one particular quest to find ones to test/check and then go back to normal. I can reset to default currently, but that undoes all my other customisation.

I'd potentially make different presets for doing quests on foot, versus walking somewhere and completing some on the way (e.g. disable the menu and opening hours ones) and just do the bits I can complete at a glance without stopping.

Sort of related to that I'd quite like some way to expose some of my lower priority quests in some cases, e.g. when walking along the road, I might not want to keep crossing, so only want to see say shop name, storeys and roof shape for the far side, but opening hours or menu stuff might be more valuable so higher priority for "my" side of the road. Pressing and holding on a quest to expose the lower ones would work, or toggling presets if it was quick enough to do.

Helium314:
Sorry it took me quite a while to reply here... I didn't have much time recently, not sure when it will get better.
I did some further small changes to the profiles, but never got to start on a proper UI. (and @westnordost was definitely right to doubt that this feature is a good thing for an inexperienced person to start with)

Still, I spent a few weeks with profiles to see how I actually use them.
By far my most common use for the profiles was loading only a few selected quests to reduce the sometimes really long time for loading. This is now irrelevant, as always all quests are loaded starting with SC 26.

Since the quest reordering bug basically always strikes when doing large changes to the quest order, I never actually had two lists with different order. I had imagined this to be useful for switching between quests on the same object e.g. lit and surface, or house number and building levels, or opening hours and wheelchair access...
So I just stuck to the usual flow of hiding quests in hope of finding an (easily) answerable alternative 'underneath'.

My last somewhat frequent use for profiles is that sometimes I want to focus on a few quests in specific situations. E.g. because I'm in a bus and can't answer some quests, or because I simply don't have time for slower quests like opening hours.

I will continue using profiles and see how often I use them, but for now close this PR as I won't work on it in the near furture.
Maybe if I find more time and wish to have proper profiles, I will re-open and continue in a few months.

Helium314:
Some small update:
I uploaded my 'presets' branch with some work done after the initial PR (still no proper UI).
The main addition is the possibility to copy profiles.
Then I added an 'invert order' button for dealing with quests not showing due to the order. This seems superior to switching profile or hiding when searching for solvable quests. (not actually related to profiles, but now I have less need for switching profiles).

Still, sometimes I actually find it very useful to have profiles:
* when traveling too fast for some quests
* when I don't want to spend time on quests that take long to answer
It seems that these two will not go away, and more or less coincide with what is mentioned in most of the issues

Issue #1989, opened 2020-07-30
No quest till manual search


kudlav:
I see no quest till manual search for quests. Maybe I have to wait a bit longer, but quests are appearing very fast after start of manual search.

**How to Reproduce**
- Open app and wait for GSP position.
- You can wait for quests - but they are not loaded

**Versions affected**
- App from Google Play - v21.2
- Android 10
- Samsung Galaxy A40

westnordost:
So, do you get a GPS position? Does the loading indicator start indicating that quests are being loaded?

kudlav:
I get a GPS position, map is centered and fixed to the position. What do
you mean by loading indicator? There is no indicator of auto-downloading
process.

--
Vladan Kudláč

Dne čt 30. 7. 2020 18:17 uživatel Tobias Zwick <notifications@github.com>
napsal:

westnordost:
Well, there should be. Do you have auto-sync on or off?

matkoniecz:
> What do you mean by loading indicator?

Quest icon at the bottom and bar scrolling around.

kudlav:
I have auto-sync disabled, I don't want to automatically upload answered quests.
Downloading quests is also connected with the sync? I so, it is very unclear. This field is just called Auto sync  without description.

westnordost:
Yes, sync goes both ways, otherwise it would be called auto-upload

kudlav:
So what includes sync? It would be useful to add description into settings.

westnordost:
Upload and download

kudlav:
Should I open new issue for feature request: Add description for sync
settings?

Dne čt 30. 7. 2020 19:04 uživatel Tobias Zwick <notifications@github.com>
napsal:

westnordost:
No, because auto-sync is on by default and it is explained when you turn it off.

matkoniecz:
It seems that English version is also not warning about manual download.

PR #2070, opened 2020-09-04
better describe implications of disabling auto-sync, fixes #2009 [ready for second review]

### LINKED ISSUE #2009: Warning in disabling auto-sync is not warning that also quest download is now manual
Triggered by #1989 by @kudlav

"To upload your changes manually..." is not mentioning that also download is manually.

Probably it is better to add new text - to reduce workload for translators and to allow mentioning where "scan quests" is available.

**How to Reproduce**
Go to settings, switch to autosync (if disabled), switch to disabled autosync.

**Versions affected**
v22.1 on Android 10

matkoniecz:
Also, night mode display of this menu was fixed as a bonus (top icon is no longer black during night but is recolored to text color).

Thanks to @kudlav by reporting problem in #1989 and to @westnordost for hints.

Screenshot_2020-09-04-09-23-47-272_de westnordost streetcomplete debug

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -119,9 +119,7 @@ class SettingsFragment : PreferenceFragmentCompat(),
                     val filled = context!!.getString(R.string.action_download)
                     val edited = context!!.getString(R.string.dialog_tutorial_download)
                     val uploadExplanation = view.findViewById<MaterialTextView>(R.id.tutorialDownloadPanel)
-                    //uploadExplanation.setText(R.string.dialog_tutorial_download)
-                    //uploadExplanation.setText(edited)
-                    uploadExplanation.setText(edited.replace("%s", filled))
+                    uploadExplanation.text = edited.replace("%s", filled)
                     context?.let {
                         AlertDialog.Builder(it)
                             .setView(view)
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -114,7 +114,7 @@ class SettingsFragment : PreferenceFragmentCompat(),
             Prefs.AUTOSYNC -> {
                 if (Prefs.Autosync.valueOf(prefs.getString(Prefs.AUTOSYNC, "ON")!!) != Prefs.Autosync.ON) {
                     val view = LayoutInflater.from(activity).inflate(R.layout.dialog_tutorial_upload, null)
-                    val filled = context!!.getString(R.string.action_download)
+                    val filled = requireContext().getString(R.string.action_download)
                     val uploadExplanation = view.findViewById<TextView>(R.id.tutorialDownloadPanel)
                     uploadExplanation.text = context!!.getString(R.string.dialog_tutorial_download, filled)
                     context?.let {
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -117,9 +117,8 @@ class SettingsFragment : PreferenceFragmentCompat(),
                 if (Prefs.Autosync.valueOf(prefs.getString(Prefs.AUTOSYNC, "ON")!!) != Prefs.Autosync.ON) {
                     val view = LayoutInflater.from(activity).inflate(R.layout.dialog_tutorial_upload, null)
                     val filled = context!!.getString(R.string.action_download)
-                    val edited = context!!.getString(R.string.dialog_tutorial_download)
                     val uploadExplanation = view.findViewById<MaterialTextView>(R.id.tutorialDownloadPanel)
-                    uploadExplanation.text = edited.replace("%s", filled)
+                    uploadExplanation.text = context!!.getString(R.string.dialog_tutorial_download, filled)
                     context?.let {
                         AlertDialog.Builder(it)
                             .setView(view)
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
```diff
@@ -111,9 +115,13 @@ class SettingsFragment : PreferenceFragmentCompat(),
             }
             Prefs.AUTOSYNC -> {
                 if (Prefs.Autosync.valueOf(prefs.getString(Prefs.AUTOSYNC, "ON")!!) != Prefs.Autosync.ON) {
+                    val view = LayoutInflater.from(activity).inflate(R.layout.dialog_tutorial_upload, null)
+                    val filled = context!!.getString(R.string.action_download)
+                    val uploadExplanation = view.findViewById<MaterialTextView>(R.id.tutorialDownloadPanel)
+                    uploadExplanation.text = context!!.getString(R.string.dialog_tutorial_download, filled)
```
westnordost, line 121: better use requireContext()

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
```diff
@@ -111,9 +115,13 @@ class SettingsFragment : PreferenceFragmentCompat(),
             }
             Prefs.AUTOSYNC -> {
                 if (Prefs.Autosync.valueOf(prefs.getString(Prefs.AUTOSYNC, "ON")!!) != Prefs.Autosync.ON) {
+                    val view = LayoutInflater.from(activity).inflate(R.layout.dialog_tutorial_upload, null)
+                    val filled = context!!.getString(R.string.action_download)
+                    val uploadExplanation = view.findViewById<MaterialTextView>(R.id.tutorialDownloadPanel)
```
westnordost, line 120: MaterialTextView? It think it is a normal text view

app/src/main/res/layout/dialog_tutorial_upload.xml
```diff
@@ -1,14 +1,28 @@
...
         android:drawablePadding="8dp"
+        app:drawableTint="@color/text"
         android:text="@string/dialog_tutorial_upload"
         android:textAppearance="@android:style/TextAppearance.Theme.Dialog"/>
-
-</RelativeLayout>
+    <View
+        android:layout_width="match_parent"
+        android:layout_height="3dp"
+        />
```
westnordost, line 19: Not sure how it looks, to nicely separate the two, you could also use a horizontal line (background color of view to `@color/divider`, layout height to 1dp + padding top and bottom)

app/src/main/res/drawable/ic_search_48dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="48dp"
+        android:height="48dp"
+        android:viewportWidth="24.0"
+        android:viewportHeight="24.0">
+    <path
+        android:fillColor="#000"
```
westnordost, line 7: You should create a copy of this with `fillColor='#fff'` for night mode

app/src/main/res/values/strings.xml
```diff
@@ -891,5 +891,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="at_housenumber">house number %s:</string>
     <string name="quest_summit_register_title">Is there a summit register at %s?</string>
     <string name="quest_is_defibrillator_inside_title">Is this defibrillator (AED) inside a building?</string>
+    <string name="dialog_tutorial_download">Quest download is also affected by auto-sync option. Use \"%s\" button in the menu to trigger download of quests.</string>
```
westnordost, line 894:
```suggestion
    <string name="dialog_tutorial_download">With auto-sync off, quests are not downloaded automatically. To do this manually, use the \"%s\" button in the menu at the location you want to download.</string>
```

westnordost (changes requested):
x

app/src/main/res/layout/dialog_tutorial_upload.xml
matkoniecz: I deliberately decided to not use line, as for me it is not about two strongly separate things, so strong separation was deliberately not used.

app/src/main/res/drawable/ic_search_48dp.xml
matkoniecz: Oh right, unlike upload icon that was special (black also during night) here separate night icon would be better.

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
matkoniecz: Weird, I added it to avoid crash. Removed.

westnordost (approved):

matkoniecz:
@westnordost It seems to be reverted in merging of new quest downloader. Is it intentional? Is the new quest downloader now ignoring autosync setting?

westnordost:
Yes, it is. The downloader ignores the autosync setting and the autosync setting has been renamed to only do stuff for uploading. I figured this makes more sense as there is little reason to download "by hand" anymore plus downloading of the map tiles happens anyway always

timothywashere:
Personally I find that quite annoying.
I have PAYG mobile data and when I turn my mobile data on whilst using StreetComplete, it's usually to download the background map tiles so I can solve quests I have already downloaded. I don't want it to automatically download new quests as I am paying 1p per MB.  I would rather control it manually.
But I'm sure I'm a small minority, probably of 1 :)

kudlav:
Hi, you are not alone. In the Czech Republic we have really expensive mobile data and low FUP limits. So it's no longer possible to use SC on 32kbps connection after the FUP has been reached.

The solution is simple: just download the area in advance using WiFi and then use the app completely offline.

westnordost:
Back when I had a non-flatrate, I always had mobile data off except when I explicitly needed to use it.

Issue #1079, opened 2018-05-22
New quest: tag seating (bench) at bus stops


SearchingForDottedLines:
Add tagging to indicate if a bus stop has seating or not per: https://wiki.openstreetmap.org/wiki/Tag:highway=bus_stop#Bus_stop

### General

Affected tag(s) to be modified/added: bench=* on nodes where highway=bus_stop and bench=* is not set
Question asked: Does this bus stop have seating?
Valid responses: Yes | No

### Checklist

Checklist for quest suggestions (see guidelines):
- [ Yes ] 🚧 To be added tag is established and has a useful purpose
- [ Yes ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [ Yes ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [ Yes ] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [ Yes ] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

Not sure if the question is better phrased as:

     Natural language: Does this bus stop have seating?
     Tag: Does this bus stop have a bench?

Basic overpass link to candidate nodes:

http://overpass-turbo.eu/map.html?Q=%5Bout%3Ajson%5D%5Btimeout%3A25%5D%3B%0A(%0A%20%20node%5B%22highway%22%3D%22bus_stop%22%5D%5B!%22bench%22%5D(-33.897171451452394%2C151.0599946975708%2C-33.862665104853846%2C151.1052918434143)%3B%0A)%3B%0Aout%20body%3B%0A%3E%3B%0Aout%20skel%20qt%3B

Potential other candidates:

Current bus stop quests also consider public_transport=platform see https://wiki.openstreetmap.org/wiki/StreetComplete/Quests

Responses:

https://taginfo.openstreetmap.org/keys/bench#values

| Response | Tag to set |
| ------------- | ------------- |
| Yes | bench=yes  |
| No | bench=no  |

matkoniecz:
Duplicate of #200

SearchingForDottedLines:
#200 seems to focus on bins more so than benches, which seemed to be the focus of discussion in the comments, but yes there is an overlap.

westnordost:
What about bus stops where the bench is mapped separately? Also, how is this information useful? Please give an example

SearchingForDottedLines:
Primary driver is accessibility.
[...]

PR #2073, opened 2020-09-04
Bench for bus stops

matkoniecz:
I am bit dubious about usefulness of this quest but it is asked after all reasonable quests were asked in a given region and I want to reduce number of open issues to 100 :)


app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
```diff
@@ -0,0 +1,44 @@
...
+class AddBenchStatusOnBusStop(o: OverpassMapDataAndGeometryApi, r: ResurveyIntervalsStore) : SimpleOverpassQuestType<Boolean>(o) {
+
+    override val tagFilters = """
+        nodes with
+        (
+          (public_transport = platform and ~bus|trolleybus|tram ~ yes)
+          or
+          (highway = bus_stop and public_transport != stop_position)
+        )
+        and physically_present != no and naptan:BusStopType != HAR
```
peternewman, line 19: Ah now I see why you were asking in the other PR. Is it worth pulling this out so it's a common rule across these two bus stop quests and any future ones. Or at least adding comments to both saying to update the other one if it changes.

Or actually TBH all of this filter apart from the bench bit could be common couldn't it?

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
```diff
@@ -0,0 +1,44 @@
...
+        and (!bench or bench older today -${r * 5} years)
+    """
+
+    override val commitMessage = "Add whether bus stop has a bench"
```
peternewman, line 23:
```suggestion
    override val commitMessage = "Add whether a bus stop has a bench"
```

peternewman (commented):
A few comments for what they're worth.

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override val tagFilters = """
+        nodes with
+        (
+          (public_transport = platform and ~bus|trolleybus|tram ~ yes)
+          or
+          (highway = bus_stop and public_transport != stop_position)
+        )
+        and physically_present != no and naptan:BusStopType != HAR
+        and (!bench or bench older today -${r * 5} years)
```
westnordost, line 20:
```suggestion
        and (!bench or bench older today -${r * 4} years)
```

(All these are using 2, 4, 8,... unless there is no good reason for an exception, let's stay with this IMO)

westnordost (approved):

HolgerJeromin:
Ref #1079

HolgerJeromin:
>  I want to reduce number of open issues to 100 :)

91! :-)

PR #2078, opened 2020-09-06
Merge surface quest and detail surface quest, attempt to make interface less confusing [accepted, waiting for merge]

### LINKED ISSUE #279: add quest to make surface=paved/unpaved more detailed

### LINKED ISSUE #1952: Indicate that categories contain other values
**Use case**
When I recently started using StreetComplete, I was looking for the "wood" and "metal" surface types (mostly for bridges and stairs), but could not find them in the app, instead leaving notes (and probably annoying my local OSM community). I did not realise how to find them until I went to report them missing as a bug and discovered #1863, which pointed out that the "generic" labels were actually categories, not final values.

**Proposed Solution**
I would suggest three changes to improve the discoverability:
1. Change the wording of the heading to include the word "category" or similar, to indicate that the selection is not a final value.
2. Remove the word "generic" from the category name. I am not describing a generic paved path, I am selecting a specific type of paving.
3. Include a visual indication that more options will appear when an option is selected, e.g. an ellipsis (…), chevron right (>), or chevron down (∨), preferably changing to its opposite when the category is expanded (i.e. a chevron right on a closed category changes to a chevron left on an expanded one).

This should also be used for building types, and any other quests using a similar UI paradigm of first selecting a category and then an item in that category.

Examples:

streetcomplete-path-categories-right
streetcomplete-path-surfaces-right

### LINKED ISSUE #2000: More surfaces available for paths
I noticed that for (foot)paths, not many surfaces are available to choose from. I especially miss "gravel" (instead "pebblestones" is available which looks similar on the photo). When selecting "show more", it shows choices that I think are only used for tracks. I would like the options for path surfaces to be more similar to those available for roads...

By the way, speaking of available reply options, I once has a quest for what was grown in an orchard, with "tomatoes" and "paprikas" available. By definition, what's grown in orchards is grown on trees, so if tomatoes were grown there, it wouldn't be called an orchard :) I was also surprised by some exotic options available for sports areas, such as Canadian & Australian football. Would be nice if these were not offered in Europe. The option "table tennis" that I needed wasn't available.

Ideally, the available answer options could be made dependent on the frequency of answers already present in that (wide) area, but that's probably not easy to code?

### LINKED ISSUE #2196: Extend DetailRoadSurface to also cover for paths/footways/cycleways
**Use case**
Currently the DetailRoadSurface quest uses the ALL_ROADS list, that doesn't cover for paths/footways/cycleways.
(I recently passed nearby one of such cases and I wondered why StreetComplete didn't show me anything to detail the surface=paved footway I was on)

**Proposed Solution**
Either:
- add paths/footways/cycleways to the query used by DetailRoadSurface
or
- add a new quest covering only the "non-road" ways (as it is now when the surface tag is missing)

From this comment I'd understand that the second option may be preferred:
https://github.com/westnordost/StreetComplete/issues/279#issuecomment-563974925

(and thanks for that particular quest, it's always nice to find an excuse to resurvey an area) :)

matkoniecz:
For now done just for roads - and review is welcomed. to implement fixes/tweaks just once, not multiple times.

Once initially reviewed I will implement it also for paths, maybe also cycleway surface part and pedestrian surface part.

It will fix #2000 and fix #1952 and fix #279 and fix #2196


app/src/main/res/authors.txt
```diff
@@ -44,6 +44,9 @@ panorama_path_surface_unpav... CC-BY 4.0         https://github.com/westnordost/
 panorama_surface_paved.jpg     CC0               https://www.goodfreephotos.com/united-states/indiana/hoosier-hill/hooiser-hill-road-near-the-top-indiana.jpg.php
 panorama_surface_unpaved.jpg   CC-BY 2.0         https://www.flickr.com/photos/ian_munroe/3633361321
 panorama_surface_ground.jpg    CC-BY 3.0         https://commons.wikimedia.org/wiki/File:Dirt_Road,_Bulacan,_Philippines_-_panoramio.jpg
```
westnordost, line 46: Any reason for keeping the panorama images?

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
```diff
@@ -117,13 +117,12 @@ object QuestModule
         AddRecyclingType(o),
         AddRecyclingContainerMaterials(o, r),
         AddSport(o),
-        AddRoadSurface(o, r), // used by BRouter, OsmAnd, OSRM, graphhopper, HOT map style
+        RoadSurface(o, r), // used by BRouter, OsmAnd, OSRM, graphhopper, HOT map style
```
westnordost, line 120: I know that the new road surface quest is not always limited to adding the surface but in case of paved etc also modifies it, but I'd rather leave the name as `AddRoadSurface` because
1. `Add`-something has somewhat established itself as the naming standard for quest types. `RoadSurface` sounds like it is a data class or enum, much like `Surface`
2. Since resurvey quests, many quests do not only add information but also modify it, but I didn't change the naming because of that

app/src/main/res/authors.txt
matkoniecz: No, this images will be removed once I extend this to all other surfaces quests.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/RoadSurface.kt
```diff
@@ -2,28 +2,30 @@ package de.westnordost.streetcomplete.quests.surface
...
+            and !surface:note
+            and segregated != yes
+            )
+        )
+        and (access !~ private|no or (foot and foot !~ private|no))
     """
-    /* ~paved ways are less likely to change the surface type */
-
-    override val commitMessage = "Add road surfaces"
+    override val commitMessage = "Add road surface info"
```
peternewman, line 28: Won't this upset the stats system?

matkoniecz: AFAIK no, stats read changeset tags - not changeset comments.

And app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt  is modified to handle that detail quest is merged into surface quest.

westnordost: correct

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
matkoniecz: name changed

app/src/main/res/values/strings.xml
```diff
@@ -131,7 +131,7 @@
     <string name="quest_buildingLevels_title_buildingPart">"How many levels does this building part have?"</string>
     <string name="quest_roofShape_title">"What basic shape does this building’s roof have?"</string>
     <string name="quest_roofShape_select_one">"Select one:"</string>
-    <string name="quest_roofShape_show_more">"Show more…"</string>
+    <string name="quest_roofShape_show_more">"Show more…"</string><!--it is used in generic list, not only for roof shape but rename would break translations-->
```
peternewman, line 134: Don't the translations get uploaded from SC to the translation system, therefore couldn't you do a sed on all the strings files to replace quest_roofShape_show_more with quest_show_more and mean no translation is required?

matkoniecz:
hm, it seems that AddPathSurfaceForm, AddRoadSurfaceForm and maybe also AddCyclewayPartSurface and AddFootwaySurface will now have almost the same code. Especially part about capturing "unpaved"/"paved" and switching form.

Main difference would be a different image used for `unpaved`, `paved` and `ground`.

And maybe `topItems` (if special ordering depending on path type or currently tagged surface value will remain).


I am tempted to create class and have inheritance in AddPathSurfaceForm and AddRoadSurfaceForm to reduce code duplication.

But I see to remember advice that avoiding code duplication is not sufficient reason to introduce inheritance. Not sure whatever I am remembering correctly, I am unable to track down this comment...

westnordost:
Hmm, I see there is a lot of code now that could potentially be shared, but most of it is overhead over the actual functionality: Letting the user explain in a text field why he selected a generic category.

Is there any particular reason or advantage of switching the layout to display such a text field over just showing a dialog in which the user ought to input the text like for the `AddOpeningHoursForm` for when the user selects that it "has no regular opening hours"? Because the dialog would be much less code and if needed, that dialog could be sourced out and reused instead of introducing some class hierarchy.

westnordost:
F.e.

```kotlin
class DescribeGenericSurfaceDialog(
    context: Context,
    onSurfaceDescribed: (txt:String) -> Unit
) : AlertDialog(context, R.style.Theme_Bubble_Dialog) {
    init {
        val view = LayoutInflater.from(activity).inflate(R.layout.quest_generic_surface_comment, null)
        val commentInput = view.findViewById<EditText>(R.id.commentInput)
        // TODO enable/disable ok button based on whether commentInput is empty (with TextWatcher?)
        setTitle(R.string.quest_surface_comment_title)
        setPositiveButton(android.R.string.ok) { _, _ ->
                val txt = commentInput.text.toString().trim()
                if (!txt.isEmpty()) { onSurfaceDescribed(txt) }
        }
        setNegativeButton(android.R.string.cancel, null)
        setView(view)
    }
}
```

westnordost:
> Is there any particular reason or advantage of switching the layout to display such a text field over just showing a dialog

...cause I only see two disadvantages actually:
1. more and more complicated code
2. the user selected either `unpaved`, `paved` or `ground`. When the layout is changed, he cannot see anymore what he selected and thus what will be tagged

matkoniecz:
> Is there any particular reason or advantage of switching the layout to display such a text field over just showing a dialog

Main and probably sole reason is that I had a clear idea how to implement this.

matkoniecz:
So thanks for the better idea, I will test it!

westnordost:
Should I finish this PR?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -29,7 +37,15 @@ class AddCyclewayPartSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyInterv
...
     override fun createForm() = AddPathSurfaceForm()
 
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.updateWithCheckDate("cycleway:surface", answer)
-    }
+    override fun applyAnswerTo(answer: DetailSurfaceAnswer, changes: StringMapChangesBuilder) {
+        when(answer) {
+            is SurfaceAnswer -> {
+                changes.updateWithCheckDate("cycleway:surface", answer.value)
+                changes.deleteIfExists("source:cycleway:surface")
```
westnordost, line 44: Why this line? Should it maybe be "cycleway:surface:note"?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddFootwayPartSurface.kt
```diff
@@ -28,7 +36,16 @@ class AddFootwayPartSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyInterva
 
     override fun createForm() = AddPathSurfaceForm()
 
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.updateWithCheckDate("footway:surface", answer)
+    override fun applyAnswerTo(answer: DetailSurfaceAnswer, changes: StringMapChangesBuilder) {
+        when(answer) {
+            is SurfaceAnswer -> {
+                changes.updateWithCheckDate("footway:surface", answer.value)
+                changes.deleteIfExists("source:footway:surface")
```
westnordost, line 43: same here

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurface.kt
```diff
@@ -39,7 +39,16 @@ class AddPathSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyIntervalsStore
 
     override fun createForm() = AddPathSurfaceForm()
 
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.updateWithCheckDate("surface", answer)
+    override fun applyAnswerTo(answer: DetailSurfaceAnswer, changes: StringMapChangesBuilder) {
+        when(answer) {
+            is SurfaceAnswer -> {
+                changes.updateWithCheckDate("surface", answer.value)
+                changes.deleteIfExists("source:surface")
```
westnordost, line 46: and here

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurface.kt
```diff
@@ -46,8 +49,17 @@ class AddRoadSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyIntervalsStore
 
     override fun createForm() = AddRoadSurfaceForm()
 
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.updateWithCheckDate("surface", answer)
+    override fun applyAnswerTo(answer: DetailSurfaceAnswer, changes: StringMapChangesBuilder) {
+        when(answer) {
+            is SurfaceAnswer -> {
+                changes.updateWithCheckDate("surface", answer.value)
+                changes.deleteIfExists("source:surface")
```
westnordost, line 56: and here

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: For now it is for consistency with deleting `source:surface` tag in basic quest. I am not entirely sure about it.

And yes, deleting `cycleway:surface:note` may be a good idea but I am also not 100% sure about it. Either way I will leave it for later.

westnordost: Ah, understood. It's the "Bing removal". Nevermind then, it's all the same to me.

Since this is a resurvey quest, I would say it is quite necessary to delete the `cycleway:surface:note` if the selection is not generic, otherwise data garbage is left behind.

matkoniecz:
re https://github.com/westnordost/StreetComplete/pull/2078#issuecomment-700269740 

Is avoiding use of `AlertDialog.Builder` deliberate here? It seems to me that it is standard way to do this and it should work.

westnordost:
You mean instead of a class, this becomes a static function that builds and an alert dialog or what?

There are other classes like this in the project, so it would only be consistent to use inheritance.

smichel17:
> You mean instead of a class, this becomes a static function that builds and an alert dialog or what?

Yes. If nothing else, it saves a level of indentation. And they can be cleanly/easily composed by breaking common groups of properties into extension functions (I can write an example shortly when on desktop, vs mobile right now).

westnordost:
Hm, let's please keep to the current pattern right now because:
- consistency: There is `TimeRangePickerDialog`, `ValuePickerDialog`, `RequestLoginDialog`, `MainMenuDialog` etc
- not make it more complicated for @matkoniecz to finish this PR by starting a discussion about code style. Let's have this discussion after this is merged if you think changing all these dialogs to static functions is an improvement

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: Added removal of `*surface:note` tags.

matkoniecz:
Nearly ready, I hoped that `getButton(DialogInterface.BUTTON_POSITIVE).isEnabled` will also recolour button to grayed out, but apparently I need to do this manually.

matkoniecz:
Technically I added button recoloring but in way that is an ugly hack (that can be polished a bit but still would be a hack).

I am unsure how one may influence styling of disabled button in AlertDialog.

I hoped that `values/themes.xml` is an answer.

I was pretty sure that this would work

```
   <style name="Theme.Bubble.Dialog" parent="Theme.MaterialComponents.DayNight.Dialog">
        <!-- The colors need to be set another time here because this theme does not inherit
             from AppTheme, but the material components theme. Not sure if there is an easier way -->
        <item name="colorPrimary">@color/primary</item>
        <item name="colorPrimaryVariant">@color/primary_variant</item>
        <item name="colorSecondary">@color/accent</item>
        <item name="colorSecondaryVariant">@color/accent_variant</item>
        <item name="colorOnPrimary">@color/on_primary</item>
        <item name="colorOnSecondary">@color/on_secondary</item>
        <item name="colorAccent">@color/button_bar_button_text</item>
```

(change in `<item name="colorAccent">@color/button_bar_button_text</item>` as `color/button_bar_button_text.xml` has recoloring for `state_enabled="false"`)

westnordost:
I really dislike hacks, workarounds are better. Is it so important that the OK button is greyed out? I think any of these alternatives would be fine:
- OK button is not clickable if the input is empty. Setting `isEnabled=false` might be enough
- tapping OK if the input is empty will show a short error message that one needs to specify text
- tapping OK if the input is empty will do nothing

Looking at how I solved it in the `OpeningHoursForm` which has a similar interface when answering "no regular opening hours", it shows a short erorr dialog in that case. It would be consistent if this quest did the same.

matkoniecz:
> Is it so important that the OK button is greyed out?

No, but it seemed to be an obvious behavior that should be somehow implementable in obvious way that I am unable to find.

If there is no reasonable way to do this, then I will skip it as not so important.

> OK button is not clickable if the input is empty. Setting isEnabled=false might be enough

This is already working this way. I will revert the hacky color change and test whatever everything works fine.

> tapping OK if the input is empty will show a short error message that one needs to specify text

And maybe implement this one for consistency.

matkoniecz:
Retested, it is now ready for review, squashing and merging (or not squashing if you prefer this way)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DescribeGenericSurfaceDialog.kt
```diff
@@ -0,0 +1,42 @@
...
+import de.westnordost.streetcomplete.R
+
+class DescribeGenericSurfaceDialog(
+    context: Context,
+    onSurfaceDescribed: (txt: String) -> Unit
+) : AlertDialog(context, R.style.Theme_Bubble_Dialog) {
+    val view = LayoutInflater.from(context).inflate(R.layout.quest_surface_detailed_answer_impossible, null)
+    val explanationInput = view.findViewById<EditText>(R.id.explanationInput);
+    init {
+        setTitle(context.resources.getString(R.string.quest_surface_detailed_answer_impossible_title))
```
westnordost, line 17: I think it would make sense to explain what the user is supposed to do here. 

`R.string.quest_surface_detailed_answer_impossible_description` should be shown as a description in this dialog,

`quest_surface_detailed_answer_impossible_title` should be changed to simply _"Describe surface"_

That the error dialog that shows if the user did not provide any description simply repeats the description is fine.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -29,7 +37,17 @@ class AddCyclewayPartSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyInterv
...
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.updateWithCheckDate("cycleway:surface", answer)
+    override fun applyAnswerTo(answer: DetailSurfaceAnswer, changes: StringMapChangesBuilder) {
+        when (answer) {
+            is SurfaceAnswer -> {
+                changes.updateWithCheckDate("cycleway:surface", answer.value)
+                changes.deleteIfExists("source:cycleway:surface")
+                changes.deleteIfExists("cycleway:surface:note")
+            }
+            is DetailingWhyOnlyGeneric -> {
```
westnordost, line 47: I think the naming of the classes is a bit off. I think it should be:

The interface/base class: `SurfaceAnswer`
The class that just includes the surface because it is already a detailed surface answer: `DetailSurfaceAnswer`
The class that includes both the description and the surface value: `GenericSurfaceAnswer` because the answer does not only contain the "detailing" but also the generic surface.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -29,7 +37,17 @@ class AddCyclewayPartSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyInterv
 
     override fun createForm() = AddPathSurfaceForm()
 
-    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
-        changes.updateWithCheckDate("cycleway:surface", answer)
+    override fun applyAnswerTo(answer: DetailSurfaceAnswer, changes: StringMapChangesBuilder) {
+        when (answer) {
+            is SurfaceAnswer -> {
+                changes.updateWithCheckDate("cycleway:surface", answer.value)
+                changes.deleteIfExists("source:cycleway:surface")
```
westnordost, line 44: Why is `source:cycleway:surface` only deleted when providing a detailed surface answer and not if a "generic" surface answer is provided? Same for the other classes that do this.
It could actually move out of the `when` no?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurface.kt
```diff
@@ -3,27 +3,30 @@ package de.westnordost.streetcomplete.quests.surface
...
+            or surface ~ ${ANYTHING_UNPAVED.joinToString("|")} and surface older today -${r * 4} years
+            or surface older today -${r * 12} years
+        or
+            (
+            surface ~ paved|unpaved
+            and !surface:note
+            and segregated != yes
+            )
+        )
+        and (access !~ private|no or (foot and foot !~ private|no))
```
westnordost, line 27: indentation is a bit off. I first thought that the `or` would conflict with the `and`

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurface.kt
```diff
@@ -3,27 +3,30 @@ package de.westnordost.streetcomplete.quests.surface
...
+        and
+        (
+            !surface
+            or surface ~ ${ANYTHING_UNPAVED.joinToString("|")} and surface older today -${r * 4} years
+            or surface older today -${r * 12} years
+        or
+            (
+            surface ~ paved|unpaved
+            and !surface:note
+            and segregated != yes
```
westnordost, line 24: `segregated` on a road? What does this mean?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
```diff
@@ -1,56 +1,43 @@
...
-    private fun determinePathType(tags: Map<String, String>): String? {
-        val pathType = tags["highway"]
-        // interpret paths with foot/bicycle/horse=designated as...
-        if ("path" == pathType) {
-            if ("designated" == tags["bicycle"]) return "cycleway"
-            if ("designated" == tags["horse"]) return "bridleway"
-            if ("designated" == tags["foot"]) return "footway"
+    override fun onClickOk(selectedItems: List<String>) {
+        val value = selectedItems.single()
+        if(value == "paved" || value == "unpaved" || value == "ground") {
```
westnordost, line 27: `ground` as well? But in the element filter, you don't search for `ground`.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
```diff
@@ -1,56 +1,43 @@
...
+                .setMessage(R.string.quest_surface_detailed_answer_impossible_confirmation)
+                .setPositiveButton(R.string.quest_generic_confirmation_yes) {
+                    _, _ -> run {
+                        DescribeGenericSurfaceDialog(requireContext()) { description ->
+                          applyAnswer(DetailingWhyOnlyGeneric(value, description))
+                        }.show()
+                    }
+                }
+                .setNegativeButton(android.R.string.cancel, null)
+                .show()
```
westnordost, line 38: I think this UI could be improved here: This could go into one dialog, a la "Can you really not specify it more detailed? Then please describe the surface here, for example...."

westnordost (changes requested):

westnordost:
Ok, all reviewed. Just small things

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceForm.kt
```diff
@@ -1,26 +1,41 @@
...
-        Item("paved", R.drawable.panorama_surface_paved, R.string.quest_surface_value_paved, null, PAVED_SURFACES.toItems()),
-        Item("unpaved", R.drawable.panorama_surface_unpaved, R.string.quest_surface_value_unpaved, null, UNPAVED_SURFACES.toItems()),
-        Item("ground", R.drawable.panorama_surface_ground, R.string.quest_surface_value_ground, null, GROUND_SURFACES.toItems())
-    )
-
-    override fun onClickOk(value: String) {
-        applyAnswer(value)
+    override fun onClickOk(selectedItems: List<String>) {
+        val value = selectedItems.single()
+        if(value == "paved" || value == "unpaved" || value == "ground") {
```
westnordost, line 25: here also: `ground`?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurface.kt
```diff
@@ -9,7 +9,7 @@ import de.westnordost.streetcomplete.data.osm.mapdata.OverpassMapDataAndGeometry
 import de.westnordost.streetcomplete.settings.ResurveyIntervalsStore
 
 class AddPathSurface(o: OverpassMapDataAndGeometryApi, r: ResurveyIntervalsStore)
-    : SimpleOverpassQuestType<String>(o) {
+    : SimpleOverpassQuestType<DetailSurfaceAnswer>(o) {
 
     override val tagFilters = """
```
westnordost, line 14: I think something is missing here in the filter: Looking for surface~paved|unpaved. Or is this deliberate?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
matkoniecz: I worried about fitting text and field on one screen - especially on small screens and in languages where the same text is much longer than in English.

Also, there is a plenty of text (though maybe it can be shortened) so I thought that in such layout there is a greater chance that it will be actually noticed and read.

But if you think that it should be changed it should fairly simple to do - let me know if I should do this.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurface.kt
matkoniecz: oh right, that is a copy paste accident

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: Right, it should be changed

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
westnordost: It's a dialog, it could potentially cover most of the screen. I don't think there is really not enough space here

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
westnordost: So this means that you should remove `|| value == "ground"` here and in the other places where you added it

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
matkoniecz: I used not `DetailSurfaceAnswer` but `SpecificSurfaceAnswer` as I think that it is more, well, specific, how it differs from `GenericSurfaceAnswer`.

`DetailSurfaceAnswer` could be detailed in some other way.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/DescribeGenericSurfaceDialog.kt
matkoniecz: > quest_surface_detailed_answer_impossible_title should be changed to simply "Describe surface"

Done

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
matkoniecz: And find new image for `ground` or drop it as an answer (if treated as synonym of `dirt`)

michaelblyons: I re-read the wiki for [`surface`](https://wiki.openstreetmap.org/wiki/Key:surface), and I think `ground` should be dropped. The description (at least in English) basically asks you to use a different value.

matkoniecz: Thanks for pointing me toward wiki statement about `ground` value.

Wiki matches rather what was originally implemented by @westnordost rather what I think.

I will start treating it as a generic value, like `surface=paved` and `surface=unpaved`.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
westnordost: ok

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
matkoniecz: I have no strong preference (for now at least, I may change my mind after using the new form of a quest for some time), so I would implement here whatever version you prefer.

But I am not entirely sure what is the preferred solution:

Reasking for `surface=ground` is unwanted.

Asking for detailed note for `surface=ground` is unwanted.

Is it necessary to keep "Are you sure that you can not determine it more specifically?" for "Ground (generic)" or some equivalent?

Is it necessary to find a new image for `ground` answer?

michaelblyons: > write something like "mix of leaves, roots, divots"

I humbly disagree. In my opinion, this is `dirt` with an appropriate `smoothness` tag.

> it will differ all the time. A few meters grass, then more earth, then some pebbles and rocks, then a puddle of mud

This is a reasonable point. But now I am in the same place as @matkoniecz. See his questions above.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
matkoniecz: screen04

vs text extracted into extra dialog

screen05

I will implement merged version but I think that one extra screen is better.

Or maybe this text can be shortened solving root problem?

matkoniecz: Done

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
matkoniecz: For now old image is used, there is no confirmation on selecting `ground`, resurvey for `ground` is not done.

westnordost: I think this is fine

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
westnordost: Ok, this is indeed a long text. I didn't realize you mention 1. differs along the way, 2. can't say 3. 255 characters, 4. an example.

You are right, in this case maybe a second dialog is indeed better.

matkoniecz: Done

westnordost (approved):
Nice, well done!

This will go into v26, so it should be merged after v25.0 is out (end of this week)

matkoniecz:
I hope that new interface will work better.

I really like that it reduces quest count while increasing features (resurvey of `paved` unpaved` specified on footway/cycleway part is a minor change, but always something).

westnordost:
Mh, one more thing: I hate to bring this up (again?), but regarding `note:surface` vs `surface:note`, it might be inconsistent to have `check_date:key` but then also `surface:note` and not `note:surface`.

What speaks for `surface:note` is that it will be displayed usually right next to `surface` in JOSM. What speaks against it is that it invades on the namespace of that tag, f.e. `recycling:note` is problematic, even `cycleway:note` might be, because validators might check that "subkeys" of `cycleway` should only be `left`, `right` and `both`. Or `ramp:note` (compare: `ramp:bicycle` etc.).

In any case, whatever comes out of this discussion now, StreetComplete should at least _understand_ the other alternative how the note on a surface can be tagged and not ask again then.

matkoniecz:
So both `footway:surface:note` and `note:footway:surface` should exclude in queries, and so on?

I can also flip what is tagged, I have no strong preference here (I do not even remember which variant is tagged right now)

westnordost:
> So both footway:surface:note and note:footway:surface should exclude in queries, and so on?

Yes. Looking at taginfo, actually the majority of currently tagged notes on surfaces is tagged not the way StreetComplete tags it.

Right now StreetComplete tags `surface:note`.

PR #2082, opened 2020-09-08
Use more Compat classes.

Isira-Seneviratne:
* Use [`LocationManagerCompat.isLocationEnabled()`](https://developer.android.com/reference/kotlin/androidx/core/location/LocationManagerCompat#islocationenabled), as [`Settings.Secure.LOCATION_MODE`](https://developer.android.com/reference/android/provider/Settings.Secure#LOCATION_MODE) has been deprecated and the SDK documentation recommends using `LocationManager.isLocationEnabled()` instead.
* Use [`LinkifyCompat.addLinks()`](https://developer.android.com/reference/kotlin/androidx/core/text/util/LinkifyCompat#addlinks_1) to make improvements to `Linkify.addLinks()` introduced in API level 28 available on older API levels.
* Use `WindowInsetsCompat`.
* Use [`ActivityCompat.recreate()`](https://developer.android.com/reference/androidx/core/app/ActivityCompat#recreate(android.app.Activity)) to work around bugs in the `recreate()` method on API levels below 28.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.java
@@ -7,6 +7,7 @@ import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
 import android.graphics.Point;
+import android.location.LocationManager;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.net.Uri;
@@ -25,6 +26,7 @@ import androidx.annotation.NonNull;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.content.ContextCompat;
+import androidx.core.location.LocationManagerCompat;
 import androidx.core.text.util.LinkifyCompat;
 import androidx.fragment.app.Fragment;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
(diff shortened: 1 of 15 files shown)


westnordost (commented):
The following compat thingies are only used for compatibility for API 21 and below. So, this is change would become meaningless and even complicates things (makes it less readable) when the min api version is set to 21:
- `elevationCompat`
- `setOnApplyWindowInsetsListenerCompat`
- `LocationUtil.isLocationOn`

For the following things, I did not find any info on how this would improve things:
- `LinkifyCompat`

Additionally, `LocationUtil.isLocationOn` seems like a dejavú, I think I remember that you included the replacement of this method with `isLocationEnabledCompat` in another PR. I already wrote that the implementation of that method looks different from `LocationUtil.isLocationOn`.

So, I don't see an added value of this PR.

Isira-Seneviratne:
> Additionally, `LocationUtil.isLocationOn` seems like a dejavú, I think I remember that you included the replacement of this method with `isLocationEnabledCompat` in another PR. I already wrote that the implementation of that method looks different from `LocationUtil.isLocationOn`.

I commented earlier that the documentation says that `Settings.Secure.LOCATION_MODE` has been deprecated and `LocationManager.isLocationEnabled()` should be used instead.

Isira-Seneviratne:
> For the following things, I did not find any info on how this would improve things:
> 
> * `LinkifyCompat`

I checked the code for the method, and it uses the method implementation for `Linkify.addLinks` on Android Pie and later.

Isira-Seneviratne:
@westnordost I added a new commit.

westnordost:
OK! Please note my last comment though: https://github.com/streetcomplete/StreetComplete/pull/2082#pullrequestreview-490342347

After upgrade to minSdkVersion 21 (it will happen, it is only a matter of time), some/all/many changes in this PR might be obsolete. So at the moment I wouldn't put time in this until the minSdkVersion hasn't been increased to 21.

Isira-Seneviratne:
Okay 👍

westnordost:
I will also close this one. The central window insets stuff is already on master, some other things don't make sense (anymore).

PR #2088, opened 2020-09-11
new photo with more surface=paving_stones variants

matkoniecz:
Screenshot_2020-09-11-20-49-00-890_de westnordost streetcomplete debug

draft as I still need to change authors.txt


westnordost:
The text obscures the other types of stone pavings a bit, could you change the cropping in a way that a little more of the other pavings is visible? Maybe also zoom in a little as long as it stays clear that the stone pavement at the top are paving plates and not concrete/asphalt. Also, compared to the other pavings (especially sett), the photo looks a little pale, maybe adapt the contrast a little?

rhhsm:
paving_stones includes natural stone (unfortunately), so it would be even better to have a photo that shows some too

matkoniecz:
> paving_stones includes natural stone (unfortunately), so it would be even better to have a photo that shows some too

If you have even better photo (or some result of merging multiple photos) I would gladly use it. This one took some effort and is clear improvement, so I will use it and try later to obtain something even better.

matkoniecz:
> Maybe also zoom in a little as long as it stays clear that the stone pavement at the top are paving plates and not concrete/asphalt. 

I really wanted to include at least one clearly square paving stone - this type is quite popular in Poland and some people were confused what fits them.

I tried new crop from other photo, but it was clearly worse. 

> Also, compared to the other pavings (especially sett), the photo looks a little pale, maybe adapt the contrast a little?

Though I played for the first time in my life with saturation/contrast/brightness tweakers with - I think - sort of acceptable results.

screen01

matkoniecz:
retry with an earlier image

different crop and attempt to balance color

(authors.txt still need to be updated, lets see whatever this one is good enough)

Screenshot_2020-09-18-21-01-08-771_de westnordost streetcomplete debug

westnordost:
This looks very fine! Thank you for the effort!

westnordost:
The PR is still marked as draft

matkoniecz:
> The PR is still marked as draft

Yes, I still need to fix authors.txt and I wanted to get confirmation that image is fine (as I was unsure which image will end used).

matkoniecz:
I went with license that you used for works in main res folder - I will leave merging for you as technically I never asked about license of the file, so to be 100% sure you can confirm by merging :)

westnordost:
Why, you made the photo

westnordost:
Maybe include the source graphic in the /res/ folder since it is not posted anywhere else?

matkoniecz:
> Why, you made the photo

Nope, this one is the final one that you took :)

matkoniecz:
> Maybe include the source graphic in the /res/ folder since it is not posted anywhere else?

Done.

PR #2090, opened 2020-09-12
Remove own useless archived SVG repo

rugk:
See https://github.com/rugk/streetcomplete-quest-svgs/issues/15

It's archived now and not used/needed anymore, so we can remove it here too.


westnordost:
Right, thanks!

PR #2091, opened 2020-09-14
Antialias pointer pin drawable

Plastix:
**Background:**
Currently the pointer pin looks jagged when rotating the map.

**Modifications:**
* Create a new paint in `PointerPinView` to enable anti-aliasing. Use it
when rendering the pointer pin drawable.

### Before:
Screenshot_1600054720

### After:
Screenshot_1600054672


westnordost:
Wow nice, thank you!

Issue #1996, opened 2020-08-02
Community Vote - Always show GPS Button?


Atrate:
Related issues: #1843, #1896, #1990, especially https://github.com/westnordost/StreetComplete/issues/1990#issuecomment-667721833.

### Background

Quite a few issues have been opened so far discussing the way the GPS button works in StreetComplete after v20.0.
This issue aims to summarise main arguments for and against the GPS button disappearing after getting a lock-on. Feel free to ping me with additional arguments and I'll add them to the list as soon as I can.

### Pros of the current state of things:

- Minimizes the amount of UI elements cluttering the map at any given time
- The functionality of the GPS button is provided by the new location pin
- It is planned to replace the +/- buttons with a slider which then would free up the right side of the screen completely

### Cons of the current state of things

- The GPS button's disappearance can cause confusion
- It is easier to simply tap a GPS button instead of searching for the pin all over the screen
- The GPS button should be very familiar from similar location-based apps such as OSMAnd, GMaps, Vespucci, etc.
- Some users have reported the location pin to be hard to tap on smaller displays
- If the location pin is very close to/overlapping a quest pin you are likely to press the wrong one

### Solution(s)

There are a number of ways to approach this issue, each will have its own emoji for voting. Feel free to ping me with additional solutions and I'll add them to the list as soon as I can.

 - Bring back the old functionality — make the GPS button not disappear at all. The new position marker can stay or go, depending on the design considerations. Emoji: 👍 

 - Leave it as it is — the GPS button will disappear when getting a good GPS lock. Emoji: 👎 

- Make the GPS button disappear only when the map is locked onto your position and following. Emoji: :confused: 
 - I don't mind either way. Emoji: :eyes:

Feel free to vote on multiple options, after all, it's all non-binding, @westnordost has the final say here.

westnordost:
(I edited the voting options because certain things I do not plan to implement in any case)

smichel17:
I want to propose a fourth option: the GPS button should disappear when when the map is locked on to your position (ie, disappears after pressing it if location is already found, or when location is found) and appears when no longer locked on (ie, when scrolling the map).

Tapping the gps pin (whether off screen or on) should center the gps on screen, but not lock on.

I volunteer to implement this, if it is accepted. I *might* be willing to implement it sooner than that so you (@westnordost) can play around with it and convince yourself it makes sense :)

westnordost:
> Maybe this option can be 😕 in the poll?

Ok

[...]

PR #2101, opened 2020-09-20
Hide gps button only when following position

smichel17:
Context in commit messages, where it belongs :)

To merge:
- [x] Get approval from @westnordost (#1996)
- [x] Make sure that this is the only place in the code that needs modification (ie, that I didn't miss anything)
- [x] Change the behavior of tapping on the location pin (just scroll to position, don't lock on)
- [ ] Decide if any refactoring should be done

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -308,7 +308,7 @@ class MainFragment : Fragment(R.layout.fragment_main),
     }
 
     override fun onClickedLocationMarker() {
-        setIsFollowingPosition(true)
+        mapFragment?.followPosition(true)
     }
 
     /* -------------------------- AbstractQuestAnswerFragment.Listener -------------------------- */
@@ -659,7 +659,7 @@ class MainFragment : Fragment(R.layout.fragment_main),
     }
 
     private fun onClickLocationPointer() {
-        setIsFollowingPosition(true)
+        mapFragment?.followPosition(true)
     }
 
     /* --------------------------------- Managing bottom sheet  --------------------------------- */
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -233,7 +233,10 @@ class MainFragment : Fragment(R.layout.fragment_main),
     /* ---------------------------------- MapFragment.Listener ---------------------------------- */
 
     override fun onMapInitialized() {
-        gpsTrackingButton.isActivated = mapFragment?.isFollowingPosition ?: false
+        val follow = mapFragment?.isFollowingPosition ?: false
+        val hide = follow && gpsTrackingButton.state !== LocationState.SEARCHING
+        gpsTrackingButton.isActivated = follow
+        gpsTrackingButton.visibility = if (hide) View.INVISIBLE else View.VISIBLE
         updateLocationPointerPin()
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -310,9 +310,7 @@ class MainFragment : Fragment(R.layout.fragment_main),
         }
     }
 
-    override fun onClickedLocationMarker() {
-        mapFragment?.followPosition(true)
-    }
+    override fun onClickedLocationMarker() {}
 
     /* -------------------------- AbstractQuestAnswerFragment.Listener -------------------------- */
 
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -566,6 +566,8 @@ class MainFragment : Fragment(R.layout.fragment_main),
         val mapFragment = mapFragment ?: return
         mapFragment.isFollowingPosition = follow
         gpsTrackingButton.isActivated = follow
+        val hideButton = follow && gpsTrackingButton.state !== LocationState.SEARCHING
+        gpsTrackingButton.visibility = if (hideButton) View.INVISIBLE else View.VISIBLE
         if (!follow) setIsCompassMode(false)
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
@@ -236,8 +236,8 @@ open class LocationAwareMapFragment : MapFragment() {
         return isFollowingPosition
     }
 
-    protected fun followPosition() {
-        if (!shouldCenterCurrentPosition()) return
+    public fun followPosition(forceCenter: Boolean = false) {
+        if (!forceCenter && !shouldCenterCurrentPosition()) return
         val controller = controller ?: return
         val targetPosition = displayedPosition ?: return
         controller.updateCameraPosition(600) {


westnordost:
- [x] Get approval from @westnordost (#1996)

It looks like the clear majority is in favour of such a compromise

westnordost (changes requested):
You need to update the button visibility always when:
1. the location state changes from/to `UPDATING`
2. the `gpsTrackingButton.isActivated` is changed

because the visibility of the button depends on these two values.

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -657,7 +659,7 @@ class MainFragment : Fragment(R.layout.fragment_main),
     }
 
     private fun onClickLocationPointer() {
-        setIsFollowingPosition(true)
+        mapFragment?.followPosition(true)
```
westnordost, line 662: Not sure what you are trying to do, but this looks like a hacky approach. See my other review comment.

smichel17: Trying to center the screen on your current position, but not lock on (that's reserved for the gps button).

I think it looks hacky because of the name of the function. It's `followPosition`, but really it just moves the view to be centered on your current position. Previously, it would only do that if the map fragment was set to following position (and this was a protected function; I made it public to enable calling it from the MainFragment); I added a switch to override that, to force a center even if it's not following.

I intentionally implemented this in few lines of code, but I'll talk more about that in reply to the review comment.

westnordost: Ah, I thought that tapping on the location pointer pin (etc) should also trigger the follow-mode.

smichel17: I thought it felt confusing/awkward to have the gps button disappear when tapping on the location pin. Also (much less important), I think sometimes it's nice to be able to return to your location without locking on.

smichel17: Note: *Currently* this happens when tapping on both the location pin (when I'm off screen) and marker (when on screen).

For the pin, it very much feels like the right thing to do / what I expect to happen.
For the marker, I think it feels a bit weird. I'm not sure it should do anything at all, actually.

smichel17:
When I first looked at `MainFragment.kt`, I thought it was one of the harder pieces of control flow to understand. Then I realized that it's basically just a container to hold and coordinate all the different parts. Each part holds its own state and they communicate between each other via callbacks on the main fragment. This made sense and, from there, I was able to figure out which pieces I needed to modify pretty quickly.

I mention this because my initial confusion came from the combination of:

- My *relative* lack of familiarity with custom xml attributes.
    - (and confusion about isActivated. Although in fairness, the Android devs are also [Um, yeah, deeply sorry about the terminology here](https://developer.android.com/reference/android/view/View#setActivated(boolean)))
- `LocationState` — it tracks whether the app has location permission, whether location is currently turned on, and whether it's searching/found; and it directly corresponds with which image/animation is displayed on the button.
- Location-related state being split between the map fragment and gps button; and its logic split between the map fragment and main fragment.

If I had unlimited time, I'd like to experiment with refactoring the location-handling code to be more concentrated in one spot. Probably something like:

- Add a new `LocationState`, `FOLLOWING`, for when the map is actually locked on.
    - Also clarify what `UPDATING` is — I think that's for when we have a location fix and we're subscribed to updates?
- Pull location state (and maybe compass mode) persistence up into the mainFragment, instead of having the map fragment and gps button each serialize their own state.
    - Alternatively, pass the gps button down to the map fragment somehow (dependency injection) and handle more of the gps logic there, instead. But I don't think this makes as much sense.
- There's a lot of `val mapFragment = mapFragment ?: return` and generally reaching down into the mapFragment or gps button to make decisions based on their state. It would be nice if we could do less of that.
    - If I *really* had a lot of time on my hands, I'd be interested in trying to refactor the map fragment so there's not so many levels of inheritance. Maybe pull the location-aware and quest-aware bits out into presenters that don't need to know anything about the view lifecycle.

I don't have unlimited time, however. And trying to change something you haven't put in the time to really understand is also a bad idea (I learned this lesson from my experience destroying a lot of nice architecture in Red Moon, which I thought was unnecessary over-engineering — and some of it *was*, but not all of it). So, for the changes currently implemented, my goals were:

- Implement just enough for a proof-of-concept, to play around with and see if you think this behavior makes sense.
    - This was what I meant by "Get approval from @westnordost", although I wasn't very clear on that.
- Change as little as possible, so each change is obvious (both, what it does and what I intended).
    - This is why I didn't rename the mapFragment's `followPosition` function — then I'd have to rename all the other places that call it, too -> more noise in the diff.

Then, ~~if~~ given that you like the direction, we can talk about what refactoring, if any, makes sense.

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
westnordost: You could remove that functionality from the marker.

westnordost:
Quite the wall of text.

> Location-related state being split between the map fragment and gps button

The gps button is just a view, it only retains its own state in order to be able to restore its instance state after it was recreated just like a text view retains its text and a radio button group retains which radio button is selected. It's the view state.
The location state in the map fragment is the actual location state that is being updated by the app's logic.

> Add a new LocationState, FOLLOWING, for when the map is actually locked on. 

Whether or not the map is locked on the location is a separate state: The app could still be searching for a location but the map should be locked on the location (once the location is found). So, you can press the GPS button while still searching to activate it, so the map will snap to your location once it does. Keeping this properties seperated also means that should the app temporarily loose the location info (f.e. location state rewerting to `ALLOWED`), the user preference to follow the position if it is available is not affected.

> There's a lot of val mapFragment = mapFragment ?: return and generally reaching down into the mapFragment or gps button to make decisions based on their state. It would be nice if we could do less of that. 

The way you say it, it sounds like this could be subject for refactor. Do you have an example?

> Then, if given that you like the direction, we can talk about what refactoring, if any, makes sense.

Ok, but well, let's first get it running. See https://github.com/westnordost/StreetComplete/pull/2101#pullrequestreview-492221773

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -566,6 +566,8 @@ class MainFragment : Fragment(R.layout.fragment_main),
         val mapFragment = mapFragment ?: return
         mapFragment.isFollowingPosition = follow
         gpsTrackingButton.isActivated = follow
+        val hideButton = follow && gpsTrackingButton.state !== LocationState.SEARCHING
```
smichel17, line 569: > You'll notice that the GPS button's location state is never read from the app logic, only updated.

Fair enough, but s/is/was/, with this implementation.

westnordost: Yes, see https://github.com/westnordost/StreetComplete/pull/2101#pullrequestreview-492221773

smichel17:
> You need to update the button visibility always when:
> 
> - the location state changes from/to UPDATING
> - the gpsTrackingButton.isActivated is changed

Done (for every place in MainActivity) — unless you also mean that the button should set its own visibility when restoring (it does not currently do this at all).

westnordost:
> unless you also mean that the button should set its own visibility when restoring

of course not, the button is a simple view

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
smichel17: Done (b8ad4d9). Leaving this conversation open since the function name is not updated yet.

smichel17: FWIW, upon reflection, I think the reason tapping the marker feels weird is that it's displayed flatly on the map, like the background, which is not tap-able; whereas the quest pins, location pin, and gps button all display as hovering above the map. So, making the marker tap-able breaks consistency. I think this is probably also my main problem with the live app's behavior.

smichel17:
It's annoying that there's only threading when discussing particular lines of code, and I can't start a comment on lines not in the changeset :( I considered writing this as a comment on some random lines so we can resolve it after, but that's annoying, too. Oh well.

> The way you say it, it sounds like this could be subject for refactor. Do you have an example?

Sure. Here's an example that:

- Is not related to the location. I'd like to better understand your intended architecture first, so I can make better suggestions.
- Is (hopefully) very clear-cut; it should either make sense or seem totally backwards. Which makes it useful for figuring out which way is intended.

Anyway, enough blabber. Here's the lines:

https://github.com/westnordost/StreetComplete/blob/784489a80838b79290401a9a4a1e822bd1f9af11/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L184-L201

IF the MainFragment is intended only to coordinate the pieces it contains, why does it have to know the current coordinates of the map here? Instead, we can just give the map the information it needs to adjust itself. This ends up reducing the number of cross-class function calls from 3 to 1. I also use a `let` to avoid the early return — scoping functions are easy to overuse but I think in this case it is an improvement:

> <details><summary>I would <em>personally</em> write it like this, but I was trying to stick to the existing style for an apples-to-apples comparison<br/>[Click to expand]</summary>
> 
> ```kotlin
>     // MainFragment
>     override fun onConfigurationChanged(newConfig: Configuration) {
>         super.onConfigurationChanged(newConfig)
>         mapFragment?.let { map ->
>             /* when rotating the screen and the bottom sheet is open, the view should not rotate
>                around its proper center but around the center of the part of the map that is not
>                occluded by the bottom sheet */
>             val previousOffset = mapOffsetWithOpenBottomSheet
>             updateMapQuestOffsets()
>             if (bottomSheetFragment != null) {
>                 map.adjustToOffsets(previousOffset, mapOffsetWithOpenBottomSheet)
>             }
>             updateLocationPointerPin()
>         }
>     }
> 
>     // MapFragment
>     public fun adjustToOffsets(oldOffset: RectF, newOffset: RectF) {
>         getViewPosition(oldOffset)?.let { pos ->
>             updateCameraPosition {
>                 position = getPositionThatCentersPosition(pos, newOffset)
>             }
>         }
>     }
> ```
> </details>

```kotlin
    // MainFragment
    override fun onConfigurationChanged(newConfig: Configuration) {
        super.onConfigurationChanged(newConfig)
        mapFragment?.let { map ->
            /* when rotating the screen and the bottom sheet is open, the view should not rotate
               around its proper center but around the center of the part of the map that is not
               occluded by the bottom sheet */
            if (bottomSheetFragment != null) {
                val previousOffset = mapOffsetWithOpenBottomSheet
                updateMapQuestOffsets()
                map.adjustToOffsets(previousOffset, mapOffsetWithOpenBottomSheet)
            } else {
                updateMapQuestOffsets()
            }
            updateLocationPointerPin()
        }
    }

    // MapFragment
    public fun adjustToOffsets(oldOffset: RectF, newOffset: RectF) {
        val oldPos = getViewPosition(oldOffset)
        if (oldPos != null) {
            val newPos = getPositionThatCentersPosition(oldPos, newOffset)
            updateCameraPosition { position = newPos }
        }
    }
```

On the other hand, if the map fragment is intended as more or less a simple view, then this would be the opposite direction of what we want — it shouldn't need to know anything about the other elements on screen around it, only the location it's supposed to display. (In that case, we could reduce api surface area by exposing the controller directly instead of proxying a bunch of function calls).

westnordost:
I'm terribly sorry, I accidentally changed your comment with my answer. I meant to post a new comment, here it is:

Your proposed change makes sense.

The map fragment is not intended as a simple view (it's not a view), but as the view controller for everything that concerns only the map. It is correct that the main fragment should only coordinate between the map, bottom sheet etc.

westnordost (commented):
I reviewed the code. However, since this spot is one of the more complex spots, I couldn't do it by just looking at the diff but had to browse through the code in the IDE as if I would want to implement it myself. 

I ended up implementing it myself, then comparing it to your version. There are a few things I did differently:
1. not extend the `followPosition` method but instead have another public function called `centerCurrentPosition`, I think this is better style
2. Do not rely on the `gpsTrackingbutton.state !== SEARCHING` but instead `mapFragment.displayedLocation != null` (= map fragment shows the location)
3. removed `onClickedLocationMarker` callback completely. No need to keep methods that are not used
4. `onLocationChanged` marks the time where the position first becomes available. This can have an effect on whether the tracking button should be visible or not, you missed that spot

I created a new PR here: 2110

smichel17:
> 1. not extend the `followPosition` method but instead have another public function called `centerCurrentPosition`, I think this is better style

Agreed. (Although I don't like the name `followPosition` in the first place; the functionality is more like `centerCurrentPositionIfFollowing`)

> 2. Do not rely on the `gpsTrackingbutton.state !== SEARCHING` but instead `mapFragment.displayedLocation != null` (= map fragment shows the location)

Ooh, that's a much nicer way of doing it; I didn't notice that variable before.

> 3. removed `onClickedLocationMarker` callback completely. No need to keep methods that are not used

:+1:

> 4. `onLocationChanged` marks the time where the position first becomes available. This can have an effect on whether the tracking button should be visible or not, you missed that spot

Actually, it was a logic error, not missed. I was thinking you'd never be able to reach that code unless you had already pressed the location button, in which case you are already following… which is true, if, like me, you leave your gps off 100% of the time unless an app requests it (ie, because I pressed the location button). But, I realize now, it also happens if your gps is already on or you turn it on from a different app (or a quick tile).

smichel17:
Your code is better quality than mine; closing in favor of #2110

westnordost:
> Agreed. (Although I don't like the name followPosition in the first place; the functionality is more like centerCurrentPositionIfFollowing)

Well, it's a private function, not intended to be on the interface. But you are right.

PR #2104, opened 2020-09-21
Emphasise being lit at night

peternewman:
Prompt people into checking the menu if it's lit 24/7, I hadn't done so, or had forgotten, so miss-tagged a number of places which are lit 24/7.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -278,9 +278,9 @@ The info you enter is then directly added to the OpenStreetMap in your name, wit
 	<string name="quest_toiletsFee_title">"Does this restroom require a fee?"</string>
     <string name="quest_baby_changing_table_toilets_title">"Does this restroom have a baby changing table?"</string>
     <string name="quest_baby_changing_table_title">"Does %s have a baby changing table?"</string>
-    <string name="quest_way_lit_title">"Is this stretch of way lit?"</string>
-    <string name="quest_way_lit_named_title">"Is %s lit here?"</string>
-    <string name="quest_way_lit_road_title">"Is this stretch of road lit?"</string>
+    <string name="quest_way_lit_title">"Is this stretch of way lit at night?"</string>
+    <string name="quest_way_lit_named_title">"Is %s lit at night here?"</string>
+    <string name="quest_way_lit_road_title">"Is this stretch of road lit at night?"</string>
     <string name="quest_way_lit_automatic">"It uses motion detection"</string>
     <string name="quest_way_lit_24_7">"It’s lit night and day"</string>
     <string name="quest_tactilePaving_title_bus">"Does this bus stop have tactile pavings?"</string>


westnordost:
So to survey if a street has lanterns, you'd expect a surveyor to sit and wait on one road 24 hours to check if it is lit all day and all night?

peternewman:
No, but there are other answers than yes:
https://wiki.openstreetmap.org/wiki/Key:lit

Indeed they're even populated in the other answers bit of SC. As mentioned I've miss-tagged a lot of tunnels and bits of subway/underground with lit=yes when it seems they should actually be lit=24/7.

> So to survey if a street has lanterns, you'd expect a surveyor to sit and wait on one road 24 hours to check if it is lit all day and all night?

Presumably you don't expect people to only tag lit=yes at night, and technically they could be lit=disused, although I don't know how you differentiate that from a theoretical lit=broken apart from I guess a whole street is unlikely to be broken.

I'm simply trying to suggest a way to nudge people into considering how tunnels should be tagged.

westnordost:
I consider the current status of the street lighting (24/7, disabled, broken, disused, limited, specific lit-hours ...) a separate property. 

Just dandily adding more possible values to `lit` is problematic because it implicitly redefines the meaning of the `lit=yes/no` key.

westnordost:
I was a little hasty on answering. So what exactly do you propose? A change in wording?

HolgerJeromin:
> As mentioned I've miss-tagged a lot of tunnels and bits of subway/underground with lit=yes when it seems they should actually be lit=24/7.

IMO an `tunnel=yes` with ` lit=yes` should be interpreted by the data consumer to 24/7. At least I do.

westnordost:
> I was a little hasty on answering. So what exactly do you propose? A change in wording?

Ah, too hasty again. This is a PR and not an issue ticket. Well, the problem with your suggestion is that it seems to suggest that the user should survey this at night. But that's not required and not particularly encouraged.

peternewman:
> I consider the current status of the street lighting (disabled, disused, limited, specific lit-hours ...) a separate property.
> 
> Just adding more and more possible values to `lit` is problematic because it implicitly redefines the meaning of the `lit=yes/no` key.

TBH I'd probably agree with you, although we're already in this mess, lit=yes has been used to mean lit at night, and 24/7 suggested as always since at least 2009:
https://wiki.openstreetmap.org/w/index.php?title=Key:lit&direction=next&oldid=322085

I guess this is similar to your speed issues and the downside of allowing people to tag things in a very loose and free-form way.

> > As mentioned I've miss-tagged a lot of tunnels and bits of subway/underground with lit=yes when it seems they should actually be lit=24/7.
> 
> IMO an `tunnel=yes` with ` lit=yes` should be interpreted by the data consumer to 24/7. At least I do.

Tunnel and lit are mentioned on the tunnel page but not the lit page although it doesn't cover the 24/7 thing, but that seems a fairly sensible assumption:
https://wiki.openstreetmap.org/wiki/Key:tunnel#How_to_map

Do you do the same for indoor too? Some of my examples were tunnels, and others were enclosed walkways/bridges. I'm not sure it's so clear-cut for indoors (or even covered) as it rather depends on usage and how open it is. An example I'm thinking of has glass brick walls but is lit all the time (probably simpler than controlling the lighting).

> > I was a little hasty on answering. So what exactly do you propose? A change in wording?
> 
> Ah, too hasty again. This is a PR and not an issue ticket.

Apologies, I figured a PR might be slightly easier than an issue and reduce your workload slightly, I can just create issues in future if that's preferred?

> Well, the problem with your suggestion is that it seems to suggest that the user should survey this at night. But that's not required and not particularly encouraged.

Yeah I was trying to nudge people into going "it's lit during the day">other answers>"Oh it's lit 24/7".

I certainly generally survey the lit status during the day (i.e. does it have lampposts), but if you do so, how do you confirm it's actually lit during the night and not disused or automatic? There are a few bits of UK motorway where they've turned off the lighting permanently and I think then eventually removed it.

westnordost:
> how do you confirm it's actually lit during the night and not disused or automatic

I won't, for me, `lit` denotes the physical existance of street lighting cause that's the thing that is easily verifiable and not subject to (quick) change.

kmpoppe:
I was like ... I've had seen this discussion before, what am I missing?! We had been talking about it in t.me/OSM_de.

In GER and AUT ONLY, there is a thing called "Laternenring" (https://de.wikipedia.org/wiki/Laternenring), also described in the German OSM wiki at https://wiki.openstreetmap.org/wiki/DE:Key:lit#Laternenring.

So, when being asked for `lit=*` in GER and AUT an answer could be "not the whole night" and the set `lit=limited` which is the preferred tagging named in the German OSM wiki. Would that be what you thought about @peternewman ?

Kai

westnordost:
That would mean that everyone who wants to map `lit` in Germany and Australia needs to know about the "Laternenring". And for other places, maybe other restrictions and/or markers to denote that the lanterns are somehow special.

peternewman:
I don't think there are markers in the UK @kmpoppe , but I know in some areas they're starting to turn them off or dim them at midnight or 1am to save money. This PR was simply driven by me going into other answers and realising I'd miss-tagged a bunch of indoor and underground lit places and trying to make that more obvious for myself and others.

westnordost:
I acknowledge your argument that the way the question is phrased and if the way is lit day and night (tunnel), users might not consider even looking into the other answers menu. A solution could be to also show "Yes, night and day" amongst "yes" and "no" but I think this would not fit into the bottom button bar.
So, then the form would need to be changed to be like the parking access form, with 4 radio buttons (yes, no, lit night and day, motion detection). Then, of course, always 2 taps are necessary to answer that question.

What do you think of that?

smichel17:
> Then, of course, always 2 taps are necessary to answer that question.

You can lower it by pre-selecting plain "yes" (or "no"), so picking *that answer* is only one tap. Then it's 1.x taps, where x depends on how often the pre-selected one is the correct answer.

> show "Yes, night and day" amongst "yes" and …

So, you're saying the difference between those two would be *night and day*? :monkey_face: :drum: :drum: :boom: :roll_eyes:

peternewman:
Would "Yes"/"No"/"24/7" fit in the existing bar as one row @westnordost ?

> You can lower it by pre-selecting plain "yes" (or "no"), so picking _that answer_ is only one tap. Then it's 1.x taps, where x depends on how often the pre-selected one is the correct answer.

I thought that was frowned upon in SC to avoid accidental answers or deliberate gaming of the system.

westnordost:
> I thought that was frowned upon in SC to avoid accidental answers or deliberate gaming of the system.

Yes

> Would "Yes"/"No"/"24/7" fit in the existing bar as one row @westnordost ?

Yes. Would it be clear though what is meant with that answer?

smichel17:
I have a much better idea, which preserves the existing one-click options and displays the 24/7 prominently:

image

peternewman:
> > Would "Yes"/"No"/"24/7" fit in the existing bar as one row @westnordost ?
> 
> Yes. Would it be clear though what is meant with that answer?

I would sat yes :smile: , but clearly wants some other input. I thought it was used in the lit quest already, but I see it says night and day. The opening hours quest offers "Open 24/7" in Other answers, so hopefully! :shrug: 

> I have a much better idea, which preserves the existing one-click options and displays the 24/7 prominently:
> 
> (I didn't think of it when I made the image, but "24/7" should probably be "Always" instead)

I'm a bit confused by the interface @smichel17 , I'd have thought never has to be in the radio buttons if you're going down that route, and yes then becomes submit, otherwise something that's lit during the night is **never** lit _sometimes_ too. You could then add "only when someone is there" or something for motion detection.

Although I thought the preference was to try and avoid listing ALL the answers on the main page.

smichel17:
@peternewman That's how the dialog would appear when opened. If the way is not lit, you pick never, and the radio buttons do not matter. If you pick "yes", then it selects the answer based on which of the radio buttons is currently selected.

That is, if you ignore the radio buttons entirely, it functions exactly the same as the current quest. To activate the new option, select the "24/7" radio button and then click "yes".

---

A variant would be to keep the answers as yes/no and change the radio buttons to a single checkbox, unchecked by default,

- [ ] The way is lit 24/7

If the user checks it, the "no" answer disappears.

westnordost:
I find your suggestion highly confusion to be honest. The same with the variant you suggested in your last comment.

smichel17:
Would it be less confusing if this text were above the radio buttons? "If it is lit, how often?"

peternewman:
My objection in this case, and in general, is people abusing radio buttons. With your initial suggestion, logically what should happen is I press never, and then **both** radio buttons go grey/disabled, a pause so I can see, and then the form submits (or ideally another button to submit). Or a separate button/form to submit the radio buttons.

Otherwise from a form perspective you are selecting:
Sometimes or 24/7
Never or Yes

> @peternewman That's how the dialog would appear when opened. If the way is not lit, you pick never, and the radio buttons do not matter.

How does the user tell the radio buttons don't matter? There is no good solution for that with radio buttons, if you start with neither selected, it looks like I need to press one before I press never (and you're back to the extra click), your current default is confusing as mentioned. The only sensible solution to that is to make never a radio button and yes submit as I suggested, but then yes is one clicks and no and 24/7 are two.

This is similar to all the web forms that abuse them by trying to do the same radio-button style behaviour with checkboxes instead. I agree they are the right selection method (as we should be picking exactly one), but that pick needs to include never too.

westnordost:
I didn't check this discussion in the last days because it was not an issue but a PR.

I'd say, moving "24/7" from the other answers menu to another option on the button bar would be ok. The only problem I see is that "24/7" is primarily known in English. In German, you'd probably rather say "rund um die Uhr, sieben Tage die Woche" (round the clock, seven days a week) or "immerzu" (incessantly).
So maybe in this case, a confirmation dialog should be shown that describes in words what this means, maybe using the same string as is used currently in the other answers menu.

I'll close this PR and you can open a new one that implements this if you like.

PR #2109, opened 2020-09-24
Multiple collection times on the same weekdays as selected before

### LINKED ISSUE #2108: Allow for multiple postbox collection times on the same day
**Use case**
When being tasked to add collection times for a postbox or opening hours for a business, the "standard" behaviour for the weekday picker is that weekdays, that were already selected are no longer activated.

In the case of the postbox I worked with today (https://www.openstreetmap.org/node/2388572099), I had to re-selected Mo-Fr three times to add 4 collection times during the week.

Also, in the not so progressive German business landscape there are still those, who have a lunch break - which leads to the same problem, needing to re-select the weekdays.

**Proposed Solution**
The `TimePickerDialog` (collection times) and `TimeRangePickerDialog` (opening hours) ought to display a checkbox "add more times for the same weekdays" that you can check (default: unchecked), which restarts the dialog to add another Time or TimeRange.

I haven't got enough experience with the app development yet, to propose a code change for this myself and would require input.

Kai

kmpoppe:
Closes #2108.

Quest for adding collection times to `amenity=post_box` nodes was changed.

Just like the quest for adding opening times, this quest will now, if there already is a row, ask if you want to add a collection time to the same weekday(s) as the row before or if you want to add more weekdays.

`CollectionTimesAdapter.addNew()` was intentionally left in the code (I'm not sure if Android Studio correctly finds all references) bur calls `addNewWeekdays()`. If this should be removed, please advise @westnordost.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
@@ -1,9 +1,11 @@
 package de.westnordost.streetcomplete.quests.postbox_collection_times
 
 import android.os.Bundle
+import android.view.Menu
 import androidx.appcompat.app.AlertDialog
 import androidx.recyclerview.widget.LinearLayoutManager
 import android.view.View
+import androidx.appcompat.widget.PopupMenu
 import androidx.recyclerview.widget.RecyclerView
 
 import java.util.ArrayList
@@ -52,7 +54,29 @@ class AddCollectionTimesForm : AbstractQuestFormAnswerFragment<CollectionTimesAn
         collectionTimesList.isNestedScrollingEnabled = false
         checkIsFormComplete()
 
-        addTimesButton.setOnClickListener { collectionTimesAdapter.addNew() }
+        addTimesButton.setOnClickListener { onClickAddButton(it) }
+    }
+
+    private fun onClickAddButton(v: View) {
+        val rows = collectionTimesAdapter.collectionTimesRows
+
+        val addTimeAvailable = rows.isNotEmpty() && rows.last() is WeekdaysTimesRow
+
+        if (addTimeAvailable) {
+            val popup = PopupMenu(requireContext(), v)
+            if (addTimeAvailable) popup.menu.add(Menu.NONE, 0, Menu.NONE, R.string.quest_openingHours_add_hours)
+            popup.menu.add(Menu.NONE, 1, Menu.NONE, R.string.quest_openingHours_add_weekdays)
+            popup.setOnMenuItemClickListener { item ->
+                when(item.itemId) {
+                    0 -> collectionTimesAdapter.addNewHours()
+                    1 -> collectionTimesAdapter.addNewWeekdays()
+                }
+                true
+            }
+            popup.show()
+        } else {
+            collectionTimesAdapter.addNewWeekdays()
+        }
     }
 
     private fun loadCollectionTimesData(savedInstanceState: Bundle?):List<WeekdaysTimesRow> =
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
@@ -60,7 +60,7 @@ class AddCollectionTimesForm : AbstractQuestFormAnswerFragment<CollectionTimesAn
     private fun onClickAddButton(v: View) {
         val rows = collectionTimesAdapter.collectionTimesRows
 
-        val addTimeAvailable = rows.isNotEmpty() && rows.last() is WeekdaysTimesRow
+        val addTimeAvailable = rows.isNotEmpty()
 
         if (addTimeAvailable) {
             val popup = PopupMenu(requireContext(), v)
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
```diff
@@ -52,7 +54,29 @@ class AddCollectionTimesForm : AbstractQuestFormAnswerFragment<CollectionTimesAn
...
         checkIsFormComplete()
 
-        addTimesButton.setOnClickListener { collectionTimesAdapter.addNew() }
+        addTimesButton.setOnClickListener { onClickAddButton(it) }
+    }
+
+    private fun onClickAddButton(v: View) {
+        val rows = collectionTimesAdapter.collectionTimesRows
+
+        val addTimeAvailable = rows.isNotEmpty() && rows.last() is WeekdaysTimesRow
```
westnordost, line 63: all rows of the `CollectinTimesAdapter` are `WeekdaysTimesRow`s, no need to check for that

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/CollectionTimesAdapter.kt
```diff
@@ -64,13 +64,25 @@ class CollectionTimesAdapter(
...
         val isFirst = collectionTimesRows.isEmpty()
         openSetWeekdaysDialog(getWeekdaysSuggestion(isFirst)) { weekdays ->
             openSetTimeDialog(12 * 60) { minutes ->
                 add(weekdays, minutes) }
         }
     }
 
+    fun addNewHours() {
+        val rowAbove = if (collectionTimesRows.size > 0) collectionTimesRows[collectionTimesRows.size - 1] else null
+        if (rowAbove !is WeekdaysTimesRow) return
```
westnordost, line 80: all rows of the `CollectinTimesAdapter` are `WeekdaysTimesRow`s, no need to check for that

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/CollectionTimesAdapter.kt
```diff
@@ -64,13 +64,25 @@ class CollectionTimesAdapter(
     }
 
     fun addNew() {
+        addNewWeekdays()
+    }
+
```
westnordost, line 69: Yes, you can delete that function

westnordost (changes requested):
Looks fine, thank you! I just have a few small comments.

Did you test it?

kmpoppe:
> Looks fine, thank you! I just have a few small comments.

Will change them in the morning, thanks.

> Did you test it?

I tested it on my phone from the Dev app that Android studio runs on the phone, yes. The UI works fine, but I don't know how to test if the changes really check out in reality. What's your way of checking if then changes you made create the desired result in the OSM Data?

westnordost:
> I tested it on my phone from the Dev app that Android studio runs on the phone, yes. The UI works fine, but I don't know how to test if the changes really check out in reality. What's your way of checking if then changes you made create the desired result in the OSM Data?

I run the app without being logged in, then I solve a quest and look at the log in Android Studio (filter by "QuestController" to remove the noise), there you can see what will be tagged.

kmpoppe:
> I run the app without being logged in, then I solve a quest and look at the log in Android Studio (filter by "QuestController" to remove the noise), there you can see what will be tagged.

Nice touch, thanks! Next enhancement: display a message box with the changes if app is in debug mode 😬

westnordost:
Right, good idea, I implemented it. It is in the settings menu -> show quest forms (last entry)

kmpoppe:
> Right, good idea, I implemented it. It is in the settings menu -> show quest forms (last entry)

96be938?

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/CollectionTimesAdapter.kt
kmpoppe: Removed function with e2a6116

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/CollectionTimesAdapter.kt
kmpoppe: Resolved with e2a6116, yet I had to still check for `null` in place, otherwise Android Studio was complaining, that `rowAbove.weekdays` might not work.
https://github.com/westnordost/StreetComplete/pull/2109/files#diff-e6478a4cea0aeba97f040ea46df5f636R76
I think that's a reasonable precaution, right?

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
kmpoppe: Resolved with e2a6116

kmpoppe:
> Did you test it?

image
image

Looks good to me. Thanks for the great testing method, make sure the other Contributors get to know about it, it'll be very handy!

kmpoppe:
@westnordost I just realized that you already put
`and (!collection_times or collection_times older today -${r * 2} years)`
in the tagFilter, without there being a proper resurvey question operation. Do you want to keep it that way for now?

westnordost:
Yes. It makes sense to revisit collection times every few years. To implement that previous collection times are shown was something I deemed to be too much effort for the gain. See https://github.com/westnordost/StreetComplete/issues/2105

westnordost (approved):

westnordost:
Looks good, merging!

PR #2119, opened 2020-09-28
Don't ask for benches if there is no shelter/cover

peternewman:
Potentially sitting in the rain has a high spam rate (e.g. looking at shelter=no there are 144 without benches and 6 with in part of Southern England)

So only ask if there is a shelter or a cover of some sort. I think I've got the logic right.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
@@ -17,6 +17,7 @@ class AddBenchStatusOnBusStop(o: OverpassMapDataAndGeometryApi, r: ResurveyInter
           (highway = bus_stop and public_transport != stop_position)
         )
         and physically_present != no and naptan:BusStopType != HAR
+        and (shelter != no or covered != no)
         and (!bench or bench older today -${r * 4} years)
     """
 


westnordost:
Only because the probability is lower doesn't mean that surveyors should not check for that. It is a quest that can be answered very quickly with just one tap, so even if in 99% of the time, the answer will be the same, it is not really a waste of time. Also, 6/144 is still around 5% by the way.

Additionally, with the same logic you apply for this PR, you could also open another PR to not ask for benches if it **has** a shelter, because most shelters will have benches.

peternewman:
> Only because the probability is lower doesn't mean that surveyors should not check for that. It is a quest that can be answered very quickly with just one tap, so even if in 99% of the time, the answer will be the same, it is not really a waste of time. Also, 6/144 is still around 5% by the way.

I found the count command, so was able to run over the whole UK:
https://overpass-turbo.eu/s/Yv6

It still averages about 4% there for shelter yes/no (I didn't check covered too).

> Additionally, with the same logic you apply for this PR, you could also open another PR to not ask for benches if it **has** a shelter, because most shelters will have benches.

I'd argue that knowing there **is** a bench is more useful than knowing there definitely isn't one (as opposed to the fact there might be one and no-one has checked). As you would plan your route finding the closest bench=yes (rather than bench!=no), compared to tactile paving say, where you may **have** to use that bus stop, so knowing there is no paving there to help locate it is useful.

westnordost:
Well, it means that if you happen to find a bench on a non-sheltered bus stop somewhere, the information value for that is much higher because it is an unexpected situation. Also, it's rainy in the UK, in other countries further south, shelters are less common because it's not that rainy. The shadow of a tree or nearby wall might be enough of a shelter. The bench-situation will be not that strongly connected to the shelter-situation everywhere. I'll close this.

peternewman:
I hadn't considered the tree aspect; I figured there would be very few places where it would be pleasant to be seated outside throughout the year either due to rain, wind or sun.

I'd also suggest that if the bench isn't part of a shelter isn't it just a bench near a bus stop, rather than **part of** it.

Fair enough though.

PR #2124, opened 2020-09-30
Remove warehouses and barns from building levels quest

Etua:
From my experience these types of buildings usually have levels indistinguishable from outside so asking for them is counterproductive and violates quest guidelines about being answerable without entering someone's property.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
@@ -35,5 +35,5 @@ private val BUILDINGS_WITH_LEVELS = arrayOf(
     "house","residential","apartments","detached","terrace","dormitory","semi",
     "semidetached_house","bungalow","school","civic","college","university","public",
     "hospital","kindergarten","transportation","train_station", "hotel","retail",
-    "commercial","office","warehouse","industrial","manufacture","parking","farm",
-    "farm_auxiliary","barn","cabin")
+    "commercial","office","industrial","manufacture","parking","farm","farm_auxiliary",
+    "cabin")


matkoniecz:
In my experience warehouses ten to be very obvious.

westnordost:
Oh, I overlooked this PR.

I think you have a point, @Etua. Historic warehouses and barns may have several easily distinguisable storeys, but modern ones are mostly just roofed forklift maneuvering spaces that consist simply of 1 storey.
On the other hand, wouldn't it be easy to simply answer "1" in this case?

Etua:
By "obvious" you mean that you mark windowless warehouses as having one level or do you encounter them rarely so that's not the issue? From my experience most of the newly built warehouses may be quite tall but you don't really know what's inside unless you enter them.

Dnia 1 października 2020 15:41:32 CEST, Mateusz Konieczny <notifications@github.com> napisał(a):
>In my experience warehouses tend to be very obvious.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/westnordost/StreetComplete/pull/2124#issuecomment-702143438
Pozdrawiam
Dominik Danelski

matkoniecz:
> do you encounter them rarely so that's not the issue

Yes. At least during OSM mapping. (I may be wrong, my exposure to OSM mapping is primarily city center of an European city and mountains/forests - both unlikely to contain a modern warehouse)

Etua:
@westnordost  "On the other hand, wouldn't it be easy to simply answer "1" in this case?"

Probably it would be a valid solution for most cases but you still have to make an assumption which isn't necessary for other quests and that case may be confusing. I wouldn't mind closing this PR if you think that the current state is OK, but after encountering a few dozen such buildings and uncertainty each time I thought that this should be discussed.

matkoniecz:
> but after encountering a few dozen such buildings and uncertainty each time I thought that this should be discussed

Now after thinking a bit more (I tend to see such warehouses out of a bus, outside cities) - I would be also confused while tagging them. So I guess that their removal makes sense.

westnordost:
Ok, let's remove it then. Thank you for being patient with me and @matkoniecz

PR #2143, opened 2020-10-08
Reorder the step quests for optimal gathering

peternewman:
Even if not travelling along the way itself.

Similar to #1583 I think this ordering will get the most step data gathered by the largest group of people.


westnordost (approved):
makes sense

PR #2148, opened 2020-10-09
Untested: Improve the username matching in the changelog

peternewman:
I've only checked the regex here: https://www.regexplanet.com/advanced/java/index.html

There's a stricter regex here:
https://github.com/shinnn/github-username-regex#githubusernameregex

Currently @CJ-Malone's mention doesn't work in v24.1

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
@@ -80,7 +80,7 @@ private fun addedLinks(description: String): String {
             val issue = matchResult.groupValues[1]
             "<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
         }
-        .replace(Regex("(?<=[\\s(]|^)@([a-zA-Z\d-]+)")) { matchResult ->
+        .replace(Regex("(?<=[\\s(]|^)@([a-zA-Z\\d-]+)")) { matchResult ->
             val contributor = matchResult.groupValues[1]
             "<a href=\"https://github.com/$contributor\">@$contributor</a>"
         }
--- a/app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
@@ -76,15 +76,12 @@ private fun readChangelog(resources: Resources) =
 
 private fun addedLinks(description: String): String {
     return description
-        .replace(Regex("([\\s,(])#([0-9]+)")) { matchResult ->
-            val s = matchResult.groupValues[1]
-            val issue = matchResult.groupValues[2]
-            "$s<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
+        .replace(Regex("(?<=[\\s(]|^)#(\\d+)")) { matchResult ->
+            val issue = matchResult.groupValues[1]
+            "<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
         }
-        .replace(Regex("([\\s,(])@(\\w[\\-\\w]*)([\\s,.\\)']|$)")) { matchResult ->
-            val s = matchResult.groupValues[1]
-            val contributor = matchResult.groupValues[2]
-            val e = matchResult.groupValues[3]
-            "$s<a href=\"https://github.com/$contributor\">@$contributor</a>$e"
+        .replace(Regex("(?<=[\\s(]|^)@([a-zA-Z\d-]+)")) { matchResult ->
+            val contributor = matchResult.groupValues[1]
+            "<a href=\"https://github.com/$contributor\">@$contributor</a>"
         }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
@@ -81,7 +81,7 @@ private fun addedLinks(description: String): String {
             val issue = matchResult.groupValues[2]
             "$s<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
```diff
@@ -76,14 +76,15 @@ private fun readChangelog(resources: Resources) =
...
 private fun addedLinks(description: String): String {
     return description
-        .replace(Regex("([\\s,(])#([0-9]*)")) { matchResult ->
+        .replace(Regex("([\\s,(])#([0-9]+)")) { matchResult ->
             val s = matchResult.groupValues[1]
             val issue = matchResult.groupValues[2]
             "$s<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
         }
-        .replace(Regex("([\\s,(])@(\\w*)")) { matchResult ->
+        .replace(Regex("([\\s,(])@(\\w[\\-\\w]*)([\\s,.\\)']|$)")) { matchResult ->
```
westnordost, line 84:
```suggestion
        .replace(Regex("([\\s,(])@([\\w-]+)")) { matchResult ->
```

I think this is enough. See

https://regexper.com/#%28%5B%5Cs%2C%28%5D%29%40%28%5B%5Cw-%5D%2B%29

Yes this is more than is allowed, but the changelog is written by me, so I won't include invalid github user names because those can't contribute

app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
```diff
@@ -76,14 +76,15 @@ private fun readChangelog(resources: Resources) =
...
             val s = matchResult.groupValues[1]
             val contributor = matchResult.groupValues[2]
-            "$s<a href=\"https://github.com/$contributor\">@$contributor</a>"
+            val e = matchResult.groupValues[3]
+            "$s<a href=\"https://github.com/$contributor\">@$contributor</a>$e"
```
westnordost, line 88:
```suggestion
            "$s<a href=\"https://github.com/$contributor\">@$contributor</a>"
```

see above

app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
peternewman: I mean by that argument, this will work too (most stuff includes _ in \w, which isn't in the GitHub list according to that repo):
```suggestion
        .replace(Regex("([\\s(])@([a-zA-Z\d-]+)")) { matchResult ->
```

As I can't see any reason you'd want `,@peternewman` gramatically it should be e.g. `@westnordost, @peternewman`.

We could add the `{0,38}` bit in, or are you not bothered about making it strict at all? I wasn't particularly expecting invalid usernames, but just it e.g. matching on other things by mistake.

Having looked at the current changelog, there is just `@username.` or `@username</li>` currently.

app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
peternewman: Yep agreed, pending discussion on the other bit.

app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
```diff
@@ -76,14 +76,15 @@ private fun readChangelog(resources: Resources) =
 
 private fun addedLinks(description: String): String {
     return description
-        .replace(Regex("([\\s,(])#([0-9]*)")) { matchResult ->
+        .replace(Regex("([\\s,(])#([0-9]+)")) { matchResult ->
```
peternewman, line 79: Actually by the same argument I think `,` is redundant in the before match here too.

We can go one better in fact and use a positive lookbehind to not even need the s bit:
`(?<=[\s(]|^)#(\d+)`

Your site can't visualise it, but this one can:
https://www.debuggex.com/

westnordost:
Ok I'll merge this PR after you tested it. I don't know if lookbehind is supported by java regex.

peternewman:
It's specifically listed here, is that sufficient:?
https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html
and here:
https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/regex/Pattern.html

westnordost:
It needs to be tested. If you don't, I'll have to.

peternewman:
Fair enough @westnordost , I'm unlikely to get to it for at least the next three days unfortunately and would need to install the IDE etc and get it building at that point. Equally I'm happy for it to wait if you're busy as it's a fairly low priority fix.

app/src/main/java/de/westnordost/streetcomplete/about/ChangelogFragment.kt
```diff
@@ -76,14 +76,12 @@ private fun readChangelog(resources: Resources) =
...
-            "$s<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
+        .replace(Regex("(?<=[\\s(]|^)#(\\d+)")) { matchResult ->
+            val issue = matchResult.groupValues[1]
+            "<a href=\"https://github.com/westnordost/StreetComplete/issues/$issue\">#$issue</a>"
         }
-        .replace(Regex("([\\s,(])@(\\w*)")) { matchResult ->
-            val s = matchResult.groupValues[1]
-            val contributor = matchResult.groupValues[2]
-            "$s<a href=\"https://github.com/$contributor\">@$contributor</a>"
+        .replace(Regex("(?<=[\\s(]|^)@([a-zA-Z\d-]+)")) { matchResult ->
```
westnordost, line 83:
```suggestion
        .replace(Regex("(?<=[\\s(]|^)@([a-zA-Z\\d-]+)")) { matchResult ->
```
syntax error in regex

peternewman: Oops, thanks @westnordost !

peternewman:
Thanks, sorry for the error and that I didn't get round to testing it.

westnordost:
No problem, I got around to testing it now

PR #2155, opened 2020-10-10
Refactor map reposition on rotation interface

smichel17:
Previously discussed in https://github.com/westnordost/StreetComplete/pull/2101#issuecomment-698393593

I have checked that this compiles, but not tested in-app.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -183,19 +183,14 @@ class MainFragment : Fragment(R.layout.fragment_main),
 
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
-        val mapFragment = mapFragment ?: return
-        /* when rotating the screen and the bottom sheet is open, the view should not rotate around
-           its proper center but around the center of the part of the map that is not occluded by
-           the bottom sheet */
+        val mapFragment = this.mapFragment ?: return
+        /* when rotating the screen and the bottom sheet is open, the view
+           should not rotate around its proper center but around the center
+           of the part of the map that is not occluded by the bottom sheet */
+        val previousOffset = mapOffsetWithOpenBottomSheet
+        updateMapQuestOffsets()
         if (bottomSheetFragment != null) {
-            val currentPos = mapFragment.getViewPosition(mapOffsetWithOpenBottomSheet)
-            updateMapQuestOffsets()
-            if (currentPos != null) {
-                val offsetPos = mapFragment.getPositionThatCentersPosition(currentPos, mapOffsetWithOpenBottomSheet)
-                mapFragment.updateCameraPosition { position = offsetPos }
-            }
-        } else {
-            updateMapQuestOffsets()
+            mapFragment.adjustToOffsets(previousOffset, mapOffsetWithOpenBottomSheet)
         }
         updateLocationPointerPin()
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
@@ -350,6 +350,14 @@ open class MapFragment : Fragment(),
 
     /* ------------------------------- Controlling the map -------------------------------------- */
 
+    public fun adjustToOffsets(oldOffset: RectF, newOffset: RectF) {
+        controller?.screenCenterToLatLon(oldOffset)?.let { pos ->
+            controller?.updateCameraPosition {
+                position = controller?.getLatLonThatCentersLatLon(pos, newOffset)
+            }
+        }
+    }
+
     fun getPositionAt(point: PointF): LatLon? = controller?.screenPositionToLatLon(point)
 
     fun getPointOf(pos: LatLon): PointF? = controller?.latLonToScreenPosition(pos)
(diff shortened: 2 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -183,21 +183,17 @@ class MainFragment : Fragment(R.layout.fragment_main),
...
-                val offsetPos = mapFragment.getPositionThatCentersPosition(currentPos, mapOffsetWithOpenBottomSheet)
-                mapFragment.updateCameraPosition { position = offsetPos }
+            /* when rotating the screen and the bottom sheet is open, the view
+               should not rotate around its proper center but around the center
+               of the part of the map that is not occluded by the bottom sheet */
+            if (bottomSheetFragment != null) {
+                map.adjustToOffsets(previousOffset, mapOffsetWithOpenBottomSheet)
             }
-        } else {
-            updateMapQuestOffsets()
```
smichel17, line 198: Trade-off here:

- Before, we called `updateMapQuestOffsets` in both branches of the check if the bottomSheetFragment is open. I didn't like this; I'd prefer to see at a glance that the offset is always updated.
- Now, I moved it outside the check. However, this means we always need to save the previous offset position, even though it's only used if the bottom sheet is open.

Overall I think this is worth it, but just wanted to point out it's a place where you might disagree.

smichel17: Also, I moved the line breaks in the comment so it stays under 80 characters per line (from 100).
In general, I think that 100 makes sense, but when I can stay at 80 characters without making other sacrifices, I prefer that. In this case, the comment takes 3 lines, either way.

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
```diff
@@ -350,6 +350,14 @@ open class MapFragment : Fragment(),
 
     /* ------------------------------- Controlling the map -------------------------------------- */
 
+    public fun adjustToOffsets(oldOffset: RectF, newOffset: RectF) {
+        controller?.screenCenterToLatLon(oldOffset)?.let { pos ->
+            controller?.updateCameraPosition {
+                position = controller?.getLatLonThatCentersLatLon(pos, newOffset)
+            }
+        }
```
smichel17, line 358: I changed these to call the controller methods directly instead of the MapFragment wrappers:

- `getViewPosition` ⇒ `screenCenterToLatLon` — I am confident this was the right choice, since it was the only use of that wrapper left.
- `getPositionThatCentersPosition` ⇒ `getLatLonThatCentersLatLon` — I think this was the right choice. The wrapper is only used in one other place (in `MainFragment`), so it opens the possibility of refactoring to remove it later.
- `updateCameraPosition` ⇒ `updateCameraPosition` — I am not sure this is a good choice. It's used in many places, so it's not going away any time soon. If it will always stay as a simple call down to the equivalent controller method, it doesn't matter either way, but if we might want to add additional behavior there, it would be better to call through the wrapper.
    - Counterpoint: I notice many of the `mapFragment` public methods are like this — they just call directly down to the equivalent controller method. I wonder if it makes sense to just expose a getter for the controller directly.

smichel17:
Also generally a few `let`s instead of `val x = x ?: return`. It's easy to overuse the scope methods, but I think this use is good.

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
westnordost: Sure, looks good! Though, while we are at cosmetics, I generally prefer the pattern
```kotlin
val mapFragment = mapFragment ?: return
```
over
```kotlin
mapFragment?.let { mapFragment ->... }
```
because one level of indentation is saved this way. (Also, imagine you have multiple such properties, then the second one becomes a christmas tree)

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
westnordost: > Counterpoint: I notice many of the mapFragment public methods are like this — they just call directly down to the equivalent controller method. I wonder if it makes sense to just expose a getter for the controller directly.

I don't think it makes sense to expose the controller to the outside. (Generally it seldomly does in my opinion)

westnordost:
Looks good!

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
smichel17: > I generally prefer the pattern
> 
> ```kotlin
> val mapFragment = mapFragment ?: return
> ```

I generally prefer functional styles — the open `return` here makes me uncomfortable. So if I were going to write it with a `let`, I'd write,

```kotlin
let mapFragment = this.mapFragment;
if (mapFragment != null) {
   /* ... */
}
```

…at which point `let` is just syntactic sugar to save a line. I still do use this form when there's an `else` branch, because this looks ugly to me

```kotlin
this.mapFragment?.let { mapFragment ->
    /* ... */
} ?: {
    /* ... */
}
```
If I could use `else` instead of `?:` there, I probably would use that form, though ;)

Unnecessary shadowing also makes me uncomfortable, so I'd either use a different variable name or disambiguate with `this`.

---

But, I do acknowledge this is mostly bikeshedding, so I'm happy to change the style so it's consistent with the rest of the code.

westnordost: > Unnecessary shadowing also makes me uncomfortable, so I'd either use a different variable name or disambiguate with this.

I'd be fine with this too

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
westnordost: I don't mind, it's very short anyway

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
smichel17: okay, I went with `this.`, since it's more in line with the other locations where we do this.

westnordost:
OK, thank you! Will merge this now

PR #2167, opened 2020-10-15
Add some upper case tests

peternewman:
As used in the UK: https://en.wikipedia.org/wiki/House_numbering#United_Kingdom


westnordost:
ok sure

PR #2175, opened 2020-10-17
Add missing import to tactile_resurvey fix

### LINKED ISSUE #2170: Crossing tactile paving question not adding surveyed tag
I'm been walking around my area confirming crossing tactile paving and just noticed all the quests reappeared after a week. Looking at some examples, it looks like StreetComplete has not adjusted the crossing at all if I confirm the existing information:

I just walked past both of these after lunch, and answered them, but there has been no update on the nodes for two years:

https://www.openstreetmap.org/node/5360243110
https://www.openstreetmap.org/node/5360243269

**How to Reproduce**
Answer a tactile paving question that was no before with no in a re-survey.

**Versions affected**
v24.1 and Android 10 (OxygenOS 10.3.5)

dbdean:
Recent change (#2172) to fix #2170 was missing an import needed to compile properly.

I can also confirm that the quest does now work correctly with a resurvey: https://www.openstreetmap.org/node/5360243108.


westnordost:
oops @peternewman 

Thanks for the correction

peternewman:
Apologies all; I really must try and get Travis or something working!

Issue #1305, opened 2018-12-30
Kerb / Curb quest surface type and tactile plates for sidewalk data


omgitsgela:
I am suggesting here two quests related to inclusive pedestrian mapping as defined in https://wiki.openstreetmap.org/wiki/Sidewalks. This has the potential to affect a large amount of kerb ramp data in a positive way.

### General
1. Define what type of kerb interface a crosswalk has to a sidewalk
Affected tag(s) to be modified/added: [kerb=*] https://wiki.openstreetmap.org/wiki/Key:kerb
Search for type=point ; barrier=kerb, which should define a single node on a highway=* segment
Ask "How high is this kerb ramp?"
Answers: flat ; lowered ; raised

2. Ask if a kerb ramp has a tactile plate that would be used by the blind
Affected tag(s) to be modified/added: [tactile_paving=*](https://wiki.openstreetmap.org/wiki/Key:tactile_paving)
Question asked: **Does this kerb ramp have a tactile plate for the blind?**
Answers: yes ;  no

### Checklist
Checklist for quest suggestions (see guidelines):
- [X] 🚧 To be added tag is established and has a useful purpose
   Purpose is to enhance the inclusive pedestrian network tagging through revision of kerb ramp data to enhance overall accessibility for patrons requiring accessible infrastructure
- [X] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
   Tags are well defined for both quests proposed.
- [X] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
   With the proper graphical elements in the displayed quest, this would be answerable by anyone.
- [X] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
   These are roughly boolean options, and so many will have the same answer, but I wouldn't consider this spam, rather, completion of features that should be documented for proper routing.
- [X] 🕓 Applies to a reasonable number of elements (Worth the effort)
   Has the potential to improve a vast amount of kerb ramp data.

### Ideas for implementation
We should use graphical depictions of the kerb ramp heights to show people what the different types are. There's very good graphics available on the Key:kerb page for each option.

We should also use a graphic to show a tactile plate, so people know what it is when asked.

pkoby:
[...]

PR #2183, opened 2020-10-20
Kerb - tactile status and height

matkoniecz:
Solves #1305 (or rather, will solve once fully implemented)

Posting with just one (simpler) quest done from both, as there is still icon needed.

I tried making one, but it is horrible (and that is why usually I am not even trying... :) ).

So if @westnordost can either make an icon or has an idea how to make some idea how to make a better one...

https://user-images.githubusercontent.com/899988/96545182-305ae100-12a8-11eb-8ba4-7c1f1681d093.jpg

Note that there will be also quest about kerb type, and I am completely at loss how to make an icon for that. I tried something along this lines and it is not really looking like going somewhere

https://user-images.githubusercontent.com/899988/96545624-09e97580-12a9-11eb-9c76-1b3d0a76587d.jpg


app/src/main/res/values/strings.xml
```diff
@@ -929,4 +929,5 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="dialog_tutorial_download">With auto-sync off, quests are not downloaded automatically. To do this manually, use the \"%s\" button in the menu at the location you want to download.</string>
 
     <string name="quest_pedestrian_crossing_island">"Does this pedestrian crossing have an island?"</string>
+    <string name="quest_tactile_paving_kerb">Is there a tactile paving on this kerb?</string>
```
peternewman, line 932: Minor tweak
```suggestion
    <string name="quest_tactile_paving_kerb">Is there tactile paving on this kerb?</string>
```

Is this supposed to be all the way along a kerb or just at a crossing, in which case isn't it covered by the existing quest?

matkoniecz: Just at the specific point of a specific crossing (in extreme cases one side of crossing may have tactile paving, second not).

matkoniecz: What about 

> Is there tactile paving on kerb here?

?

matkoniecz:
What would be a good question?

Is there tactile paving on this kerb?
Is there tactile paving on kerb here?
Is there tactile paving on kerb on this crossing?

In #1305 @omgitsgela proposed

> Does this kerb ramp have a tactile plate for the blind?

but not all crossing have kerb ramps, and I think that limiting to just kerb ramps is not needed nor wanted

westnordost:
Can you give me push access to this branch if you haven't yet? I am going to merge master and fix the query. I don't want to expect this of you.

matkoniecz:
"Allow edits and access to secrets by maintainers" is enabled

westnordost:
Okay, that was... quite the effort. I also added tests, the icons and fixed a copy&paste mistake in `AddKerbHeight`. I hope the rest there is to do here will see the light of release soon.

westnordost:
Theoretically it is now possible to also look for nodes as described in point 3 in https://github.com/westnordost/StreetComplete/issues/1305#issuecomment-688333976 : 
find all nodes that are
- at the intersection **endpoints** of a footway with footway=sidewalk and a footway with footway=crossing
- and that are part of only exactly 2 footways, i.e. there is no "T-intersection"

But I didn't bother with it in the merge. If you think this (=not adding barrier=kerb) is a common tagging and the situation described in point 3 should be recognized too, feel free to add this logic to KerbUtil.

westnordost:
Ok whatever, I also implemented this. It is quite complex and I am not sure if this was the first thing you should do when learning to write programmatic filters in kotlin code

westnordost:
But you should review this and report if you don't understand anything or something seems wrong.

matkoniecz:
Thanks for all that changes, for icons, case 3 implementation, updating to new querying and so on.

Also, new osmapi finally actually allows adding the first tag with SC :) ( https://www.openstreetmap.org/node/4103404566/history )

app/src/main/res/values/strings.xml
```diff
@@ -927,4 +927,9 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_pedestrian_crossing_island">"Does this pedestrian crossing have an island?"</string>
     <string name="quest_surface_detailed_answer_impossible_title">Describe surface</string>
     <string name="open_url">Open URL</string>
+    <string name="quest_tactile_paving_kerb_title">Is there tactile paving on the kerb on this crossing?</string>
+    <string name="quest_kerb_height_title">What is the kerb height on this crossing?</string>
```
westnordost, line 931: Actually, shouldn't it say "What is the type/height of this **kerb**", rather than crossing? After all, we ask for each kerb individually.

matkoniecz:
screen01

I have not really found ones differing only by kerb height, but together with text it should be good enough.

I plan to align kerb height to something more similar, rebalance tone/light levels and add ramp photo - sadly, in probably a noticeably different style.

For reference tagged values would be `raised`, `lowered`, `flush`.

westnordost:
Good pictures!
you mean to have `lowered` a little higher?

matkoniecz:
>  you mean to have lowered a little higher?

Yes, the same for flush - to have bottom of kerb on the same height of photos. Currently it looks a bit untidy.

matkoniecz:
And for the question, I will change it as pointed out in the earlier comment.

> Is there tactile paving on this kerb for this crossing?

Is probably suboptimal Engrish, feedback welcomed.

> Is there tactile paving on this kerb?

feels not ideal, as  question is about tactile elements just on crossing, not for the entire kerb (but it is kind of implied by quest being for point, not for line...)

matkoniecz:
Just noticed: there is a possibility that there is no kerb whatsoever at given location (possible as result of https://github.com/westnordost/StreetComplete/pull/2183/commits/b51bee6361143cad0c819551df779862e837f4c7 )

What should be tagged in such case? `kerb=no`? https://taginfo.openstreetmap.org/tags/kerb=no - seems decently popular

westnordost:
I just made some photos. All are CC-0. Feel free to use them if you think any of these are a better fit.

westnordost:
> What should be tagged in such case? kerb=no? https://taginfo.openstreetmap.org/tags/kerb=no - seems decently popular

Yeah maybe. After all, there is no `barrier=kerb=no`.

Ah, by the way, StreetComplete should maybe complement `barrier=kerb` if it is missing (and the answer is not "no kerb") ?

matkoniecz:
> Ah, by the way, StreetComplete should maybe complement barrier=kerb if it is missing (and the answer is not "no kerb") ?

Is there some obvious way to distinguish "this untagged node in on crossing of `barrier=kerb` and `highway=footway`" (adding `barrier=kerb` on node is unnecessary) and "this untagged node in on transition from `footway=sidewalk` to `footway=crossing`" (adding `barrier=kerb` would be useful).

matkoniecz:
screen02

Now I am going to try balance colors of multiple photos for the first time in my life, lets see how it will go.

matkoniecz:
screen03

I am really unhappy about ramp not matching at all, but I was unable to find any such object in my city or a better image of it. And it is required to make solving less confusing in areas where curb cut is present.

note to self: remember to edit UK translation once it is merged and available for translating

app/src/main/res/values/strings.xml
```diff
@@ -927,4 +927,11 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_pedestrian_crossing_island">"Does this pedestrian crossing have an island?"</string>
     <string name="quest_surface_detailed_answer_impossible_title">Describe surface</string>
     <string name="open_url">Open URL</string>
+    <string name="quest_tactile_paving_kerb_title">Is there tactile paving on this kerb for this crossing?</string>
+    <string name="quest_kerb_height_title">What is the height of this kerb?</string>
+    <string name="quest_kerb_height_flush">Same level as road surface</string>
+    <string name="quest_kerb_height_lowered">Lowered, still above road surface</string>
+    <string name="quest_kerb_height_raised">High kerb</string>
+    <string name="quest_kerb_height_answer_no_kerb">There is no kerb on this side of crossing</string>
+    <string name="quest_kerb_height_lowered_ramp">Curb cut</string>
```
FloEdelmann, line 936: Shouldn't this be British English like the ones above?

```suggestion
    <string name="quest_kerb_height_lowered_ramp">Kerb cut</string>
```

matkoniecz: Good point, though SC is in US English with UK English handled as translation for such cases. I will fix this.

app/src/main/res/values/strings.xml
```diff
@@ -927,4 +927,11 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_pedestrian_crossing_island">"Does this pedestrian crossing have an island?"</string>
     <string name="quest_surface_detailed_answer_impossible_title">Describe surface</string>
     <string name="open_url">Open URL</string>
+    <string name="quest_tactile_paving_kerb_title">Is there tactile paving on this curb for this crossing?</string>
```
westnordost, line 930: "on this curb for this crossing" - why not simply "on this curb"?

westnordost (commented):
What about `kerb=no`?

matkoniecz:
> What about kerb=no?

It is available via "can't say" menu.

> Also, I don't have an idea what a "curb cut" is supposed to be, also not from the photo

https://en.wikipedia.org/wiki/Curb_cut

A small ramp allowing to get stroller/wheelchair/bicycle more easily through curb. Not sure how to improve it, I may look again for a photo. Maybe also modify description. Is "curb cut (small ramp)" any better?

app/src/main/res/values/strings.xml
matkoniecz: Because it is asking about tactile paving status just for crossing, not entire curb. But maybe it is communicated by quest element being a point.

matkoniecz:
I tweaked description for `kerb=lowered`, hopefully it actually is better or not worse.

 screen04

app/src/main/res/values/strings.xml
westnordost: "on this curb here"?

westnordost:
> A small ramp allowing to get stroller/wheelchair/bicycle more easily through curb. Not sure how to improve it, I may look again for a photo. Maybe also modify description. Is "curb cut (small ramp)" any better?

Where is the difference then to "lowered" and/or "flat"?

matkoniecz:
> Where is the difference then to "lowered" and/or "flat"?

In tagging: none - both "a bit higher than road surface" and "curb cut" is tagged as `kerb=lowered`

In real use as an obstacle: there is some difference, depending on user (for strollers difference would be likely minimal, but for a older person ramp may become a barrier, for wheelchair users that are weaker it also may become a barrier, for cyclists it would be likely a bit more annoying). But difference between "ramp" and "lowered kerb" are not so great (I think).

In real use for blind: `kerb=flush` is considered as problematic as it is basically undetectable, unlike lowered kerb that still has some height difference that is detectable by blind. Not sure about how curb cut works here and whatever it also hides kerb transition from people not using sight to navigate.

In mapping: With just three photos and their description I would not guess which one is supposed to be used for curb cuts, that is why I added fourth option. This feature appears popular in (for example) USA, at least according to my checks.

westnordost:
If a curb cut is a curb ramp, maybe that would be the better wording for it because it uses simpler more self-explanatory words. From the picture, as said, I don't understand what a kerb cut is, maybe also because there is no sense of dimension. How broad is that ramp? My estimate from loking at the picture would be about 30cm, but this would still not be enough at all for wheelchair users.

Also, by the way, if there are exactly 4 choices, it would be nice but is not required to use 2 columns (with broader pictures) like in the steps ramp quest.

matkoniecz:
> How broad is that ramp?

No idea - https://en.wikipedia.org/wiki/Curb_cut describes it as "pram ramp with tactile paving that connects an off-street walking trail to a road"

> If a curb cut is a curb ramp, maybe that would be the better wording for it because it uses simpler more self-explanatory words.

OK, I will change it that way

> My estimate from loking at the picture would be about 30cm, but this would still not be enough at all for wheelchair users.

Good point. Maybe it would be better to tag lowered kerb and kerb ramps in way that allows distinguishing them? Still, I would prefer to make a quest using current (maybe suboptimal tagging). And immediately after improve tagging scheme as a separate project, rather doing both as one project together.

matkoniecz:
@westnordost Do you think that it is ready or that some of what is discussed is a blocker? (I am considering it as finished, please let me know if you think otherwise)

(if you simply had no time/state of mind to review - I will wait patiently, no problem, I just wanted to avoid case of both sides waiting for another)

westnordost:
I've reviewed it, it is fine. As optional, there is still

> Also, by the way, if there are exactly 4 choices, it would be nice but is not required to use 2 columns (**with broader pictures**) like in the steps ramp quest.

however, not a blocker. I did not merge it yet because it contains new strings and master is for v26 so far.

matkoniecz:
Aaaa, right. Switching to two colums will not uncrop images :)

matkoniecz:
I need to fix false positive at https://www.openstreetmap.org/node/1079816519#map=19/50.07144/19.92905

westnordost:
Actually, here is a good example where test driven development would make sense. First write the test that documents what behavior you would like to see (which special case you want to catch), then write the code. The giant comment you put into the code would be better as a test. Good tests are documentation.

To me the linked special case looks like you'd want to count the connections with `highway=footway + footway=crossing` not only with sidewalks but with any way that could be a sidewalk - any footway, path or cycleway that is not a crossing.

westnordost:
I think your implementation does not work for 

image

(You should write that as a test.)

What you want to do is that any crossing end node that has a connection with any other way (of path, cycleway, footway) that is not a sidewalk is disqualified.

matkoniecz:
> I think your implementation does not work for

I am pretty sure that it works - `= -1` is applied, not `-= 1`

> of path, cycleway, footway

Hmm, and probably it includes `highway=cycleway` without `foot` tag that right one are skipped in footwaysFilter.

The same for `highway=construction` (possibly with `construction=footway`)

> Actually, here is a good example where test driven development would make sense. First write the test that documents what behavior you would like to see (which special case you want to catch), then write the code. The giant comment you put into the code would be better as a test. Good tests are documentation.

+1

westnordost:
I am sorry, I accidentally edited your comment instead of answering! Restired your comment.

>> I think your implementation does not work for

> I am pretty sure that it works: `= -1` is applied, not `-= 1`

Ah, right, well.

matkoniecz:
It iseems that adding "node is not on any way that is not terminating on it" and "node has no attached ways that are neither `footway=crossing` nor `footway=sidewalk`" checks should work, hopefully it will not be horribly slow.

matkoniecz:
Huh, I expected that this checks would be horrifyingly unusuably slow and it appears to be fine.

matkoniecz:
Uncropping is not working exactly like I expected, but it is an improvement anyway.

screen14

screen13

westnordost:
uncropping:
Just have a look at the graphics for the steps ramp quest, then you know how big they must be.

app/src/main/java/de/westnordost/streetcomplete/quests/kerb_height/KerbUtil.kt
```diff
@@ -0,0 +1,101 @@
...
+        .flatMap { it.nodeIds.firstAndLast() }
+        .mapNotNull { getNode(it) }
+
+    val unknownEndNodes = footways
+        .filter { it.tags?.get("footway") != "sidewalk" && it.tags?.get("footway") != "crossing" }
+        .flatMap { it.nodeIds.firstAndLast() }
+        .mapNotNull { getNode(it) }
+
+    val allNodesOnNotFootways = ways
+        .filter { footwaysFilter.matches(it) == false }
```
westnordost, line 35:
```suggestion
        .filterNot { footwaysFilter.matches(it) }
```

westnordost:
I refactored the KerbUtil code to be expressed less laborious and hopefully be better readable. Should also be faster.

matkoniecz:
screen15

westnordost:
One more thing: Did you add the quest types to count towards certain achievements?

westnordost (approved):

matkoniecz:
> Note to self: make sure that GB translations are done (curb vs kerb)

I remembered it, done :)

davidpnewton:
There's a problem with the tagging used with this quest. Kerb=lowered is absolutely fine, and although I haven't tagged it yet I'd imagine that kerb=flush is used for flush kerbs.

However the quest shows a picture of a kerb with the following text, "High curb" and tags the result as kerb=raised. The problem is that the picture is NOT of a raised kerb, it is of a normal height kerb. A raised kerb where kerb=raised is the appropriate tag is at places like bus stops where the kerb is even higher than it along normal sections of the road. See this Google Streetview image for an example of what I mean:

https://www.google.com/maps/@52.2622662,-1.5256606,3a,66.7y,170.41h,74.77t/data=!3m6!1e1!3m4!1sSRsZVJzy2ff3r8GYX8jrNw!2e0!7i13312!8i6656

Behind the bus is a dropped kerb for two driveways. In front of the bus is a normal height kerb. Alongside the bus is a raised kerb. This is tagged kerb=lowered for the driveway, kerb=raised for the bus stop itself, kerb=flush for the kerb in between the bus stop and the main carriageway and kerb=yes for the rest (because there is no specific tag for a kerb of normal height that is present). Those tags are applied to a way which is traced along the kerb at the edge of the road, but of course the same situation applies equally to the nodes which this quest deals with.

These are OSM ways 873137651, 873137652, 873137653 and 856824474.

So I'd say this quest needs tweaking to have another option available to solve it. Instead of flush, lowered, raised, ramp and other as answers have flush, lowered, raised, ramp, normal and other for the answers. Normal would simply be a kerb which is present but is not flush, raised, lowered or a ramp and would simply be tagged kerb=yes (unless someone can come up with a better tag for that). For the high kerb I'd also suggest a different image is needed showing a raised kerb at a bus stop or similar.

westnordost:
Change the wording from "high curb" to "normal curb" then? @matkoniecz

davidpnewton:
That's not the whole issue. The phraseology and semantic meaning is a bit awkward to deal with in isolation, but in context of the quest it makes perfect sense. For the current picture with "high curb" change it to "normal curb" and change the tagging from kerb=raised to kerb=yes (or kerb=present or whatever is appropriate), and also add in a further, distinct answer for raised kerb, tagged as kerb=raised with a picture showing a bus stop or similar with a kerb that's distinctly higher than normal.

For a more dramatic image of what I mean by something that would be tagged kerb=raised, rather than kerb=yes see the image captioned Kassel Access Kerb used at a bus stop at Cardiff Bay in the following blog post:

https://therantyhighwayman.blogspot.com/2013/08/kerb-your-enthusiasm.html

So in that picture the brick portion of the kerb edge would be tagged kerb=yes and the long, concrete slabs would be tagged kerb=raised. Hopefully that illustrates what I mean about the distinction better than the Google Streetview image I linked to in the previous comment.

matkoniecz:
Disclaimer: I am not a native speaker

>  change it to "normal curb"

The problem with using "normal" is that what is normal depends on a location. For example in some place lowered curb may be normal. In such case I felt that it would be quite confusing.

"raised curb" may be fine, though for me it felt a bit confusing as anything not flush is raised - more or less

davidpnewton:
Like I said, it's awkward to deal with in isolation, and it's awkward to describe independently. That Google Streetview example I provided shows the four main kerb situations, and shows that they need to be distinct. kerb=raised is a tag which is used for something else than it is being applied to by the app. Therefore that needs to be fixed. The questions are what tag do we use for kerbs which are not flush, dropped or raised when compared to the usual for the area, and how do we word things so that people understand what is meant by "usual for the area"?

On the tag front kerb=yes does seem to be the tag used most often. If that's the case then we just need to find a form of words that unambiguously, and succinctly conveys the concept.

goldfndr:
Can the difference between this higher "raised" kerb and the "yes" (medium?) kerb be quantified? It doesn't look much higher.

I can contribute pictures of rolled kerbs, they are present in at least two residential areas locally. (Most of the driveways where present don't have aprons, they use the sidewalk as is, but some homeowners buy long supplemental risers that reside in the gutter and have a gap for storm water flow and effect a ramp.) But not with the same sidewalk texture, here there are large (~1×1½ m²) slabs of concrete for the sidewalk.

PR #2195, opened 2020-10-24
Fix some changelog formatting

peternewman:


matkoniecz:
@westnordost Is it safe to edit in https://github.com/westnordost/StreetComplete/releases/edit/v25.0 ?

westnordost:
Yes it is, Done that

PR #2207, opened 2020-10-27
Use try-with-resources.

Isira-Seneviratne:
[Try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) is supported on API levels below 19 when Java 8 support is enabled.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.java
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.java
@@ -114,24 +114,14 @@ public class CountryInfos
 	private CountryInfo loadCountryInfo(String countryCodeIso3166) throws IOException
 	{
 		String filename = countryCodeIso3166+".yml";
-		InputStream is = null;
-		try
+		try (InputStream is = assetManager.open(BASEPATH + File.separator + filename))
 		{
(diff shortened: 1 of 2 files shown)


westnordost:
Looks reasonable. Did you test it? Both places are critical places within the app, an error there will have very serious consequences (data corruption).

Isira-Seneviratne:
> Looks reasonable. Did you test it? Both places are critical places within the app, an error there will have very serious consequences (data corruption).

I tried updating a note and didn't encounter any crashes or anything like that.

westnordost:
Mh, the serializer is only used for notes if you attached a picture to the note

Isira-Seneviratne:
> Mh, the serializer is only used for notes if you attached a picture to the note

Oh, I see. In that case, what about when adding road information?

westnordost:
Yes, solving a quest and uploading the change should be enough as it serializes the changes once and then deserializes it for upload. If the upload is correct, then the change did not create any problems

westnordost:
Is that what you meant with "adding road information"?

Isira-Seneviratne:
> Is that what you meant with "adding road information"?

Yeah.

westnordost:
Ok, then this is safe. Thank you!

PR #2208, opened 2020-10-27
WIP: Add initial Room support.

Isira-Seneviratne:
Add a `RoomDatabase` implementation, migrations and entities, and switch to `SupportSQLiteOpenHelper`. This PR is limited to those changes to keep it as simple as possible, with the intent being to fully switch to Room over multiple PRs: https://medium.com/androiddevelopers/incrementally-migrate-from-sqlite-to-room-66c2f655b377.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,6 @@
-import java.util.Properties
-import java.net.URI
 import java.io.FileInputStream
+import java.net.URI
+import java.util.*
 
 plugins {
     id("com.android.application")
@@ -38,6 +38,16 @@ android {
         versionCode = 2501
         versionName = "25.1"
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+
+        javaCompileOptions {
+            annotationProcessorOptions {
+                arguments += arrayOf(
+                    "room.schemaLocation" to "$projectDir/schemas",
+                    "room.incremental" to "true",
+                    "room.expandProjection" to "true"
+                )
+            }
+        }
     }
 
     buildTypes {
@@ -99,6 +109,7 @@ dependencies {
     val mockitoVersion = "2.28.2"
     val kotlinxVersion = "1.3.8"
     val daggerVersion = "2.14.1"
+    val roomVersion = "2.2.5"
 
     // debugging
     debugImplementation("com.squareup.leakcanary:leakcanary-android:2.4")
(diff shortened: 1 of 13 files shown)


matkoniecz:
Why migrate from SQLite to Room? (sorry if that is obvious, but I lack knowledge here)

Isira-Seneviratne:
> Why migrate from SQLite to Room? (sorry if that is obvious, but I lack knowledge here)

[Room](https://developer.android.com/training/data-storage/room) provides an abstraction layer over SQLite, so performing CRUD operations can be done without having to write boilerplate code.

matkoniecz:
I am NOT experienced in this area, bur from my perspective entities and other Room-specific code is a boilerplate. What worse it is sn Android specific boilerplate rather than universal and clear SQL, adding even more unclear layers of abstraction.

(though my experience here is minimal, I am curious about what @westnordost thinks about this)

Isira-Seneviratne:
> I am NOT experienced in this area, bur from my perspective entities and other Room-specific code is a boilerplate. What worse it is sn Android specific boilerplate rather than universal and clear SQL, adding even more unclear layers of abstraction.
> 
> (though my experience here is minimal, I am curious about what @westnordost thinks about this)

I meant that manually writing code using `Cursor` and `SQLiteOpenHelper` isn't required.

westnordost:
Thank you for your effort, @Isira-Seneviratne , but I fear you could have saved yourself this effort.

I am with @matkoniecz here. The database layer is already well readable and quite short in terms of code thanks to some convenience extension functions. Room may make access for CRUD operations even shorter, but introduces a new framework. The more frameworks there are, the more frameworks (new) contributors need to learn to use and understand, thus increases the entry barrier to contribute to this project. And of course, adding new dependencies has always downsides as well as one becomes dependent on that that dependency may have bugs and may not be properly maintained (one day). (Though since it is from Google, not that much concern here.)

Additionally, to keep options open for a possible port to iOS, if any abstraction layer is introduced, it should actually be one without external dependencies on a Java-library, let alone a Java-library dependent on the Android environment.

Looking at the code of this PR itself, there are a few more reasons not to use this, but the other reasons states should be enough.

Isira-Seneviratne:
> Thank you for your effort, @Isira-Seneviratne , but I fear you could have saved yourself this effort.
> 
> I am with @matkoniecz here. The database layer is already well readable and quite short in terms of code thanks to some convenience extension functions. Room may make access for CRUD operations even shorter, but introduces a new framework. The more frameworks there are, the more frameworks (new) contributors need to learn to use and understand, thus increases the entry barrier to contribute to this project. And of course, adding new dependencies has always downsides as well as one becomes dependent on that dependency - it may have bugs and may not be properly maintained (one day). (Though since it is from Google, not that much concern here.)
> 
> Additionally, to keep options open for a possible port to iOS, if any abstraction layer is introduced, it should actually be one without external dependencies on a Java-library, let alone a Java-library dependent on the Android environment.
> 
> Looking at the code of this PR itself, there are a few more reasons not to use this, but the other reasons stated should be enough.

Okay, I see.

Issue #2177, opened 2020-10-18
Add categories to the "What kind of building is this?" Quest


RiffLord:
**Use case**

Some areas have many abandoned buildings and/or ruins which can't be adequately classified by the categories available in the _What kind of building is this?_ quest. I believe it would be useful to be able to mark these buildings as abandoned.

On a related note, I believe the quest would also benefit from having a category for historical buildings which could encompass used, unused, public or private buildings of historical relevance which otherwise wouldn't fit into any of the currently available categories. While many historical buildings are used for purposes which can be classified with the options available (Civic, Religious or Commercial buildings) there are some which don't fall into these categories and might otherwise remain unmarked.

**Proposed Solution**

Add an _Abandoned building/Ruin_ and a _Historical building_ option to the _Other_ category in the options for the quest.

Something like this (the icons were repurposed from those available just as a placeholder):

![Untitled](https://user-images.githubusercontent.com/47884610/96376329-8df30e80-117e-11eb-9a92-cf684be117c0.jpg)

matkoniecz:
Note that abandoned bunker should still be answered as "bunker", abandoned shed as "shed" and so on.

Historic buildings answer - which tag would be used in this case?

RiffLord:
Regarding generic abandoned buildings, what about other categories which might cause confusion? 
For example, in the area where I live there is the ruin of a restaurant. If I were to classify it as a Shop per the app's guidelines, wouldn't it show up as a result along with other commercial activities for people using OSM to find services in the area?

As for the historic buildings, I assume that simply using a _Historic_ tag would be too vague perhaps?
In that case what about creating a separate Historic category, including tags such as _Castle_, _Fort_, _Watchtower_, _Observation Post_, etc?

The reasoning behind this is that while all over the world buildings such as these are often museums or tourist attractions and can be tagged appropriately through the app, what about the similar places which don't have this function?
How do we tag those historic buildings which remain officially unused but accessible to people?

westnordost:
It would be possible to tag something like `abandoned=yes` or `historic=yes`. These tags are even well in use.

RiffLord:
So these characteristics should be specified by editing tags after the fact on OpenStreetMaps rather than adding options to the quest in the StreetComplete app?

westnordost:
No, on the contrary! What I meant to say is that it would be quite possible to add these two options in streetcomplete because there is an accepted tagging for that

[...]

PR #2214, opened 2020-10-30
Add building types to "What kind of building is this?" quest

RiffLord:
Solves issue #2177


app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
```diff
@@ -37,7 +37,10 @@ class AddBuildingType (o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestTy
             changes.delete("building")
             changes.add("man_made", manMade)
         } else {
-            changes.modify("building", answer)
+            if (answer.contains("historic") || answer.contains("abandoned") || answer.contains("ruin"))
+                changes.add(answer, "yes")
+            else
+                changes.modify("building", answer)
         }
```
westnordost, line 44: You could pull up that else. I.e. if ... else if... else

Also, why `answer.contains("historic")` and not `answer == "historic"`?

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypes.kt
```diff
@@ -58,6 +58,10 @@ enum class BuildingType(val item: Item<String>) {
     SERVICE       (Item("service",    R.drawable.ic_building_service,    R.string.quest_buildingType_service, R.string.quest_buildingType_service_description)),
     HANGAR        (Item("hangar",     R.drawable.ic_building_hangar,     R.string.quest_buildingType_hangar, R.string.quest_buildingType_hangar_description)),
     BUNKER        (Item("bunker",     R.drawable.ic_building_bunker,     R.string.quest_buildingType_bunker)),
+    //  TODO: replace placeholder icons
```
westnordost, line 61: I'll do that after the merge

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
```diff
@@ -37,7 +37,10 @@ class AddBuildingType (o: OverpassMapDataAndGeometryApi) : SimpleOverpassQuestTy
             changes.delete("building")
             changes.add("man_made", manMade)
         } else {
-            changes.modify("building", answer)
+            if (answer.contains("historic") || answer.contains("abandoned") || answer.contains("ruin"))
```
westnordost, line 40: I think it is _ruin**s**_, not _ruin_.

westnordost (changes requested):
See my comments.

I didn't follow the discussion in that ticket yet whether instead `abandoned:building` should be used or not. Will do that tomorrow.
Maybe it would make sense to extend the filter at the top of the file to also exclude buildings that have been tagged with such prefixes.

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypes.kt
RiffLord: Should I remove this comment before updating the PR?

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
```diff
@@ -36,8 +37,9 @@ class AddBuildingType : OsmFilterQuestType<String>() {
             val manMade = answer.split("=")[1]
             changes.delete("building")
             changes.add("man_made", manMade)
-        } else {
+        } else if (answer == "historic" || answer == "abandoned" || answer == "ruins")
+            changes.add(answer, "yes")
```
westnordost, line 41: Hmm, this will crash the app whenever there was a previous value for either of `history`, `abandoned` or `ruins` that was not `yes`.
```suggestion
            changes.addOrModify(answer, "yes")
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
```diff
@@ -20,6 +20,7 @@ class AddBuildingType : OsmFilterQuestType<String>() {
          and !amenity
          and !leisure
          and location != underground
+         and abandoned != building
```
westnordost, line 23:
```suggestion
         and abandoned != building
         and abandoned:building != yes
```

westnordost:
I added the icons to master. You just need to merge master int your branch and use the new icons

westnordost:
Looking good, thank you!

Issue #2230, opened 2020-11-02
Inconsistent cycleway quest for residential highways


HolgerJeromin:
https://www.openstreetmap.org/way/385399447
```
highway=residential
lit=no
maxspeed=30
name=[set]
source:maxspeed=DE:zone30
surface=asphalt
```
shows the cycleway quest. Probably fine.

https://www.openstreetmap.org/way/387842231
```
highway=residential
lit=no
maxspeed=30
name=[set]
surface=asphalt
zone:maxspeed=DE:30
```
showed the quest on the first load (small area) and was hidden after a few seconds.

Perhaps there is a good reason for that, but this is surprising at least.

**Versions affected**
Android v10
SC 26.0 beta 3

matkoniecz:
> Perhaps there is a good reason for that, but this is surprising at least.

Is it maybe blocked by other nearby, higher priority quest? Try moving cycleway quest to top of list in settings.

westnordost:
When download happened for https://www.openstreetmap.org/way/387842231 , were you standing west or east of that way?

westnordost:
Anyway, I am pretty you were standing west of it.

The mentioned way is on several tiles on zoom level 16. A tiny part of it is on tile https://tile.openstreetmap.org/16/34314/21685.png and the rest is on https://tile.openstreetmap.org/16/34315/21685.png

What happened is the following:
The app first downloaded the data for only the left tile. A quest was created for way 387842231 because no footway was within 15m of it it.
Then, the app downloaded the data for a bigger area, an area in which both mentioned tiles were included. Included in the returned data was this footway https://www.openstreetmap.org/way/660700307 . Since the way 387842231 was now within 15m of a footway, the current quest for that way was deleted because it turned out to be an element for which the cycleway quest should not be created after all.

One solution would be to not create quests for ways that intersect with the bounding box. This way, the created quests would be very stable. But on the other hand, this would mean that there will be blind spots the edges of zoom16 tiles. Additionally, most streets will be so long that they intersect with the edge(s) of a zoom16 tile one way or another. So, this solution would mean that there will be almost no quests at all.

Because of this, I decided to not fix this: After all, what the quest really wants to exclude consistently are sidewalks that run alongside the street. There are very few situations where a sidewalk running along a street would not be present in the same tile(s) as the street it belongs to. Most cases of footways that are near streets but not (always) present in all the tiles the street is will be some footway leading away from the street itself, like in this case.

westnordost:
[...]

PR #2231, opened 2020-11-02
Only exclude a way if the there is a footway of which at least one se…

westnordost:
…gment fulfills the following conditions for any segment of this way

- is within 15m
 - angle difference of the two segments are max 45°

#2230


westnordost:
TODO: also add tests

westnordost:
and also do not use a constant 15m but vary it according to `width` and `lanes` tags.

PR #2233, opened 2020-11-02
encourage tagging building type rather than historic=yes

matkoniecz:
for example closed historic church ca be tagged as building=church what seems more useful than historic=yes

(spotted while bringing Polish translation to 100% (again :) ))


matkoniecz:
Not entirely sure is `an unknown or unclear` or `an unknown or an unclear` is better.

michaelblyons:
The ELL question isn't quite as relevant here because "unknown" and "unclear" are adjectives, not nouns. Personally, I'd leave the text like 49b3cd0, but 8b0cbcb will still be understood by English speakers.

- 49b3cd0: "an {unknown or unclear} purpose"
- 8b0cbcb: "an unknown [implied expected noun] or an unclear purpose"

westnordost:
Ok, I'll change it like @michaelblyons suggested

PR #2237, opened 2020-11-04
ask the surface of roads with surface=cobblestone because it is generic

westnordost:
@matkoniecz what do you think? I was thinking to make a separate quest out of this, but I think this is easier and has no real downsides.


westnordost:
So, `surface=cobblestone` could be natural (unhewn) cobblestone or sett. In most cases it is sett. I was thinking to present these two options in a separate quest, plus a "show more" button that shows the rest, in case it has been tagged wrongly. But on the other hand, we had this discussion with `DetailRoadSurface` (paved/unpaved) and the solution was to just put everything into the same quest type.

matkoniecz:
The only downside that I see is that in a separate quest people are less likely to be confused and miss `sett` or `unhewn_cobblestone` option.

But probably it should not be a big problem, right?

westnordost:
I think not. I'll merge it then.

PR #2244, opened 2020-11-06
kosher quest

matkoniecz:
First question: what is the method now to make `R.drawable.ic_quest_` icons?

I used to open this big svg file with icons, copy icon to new field (with pin), edit it a bit, and later copy it to separate file and import that in Android Studio.

Now I am unsure whatever I am supposed to import drawable without pin? Or put `StreetComplete/res/graphics/quest icons` file into pin somehow?

--------------------

Anyway, it is kind of akademic as my K is horrific anyway. I am relatively happy compared to my earlier icon attempts but... So if @westnordost wants to make a better K I would be thankful :)

(Why K? I thought about using Star of David as an icon, but I consulted with a relevant OSM community and they were sure that K would be a far better symbol here than Star of David).

-----------------

I asked some people (OK, many people) and sadly this needs to be disabled by default as you need to go inside and ask to be sure that something is not kosher. In many cases you can easily discover that something is kosher from outside, but confirming non-kosher status is harder.

So it is not restricted to just Israel, it allows to enable quest if you spot kosher restaurant/fast_food/butcher outside Israel and tag without leaving StreetComplete. Or tag kosher status if you are outside Israel in area with some kosher restaurants.

Risk here is someone enabling it and adding "not kosher" to 200 restaurants in area where nothing at all is kosher. But hopefully people are not going to do that, in the worst case country restriction can be added later.


app/src/main/res/values/strings.xml
```diff
@@ -474,6 +474,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="undo_confirm_negative">Cancel</string>
     <string name="enable_quest_confirmation_title">Enable this quest type?</string>
     <string name="default_disabled_msg_go_inside">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type.</string>
+    <string name="default_disabled_msg_go_inside_usefullness_warning">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type. Please, do not enable this quest if answer would be "no" for all objects.</string>
```
FloEdelmann, line 477: Typo:

```suggestion
    <string name="default_disabled_msg_go_inside_usefulness_warning">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type. Please, do not enable this quest if answer would be "no" for all objects.</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddKosher.kt
```diff
@@ -0,0 +1,34 @@
...
+        )
+        and name and (
+          !diet:kosher
+          or diet:kosher != only and diet:kosher older today -4 years
+        )
+    """
+    override val commitMessage = "Add kosher status"
+    override val wikiLink = "Key:diet:kosher"
+    override val icon = R.drawable.ic_quest_kosher
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside_usefullness_warning
```
FloEdelmann, line 25:
```suggestion
    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside_usefulness_warning
```

FloEdelmann (changes requested):

matkoniecz:
@FloEdelmann Thanks, fixed!

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddKosher.kt
```diff
@@ -0,0 +1,34 @@
...
+    override val wikiLink = "Key:diet:kosher"
+    override val icon = R.drawable.ic_quest_kosher
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside_usefulness_warning
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_dietType_kosher_name_title
+
+    override fun createForm() = AddDietTypeForm.create(R.string.quest_dietType_explanation_kosher)
+
+    override fun applyAnswerTo(answer: String, changes: StringMapChangesBuilder) {
+        changes.updateWithCheckDate("diet:vegan", answer)
```
ENT8R, line 32: Looks like a copy and paste error

```suggestion
        changes.updateWithCheckDate("diet:kosher", answer)
```

matkoniecz:
I should also probably roll back pointless whitespace changes done to `app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt`

matkoniecz:
> Will the kosher quest be by default disabled?

Yes, as you need to go inside and ask to be sure that something is not kosher. Some places make clear that they are a kosher location by prominent signage visible outsdide, but remaining ones may be kosher or not. Kosher certificate may hanging somewhere inside, sometimes prominently displayed, sometimes less prominently, sometimes available on request.

So you need at least to go inside, and maybe even ask.

westnordost:
> First question: what is the method now to make R.drawable.ic_quest_ icons?
> Now I am unsure whatever I am supposed to import drawable without pin? Or put StreetComplete/res/graphics/quest icons file into pin somehow?

What pin? You just 

image

and then you 

image

westnordost:
> K

Just  K? The German Wikipedia has some pictures of certification logos:
https://de.wikipedia.org/wiki/J%C3%BCdische_Speisegesetze#Zertifizierung

app/src/main/res/values/strings.xml
```diff
@@ -474,6 +474,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="undo_confirm_negative">Cancel</string>
     <string name="enable_quest_confirmation_title">Enable this quest type?</string>
     <string name="default_disabled_msg_go_inside">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type.</string>
+    <string name="default_disabled_msg_go_inside_usefulness_warning">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type. Please, do not enable this quest if answer would be "no" for all objects.</string>
```
westnordost, line 477: I am not sure if it is StreetComplete's task to warn about "self-imposed" spam. I know, there is the guideline 

> 💤 No spam: It must be possible to determine whether the quest should reasonably be asked. A quest which is to be answered in 99% of the time with the same answer is not a good quest, don't bore the users. 

, however, this concerns the quest design of enabled quests. If the user enables the quest by himself, he is responsible himself for boring himself.

app/src/main/res/values/strings.xml
```diff
@@ -935,4 +936,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_pedestrian_crossing_island">"Does this pedestrian crossing have an island?"</string>
     <string name="quest_surface_detailed_answer_impossible_title">Describe surface</string>
     <string name="open_url">Open URL</string>
+    <string name="quest_dietType_kosher_name_title">Does %s have kosher products in the offer?</string>
```
westnordost, line 939:
```suggestion
    <string name="quest_dietType_kosher_name_title">Does %s offer kosher (%s) products?</string>
```
Maybe add the Hebrew word [כשרות](https://he.m.wikipedia.org/wiki/%D7%9B%D7%A9%D7%A8%D7%95%D7%AA) in brackets behind the translation? (Also requires to override `getTitleArgs`)

matkoniecz:
> Maybe add the Hebrew word כשרות in brackets behind the translation? (Also requires to override getTitleArgs)

I am not sure is it adding anything, I would expect "kosher" to be clear and if someone is unfamiliar with text on signage distinguishing between yes/only would be anyway problematic.

app/src/main/res/values/strings.xml
matkoniecz: It is mostly to discourage someone from spamming mostly useless tag in location where nothing whatsoever is kosher. But that may be overly cautious.

westnordost:
But you just merged that suggestion. Anyway, now the modification of `getTitleArgs` is missing.
The reason behind my suggestion is that some certification labels may mention that word (i.e. do not have a "K"), so users can recognize it better.

app/src/main/res/values/strings.xml
westnordost: I think it is superfluous

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddKosher.kt
```diff
@@ -0,0 +1,34 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
+
+class AddKosher : OsmFilterQuestType<String>() {
+
+    override val elementFilter = """
+        nodes, ways with
+        (
+          amenity ~ restaurant|cafe|fast_food
+          or amenity = ice_cream
+          or shop = butcher
```
peternewman, line 15: shop = supermarket too and bakery:
https://www.kosher.com/learn/about-kosher-laws/kosher-bread

westnordost:
If you finish this this weekend, it can go into the next version

app/src/main/res/values/strings.xml
matkoniecz: What about

> This quest type is disabled by default because it is not useful in some regions and you will likely have to enter the place to be able to answer a quest of this type.

?

westnordost: 🤷 
Sounds better.

matkoniecz:
> Just K? The German Wikipedia has some pictures of certification logos:
> https://de.wikipedia.org/wiki/J%C3%BCdische_Speisegesetze#Zertifizierung

> The K-Logo could be used (maybe some details deleted because they are too small)

This one sadly looks like a copyright violation, but based on ideas from such certification logos I tried again. Result seems less horrifically ugly.

screen

matkoniecz:
I noticed a bug that diet explanation is not show at all, it happens also in vegetarian and vegan quests - in a released version.

I will try to debug it.

matkoniecz:
I have found source of a bug: https://github.com/streetcomplete/StreetComplete/blob/cc7624d2f8c83c037f4a078eb56dadd3b1cd80a2/app/src/main/java/de/westnordost/streetcomplete/settings/ShowQuestFormsActivity.kt#L119

Now I need to figure out how bundleOf can be merged

matkoniecz:
Huh, that is embarrassing. I have no idea how to merge bundles.

matkoniecz:
Marking as ready for review as bug is not critical and is happening in base SC anyway.

matkoniecz:
It seem to me that https://github.com/streetcomplete/StreetComplete/pull/2244/commits/77eb5764fbdcf0025ed5a26a0d70689b6f599cfe is a bit ugly, but I have no better idea how to do it.

matkoniecz:
Now with diet info bug gone:

screen01

westnordost (commented):
Should also be part of an achievement, the citizen one I guess

matkoniecz:
Done.

matkoniecz:
I probably should also cleanup res/graphics/quest icons/kosher.svg - https://github.com/streetcomplete/StreetComplete/commit/e918d84386244117f927d6bbb51edb04c666b972

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddKosher.kt
peternewman: Did the OSM Israel people suggest bakery wasn't worthwhile then @matkoniecz ?

PR #2247, opened 2020-11-07
for cycleway quest skip ones explicitly marked as with separately mapped sidewalks

matkoniecz:
for example https://www.openstreetmap.org/way/443761961#map=19/50.07290/19.88914


matkoniecz:
wait, I can just commit directly for such small suff

FloEdelmann:
The "Displayed separately on map" option needs to be added to the cycleway quest as well.

> wait, I can just commit directly for such small stuff

Opening a PR notifies watchers of the repo (like me). If you committed directly, I wouldn't have seen this change.

matkoniecz:
> The "Displayed separately on map" option needs to be added to the cycleway quest as well.

It is a bit tricky as map style is not allowing to distinguish between "footway is mapped as a separate way" and "cycleway is mapped as a separate way" so in such cases note is needed (or check in Vespucci or OSM Go !).

While for footways it is is extremely unusual to have cycleway, without footway which is unusuable by pedestrians.

PR #2269, opened 2020-11-14
Add lanes quest

### LINKED ISSUE #856: Lanes for streets
### General

We in Germany got many country roads which doesn't have center divider line and are very narrow. This would be a very useful information for routers, because you cannot drive that fast on those roads.

So it would be interesting of a unclassified road has 2 lanes, 1.5 or just 1 lane, where you need to find a wide place to pass other vehicles.

In cities this is a very bad idea to implement this, cause the lane numbers often change, which would make way splits necessary, but in rural environments this would be possible.

### Checklist

### Ideas for implementation

**Element selection:** 

If there's no lane tag for a highway=unclassified|tertiary|track(+grade1)|service" and it's not inside a city (maybe determine by the speed restrictions) than ask for the lane numbers: 1, 1.5, 2.

**Proposed GUI:** 

If 3 or more as answer should be possible we need to display the street like for bike lanes, to make it possible to select a direction per lane count.

I don't feel this is necessary, most lane counts should be below 3.

westnordost:
Fixes #856


app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
```diff
@@ -0,0 +1,72 @@
...
+            changes.deleteIfExists("turn:lanes:both_ways")
+        }
+
+        when(answer) {
+            is MarkedLanes -> {
+                if (answer.count == 1) {
+                    changes.deleteIfExists("lanes:forward")
+                    changes.deleteIfExists("lanes:backward")
+                } else {
+                    changes.modifyIfExists("lanes:forward", (answer.count / 2).toString())
```
matkoniecz, line 49: If I am reading correctly it is displayed as `answer.count - answer.count / 2`

From `setTotalLanesCount` 

```
            leftSide = lanes / 2
            rightSide = lanes - lanes / 2
```

(this time I started from looking at source code, it may have noticeable effects even for odd lane count, but seemed suspicious)

app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
```diff
@@ -0,0 +1,72 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
+
+class AddLanes : OsmFilterQuestType<LanesAnswer>() {
+
+    override val elementFilter = """
+        ways with
+          highway ~ ${ROADS_WITH_LANES.joinToString("|")}
+          and (!lanes or lanes !~ [1-9][0-9]*)
```
matkoniecz, line 12: What about cases with known bicycle lanes (bicycle lanes are not counted in `lanes`)? Maybe skip them?

See say https://taginfo.openstreetmap.org/tags/cycleway=lane#combinations - about 130 000 of `cycleway=lane` have no `lanes` tag and mistake here by user is quite likely.

app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
matkoniecz: `answering marked lanes with previous marked for each side answer` test seems to ensure that it is not a problem.

app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
```diff
@@ -0,0 +1,72 @@
...
+            }
+            is MarkedLanesSides -> {
+                changes.addOrModify("lanes:forward", answer.forward.toString())
+                changes.addOrModify("lanes:backward", answer.backward.toString())
+            }
+        }
+    }
+
+    companion object {
+        private val ROADS_WITH_LANES = listOf(
```
matkoniecz, line 65: I was unsure about _links but for motorway_link it seems that asking will be useful: http://overpass-turbo.eu/s/107h (1/3 of cases where `lanes` is tagged motorway link is larger than smallest possible one)


For primary link about 1/5: http://overpass-turbo.eu/s/107i

Yes, it is possible that every single interesting `lane` value was tagged for them in some regions - but even if in some regions asking for lane count on links is not useful anymore then it still will not be a spam, links are not appearing so often

app/src/main/res/values/strings.xml
```diff
@@ -243,10 +243,14 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_cycleway_title2">"Is there a cycleway here? What type?"</string>
     <string name="quest_cycleway_resurvey_title">"Is this still the cycleway situation here?"</string>
     <string name="quest_serviceType_title">"What kind of service road is this?"</string>
-    <string name="quest_streetLanes_title">"How many lanes does this road have?"</string>
-    <string name="quest_streetLanes_description">"If there are no road markings, it has no lanes."</string>
-    <string name="quest_streetLanes_answer_noLanes">"It has no marked lanes"</string>
-    <string name="quest_streetLanes_noLanes_estWidth">"Please estimate how many cars would fit alongside each other on this road."</string>
+    <string name="quest_lanes_title">"How many lanes for cars does this road have?"</string>
+    <string name="quest_lanes_answer_lanes">"Marked lanes"</string>
+    <string name="quest_lanes_answer_lanes_description">"Specify the total number of usable marked lanes (both sides). So, lanes occupied with parking cars don't count."</string>
```
matkoniecz, line 248: Maybe "lanes occupied with parking cars" -> "lanes for parking cars"

"occupied" seems possible to be misinterpreted as "occupied right now"? "for" seems shorter and also correct and without this issue

Not sure how to handle edge case of lanes available for driving for part of the day and as parking during evening and night.

app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
westnordost: Yes, bicycle lanes are not counted in lanes. The quest also asks for "lanes for cars".

I don't understand your concern. Skip what? Roads with cycle lanes? Why?

app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
westnordost: In general the `answer.count` should always be an even number, except for oneway roads.
Lanes count passe to `setTotalLanesCount` may be odd, because when selecting unmarked lanes, it is allowed to specify an odd number.

app/src/main/res/values/strings.xml
westnordost: I don't want it to sound like that the user has to look for whether parking is allowed here or not but just look at the de-facto state of the road. If a side of a road is used for parking, even when everyone went to work, there will be one or another car left to denote that cars park here I think.

What about "So, lanes used by parking cars don't count"?

westnordost: I think I'll leave it as it is for now, because
- "lanes for parking cars" slightly suggests that user ought to look for official markings or signs
- "lanes used by parking cars" sounds not as nice as "occupied with" IMO and "occupied" also suggests (and I think it ought) that if there are a few parking cars scattered around (which act like traffic calming) that it is doesn't mean that the lane should not count at all

sun-geo:
Quest question wording, maybe consider to change
from
"How many lanes for cars does this road have?"
to
"How many lanes does this road have for cars?"
?

westnordost:
Don't see a big difference. Is English your native tongue?

smichel17:
If it is only bike lanes we are concerned with, I would say, something like

How many lanes does this road have? Not counting bike lanes.

Atrate:
@sun-geo I feel like having the Subject at the end is easier for non-native speakers, since that's how making questions in English is taught at a base level - by inverting the SVO structure.

Splitting the object and having it appear before V+S with "for cars" referencing it may also add to confusion for non-native speakers.

RubenKelevra:
Hey @westnordost,

thanks that quest is really nicely made!

Two things could be improved I think:

- Can you slow down the cars zipping by? They are a bit fast I think and might be confusing.
- How are the users expected to treat a road which is 'basically' wide enough but there are parked cars on one or both side, so the road can only be used in one direction at a time? 🤔

RubenKelevra:
I noticed that we only tag the number of lanes when the user is answering that there are lane markings. Why? We could set `lane_markings=yes` to explicit tag that the road is marked with 2 individual lanes instead of relying on assumptions.

matkoniecz:
> `lane_markings=yes` to explicit tag

This seems too verbose tagging for me...

> so the road can only be used in one direction at a time?

I would expect it to be tagged as a single lane.

RubenKelevra:
> > `lane_markings=yes` to explicit tag
> 
> This seems too verbose tagging for me...

Well, we asked the user explicitly for road markings, so why not tag it explicitly? Should come handy for autonomous vehicles for example - as a possible data consumer. If it's yes, they can expect to find some kind of road makings in the middle, if no they won't try to. 

I mean, it doesn't hurt to explicitly tag this while it can this way be destinguished from roads which have just some random number on them without any info if there are road markings. 


> > so the road can only be used in one direction at a time?
> 
> I would expect it to be tagged as a single lane.

Okay, how would you tag this?

IMG_20201117_151311__01

westnordost:
Difficult, so cars pass each other here or does one car need to wait back there until the first passed the narrowish section?

westnordost:
From just the picture I'd assume that two cars can pass each other

westnordost:
> How are the users expected to treat a road which is 'basically' wide enough but there are parked cars on one or both side, so the road can only be used in one direction at a time? 🤔

`lanes=1` definitely. Discussion on improving the wording can happen here: #2284

RubenKelevra:
> Difficult, do cars pass each other here or does one car need to wait back there until the first passed the narrowish section?

They do both... 🤔 

Sometimes a driver waits and the other one passes and sometimes the pass each other slowly. But in general this has very little traffic - so hard to observe.

They can pass each other, but only if the cars are not large SUVs etc. 

That's basically where I started to tag 1.5 lanes - you can pass, but you have to slow down below the speed limit to do this.

RubenKelevra:
> From just the picture I'd assume that two cars can pass each other

True, but SC asks for 'comfortable'... 🤔

westnordost:
Hm, I don't know. The wiki amongst other things says this:

>  Many ways have not yet been tagged with the total number of lanes at all points, but only with the number of through lanes of a longer section. Therefore, data consumers can mostly treat the lanes tag as a minimum rather than an exact number. 

This suggests that rather it should be rounded down in case of doubt.

RubenKelevra:
Alright :)

thibaultmol:
Is it intended behaviour that the 'ok' button doesn't show up when you select only 1 lane?
https://photos.app.goo.gl/XDRUg3QiLpvsQx5k9

matkoniecz:
> Is it intended behaviour that the 'ok' button doesn't show up when you select only 1 lane?

No, and this bug is reported and fixed ( https://github.com/westnordost/StreetComplete/issues/2288 ).

In the next update, maybe already available for you, it will be fixed (see https://github.com/westnordost/StreetComplete/releases/tag/v27.0-beta2 )

sun-geo:
> Don't see a big difference. Is English your native tongue?

Okay, just was thinking if it is easier to read/understand. No. So seems fine if it stays like it is.

HolgerJeromin:
Sorry for commenting in a closed issue but: 
Should we add a hint what to to do with the turn lane area before a crossing? Mostly the user should split the way at the correct position:
image

https://wiki.openstreetmap.org/wiki/Key:turn#Motorway_with_links_and_destinations

PR #2275, opened 2020-11-15
Update README.md

jamescridland:
The latest release notes aren't easily visible on the mobile version of Github in particular (they hide at the bottom of a long page, and it's unclear that the release functionality has a changelog in it). It would be great to make them more obvious, since they show that this project is alive and being worked upon.


westnordost:
Sure, why not

PR #2335, opened 2020-12-02
Existence

westnordost:


FloEdelmann (changes requested):

HolgerJeromin:
I got asked if there is still a waste basket: https://www.openstreetmap.org/node/6853026476
This is imo very early after one year.

westnordost:
It should be every 2 years. Do you have "ask more often" activated in the settings?

HolgerJeromin:
> It should be every 2 years. Do you have "ask more often" activated in the settings?

Yes i do. I am fine with beeing spammed :-P 

Every 2 years is probably ok-ish.

matkoniecz:
Every two years for waste basket still fills very often.

Lukas458:
Hi, I have a question concerning the time periods of resurvey. Is the actual setting like this, that amenity=telephone nodes get surveyed every 2 years? My question is, why was https://www.openstreetmap.org/node/270119900/history resurveyed ten, before 2 years were over in the check_date of the feature?

westnordost:
In the app settings, you can specify to ask 2x more often or 2x less often. I assume that the user changed this option from the default.

Lukas458:
Okay well, I see. So the most often period to ask would then be half a year for objects which have the SC-default setting to be asked for every 2 years?

michaelblyons:
> Okay well, I see. So the most often period to ask would then be half a year for objects which have the SC-default setting to be asked for every 2 years?

If I understand correctly, no. The default for `amenity=telephone` is 2 years. You can tweak your own settings to make that 2x _more often_ (every 1 year) or 2x _less often_ (every 4 years).

In no case would a 2 year item be updated every 0.5 years.

westnordost:
Correct

Lukas458:
Okay, thanks for your answer.

Lukas458:
Hi, I have some question which came to my mind,
maybe it would be possible to ask the existence question also for the objects _amenity=ticket_validator_ and _emergency=phone_? I think these are also little "street objects" which might be worth to resurvey, let's say all 4 years.
Or do we need another issue ticket for that?

matkoniecz:
> emergency=phone

See #2744

> amenity=ticket_validator

If worth resurveying it may be worth adding there (no idea how often such things move/are getting removed and would be mapped in remapping)

PR #2363, opened 2020-12-10
Update some links to the org

peternewman:


matkoniecz:
@westnordost Is it deliberate that it is  streetcomplete organisation, not StreetComplete?

westnordost:
Yeah, kinda, because the other orgs I know also use lowercase. Anyway,

https://github.com/sTrEeTcOmPlEtE/StreetComplete/pull/2363

works too

westnordost:
Thanks, @peternewman

peternewman:
> Thanks, @peternewman

Well it seemed only fair given I'd generated some extra work.

westnordost:
Well, thanks also for the suggestion to move it into an org. I think this is the right step and I am glad that you would like to help in this regard.

This will make it easier in the future to add more people that can help triage the issues or even make small corrections on the app. As you all know, I can not forever keep up this kind of activity in this intensity, so this is one step to move the project away from being carried on by my and more into the (OSM) community.

PR #2377, opened 2020-12-16
surface for pitches, fixes #1170

### LINKED ISSUE #1170: [Quest] surface of leisure=pitch
### General
Affected tag(s) to be modified/added: [surface of leisure:pitch](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpitch)
Question asked: **What type of surface does this pitch have?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose (in tennis or football surface changes playing technique slightly, also one may prefer other types of surface than concrete or asphalt)

### Ideas for implementation
Picture selection quest.

**Element selection:** 
ways with leisure=pitch and without surface tag, nodes could be too ambiguous as few pitches that are close together may be tagged as one

matkoniecz:


matkoniecz:
I still need to cobble together some icon and find matching photos, but everything else is ready for review (I will finish also kosher quest, but today I had urge to write something new rather than finish something)

matkoniecz:
Though obviously if @westnordost has an idea and time for icon I would be very happy to use it :)

matkoniecz:
Now I need to figure out how to handle paved/unpaved/ground.

I though about dropping ground and having some mix of paving stones and asphalt for paved.

westnordost:
Thanks for your work so far on this!

> I had urge to write something new rather than finish something

I consider this a bad habit and try to avoid it whenever possible.

> Though obviously if @westnordost has an idea and time for icon I would be very happy to use it :)

There is already an icon. See pitch_surface.svg

> I though about dropping ground and having some mix of paving stones and asphalt for paved.

Especially ground should not be dropped. Many pitches are ground surface, even in developed countries for more or less "informal" pitches.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -0,0 +1,49 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+
+
+class AddPitchSurface : OsmFilterQuestType<SurfaceAnswer>() {
+        override val elementFilter = """
+        ways with leisure=pitch
+        and sport ~ soccer|tennis|basketball|equestrian|athletics|volleyball|skateboard|bmx|american_football|rugby_union
```
westnordost, line 12: Why limit the sports?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -0,0 +1,49 @@
...
+                    changes.deleteIfExists("surface:note")
+                }
+                is GenericSurfaceAnswer -> {
+                    changes.updateWithCheckDate("surface", answer.value)
+                    changes.addOrModify("surface:note", answer.note)
+                }
+            }
+            changes.deleteIfExists("source:surface")
+        }
+}
```
westnordost, line 49: Check the indent in this class

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurfaceForm.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override val itemsPerRow = 3
+
+    override fun onClickOk(selectedItems: List<String>) {
+        val value = selectedItems.single()
+        if(value == "paved" || value == "unpaved") {
+            AlertDialog.Builder(requireContext())
+                .setMessage(R.string.quest_surface_detailed_answer_impossible_confirmation)
+                .setPositiveButton(R.string.quest_generic_confirmation_yes) {
+                    _, _ -> run {
```
westnordost, line 24: `run`? Think this can be removed

westnordost:
The pictures for the new surfaces look good. However, this

> I considered also https://commons.wikimedia.org/wiki/File:200_metres_track_running_start_line.jpg for generified surface=tartan

looks much better than what is currently checked in.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
matkoniecz: I got scared by by https://taginfo.openstreetmap.org/tags/sport=table_tennis (confusing), https://taginfo.openstreetmap.org/tags/sport=shooting#combinations (where quest would be pointless), https://taginfo.openstreetmap.org/tags/sport=beachvolleyball#combinations (here it would be pointless, it should be always sand), and baseball for example has clay + grass pitches (multiselect of surfaces is in principle possible).

https://taginfo.openstreetmap.org/tags/sport=golf#combinations should be both pointless and requiring multiple surfaces, https://taginfo.openstreetmap.org/tags/sport=boules is always surface=compacted etc

matkoniecz:
> I consider this a bad habit and try to avoid it whenever possible.

I know, but there is only one ,y PR open and I finished multiple other things so I thought that it should be OK.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
westnordost: Right, but then your list of sports is a little small. Maybe have a negative list instead?

matkoniecz: Which sports are missing?

For negation it would be necessary to add syntax that would have regexp matching without adding `^` `$` (to skip also say `shooting;archery`, `ice_skating;ice_hockey`, `cricket;australian_football`, `skating;hockey`, `discus_throw;hammer_throw` etc).

Also, based on my experience with blacklisting values - that is just never ending quest for adding new variables, so I would strongly prefer whitelisting. I used blacklist in some places in my code and it was very irritating in its outcome.

matkoniecz:
> The pictures for the new surfaces look good. However, this

> >  I considered also https://commons.wikimedia.org/wiki/File:200_metres_track_running_start_line.jpg for generified surface=tartan

> looks much better than what is currently checked in.

Changed.

And just in case that it repeats - for pitch specific `surface=pvaed` image is (a bit ironically) from https://commons.wikimedia.org/wiki/Category:Concrete_roads_in_the_United_States

https://commons.wikimedia.org/wiki/File:M%C3%BClheim_an_der_Ruhr,_Radschnellweg_Ruhr,_2020-04_CN-01.jpg
https://commons.wikimedia.org/wiki/File:2014-08-29_15_35_59_View_northwest_along_Stuyvesant_Avenue_at_the_border_of_Trenton_and_Ewing,_New_Jersey,_with_concrete_pavement_likely_dating_to_the_1950s.JPG
https://wiki.openstreetmap.org/wiki/File:Witosa_bike_2.jpg
https://commons.wikimedia.org/wiki/File:M%C5%82yn%C3%B3wkaKr%C3%B3lewska-OkoliceAleiKijowskiej-POL,_Krak%C3%B3w.jpg
https://commons.wikimedia.org/wiki/File:Passannante_Ballfield_td_(2019-03-27)_23.jpg
https://commons.wikimedia.org/wiki/File:Bulgarska_Poznan_bike.JPG

were among considered sources (I wanted concrete + asphalt, not asphalt + paving_stones)

matkoniecz:
> There is already an icon. See pitch_surface.svg

Great! Is it OK to import also other not yet import quest icons into drawables? Or is it preferable to keep them this way for easier editing?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
matkoniecz: reintended everything using AS

matkoniecz:
What remained:

https://github.com/streetcomplete/StreetComplete/pull/2377#discussion_r544679711 - sport filtering ("only this" vs "except"), possibly some sports are missing.

Done, not sure about it:

- images for ground and unpaved - not the biggest fan of them but plenty of searching failed to find something better
- not showing `surface=wood`, `paving_stones` `metal` `compacted`, `unhewn_cobblestone` and other surfaces. I dropped them deliberately as I am not aware about pitches with them and I assumed that what is in OSM is a tagging mistake.

But maybe all surfaces up to and including rock should be available?

matkoniecz:
Done, also not sure about it:

position on the list of quests - put in last category but first as it seems more likely to be useful than other hings listed there, and I expect it to be nice in solving (maybe someone will find pitch that (s)he was unaware of, it may detect some pitches that are gone but not deleted from OSM, not generating entire flocks of quests like backrest one etc)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
westnordost: > Which sports are missing?

Uf, I will not give you a list, but just skimming that I noticed canadian football, australian football, softball, baseball, cricket, that other rugby, gaelic games and sepak takraw.

westnordost: > For negation it would be necessary to add syntax that would have regexp matching without adding ^ $

`.*ice_hockey|ice_skating|table_tennis.*`

westnordost:
> Great! Is it OK to import also other not yet import quest icons into drawables? Or is it preferable to keep them this way for easier editing?

It's ok

westnordost:
I think all surfaces should be available

matkoniecz:
> I think all surfaces should be available

Now also surfaces are available.

I was thinking about surface=parquet but it has basically no use (thought it is mentioned for example at https://wiki.openstreetmap.org/wiki/Tag:sport%3Dbasketball )

matkoniecz:
I am getting some weird crashes - almost certainly unrelated, but...

westnordost:
but?

matkoniecz:
I worried that it may be caused by bug in this PR, but was rather caused by messing with how SC is installed (manual installs) and bugs in my fork - not in SC or in this PR.

Since I uninstalled and installed fresh SC it works well.

matkoniecz:
I tested it a bit, and it works.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -0,0 +1,56 @@
...
+class AddPitchSurface : OsmFilterQuestType<SurfaceAnswer>() {
+    private val sportValuesWherePitchSurfaceQuestionIsInteresting = listOf(
+        "soccer", "tennis", "basketball", "equestrian", "athletics", "volleyball", "skateboard",
+        "bmx", "american_football", "rugby_union", "badminton", "pelota", "horse_racing",
+        "disc_golf", "futsal", "cycling", "gymnastics"
+    )
+
+    override val elementFilter = """
+        ways with leisure=pitch
+        and sport ~ ${sportValuesWherePitchSurfaceQuestionIsInteresting.joinToString("|")}
```
westnordost, line 19:
```suggestion
        and sport ~ .*${sportValuesWherePitchSurfaceQuestionIsInteresting.joinToString("|")}.*
```
... to catch pitches with multiple sports like `sport=soccer;basketball`?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -0,0 +1,56 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+
+
+class AddPitchSurface : OsmFilterQuestType<SurfaceAnswer>() {
+    private val sportValuesWherePitchSurfaceQuestionIsInteresting = listOf(
+        "soccer", "tennis", "basketball", "equestrian", "athletics", "volleyball", "skateboard",
+        "bmx", "american_football", "rugby_union", "badminton", "pelota", "horse_racing",
+        "disc_golf", "futsal", "cycling", "gymnastics"
+    )
```
westnordost, line 15: \+
multi
fitness
bowls, boules
netball
handball, team_handball
hockey, field_hockey
futsal
badminton
padel
pelota
rugby_league
gaelic_games
horseshoes
australian_football
racquet
?

Well, I could go on. Just on page 2 of taginfo.

matkoniecz: multi - may have "bad" values and will be anyway retagged by higher priority quest for selecting sport of a pitch
fitness - not tagged on pitches, and what is tagged looks in almost all cases like a tagging mistake (leisure=fitness_station)
bowls - on review it makes sense, added
boules - I thought that it is `surface=compacted` but on recheck of what is mapped turned out to be not true, added
netball - rechecked, added
handball, team_handball - makes sense, added
hockey - retagged by sport quest, ambiguous, may mean ice hockey ( https://wiki.openstreetmap.org/wiki/Tag:sport%3Dhockey )
field_hockey - added
futsal - is listed already
badminton - is listed already
padel - added
pelota - is listed already
rugby_league - added
gaelic_games - added
horseshoes - added
australian_football - Wikipedia claims that it is always played on grass, but looking at what is actually tagged it turned to be untrue, added
racquet - added

I must say that blacklist seems to now a better idea than initially.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
matkoniecz: Baseball was skipped deliberately as it is standard that field will have parts of different surfaces (grass + clay), at least a bit better ones (according to what people from USA told me).

As I understand the same is true for cricket: https://en.wikipedia.org/wiki/Cricket#/media/File:Muralitharan_bowling_to_Adam_Gilchrist.jpg (though looking again at actual tagging, it seems to be done differently? Added after all)

Others are added already or I added them right now.

And blacklist would need to be a bit longer what in turn convinces me that whitelist was not so bad idea...

`.*multi|ice_hockey|([^_]^)hockey|ice_skating|table_tennis|baseball|beachvolleyball|climbing|shooting|archery|scuba_diving|canoe|skiing|free_flying.*`

(for example there are some marked canoe pitches in places where boats are stored or member meet, asking about ice_skating or beachvolleyball is quite pointless etc)

matkoniecz: And further ones that would end on a blacklist: 10pin model_aerodrome skating billiards trampoline 9pin rowing sailing climbing_adventure long_jump ice_stock ski_jumping (I have not checked whatever this is tagged on `leisure=pitch` but after canoe pitches on land I assume that at least some of them are).

Not on pitches for now but maybe would be better to blacklist just in case: orienteering

Not sure what to do with bullfighting and chess (chess on areas is a tagging for oversized chess boards where asking about surface would make sense, but question there wold be confusing)

On additional check: everyone tags baseball pitches with just a single value, so it is back on a whitelist and should be dropped from blacklist.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
westnordost: > multi - may have "bad" values and will be anyway retagged by higher priority quest for selecting sport of a pitch 

Not it won't.

westnordost:
Maybe exclude all sport pitches tagged with `indoor=yes`, too

matkoniecz:
And next part of looking at list of sports:

skipped: polo (sems to be laways grass), boxing (noone tags surface on pitches - 3 cases worldwide), martial_arts and weightlifting (even less), judo and taekwondo and pilates and crossfit (even less), rc_car and car_racing (confusing for users, likely includes paved and unpaved areas), dog_racing (rarely have pitches with surface tagged), karate (the same), dog (????????)

Some were added: paddle_tennis korfball petanque pickleball lacrosse croquet four_square tetherball shot-put

Researched till 7th page of https://taginfo.openstreetmap.org/keys/sport#values - I think that it is enough...

matkoniecz:
> Maybe exclude all sport pitches tagged with indoor=yes, too

I though about it but in the end just copied what was for sport.

But given that indoor ones are likely to have parquet or some weird synthetic surface AND will be likely hard to survey it is definitely better to exclude them.

I will also exclude cases where `leisure=pitch` is also tagged with `buulding=*` (https://taginfo.openstreetmap.org/tags/leisure=pitch#combinations - 10k cases)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
matkoniecz: > Not it won't.

Right, I was convinced that I remember well and failed to verify it. Added to list of accepted values.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -0,0 +1,61 @@
...
+        "handball", "team_handball", "field_hockey", "padel", "horseshoes", "tetherball",
+        "gaelic_games", "australian_football", "racquet", "rugby_league", "rugby_union", "rugby",
+        "canadian_football", "softball", "sepak_takraw", "cricket", "pickleball", "lacrosse",
+        "roller_skating", "baseball", "shuffleboard", "paddle_tennis", "korfball", "petanque",
+        "croquet", "four_square", "shot-put",
+    )
+    
+    override val elementFilter = """
+        ways with leisure=pitch
+        and sport ~ "(^|.*;)(${sportValuesWherePitchSurfaceQuestionIsInteresting.joinToString("|")})(${'$'}|;.*)"
```
westnordost, line 23:
```suggestion
        and sport ~ "(^|.*;)(${sportValuesWherePitchSurfaceQuestionIsInteresting.joinToString("|")})($|;.*)"
```
Doesn't this work?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -0,0 +1,61 @@
...
+            and !surface:note
+            and !note:surface
+          )
+        )
+    """
+
+    override val commitMessage = "Add pitch surfaces"
+    override val wikiLink = "Key:surface"
+    override val icon = R.drawable.ic_quest_pitch_surface
+    override val isSplitWayEnabled = true
```
westnordost, line 40: split way enabled for pitches? This must be an error, pitches are areas, aren't they?

app/src/test/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurfaceTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    private val questType = AddPitchSurface()
+
+    @Test fun `not applicable to pitch area without tagged sport`() {
+        val mapData = TestMapDataWithGeometry(listOf(
+            OsmWay(1L, 1, listOf(1,2,3), mapOf(
+                "leisure" to "pitch",
+            ))
+        ))
+        assertEquals(0, questType.getApplicableElements(mapData).toList().size)
+    }
```
westnordost, line 19: The test is unnecessarily complex. For `OsmFilterQuestType`s, `getApplicableElements` directly refers to `isApplicableTo`, also, you are testing very similar things, so lets add a few helper functions
```suggestion
    @Test fun `not applicable to pitch area without tagged sport`() {
        assertIsNotApplicable("leisure" to "pitch")
    }
    
    private fun assertIsApplicable(vararg vararg pairs: Pair<K, V>) {
        assertTrue(questType.isApplicableTo(create(*pairs)))
    }
    
    private fun assertIsNotApplicable(vararg vararg pairs: Pair<K, V>) {
        assertFalse(questType.isApplicableTo(create(*pairs)))
    }
    
    private fun create(vararg vararg pairs: Pair<K, V>) = OsmWay(1L, 1, listOf(1,2,3), mapOf(pairs))
```

If you do this for all the tests, you'll shrink it somewhat. Add more helper functions like

```kotlin

```

westnordost (commented):
The quest should maybe be part of an achievement, too. The outdoors one.

app/src/test/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurfaceTest.kt
westnordost: Maybe even put it all in one test, I know it is considered not that good of a style but if the `assert...` statement already reads self-explanatory, it feels like unnecessary duplication to have the title of the test also in the function name.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
matkoniecz: If anyone sees anything missing/wrong or thinks that further research is needed - let me know.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
matkoniecz: In the end I think that having exemption list and valid list is equally irritating, and false negatives are better than false positives.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
westnordost: It doesn't?

app/src/test/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurfaceTest.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    @Test fun `not applicable to pitch area with single invalid sport`() {
+        assertIsNotApplicable("leisure" to "pitch", "sport" to "table_tennis")
+    }
+
+    @Test fun `applicable to pitch area with both valid and invalid sport`() {
+        assertIsApplicable("leisure" to "pitch", "sport" to "multi;table_tennis")
+    }
+
+    @Test fun `not applicable to pitch area with with sport value that is substring of valid one`() {
```
westnordost, line 40:
```suggestion
    @Test fun `not applicable to pitch area with sport value that is substring of valid one`() {
```

westnordost (approved):

PR #2380, opened 2020-12-17
clarify that it applies to just undo within SC

matkoniecz:


matkoniecz:
If this would be accepted it would be probably preferable to do it as a new string.

westnordost:
Hm, rather not. The user doesn't need to know that it can be undone **somehow**. If you delete a file, it will also not warn you that "You cannot undo this in the Explorer" but "You cannot undo this".

Issue #2285, opened 2020-11-16
Provide hint in "can't say" menu for cycleways than you can tag "cycleway:both=no" with SC


matkoniecz:
**Use case**
Multiple users left massive amount of notes "there is no cycleway of any kind here".

I dismissed it, but it keeps happening. This time in Poland with 90 notes by user who was confused by StreetComplete interface.

http://resultmaps.neis-one.org/osm-notes-country?c=Poland - part of notestorm here
![x9wUJp1](https://user-images.githubusercontent.com/899988/99306900-16d09900-2856-11eb-8ab8-5fcf417b5e0c.png)

Triggered by https://forum.openstreetmap.org/viewtopic.php?pid=808508#p808508

**Proposed Solution**
Have in "can't say" menu entry for "no cycleway of any kind here" that would either tag `cycleway:both=no` or educate user how to solve the quest.

Maybe also make "no cycleway" answer more prominent with some big red X as part of the icon in selecting answers?

EDIT: more links https://www.openstreetmap.org/note/2465507 https://www.openstreetmap.org/note/2465506 https://www.openstreetmap.org/note/2465505 https://www.openstreetmap.org/note/2455823

matkoniecz:
It happened earlier, more than once, but I kept dismissing it as "isolated incident, also interface changed since that time".

FloEdelmann:
> Maybe also make "no cycleway" answer more prominent with some big red X as part of the icon in selecting answers?

+1 from me for this suggestion. I've needed to search the correct graphic for "no cycleways" multiple times. Also, the sidewalks quest's answers do have a red :x: and a green :heavy_check_mark:, so it'd make the user interface more consistent.

westnordost:
Should the answer option educate people about how to select it, or (pre)select it?

matkoniecz:
> Should the answer option educate people about how to select it, or (pre)select it?

Probably educate. Someone confused about this is likely to be confused how to fill it.

It is also possible that someone failed to consider some more exotic bicycle infrastructure such as contraflow without a dedicated lane.

But it would offer a quicker way to solve this common case: "can't say" -> "select case" -> "confirm"

travnick:
The whole issue was because the person that was adding the comments didn't see the "no cycle way" option on mobile because of screen size/not obvious icon.

Sorry for that.

westnordost:
Assuming you are that person, how could this be made more clear?

travnick:
[...]

PR #2385, opened 2020-12-18
make 'no bicycle infrastructure here' more visible, first part of #2285

matkoniecz:
I want to add also hint, but it is also useful and would be nice to refer to some clearly visible symbol

my main worry here is risk that people will interpret it as 'no cycling allowed' but hopefully it is unlikely


matkoniecz:
screen18

matkoniecz:
Now that I look at it - maybe also move "none" to the first position?

matkoniecz:
Tested in country with left-hand traffic, in night mode, in night mode in country with left-hand traffic.

theraser:
Thanks, nice! Would it be possible to use exactly the same icon as for sidewalks for a more similar feeling/look? It seems that the ❌ in your version is more in the middle of the icon that at the sidewalks one. Moving it more to the left/right maybe would avoid your concerns:

> my main worry here is risk that people will interpret it as 'no cycling allowed' but hopefully it is unlikely

matkoniecz:
Yes, it is moved to a side.

This was deliberate to avoid interpreting it as "no cycleway along cariageway" but rather as "no cycleway infrastructure at all".

I though also about changing it to another symbol like a sad face, or empty set symbol but I was not convinced that it would be more clear.

westnordost:
The X would be visible also in the preview, wouldn't it? That's not consistent with the sidewalk quest

matkoniecz:
> The X would be visible also in the preview, wouldn't it?

Yes.

> That's not consistent with the sidewalk quest

So bikeway quest should be modified to allow displaying different graphic in selection menu and different in a lane view?

matkoniecz:
Personally I have no trouble with the difference but having it consistent is also fine.

westnordost:
I'd say so, yes

matkoniecz:
I see two options: 

(1) special casing this single situation in showCyclewaySelectionDialog

(2) duplicate entire `getLeftHandTrafficIconResId` `getRightHandTrafficIconResId`

Version (1) seemed nicer despite hard-coding, but it is happening in the different file. I will go with (2).

westnordost:
(3) special-casing this situation in 
```kotlin
fun Cycleway.getIconResId(isLeftHandTraffic: Boolean): Int =
    if (this == NONE) R.drawable.ic_cycleway_none_with_X
    else getPuzzleIconResId()

fun Cycleway.getPuzzleIconResId(isLeftHandTraffic: Boolean): Int =
    if (isLeftHandTraffic) getLeftHandTrafficIconResId() else getRightHandTrafficIconResId()
```

and use `getPuzzleIconResId` in `AddCyclewayForm`

matkoniecz:
I ended special casing in a bit different location, hopefully it is also OK.

westnordost (approved):

PR #2397, opened 2020-12-22
OpenStreetCam rebranded as KartaView

matkoniecz:
Description has no need for change.

Not sure whatever change id or keep it the same (what is the impact here?).

Not sure whatever https://kartaview.org/map/ or https://kartaview.org is preferable

Logo also needs to be updated.


westnordost:
Current impact is none, because it is not included in any achievement. Thank you for the info, I'll replace it real quick

matkoniecz:
> it is not included in any achievement.

Maybe it would fit thematically achievement car? Most people contributing have it in their car, AFAIK.

westnordost:
Mh yeah, maybe it would. You can create a PR for that if you like

matkoniecz:
OK, I will wait for a new id.

PR #2402, opened 2020-12-22
Construction quest now has an option "exact date is known" fixes #1341

### LINKED ISSUE #1341: Construction quest should have option "exact date is known"
**Use case**
Let's say I have an construction quest (https://github.com/westnordost/StreetComplete/issues/685). Actually it is not rare in Germany (at least) to see signs or so for bigger projects explaining the project and when it ends (estimated) or so.

**Proposed Solution**
As such, maybe we can use this info and add a "Other answer" -> "End date is known", which then opens a date picker (maybe only month/year though, as that is often only given).

matkoniecz:
available via can't say menu, a bit hidden but as discussed in #1341
it is hard to find a better place for it


app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedConstructionForm.kt
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    private fun setFinishDate() {
+        val tomorrow = Calendar.getInstance()
+        tomorrow.add(Calendar.DAY_OF_MONTH, 1)
+        val year = tomorrow.get(Calendar.YEAR)
+        val month = tomorrow.get(Calendar.MONTH)
+        val day = tomorrow.get(Calendar.DAY_OF_MONTH)
+        val dpd = DatePickerDialog(requireContext(), { _, yearSelected, monthSelected, dayOfMonthSelected ->
+            applyAnswer(OpeningDateAnswer(Date(yearSelected-1900, monthSelected, dayOfMonthSelected)))
```
westnordost, line 26: this Date-constructor is deprecated, see https://docs.oracle.com/javase/7/docs/api/java/util/Date.html#Date(int,%20int,%20int)

app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedHighwayConstruction.kt
```diff
@@ -1,15 +1,17 @@
...
+import android.util.Log
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.meta.ALL_ROADS
 import de.westnordost.streetcomplete.data.meta.SURVEY_MARK_KEY
 import de.westnordost.streetcomplete.data.meta.toCheckDateString
 import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
 import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
-import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+import de.westnordost.streetcomplete.quests.bus_stop_ref.BusStopRef
+import de.westnordost.streetcomplete.quests.bus_stop_ref.NoBusStopRef
```
westnordost, line 11: ?

app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedBuildingConstruction.kt
```diff
@@ -1,14 +1,14 @@
 package de.westnordost.streetcomplete.quests.construction
 
+import android.util.Log
```
westnordost, line 3: superfluous?

app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedHighwayConstruction.kt
```diff
@@ -1,15 +1,17 @@
 package de.westnordost.streetcomplete.quests.construction
 
+import android.util.Log
```
westnordost, line 3: superfluous?

app/src/main/res/values/strings.xml
```diff
@@ -956,4 +956,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
     <string name="quest_shop_gone_replaced_answer_error">"Don't know what this is. Consider leaving a note."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What\'s here now?</string>
+    <string name="quest_construction_completes_at_known_date">Completes at known date</string>
+    <string name="quest_construction_completion_date_title">When object will be open?</string>
```
westnordost, line 960:
```suggestion
    <string name="quest_construction_completion_date_title">When will it be complete?</string>
```

Though, not sure if a description here is necessary at all. After all, it should come to no surprise to the user to be asked to provide a date when he previously selected "Completes at known date"

app/src/main/res/values/strings.xml
```diff
@@ -956,4 +956,6 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
     <string name="quest_shop_gone_replaced_answer_error">"Don't know what this is. Consider leaving a note."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What\'s here now?</string>
+    <string name="quest_construction_completes_at_known_date">Completes at known date</string>
```
westnordost, line 959:
```suggestion
    <string name="quest_construction_completes_at_known_date">Finished at date...</string>
```
Even if you find your wording better, the "..." is important.

matkoniecz: I will go with yet another wording - "Opens at date…" (tripple dot character + make clear that it is when road be open, not when construction is finished - what matters in some cases)

app/src/main/res/values/strings.xml
matkoniecz: Sometimes I missclick or get interrupted and forget what I was doing, so I quite like to have confirmation what I am doing on every screen. But I am not strongly attached to it.

westnordost:
You chose the "open"-wording. However, this may apply to roads and businesses, but you wouldn't be able to say that a "apartment building is open".

matkoniecz:
> You chose the "open"-wording. However, this may apply to roads and businesses, but you wouldn't be able to say that a "apartment building is open".

OK, so I need to change it back. I basically learned about one more difference between Polish and English.

westnordost:
"apartment building is open" sounds to me like maybe it is "open" in the sense of that the roof is missing and thus at the top, it's open. ;-)

In German, you can say that a thing is "eröffnet" (=opened), but you can only say that for something public because it suggests that it is open to people to enter, i.e. also not for apartments.

westnordost:
Looks good now! Did you test it?

matkoniecz:
I changed again to "Will be finished at date…" to make clear that it is not about recording completion date that already happened.

I tested it, though I am running right now retest after all this text changes.

matkoniecz:
Still works, as expected.

westnordost (approved):

PR #2403, opened 2020-12-22
show mapillary and kartaview

matkoniecz:
not strongly fitting anywhere, but car fits best as the are typically used via cameras mounted in cars
I considered putting them also significantly later - level 12 or something, as only quite dedicated OSM mapper is likely to be interested in it


westnordost:
I'd also say to put them considerably later, because they are not directly things that have something to do with OSM while the other things mentioned there have.

PR #2404, opened 2020-12-22
put no cycleway infrastructure as the first answer in cycleway quest

matkoniecz:
in attempt to reduce amount of OSM notes about what is solvable in SC

sorry for people who are familiar with position of items, hopefully it will not be too bad
next part of #2285


westnordost:
ok... however, before it was also visible in the selection. I don't think it makes a big difference either way.

matkoniecz:
After closing 200+ notes "there is no cycleway here" and explaining to several people that it is actually taggable with SC - I want to use every single possible option to make it clearly visible.

Next one is tutorial in "can't say" menu.

PR #2405, opened 2020-12-22
 provide special tutorial how to tag case of no cycleway

### LINKED ISSUE #2285: Provide hint in "can't say" menu for cycleways than you can tag "cycleway:both=no" with SC

matkoniecz:
why no direct answer?
- person confused by this may be confused by this quest in general and it is good to have info how to use it
- directly tagging cycleway:both=no is likely to cause incomplete data for example in case of contraflows without contraflow lane,
I hope that someone seeing selections will be more likely to answer correctly

fixes #2285


app/src/main/res/values/strings.xml
```diff
@@ -956,4 +956,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
     <string name="quest_shop_gone_replaced_answer_error">"Don't know what this is. Consider leaving a note."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What\'s here now?</string>
+    <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway of any kind</string>
+    <string name="quest_cycleway_answer_no_bicycle_infrastructure_explanation">As usual tap at one side of a road, select a proper answer and do the same for the other side. See all options, some cases are not obvious, like oneway road allowing cyclists in two directions!</string>
+    <string name="quest_cycleway_answer_no_bicycle_infrastructure_title">Marking that there is no cycleway</string>
```
westnordost, line 961:
```suggestion
    <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway at all</string>
    <string name="quest_cycleway_answer_no_bicycle_infrastructure_explanation">As usual, tap on one side of the illustration, select a proper answer and do the same for the other side. Note all the available options: some cases are not obvious, like a oneway road that allows cyclists in both directions!</string>
    <string name="quest_cycleway_answer_no_bicycle_infrastructure_title">Selecting that there is no cycleway</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/bikeway/AddCyclewayForm.kt
```diff
@@ -30,9 +31,20 @@ class AddCyclewayForm : AbstractQuestFormAnswerFragment<CyclewayAnswer>() {
         if (!isDefiningBothSides && isNoRoundabout) {
             result.add(OtherAnswer(R.string.quest_cycleway_answer_contraflow_cycleway) { showBothSides() })
         }
+        result.add(OtherAnswer(R.string.quest_cycleway_answer_no_bicycle_infrastructure) { noBikewayHereHint() },)
```
westnordost, line 34:
```suggestion
        result.add(OtherAnswer(R.string.quest_cycleway_answer_no_bicycle_infrastructure) { noBikewayHereHint() })
```

westnordost (approved):

PR #2409, opened 2020-12-23
Ignore parking=street_side unless access=unknown

### LINKED ISSUE #2408: Parking access might omit parking=street_side
A new user added `access=yes` to a parking area mapped as `amenity=parking` with `parking=street_side` based on the parking access quest. This is not wrong of course, but for parking amenities tagged with `parking=street_side` the overwhelming majority is likely to be found at the sides of public access roads, and any conditions tagged on it are probably better done in bulk for all such parking areas in an area, along with any `parking:lane` tagging on the roads themselves. As the default access value already is `yes` for this kind of amenity, it is a bit redundant.

I would suggest removing `parking=street_side` from the parking access quest. The access-question makes a lot of sense for `parking=surface`, but may waste user's efforts for `parking=street_side`.

What do you think?

jdhoek:
Fixes #2408.

This does allow `parking=street_side` when `access=unknown`, which seems sensible.

-----

I am a Java developer, but I don't work with Kotlin, and don't have a smartphone I can test this on. I have only amended the `elementFilter`, and tested the filter string with JOSM on Overpass. It seems to work, but please review for formatting and correctness.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddParkingAccess.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddParkingAccess.kt
@@ -6,7 +6,22 @@ import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
 
 class AddParkingAccess : OsmFilterQuestType<String>() {
 
-    override val elementFilter = "nodes, ways, relations with amenity=parking and (!access or access=unknown)"
+    // Exclude parking=street_side lacking any access tags, because most of
+    // these are found alongside public access roads, and likely will be
+    // access=yes by default. Leaving these in makes this quest repetitive and
+    // leads to users adding lots of redundant access=yes tags to satisfy the
+    // quest. parking=street_side with access=unknown seems like a valid target
+    // though.
+    //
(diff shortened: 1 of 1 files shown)


matkoniecz:
> don't have a smartphone I can test this on

You can run app in an Android Emulator on PC - see https://developer.android.com/studio/run/emulator

app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddParkingAccess.kt
```diff
@@ -6,7 +6,22 @@ import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
...
+    // leads to users adding lots of redundant access=yes tags to satisfy the
+    // quest. parking=street_side with access=unknown seems like a valid target
+    // though.
+    //
+    // Cf. #2408: Parking access might omit parking=street_side
+    override val elementFilter = """
+        nodes, ways, relations with amenity=parking
+        and (
+            access=unknown
+            or (!access and -parking=street_side)
```
westnordost, line 21:
```suggestion
            or (!access and parking != street_side)
```

westnordost:
Thank you!

PR #2417, opened 2020-12-26
exclude addresses with implied addr:street from addr:street quest

### LINKED ISSUE #2112: "What Street" quest should not be asked for house numbers that belong to an interpolation which already defines the street
Currently, streetComplete unnecessarily asks for a street name in cases where the house number belongs to an interpolation that already defines the street.

Example:

- Node [3035882703 ](https://www.openstreetmap.org/node/3035882703)defines addr:housenumber=96
- It is a member of way [350655164](https://www.openstreetmap.org/node/3035882703)
- way 350655164 defines addr:street=The Sunny Road

Asking this quest in the above circumstance wastes questor's time asking for information which is already known. It is also data inefficient, because answering the quest causes the OSM database to contain multiple unnecessary duplication of information that is more efficiently stored in the interpolation. Questor's time would be better spent answering other quests and the unnecessary duplication would be eliminated.

**Proposed Solution**

Before asking this quest, StreetComplete should check if the house number belongs to an interpolation, and should eliminate it if the interpolation contains an addr:street.

matkoniecz:
fixes #2112

Let me know if it is something that could be commited directly, without opening PR. (I suspect so but was unsure)


westnordost (changes requested):
Looks good!

However, there are tests for the other cases, so for this case there should also be a test case in `AddAddressStreetTest`

matkoniecz:
tests added

westnordost (approved):
👍

westnordost:
You reduced the issue count to below 70!

matkoniecz:
I just wanted to post about it, though it is rather 80 :)

PR #2425, opened 2020-12-27
mention use iD presets as relevant to translations

matkoniecz:
triggered by https://github.com/streetcomplete/StreetComplete/discussions/2416#discussion-227909

I am really unhappy about clumsy wording here


westnordost (commented):
meh, it's ok

PR #2443, opened 2021-01-01
Set BottomSheetBehaviour to expanded in note creation to fix keyboard overlapping input area

### LINKED ISSUE #2442: Note creation dialog gets covered by keyboard
When creating a note the input field gets covered by the keyboard instead of shifting up as other input fields do:

image

**How to Reproduce**

1. Long press on the map
2. Select "Create new note"
3. Tap text area

**Versions affected**

Android: 11 (RQ1A.201205.003)
StreetComplete: v29.0-beta1

Elleo:
Fixes #2442

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
@@ -30,9 +30,7 @@ abstract class AbstractCreateNoteFragment : AbstractBottomSheetFragment() {
         val view = inflater.inflate(layoutResId, container, false)
 
         val bottomSheet = view.findViewById<LinearLayout>(R.id.bottomSheet)
-        if (resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE) {
-            BottomSheetBehavior.from(bottomSheet).state = BottomSheetBehavior.STATE_EXPANDED
-        }
+        BottomSheetBehavior.from(bottomSheet).state = BottomSheetBehavior.STATE_EXPANDED
 
         val content = view.findViewById<ViewGroup>(R.id.content)
         content.removeAllViews()


westnordost:
Hmm, weird that. So, expanding the layout fixes it for you?

In `AbstractBottomSheetFragment` (super class of the `AbstractCreateNoteFragment` class), there is code that should automatically expand the bottom sheet when the keyboard is opened. Does this not work anymore on Android 11?

https://github.com/streetcomplete/StreetComplete/blob/8083c1575e1e865c35ae8f76941849be37ddc3fa/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractBottomSheetFragment.kt#L42-L55

westnordost:
Right, I can reproduce it with an Android 11 emulator. Since there is no API method to find out whether the keyboard has been opened/closed, I had to write the above method that registers if the view bottom changed by a certain degree. Possibly because of the smooth keyboard pop up animation, this hack does not work anymore.

Elleo:
Thanks for taking a look at this. Here's an extra data point:

The bottom of the create note dialog displays below the top of the android bottom bar overlay:

Screenshot_20210102-143716

Manually dragging the dialog up before tapping the input area causes it to behave correctly when the keyboard opens.

Expanding the bottom sheet causes it to display correctly above the android system overlay from the start, but I'm not sure if this is the correct way to fix this (this is my first interaction with the codebase)

Thanks,
Mike

westnordost:
I tried to fix this. Now on my Android 11 emulator and Android 10 phone, I can't reproduce this anymore. Would you like to try if it is fixed for you as well? I committed it to master.

Elleo:
Just gave it a quick test and that seems to fix it for me too, thanks :slightly_smiling_face:

westnordost:
Cool! Thank you too for the report and even the initiative to try to fix it yourself!

Issue #2454, opened 2021-01-03
Feature Request: crossing island quests should show up for crossings with connected footways


dbdean:
**Use case**

As I have mapped most of the sidewalks in Armidale, NSW, Australia (but not in enough detail to map most islands as road separations), I don't get any crossing island quests because of the decision to not show them for `highway`=`crossing` when there is an attached `highway`=`path|footway|cycleway|pedestrian`. I understand this is to avoid getting spammed with highway island quests (as discussed in https://github.com/streetcomplete/StreetComplete/pull/2030) when the islands are mapped in detail, but I think there might be a better way.

**Proposed Solution**

I believe that the connection of a footpath to a crossing shouldn't matter. The characteristics of the road should be what determine if there could be a crossing. For example, see the following section around https://www.openstreetmap.org/#map=19/-30.51435/151.66405 in Armidale, NSW, Australia:

![image](https://user-images.githubusercontent.com/12725310/103490597-11182c80-4e71-11eb-9cb9-1ab7a4764f26.png)

I think that the crossing quest should show up for the green crossings, but not the blue crossings, and the distinguishing characteristic is that Rusden St (and the small service lane to the right) are single-lane one-way roads.

I propose that the quest is changed to show on all `highway`=`crossing` nodes where the road is not one-way. If the road is one-way, I could also see a potential for including it only if it had 2 or more lanes.

dbdean:
I would be happy to have a go at implementing this myself and testing it out in a PR if the general principle seems ok.

westnordost:
The quest was done by @kmpoppe , the original PR is here #2030

Maybe somewhere in the discussion the reasoning behind this filter can be found. I may have a look at it tomorrow.

peternewman:
> I propose that the quest is changed to show on all `highway`=`crossing` nodes where the road is not one-way. If the road is one-way, I could also see a potential for including it only if it had 2 or more lanes.

Two lanes of one-way traffic will still get some false positives at junctions, see for example:
https://www.openstreetmap.org/node/256794584

Probably also at things like motorway/large A road junctions.

Allowing it for three would probably reduce that further, although generally I'd imagine the number of crossings missed by just ignoring for all one-way roads crossing crossings would be low enough to be acceptable and certainly fewer misses than the current situation.

One way roads was mentioned by @matkoniecz in the issue and a query found:
> Maybe skipping crossings on oneway roads would be way to get rid of them (and also of some valid quests)?

[...]

PR #2455, opened 2021-01-04
Crossing island quest with footways

dbdean:
WIP on allowing crossing:island quests when footways/cycleways etc are mapped separately. 

See https://github.com/streetcomplete/StreetComplete/issues/2454 for issue.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
@@ -23,6 +23,7 @@ class AddCrossingIsland : OsmElementQuestType<Boolean> {
     private val excludedWaysFilter by lazy { """
         ways with 
           highway and access ~ private|no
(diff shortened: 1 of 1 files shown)


westnordost:
I read up on the issue and I can't think of an example now where the road is **not** oneway but a footway that crosses the road should not have a `crossing:island=yes/no`. Or did I miss something?

So from my point of view, this could be merged. @kmpoppe @matkoniecz any comments?

Why is this WIP?

dbdean:
I hadn't tested it a the time of creating the PR. Now that I have run it on my phone and done a bit of mapping, I'm happy with the way it works now. I've removed the WIP label and am happy for this to be merged if @kmpoppe and/or @matkoniecz have no objections.

matkoniecz:
Maybe it was intended to exclude small roads?

In any case, I am not opposed to such change in oneway filtering.

matkoniecz:
I have not tested this PR but I see no problems here.

westnordost:
Okay, let's merge it then, thank you!

PR #2469, opened 2021-01-08
Simplify the sending of emails

tapetis:
This pull request

- adds a new utility function to send emails, thus unifying the previous two separate implementations for it,
- removes the therefore no longer necessary `SENDTO` `intent` element in the `AndroidManifest`'s `queries` list, and
- removes duplicate `tryStartActivity()` implementations.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -26,11 +26,6 @@
             <action android:name="android.intent.action.VIEW"/>
             <data android:scheme="geo"/>
         </intent>
-        <!-- send crash report mail -->
-        <intent>
-            <action android:name="android.intent.action.SENDTO"/>
-            <data android:scheme="mailto"/>
-        </intent>
     </queries>
 
     <application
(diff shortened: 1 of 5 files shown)


westnordost:
> removes the therefore no longer necessary SENDTO intent element in the AndroidManifest's queries list, and

Why it is no longer necessary?

tapetis:
Starting an intent and letting the operating system decide who handles it is still possible on Android 11 (API 30) without an entry in the `queries` list. Just like with older Android versions,  an `ActivityNotFoundException` is thrown when the intent cannot be opened in another app. [The privacy changes in Android 11 just prevent](https://developer.android.com/about/versions/11/privacy/package-visibility) iterating over all activities that support a particular intent. One method that previously allowed this without a `queries` list is the `Intent.resolveActivity()` function, which is no longer used in this pull request.

https://github.com/streetcomplete/StreetComplete/blob/5df2040f07db3feb23eee4c2bb78ed40885f4255/app/src/main/java/de/westnordost/streetcomplete/util/CrashReportExceptionHandler.kt#L124

westnordost:
Ah, okay

westnordost:
The PR looks good to me! Did you test your change?

tapetis:
I tested both ways to send emails on an Android 11 emulator. I have also disabled the built-in email app to see if the toast works.

westnordost:
Nice, thank you!

PR #2475, opened 2021-01-09
Enable pitch surface quest for `leisure=track`

### LINKED ISSUE #2468: Enable pitch surface quest also for `leisure=track`
### General
Affected tag(s) to be modified/added: [surface](https://wiki.openstreetmap.org/wiki/Key:surface)
Question asked: **What surface does this sport track have?**

### Checklist
- [X] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [X] 💤 Not an overwhelming percentage of elements have the same answer (No spam):
      Most will probably have `surface=tartan`, but there are enough that have a different surface: https://overpass-turbo.eu/s/1294
- [X] 🕓 Applies to a reasonable number of elements (Worth the effort):
      Taginfo shows that only 17.35% of 87 274 objects with `leisure=track` have a `surface` tag: https://taginfo.openstreetmap.org/tags/leisure=track#combinations

### Ideas for implementation

**Element selection:**
`leisure=track and !surface and (!access or access!=private)` (+ maybe `and sport`)

**Proposed GUI:**
Same as "What surface does this sport field have?" quest (#1170 / #2377).

FloEdelmann:
Closes #2468.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
@@ -43,10 +43,10 @@ class AddPitchSurface : OsmFilterQuestType<SurfaceAnswer>() {
     override val icon = R.drawable.ic_quest_pitch_surface
 
     override fun getTitle(tags: Map<String, String>) =
-        if (tags.get("leisure") == "pitch")
-            R.string.quest_pitchSurface_title
-        else
+        if (tags.get("leisure") == "track")
             R.string.quest_pitchSurface_title_track
+        else
+            R.string.quest_pitchSurface_title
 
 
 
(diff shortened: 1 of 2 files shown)


matkoniecz:
Thanks! From looking at code everything seems fine to me, though I have not actually tested the code.

FloEdelmann:
I have tested it on the emulator. How can I see the OSM changeset I made there? (Even though it's not actually uploaded to OSM.)

matkoniecz:
You can look in a log, it mentions applied tags there.

(let me know if you need help in finding it)

FloEdelmann:
Thanks, I see a message which looks fine :)

> 21-01-09 19:19:55.168 7678-7678/de.westnordost.streetcomplete.debug D/QuestController: Solved a AddPitchSurface quest: ADD "surface"="tartan"

westnordost (approved):

Issue #1892, opened 2020-06-11
iOS version


westnordost:
### Edit: see [this comment](https://github.com/streetcomplete/StreetComplete/issues/1892#issuecomment-1671074967) for the current state of affairs

---

Quite a few times, it has been requested that an iOS version of this app is made available. This ticket shall serve as an explanation why this isn't possible but also as an entry point for why it is perhaps possible after all, but with enormous effort. So if you are an iOS developer and interested in creating an iOS version, also read this.

## It's not possible
StreetComplete is developed as a native Android app and iOS is a completely different platform. There is no compatibility at all. Usually, if you see an app that is both available for Android and iOS, it is either not native (for example, just a website being displayed within an app or using a cross-platform framework such as ReactNative) or you are looking at two completely different applications, likely developed by two teams. This is my experience as a professional app developer.

## Maybe it's possible
So, this is interesting if you are an iOS developer yourself. It is not planned from my side to do this.

### Kotlin Multiplatform Mobile
This app is written in Kotlin, a modern programming language that is quite similar to Swift. Originally, Kotlin was built to run on the JVM, but by now, it can be transpiled into JavaScript, and with Kotlin Native, also to machine code, just like Swift.
This makes it possible to write iOS apps with Kotlin. Thus, Kotlin code of this project that is not dependent on any Java library (Http library etc) and also not dependent on the Android platform (UI, sensors, persistance management etc) could be used 1:1 in an iOS port of this app.
Such "pure" Kotlin code would be put into a common module that is shared between the iOS and the Android app.
However, not a big percentage of the total code in this project is such pure Kotlin code that has no dependencies to any other Java or Android dependencies. Most of the iOS app would still need to be written from scratch, in Swift or better in Kotlin, if possible. On the other hand, some of the code that is currently dependent on some Java library or Android dependency could be changed to not be dependent on it directly anymore but on a wrapper interface. Then, only that wrapper would need to also be implemented for iOS.
<strike>
### Multi-OS Engine
The MOE is a little different approach. So, the JVM of Android is called ART and is open source. On Android, we write apps that run within this virtual machine provided by the Android system. For iOS, there is no such virtual machine, apps run natively. But, what we could do, is to write an app that comes bundled with the ART virtual machine and does nothing else than to load the provided JVM bytecode. This is the approach of MOE.
[...]

PR #2493, opened 2021-01-11
Convert several Java files to Kotlin

FloEdelmann:
I hope this kind of PR is fine for you. I followed these steps for each of the commits:

1. used Android Studio's "Convert Java File to Kotlin File" feature
2. manually formatted the code
	* mostly restored empty lines and comment formatting
	* sometimes changed it to more idiomatic Kotlin
3. fixed compile errors in other files, if any

Please tell me if I missed any obvious "idiomatic Kotlin" opportunities, as I'm new to Kotlin.


westnordost:
Yes, some of it is ok to change to Kotlin. Prefs is, ApplicationConstants is.

However, `KryoSerializer` is too critical that I want to touch it at all. For example, in the Java version, there was the `java.util.HashMap` at the first place in the `registeredClasses` array. In the Kotlin version, it is not clear if it is still `java.util.HashMap` or `kotlin.collections.HashMap`. If that makes any difference. In any case, I'd not like to take the risk that something could break there. Maybe one day I'll convert it, but at my own direction then.

The reason why  `CountryInfos.java` and `Abbreviations` has not been converted yet is because I am playing with the thought or at least keep open the option to put each into a small separate library. For this reason, it should remain as .java.

westnordost:
So if you revert the changes not related to these two files, I can merge it.

westnordost:
If you feel like helping out by doing some small refactorings, there is something else that could be done - decoupling the data-side of quests (more) from UI and Android:

### Task
For the cycleway quest, the cycleway answer options are cleanly seperated into two files: `Cycleway.kt` contains the "pure data" (the enum) and `CyclewayItem` contains the view-related part of a cycleway answer option.

This is better than how it is done for for example the surface quest (and others). There, in `Surface.kt`, the enum is not pure data because is contains reference to view stuff and also Android-specific stuff. So for these quest(s), the data class should be refactored to be like for the cycleway class. 

### Example

```kotlin
enum class Surface(val value: String) {
  ASPHALT("asphalt"),
  ...
}

// in SurfaceItem.kt
fun Surface.asItem(): Item<Surface> = ...
```

### Reason
Apart from it being good style, it is a step towards making #1892 a little less work to do

westnordost:
(If you are up for it, I can create a ticket for that so it can be properly referred to)

FloEdelmann:
I'm not sure how much time that will take for a Kotlin novice like me, but I will try :)

westnordost (approved):

westnordost:
👍 thanks, merged!

FloEdelmann:
Are any of the remaining Java files eligible to be converted to Kotlin?

### Critical → should not be converted

* [`KryoSerializer.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/util/KryoSerializer.java)

### Could be extracted into a library → should not be converted

* [`CountryInfo.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.java)
* [`CountryInfos.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.java)
* [`CountryInfosTest.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/androidTest/java/de/westnordost/streetcomplete/data/meta/CountryInfosTest.java)
* [`Abbreviations.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/data/meta/Abbreviations.java)
* [`AbbreviationsByLocale.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/data/meta/AbbreviationsByLocale.java)

### Dagger 2 can't find these as Kotlin files → should not be converted

* [`DownloadModule2.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/data/download/DownloadModule2.java)
* [`UploadModule2.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/data/upload/UploadModule2.java)

### Remaining → :question: 

* [`FlattenIterable.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/util/FlattenIterable.java)
* [`MultiIterable.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/util/MultiIterable.java)
* [`Serializer.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/util/Serializer.java)
* [`LocationUtil.java`](https://github.com/streetcomplete/StreetComplete/blob/v29.1/app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.java)

westnordost:
Serializer is closely related to KryoSerializer... and it is just 2 lines of code
For FlattenIterable and MultiIterable.... are they actually used? Maybe they can be converted. Kotlin is a little stricter regarding usage of generics (of generics).
LocationUtil... there was some other open PR I think, that touched that file, which is why I thought I'd rather postpone converting that.

FloEdelmann:
* `Serializer` is closely related to KryoSerializer → fair enough
* `FlattenIterable` is not used at all (but there is a test for it) → should it be removed?
* `MultiIterable` is used in three places, so I'll try to convert it to Kotlin and pay attention to generics
* `LocationUtil` is indeed touched by #2031 and #2082.
  But so was `MainActivity.java`, which you already converted to Kotlin :sweat_smile: 
  @Isira-Seneviratne Would it be okay for you to resolve another merge conflict in those PRs when `LocationUtil` is converted to Kotlin?

westnordost:
> MultiIterable.java is used in three places, so I'll try to convert it to Kotlin and pay attention to generics

Maybe there is a more kotlin-y way to do that. Something like 
`for(item in (collection1.asSequence() + ollection2.asSequence() + ...))`
maybe?

Issue #2494, opened 2021-01-12
Separate view data in quest data classes


westnordost:
### Task
For the cycleway quest, the cycleway answer options are cleanly seperated into two files: [`Cycleway.kt`](https://github.com/streetcomplete/StreetComplete/blob/cc7624d2f8c83c037f4a078eb56dadd3b1cd80a2/app/src/main/java/de/westnordost/streetcomplete/quests/bikeway/Cycleway.kt) contains the "pure data" (the enum) and [`CyclewayItem`](https://github.com/streetcomplete/StreetComplete/blob/cc7624d2f8c83c037f4a078eb56dadd3b1cd80a2/app/src/main/java/de/westnordost/streetcomplete/quests/bikeway/CyclewayItem.kt) contains the view-related part of a cycleway answer option.

This is better than how it is done for for example the surface quest (and others). There, in [`Surface.kt`](https://github.com/streetcomplete/StreetComplete/blob/25d0984bcedc5d4daf617486f2c71d22dbc03633/app/src/main/java/de/westnordost/streetcomplete/quests/surface/Surface.kt), the enum is not pure data because it contains reference to Android-specific view stuff. So for these quest(s), the data class should be refactored to be like for the cycleway class. 

### Example

So it should be....
```kotlin
enum class Surface(val value: String) { // or osmValue?
  ASPHALT("asphalt"),
  // etc....
}

// in SurfaceItem.kt
fun Surface.asItem(): Item<Surface> {
   // ... well, see CyclewayItem.kt, ...
}
```

### Reason
Apart from it being good style generally to seperate UI from data logic, it is a step towards making #1892 or any other port from Android to something else a little less work to do.

### Quests where this "old style" is used
- BuildingType
- Surface

### And then
And there is more. Quest types should no longer have a `String` as parameter but an enum. 

[...]

Issue #2028, opened 2020-08-16
Android Unit Tests broken


westnordost:
Unit tests of any Android project are split into two sets. Those that can be executed on the local machine because they do not depend on the Android environment (=test) and those that are executed within the Android environment because they depend on Android APIs in some way (=androidTest).

Mockito is a mocking library used for convenient testing.

Some android unit tests fail since some time with (mostly) this error message:
`java.lang.ClassNotFoundException: Didn't find class "org.objenesis.ObjenesisStd" on path: DexPathList`

I know as much as Objenesis is related to Mockito. So this seems to be a problem with Android Unit tests + Mockito + maybe! Proguard/R8, unfortunately I haven't been able to get to the bottom of this, any help is highly appreciated.

westnordost:
I am closing this in favour of #2521. In the long term, that ticket makes more sense than dabbling with android unit test weirdness.

PR #2495, opened 2021-01-12
Separate view data for Surface and BuildingType

FloEdelmann:
Fixes first part of #2494.


app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceForm.kt
```diff
@@ -5,19 +5,15 @@ import de.westnordost.streetcomplete.R
...
+    override val items: List<Item<Surface>>
+        get() = (PAVED_SURFACES + UNPAVED_SURFACES + GROUND_SURFACES + GENERIC_SURFACES).toItems()
 
     override val itemsPerRow = 3
 
-    override fun onClickOk(selectedItems: List<String>) {
+    override fun onClickOk(selectedItems: List<Surface>) {
         val value = selectedItems.single()
-        if (value == "paved" || value == "unpaved") {
+        if (value.shouldDescribe) {
```
westnordost, line 16: Hmm, not sure if this is an improvement. I'd rather have the check here or maybe rename `shouldDescribe` to `isGeneric`?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/Surface.kt
```diff
@@ -42,3 +41,18 @@ val UNPAVED_SURFACES = listOf(
...
+    CLAY, ARTIFICIAL_TURF, TARTAN, DIRT,
+    FINE_GRAVEL, PAVING_STONES, COMPACTED,
+    SETT, UNHEWN_COBBLESTONE, GRASS_PAVER,
+    WOOD, METAL, GRAVEL, PEBBLES,
+    ROCK
+)
+
+val GENERIC_SURFACES = listOf(
+    PAVED, UNPAVED, GROUND
+)
```
westnordost, line 56: Good idea to have these lists not defined in the view code (the form) 👍

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
```diff
@@ -0,0 +1,76 @@
+package de.westnordost.streetcomplete.quests.building_type
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.view.image_select.Item
+import de.westnordost.streetcomplete.quests.building_type.BuildingType.*
+
+fun List<BuildingType>.toItems() = this.map { it.asItem() }
+
+fun BuildingType.asItem(): Item<String> = when(this) {
```
westnordost, line 9: Here, it should be 

`fun BuildingType.asItem(): Item<BuildingType> = when(this) {`

not string

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
```diff
@@ -0,0 +1,76 @@
+package de.westnordost.streetcomplete.quests.building_type
+
+enum class BuildingType(val osmTag: String?, val subTypes: List<BuildingType>? = null) {
```
westnordost, line 3: Interesting idea, to include both key and value information in here. I think this makes sense. However, then it would make more sense to have two fields for that: `osmKey` and `osmValue`

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
```diff
@@ -0,0 +1,76 @@
...
+    ABANDONED     ("abandoned=yes"),
+    RUINS         ("ruins=yes"),
+
+    RESIDENTIAL   ("building=residential", listOf(DETACHED, APARTMENTS, SEMI_DETACHED, TERRACE, HOUSE, FARM, HUT, BUNGALOW, HOUSEBOAT, STATIC_CARAVAN, DORMITORY)),
+    COMMERCIAL    ("building=commercial", listOf(OFFICE, INDUSTRIAL, RETAIL, WAREHOUSE, KIOSK, HOTEL, STORAGE_TANK)),
+    CIVIC         ("building=civic", listOf(SCHOOL, UNIVERSITY, HOSPITAL, KINDERGARTEN, SPORTS_CENTRE, TRAIN_STATION, TRANSPORTATION, COLLEGE, GOVERNMENT, STADIUM)),
+    RELIGIOUS     ("building=religious", listOf(CHURCH, CATHEDRAL, CHAPEL, MOSQUE, TEMPLE, PAGODA, SYNAGOGUE, SHRINE)),
+    FOR_CARS      (null, listOf(GARAGE, GARAGES, CARPORT, PARKING)),
+    FOR_FARMS     (null, listOf(FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK)),
+    OTHER         (null, listOf(SHED, ROOF, SERVICE, HUT, TOILETS, HANGAR, BUNKER, HISTORIC, ABANDONED, RUINS));
```
westnordost, line 69: This will probably not work like that unless `fun BuildingType.asItem()` works recursively.

westnordost (changes requested):
Wow, you are fast. nice work! I got a few comments though.

Also, did you check if you need to change any tests as well? Try executing the tests to see if they still compile

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
westnordost: FOR_CARS, FOR_FARMS etc. are rather an UI category, so question is whether they should be part of the data at all or not just part of the UI data

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceForm.kt
FloEdelmann: This check is duplicated across all surface quest forms, that's why I extracted it.

First, I just had `GENERIC_SURFACES.contains(value)`, but then I saw [`ac55db8` (#2078)](https://github.com/streetcomplete/StreetComplete/pull/2078/commits/ac55db8e5f1013ec9a5b986a6758fba8baa12202), which explicitly excludes `ground`.

I think `isGeneric` sounds a bit like a shorthand for `GENERIC_SURFACES.contains(value)`.

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
FloEdelmann: When it returns a `Item<BuildingType>`, the logic in [`AGroupedImageListQuestAnswerFragment.kt`](https://github.com/streetcomplete/StreetComplete/blob/9820af988819cae4e0ee33a933c5b53cfc29919f/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestAnswerFragment.kt#L99-L128) (don't accept a `null` answer) won't work anymore, because the answer is then a non-null Item instead of a null String:

https://github.com/streetcomplete/StreetComplete/blob/9820af988819cae4e0ee33a933c5b53cfc29919f/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestAnswerFragment.kt#L99-L128

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
FloEdelmann: Indeed, that's a regression in [`644ee41` (#2495)](https://github.com/streetcomplete/StreetComplete/pull/2495/commits/644ee41b196beb7eebd2bc444a7d4e3e105a496c). I'll see what I can do instead.

FloEdelmann:
Running "de.westnordost.streetcomplete (test)" works, all the tests pass there.

However, running "de.westnordost.streetcomplete (androidTest)" yields this error in a lot of (unrelated) files:

```
java.lang.IllegalStateException: Could not initialize plugin: interface org.mockito.plugins.InstantiatorProvider2 (alternate: interface org.mockito.plugins.InstantiatorProvider)
	at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:74)
	at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
	at $Proxy13.getInstantiator(Unknown Source)
	at org.mockito.internal.creation.bytebuddy.SubclassByteBuddyMockMaker.createMock(SubclassByteBuddyMockMaker.java:44)
	at org.mockito.android.internal.creation.AndroidByteBuddyMockMaker.createMock(AndroidByteBuddyMockMaker.java:39)
	at org.mockito.internal.util.MockUtil.createMock(MockUtil.java:35)
	at org.mockito.internal.MockitoCore.mock(MockitoCore.java:62)
	at org.mockito.Mockito.mock(Mockito.java:1908)
	at org.mockito.Mockito.mock(Mockito.java:1817)
	at de.westnordost.streetcomplete.data.notifications.NewUserAchievementsDaoTest.addPop(NewUserAchievementsDaoTest.kt:34)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at androidx.test.internal.runner.junit4.statement.RunBefores.evaluate(RunBefores.java:80)
	at androidx.test.internal.runner.junit4.statement.RunAfters.evaluate(RunAfters.java:61)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:56)
	at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:395)
	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2205)
Caused by: java.lang.NoClassDefFoundError: Failed resolution of: Lorg/objenesis/ObjenesisStd;
	at org.mockito.internal.creation.instance.ObjenesisInstantiator.<init>(ObjenesisInstantiator.java:16)
	at org.mockito.internal.creation.instance.DefaultInstantiatorProvider.<clinit>(DefaultInstantiatorProvider.java:13)
	at java.lang.Class.classForName(Native Method)
	at java.lang.Class.forName(Class.java:454)
	at java.lang.Class.forName(Class.java:379)
	at org.mockito.internal.configuration.plugins.DefaultMockitoPlugins.create(DefaultMockitoPlugins.java:68)
	at org.mockito.internal.configuration.plugins.DefaultMockitoPlugins.getDefaultPlugin(DefaultMockitoPlugins.java:45)
	at org.mockito.internal.configuration.plugins.PluginLoader.loadPlugin(PluginLoader.java:67)
	at org.mockito.internal.configuration.plugins.PluginRegistry.<init>(PluginRegistry.java:36)
	at org.mockito.internal.configuration.plugins.Plugins.<clinit>(Plugins.java:19)
	at org.mockito.internal.configuration.plugins.Plugins.getMockMaker(Plugins.java:35)
	at org.mockito.internal.util.MockUtil.<clinit>(MockUtil.java:24)
	at org.mockito.internal.util.MockUtil.typeMockabilityOf(MockUtil.java:29)
	at org.mockito.internal.util.MockCreationValidator.validateType(MockCreationValidator.java:22)
	at org.mockito.internal.creation.MockSettingsImpl.validatedSettings(MockSettingsImpl.java:240)
	at org.mockito.internal.creation.MockSettingsImpl.build(MockSettingsImpl.java:228)
	at org.mockito.internal.MockitoCore.mock(MockitoCore.java:61)
	... 32 more
Caused by: java.lang.ClassNotFoundException: Didn't find class "org.objenesis.ObjenesisStd" on path: DexPathList[[zip file "/system/framework/android.test.runner.jar", zip file "/system/framework/android.test.mock.jar", zip file "/system/framework/android.test.base.jar", zip file "/data/app/~~SdYUqSRd782jfc3ZT49njQ==/de.westnordost.streetcomplete.debug.test-zUOpAoBsp-0jMpZh-Dnsrg==/base.apk", zip file "/data/app/~~HR57ElSadBtTbGZlCGUSLg==/de.westnordost.streetcomplete.debug-3kwzh2RqKnLZLPYm6y-xyw==/base.apk"],nativeLibraryDirectories=[/data/app/~~SdYUqSRd782jfc3ZT49njQ==/de.westnordost.streetcomplete.debug.test-zUOpAoBsp-0jMpZh-Dnsrg==/lib/x86, /data/app/~~HR57ElSadBtTbGZlCGUSLg==/de.westnordost.streetcomplete.debug-3kwzh2RqKnLZLPYm6y-xyw==/lib/x86, /data/app/~~SdYUqSRd782jfc3ZT49njQ==/de.westnordost.streetcomplete.debug.test-zUOpAoBsp-0jMpZh-Dnsrg==/base.apk!/lib/x86, /data/app/~~HR57ElSadBtTbGZlCGUSLg==/de.westnordost.streetcomplete.debug-3kwzh2RqKnLZLPYm6y-xyw==/base.apk!/lib/x86, /system/lib, /system_ext/lib]]
	at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:207)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
	... 49 more
```

It seems like not all dependencies were installed correctly, right?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceForm.kt
westnordost: Hm thats true. But maybe then `shouldBeDescribed`?

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
westnordost: Except if `FOR_CARS` would not actually be a BuildingType but just a categorization for the UI. Which is kinda true.

westnordost:
Yes, I assume it is not due to your changes. See #2028 , don't know what is the cause

FloEdelmann:
I think I found a nice solution to work around the issues I commented above :)

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
```diff
@@ -0,0 +1,121 @@
...
+    BuildingType.SHED, BuildingType.HUT
+)
+
+enum class BuildingTypeCategory(val type: BuildingType?, val subTypes: List<BuildingType>) {
+    RESIDENTIAL(BuildingType.RESIDENTIAL, listOf(
+        BuildingType.DETACHED, BuildingType.APARTMENTS, BuildingType.SEMI_DETACHED,
+        BuildingType.TERRACE, BuildingType.HOUSE, BuildingType.FARM, BuildingType.HUT,
+        BuildingType.BUNGALOW, BuildingType.HOUSEBOAT, BuildingType.STATIC_CARAVAN,
+        BuildingType.DORMITORY
+    )),
```
westnordost, line 88: Interesting solution, nice work!

Maybe it would improve readability if you could cut out the `BuildingType.*` in this file.

It should work if you 
```kotlin
import de.westnordost.streetcomplete.quests.building_type.BuildingType.*
import de.westnordost.streetcomplete.quests.building_type.BuildingTypeCategory.*
```

smichel17: I don't think that will work — `BuildingType` is already declared in the same file.

However, I think a scope function should work:

```kotlin
    RESIDENTIAL(BuildingType.RESIDENTIAL, with(BuildingType) { listOf(
        DETACHED, APARTMENTS, SEMI_DETACHED, TERRACE, HOUSE, FARM, HUT,
        BUNGALOW, HOUSEBOAT, STATIC_CARAVAN, DORMITORY
    )}),
```

Didn't test this; don't take my word that it will work.

smichel17: Personally I would prefer this even if BuildingType were declared in a different package — as someone who seldom uses an IDE, wildcard imports make it a real pain to figure out where something comes from.

FloEdelmann: @smichel17 Unfortunately, your suggestion doesn't work:

```
…/quests/building_type/BuildingType.kt: (84, 9): Unresolved reference: DETACHED
```

westnordost:
Nice! ready to merge?

FloEdelmann:
No objections from my side :)

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
smichel17: Ah, that's sad. …but the import trick did work! I'm surprised — didn't realize you could import a package from within itself.

smichel17: For posterity, this would also work (tested this time), although I prefer the wildcard import given the definitions are contained in the same file
```kotlin
listOf("STATIC_CARAVAN", "DORMITORY").map(BuildingType::valueOf)
```

westnordost: > listOf("STATIC_CARAVAN", "DORMITORY").map(BuildingType::valueOf)

Goodby compile time safety, hello RuntimeException

smichel17: I call it a [compiler bug](https://www.staging-typescript.org/docs/handbook/literal-types.html#string-literal-types)

PR #2501, opened 2021-01-14
Extract quest data models

### LINKED ISSUE #2494: Separate view data in quest data classes

FloEdelmann:
Follow-up to #2495. Fixes second part and closes #2494.


app/src/main/java/de/westnordost/streetcomplete/quests/kerb_height/KerbHeight.kt
```diff
@@ -0,0 +1,9 @@
+package de.westnordost.streetcomplete.quests.kerb_height
+
+enum class KerbHeight(val osmValue: String) {
+    RAISED("raised"),
+    LOWERED("lowered"),
+    FLUSH("flush"),
+    KERB_RAMP("lowered"),
```
westnordost, line 7: Wait a moment, @matkoniecz , is this a bug? That both kerb ramp ans lowered have the osm value "lowered"?

app/src/main/java/de/westnordost/streetcomplete/quests/sport/Sport.kt
```diff
@@ -0,0 +1,47 @@
+package de.westnordost.streetcomplete.quests.sport
+
+enum class Sport(val osmValue: String) {
+    // sorted by ~worldwide usages, minus country specific ones
+    MULTI("multi"),
+
+    // 250k - 10k
```
westnordost, line 7: the comments about the usage numbers should actually go to the quest form, because this is where the order matters. The order does not matter in an enum (read: should not)

westnordost:
🚀

app/src/main/java/de/westnordost/streetcomplete/quests/kerb_height/KerbHeight.kt
matkoniecz: Not a bug. It is a bit stupid to present fake distinction, but I have not found a better interface.

See https://wiki.openstreetmap.org/wiki/Key:kerb#Tagging 

It was less explicit in the proposal (that never went to vote), but present from start - see https://wiki.openstreetmap.org/wiki/File:FI-Raised-regular-lowered.png from https://wiki.openstreetmap.org/w/index.php?oldid=1207438

Tag to distinguish this would be nice but I will leave it to others. 

I opened https://wiki.openstreetmap.org/wiki/Talk:Key:kerb#Both_kerb_ramps_and_lowered_kerbs_are_tagged_in_the_same_way to increase chance of someone inventing tagging for this.

app/src/main/java/de/westnordost/streetcomplete/quests/sport/Sport.kt
westnordost: But I can just do that after the merge

westnordost:
So, all fine, I'll merge it. Thank you for your work!

app/src/main/java/de/westnordost/streetcomplete/quests/sport/Sport.kt
FloEdelmann: In the quest form, the list is sorted by country usage anyway, so order there shouldn't matter either.

westnordost: only the top few

Issue #2507, opened 2021-01-15
bike parking fee 


DerDings:
### General
Affected tag(s) to be modified/added: [fee](https://wiki.openstreetmap.org/wiki/Key:fee)
Question asked: **does it cost a fee to park a bike here?**

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
only if thoughtfully filtered
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation
 identically to the car parking fee quest

**Element selection:** 
`amenity=bicycle_parking and access!~no|private and (name or bicycle_parking~building|lockers|shed or capacity>=100) and !fee and !fee:conditional`

I know most bicycle parkings dont't cost any fee, so I did some research.
I queried all bicycle parkings somewhere in central europe (with a bbox that didn't make overpass-turbo complain about the amount of data. I didn't record that, it was somewhat the size of France+Germany) with a given `bicycle_parking` tag to a csv file. Then I did some computer-noob level excel analysis. I excluded all Items tagged with `access~no|private`. This is what I found out:

(0/grey: `!fee`; no/orange: `fee=no`; yes/blue: `fee and fee!=no`) data: © OpenStreetMap contributors, obtained 2020-11-30
1. named bicycle parkings are likely to cost fee 
![grafik](https://user-images.githubusercontent.com/42044179/104765158-bcc25600-5768-11eb-9d10-5d0f7d89a61a.png)
0: `!name`, 1: `name`

2. within the _unnamed_ bicycle parkings, `bicycle_parking~building|lockers|shed` have a notable share of items with a fee. 
![grafik](https://user-images.githubusercontent.com/42044179/104765195-ce0b6280-5768-11eb-92ad-6e65c3a81b24.png)

3. within the _unnamed_ bicycle parkings which are _not_ `bicycle_parking~building|lockers|shed`, the larger ones (103<capacity<153) are likely to cost a fee. 
![grafik](https://user-images.githubusercontent.com/42044179/104765234-df546f00-5768-11eb-9223-7de670257f53.png)
[...]

PR #2509, opened 2021-01-16
Add drinking water quest

### LINKED ISSUE #549: New quest: Is water drinkable? (drinking_water)
[drinking_water](https://wiki.openstreetmap.org/wiki/Key:drinking_water) seems to be promising for a quest.

* [ ] Easy to answer, i.e. everyone can easily solve it (from the outside)
* [x] Easy user interface. (quite easy, depends on how much details to be implement)
* [x] Has a useful purpose. (important information)
* [x] Established tag. (seems to be; ~50 000 times used)
* [x] Low unanswerable quests. (i.e. low false positives, as every quest can be answered; at least if there is a sign)
* [x] It can be determined, if the quest is needed to be asked. (i.e. there is drinking_water=no)

It should be asked for [the things mentioned here](https://wiki.openstreetmap.org/wiki/Key:drinking_water#Applies_to_.28among_other_things.29), maybe only exclude toilets (they are too common?) and some other tags, where it is not alwayss clear that there is water at all (i.e. rest_area/camp_site).

FloEdelmann:
Closes #549.


app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override val elementFilter = """
+        nodes with (
+            man_made = water_tap
+            or man_made = water_well
+            or amenity = toilets
+            or natural = spring
+        )
+        and access !~ private|no and indoor != yes
+        and !drinking_water and amenity != drinking_water
```
FloEdelmann, line 17: Maybe even exclude any nodes with `drinking_water:legal`?

```suggestion
        and !drinking_water and !drinking_water:legal and amenity != drinking_water
```

westnordost:
Nice initiative! Also, nice icon! 

But phew, of all the other quest suggestions you pick that one to implement! If I remember right from the linked ticket, the discussion around this quest ended in a limbo, with all that `drinking_water:legal` and stuff.
Could you summarize your position on this and how you would solve (/solved) the issues mentioned in that ticket?

FloEdelmann:
> Nice initiative! Also, nice icon!

Thanks :blush: 

> But phew, of all the other quest suggestions you pick that one to implement!

It seemed like the easiest for me to implement :sweat_smile: 

> Could you summarize your position on this and how you would solve (/solved) the issues mentioned in that ticket?

I tried to make the options clear through wording:

* If it's signed, it's easy to answer.
* If it's not signed, and easy to decide (like https://goo.gl/K2hp3W vs. https://goo.gl/PKuP4U), it's easy to answer, too.
* If it's not signed, and difficult to decide, users might tap on "Other answers" and choose "seems drinkable", or skip the quest (with or without a note).
* If it's signed, but clearly wrong (e.g. in restaurant toilets etc., or by applying local knowledge), none of the answers are applicable, so users will tap on "Other answers" → "Sign is wrong", and leave a note. Then it will be decided case by case.

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    override val commitMessage = "Add whether water is drinkable"
+    override val wikiLink = "Key:drinking_water"
+    override val icon = R.drawable.ic_quest_drinking_water
+    override val isDeleteElementEnabled = true
```
FloEdelmann, line 23: Not sure if this is necessary in this quest.

westnordost:
Okay, I am sorry, I thought I wrote something on the #549 ticket before, but I did not.

Let's do the following:
Throw the uncommon answers and/or answers that require explanation all into one bucket, into the "Can't say" bucket:
    - It's better to leave a note (with picture) than to delete the node if the node can't be found because as you say, it might be a faucet in a toilet or something like that
    - sign is wrong: Better have a note+photo here, cause this may be quite subjective
    - seems drinkable: This answer is too close to just simply "yes, but no sign" anyway. If the surveyor wants to express that he is not 100% sure, better let him leave a note

Regarding the answer options remaining, I am not completely happy about these. At least, the wording is too long. Also, it should avoid "yes" and "no" because there are not 2 but 4 answer options. Maybe:

"Potable, stated by a sign"
"Non-potable, stated by a sign"
"Potable, obvious from the context"
"Non-potable, obvious from the context"

Still a lot of text. Maybe (additionally) use pictures? 
image

Are really all answer options realistic and necessary? I.e. does none of the options require explanation (-> create note) and none of the options is fuzzy enough that users can not always be sure about it (-> create note)?

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
```diff
@@ -0,0 +1,36 @@
...
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
+
+class AddDrinkingWater : OsmFilterQuestType<DrinkingWater>() {
+
+    override val elementFilter = """
+        nodes with (
+            man_made = water_tap
+            or man_made = water_well
+            or amenity = toilets
```
westnordost, line 13: should remove that. Too common and it will be always the same answer that depends on the region. ("Yes, no sign" or "no, no sign")

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
westnordost: yes. If only `drinking_water:legal` is set but not `drinking_water`, it will have a good reason

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
```diff
@@ -0,0 +1,36 @@
...
+        and access !~ private|no and indoor != yes
+        and !drinking_water and amenity != drinking_water
+    """
+
+    override val commitMessage = "Add whether water is drinkable"
+    override val wikiLink = "Key:drinking_water"
+    override val icon = R.drawable.ic_quest_drinking_water
+    override val isDeleteElementEnabled = true
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_drinking_water_title
```
westnordost, line 25: I think it makes sense to have a different question for the different things to ask for. I.e. the user should know if (s)he is looking for a faucet, or a natural spring, or a water well.

However, we could try something different here.

We have the feature dictionary. If all of man_made = water_tap, man_made = water_well etc. are in the id presets, the question could also appear like this:

`Is the water drinkable here? (Natural spring)`

This helps to keep down the number of strings that must be translated.

westnordost: Whether a preset for a thing exists, can be checked by searching through the `presets.json` (not the brands one)

FloEdelmann: They are indeed all included in `presets.json`:

* [`man_made=water_tap`](https://github.com/openstreetmap/id-tagging-schema/blob/9fdecc983443ce75185c2923caae9f7703f2cae4/dist/presets.json#L15735)
* [`man_made=water_well`](https://github.com/openstreetmap/id-tagging-schema/blob/9fdecc983443ce75185c2923caae9f7703f2cae4/dist/presets.json#L15768)
* [`natural=spring`](https://github.com/openstreetmap/id-tagging-schema/blob/9fdecc983443ce75185c2923caae9f7703f2cae4/dist/presets.json#L16297)

How can I access those from the code?

westnordost: See for example `AddPlaceName.kt`, `getTitle` and `getTitleArgs`

westnordost: I think `arrayOf(featureName.value.toString())` should be fine too. In the worst case (-> somehow that preset was removed), the question would appear as `Is the water drinkable here? (null)`

FloEdelmann:
> Throw the uncommon answers and/or answers that require explanation all into one bucket, into the "Can't say" bucket

Okay, done.

> "Potable, stated by a sign"
> "Non-potable, stated by a sign"
> "Potable, obvious from the context"
> "Non-potable, obvious from the context"

Those are fine for me, too.

Regarding *potable* vs. *drinkable*: *Potable* seems to be a bit more technically correct ("safe to drink, free from poisons etc.") and also more widespread. However, *drinkable* is also not uncommon, seems to be used interchangeably in colloquial language, and maybe might make it easier for non-native speakers to understand (at least I've never heard *potable* before). On the other hand, the roman languages all seem to use a word similar to *potable*.

> Maybe (additionally) use pictures?

Those limit the possible text length even more (especially in translations). Also, is it clear from the accompanying description that the sign does not have to look exactly like that? There is a wide variety of "drinking water" / "no drinking water" signs, see https://www.ecosia.org/images?q=trinkwasser%20schild and https://www.ecosia.org/images?q=drinking%20water%20sign

> Are really all answer options realistic and necessary? I.e. does none of the options require explanation (-> create note) and none of the options is fuzzy enough that users can not always be sure about it (-> create note)?

There exist both "drinking water" and "no drinking water" signs, so the first two options are realistic and necessary. When there is a sign, the user can assume that it's correct, so no note is necessary.

Only in the case that the user has local knowledge that the sign is wrong, they might be tempted to choose one of the "obvious from the context" answers instead of leaving a note. I think that case is better handled by my original wording.

When there is no sign, in many cases it's obvious (e.g. [here](https://goo.gl/K2hp3W), [here](https://goo.gl/PKuP4U) or [here](http://cdn.onlyinyourstate.com/wp-content/uploads/2016/02/Ice_Cold_Spring_Water-01.jpg)), but of course there are also cases where it's not obvious (e.g. [here](https://s3.wasabisys.com/saudiawebdesigncompany/2017/09/garden-faucet-design-ideas-faucet-extension-lee-valley-tools-regarding-dimensions-957-x-1300.jpg) or [here](https://wv-oberdiessbach.ch/schwand%20aussen.JPG)). I think then, none of the options are fitting enough to choose them instead of leaving a note.

FloEdelmann:
What about this wording?

**Is the water potable here? (%s)**

* Potable, stated by a sign
* Non-potable, stated by a sign
* Clearly potable (no sign)
* Clearly non-potable (no sign)

westnordost:
I think we need some native speakers on this. @peternewman @smichel17 ?

smichel17:
- "Potable" is the correct term. In my *opinion* everyone *should* know it, but in reality I am not not sure everyone *will*. So, I would also include "safe to drink".
- I have no opinion on whether it is better to omit "potable" entirely or use a mix of both, like in the example below.
- I have no opinion on which parts use which; they are freely interchangeable.
- You may wish to include a note to translators that they can translate either way.
    - One option: always use potable in the source set and then provide an en_US, translation which uses "safe to drink"
- You could provide a graphic but I would be careful not to imply anything unintentional — when there are images, people tend to skip over the words.
    - In particular, I'm worried the faucet icon might be misinterpreted as "This question asks about the type of water source". Not such a big deal for the quest icon (unless you will have a "what type of water source is this" quest later, in which case, consider using this icon for that), but if you put images in the answer, perhaps a water fountain or emoji-drinking-from-a-cup would be a better option.
- (signed) vs (unsigned) is clearest, in my opinion.

> ### Is this water source safe to drink?
> - [ ] Potable (signed)
> - [ ] Potable (unsigned)
> - [ ] Not potable (signed)
> - [ ] Not potable (unsigned)

peternewman:
Potable may be used on some signs and I'm aware of it from camping etc, but as @smichel17 says, I wonder if it's too technical. I think drinkable/safe-to-drink may be more widely understood.

I like @smichel17 's suggestion, it seems to cover all the boxes, as long as you don't think potable means unsafe to drink (which I guess the picture(s) would help with)...

What's wrong with the horse trough? In the UK that would probably be fed by mains pressure water, so is probably safer than drinking out of a stream, yet I've done the latter but not the former... :laughing: Maybe not while the horse is there too, although with that one having a hose rather than a ball cock, you could pull it out and use it pretty cleanly.

It's also interesting looking at your signs that the German is clearly drinking water, where as the French go with potable.

smichel17:
> as long as you don't think potable means unsafe to drink

I was flip-flopping about which way it should be, but you're right, the way I put it is ambiguous; swapping would solve that problem (while retaining the technical word, so that anyone who knows it, knows exactly what they're mapping):

> ### Is this water source potable?
> - [ ] Safe to drink (signed)
> - [ ] Safe to drink (unsigned)
> - [ ] Not safe to drink (signed)
> - [ ] Not safe to drink (unsigned)

It is a bit lengthier, but I think short enough in English, and other translations can use that language's equivalent of potable if they need the space.

> I think drinkable/safe-to-drink may be more widely understood.

I think drinkable is too vague. Most people will read it as "safe to drink", but I can imagine some people reading it as "possible to drink", and answer that river water which tastes fine but is not safe (due to bacteria) is "drinkable".

westnordost:
What's a ball cock?

westnordost:
I think it is best to both mention "safe to drink" and "potable" in the form, so that it is clear that it is pretty synonymous. "Potable" is also shorter.

peternewman:
> I was flip-flopping about which way it should be, but you're right, the way I put it is ambiguous; swapping would solve that problem (while retaining the technical word, so that anyone who knows it, knows exactly what they're mapping):

That's perfect!

> > I think drinkable/safe-to-drink may be more widely understood.
> 
> I think drinkable is too vague. I think most people will read it as "safe to drink", but some may read it as "possible to drink", and answer that river water which tastes fine but is not safe (due to bacteria) is "drinkable", or something like that.
> "Safe to drink" is still a little ambiguous, because or varying tolerance to risk, but much less so.

Yeah fair point. I was more meaning in comparison to potable.

> What's a ball cock?

Nothing rude, just a float valve, used in traditional UK toilets too. Horse drinks water, water level drops, float valve opens, water trough refills. I can't see anything like that in that image, hence my confusion. This isn't a common UK design but it shows the operation:
https://www.jamessmithfencing.co.uk/product/concrete-water-trough/concrete-water-trough-75-gallon/

> > It's also interesting looking at your signs that the German is clearly drinking water, where as the French go with potable.
> 
> Where, on the wiki?

No, @FloEdelmann 's image searches:
https://www.ecosia.org/images?q=trinkwasser%20schild and https://www.ecosia.org/images?q=drinking%20water%20sign

> I think it is best to both mention "safe to drink" and "potable" in the form, so that it is clear that it is pretty synonymous. "Potable" is also shorter.

I think @smichel17 does that quite elegantly, or do you think it needs to say "potable (safe to drink)"?

FloEdelmann:
Thanks for your input!

I've changed the wording to @smichel17's second suggestion:

**Is this water source potable? (*Natural Spring*)**

- Safe to drink (signed)
- Safe to drink (unsigned)
- Not safe to drink (signed)
- Not safe to drink (unsigned)

> You may wish to include a note to translators that they can translate either way.

The strings are now called `quest_drinking_water_potable_signed` etc.
Do you think that implicit hint is sufficient enough for translators?

smichel17:
I don't know how it appears in POEditor but assuming the string name is visible, I think that it is probably good enough. But it's also hard for me to know, since I've thought about it, vs seeing it for the first time… but also if a direct translation of "potable" is well known in that language, I would somewhat expect them to use that word anyway, as opposed to a word-by-word translation of "safe to drink". So yeah, probably good enough. If you really want, you could send one string to a translator and ask them to try translating it and then link to this conversation and ask if the flexibility in phrasing was understood :shrug:

FloEdelmann:
> I don't know how it appears in POEditor

It does:

matkoniecz:
> It seemed like the easiest for me to implement sweat_smile

Obviously, I may be wrong but #2507 looks nice (reuse of existing code or maybe even simple yes/no quest), well researched, propably has no tagging traps.

FloEdelmann:
So what is left to do here?

* I believe the wording is as simple, concise and short as possible now.
* Translators have an additional hint in the string names (which are shown in POEditor).
* All fuzzy options have been removed in favor of the "Can't say" option.
* The "signed" options are easy to determine and have a clear tagging scheme.
* The "unsigned" options are more fuzzy and depend on the user's judgement. I think that most false conclusions are prevented by the wording though.

The quest does still need an achievement category. Is it more fitting for "outdoors" or "citizen"?

matkoniecz:
> The quest does still need an achievement category. Is it more fitting for "outdoors" or "citizen"?

outdoors to me, though quest can be in two categories (at least one is already)

westnordost (approved):

westnordost:
Yeah, I'd add it to outdoors

westnordost (approved):

PR #2511, opened 2021-01-17
Extract remaining quest data models

FloEdelmann:
Follow-up to #2501. References #2494.

Now the only quests extending `OsmFilterQuestType<String>` are those with a free-text input field (with suggestions), namely `AddAtmOperator` and `AddChargingStationOperator`.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/DietType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/DietType.kt
@@ -1,0 +1,7 @@
+package de.westnordost.streetcomplete.quests.diet_type
+
+enum class DietType(val osmValue: String) {
+    DIET_YES("yes"),
+    DIET_NO("no"),
+    DIET_ONLY("only"),
+}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/internet_access/InternetAccess.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/internet_access/InternetAccess.kt
@@ -1,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.internet_access
+
+enum class InternetAccess(val osmValue: String) {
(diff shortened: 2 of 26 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/DietType.kt
```diff
@@ -0,0 +1,7 @@
+package de.westnordost.streetcomplete.quests.diet_type
+
+enum class DietType(val osmValue: String) {
```
westnordost, line 3: Under the name "diet type", I'd expect values such as "vegan", "halal", "vegetarian" etc.

Right now, no better name for the enum comes to my mind... HasDiet?

app/src/main/java/de/westnordost/streetcomplete/quests/internet_access/InternetAccess.kt
```diff
@@ -0,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.internet_access
+
+enum class InternetAccess(val osmValue: String) {
+    WLAN("wlan"),
```
westnordost, line 4: should rename the enum to "WIFI", WLAN is only really used in German.
The OSM value must stay "wlan" of course.

westnordost (commented):
Oh, nice work! I didn't realize there were so many quests who do **not** use the image picker that were still string-based.

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/DietType.kt
FloEdelmann: `DietAvailability`?

westnordost: sounds good

westnordost (approved):

PR #2513, opened 2021-01-17
Exclude strictly vegetarian/vegan places from kosher quest

FloEdelmann:
Because of https://wiki.openstreetmap.org/wiki/Key:diet:

> Places tagged as strictly vegetarian or vegan should be implicitly understood by renderers as also being kosher and halal.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddKosher.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddKosher.kt
@@ -13,6 +13,7 @@ class AddKosher : OsmFilterQuestType<String>() {
           amenity ~ restaurant|cafe|fast_food|ice_cream
           or shop ~ butcher|supermarket|ice_cream
         )
+        and diet:vegetarian != only and diet:vegan != only
         and name and (
           !diet:kosher
           or diet:kosher != only and diet:kosher older today -4 years


westnordost:
Makes sense, thanks!

HolgerJeromin:
Why `only`? 
If there is some (real full featured) vegan/vegetarian meal you can get full if you only eat kosher.

matkoniecz:
But place may be:

- not kosher at all (if there is not a separate set of utensils/plates/pots for unclean meat)
- serve kosher and not kosher things
- serve only kosher things

> Places tagged as strictly vegetarian or vegan should be implicitly understood by renderers as also being kosher and halal. 

I suspect that it may be not true and that you can be fully vegan and not kosher (depending on strictness level I guess).

georg-d:
Did you talk to people living that diet? I was wondering why even pure bottled water needs a "kosher" sign and was told that even 100% vegetarian food products like water or apple juice may or may not be kosher - depending on how they have been purified/filtered (several filters contain parts from pork/pig). Same applies to halal.

westnordost:
Hm okay..., can you link a source for that?

georg-d:
Sorry, for this special example not, it was verbal with several muslims and some people that lived in Israel. But to underline why I question whether it's valid to deduct "vegetarien is always kosher", have a look at e.g. https://en.wikipedia.org/wiki/Kashrut#Kosher_utensils telling you'd need two types of dishes or - depending on their material - need at least do special treatment between uses, or https://en.wikipedia.org/wiki/Kashrut#Passover_laws telling some types of bread may not be kosher depending on time of the year respectively /holiday, or https://en.wikipedia.org/wiki/Kashrut#Vegetables some food cooked exclusively by non-Jews is not kosher. But things are not only dependant on just day, date/holiday, ingedients and preparation crew & material, but also e.g. on the "flavor"/"school" of Jewish believe _Due to differences in kashrut standards held by different organizations, the hechsheirim of certain Jewish authorities may at times be considered invalid by other Jewish authorities.[94]_ https://en.wikipedia.org/wiki/Kashrut#Product_labeling_standards

westnordost:
Thank you!

Well then it looks like, this change should be reverted. What do you think, @FloEdelmann?

FloEdelmann:
It seems you are right:

https://www.chabad.org/library/article_cdo/aid/547235/jewish/Are-vegan-restaurants-automatically-kosher.htm

https://www.iamgoingvegan.com/is-vegan-kosher/

So we should revert this change and also remove the sentence from the wiki page (and its translations).

westnordost:
ok, reverted

FloEdelmann:
Interestingly, the same is true for halal food, too: https://crueltyfreereviews.com/articles/is-vegan-food-halal/

I'll update the English, German and French wiki pages.

PR #2517, opened 2021-01-19
Add bike parking access and fee quests

### LINKED ISSUE #2496: new quest - bike parking access
### General
Affected tag(s) to be modified/added: [access](https://wiki.openstreetmap.org/wiki/Key:access)
Question asked: **Is it restricted who may park their bike here?**

### Checklist
see below
see below

About the "no spam" and "useful purpose" points:

A few years ago, I would have considered this quest spammy. Most bicycle parkings were either public, or for customers. I didn't know any case where a shop owner checked if their bicycle parking was really only used by customers. Thus, the access tag wasn't really necessary.
Now I see some change here. Many newly-built apartment or office buildings come with bicycle parkings, only for their residents/employees. Where I live, in the Ruhr Area, we're just at the beginning of this process. But having a look at advanced countries like Denmark, this will develop rapidly.
Even in areas where there are very few private bicycle parkings (yet), they're at least divided in public/customers. Once the bicycle may become more popular with the masses, shop owners may start enforcing to have their parkings used by customers only. So tagging this here is also useful.

### Ideas for implementation
 identically to the car parking access quest

**Element selection:** 
`amenity=bicycle_parking and bicycle_parking!=informal and (!access or access=unknown`

**Metadata needed:** 

**Proposed GUI:**
can be copied from car parking access quest, solely the question itself should be changed.

### LINKED ISSUE #2507: bike parking fee

FloEdelmann:
Closes #2496, closes #2507.

As @matkoniecz noted in https://github.com/streetcomplete/StreetComplete/pull/2509#issuecomment-762879274, this was much easier to implement than #2509. Also it should be less controversial :wink:
However, I had already started coding #2509 when #2507 was opened, so I stuck with it first.

This also includes an icon for #2496, as I created that icon first by mistake, and then didn't want to throw it away :smile: 
**EDIT:** Now this PR also includes a full quest for that icon :wink:


FloEdelmann:
So, I just learned that sadly [renaming a branch](https://github.blog/changelog/2021-01-19-support-for-renaming-an-existing-branch/) from which a PR is open will close the PR.

westnordost:
I think if this is added, also #2496 should be added (with a higher priority, so it does show first). I know, there is disagreement whether it should be asked or not, but I think everyone agrees that at leeast for shed/building/lockers it makes sense to ask.

Would you include a "add bicycle parking access" quest for only shed/building/lockers as well into this PR?

app/src/main/java/de/westnordost/streetcomplete/quests/parking_fee/FeeAnswer.kt
```diff
@@ -1,11 +1,38 @@
...
+        changes.addOrModify("fee:conditional", "yes @ (${openingHours})")
+    }
+}
+
+data class HasFeeExceptAtHours(val openingHours: OpeningHoursRuleList) : FeeAnswer() {
+    override fun applyTo(changes: StringMapChangesBuilder) {
+        changes.updateWithCheckDate("fee", "yes")
+        changes.addOrModify("fee:conditional", "no @ (${openingHours})")
+    }
+}
```
westnordost, line 38: Actually, it would be better to have the `applyTo` as an extension function (or just static function). To apply an answer to a StringMapChangesbuilder is not an essential part of a fee answer, rather it is a functionality of the quest.

westnordost (changes requested):

matkoniecz:
> I think everyone agrees that at leeast for shed/building/lockers it makes sense to ask.

+1

app/src/main/java/de/westnordost/streetcomplete/quests/parking_fee/AddBikeParkingFee.kt
```diff
@@ -9,7 +9,7 @@ class AddBikeParkingFee : OsmFilterQuestType<FeeAnswer>() {
     // element selection logic by @DerDings in #2507
     override val elementFilter = """
         nodes, ways, relations with amenity = bicycle_parking
-        and access !~ no|private
+        and access and access !~ no|private
```
westnordost, line 12: Maybe 
```suggestion
        and access ~ yes|customers|public
```
to be consistent with the car parking access quest?

westnordost: For example you probably wouldn't want to ask for `access=permissive`

westnordost (approved):

westnordost (approved):

westnordost:
👍

michaelblyons:
For "It is open to the public, but there's a yearly fee and a _huge_ waiting list to get one of the spots" cases, should I just mark it "private?" "Public" only means walk-ins welcome, right?

FloEdelmann:
@michaelblyons If in doubt, better leave a note. In this case, I'd say "private" is the right choice though: https://wiki.openstreetmap.org/wiki/Key:access#List_of_possible_values

Issue #477, opened 2017-08-13
(Wiki) site with list of all quests


rugk:
### Reasoning/user "stories"

For the following reasons, it would be nice to have a complete list of all quests available:
* For issue creators suggesting new quests/ideas, to see whether a quest is already implemented or there is a similar quest (ehm… https://github.com/westnordost/StreetComplete/issues/394 🙄)
* For developers/contributors, to check whether there is a similar quest (code) they can reuse.
* For users as "advertisement". "See how much quests you can answer here!" (potentially in conjunction with "And you can contribute your own ones!"). You could include it in app description, etc.
* For the curious, [who don't know what is "hidden" behind these icons](https://github.com/westnordost/StreetComplete/blob/master/res/quest_icons.svg), but like to know…
* For OSM mappers, so they can see what things this app can modify/comment, etc.

And for various other user questions:
* For "Ehh, I saw an icon there once, but I do not remember what type of quest this is?"
* For "Hmm, can I also use this quest to answer XY?"
* For "I only see quests about street? Does this mean **Street**Complete is only about streets?"

### Concrete suggestion

I image a table or a similar format, which should contain the following elements (in the following or a similar order, ordered by importance):
1. Quest icon.
2. Quest name (short)
3. Quest question
4. What the quest means, if it is not clear OR an example OR a screenshot (may take too much space, however, but on the other hand also shows how the user input/selection is done)
5. Link to code where it is implemented, and – if it applies – link to resource files (YAML) used
6. Link to issue/discussion/PR, if there is one
7. (optional) link to example changeset or similar explanation/depiction of what may/is changed by this quest
8. (optional) since when (version number and/or date) the quest is included

I tried to include all the interests from the "user stories" above. Anyway, before anyone (including me) starts to make this table, let's first finish this discussion in this issue about **what** to include and in **which order**.

#### Why wiki?
Just because that may be a good thing for all users to maintain. And your [wiki](https://github.com/westnordost/StreetComplete/wiki) needs content… :wink:

westnordost:
Within the app, there will be a list because #35. It will have the
- quest icon
- the quest question
but not the name, because it does not exist yet and it would require another translation for each quest (and also not the other things).

This covers a few of your user stories. 

[...]

PR #2523, opened 2021-01-21
Add automatically updated quest list

FloEdelmann:
I've had a look at the [OSM Wiki quest list](https://wiki.openstreetmap.org/wiki/StreetComplete/Quests) today, and noticed that some information could be automatically generated. Especially updating the default priority manually (which is a bit outdated there) is quite tedious.

So I wrote a JavaScript script (because that's the language I'm most fluent in) that generates a Markdown file in the repository root directory. It is automatically called by a GitHub Actions workflow whenever new commits are pushed to the `master` branch.

This quest list is not meant to replace the wiki page, but rather to enhance it, and to make it easier to keep the wiki page up to date.


westnordost:
Hmm, yes, updating the default priority manually on the wiki is tedious. This is why I don't do it and would advise those who maintain that wiki page (@rugk et al) to not include it at all.

Apart from that, I don't see that much of a use case / advantage for this file. 
- The icons are already in the wiki
- The links to the code are already in the wiki
- The question text is already in the wiki
- the since-version-information is already in the wiki - and also more precise. The automatically generated information is wrong to a degree where it doesn't make sense to display it at all in my opinion

So the only advantage here is to see the default order of the quests. But this can already be seen from within the app and also by looking at that one source code file where they are neatly arranged - with comments as to why.

westnordost:
One has to ask oneself why there needs to be such a list at all, outside the app.

I see only one use case:

(Experienced) OpenStreetMap users who want to look up what a certain quest does and tags, maybe triggered by a weird changeset or whatnot. With this use case, it is most important to see:
- What question is being asked
- How the interface looks like
- What is being tagged
- For which elements is it asked
- To see the discussions that led to exactly this implementation

The list in the wiki fulfills this use case mostly (except showing how the interface looks like), because it links to the actual code, the pull request and issue, also shows a brief summary for what it is being asked and what it tags.
The automatically generated list does almost do nothing of this.

matkoniecz:
> One has to ask oneself why there needs to be such a list at all, outside the app.

It is also useful to list what StreetComplete can do visiting website is lower barrier to entry than installing an Android app.

rugk:
Hi @FloEdelmann,
first of all great idea and thanks putting so much work into it. Speaking from the wiki-maintenance side here, I totally support your point that especially the priority thing is annoying to maintain. (Overpass queries were even worse in the past, but we dropped that.)
Unfortunately see it similar. Information duplication (especially when it may partially be wrong) is not a good thing usually.

So yes, automation can always help us here, but IMHO a separate Markdown page is not a good place for that. When we have started the wiki page (a long time ago this was: https://github.com/streetcomplete/StreetComplete/issues/477), the intention was mostly to provide a accessible wiki page list for non-technical people and a good overview for many use cases (contributors; OSM mappers etc.).
Also, what became clear to me only later is that the OSM wiki (i.e. Wikimedia) software (now with VisualEditor) is clearly superior to the GitHub wiki or any markdown page, because it is easier to edit (visually, no code needed), especially concerning the table editor and the features it offers that go way beyond simple Markdown tables (table sorting, templates etc.). With Markdown it would be a pain to maintain it like that.

That al said, I am always for automating things. And the real advantage of this PR (and if I understood correctly also the idea that sparked it) was the _default priority_ column.
And here I'd actually say that this column useful. It may be some work, but if it is automated that is a work that allows users to sort the table as they see it in the app, which is useful, because they can then find the quest they want to disable in the app or so. Or to see the quests regarded as "most important" first etc.

So this would be helpful if we could integrate that result (easily) in the OSM wiki page. Thus, maybe we can generate Wiki syntax instead of markdown, @FloEdelmann? Or somehow make it possible to copy and paste that column into the Visual Editor page (is that possible? I have not tried that.)
And if that works, it also does not have to be in this repository – where it duplicates information, but we may hide it in a separate one – [another community project as the ones we know already](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#streetcomplete-related-projects). I could maybe imagine a website that does this as a standalone thing – triggered on access of the website? If it is possible to do in a browser?

As for the other generated information they can usually be easily copied into the OSM wiki page anyway (from release notes, only the SVG needs to be searched in the source code tree). So having that generated Markdown page provides little advance – though it may somewhat ease copying that stuff.
In any case I would want to use this here to make contributing to the OSM wiki page easier and integrate the information there (note Wikimedia bots would also be possible, but likely yet again more difficult to create) instead of duplicating the information, if that is somehow possible.

rugk (changes requested):

res/graphics/quest icons/wheelchair_outside.svg
```diff
@@ -0,0 +1 @@
+/home/flo/www/StreetComplete/res/graphics/quest icons/toilets_wheelchair.svg
```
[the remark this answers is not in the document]
FloEdelmann: The change wasn't wrong, I've symlinked `wheelchair_outside.svg` to `toilets_wheelchair.svg`, as the former had no separate icon. However, this is not needed anymore (because I dropped the icon information), so I reverted the commit.

The rename of `guidepost.svg` to `board_type.svg` because the corresponding Android drawable is also called `ic_quest_board_type` and not `ic_quest_guidepost`: https://github.com/streetcomplete/StreetComplete/blob/86b082850c0ea13c0a942255edd7ce5a94808460/app/src/main/java/de/westnordost/streetcomplete/quests/board_type/AddBoardType.kt#L16

Now, all quest icons' drawable names match the SVG file names.

FloEdelmann:
@westnordost This PR wasn't meant to be merged as is, but rather as a proof of concept to show what automation is possible, and as a way to get feedback. And that worked perfectly :sweat_smile: 

@rugk Thanks for your answer!

I've updated the PR to generate a CSV (comma-separated values) file instead: https://github.com/FloEdelmann/StreetComplete/blob/quest-list/quests.csv

Note that you can easily see there that two quests in the wiki page are no longer in the source code: 

* What surface does this road have here?
* What specific surface does the road […] have here?

… and four quests are in the source code but not yet in the wiki page:

* What surface does this piece of road have?
* Is it restricted who may park a bike here?
* Do you have to pay to park a bike here?
* Is the water potable here? […]

When the CSV file is opened with LibreOffice Calc (or probably Excel), you can copy the *Default priority* column and paste it into the OSM Wiki's VisualEditor. For that to work, I needed to use the same order as in the wiki page, so the script now fetches the wiki page content from the MediaWiki API.

---

My idea for a possible workflow is like this:

The script stays in this repo and runs automatically when new commits are pushed to the `master` branch. The CSV file is not committed to the repo, but rather a "build artifact" (see e.g. this guide to learn about artifacts: https://elanderson.net/2020/06/github-use-actions-to-publish-artifacts/).

The artifact can be downloaded from the GitHub Actions workflow run page (see example workflow run here: https://github.com/OpenLightingProject/open-fixture-library/actions/runs/515967387), and opened in LibreOffice.

Then all old quests are removed from the wiki page and the corresponding rows removed in LibreOffice.

Then all new quests are added to the wiki page and manually sorted in LibreOffice to the new location.

Then the *Default priority* column can be copied and pasted into the VisualEditor. That's it :slightly_smiling_face:

FloEdelmann:
Leaving the script in this repo rather than a separate one has the advantage that it can run automatically when new commits are added to the `master` branch.

Running in the browser on demand is not feasible because it needs to fetch information from many different files from this repo. So it'd need GitHub API access. Also, copying from tables in browsers is always a mess :smile:

rugk:
@FloEdelmann Thanks! I have not tested it, but from what you write it seems to indeed may maintenance of the wiki easier.

And that LibreOffice CSV trick is a very neat idea! :smiley: 
I've tested that and VisualEditor happily accepts that (just need to use the version of the next commit, as you've fixed the misnamed quests. :smiley: 

> The CSV file is not committed to the repo, but rather a "build artifact" (see e.g. this guide to learn about artifacts: https://elanderson.net/2020/06/github-use-actions-to-publish-artifacts/).

Totally agree this is the correct approach. Also, it's then hidden and we may only give a tip to that information to Wiki contributors (in a Wiki comment or whereever) that this data is available.
Quite like the "Open data" approach we then have with this CSV.

.github/workflows/generate-quest-list.yml
```diff
@@ -0,0 +1,19 @@
...
+  generate-quest-list:
+    name: Generate quest list
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2.1.4
+        with:
+          node-version: 14
+      - run: node .github/generate-quest-list.mjs
+      - uses: stefanzweifel/git-auto-commit-action@v4
```
rugk: Well then… don't commit it, but upload the artifact :slightly_smiling_face:

rugk: Ah, I see you've changed that already. :smiley:

rugk (approved):
LGTM, only thing I would suggest is to maybe structure the JS in [Classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes).

peternewman:
I was just going to come and say the readme was quite useful for some stuff and it's vanished! :cry: 

Given its updated automagically it means it's always going to be current with what I might be running, which is useful for commenting on issues, particularly (https://github.com/streetcomplete/StreetComplete/discussions/2457), whereas the wiki needs people to update it. As to why, it's easier to view/less scrolling than viewing the list on my phone, and I can copy/paste as well as typing on a proper keyboard!

What about making a gh-pages branch which hosts (automatic copies) of the various Markdown files, as well as an auto-generated version of your previous markdown table @FloEdelmann ? Then it could have an "official" website home:
http://streetcomplete.github.io/StreetComplete/

Or is the OSM Wiki felt to be that home for SC?

If that was useful, see the links at the top of this Flo, which point to a branch and some other options where I've done or am investigating similar:
https://github.com/mjemmeson/Date-Holidays-GB/pull/11#issue-555351321

.github/generate-quest-list.mjs
```diff
@@ -0,0 +1,331 @@
...
+    const oldQuests = quests.filter(quest => quest.wikiOrder > -1);
+
+    const csvLines = [
+        '"Quest Name", "Question", "Default Priority", "Wiki Order"',
+        ...wikiTable.filter(quest => quest.isOutdated).map(quest => quest.csvString),
+        ',,,',
+        ...newQuests.map(quest => quest.csvString),
+        ',,,',
+        ...oldQuests.map(quest => quest.csvString),
+    ];
```
peternewman, line 328: Surely there's an npm module to do writing to a CSV properly?

FloEdelmann: I tried to keep it dependency-free (apart from Node.js), as proper dependency management requires an extra `package.json` file and ongoing maintenance. Writing the CSV file manually may not as pretty, but I think it's good enough for a script like this.

FloEdelmann:
Pros and cons of my original automatically updated [Markdown quest list](https://github.com/streetcomplete/StreetComplete/blob/dc9981671d6a6154538fa909f4ce934278d9cb74/quests.md), compared to the Wiki page:

```diff
+ always updated
+ no manual maintenance
+ includes quest class name (with link to Kotlin file)
- version information unreliable (in such a grade that it's mostly useless)
- no "Asked for elements" column
- no "Modified tags" column
- no "Notes" column → this will never be possible to add automatically
- no "Code" column with link to issue, PR, code directory
- information duplication → which list is authoritative (for which attributes)?
```

It was never meant to replace the wiki list, as too much from it is missing. However, using it as an assistance for wiki editors is a good idea, but not the best execution either, as @westnordost and @rugk showed.

I think the [automatically updated CSV list](https://github.com/streetcomplete/StreetComplete/pull/2523#issuecomment-769231699) is a good help for this purpose though.

---

Whether StreetComplete should get its own homepage should be discussed separately.
If it's gonna be GitHub Pages, I'd suggest http://streetcomplete.github.io/ instead of http://streetcomplete.github.io/StreetComplete though. Then it'd be a separate `streetcomplete.github.io` repo in the `streetcomplete` organization.

westnordost:
I didn't follow the whole discussion here.

So an automatically updated quest list could be produced whose purpose is to aid wiki editors, a far as I understand. What information should this automatically generated information contain? I'd say, the **difference** to the wiki page would be most helpful:
- the English title if it differs what can be found on the wiki
- the link to the source file if it differs
- the order if it differs(?)

Earlier in thread, a CSV was linked with only the information from this repos. This means that the task of finding the differences is still up to the wiki editors. Maybe this step could be automated, so that the script only outputs the "warnings" when it found a difference

FloEdelmann:
The CSV file ([example](https://github.com/FloEdelmann/StreetComplete/blob/e2a65272d0c855337cf16717a5c9e8a1115a2c0e/quests.csv)) contains 3 sections:

1. All quests that are in the wiki but not in the repo
2. All quests that are in the repo but not in the wiki
3. All quests found in both places that could be matched

Matching happens by quest title, so if a quest title is wrong in the wiki, it shows up in sections 1 and 2.

The priority column can be copied as is from the CSV to the Wiki Editor once the quests are manually matched. See the workflow in my earlier [comment](https://github.com/streetcomplete/StreetComplete/pull/2523#issuecomment-769231699) (below the horizontal line).

westnordost:
So, basically this is the final version and ready to review/merge?

.github/generate-quest-list.mjs
peternewman: Yeah I guess you're safe until a quest gets a comma or something, although I guess Java may stop that happening.

peternewman:
Nice abuse of diff @FloEdelmann !

> I think the [automatically updated CSV list](https://github.com/streetcomplete/StreetComplete/pull/2523#issuecomment-769231699) is a good help for this purpose though.

Yeah, although downloading an opening a CSV is a bit more hassle if you just want to copy a new quest name.

> Whether StreetComplete should get its own homepage should be discussed separately.

Yeah agreed as a whole, I'm not sure why the CSV and MarkDown versions couldn't be evergreen via GH Pages. Which would also work around the expiry of logs and exported items.

> If it's gonna be GitHub Pages, I'd suggest http://streetcomplete.github.io/ instead of http://streetcomplete.github.io/StreetComplete though. Then it'd be a separate `streetcomplete.github.io` repo in the `streetcomplete` organization.

Shouldn't that be for the project as a whole (i.e. all the linked software, rather than just the App?

westnordost:
Sorry that I haven't reviewed or merged this PR yet. I understand that it is helpful for people that keep the wiki up to date and thus should be merged. I currently have a lot to do and this isn't at the top of my list. I will come around to it eventually though!

Btw I wonder if github allows also hooks being written in kotlin. If it was kotlin, it would be easier for me to maintain it.

rugk:
> Btw I wonder if github allows also hooks being written in kotlin. If it was kotlin, it would be easier for me to maintain it.

It's Ubuntu VMs. So anything can be executed there, basically…

FloEdelmann:
I've converted the script to Kotlin. From my side, this PR is ready now. A sample GitHub Actions workflow run can be seen in my fork: https://github.com/FloEdelmann/StreetComplete/actions/runs/563454452

westnordost:
> I've converted the script to Kotlin.

Wow! Damn! That's cool!

westnordost:
Maybe if I crib from this script, I can write another script some time that collects the information what tags are tagged by StreetComplete for taginfo, to replace the hoplessly outdated https://github.com/goldfndr/StreetCompleteJSON with something automatically generated.

.github/generate-quest-list.main.kts
```diff
@@ -0,0 +1,204 @@
...
+
+fun getWikiTableContent(): String {
+    val page = "StreetComplete/Quests"
+    val section = 1 // "Released quest types" section
+
+    val apiUrl = URL("https://wiki.openstreetmap.org/w/api.php?action=parse&format=json&prop=wikitext&formatversion=2&page=$page&section=$section")
+
+    val jsonString = apiUrl.openStream().bufferedReader().use { it.readText() }
+
+    return Json.decodeFromString<ApiResult>(jsonString).parse.wikitext
```
westnordost, line 161: Ah, that's clever to use the serialization library for parsing JSON

.github/generate-quest-list.main.kts
```diff
@@ -0,0 +1,204 @@
...
+
+val csvFile = projectDirectory.resolve("quest-list.csv")
+
+val noteQuestName = "OsmNoteQuest"
+val noteQuestFile = sourceDirectory.resolve("data/osmnotes/notequests/OsmNoteQuestType.kt")
+
+val wikiRowSpan2 = " rowspan=\"2\" |"
+
+val questFileContent = sourceDirectory.resolve("quests/QuestModule.kt").readText()
+val questNameRegex = Regex("(?<=^ {8})[A-Z][a-zA-Z]+(?=\\()", RegexOption.MULTILINE)
```
westnordost, line 26: Ui, I suppose this is ok, though this might be a little unstable.

westnordost (approved):
I didn't try to understand the whole script line-by-line.

What I can say generally about it is that it seems to me to be very well done! Also cool: The Kotlin script is 60% shorter than the JavaScript version.

When not looking at this script from within an IDE, it would help if there was a clear separation of global variables from functions and classes. In other words, if classes and functions would not access globals but instead get passed in what they need (and output what they produce, i.e. not modify anything). Or yet in other words, confine use of global variables to the plane script-part of the file (until line 36).
I noticed for example that `writeCsvFile` uses the `quests` global variable, and the `wikiTable` and probably some more and had to scroll way up to see what exactly they are.

westnordost:
Out of curiousity, how did you develop/test this script and the javascript one before? How were you able to run it for testing if everything works alright?

.github/generate-quest-list.main.kts
FloEdelmann: It is indeed. Can you think of a smarter way?

FloEdelmann:
You can run the script manually with the following command:

```
kotlinc -script -Xplugin=/path/to/kotlinx-serialization-compiler-plugin.jar .github/generate-quest-list.main.kts
```

Unfortunately, you need to manually pass that `kotlinx-serialization` compiler plugin path then (`/path/to/` is `/snap/kotlin/current/lib/` for me on Ubuntu, like in the GitHub Actions workflow).

Previously with the JavaScript version, I could run `node .github/generate-quest-list.mjs` after installing [Node.js](https://nodejs.org/).

The script then generates the CSV locally, which I inspected manually.

.github/generate-quest-list.main.kts
westnordost: Maybe look at all files in the `/quests` directory and search for `class ([A-Z][a-zA-Z]+)\s*:\s*(OsmElementQuestType|OsmFilterQuestType|QuestType)`, i.e. all classes that inherit from QuestType or a subclass/interface of it.
This is also a bit unstable of course, because there might be new subclasses/interfaces in the future.

FloEdelmann: We'd still need to parse the `QuestModule` though, to find out the order (default priority).

smichel17: I think it's fine if the regexes are unstable — as long as updating the script is less work than manually updating the wiki, it's a net gain.

That said, it may not be a net gain if the regex breaks *silently* and produces wrong results. I've only read the comments here, not dug into the code, but I wonder if there is a way to put a sanity check on the regex result and have it fail if the sanity check fails. I'm thinking something along the lines of a hard-coded assertion about the number of quest names we expect to be found, where the script fails if it's the wrong number. Of course, this will break every time a quest is added or removed, but that's just the point. If it's expected breakage (because a quest was added/removed), it's simple enough to update the number. If it's unexpected, then the check has done its job.

westnordost: right

westnordost:
Huh, why didn't I merge this yet?

westnordost:
Ok it looks like I was waiting if you wanted to do anything on my remarks, which you did. So, thank you :-)

FloEdelmann:
@rugk Now that this is merged, we can see the first workflow run with the generated CSV file: https://github.com/streetcomplete/StreetComplete/actions/runs/582771780

Here is a list with all workflow runs (the newest one will be on the top): https://github.com/streetcomplete/StreetComplete/actions/workflows/generate-quest-list.yml

I think we should link to that page from a comment on the wiki page, right?

.github/generate-quest-list.main.kts
peternewman: Sorry I meant to comment sooner. Now this is also Kotlin can't you just load QuestModule and access the data from that directly the same way the app does? Which will then be fine until QuestModule gets changed.

FloEdelmann: That seems not possible, because the script is not built together with the other source files. Also if it was, we'd need to supply suitable (probably mocked) parameters.

If someone finds a solution for this, feel free to change it :)

rugk:
Yes, feel free to link it. Maybe only in a comment, so only editors can see it… although, maybe it would be helpful for users who notice an error to see.

FloEdelmann:
### Short summary

In this repository, a [Kotlin script that generates a quest list CSV file](https://github.com/streetcomplete/StreetComplete/blob/master/.github/generate-quest-list.main.kts) runs automatically [every time new commits are pushed to the `master` branch](https://github.com/streetcomplete/StreetComplete/blob/master/.github/workflows/generate-quest-list.yml).

When updating the [StreetComplete quest list in the OSM Wiki](https://wiki.openstreetmap.org/wiki/StreetComplete/Quests), follow these steps:

1. Download the CSV file of the [latest workflow run](https://github.com/streetcomplete/StreetComplete/actions/workflows/generate-quest-list.yml) and open it in a spreadsheet editor (e.g. Excel or LibreOffice Calc).
2. If the CSV file contains old quests (present in wiki page, but not in repository), remove them from the wiki table and from the spreadsheet.
3. If the CSV file contains new quests (present in repository, but not in wiki page), add them to the wiki table and manually sort the respective rows in the spreadsheet to the new location.
4. Copy the *Default priority* column and paste it into the VisualEditor of the wiki page.

rugk:
@FloEdelmann Having used that table now "in production", I have one thing that could be improved: In the markdown version we had a column for the quest icon IIRC.
For [that new crown quest](https://wiki.openstreetmap.org/w/index.php?title=StreetComplete%2FQuests&type=revision&diff=2131008&oldid=2130991) it took me some minutes to actually find the correct icon due to the different naming (I only looked at the source code later), so one idea may be to
Or we introduce the Markdown table again, which can be used to copy the link – though I guess this also is not a lot more convenient than the CSV, so yeah… just thinking aloud here…

Issue #2519, opened 2021-01-21
Add possibility to mark sidewalks


yyatsyn:
Possibility to mark sidewalks along the street (both, left, right, none -- as [here](https://wiki.openstreetmap.org/wiki/Key:sidewalk)) would be great as such information can help for pedestrian navigation. This feature will require taking directions of streets' ways into account.

FloEdelmann:
This was implemented already in version 10: https://wiki.openstreetmap.org/wiki/StreetComplete/Quests#cite_ref-:8_16-0

How, we're at version 29. Didn't the quest show up for you?

yyatsyn:
thanks! weird that I've never seen this question about sidewalks popping up

westnordost:
You can move the priority of the quest further up so that it is not obscured by other quests.

Also, it is only shown for street segments for which this expression is true:

```
        ways with
          highway ~ primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential
          and area != yes
          and motorroad != yes
          and !sidewalk and !sidewalk:left and !sidewalk:right and !sidewalk:both
          and (
            !maxspeed
            or maxspeed > 8
            or (maxspeed ~ ".*mph" and maxspeed !~ "[1-5] mph")
          )
          and surface !~ unpaved|compacted|gravel|fine_gravel|pebblestone|grass_paver|ground|earth|dirt|grass|sand|mud|ice|salt|snow|woodchips
          and lit = yes
          and foot != no and access !~ private|no
          and foot != use_sidepath
```

... and also, there are no separately mapped sidewalks nearby.

yyatsyn:
Hm, why lit=yes is required?
Most of streets in my town are with lit=no or lit=<NA>, but they do have
sidewalks on one or both sides which I would find useful to map.
Best regards,
Yury

westnordost:
Because the quest should only be asked for roads that are likely within a built-up area. There is no precise tag to find this out (there is no `urban=yes` or anything), lit is quite fuzzy way to exclude roads in the countryside

riQQ:
Maybe you could do `and (lit = yes or highway = residential)` to ignore `lit` for `residential` streets.

PR #2531, opened 2021-01-24
Show sidewalk quest on residential streets regardless of lit tag

riQQ:
See #2519


westnordost:
That's fine, ok! Thanks!

Issue #2336, opened 2020-12-02
Memory Allocation Failure / Crash on Startup


nikhiljha:
While running with grapheneOS, the app crashes on startup. I think I've narrowed this down to the hardened malloc implementation that grapheneOS uses, indicating memory management issues with the `com.mapzen.tangram` package.

```
12-02 10:04:08.935 26927 26927 E AndroidRuntime: FATAL EXCEPTION: main
12-02 10:04:08.935 26927 26927 E AndroidRuntime: Process: de.westnordost.streetcomplete, PID: 26927
12-02 10:04:08.935 26927 26927 E AndroidRuntime: java.lang.RuntimeException: Unable to create a native Map object! There may be insufficient memory available.
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapController.<init>(MapController.java:163)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView.getMapInstance(MapView.java:276)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView.initMapController(MapView.java:173)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView$1.onLibraryReady(MapView.java:120)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView$1InitTask.onPostExecute(MapView.java:236)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView$1InitTask.onPostExecute(MapView.java:224)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.AsyncTask.finish(AsyncTask.java:771)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.AsyncTask.access$900(AsyncTask.java:199)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:788)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:106)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:223)
[...]

Issue #1810, opened 2020-05-02
Quest focus-ring appears shifted.


Sequynth:
**How to Reproduce**
It happens for every quest type. I tried complete reinstall, still appears shifted

**Versions affected**
v19 (fDroid)
Android 6.0
LG G4 H815

smichel17:
It's fixed, just not released yet.

Sequynth:
I got the newest version from fDroid today, but the shift is still present, albeit smaller than before
![Capture+_2020-05-10-14-42-32](https://user-images.githubusercontent.com/29228990/81500774-9fcdad00-92d4-11ea-8919-e6931173bb9c.png)

westnordost:
If you rotate the map, does the selection ring "wander" or does it stay at the same position relatively to the pin?

smichel17:
Always the same relative position (slightly below, slightly left).

westnordost:
What's your device? Specifically, what is the DPI of your device?

smichel17:
Device is a Samsung Galaxy S6, specifically the SM-G920A.

xxxhdpi, if that's what you were looking for

westnordost:
Ok, then it is likely a problem with tangram-es which apparently treats pixel offsets for elements in layers differently than for markers.

I can reproduce it on a xxxhdpi emulator. On a hdpi device, the circle offsets slightly to the right.

Though, the backlog of bugs I reported in tangram-es is already so full, I kind of don't want to add even more there as long as the others are not fixed. Partly because I will have to research the exact problem myself as well and create a minimal example. Perhaps someone else would like to do that and report the exact problem? Setup would be a map data layer consisting of a single dot at a certain position and a map marker at the same position, then, add a certain same offset for both in the style.

Sequynth:
My density is: 1440 x 2560 pixels, 16:9 ratio (~538 ppi density). I cannot convert that  to '*hdpi' units like smichel17.

Same behaviour as he describes, offset is constant under rotation.

Does the tangram issue also explain, why the pin is is not exactly on the steps? Its position also changes under rotation (probably rotating around the point it should be contantly point to)

westnordost:
Maybe

RubenKelevra:
I'm not really a fan of pixel peaking on rendering libraries, but I can confirm a very slightly wrong position of the pins.

It's around 10% of the pin size too far down, and stay too far down on the rotation.

Android 9
18:9  6" display with 1080 x 2160px (401ppi)


[...]

PR #2548, opened 2021-02-02
upgrade tangram-es dependency

westnordost:
- fixes workarounded crash (https://github.com/tangrams/tangram-es/issues/2136)
- fixes misplacement of quest ring #1810 (and quest pins)
- fixes workarounded crash on Android API 30 #2336 (https://github.com/tangrams/tangram-es/issues/2215)

Needs someone which an android 11 phone that can test if the app still starts correctly.


westnordost:
Thanks @Helium314  for testing!

PR #2551, opened 2021-02-03
Update Czech translation of path split description

aikencz:
The original translation is not exact - it applies only path surfaces and it may confuse users. I've modified the translation to apply to all path properties.


FloEdelmann:
Please see [CONTRIBUTING.md](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app) for how to contribute translations.

aikencz:
OK, thanks for information.

Issue #2189, opened 2020-10-21
a simple team modus for mapping parties


RubenKelevra:
**Use case**
Currently, it's complicated to use SC with multiple people, since SC shows all quests to everyone. In a mapping party you usually don't want to spread and have everyone walking a street alone, but walk in small groups. Walking the same route with multiple people will lead to many conflicts in the edits, reducing the efficiency because some data will be captured multiple times.

**Proposed Solution**
Create a simple team modus, where you enter the total number of team members, and your member number.

Every team member will only see his part of the quests, which are filtered based on the IDs from OSM database with a simple modulo operation on the ID.

This way there are no conflicts.

While having the team modus active, SC should track in which area the user is using SC. After exiting the team modus again, those areas' quest cache should be invalidated automatically to make sure that the quests solved by the other team members won't be displayed the next time the user want to map there.

westnordost:
Hm! This is a really nice idea. I really like how simple it is conceptually!

What would you propose - how should the UI look like for that?

matkoniecz:
Have setting "Split items among people" asking to enter how many people are in group and which one is your number? Maybe with info that each number should be unique.

Maybe show toast on start reminding that you see only small part of quests if that setting is enabled.

matkoniecz:
Main problem is way splitting, but I think that it is survivable even if slightly less than half of split ways are not handled. There is a possibility to have a local database with exceptions for modulo rule and split ways getting added there (as some of them would be not matched by modulo rule) but it should be fine also without that.

RubenKelevra:
> Hm! This is a really nice idea. I really like how simple it is conceptually!
> 
> What would you propose - how should the UI look like for that?

Well, I would keep it very simple.

Just move the 'Settings' in the menu to the second row and put a Team-Modus on the right side with a group with 3 stick persons.

Then ask for the number of team members and then show colors or maybe even better, some cool avatars - if you find some free ones like those on Netflix for example and write "each team member needs to use a unique avatar."

> Main problem is way splitting, but I think that it is survivable even if slightly less than half of split ways are not handled. 

Even that isn't an issue since the way is only shown to one person. If he/she cuts it, the cache on the other phones wouldn't know that. So as long as the group stays somewhat together, the cut ways shouldn't show up.

Maybe we could just delay the upload of cut ways some minutes - but that leaves the possibility that the data isn't uploaded at all because the app is closed while the timer is running.

rugk:
[...]

PR #2557, opened 2021-02-07
Team mode

FloEdelmann:
I started working on #2189.

Now I'm stuck; I don't know where to implement the actual OSM element filter logic. But I wanted to share my (early) progress anyway. I appreciate any feedback and help on where/how to continue. Also, taking my commits and adding new code yourself is welcome, too!


FloEdelmann:
### Main menu

There is now a "Team Mode" button. When tapping that, the main menu closes and the team mode dialog opens.



### Team Mode dialog

#### Choosing team size

When entering an out-of-bounds number (I allowed team sizes of 2 to  12, quite arbitrarily), a hint is shown.

 

#### Choosing an avatar

When entering a valid number, that many avatars are displayed to choose from, alongside an explanation that every team member has to choose one on their own phone. A different set of avatars is shown for each number, and on every day ([logic here](https://github.com/FloEdelmann/StreetComplete/blob/d868c4b0002321bea1e7eb44683f74c9816d6fdf/app/src/main/java/de/westnordost/streetcomplete/controls/TeamModeDialog.kt#L110-L116)). I used these SVGs: https://www.svgrepo.com/collection/animals-6/

 

**Issue:** The last row of avatars is always cut off. I couldn't get the [`RecyclerView`](https://github.com/FloEdelmann/StreetComplete/blob/d868c4b0002321bea1e7eb44683f74c9816d6fdf/app/src/main/res/layout/dialog_team_mode.xml#L70-L81) to scroll properly without being cut off.

#### Confirming

When one avatar is chosen, it's highlighted so that all team members can compare their choice, and an "Ok" button is shown, similar to the quest layout.

 

**Issue:** The button overlaps the last avatar icon. When wrapping the `ConstraintLayout` in a `RelativeLayout` with no padding (so that I can put the button there, so that it is pushed more towards the dialog edges), it completely messes up the `RecyclerView` height and cuts off the number input on the right side.

westnordost:
Wow, really nice! The wording is on point, I think! Thank you for the screenshots!

A few early comments:
- a floating action button in a dialog is very uncommon, it would be more consistent to use the button bar
- I am not sure if I like the "animal avatars". The idea to not let users choose "a number from 1 to number of people" is good and fair, but introducing the concept of avatars may add some confusion as avatars come with certain expectations tied to them. It may be more clear to let the user **choose a color**: red, blue, green, yellow, black, orange, violet etc. 

> Now I'm stuck; I don't know where to implement the actual OSM element filter logic
The logical place would be `VisibleQuestSource`. This class would have another new dependency called `TeamModeQuestFilter` or something like this with one public method like `isVisible(quest: Quest): Boolean`. That class would just always return true if team mode is not on, and if team mode is on, return `quest.id % teamSize == teamIndex`. `TeamModeQuestFilter` would probably save the team mode in the shared preferences.
Then, that class should maybe have another two methods: 1. setting the team mode (on and off) and 2. checking if the team mode is on. Finally, it should be possible to register a listener onto that class so that other parts of the UI can be notified whenever the mode changes: The `VisibleQuestSource` would be one of those listeners, the TBD UI component that shows whether team mode is on (and in what team one is) would be another.

HolgerJeromin:
+1 for colors instead of animals.

westnordost:
If you need more input, or if you are stuck, feel free to press the "request review" button. Currently I have not so much time, so I don't monitor all the tickets.

peternewman:
>     * I am not sure if I like the "animal avatars". The idea to not let users choose "a number from 1 to number of people" is good and fair, but introducing the concept of avatars may add some confusion as avatars come with certain expectations tied to them. It may be more clear to let the user **choose a color**: red, blue, green, yellow, black, orange, violet etc.

Colour-blindness and how different phones render different colours (and dark mode)?

How about shapes or patterns instead?

FloEdelmann:
Thanks @westnordost.

I think I'm done with the UI and saving the preference, but the actual quest filtering is still missing. In [`VisibleQuestsSource`](https://github.com/streetcomplete/StreetComplete/pull/2557/commits/fd71e5fdc8d2813b648ce13ecaef5880488591a9) I have either access to whether team mode is enabled or not, or to which quests are visible. I don't know how/where to combine that information. A little help here would be much appreciated :blush:

@peternewman I also though about color blindness, so I added letters to the color circles. Also, the positions of the circles are now consistent (it doesn't make sense to randomize these anymore like I had done with the animal avatars), so there are three factors to identify (and match) an option: color, letter and position.

Here are some updated and new screenshots :tada: 

### Team mode dialog

 

### When team mode is activated

matkoniecz:
> Seems like I can't "request a review" from you though, since you have not reviewed the PR (I can only click "re-request review" then).

I could press this button, done

matkoniecz:
Is there something in main view that reminds that team mode is enabled? Reminder that only part of quests is shown would be useful.

FloEdelmann:
> something in main view that reminds that team mode is enabled

The chosen team mode color circle is displayed as a little badge for the main menu button (see the third screenshot in https://github.com/streetcomplete/StreetComplete/pull/2557#issuecomment-778077358). Is that too subtle?

matkoniecz:
I missed it, I though about something like following (feel free to ignore me, I am just trying to simulate what may go wrong with confused/slightly clueless user).

Untitled

westnordost (changes requested):
Regarding the VisibleQuestsSource, locally on my branch it looks completely different now, so I'll just delay advice on this until the other things have been addressed I found in the review.

In short, all `Quest`s that are output from that class must be filtered through the `TeamModeFilter` before they are being returned. Both when quests are being queried by bbox and also in listeners.

westnordost:
Oh yea, additionally I concur with @matkoniecz that there should be some kind of heads-up that you are (still) in team mode.

I suggest a simple toast each time the main menu button with the little annotation becomes visible: "Team mode active"

HolgerJeromin:
Teams, Teamviewer, Chrome paint a thick border around the screen when there is an active screenshare. Same could be done here:
image

matkoniecz:
I admit that at least for me it would be not be recognizable as indicator what is happening (as person who never used screenshare so far).

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
```diff
@@ -0,0 +1,62 @@
...
+/** Controller for filtering all quests that are hidden because they are shown to other users in
+ *  team mode. Takes care of persisting team mode settings and notifying listeners about changes */
+@Singleton class TeamModeQuestFilter @Inject internal constructor() {
+    /* Must be a singleton because there is a listener that should respond to a change in the
+     *  shared preferences */
+
+    @Inject internal lateinit var prefs: SharedPreferences
+
+    private val teamSize: Int get() = prefs.getInt(Prefs.TEAM_MODE_SIZE, -1)
+    val teamIndex: Int get() = prefs.getInt(Prefs.TEAM_MODE_INDEX, -1)
```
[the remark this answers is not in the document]
FloEdelmann: I think it should rather be `teamSize` and `indexInTeam`.

There will only be one team conceptually (the group of people who are mapping the same area together). When they all enable team mode, everyone in the group has a unique index (in the team).

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
```diff
@@ -0,0 +1,62 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Controller for filtering all quests that are hidden because they are shown to other users in
+ *  team mode. Takes care of persisting team mode settings and notifying listeners about changes */
+@Singleton class TeamModeQuestFilter @Inject internal constructor() {
+    /* Must be a singleton because there is a listener that should respond to a change in the
+     *  shared preferences */
+
+    @Inject internal lateinit var prefs: SharedPreferences
```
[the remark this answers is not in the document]
FloEdelmann: It did work in the emulator, but I'm fine with changing it.

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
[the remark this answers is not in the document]
westnordost: Sure, whatever you think is not ambiguous. "Group" and "Team" are interchangable terms then?

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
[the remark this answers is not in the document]
westnordost: Hmm... weird .... who injects the prefs then? Then this is an automatism I didn't know about. Anyway, that lateinit var stuff is just a workaround for Android doing the instantiation of activities and fragments and thus normal dependency injection via the constructor is not possible.

peternewman:
> I admit that at least for me it would be not be recognizable as indicator what is happening (as person who never used screenshare so far).

Presumably in conjunction with something in the menu it's just another nudge though.

app/src/main/res/values/strings.xml
```diff
@@ -825,6 +825,13 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_charging_station_operator_title">"Who is the operator of this charging station?"</string>
     <string name="quest_charging_station_capacity_title">"How many cars can be charged at this charging station at the same time?"</string>
 
+    <string name="team_mode">Team Mode</string>
+    <string name="team_mode_exit">Exit Team Mode</string>
+    <string name="team_mode_description">When mapping the same area together in a group, you can enter team mode to split up quests between all team members.</string>
+    <string name="team_mode_team_size_label">How many people are you?</string>
```
peternewman, line 831: Sorry but this is rather Germanic, one of these two is probably more English.
```suggestion
    <string name="team_mode_team_size_label">How many people are there?</string>
```
```suggestion
    <string name="team_mode_team_size_label">How many people in your team?</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -825,6 +825,13 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="team_mode_team_size_hint">Team mode works for groups between 2 and 12 people.</string>
```
peternewman, line 832:
```suggestion
    <string name="team_mode_team_size_hint">Team mode works for groups of between 2 and 12 people.</string>
```

peternewman (commented):
Some language improvements

peternewman:
> @peternewman I also thought about color blindness, so I added letters to the color circles. Also, the positions of the circles are now consistent (it doesn't make sense to randomize these anymore like I had done with the animal avatars), so there are three factors to identify (and match) an option: color, letter and position.

I considered suggesting letters or numbers, but was wary of how it fits with people using non-Latin scripts, but I guess it's just a symbol at worst, and combined with the colour and position should cover it for all users.

Although position becomes irrelevant when selected.

app/src/main/res/values/strings.xml
```diff
@@ -825,6 +825,13 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="team_mode_choose_color">Everybody in the group has to choose a unique color (on their own phone) now.</string>
```
peternewman, line 833: It's not just a unique colour now, but maybe unique dot is more confusing?

app/src/main/res/values/strings.xml
westnordost: What about "How many are you?" I mean, what else?

app/src/main/res/values/strings.xml
westnordost: No I think color is better. You suggested yourself to show a boundary in the color chosen.

peternewman: Yeah fair enough.

app/src/main/res/values/strings.xml
peternewman: It still doesn't sound quite right to a native.

If you want something shorter:
```suggestion
    <string name="team_mode_team_size_label">How many people?</string>
```

peternewman: Or
```suggestion
    <string name="team_mode_team_size_label">How many of you are there?</string>
```

peternewman: https://ell.stackexchange.com/questions/181748/how-many-are-you-guys-or-how-many-people-are-you-guys-which-one-is-correct

app/src/main/res/values/strings.xml
FloEdelmann: Does this sound a bit nicer?

```suggestion
    <string name="team_mode_team_size_hint">Team mode works for groups of 2 to 12 people.</string>
```

peternewman: It certainly sounds less wordy!

app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuDialog.kt
```diff
@@ -16,7 +16,7 @@ import kotlinx.android.synthetic.main.dialog_main_menu.view.*
 class MainMenuDialog(
     context: Context,
     onClickDownload: () -> Unit,
-    teamModeQuestFilter: TeamModeQuestFilter
+    val teamModeQuestFilter: TeamModeQuestFilter
```
FloEdelmann, line 19: Either I pass the quest filter in here, so that I can access its `enableTeamMode`/`disableTeamMode` functions and `isEnabled`/`indexInTeam` properties, or I pass them as separate parameters. Do you prefer callbacks and separate parameters here, too?

FloEdelmann (commented):
~I think all your comments should be resolved now.~ Oh, I just remembered that something is missing: Using the programmatically created color circles in the main menu button and dialog.

If it makes sense for me to wait until your branch is merged, so that we can avoid merge conflicts in `VisibleQuestsSource`, just let me know.

westnordost:
> If it makes sense for me to wait until your branch is merged, so that we can avoid merge conflicts in VisibleQuestsSource, just let me know.

Nah, my refactor work will probably take longer than this PR. The only thing I am concerned about is that it may not be received favourable to launch a feature that promotes to meet up in groups in mids of the pandemic. The monthly OSM "Stammtische", mapathons and other meet-ups are/should also not happening right now in real life. Surely, such feature could be used by flatmates or couples to have something to do on a walk.

app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuDialog.kt
westnordost: I have no preference here. Well... maybe slightly twards callbacks, because then the control-logic is all in one place (in MainMenuButtonFragment)

smichel17:
> The only thing I am concerned about is that it may not be received favourable to launch a feature that promotes to meet up in groups in midst of the pandemic.

On the other hand, going on a walk outdoors is one of the safest in-person activities you can do with someone outside your bubble during the pandemic. Plenty of ventillation, easy to maintain distance, and wearing a mask in the winter can even be nice, to keep your face warm :) I think it would be okay as long as the changelog included a note acknowledging the pandemic.

An alternative would be to merge the PR but comment out the pieces of the UI, so the feature cannot be accessed. Then you can enable the feature later, but avoid extra work resolving merge conflicts.

app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
```diff
@@ -69,9 +71,16 @@ import javax.inject.Singleton
         }
     }
 
+    private val teamModeChangeListener = object : TeamModeQuestFilter.TeamModeChangeListener {
+        override fun onTeamModeChanged(enabled: Boolean) {
+            // TODO: Not yet implemented
```
FloEdelmann, line 76: Here we have access to whether team mode is enabled or not, but how can we loop through all quests and update their visibility then?

FloEdelmann (commented):
I've rebased the PR onto the `master` branch and cleaned up / combined the commits to make them as easy to review as possible.

My question on how to proceed remains: In `VisibleQuestsSource` I have either access to whether team mode is enabled or not, or to which quests are visible. I don't know how/where to combine that information.

app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
westnordost: Right, this class doesn't know that. So that is why when the switch is flipped, the class should get all the visible quests from the two controllers by itself, sort them into invisible and visible quests and call the `onUpdatedVisibleQuests` listener with that info.
Yes, this may lead to already visible quests being added twice and already not-visible quests to be deleted from view twice. You need to check if the listener(s), mainly it should be the QuestPinLayerManager, are OK with that.

Hmm, but then of course,... the question is fro which bbox the visible quests should be retrieved from DB. Hmm.

I have had the same problem on my development branch, this is why I extended the listener like this:
```kotlin
    interface Listener {
        /** Called when given quests in the given group have been added/removed */
        fun onUpdatedVisibleQuests(added: Collection<Quest>, removed: Collection<Long>, group: QuestGroup)
        /** Called when something has changed which should trigger any listeners to update all */
        fun onInvalidated()
    }
```
... with an onInvalidated() method. The `QuestPinLayerManager` does this then:

```kotlin
    override fun onInvalidated() {
        clear()
        onNewScreenPosition()
    }
```

to re-fetch the new quests.

app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
```diff
@@ -38,12 +46,46 @@ class MainMenuButtonFragment : Fragment(R.layout.fragment_main_menu_button),
...
+    override fun onTeamModeChanged(enabled: Boolean) {
+        if (enabled) {
+            requireContext().toast(R.string.team_mode_enabled, Toast.LENGTH_LONG)
+            teamModeColorCircle.popIn()
+            TeamModeColorCircle.setViewColorsAndText(teamModeColorCircle, teamModeQuestFilter.indexInTeam)
+        } else {
+            requireContext().toast(R.string.team_mode_disabled, Toast.LENGTH_SHORT)
+            teamModeColorCircle.popOut()
+        }
     }
```
FloEdelmann, line 89: Should we invalidate the quests here when team mode gets disabled, as suggested in the original issue #2189?

If so, is `downloadedTilesDao.removeAll()` the right function to call? It doesn't seem start a new download immediately though.

FloEdelmann (commented):
Filtering quests now works :tada:

Thanks @westnordost for your guidance! I've got one more question, otherwise this is ready to review :blush:

app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
westnordost: No, that's not the right function to call. That makes the app forget that it downloaded anything, so it downloads again everything on the next opportunity.

On invalidation, the `QuestPinLayerManager` which is a listener to the `VisibleQuestsSource` should simply call `clear()` on itself (=removes all the quests + forgets that it retrieved from database any quests). This should next time the map is moved re-load all the quests from db, i.e. make `QuestPinLayerManager` to ask `VisibleQuestsSource` for quests.
Maybe to make this happen immediately and not just after the user panned the map, call `onNewScreenPosition()` after `clear()`.

FloEdelmann: > That makes the app forget that it downloaded anything, so it downloads again everything on the next opportunity.

That's the idea though. From #2189:

> After exiting the team modus again, those areas' quest cache should be invalidated automatically to make sure that the quests solved by the other team members won't be displayed the next time the user want to map there.

So we actually want to re-download the quests to make sure that other team members' solved quests don't show up after exiting team mode.

The changes in `QuestPinLayerManager` are already implemented in [`68d3a2c` (#2557)](https://github.com/streetcomplete/StreetComplete/pull/2557/commits/68d3a2cce0bf82ae2fc8a63371d5221acb595279).

westnordost: > That's the idea though. From #2189:

>>    After exiting the team modus again, those areas' quest cache should be invalidated automatically to make sure that the quests solved by the other team members won't be displayed the next time the user want to map there.

Ah right, makes sense. Hmm, but it only "invalidates" them, so they won't actually get updated after the next download. The `QuestAutoSyncer ` does the automatic downloading. So, that class should maybe also listen to changes on the team mode state. If the team mode got turned off, it should trigger an auto-download (after `downloadedTilesDao.removeAll()` has been called).

Helium314:
I did a short test of the team mode (team of 2), as it looks like it is almost finished.
Some comments:

1. When enabling team mode, the field for entering team size should have a blinking cursor (and the keyboard could pop up immediately, as it is needed anyway)

2. I don't like the changed position of the _settings_ entry in the menu, it's a bit of an unnecessary change in th UI. But this may just be personal taste, I guess after a few days/weeks I will get used to the layout

3. thank you! It works well!

westnordost:
I noticed you committed something here. I lost track - is it ready to merge? Is another review required?

RubenKelevra:
Not sure what's wrong with the animals as avatars... they looked really cute. Just just A, B, C, D is pretty boring tbh. 

Just my two cents. 

Thanks for the efforts implementing my ticket! 😁

westnordost:
> By the way, do we have a quest counter for the team mode? I feel like it would be cool to be able to compare the progress for each team member (on their phone) instead or showing the global counter. :)

If you add something like this later to a post, it will be missed by people who only read the updates via email. Anyway, as far as I know, Flo didn't implement anything like that. I agree that a feature like that might be in the spirit of that mode. But something like this could always be added later.

RubenKelevra:
> > By the way, do we have a quest counter for the team mode? I feel like it would be cool to be able to compare the progress for each team member (on their phone) instead or showing the global counter. :)
> 
> If you add something like this later to a post, it will be missed by people who only read the updates via email. Anyway, as far as I know, Flo didn't implement anything like that. I agree that a feature like that might be in the spirit of that mode. But something like this could always be added later.

I just tapped the "comment" button accidentally and didn't thought about the mail users. Next time I'll just create the full comment. 😁

But I agree that this is a nice-to-have and shouldn't stop this PR. 


@FloEdelmann I've posted the first pictures posted here as an appetizers on Twitter without realizing that the approach changed and those animals won't be used. My bad - sorry!

Anyway: There was a pretty good response to that news!

How about creating a poll there if they prefer the Animals or the Alphabetical Avatars? 🤔

westnordost:
Nah, no poll. The reasons for choosing colors over animals are:
- slightly better distinguishable. There are many brownish animal heads, some of these could not be told apart on first sight
- better communicable. "I take blue" is easer than "I take the light brown animal that looks like there are little crosses on his ears"
- better recognizable when displayed small (i.e. above the main menu button)
- consistent with other games: in boardgames, and computer games, the team is usually always defined by a color; which character/faction you choose is a different matter and has usually nothing to do with in which team you are but has other perks. This is why I mentioned that choosing avatars comes with other expectations connected with it that may lead to confusion
- you already got an OSM avatar. Confusion might arise in what way these two interact, if at all. Of course they have nothing to do with each other, but it may still lead to confusion. Users are easily confused.

The letters on the color was @FloEdelmann 's decision, I guess he added it so it is even more easy to communicate in which team you are. So instead of saying "I take aquamarine", you can say "I take the D" or something like that. I am neutral towards that and slightly lean towards not even having letters there, but just colors (that then must be very easy to tell apart). But this can all be changed later. In case we want to keep the labels, I'd also slightly lean towards numbers instead of letters because numbers are more international than latin letters, I'd even go so far that these are understood and used everywhere, even if the script used in that place has own numbers (such as in Arabia, India and South East Asia).

matkoniecz:
> The letters on the color was @FloEdelmann 's decision, I guess he added it so it is even more easy to communicate in which team you are. So instead of saying "I take aquamarine", you can say "I take the D" or something like that. I am neutral towards that and slightly lean towards not even having letters there, but just colors (that then must be very easy to tell apart).

letters may be useful for colorblind people (also, in some board games there are also shapes in addition to colors - see https://www.dangergames.co/blog/designing-games-for-colorblind-players/ )

screen06

peternewman:
> > The letters on the color was @FloEdelmann 's decision, I guess he added it so it is even more easy to communicate in which team you are. So instead of saying "I take aquamarine", you can say "I take the D" or something like that. I am neutral towards that and slightly lean towards not even having letters there, but just colors (that then must be very easy to tell apart).
> 
> letters may be useful for colorblind people (also, in some board games there are also shapes in addition to colors - see https://www.dangergames.co/blog/designing-games-for-colorblind-players/ )

Yeah we shouldn't be adding features which actively inhibit colour-blind users in this day and age! They will also deal with people's screens having different rendering/brightness/contrast especially in bright light, to ensure you can tell the potential 12 colours apart. But as you say letters, numbers or even basic symbols (e.g. square, triangle, circle, star etc) would do the job. I wonder if a screen reader would cope with the letters or numbers as is...

Also in terms of using just colours, see https://en.wikipedia.org/wiki/The_dress and https://colornames.org/ and similar, one person's aquamarine is another person's blue!

FloEdelmann:
Thanks for all your comments!

@westnordost Yes, this is ready for review. I'll try to test on real devices in the next few days.

@RubenKelevra Regarding the animal avatars: I also thought they were a cute idea to make the selection more playful. However, additional to the problems that @westnordost noted: I'd like to have a bit of variation between avatars, which is not that easy. This variation can be based the team size (but then you'd still get the same two avatars every time you're mapping with two people) and/or the time. You can see how I implemented that [here](https://github.com/streetcomplete/StreetComplete/commit/3cd9cb18cc0ac150529ee56786f721649df8238f#diff-cf4dee20bc508dfbf2d951317003aec9d6b303d34e8bad457b03c825ea3dc266R101-R107) (the `total` parameter is the team size there). Being based on time, it inherently has the problem of syncing: Two phones may have slightly different times set; or one person just happens to open the dialog right before a new selection is chosen, and the other one right after that. Then they end up with different sets of avatars.

Making the selection more boring with color circles also avoids all of these problems.

@RubenKelevra Regarding the team mode counter: I won't implement this in this PR. Let's first find out together if this works as expected; then it can later be improved. Good idea though!

Regarding the letters on the color circles: As @peternewman already said: There needs to be an additional way to distinguish the color circles for colorblind people, but also to mitigate differences in displayed colors. I don't really care whether it's latin letters, arabic digits, or other symbols. Coming up with 12 distinct shapes that still are simple and easy to recognize is actually quite hard though. Also, using letters or digits is much easier to code and scale (e.g. we might decide to increase the maximum team size to 15, then we'd need to come up with 3 more shapes).

RubenKelevra:
@FloEdelmann just use the day of the month for the avatar variation. It's unlikely that two people have two different dates in their phone. :)

peternewman:
> @FloEdelmann just use the day of the month for the avatar variation. It's unlikely that two people have two different dates in their phone. :)

Timezones and people like Australia/NZ may be a day ahead depending on how they're running their phone with local timezone or world clock and a location!

> @RubenKelevra Regarding the animal avatars: I also thought they were a cute idea to make the selection more playful. However, additional to the problems that @westnordost noted: I'd like to have a bit of variation between avatars, which is not that easy. This variation can be based the team size (but then you'd still get the same two avatars every time you're mapping with two people) and/or the time.

I would have thought the variation might be a disadvantage personally. E.g. when playing a colour based board game, or a counter based one (e.g. Monopoly), in a regular group of players certain people are always likely to take certain choices so it almost becomes a given, which removes choosing time and also means you don't get confused when who you are (although obviously the latter bit isn't relevant to SC once you've made your choice). If that changes each time you've got to pick and communicate that pick each time.

westnordost:
@FloEdelmann: I am currently reviewing it and making small changes to it. If you don't mind, I'd also change from letters to numbers because I think it is slightly better. Ok?

westnordost:
Oh, I get now why you chose letters over numbers. A number is confusing when shown as a little colored circle on top of the main menu button

westnordost (approved):

RubenKelevra:
> > @FloEdelmann just use the day of the month for the avatar variation. It's unlikely that two people have two different dates in their phone. :)
> 
> Timezones and people like Australia/NZ may be a day ahead depending on how they're running their phone with local timezone or world clock and a location!


Just ignore the timezones. UTC should be everywhere the same.

If this isn't fault proof enough I would recommend asking the OSM API for the current time. I think HTTP is still posting the current time on each response, no? 🤔

peternewman:
> Just ignore the timezones. UTC should be everywhere the same.

Yeah fair point, although doesn't deal with the IMHO other downsides.

> If this isn't fault proof enough I would recommend asking the OSM API for the current time. I think HTTP is still posting the current time on each response, no? thinking

That doesn't cover the users who use it offline and just sync via Wi-Fi occasionally/when they return home, which is effectively something the whole of the rest of this PR enables, otherwise you could just keep uploading and rescanning I think.

app/src/main/res/values/strings.xml
FloEdelmann: @westnordost Why did you change this back to the string I had initially?

app/src/main/java/de/westnordost/streetcomplete/controls/TeamModeColorCircleView.kt
```diff
@@ -0,0 +1,54 @@
...
+import de.westnordost.streetcomplete.R
+import kotlinx.android.synthetic.main.view_team_mode_color_circle.view.*
+
+class TeamModeColorCircleView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : ConstraintLayout(context, attrs, defStyleAttr) {
+
+    fun setTeamIndex(index: Int) {
```
FloEdelmann, line 17: Shouldn't this be `setIndexInTeam` again to avoid confusion?

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
```diff
@@ -59,4 +59,8 @@ import javax.inject.Singleton
     fun removeListener(listener: TeamModeChangeListener) {
         listeners.remove(listener)
     }
+
+    companion object {
+        const val MAX_TEAM_SIZE = 12
```
FloEdelmann, line 64: I thought you didn't want to hardcode this but return the colors array size instead?

FloEdelmann (commented):
Thanks @westnordost for your changes! I got a few post-merge questions though.

app/src/main/java/de/westnordost/streetcomplete/controls/TeamModeColorCircleView.kt
westnordost: Hm, right

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
westnordost: Well, now that the Colors object is gone (converted into view), it seemed wrong to me to have this information sourced from a view class. So I thought I put it where the TeamMode logic has its core.

app/src/main/res/values/strings.xml
westnordost: Oh, it's just that it appeared to be better in my ears, I didn't remember this discussion

RubenKelevra:
> > Just ignore the timezones. UTC should be everywhere the same.
> 
> Yeah fair point, although doesn't deal with the IMHO other downsides.
> 
> > If this isn't fault proof enough I would recommend asking the OSM API for the current time. I think HTTP is still posting the current time on each response, no? thinking
> 
> That doesn't cover the users who use it offline and just sync via Wi-Fi occasionally/when they return home, which is effectively something the whole of the rest of this PR enables, otherwise you could just keep uploading and rescanning I think.

I just want the cute Animals back 😬

I mean we don't change the colors either. So two people has to be red and blue too all the time, right? 🤔

The A, B, C, D is just outright boring.

If you think daily changes are bad, just do it weekly. Hard to belive that people go team mapping on Sunday nights with phones having clocks completely off. 😬

app/src/main/res/values/strings.xml
FloEdelmann: Fixed in #2613.

app/src/main/java/de/westnordost/streetcomplete/controls/TeamModeColorCircleView.kt
FloEdelmann: Fixed in #2613.

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/TeamModeQuestFilter.kt
FloEdelmann: Fixed in #2613.

PR #2567, opened 2021-02-11
Check parking tickets every four time periods too

peternewman:
Around my area, a lot of parking is now changing to pay by phone so machines are no longer required, and are presumably being removed/not replaced.

They were previously not being checked at all.

I note fuel vending machines aren't being checked at all either.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
@@ -34,7 +34,7 @@ class CheckExistence(
             or amenity = post_box
             or leisure = picnic_table
             or leisure = firepit
-            or amenity = vending_machine and vending = public_transport_tickets
+            or amenity = vending_machine and vending ~ parking_tickets|public_transport_tickets
             or tourism = information and information ~ board|terminal|map
             or advertising ~ column|board|poster_box
           )


matkoniecz (approved):
I have tested it and it seems both good idea and working code

Not merging just in case it was intentional decision done during resurvey implementation

screen10

peternewman:
Great thanks @matkoniecz .

Any feeling on fuel vending machines from your side, which like parking machines were, is effectively excluded everywhere.

matkoniecz:
For me it seems to make sense to ask about all vending machines, but maybe it should be reduced how often it happens...

It should be just checked that type of them will be recognized (included in iD presets).

westnordost:
I think I did forget parking tickets vending machines, thanks!

Though, I did not forget the fuel vending machines. Usually, those are part of a fuel station, so whether they are still there or not is connected with the whole fuel station being present or not. So the quest shouldn't ask about it cause that would only update the data partially.

peternewman:
Makes sense @matkoniecz 

> I think I did forget parking tickets vending machines, thanks!

No worries.

> Though, I did not forget the fuel vending machines. Usually, those are part of a fuel station, so whether they are still there or not is connected with the whole fuel station being present or not. So the quest shouldn't ask about it cause that would only update the data partially.

Fair enough, makes sense.

matkoniecz:
Opened #2570 for packstations.

PR #2576, opened 2021-02-15
Add postbox royal cypher quest

### LINKED ISSUE #2563: Suggested Quest: Royal Cypher for post boxes
### General

On items with tag [Tag:amenity=post_box] 

Note: applies to UK only (could be applied to some Commonwealth territories if requested by local mappers)

Affected tag(s) to be modified/added: [Tag:royal_cypher=*] ( https://wiki.openstreetmap.org/wiki/Key:royal_cypher )

Question asked: **What is the royal cypher on this post box**

### Checklist
- [x ] 🚧 To be added tag is established and has a useful purpose
- [x ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x ] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x ] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

A set of radio buttons with the values as listed at https://wiki.openstreetmap.org/wiki/Key:royal_cypher from which only one, or "no value", may be selected. Can use thumbnail mages fro that age if desired. Since "EIIR" is the commonest value it should be prominent. "Scottish Crown" will be the commonest value in Scotland.

Note: useful purpose is to aid enthusiasts, conservationists, and photographers looking for boxes with older or rare cyphers.

arrival-spring:
Fixes #2563 

I read what was said in the issue about the images, and to an extent I agree, but then it was a bit tricky to find decent svgs for all of the cyphers, that also looked like what was on postboxes, so I decided that having real examples was probably better.

Also I've just used the post icon because I wasn't sure what would represent it better. But maybe something could be added to make it distinct?

First time contributing, so I might have missed things or made mistakes, but it all seems to be working as expected for me.


app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/PostboxRoyalCypher.kt
```diff
@@ -0,0 +1,12 @@
+package de.westnordost.streetcomplete.quests.postbox_royal_cypher
+
+enum class PostboxRoyalCypher(val osmValue: String, val wikidataValue: String) {
+    VICTORIA("VR", "Q33102356"),
+    EDWARDVII("EVIIR", "Q33102324"),
+    GEORGEV("GR", "Q33102273"),
+    EDWARDVIII("EVIIIR", "Q33102195"),
+    GEORGEVI("GVIR", "Q33102169"),
+    ELIZABETHII("EIIR","Q33102113"),
+    SCOTTISHCROWN("scottish_crown", ""),
```
westnordost, line 10: Why tagging the wikidata at all? Isn't or shouldn't the wikidata be assigned to that OSM tag already? That seems like duplication.

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/AddPostboxRoyalCypher.kt
```diff
@@ -0,0 +1,33 @@
...
+class AddPostboxRoyalCypher : OsmFilterQuestType<PostboxRoyalCypher>() {
+
+    override val elementFilter = """
+        nodes, ways, relations with
+          amenity = post_box
+          and !royal_cypher
+    """
+    override val commitMessage = "Add postbox royal cypher"
+    override val wikiLink = "Key:royal_cypher"
+    override val icon = R.drawable.ic_quest_mail
```
westnordost, line 18: Why, a crown of course!

westnordost (changes requested):
Technically, all fine implemented!

Regarding the pictures: Well, not that happy about it. Because, 
- of the reasons mentioned already
- they are cropped to different sizes
- they are in the wrong directory - for raster graphics, there must be a version for xxhdpi (384x384), a xhdpi (256x256), a hdpi (192x192) and a mpdi (128x128) version of each picture. This is always a hassle to do, which is why I use vector graphics whenever possible

In Inkscape, it is possible to trace bitmaps to a vector graphic:
https://inkscape.org/doc/tutorials/tracing/tutorial-tracing.html
So one just needs to find pictures from which they are clearly tracable, ... I found this website
https://www.postalmuseum.org/blog/royal-cypher-appearances/
(Maybe it is easier without the crown)

westnordost:
There is also https://commons.wikimedia.org/wiki/Category:Royal_monograms_of_the_United_Kingdom but most of them are ridiculuously large (file size), so I think a simple monochrome SVG would be better

matkoniecz:
> they are in the wrong directory - for raster graphics, there must be a version for xxhdpi (384x384), a xhdpi (256x256), a hdpi (192x192) and a mpdi (128x128) version of each picture. This is always a hassle to do, which is why I use vector graphics whenever possible

I wrote tiny script to solve this https://github.com/matkoniecz/rescaling_for_android

Relies on imagemagick that may be Linux only.

westnordost:
I created a quest icon for you to use. It is a simplified version of the Elizabeth II crown. It is `ic_quest_crown` on the master branch:

crown

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -722,4 +722,5 @@ object AchievementsModule {
     // maybe "emergency"
     // AddFireHydrantType AddIsDefibrillatorIndoor
     //
+    // AddPostboxRoyalCypher
```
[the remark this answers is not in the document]
westnordost: And maybe also to "rare"

westnordost:
**EVIIR** make this monochrome, done: https://commons.wikimedia.org/wiki/File:Royal_Monogram_of_King_Edward_VII_of_Great_Britain_goldenrod.svg
**VR** https://commons.wikimedia.org/wiki/File:Royal_Cypher_of_Queen_Victoria.svg
**EIIR** can be taken from https://commons.wikimedia.org/wiki/File:Royal_Cypher_of_Queen_Elizabeth_II.svg
**GVIR** good source to trace from is this https://de.m.wikipedia.org/wiki/Datei:MonogramGeorgeVI.jpg
The remaining ones can I think pretty well be traced from http://www.bechler.free-online.co.uk/gc/RoyalCyphers2.jpg

app/src/main/res/values/strings.xml
```diff
@@ -564,6 +564,15 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
     <string name="quest_collectionTimes_add_times">Add collection time</string>
     <string name="quest_collectionTimes_answer_no_times_specified">No times specified</string>
+    <string name="quest_postboxRoyalCypher_title">What is the royal cypher on this postbox?</string>
+    <string name="quest_postboxRoyalCypher_type_victoria">Victoria (1837 to 1901)</string>
+    <string name="quest_postboxRoyalCypher_type_edwardvii">Edward VII (1901 to 1910)</string>
+    <string name="quest_postboxRoyalCypher_type_georgev">George V (1910 to 1936)</string>
+    <string name="quest_postboxRoyalCypher_type_edwardviii">Edward VIII (1936)</string>
+    <string name="quest_postboxRoyalCypher_type_georgevi">George VI (1936 to 1952)</string>
+    <string name="quest_postboxRoyalCypher_type_elizabethii">Elizabeth II (1952 to present)</string>
+    <string name="quest_postboxRoyalCypher_type_scottish_crown">The Crown of Scotland</string>
```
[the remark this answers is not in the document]
arrival-spring: I did think that a bit when I was adding the labels in.
One thought the other way though, if a form of this image was the option to select: https://commons.wikimedia.org/wiki/File:Royal_Cypher_of_Queen_Elizabeth_II.svg
Would the user tag all of these as that?
https://www.geograph.org.uk/photo/6269863
https://www.geograph.org.uk/photo/4895153
https://www.geograph.org.uk/photo/1135599

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/PostboxRoyalCypher.kt
arrival-spring: I guess so, it was there on the wiki so I assumed it was something that might be helpful to have tagged.
Surely something similar could be said about a lot of wikidata tags (e.g. for brands), but then I don't really know what uses the wikidata tags.

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
westnordost: I think so, the cypher is pretty unambiguous.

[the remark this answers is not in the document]
arrival-spring: I know they're pretty similar, but how different is too different?
Maybe a description line to say something like "Select the one that most closely matches"?

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/PostboxRoyalCypher.kt
westnordost: Ok then lets remove the wikidata tagging.

arrival-spring:
Thanks for all of the feedback here, I clearly hadn't looked hard enough for images.
I think I've made all of those changes now.
I'm not very happy with the Scottish Crown option, because it just looks like a crown and most postboxes will have a crown in addition to the initials. Maybe that one does need a description?

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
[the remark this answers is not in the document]
arrival-spring: Added to both

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/PostboxRoyalCypher.kt
arrival-spring: Have done

westnordost:
Hm, something went wrong with you pulling from master. Did you rebase instead of merge or something? Anway, this PR now also includes changes already made on master. And also still some leftovers from the JPEGs.
The SVG graphics used to generate the android drawables are also not included. Surely, you needed to add/modify the SVGs I linked to make them monochrome, to resize them to 128x128 or what the correct size was, to create the trace from the pictures etc.

Regarding the Scottish Crown: Well, even if it is labelled "Scottish Crown", it will not be so helpful because not everybody will know how exactly that crown looks. Maybe instead "Just a crown" as a label.

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -722,4 +724,5 @@ object AchievementsModule {
...
```
westnordost, line 727:
```suggestion
```

westnordost:
#2583

PR #2578, opened 2021-02-15
Add detail to changeset comment for traffic signals sound and vibration quests

### LINKED ISSUE #2577: Add detail to crossing sound signals and vibration changeset message
When the traffic signals sound and traffic signals vibration quests are answered the comment on the changeset is 
"Add traffic_signals:vibration tag", e.g. https://www.openstreetmap.org/changeset/99294493
Or "Add traffic_signals:sound tag", e.g. https://www.openstreetmap.org/changeset/99263090

All other quests (I think) have more meaningful comments, e.g. "Add whether pedestrian crossing has an island" (https://www.openstreetmap.org/changeset/99305683) etc.

(All changesets are just random recent ones I found through OSMCha).

How about "Add whether traffic signals have sound signals" and "Add whether traffic signals have tactile indication that it's safe to cross".

I'm up for fixing this (easy change, but I want to contribute).

Also related, is there a reason for changeset comments not being localised? If it's an enhancement that's wanted then I'll open a different issue for it.

arrival-spring:
Closes #2577 

Add more meaningful changeset comments when sound and vibration quests are solved.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_sound/AddTrafficSignalsSound.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_sound/AddTrafficSignalsSound.kt
@@ -18,7 +18,7 @@ class AddTrafficSignalsSound : OsmFilterQuestType<Boolean>() {
         )
     """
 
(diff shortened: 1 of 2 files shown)


westnordost:
makes sense, thanks!

PR #2584, opened 2021-02-16
Small refactoring / cleanup

FloEdelmann:
Extracted from #2557 to make the diff more manageable.

* Fix Android Studio warnings in main menu dialog layout
	- use `app:drawableXCompat` instead of `android:drawableX`
	- rearrange attributes
* Fix disabled dialog button style
* Extract main menu button into separate fragment

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
@@ -1,0 +1,104 @@
+package de.westnordost.streetcomplete.controls
+
+import android.net.ConnectivityManager
+import android.os.Bundle
+import android.view.View
+import android.widget.Toast
+import androidx.appcompat.app.AlertDialog
+import androidx.core.content.getSystemService
+import androidx.fragment.app.Fragment
+import de.westnordost.osmapi.map.data.BoundingBox
+import de.westnordost.streetcomplete.ApplicationConstants
+import de.westnordost.streetcomplete.Injector
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.download.DownloadController
+import de.westnordost.streetcomplete.ktx.toast
+import de.westnordost.streetcomplete.map.MainFragment
+import de.westnordost.streetcomplete.util.area
+import de.westnordost.streetcomplete.util.asBoundingBoxOfEnclosingTiles
+import de.westnordost.streetcomplete.util.enclosingBoundingBox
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private fun isConnected(): Boolean {
+        val connectivityManager = context?.getSystemService<ConnectivityManager>()
+        val activeNetworkInfo = connectivityManager?.activeNetworkInfo
+        return activeNetworkInfo != null && activeNetworkInfo.isConnected
+    }
+
+    private fun downloadDisplayedArea() {
+        val displayArea = (requireParentFragment() as MainFragment).mapFragment?.getDisplayedArea()
```
westnordost, line 65: This strongly couples this fragment to the MainFragment.

The ususal pattern used in this app for fragments to communicate with their parent is to declare 
```kotlin
    interface Listener {
        fun getDownloadArea(): BoundingBox
    }
    private val listener: Listener? get() = parentFragment as? Listener ?: activity as? Listener
```
, then change the code here to
```kotlin
val displayArea = listener?.getDownloadArea() ?: return
```

This way, the parent could be an activity or fragment, anything that implements that interface.

app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+    }
+
+    private fun downloadAreaConfirmed(bbox: BoundingBox) {
+        var bbox = bbox
+        val areaInSqKm = bbox.area() / 1000000
+        // below a certain threshold, it does not make sense to download, so let's enlarge it
+        if (areaInSqKm < ApplicationConstants.MIN_DOWNLOADABLE_AREA_IN_SQKM) {
+            val cameraPosition = (requireParentFragment() as MainFragment).mapFragment?.cameraPosition
```
westnordost, line 96: same here

westnordost (changes requested):
Very cool, thanks for splitting it up!

app/src/main/java/de/westnordost/streetcomplete/controls/MainMenuButtonFragment.kt
westnordost: Maybe that part and the deciding if it is too large or too small should go into whoever implements the `getDownloadArea`, what do you think?

westnordost (approved):

westnordost:
👍

PR #2591, opened 2021-02-17
Repeat last answer for sidewalk/bicycle lane

### LINKED ISSUE #2542: Repeat last answer for bicycle lane quests
**Use case**
On some roads, there are really many bicycle lane quests. I walked a street about 200meters long and answered 8 times: "no, there is no bicycle lane", "no, again not", "still not", "nope, nothing changed", "no, still the same road without any bicycle lane", ... (and that for both sides). That is about 6 taps for 1 quest.

**Proposed Solution**
- Most simple solution would be "repeat my last answer".
- A more sophisticated solution might be to predict (if there is no data available) the bicycle lane situation for the next quests. If I answered "no bicycle lanes here on both sides" on the beginning of a street, it might be sane to suggest "no bicycle lanes on both sides" for the rest of the road (at least I could just tap OK then). If anything changes, I can still change my answer for the next bit of the street.

arrival-spring:
fixes #2542 

Here's a first attempt at adding a 'repeat last answer' button for quests about sides of the road.

So far I've just added it for the sidewalks quest as that seemed like a simple place to start, because it only has yes or no options for each side of the road.

Before pressing the button:
image
After pressing the button:
image

Done:
The button shows up only if you have answered the quest in this session and tapping it puts the values into the form.
Tapping it then swaps the button over so that a second tap would swap the sides of the road over (but only if there's a different answer for each side).
The button rotates with the map (the same as the compass and main puzzle bit do)

Still to do:
Make the button look better, it's not as clear as it could be right now because it's sort of too wide for the button. I could do with some help here.
Move the button (possibly only once it's been pressed) so that it doesn't overlap the OK button, again could probably do with some help/pointers here.

Once this seems ok I'll then look at the bicycle lane quest (they use the same layout so at the moment a blank button shows up). But it will use basically the same code so actually I'll look to separate the code out so they can both use it (and any other quests now or in the future, as appropriate).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/StreetSideRotater.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/StreetSideRotater.kt
@@ -11,7 +11,8 @@ import de.westnordost.streetcomplete.view.StreetRotateable
 class StreetSideRotater(
     private val puzzle: StreetRotateable,
     private val compassView: View,
-    geometry: ElementPolylinesGeometry
+    geometry: ElementPolylinesGeometry,
+    private val lastAnswerButton: StreetRotateable? = null
 ) {
     private val wayOrientationAtCenter = geometry.getOrientationAtCenterLineInDegrees()
     private val uiHandler = Handler(Looper.getMainLooper())
@@ -28,6 +29,7 @@ class StreetSideRotater(
         puzzle.setStreetRotation(wayOrientationAtCenter + rotation.toDegrees())
         compassView.rotation = rotation.toDegrees()
         compassView.rotationX = tilt.toDegrees()
+        lastAnswerButton?.setStreetRotation(wayOrientationAtCenter + rotation.toDegrees())
     }
 
     private fun Float.toDegrees() = (180 * this / Math.PI).toFloat()
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalkForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalkForm.kt
@@ -26,7 +26,6 @@ class AddSidewalkForm : AbstractQuestFormAnswerFragment<SidewalkAnswer>() {
     override val contentPadding = false
 
     private var streetSideRotater: StreetSideRotater? = null
-    private var rotateTheButton: StreetSideRotater? = null
     private var leftSide: Sidewalk? = null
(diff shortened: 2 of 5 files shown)


westnordost:
Wow, man, you are productive! 

I haven't looked at the code yet, let me just make a few notes I made for myself (in my head) for when I'd implement this:

1. Into the store that is used to store the last picked answer, it is necessary to not only store that last answer but also the orientation of the road at the location of the pin (=the rotation at which the preview was displayed). This information can then be used to switch the left and right side for the next way if the rotation of that way differs of more than -/+ 90°. Basically, this implementation is to prvent what seems to be happening in your second screenshot(?) - that a sidewalk continues on the same side, but the OSM-way is reversed.

2. It seems the prerequisite for a button that is small enough to recognized as one would be to make the roadway-part of the used illustration smaller. If all else fails, for every street side illustration graphic used, we need another one with less roadway visible. From the screenshot, it looks like you are using the StreetSidePuzzle for the small illustration, too. That class has a lot of complexity that is not necessary for a static button like this - animation, replacing the images, scaling, repating the graphic along the Y-axis etc. I think it may be the easiest to create another View class that does exactly what we want for this button: Just display a graphic on the left and on the right side; and probably: show only the right 50% of that graphic, cut the rest. On the other hand, I just looked at the cycleway graphics and some display the cycleway at more than 50% of that graphic. So maybe this solution: Don't rotate the street shown in that button at all, instead that button will have a ratio of 2:1 (or maybe 3:2 if the graphic is cut a little) and is shown always upright. If the street rotation (by rotating the map) exceeds -/+ 90°, what's on the button will be switched left/right.

arrival-spring:
1. Ah yes, that makes sense, I'll do that then.
I should have put captions (I've now edited them in), the first screenshot is before the button is pressed (remembering the last answer) and the second screenshot is after the button has been pressed (it swaps the sides over).

2. Yes, I did just use the StreetSidePuzzle for now for the button, so will go and make a new layout for this button, as that makes sense.

westnordost:
It looks like this PR is abandoned. @arrival-spring , do you still intend to finish this PR?

arrival-spring:
I had intended to finish it, but I think I got stuck at around this point and have since been busy with other things so realistically I don't think I'm going to finish it, sorry about that.

westnordost:
Ok no problem. I have looked at the code so far and I think it would be easier for someone wanting to do this to start anew (maybe even for you, as the code is half a year old) than to try to continue from the current code, so I'll close this.

Issue #2512, opened 2021-01-17
improve quest "Check if element still exists" for newspaper vending machines


Strubbl:
**Use case**
I want to suggest an improvement for the quest "Check if element still exists" for newspaper vending machines (NVM). There are NVM which have a name tag. Those ones display that name during the survey question. This is very helpful if there are two or more NVM next to each other.

But there are also NVM which do not have a name, but a brand tag. It would be nice to see the brand value in the survey question, esp. when there is no name tag of that node.

To show you what i mean, this is a NVM with a name tag:
https://www.openstreetmap.org/node/4192397931

And this one has no name, but a brand:
https://www.openstreetmap.org/node/4253350433

This is an example place, where you have three NVM next to each other:
https://www.openstreetmap.org/node/4253350433
https://www.openstreetmap.org/node/4253350432
https://www.openstreetmap.org/node/4253350431

**Proposed Solution**
Add the value of the brand tag to the question in case it exists and no name tag exists.

I fear there might be an impact on other survey questions of that "Check if element still exists" type I do not if that is a real problem and if yes, i do not know how to address this problem yet. Maybe make a separate quest only for NVM?

westnordost:
Right, this was kind of missing

peternewman:
It might also be worth falling back to operator and possibly network too.

PR #2594, opened 2021-02-19
Minor improvements for quests that display OSM element's name

### LINKED ISSUE #2593: Show name for charging stations
Often (37% according to taginfo) charging stations have a `name` tag.
I think it would be beneficial if SteetComplete would show this name when asking for the number of cars that can be charged.

I already found some cases where no charging station was visible (and answering 'does not exist' (I think) would delete the node)
But the charging stations actually did exist in a nearby parking garage, and simply were not tagged with `level=-1` or similar.
All of those charging stations had a name indicating they belong to a garage, so by showing the name, wrong deletions might be prevented.

FloEdelmann:
* Add `T?.asSingleArray()` and `Map.containsAnyKey(...keys)` extension functions
  → to make the code more concise
* Fix `brand` tag being ignored in `CheckExistence` quest
  → It was added in #2512 / 245fffe863f874ae5041f8236acb6dc041e3610e but only to the `getTitleArgs` function, not to the `getTitle` function
* Fall back to operator in `CheckExistence` quest
  → as suggested by @peternewman in https://github.com/streetcomplete/StreetComplete/issues/2512#issuecomment-764587826
* Show name/brand/operator for charging station quests
  → Closes #2593. I've also taken the `brand` and `operator` tags into account, since those are more widespread than `name` in my area.
* Lint `getTitle` function in all quests
  → to make the code more concise and consistent


app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
```diff
@@ -6,14 +6,13 @@ import de.westnordost.streetcomplete.data.quest.QuestType
...
 
 /** Quest type where each quest refers to an OSM element */
 interface OsmElementQuestType<T> : QuestType<T> {
 
-    fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
-        val name = tags["name"] ?: tags["brand"]
-        return if (name != null) arrayOf(name) else arrayOf()
-    }
+    fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> =
+        (tags["name"] ?: tags["brand"]).asSingleArray()
```
FloEdelmann, line 15: Should we include `tags["operator"]` as a third fallback here?

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
```diff
@@ -63,7 +63,7 @@ class AddAcceptsCash(
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
         val name = tags["name"] ?: tags["brand"]
```
FloEdelmann, line 65: Is `tags["brand"]` correct here? A few lines above, it explicitly excluded.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -88,12 +88,14 @@ class AddWheelchairAccessBusiness(
...
-        else                      R.string.quest_wheelchairAccess_name_title
+        if (hasFeatureName(tags))
+            R.string.quest_wheelchairAccess_name_type_title
+        else
+            R.string.quest_wheelchairAccess_name_title
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
         val name = tags["name"] ?: tags["brand"]
-        return if (name != null) arrayOf(name,featureName.value.toString()) else arrayOf()
+        return if (name != null) arrayOf(name, featureName.value.toString()) else arrayOf()
```
FloEdelmann, line 98: `name` is explicitly included in the element filter above. Should we drop `tags["brand"]` here, or change the filter to `(name or brand)`? Also the condition can be dropped, right?

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
peternewman: That (and network) would seem easier to me than trying to guess and keep up to date with what is used where. I assume it will just return the first one?

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
peternewman: Is that because if the hasFeatureName matched a brand (e.g. McDonalds), you don't want it to say:
"Does %1$s (%2$s) accept cash payment?"
"Does McDonalds (McDonalds) accept cash payment?"

But this would be acceptable:
"Does McDonalds (Fast Food Restaurant) accept cash payment?"

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
FloEdelmann: Yes, `tags["name"]` returns either a `String` or `null`, and the `?:` operator evaluates to the right side if the left is `null`.

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
FloEdelmann: But that's not what would happen:

Given the following tags:

```
amenity=fast_food
name=McDonald's
brand=McDonald's
```

* `hasFeatureName(tags)` returns `true`
* `tags.containsKey("brand")` returns `true`
* So, `R.string.quest_accepts_cash_title` ("Does %s accept cash payment?") is used
* `getTitleArgs` returns `arrayOf("McDonald's", "Fast Food Restaurant")`
* All in all, this results in "Does *McDonald's* accept cash payment?"

peternewman: Ah yes, sorry I thought hasFeatureName included the NSI presets so would match on the brand as well.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
westnordost: No, we shan't because the operator could be "Max Mustermann". Only for certain elements it makes sense to display that as a "name" (f.e. for ATMs it may make sense)

app/src/main/java/de/westnordost/streetcomplete/ktx/Collections.kt
```diff
@@ -73,3 +73,6 @@ inline fun <T> Iterable<T>.sumByFloat(selector: (T) -> Float): Float {
 
 fun <T> Collection<T>.containsExactlyInAnyOrder(other: Collection<T>): Boolean =
     other.size == size && containsAll(other)
+
+/** If null, returns an empty array. Otherwise, returns an array with the value as the only element */
+inline fun <reified T> T?.asSingleArray(): Array<T> = if (this == null) arrayOf() else arrayOf(this)
```
westnordost, line 78: Hm. There is a reason why `getTitleArgs` returns an array and not a `String?` - because there may be several args.

I would very much prefer a `arrayOfNotNull` function instead. There is already a `listOfNotNull` function, so this would be consistent.

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingBusStop.kt
```diff
@@ -31,12 +31,11 @@ class AddTactilePavingBusStop : OsmFilterQuestType<Boolean>() {
...
-            if (hasName) R.string.quest_tactilePaving_title_name_tram
-            else         R.string.quest_tactilePaving_title_tram
-        } else {
-            if (hasName) R.string.quest_tactilePaving_title_name_bus
-            else         R.string.quest_tactilePaving_title_bus
+        return when {
+            isTram && hasName ->    R.string.quest_tactilePaving_title_name_tram
+            isTram ->               R.string.quest_tactilePaving_title_tram
+            hasName ->              R.string.quest_tactilePaving_title_name_bus
+            else ->                 R.string.quest_tactilePaving_title_bus
```
westnordost, line 38: Matter of taste I guess (the other as well). I slightly prefer the current version but I am fine with your version too

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
westnordost: Right, it could be `(name or brand)`

app/src/main/java/de/westnordost/streetcomplete/quests/charging_station_operator/AddChargingStationOperator.kt
```diff
@@ -13,7 +13,11 @@ class AddChargingStationOperator : OsmFilterQuestType<String>() {
     override val icon = R.drawable.ic_quest_car_charger
     override val isDeleteElementEnabled = true
 
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_charging_station_operator_title
+    override fun getTitle(tags: Map<String, String>): Int =
+        if (tags.containsAnyKey("name", "brand"))
+            R.string.quest_charging_station_name_operator_title
+        else
+            R.string.quest_charging_station_operator_title
```
westnordost, line 20: This quest is not asked for charging station that have a name, brand or operator

westnordost: and also should not be, of course

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
westnordost: Right, there is a mix-up.

In some places 
```kotlin
    private fun hasFeatureName(tags: Map<String, String>?): Boolean =
        tags?.let { featureDictionaryFuture.get().byTags(it).isSuggestion(false).find().isNotEmpty() } ?: false
```
is used to find out if a feature has a "real" preset name.
In other places, a function with the same name is used to find if a feature can be named at all somehow.

I'd be grateful if you cleaned that up. I think in most cases (for the quest title string), what we actually want is the feature name, not the brand/NSI-name. For a few quests, we need "any" name.

app/src/main/java/de/westnordost/streetcomplete/quests/charging_station_operator/AddChargingStationOperator.kt
peternewman: Is it not the case that they could have a brand or operator and need a name set (e.g. at a service station)?

For example this company seem to name them (effectively based on their location), while the brand/operator would be Pod Point:
https://charge.pod-point.com/

Maybe this wants to be a separate issue to discuss it?

westnordost: It could, but any of the three is usually enough to identify what kind of charging station it is, i.e. derive what socket types it supports etc.
As with ATMs, it's not that well standardized what exactly should go into brand, what into operator and what into name.

peternewman: Yeah it would be fine for the how many sockets quest. I was referring to this comment of yours:

> This quest is not asked for charging station that have a name, brand or operator

Given this is the AddChargingStationOperator quest we're discussing it in.

> As with ATMs, it's not that well standardized what exactly should go into brand, what into operator, what into network and what into name.

NSI is working hard to improve on that...

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
peternewman: I was suggesting as the THIRD option, i.e. to fall back on if there is no name or brand.

All of these things in NSI are operator based:
https://nsi.guide/index.html?t=operators

Some, such as a lot of charging stations, don't have a brand set:
https://nsi.guide/index.html?t=operators&k=amenity&v=charging_station

peternewman: Or does this mean it's a choice between "Max Mustermann" or ATM? I thought from the other code if it had a name-like field and a feature type you'd get "Does this `ATM` (`Max Mustermann`) exist still?"

westnordost:
If you can finish this today, the time frame would be right to include it in the next major release. (PRs that include new strings are always only merged into the next major release and not into a minor releaes to give translators enough time to translate)

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt
westnordost: I already solved this in the frame of a bug I just fixed! I hope you did not start working on it!

westnordost: Oh no, I didn't. But I can do that after merge if that is the last thing that is missing for htis PR

westnordost (approved):

westnordost:
Thank you, @FloEdelmann ! :-)

Issue #1988, opened 2020-07-30
collection_times value incorrectly formatted


JacLavi:
This the way SC formats collection_times for a postbox

collection_times=Mo-Fr 16:30, Sa 12:00

https://www.openstreetmap.org/node/1395865959

which is reported as wrong by Osmose according to the Wiki : 

https://wiki.openstreetmap.org/wiki/Key:collection_times

and should be : 

collection_times=Mo-Fr 16:30; Sa 12:00  (semi-colon)

**How to Reproduce**

**Versions affected**  :

SC 10.2 

Android 9

matkoniecz:
both version are correct, Osmose is wrong to complain https://openingh.ypid.de/evaluation_tool/?EXP=Mo-Fr%2016%3A30%2C%20Sa%2012%3A00&lat=48.7769&lon=9.1844&mode=1&DATE=1596111000000 and https://wiki.openstreetmap.org/wiki/Key:opening_hours

Issue #1786, opened 2020-04-23
Postbox Collection Time Separator (and possibly opening hours too)


peternewman:
The collection_times separator used by StreetComplete is a comma, e.g. `Mo-Fr 17:00, Sa 12:45`.

However everything seems to favour semicolon then space between days (including @rjw62 's reporting page, which initially flagged this up to me):
https://wiki.openstreetmap.org/wiki/Key:collection_times#Examples
https://wiki.openstreetmap.org/wiki/Key:opening_hours#Simple_examples
https://osm.mathmos.net/postboxes/osm-collection-times.html#details
https://projets.pavie.info/yohours/?oh=Mo-Fr%2009:00;%20Sa%2007:00 - Works
https://projets.pavie.info/yohours/?oh=Mo-Fr%2009:00,%20Sa%2007:00 - Doesn't work

Although the summary syntax says it would be two rulesets separated by a semicolon:
https://wiki.openstreetmap.org/wiki/Key:opening_hours#Summary_syntax

The supposed grammar has additional_rule_separator:
https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#Syntax_diagram

And seems to imply either would work:
https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:additional_rule_separator

> An additional rule is treated exactly the same as a normal rule, except that an additional rule does not overwrite the day for which it applies (unlike the <normal_rule_separator> which starts always with a new, empty day, deleting any previous rules applying for the given day).

So clear as mud!

However in terms of normal usage, I added stats to my previous spreadsheet and it found no commas from when I did that snapshot:
https://docs.google.com/spreadsheets/d/1twJhvSKajD2vtn71AEPF0e3_o2gUz3mf6qaNcytkVjM/edit?usp=sharing

As mentioned, and given I believe they use the same code, opening hours may be affected too?

**How to Reproduce**
Fill in a postbox collection time quest with different times on different days.

**Versions affected**
Android 9, SC v18.0

This appears to be since SC 11 for some people:
https://www.openstreetmap.org/changeset/69425227

westnordost:
This came up a couple of times and been discussed a couple of times, so I am closing this as duplicate.

> So clear as mud!

That's a matter of opinion. The `additional_rule_separator` reflects the real life intuitive usage, i.e. if you would copy 1:1 the opening hours plate on shops, that separator would be the one you would want to use. Example opening hours:

[...]

Issue #1578, opened 2019-09-23
Postbox collection time uses commas instead of semicolons.


Mbodin:
Hi, I noticed the history of this object: https://www.openstreetmap.org/node/450211913/history
In a nutshell I have added the collection time through StreetComplete in this changeset: https://www.openstreetmap.org/changeset/72027930 (the interface of StreetComplete makes it perfect for this kind of changes)
StreetComplete then added the value “Mo-Fr 09:00, Sa 07:00” to the key “collection_times”. I think that “Mo-Fr 09:00; Sa 07:00” was expected instead (with a semicolon in place of the comma). This has been changed in the meantime for this particular postbox.

**How to Reproduce**
It seems that this happens on all postbox that I filled through StreetComplete (here is another example: https://www.openstreetmap.org/node/6772668486 ).

**Versions affected**
StreetComplete version 14.0, Android version 7.1.2 N2G47H

westnordost:
Hi @Mbodin 

Actually, [both versions are correct](https://openingh.openstreetmap.de/evaluation_tool/?EXP=Mo-Fr%2009%3A00%2C%20Sa%2007%3A00&lat=48.7769&lon=9.1844&mode=1).

This is a duplicate of #1402

Mbodin:
Right. I was basing myself on these two webpages, which seem to strongly advise to use the semicolon to separate different day ranges:
https://wiki.openstreetmap.org/wiki/Key:collection_times
https://wiki.openstreetmap.org/wiki/Key:opening_hours
Anyway, sorry about the duplicate, I did look for such, but missed this one.

Issue #1402, opened 2019-05-23
Postbox Quest giving incorrectly formatted tag entries


jambamkin:
It's been pointed out to me that when I am added collection times to postboxes that have a different collection times on the weekend it is tagged as '=Mo-Fr␣09:00,␣Sa␣07:00' when the correct format is '=Mo-Fr␣09:00;␣Sa␣07:00'.

westnordost:
Both are correct.

PR #2604, opened 2021-02-22
Fix collection_times formatting

TurnrDev:
[Key:collection_times](https://wiki.openstreetmap.org/wiki/Key:collection_times#Examples) - days should be separated with a semicolon. This fixes that

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt
@@ -19,10 +19,10 @@ class AddPostboxCollectionTimesTest {
     @Test fun `apply collection times answer`() {
         questType.verifyAnswer(
             CollectionTimes(listOf(
-                WeekdaysTimes(Weekdays(booleanArrayOf(true)), mutableListOf(60)),
+                WeekdaysTimes(Weekdays(booleanArrayOf(true)), mutableListOf(60, 65)),
                 WeekdaysTimes(Weekdays(booleanArrayOf(false, true)), mutableListOf(120))
             )),
-            StringMapEntryAdd("collection_times", "Mo 01:00, Tu 02:00")
+            StringMapEntryAdd("collection_times", "Mo 01:00,01:05; Tu 02:00")
         )
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
@@ -63,7 +63,7 @@ class AddPostboxCollectionTimes : OsmFilterQuestType<CollectionTimesAnswer>() {
                 changes.add("collection_times:signed", "no")
             }
             is CollectionTimes -> {
-                changes.updateWithCheckDate("collection_times", answer.times.joinToString(", "))
+                changes.updateWithCheckDate("collection_times", answer.times.joinToString("; "))
             }
         }
     }


app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt
```diff
@@ -19,10 +19,10 @@ class AddPostboxCollectionTimesTest {
     @Test fun `apply collection times answer`() {
         questType.verifyAnswer(
             CollectionTimes(listOf(
-                WeekdaysTimes(Weekdays(booleanArrayOf(true)), mutableListOf(60)),
+                WeekdaysTimes(Weekdays(booleanArrayOf(true)), mutableListOf(60, 65)),
                 WeekdaysTimes(Weekdays(booleanArrayOf(false, true)), mutableListOf(120))
             )),
-            StringMapEntryAdd("collection_times", "Mo 01:00, Tu 02:00")
+            StringMapEntryAdd("collection_times", "Mo 01:00,01:05; Tu 02:00")
```
TurnrDev, line 25: I also added an example of multiple collections a day. To really test `WeekdaysTimes` and `AddPostboxCollectionTimes.applyAnswerTo` :p

matkoniecz:
See https://github.com/streetcomplete/StreetComplete/issues?q=is%3Aissue+semicolon+opening+hours+is%3Aclosed

Especially  #1786  #1578 #1402 

Has anything changed since https://github.com/streetcomplete/StreetComplete/issues/1578#issuecomment-534069787 ?

westnordost:
Please see any of these tickets/comments:
- https://github.com/streetcomplete/StreetComplete/issues/1988#issuecomment-666327210
- https://github.com/streetcomplete/StreetComplete/issues/1786#issuecomment-618386613
- https://github.com/streetcomplete/StreetComplete/issues/1402
- https://github.com/streetcomplete/StreetComplete/issues/1578

westnordost:
@matkoniecz beat me to it...

TurnrDev:
If you really wanna go this route, edit opening_hours to use a comma instead of the semicolon. Please, for the sake of consistently, it's driving me insane.

If both are perfectly correct and semicolon is used more, why do you keep pushing for a comma. It's not intuitive given knowledge of the rest of the tagging on OSM.

TurnrDev:
Had a little discussion on the [OpenStreetMap World Discord](https://discord.com/channels/413070382636072960/413070502580453387/813506349119307806). This is the consensus:


* `;` is used as a delimiter on all tags except opening_hours varients, where `; ` us used
* Whilst `, ` is technically correct because, whilst it doesn't reset any attributes from the last rule, it can be used in place of `; ` if you explicitly set all the rules after it. Which is a side effect of the design, and not a function of (otherwise `; ` would be redundant and not accepted)
* westnordost feels `, ` is "more intuitive" (subjective) so chose this over `; `, going against the community on a tool that doesn't give the user freedom to modify tags before saving.
* Despite countless Issues and even a Pull Request, westnordost still fights this community on this tag.

Sounds like another `landuse=reservoir` to me...

westnordost:
Sorry, you do not bring anything new to the table. I'll close this.

westnordost:
But actually, I don't care that much whether it is comma or semikolon. If you provide a PR, with tests, I'll merge it.

The condition is that if the weekdays the user provided in the UI collide with each other, the comma separator must be used and that the change is well tested (with unit tests)

So:

`Mo-Fr 16:00; Sa 12:00` is fine with semicolon.

But in this case, 
`Mo-Fr 09:00, We 16:00`
a comma must be used.

For this, I don't care if only commas are used or only for those rules whose weekdays collide
`Mo-Fr 09:00, We 16:00; Sa 12:00`

TurnrDev:
> If you provide a PR, with tests, I'll merge it.
That's what this is.

matkoniecz:
Note

> But in this case,
> `Mo-Fr 09:00, We 16:00`
> a comma must be used.

See also #2606 that adds test that is missing right now.

TurnrDev:
`Mo-Fr 09:00, We 16:00` would this be "two collection on Wednesday?" But `Mo-Fr 09:00, We 16:00` is one collection on Wednesday, yeah?

matkoniecz:
`Mo-Fr 09:00, We 16:00` means "two collection on Wednesday, one 9:00 one 16:00" `But Mo-Fr 09:00; We 16:00` (with ;) means one collection on Wednesday, on 16,

TurnrDev:
> 
> 
> `Mo-Fr 09:00, We 16:00` means "two collection on Wednesday, one 9:00 one 16:00" `But Mo-Fr 09:00; We 16:00` (with ;) means one collection on Wednesday, on 16,

Gotcha! I'll continue working on this :)

@ westnordost Can you reopen so I can change to draft?

Merged in test from #2606

PR #2606, opened 2021-02-23
document case where comma is needed

matkoniecz:
also, mention that for simple cases semicolons are fine, as long as it is not breaking complex cases

I added test for overlapping ranges so @TurnrDev  and other people interested in moving from `,` to `;` will know what is the actual blocker.

Not committing directly as apparently people have strong feelings about this topic and would be nice to check whatever I missed something.

This change has no effect on how StreetComplete behaves, it is only documentation of case where "lets replace all commas with semicolons for day range separation" breaks down.

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt
@@ -22,7 +22,22 @@ class AddPostboxCollectionTimesTest {
                 WeekdaysTimes(Weekdays(booleanArrayOf(true)), mutableListOf(60)),
                 WeekdaysTimes(Weekdays(booleanArrayOf(false, true)), mutableListOf(120))
             )),
+            // here ; would be fine as well instead of ;
+            // see https://github.com/streetcomplete/StreetComplete/pull/2604#issuecomment-783823068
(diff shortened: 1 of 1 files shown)


app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt
```diff
@@ -22,8 +22,23 @@ class AddPostboxCollectionTimesTest {
                 WeekdaysTimes(Weekdays(booleanArrayOf(true)), mutableListOf(60)),
                 WeekdaysTimes(Weekdays(booleanArrayOf(false, true)), mutableListOf(120))
             )),
+            // here ; would be fine as well instead of ;
```
HolgerJeromin, line 25:
```suggestion
            // here ; would be fine as well instead of ,
```

westnordost:
It's nice of you to provide an example. It should be noted that this is just one case of several that need to be tested. Multiple day ranges could collide independent of each other for example. Or, of many rules, only the very last collides with the first, third and fifth rule, for example.

matkoniecz:
> Multiple day ranges could collide independent of each other for example

What you mean here?

> Or, of many rules, only the very last collides with the first, third and fifth rule, for example.

I added last vs first collision.

westnordost:
> What you mean here?

Mo-Fr 8:00, We 12:00; Sa-So 12:00, So 16:00

westnordost (approved):

PR #2665, opened 2021-03-18
Support cycleway=separate (fixes #2525)

### LINKED ISSUE #2525: Support bicycle=use_sidepath
**Use case**
It shall be possible to choose "There is a bicycle path belonging to this road which is tagged separately and mandatory to use.

https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath

This should affect the question "Is there a cycleway? Which kind".

Currently there are the following options: "none", and several kinds of cycle lanes that are part of the same road.

**Proposed Solution**

- add an option "separately tagged cycleway".
- In a second step, ask: "Is the separately tagged cycleway mandatory to use?" Only if the answer is "Yes", apply "use_sidepath".

Something similar seems to be in place for footpaths, see screenshot here: https://twitter.com/StreetComplete/status/1352995173291196416

westnordost:
Add support for cycleway=separate. In particular, lets the user choose for each side individually "displayed separately on the map".

For each side individually, because this might very well be the real-life occurance: One side is mapped on the road-way, the other is mapped as a separate way.

However, when `cycleway:*=separate` has been added once for a road, the question will not be asked again periodically as with for other cycleway values.

Screenshot_1616026744
Screenshot_1616026755

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bikeway/AddCycleway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bikeway/AddCycleway.kt
@@ -228,6 +228,9 @@ class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
             BUSWAY -> {
                 changes.addOrModify(cyclewayKey, "share_busway")
             }
+            SEPARATE -> {
+                changes.addOrModify(cyclewayKey, "separate")
+            }
             else -> {
                 throw IllegalArgumentException("Invalid cycleway")
             }
@@ -259,7 +262,6 @@ class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
         changes.deleteIfExists("$cyclewayKey:oneway")
         changes.deleteIfExists("$cyclewayKey:segregated")
         changes.deleteIfExists("sidewalk$sideVal:bicycle")
-
     }
 
     companion object {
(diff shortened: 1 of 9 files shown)


matkoniecz:
How user is supposed to be able to answer that if using StreetComplete as single app?

See https://github.com/streetcomplete/StreetComplete/issues/2525#issuecomment-766106872

> Note that StreetComplete user may be unable to know whatever answer 

> > There is a bicycle path belonging to this road which is tagged separately and mandatory to use

> is correct, as cycleways and footways are both shown in the same way.

> To check that one would need to look at OSM data using some other tool.

is it option for users who map given area and know what is mapped where?

westnordost:
The app would be using `cycleway:left/right/both=separate`, not `bicycle=use_sidepath`, so the element of access restrictions (whether or not that separately mapped cycleway is mandatory or not) is removed.

matkoniecz:
Lets say there is 

cycleway along road in reality 
in OSM there is outdated `highway=footway`

is it OK to tag `cycleway:left=separate` in such case?

westnordost:
What do you mean with outdated?

matkoniecz:
There is a cycleway but it is not mapped in OSM. There is instead mapped a `highway=footway`.

Is it OK to use `cycleway=separate` in such case?

westnordost:
Yes, I'd say so. Intermixing the two schemes (separate vs on road-way) is already problematic. The app will support mixing it per direction but, I don't think it is deemed acceptable to mix separate footway as sidewalk + cycle track tagged on road-way. I have never seen it mapped like this as well.

PR #2667, opened 2021-03-20
fix isRoofProbablyVisibleFromBelow

### LINKED ISSUE #2666: roof:shape quest is shown for very high buildings with no roof:levels
@Helium314 wrote:

> @westnordost I just got the roof quest for an 8 level building without roof:levels, despite 8/(0+2) being larger than 2.

> In https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/roof_shape/AddRoofShape.kt it looks like isRoofProbablyVisibleFromBelow always returns null if roof:levels does not exist (line 53). Since the result of isRoofProbablyVisibleFromBelow is only compared to false, null is treated like true (lines 34 and 43).
No roof:levels means that the roof shape quest is always shown (at least if roofs are not usually flat in my country)

The app should probably interpret a missing `roof:levels` as `roof:levels=0` for the roof shape quest.

Helium314:
> The app should probably interpret a missing roof:levels as roof:levels=0 for the roof shape quest.

fix #2666 
Additionally now the quite common fractional value of `roof:levels` (often 0.5 or 1.5) is now understood correctly. (according to [this](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS40LjMwIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsImpzQ29kZSI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiLyoqXG4gKiBZb3UgY2FuIGVkaXQsIHJ1biwgYW5kIHNoYXJlIHRoaXMgY29kZS4gXG4gKiBwbGF5LmtvdGxpbmxhbmcub3JnIFxuICovXG5cbmZ1biBtYWluKCkge1xuICAgIHZhbCByb29mTGV2ZWxzID0gXCIxLjVcIlxuICAgIHByaW50bG4ocm9vZkxldmVscz8udG9GbG9hdE9yTnVsbCgpID86IDApXG4gICAgcHJpbnRsbihyb29mTGV2ZWxzPy50b0ludE9yTnVsbCgpID86IDApXG59In0=))

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/roof_shape/AddRoofShape.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/roof_shape/AddRoofShape.kt
@@ -50,7 +50,7 @@ class AddRoofShape(private val countryInfos: CountryInfos) : OsmElementQuestType
 
     private fun isRoofProbablyVisibleFromBelow(tags: Map<String,String>?): Boolean? {
         if (tags == null) return null
-        val roofLevels = tags["roof:levels"]?.toIntOrNull() ?: return null
+        val roofLevels = tags["roof:levels"]?.toFloatOrNull() ?: 0
         val buildingLevels = tags["building:levels"]?.toIntOrNull() ?: return null
         if (roofLevels < 0 || buildingLevels < 0) return null
         return buildingLevels / (roofLevels + 2f) < 2f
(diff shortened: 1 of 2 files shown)


westnordost:
Hey, thank you! But it would be required to also add a test that confirms that it works

westnordost:
(While being in that class, press Ctrl+Shift+T to directly jump to the test class)

westnordost:
If you are not up to it, let me know. I will do that then

Helium314:
I added some tests and switched to using float for all levels for consistency, even though it hardly matters in the case if `building:levels`
I hope the tests are acceptable:
- included high buildings with no or 0 `roof:levels`
- added test for high buildings with many `roof:levels`
- changed test for 0 or no `roof:levels` to be unknown only for buildings with no or few levels
- included fractional `roof:levels` and `building:levels`

westnordost (approved):

westnordost:
Looks good, thank you!

PR #2674, opened 2021-03-24
Include icon information in quest list CSV

FloEdelmann:
This implements @rugk's request in https://github.com/streetcomplete/StreetComplete/pull/2523#issuecomment-804395330:

> In the markdown version we had a column for the quest icon IIRC.
> For [that new crown quest](https://wiki.openstreetmap.org/w/index.php?title=StreetComplete%2FQuests&type=revision&diff=2131008&oldid=2130991) it took me some minutes to actually find the correct icon due to the different naming (I only looked at the source code later), so one idea may be to add another column which gives me the link to the SVG icon, so one can upload it more easily.

Basically, it is [`2c4450f` (#2523)](https://github.com/streetcomplete/StreetComplete/pull/2523/commits/2c4450fc1fd77dc1743e979ad2fc008b1f23bbc7) reverted and translated to Kotlin.

For this to work, all `ic_quest_*` drawables need a corresponding SVG file in `/res/graphics/quest icons/` with the same name. That's why I also noticed that `ic_quest_wheelchair_outside` is actually the same as `ic_quest_toilets_wheelchair`: The former has no SVG version.


westnordost:
👍

PR #2677, opened 2021-03-25
Fix typo in changeset message

eginhard:


westnordost:
Oh, haha, @matkoniecz strikes again ;-) Didn't notice it, thanks!

(He implemented this quest and he is not a native speaker)

smichel17:
Being a native speaker is no protection… I misspell receive (as recieve) nearly every time I type it :P

westnordost:
by the way, "ei" is pronounced like "aye" in German

matkoniecz:
aarghhh.

whatever vs whether is my most common mistake. In addition to missing "a"/"an"/"the" - but this is explainable as Polish is without such grammatical construct. But "whatever" replacing "whether" is some weird habit.

Thanks for catching it!

Issue #2506, opened 2021-01-15
Layers - Technical Planning


westnordost:
For the Layers #2461 concept, there is some deep code restructuring that has to be made. Here is an attempt to summarize it.

# Summary

- All OSM data downloaded plus generated geometry must be persisted. The app should take care of dumping old unneeded OSM data automatically.
- The locally persisted OSM data and geometry must be accessible (loaded into memory) efficiently by bounding box because not all local data should be displayed / processed at the same time.
- any change made on the OSM data (tag changes, deletions, split ways, reverts of those) must be added to a single persisted queue of actions (~diff) to be performed
- when the OSM data is loaded into memory for display (in layers), for quest creation etc., the queue of actions must be performed on it first so to work with a "working set" of data rather than only the original data downloaded from OSM
- on upload, this queue of actions is applied to the original OSM data one by one and the affected elements updated from the server until that queue is empty
 
# Milestones

1. Change the download, to keep all the OSM data and manage its cleanup. 
2. Read-only version of a layer (vertically): data management, layer definition, UI
3. Change the way changes (actions) are stored, treated and processed

# M1

## Download

On download of OSM data, element geometry for all downloaded elements must be created and persisted. Currently, this is only done for any elements that get a quest.

In detail:
- rename `QuestDownloader` to `OsmDownloader` or similar. Generalize it. Put the notes download into own class
- the `OsmApiQuestDownloader` should not download OSM data by itself (so, rename) but having it handed to it. Maybe by `OsmDownloader` or maybe there is some listener that fires whenever data in the OSM elements table changed and the `OsmApiQuestDownloader` is informed of that, updates the quests based on that. Maybe there is an oppurtunity to merge the logic in `OsmApiQuestDownloader` with `OsmQuestGiver`

## Data Cleanup

A last update timestamp for each element must be added to be able to periodically dump old data. Based on that, data older than X days not referenced by an action must be dumped.

In detail:
- add last update timestamp to NodesTable, WaysTable, RelationsTable
- add cleanup function to the associated DAOs, call it f.e. on each application startup, or on each download. The "fresh" time should be the same as the time it takes for unsolved quests to be deleted. Same mechanism
- note: only data should be deleted that is not referenced by other tables, i.e. the OsmQuest table or later the Actions table or whatever one may call it
- a user option to control how long "quest data" should be stored would be possible, but optional

## Data access

[...]

PR #2678, opened 2021-03-25
Replace osmapi's data classes with pure Kotlin ones

FloEdelmann:
As discussed in https://github.com/streetcomplete/StreetComplete/issues/2506#issuecomment-806163270.

@westnordost I'm opening this PR early, so you can tell me if I'm on the right track.


westnordost:
Jesus, you are fast!

I'll skim through it really quick

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/BoundingBox.kt
```diff
@@ -0,0 +1,41 @@
...
+    } else listOf(this)
+}
+
+// TODO check usages
+fun BoundingBox.getAsLeftBottomRightTopString(): String =
+    listOf(min.longitude, min.latitude, max.longitude, max.latitude)
+        .joinToString(",") { it.format(7) }
+
+fun BoundingBox.toOsmApiBoundingBox() =
+    OsmApiBoundingBox(min.latitude, min.longitude, max.latitude, max.longitude)
```
westnordost, line 41: Conversions to Osmapi-Data classes should ideally not be in the same file as the data class definitions, so it can be split easily into a portable portion and a non-portable portion. Best have this in the same class where the wrapper code is all located (f.e. in `JavaOsmMapDataApi` or similar. Have a look at `AndroidDatabase` to see how I realized this separation).

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/Note.kt
```diff
@@ -0,0 +1,75 @@
...
+    status.toNoteStatus(),
+    comments.map { it.toNoteComment() }
+)
+
+fun OsmApiNote.Status.toNoteStatus() = when(this) {
+    OsmApiNote.Status.OPEN   -> Note.Status.OPEN
+    OsmApiNote.Status.CLOSED -> Note.Status.CLOSED
+    OsmApiNote.Status.HIDDEN -> Note.Status.HIDDEN
+    else -> throw NoSuchFieldError()
+}
```
westnordost, line 43: see last comment

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesApi.kt
```diff
@@ -0,0 +1,120 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.toOsmApiBoundingBox
+
+
+/**
+ * Creates, comments, closes, reopens and search for notes.
+ * All interactions with this class require an OsmConnection with a logged in user.
+ *
+ * @see NotesDao which is internally used.
+ */
+class NotesApi(osm: OsmConnection) {
```
westnordost, line 19: NotesApi should be an interface. Classes that use this interface should only know the interface and don't worry about how it is constructed.

F.e. `JavaNotesApi` (or `OsmNotesApi`, `NotesApiImpl` or another name that makes sense to you) would then be the implementation of this interface for the Java/Android platform.

So, classes that need `NotesApi` to work (f.e. the NotesUploader) just declare that they need a NotesApi as constructor parameter and in the appropriate module, it is provided.

You can peek at `AndroidDatabase` on how I did this.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesApi.kt
```diff
@@ -0,0 +1,120 @@
...
+     * Create a new note at the given location
+     *
+     * @param pos position of the note.
+     * @param text text for the new note. Must not be empty.
+     *
+     * @throws OsmAuthorizationException if this application is not authorized to write notes
+     *                                   (Permission.WRITE_NOTES)
+     *
+     * @return the new note
+     */
```
westnordost, line 32: Hmm right, this means that the exception classes (etc) need to be wrapped as well.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesApi.kt
```diff
@@ -0,0 +1,120 @@
...
+        notesDao.getAll(bounds.toOsmApiBoundingBox(), search, handler.toOsmApiNoteHandler(), limit, hideClosedNoteAfter)
+
+    /**
+     * Get a number of notes that match the given filters.
+     *
+     * @param handler The handler which is fed the incoming notes
+     * @param filters what to search for. I.e. `QueryNotesFilters().byUser(123).limit(1000)`
+     */
+    fun find(handler: Handler<Note>, filters: QueryNotesFilters?) =
+        notesDao.find(handler.toOsmApiNoteHandler(), filters)
```
westnordost, line 119: Is this even used in the code? The wrapper only needs to offer what is actually used in the code.

westnordost:
If you have local progress on it, it would be good if you pushed that to your branch because then I could in parallel already start working on things that depend on this refactor, such as reverting deletion of POIs and kotlinx-serialization

FloEdelmann:
@westnordost Sorry for the delay; I'll push my local progress and continue on this now.

FloEdelmann:
@westnordost All instances of the old NotesDao should now be replaced. The app compiles and runs, but only after clearing the data.

I guess that's because in the `osm_notes` database table, the old serialized `NoteComment` and `NoteComment.Action` classes are serialized, while the code expects the new classes instead. I noticed that there are no migrations in the `elements-first-class` branch anymore, so should I add one there?

Could you review the current state? If this is fine, I'll replace the other data classes afterwards.

westnordost:
> I noticed that there are no migrations in the elements-first-class branch anymore, so should I add one there?

No, v32.0 will be a fresh start.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
```diff
@@ -103,7 +99,8 @@ class NoteEditsUploader @Inject constructor(
         }
     }
 
-    private fun uploadCreateNote(pos: LatLon, text: String): Note = notesApi.create(pos, text)
+    private fun uploadCreateNote(pos: LatLon, text: String): Note =
+        notesApi.create(pos.toLatLon(), text)
```
westnordost, line 103: this looks wrong. LatLon is already a LatLon, why does it have a method called `toLatLon`?

In any case, `notesApi` should accept a `de.westnordost.streetcomplete.data.osmnotes.LatLon`.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesDownloader.kt
```diff
@@ -44,6 +40,6 @@ private suspend fun NotesApi.getAll(
     hideClosedNoteAfter: Int
 ): List<Note> = withContext(Dispatchers.IO) {
     val notes = ArrayList<Note>()
-    getAll(bbox, notes::add, limit, hideClosedNoteAfter)
+    getAll(bbox.toBoundingBox(), notes::add, limit, hideClosedNoteAfter)
```
westnordost, line 43: this looks wrong. BoundingBox is already a BoundingBox, why does it have a method called toBoundingBox ?

In any case, notesApi should accept a de.westnordost.streetcomplete.data.osmnotes.toBoundingBox .

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
```diff
@@ -91,7 +92,7 @@ import javax.inject.Singleton
 
     private fun createQuestForNote(note: Note, blockedNoteIds: Set<Long> = setOf()): OsmNoteQuest? =
         if(note.shouldShowAsQuest(userStore.userId, showOnlyNotesPhrasedAsQuestions, blockedNoteIds))
-            OsmNoteQuest(note.id, note.position)
+            OsmNoteQuest(note.id, note.position.toOsmLatLon())
```
westnordost, line 95: The data classes from `de.westnordost.osmapi` should not occur at all in the StreetComplete source code anymore except in the `NotesApiImpl` class

app/src/test/java/de/westnordost/streetcomplete/testutils/TestDataShortcuts.kt
```diff
@@ -65,26 +66,15 @@ fun note(
...
     timestamp: Long = 0,
     comments: List<NoteComment> = listOf(comment("test", NoteComment.Action.OPENED))
-) = Note().also {
-    it.id = id
-    it.comments = comments
-    it.dateCreated = Date(timestamp)
-    it.position = position
-    it.status = Note.Status.OPEN
-}
+) = Note(position.toLatLon(), id, timestamp, null, Note.Status.OPEN, comments)
```
westnordost, line 69: .toLatLon() - see comment(s) above.

westnordost (changes requested):
Hmm, maybe there was a misunderstanding. It looks like the code is (still) littered with usages of the osmapi data classes.
The goal of this refactor is to "ban" those data classes into the `NotesApiImpl` and `MapDataApiImpl` classes.

I think it is counterproductive if I add a review comment below every occurance of `.toLatLon()` etc., it will be more efficient to Ctrl+F for any remaining occurances.

app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
```diff
@@ -26,11 +28,11 @@ object OsmApiModule {
 
     @Provides fun userDao(osm: OsmConnection): UserApi = UserApi(osm)
 
-    @Provides fun notesDao(osm: OsmConnection): NotesApi = NotesApi(osm)
+    @Provides fun notesDao(osm: OsmConnection): NotesApi = NotesApiImpl(osm)
```
westnordost, line 31:
```suggestion
    @Provides fun notesApi(osm: OsmConnection): NotesApi = NotesApiImpl(osm)
```

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NoteDao.kt
```diff
@@ -5,16 +5,14 @@ import de.westnordost.osmapi.map.data.BoundingBox
...
 
 import javax.inject.Inject
 
 import de.westnordost.streetcomplete.util.Serializer
 import de.westnordost.osmapi.map.data.OsmLatLon
-import de.westnordost.osmapi.notes.Note
-import de.westnordost.osmapi.notes.NoteComment
 import de.westnordost.streetcomplete.data.CursorPosition
 import de.westnordost.streetcomplete.data.Database
+import de.westnordost.streetcomplete.data.osm.mapdata.LatLon as SCLatLon
```
westnordost, line 15: no need to name it this way

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestHidden.kt
```diff
@@ -12,6 +13,6 @@ data class OsmNoteQuestHidden(
 ) : Edit {
     override val key: OsmNoteQuestHiddenKey get() = OsmNoteQuestHiddenKey(OsmNoteQuestKey(note.id))
     override val isUndoable: Boolean get() = true
-    override val position: LatLon get() = note.position
+    override val position: LatLon get() = note.position.toOsmLatLon()
```
westnordost, line 16: OsmNoteQuestHidden should use the LatLon from StreetComplete

app/src/main/java/de/westnordost/streetcomplete/util/KryoSerializer.java
```diff
@@ -58,10 +58,10 @@
 			StringMapEntryAdd.class,
 			StringMapEntryDelete.class,
 			StringMapEntryModify.class,
-			NoteComment.class,
-			NoteComment.Action.class,
+			NoteComment.class, // TODO(Flo): Replace this with Object.class
+			NoteComment.Action.class, // TODO(Flo): Replace this with Object.class
 			Date.class,
-			User.class,
+			User.class, // TODO(Flo): Replace this with Object.class
```
westnordost, line 64: Why? It should still be persisted

westnordost: Ah, I understand. Well, as v32.0 will make a clean slate, incompatible serialization is OK in this case as the whole DB is cleared anyway.
(Plus, Kryo will be replaced with kotlinx-serialization anyway)

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesApi.kt
```diff
@@ -0,0 +1,61 @@
...
+     * @param handler The handler which is fed the incoming notes
+     * @param limit number of entries returned at maximum. Any value between 1 and 10000
+     * @param hideClosedNoteAfter number of days until a closed note should not be shown anymore.
+     *                            -1 means that all notes should be returned, 0 that only open notes
+     *                            are returned.
+     *
+     * @throws OsmQueryTooBigException if the bounds area is too large
+     * @throws IllegalArgumentException if the bounds cross the 180th meridian
+     */
+    fun getAll(bounds: BoundingBox, handler: (Note) -> Unit, limit: Int, hideClosedNoteAfter: Int)
```
westnordost, line 60: Actually, an interface like 

```suggestion
    fun getAll(bounds: BoundingBox, limit: Int, hideClosedNoteAfter: Int): List<Note>
```

would probably be more convenient

app/src/test/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploaderTest.kt
```diff
@@ -77,7 +79,7 @@ class NoteEditsUploaderTest {
 
         upload()
 
-        verify(notesApi).create(pos, "abc")
+        verify(notesApi).create(pos.toLatLon(), "abc")
```
westnordost, line 82: X

app/src/test/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploaderTest.kt
```diff
@@ -175,7 +177,7 @@ class NoteEditsUploaderTest {
 
         upload()
 
-        verify(notesApi).create(pos, "test\n\nAttached photo(s):\nx\ny\nz")
+        verify(notesApi).create(pos.toLatLon(), "test\n\nAttached photo(s):\nx\ny\nz")
```
westnordost, line 180: X

FloEdelmann:
> The goal of this refactor is to "ban" those data classes into the `NotesApiImpl` and `MapDataApiImpl` classes

Sorry, I should have cleared that up before asking for your review: Rewriting all note-related code to be completely free of osmapi classes is quite difficult before the `MapDataApi` is done, too.

So for now, as a preliminary state that compiles and runs, I've not completely migrated usages of osmapi's `LatLon` and `BoundingBox`, but rather converted them via the extension functions to the new Kotlin data classes. (I hope this explains the `TODO: Make this private` comments there.)

That will of course be done later once `MapDataApi` is not expecting osmapi classes anymore.

westnordost:
Ah okay, I already suspected that this may be the reason for this.

westnordost:
FYI #2740

westnordost:
I now replaced `Date` with `Instant` in the osmapi library (not released yet): https://github.com/westnordost/osmapi/tree/instant

So as soon as the osmapi is wrapped, faster date parsing can be added. Since kotlinx-datetime is not really stable yet and the Java 8 time API is very similar to it, I think it is also viable to first only migrate to Java 8 time classes and not right away to kotlinx-datetime.

If you have progress on this, please always push your work, even if it does not currently compile. This way, I can also contribute to it without fear that I refactor something that you already did.

FloEdelmann:
I'm currently working on it. I first try to reduce the usage of osmapi's `LatLon` and `BoundingBox`.

westnordost:
Nice! Ok I do not intend to do anything more today. The comment regarding pushing your work was more along the lines of that a completed chunk of work doesn't lie around for several days. I didn't want to put stress on you to do atomic commits or something.

FloEdelmann:
A little more progress: The osmapi classes are now replaced with the new Kotlin data classes in most places. Map display and download seems to work fine.

A few problems:

1. When opening a quest, the app crashes with this error:

	```
	java.lang.IllegalArgumentException: Illegal value type de.westnordost.streetcomplete.data.osm.mapdata.Way for key "element"
	    at androidx.core.os.BundleKt.bundleOf(Bundle.kt:99)
	    at de.westnordost.streetcomplete.quests.AbstractQuestAnswerFragment$Companion.createArguments(AbstractQuestAnswerFragment.kt:443)
	    at de.westnordost.streetcomplete.map.MainFragment.showQuestDetails(MainFragment.kt:785)
	```

	But `Way` is a subtype of `Element`? Am I missing something?
2. When running the "Tests in 'de.westnordost.streetcomplete'", 6 of them fail with the following error:

	```
	java.lang.IllegalStateException: Could not initialize plugin: interface org.mockito.plugins.InstantiatorProvider2 (alternate: interface org.mockito.plugins.InstantiatorProvider)
	```

	image
3. When running the "streetcomplete in StreetComplete.app" tests, 4 of them fail:
	1. `ElementEditUploaderTest`: `handles changeset conflict exception`
		  ```
		  java.lang.ClassCastException: kotlin.Unit cannot be cast to de.westnordost.osmapi.map.ElementUpdates
			  at de.westnordost.streetcomplete.data.MapDataApi$MockitoMock$2021988431.uploadChanges(Unknown Source)
			  at de.westnordost.streetcomplete.data.osm.edits.upload.ElementEditUploader.uploadInChangeset(ElementEditUploader.kt:43)
			  at de.westnordost.streetcomplete.data.osm.edits.upload.ElementEditUploader.upload(ElementEditUploader.kt:36)
			  at de.westnordost.streetcomplete.data.osm.edits.upload.ElementEditUploaderTest.handles changeset conflict exception(ElementEditUploaderTest.kt:66)
		  ```

	2. `OsmNoteQuestControllerTest`: `get note quest with comment from user returns null`
		  ```
		  java.lang.AssertionError: expected null, but was:<OsmNoteQuest(id=1, position=LatLon(latitude=0.0, longitude=0.0))>
			  at de.westnordost.streetcomplete.data.osmnotes.notequests.OsmNoteQuestControllerTest.get note quest with comment from user returns null(OsmNoteQuestControllerTest.kt:147)
		  ```

	3. `SphericalEarthMathTest`: `two lines intersect at endpoints` and `two lines intersect at start points`
		```
		Expected :LatLon(latitude=0.0, longitude=0.0)
		Actual   :LatLon(latitude=-0.0, longitude=0.0)
		```

And of course, the *TODO*s still have to be resolved.

@westnordost I'd be glad if you could help me on the issues I mentioned. Thanks :)

westnordost:
> When opening a quest, the app crashes with this error: []...  But Way is a subtype of Element? Am I missing something?

The `Element` from osmapi implements java `Serializable`. `Serializable`s may be added as values into a bundle, but not any class.
The easy fix would be of course to mark them as `Serializable`, too. Like it is the case with `Serializable`. This of course won't work for a possible iOS port, but this is still blue-sky anyway. So I guess we could go with that for now.

Final solutions for this could be either:
1. use kotlinx-serialization to serialize and de-serialize the element (and element geometry) from / to string
2. subtype the `AbstractQuestAnswerFragment` and put any OSM element related stuff into that subclass; that subclass fetches the OSM element itself.

> When running the "Tests in 'de.westnordost.streetcomplete'", 6 of them fail with the following error

Yes, this is normal, you can ignore that. They fail because of some problem with Mockito + Android tests.

app/src/main/java/de/westnordost/streetcomplete/ktx/Element.kt
```diff
@@ -1,36 +1,35 @@
...
     return when (this) {
-        is Node -> OsmNode(newId, newVersion, position, tags, null, newDateEdited)
-        is Way -> OsmWay(newId, newVersion, ArrayList(nodeIds), tags, null, newDateEdited)
-        is Relation -> OsmRelation(newId, newVersion, ArrayList(members), tags, null, newDateEdited)
-        else -> throw RuntimeException()
+        is Node -> Node(newId, position, newTags, newVersion, newTimestampEdited)
+        is Way -> Way(newId, ArrayList(nodeIds), newTags, newVersion, newTimestampEdited)
+        is Relation -> Relation(newId, ArrayList(members), newTags, newVersion, newTimestampEdited)
     }
 }
```
westnordost, line 19: All data classes in Kotlin already have a `copy` method, so this is superfluous now.

westnordost (commented):
Wow, the sheer volume of this refactor!

... I am currently looking through it and small things that catch my eye, I'll solve directly. 

However, I will pause now and continue this in the evening today. I didn't get to the tests yet.

app/src/main/java/de/westnordost/streetcomplete/ktx/Element.kt
FloEdelmann: But `Element` is not a data class. At least in one place, `copy` is called on an `Element`, so we'd then need to differentiate between `Node`/`Way`/`Relation` there.

westnordost: Ah, ok

westnordost:
Ok, I read all the and made small adaptions/corrections.

I didn't have a closer look at the last failing test. There is something that needs to be considered anyway:

The OSMAPI data classes would automatically set itself to `isModified=true` when any tag was changed. The Kotlin data classes do not have this automatism anymore. So, we need to find a new place for that. Maybe something like

```
someElement.edit {
    tags = ...
}
```
...automatically sets the edit timestamp to now and sets `isModified=true`?

Generally, the elements passed to 
```fun uploadChanges(changesetId: Long, elements: Collection<Element>): MapDataUpdates```
are sorted into three categories: Updated elements, created elements and deleted elements. This is controlled by the properties isModified, isDeleted and isNew.

**Comment by westnordost** [comment]
So, what is left to do is:

- [ ] The exceptions the methods in NotesApi throw must also be defined in Kotlin in StreetComplete. It is ok to simplify/merge them if they are each handled the same by the code that uses the api.
- [ ] The exceptions the methods in MapDataApi throw must also be defined in Kotlin in StreetComplete. It is ok to simplify/merge them if they are each handled the same by the code that uses the api.
- [ ] `NotesApiImpl(osm: OsmConnection?)` osm should not be nullable
- [ ] solve the `TODO(Flo): Make this private` todos
- [ ] make the test for the uploader not fail anymore. You could leave this to me. This is related to my last comment. I was thinking maybe it is better if the signature of the upload method would be 
```kotlin
fun uploadChanges(changesetId: Long, elementsChanges: MapDataChanges): MapDataUpdates

data class MapDataChanges(
    val creations: Collection<Element>,
    val modifications: Collection<Element>,
    val deletions: Collection<ElementKey>
)
``` 
instead of propping that information on each individual element.

FloEdelmann:
:ballot_box_with_check: `NotesApiImpl(osm: OsmConnection?)` osm should not be nullable
:ballot_box_with_check: solve the `TODO(Flo): Make this private` todos

westnordost:
I'll merge this PR into master now as it is almost done minus

- [ ] The exceptions the methods in NotesApi throw must also be defined in Kotlin in StreetComplete. It is ok to simplify/merge them if they are each handled the same by the code that uses the api.
- [ ] The exceptions the methods in MapDataApi throw must also be defined in Kotlin in StreetComplete. It is ok to simplify/merge them if they are each handled the same by the code that uses the api.

for which I'll create new issues. These two however don't block already testing the "Release Candidate". When everything is in master, it'll be easier to keep track of the remaining bugs.

PR #2680, opened 2021-03-25
add shed and garage to farm buildings

matkoniecz:
I am pretty sure that it would be an improvement (I just run into it while tagging), but in a bit subjective part, so I opened a PR

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
@@ -98,7 +98,7 @@ enum class BuildingTypeCategory(val type: BuildingType?, val subTypes: List<Buil
         GARAGE, GARAGES, CARPORT, PARKING
     )),
     FOR_FARMS(null, listOf(
-        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK, SHED, GARAGE
+        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK, SHED
     )),
     OTHER(null, listOf(
         SHED, ROOF, SERVICE, HUT, TOILETS, HANGAR, BUNKER, HISTORIC, ABANDONED, RUINS
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
@@ -98,7 +98,7 @@ enum class BuildingTypeCategory(val type: BuildingType?, val subTypes: List<Buil
         GARAGE, GARAGES, CARPORT, PARKING
     )),
     FOR_FARMS(null, listOf(
-        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK
+        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK, SHED, GARAGE
     )),
     OTHER(null, listOf(
         SHED, ROOF, SERVICE, HUT, TOILETS, HANGAR, BUNKER, HISTORIC, ABANDONED, RUINS


peternewman:
When is a shed not a barn, and other philosophical questions?

westnordost:
I wouldn't include garage. Garages are for cars. What does this have to do with farms? A tractor would either go into a barn or a shed (if it is really small, i.e. just an enclosure for the tractor and nothing else).

matkoniecz:
> I wouldn't include garage. Garages are for cars. What does this have to do with farms? A tractor would either go into a barn or a shed (if it is really small, i.e. just an enclosure for the tractor and nothing else).

on farms in Poland it is typical to have a car garage for car, more typical than in city due to more space available

westnordost:
But wouldnt a user search for a garage in the "for cars" menu point? After all, the garage is also not in "residential".

On March 30, 2021 8:05:27 AM GMT+02:00, Mateusz Konieczny ***@***.***> wrote:
>> I wouldn't include garage. Garages are for cars. What does this have
>to do with farms? A tractor would either go into a barn or a shed (if
>it is really small, i.e. just an enclosure for the tractor and nothing
>else).
>
>on farms in Poland it is typical to have a car garage for car, more
>typical than in city due to more space available
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/2680#issuecomment-809937452

-- 

matkoniecz:
In my case I was tagging things on a small farm area, so I clicked "on a farm". Obviously, I selected "for cars instead", but it seems that listing this two that are often appearing on farms would be helpful

> After all, the garage is also not in "residential".

but garage is not fitting "residential" and fitting "on a farm".

smichel17 (changes requested):
There are two different philosophies here. One is tagging by use (*For* cars) and the other is tagging by location (*On* a farm).

…hmm. That's not quite what I want to get at. Maybe it's more like, one philosophy is that the categories should be exclusive. So, "on a farm" means buildings you'd *only* find on a farm and "for cars" means buildings that are *only* (okay, or *mostly*) for cars. 

…hmm. That's also not quite it. I am having a hard time putting the right words on it. Maybe it's more like, "if you find a specific building type that fits, and then you scroll up and tag it in the generic category it's under instead, it should be the right tag." So if I find garage under "for cars", it should be correct if I scroll up and tag it as a generic "for cars" building instead (whatever tags that applies), and so it would be wrong to put it under farm buildings unless it's okay to tag as a generic farm building.

---

Whew. Hopefully you can piece together what I'm getting at from those different attempts at explaining.

The other philosophy, which I think drives this PR, is easier to describe. It is more like, "group together all the buildings which are likely to be found in the area I am currently mapping." So, if you are on a farm, first you choose "on a farm", and then all the buildings you'd expect to see should be there. Extending that philosophy, I think you'd also make the argument for *also* putting shed and garage into the residential category, or maybe splitting the residential category into rural and urban.

I'm mostly in favor of keeping the current approach, where buildings only appear under one category. I think there are too many combinations of buildings you'd expect to find near one another to create categories for each of those, and the category boundaries are much more vague, so it's hard for someone to know where to look if they already know which building type they're looking for, but don't remember which category it appears in. Also, the user is already creating their own custom collection of buildings are common in their area, the cache of recently selected answers.

smichel17:
Aah, and of course just after I post I think of a better analogy to explain the difference. It is like the two different ways of organizing code. Say you have four classes, two views and two controllers, one each for the map screen and the profile screen.

- One way of organizing is to create a `Views` folder with `MapView` and `ProfileView`, and `Controllers` folder with the other two.
- The other way is to create a `Map` folder with `MapController` and `MapView`, and a `Profile` folder with the other two.

I also favor the first approach when coding, unless you have a really large number of classes and you can break them down into groups of modules which do not share much code (and then organize the classes in each module the first way).

The promise of the second way is that when you're working on the map, it's nicer because all the classes you'll need to touch are in the same folder so you don't need to jump around as much. I tried it in Red Moon based on the recommendation of some blog post or other and found that in practice it's very hard to achieve such clear separation of concerns; all it actually did was introduce another layer of abstraction which made it harder to remember what folder something was in.

westnordost:
You surely use a lot of words, @smichel17

smichel17:
For a year or two, I was focusing heavily on writing very concisely. This had the effect of all my messages taking an extremely long time to compose. Now I am experimenting to find a balance between speed of writing and reading.

westnordost:
So anyway, I'd say "on a farm", there may even be more things... roofs, static caravans, huts, ruins, service buildings etc.

Especially for garages, I think it makes no sense to include it in the "on a farm" category because the "for cars" category is actually shown **above** the "on a farm" category.

For shed, yeah, I understand it is something like a "rest" category, where one only looks into if it's not possible to find anything in any other category, so it may make sense to include it in another category too. Since a shed is common enough on a farm, I think it's ok to include that.

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
```diff
@@ -98,7 +98,7 @@ enum class BuildingTypeCategory(val type: BuildingType?, val subTypes: List<Buil
         GARAGE, GARAGES, CARPORT, PARKING
     )),
     FOR_FARMS(null, listOf(
-        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK
+        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK, SHED, GARAGE
```
westnordost, line 101:
```suggestion
        FARM, FARM_AUXILIARY, SILO, GREENHOUSE, STORAGE_TANK, SHED
```

westnordost:
And I merged that.

PR #2691, opened 2021-03-28
replace QuestGroup + id with each quest type defining its own key

westnordost:
Reasons:
1. OSM quests are already unique by element id + element type + quest type. No need for a second id
2. The same for OSM note quests: the note id is the id
3. Most importantly: when elements update, their position might change. The position is (also) persisted in the OSM Quests Table. Since there is no UPSERT in SQLite, quests were inserted with INSERT OR IGNORE, i.e. if it was already there, it was not inserted. This meant that the position was also not updated. With this extra ID gone, one can safely replace everything when inserting.

When for example splitting a way, the position updates. So, the situation was that after splitting a way, the marker for the split way was still at the old position.
There are other ways to fix this , but I think this is a nice solution. Another way to solve this would have been to not persist the position in the quest table in the first place but when displaying quests, getting all OSM element geometry in the given bounding box, then find the quests for these elements. But this would require to always get all the OSM element geometry from database rather than only those which have a quest.


westnordost:
and it's 25 lines shorter!!1

app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
```diff
@@ -72,13 +72,13 @@ import javax.inject.Singleton
...
-    fun getAllVisible(bbox: BoundingBox, questTypes: Collection<String>): List<QuestAndGroup> {
+    fun getAllVisible(bbox: BoundingBox, questTypes: Collection<String>): List<Quest> {
         if (questTypes.isEmpty()) return listOf()
         val osmQuests = osmQuestSource.getAllVisibleInBBox(bbox, questTypes)
         val osmNoteQuests = osmNoteQuestSource.getAllVisibleInBBox(bbox)
 
-        return osmQuests.filter(::isVisible).map { QuestAndGroup(it, QuestGroup.OSM) } +
-               osmNoteQuests.filter(::isVisible).map { QuestAndGroup(it, QuestGroup.OSM_NOTE) }
+        return osmQuests.filter(::isVisible).map { it } +
+               osmNoteQuests.filter(::isVisible).map { it }
```
FloEdelmann, line 81: Isn't `.map { it }` a no-op?

westnordost: oops, you are right. This was some remains from the last implementation

PR #2699, opened 2021-03-30
assume track to be a line when on a closed way

### LINKED ISSUE #2698: "Add pitch surface" quest display issue for sport=running
The "add pitch surface" quest has a display issue for sport=running. Instead of highlights just the way, it highlights the whole surface inside the way.

**How to Reproduce**
See [this POI](https://www.openstreetmap.org/way/922945221) for exemple

**Versions affected**
v31.1

matkoniecz:
fixes #2698

previously it was assumed to be an area when on closed way, and breaking when tagged as way without having explicit area tag
now it will be assumed to be line, even when on closed way and breaking when tagged as area without having an explicit tag

if SC can detect in quest filter closed ways, maybe adding explicit `yes/no` can be an useful quest? It can be often done from imagery, but would avoid such confusion within SC itself.

screen05

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/ktx/Element.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/ktx/Element.kt
@@ -44,7 +44,7 @@ private val IS_AREA_EXPR = """
     or (emergency and emergency !~ yes|no)
     or historic
     or landuse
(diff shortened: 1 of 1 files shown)


westnordost:
ok. I read that osm carto does it the same

Issue #2716, opened 2021-04-02
Ask for opening hours for shops tagged name:signed=no


rhhsm:
Am I right that shops tagged with name:signed=no are not asked for their opening hours? Where I am in Bulgaria, most small shops such as convenience stores, greengrocers, copy shops, fast food outlets, etc. don't have names, just descriptive signs of what they have for sale ("fruits&vegetables", "pizza slices", etc.) or what they are ("mini market", "food shop", etc.), so when I am asked for their name, I answer "no sign". These shops then don't appear again for the opening hours quest, but most of them do have opening hours displayed. As a result, I guess about 10-20% of shops in my area don't have opening hours on OSM.

Helium314:
Yes... it seems this was not updated when SC switched from tagging `noname=yes` to `name:signed=no`

PR #2718, opened 2021-04-02
allow adding opening hours for name:signed = no

Helium314:
see #2716
(I did not actually test this change, but it seems obvious to me)


westnordost:
Right. Thanks for the fix!

PR #2724, opened 2021-04-09
Use disused:shop instead of shop=vacant

### LINKED ISSUE #2707: Use disused:shop instead of shop=vacant
If a shop is closed StreetComplete removes all tags and adds [`shop=vacant`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dvacant). This tag has become rather controversial, though. This place is no longer a shop, so it shouldn't get tagged as a shop. Also, every data consumer needs to add a special handling for `shop=vacant`, i.e. for excluding it when rendering shops. Therefore it would be better to use the [lifecycle prefix](https://wiki.openstreetmap.org/wiki/Lifecycle_prefix) instead, i.e. `disused:shop=<former shop value>`.

matkoniecz:
fixes #2707

And yes, it is (now) properly tested.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
@@ -68,7 +68,7 @@ class ShopGoneDialog(
         getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener {
             when (selectedRadioButtonId) {
                 R.id.vacantRadioButton -> {
-                    onSelectedFeature(mapOf("shop" to "vacant"))
+                    onSelectedFeature(mapOf("disused:shop" to "yes"))
                     dismiss()
                 }
                 R.id.replaceRadioButton -> {
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquest/OsmElementQuestType.kt
@@ -42,7 +42,7 @@ interface OsmElementQuestType<T> : QuestType<T> {
      *  - to be part of a(nother) way
      *  - to house a second POI on the same element
      *  - to be a kind of element where deletion is not recommended, (f.e. a shop should rather
-     *    be set to shop=vacant until there is another one)
+     *    be set to disused:shop=yes until there is another one)
      *  ...should be deletable */
     val isDeleteElementEnabled: Boolean get() = false
 


app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
```diff
@@ -68,7 +68,7 @@ class ShopGoneDialog(
         getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener {
             when (selectedRadioButtonId) {
                 R.id.vacantRadioButton -> {
-                    onSelectedFeature(mapOf("shop" to "vacant"))
+                    onSelectedFeature(mapOf("disused:shop" to "yes"))
```
HolgerJeromin, line 71: Reading the code we had:
`shop=supermarket` => `shop=vacant`
and now we have:
`shop=supermarket` => unchanged
**added** `disused:shop=yes`

I think we should do
`shop=supermarket` => `disused:shop=yes`

matkoniecz: definitely, not sure how I managed to miss it

matkoniecz: Or maybe it actually works? Looking again at code it works the same way as other feature replacement (`onSelectedFeature(feature.addTags)`)

Yeah, it works.

`D/QuestController: Replaced NODE #6325607440 in frame of quest AddOpeningHours with DELETE "shop"="gift", DELETE "noname"="yes", ADD "disused:shop"="yes"`

westnordost:
Yes, it works because it ultimatively calls `QuestController::createReplaceShopChanges` which removes **ALL** tags from that element except these:
```kotlin
private val KEYS_THAT_SHOULD_NOT_BE_REMOVED_WHEN_SHOP_IS_REPLACED = listOf(
    "landuse", "historic",
    // building/simple 3d building mapping
    "building", "man_made", "building:.*", "roof:.*",
    // any address
    "addr:.*",
    // shop can at the same time be an outline in indoor mapping
    "level", "level:ref", "indoor", "room",
    // geometry
    "layer", "ele", "height", "area", "is_in",
    // notes and fixmes
    "FIXME", "fixme", "note"
).map { it.toRegex() }
```

westnordost (approved):

westnordost:
By the way, you chose to tag `disused:shop=yes` and not `disused:$previousKey=$previousValue`. Was this deliberate?

matkoniecz:
> By the way, you chose to tag disused:shop=yes and not disused:$previousKey=$previousValue. Was this deliberate?

`$previousValue` is deliberately ignored. See https://github.com/streetcomplete/StreetComplete/issues/2707#issuecomment-811265793 https://github.com/streetcomplete/StreetComplete/issues/2707#issuecomment-816471983

`$previousKey` - it is always shop, right? (and if not then it is not worse anyway as it replaces `shop=vacant`)

westnordost:
Ok, just asking. 

No in some cases it may be an amenity. For example `amenity=pharmacy`

FloEdelmann:
@westnordost There is a typo in the changelog for this PR:

```diff
- Vacant shops are not tagged as disused:shop=*
+ Vacant shops are now tagged as disused:shop=*
```

I'm currently not at a computer, so I can't open a PR directly.

westnordost:
oh damn, the new version is released already

riQQ:
The GitHub release also has the typo

PR #2728, opened 2021-04-10
clarify crossings where there is a blinking traffic light [review of proposed change is welcomed]

matkoniecz:
it is not controlling crossing, so traffic lights is not applicable
but it matches the current answer

And at least I got confused (OSM Wiki check fixed it, but it should not be expected from a typical user)

it is about crossings like

IMG_20210405_121457

draft as in case of acceptance I need to realign photo - or get help to express it in a shorter from

screen01

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/crossing_type/AddCrossingTypeForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/crossing_type/AddCrossingTypeForm.kt
@@ -8,7 +8,7 @@ import de.westnordost.streetcomplete.view.image_select.Item
 class AddCrossingTypeForm : AImageListQuestAnswerFragment<CrossingType, CrossingType>() {
 
     override val items = listOf(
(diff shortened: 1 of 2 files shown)


TurnrDev:
This would be equivalent to the UK Zebra Crossing, right?

matkoniecz:
> This would be equivalent to the UK Zebra Crossing, right?

Roughly similar, main difference will disappear on Summer 2021 with a legislation change. 

Currently pedestrians do not have priority (over wheeled traffic) on marked crossings in Poland.

kmpoppe:
What's the tagging associated with that blinking light?

matkoniecz:
In this quest I would expect "marked crossing" to be selected.

peternewman:
Possibly too subtle, but "Traffic controlled by (traffic) lights" might be more explicit.

TurnrDev:
> What's the tagging associated with that blinking light?

This is what I have used on a similar crossing near me: [](https://www.openstreetmap.org/node/320514912)

westnordost (approved):

Issue #2731, opened 2021-04-10
Switch to using released NSI rather than the latest commit branch?


matkoniecz:
@bhousel about NSI use in https://github.com/streetcomplete/StreetComplete/blob/master/buildSrc/src/main/java/UpdateNsiPresetsTask.kt :

> Also - it makes me nervous when people just pull a file off our main branch.
> NSI publishes releases so downstream projects don't need to worry about breakage.
>
> More info here:
> https://github.com/osmlab/name-suggestion-index/wiki/Downloading-the-Data
>
> They should probably fetch the latest stable presets with a semver-range URL like:
> https://cdn.jsdelivr.net/npm/name-suggestion-index@5/dist/presets/nsi-id-presets.min.json

Likely it would require just changing URL in https://github.com/streetcomplete/StreetComplete/blob/master/buildSrc/src/main/java/UpdateNsiPresetsTask.kt + testing, posting as issue so it will not be forgotten

westnordost:
Yes, that would suffice and yes, I agree.

matkoniecz:
Done by #2733

Thanks to @UKChris-osm (hint: you can have "fixes #2733" in commit - on merge GitHub will automatically close the issue)

PR #2733, opened 2021-04-10
Changed NSI URL to released minimalised CDN copy.

UKChris-osm:
Changed NSI URL to released minimalised CDN copy, as opposed to the direct repository URL #2731

Very new to this repository so my apologies if this is incorrectly pulled.


westnordost:
Thanks!

Issue #2726, opened 2021-04-10
Check for existence of grit bins


arrival-spring:
Grit bins can be removed or moved, so their existence should be checked for every few years.

There are 19000 objects tagged as `amenity=grit_bin`, so plenty of them to check.

matkoniecz:
`amenity=grit_bin` is in iD presets - so it is not blocked like #2679

As far as I see it makes sense

westnordost:
Makes sense, yeah.

matkoniecz:
implemented by @arrival-spring - thanks!

georg-d:
Dear all, does the quest design consider that many grit bins are only visible during winter time but collected & stored centrally during the summer time? In other words, do we avoid the situation that people are asked "does it still exist?" during summer time  and correctly answer "no" but the grit bin is added again few months laterby some mapper?

matkoniecz:
@georg-d Not at all. How such grit bins are marked as present seasonally? With `seasonal=yes` or other value of `seasonal` tag? (it would be possible to exclude grit bins tagged as seasonal)

(if it is not indicated at all, I would consider this to be quite problematic way of tagging)

georg-d:
@matkoniecz https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dgrit_bin does not suggest any tags concerning season. But as these features are obviously only required in cold time, personally, I never even thought of marking them explicitly as seasonal - it seemed like explicitly telling plant pots may be moved to the inside during winter, or equipment of open air bathing areas (volley ball nets, pontons for swimmers,...). 

Back to the quest: Would it be difficult to show the quest only during "winter time" of current position/region? If no, that would be an effective solution. Alternatively, we may simply refine the wording, e.g. extending to "Does it still exist? (grit bin) Note: May be present only in cold season."

westnordost:
Yes, it would not be trivial. Then maybe remove checking for the existence of grit bins again.

matkoniecz:
I posted https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dgrit_bin#Marking_grit_bins_placed_only_during_winter.3F and will commit removal of grit bin checks :(

peternewman:
It's not worth doing this by country is it? In the UK (or at least England) they are present 365 days of the year.

matkoniecz:
Probably no.

georg-d:
[...]

PR #2734, opened 2021-04-10
Check for grit bins every 4 years

arrival-spring:
4 years seems like a sensible frequency to check these
Clsoes #2726

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
@@ -32,6 +32,7 @@ class CheckExistence(
             or amenity = bench
             or amenity = waste_basket
             or amenity = post_box
+            or amenity = grit_bin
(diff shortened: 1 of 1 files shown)


arrival-spring:
Closes #2726

matkoniecz:
Looks obvious, but have you tested it?

HolgerJeromin:
> Closes #2726

This needs to be in the first comment.

arrival-spring:
In that it shows up for a grit bin near me that was last edited more than 5 years ago

matkoniecz:
> In that it shows up for a grit bin near me that was last edited more than 5 years ago

Ideally you would also solve this quest and see in logs is it acting as expected (in this case bug in ridiculously unlikely but I found some when I tested "obvious" changes).

I see in logs `D/QuestController: Solved a CheckExistence quest: ADD "check_date"="2021-04-10"` and `D/QuestController: Deleted NODE #4029804614 in frame of quest CheckExistence` so it seems working as expected :)

PR #2737, opened 2021-04-12
Add sports pitch lighting quest

### LINKED ISSUE #2639: Sport pitch lighting
Since the app is already asking for the sport and surface of a pitch, I thought asking users for lit=yes/no on pitches would be a good idea for a new quest.

The question would be: Is this sport pitch lit? or Does this sport pitch have floodlights?

Useful for football pitches, skateparks etc.

eginhard:
Fixes #2639.

This is basically just a mix of the WayLit and PitchSurface quests. I noticed that the pitch surface quest is restricted to certain sports where it makes sense. I don't think that restriction is required for lighting.


app/src/main/java/de/westnordost/streetcomplete/quests/pitch_lit/PitchLit.kt
```diff
@@ -0,0 +1,10 @@
+package de.westnordost.streetcomplete.quests.pitch_lit
+
+enum class PitchLit(val osmValue: String) {
+    NIGHT_AND_DAY("24/7"),
+    AUTOMATIC("automatic"),
```
westnordost, line 5: I don't think these two are reasonable values for this.

24/7 is meant for for example lighting in tunnels (or indoors). But with the filter above, you already filter out any pitches that may be indoor.

automatic doesn't really make sense for pitches: It turns on when something moves? Oh, better not standon the tennis court, otherwise the lights go off again!

So, this could be a Boolean. Yes, no

westnordost (changes requested):
Looks good, just one thing:

app/src/main/java/de/westnordost/streetcomplete/quests/pitch_lit/PitchLit.kt
eginhard: Yes, that makes sense. I removed those values.

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/pitch_lit/AddPitchLit.kt
```diff
@@ -0,0 +1,38 @@
...
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.ktx.toYesNo
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+
+class AddPitchLit : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
+        ways with (leisure=pitch or leisure=track)
+        and (access !~ private|no)
+        and indoor != yes and !building
```
FloEdelmann, line 15: Maybe `(!building or building = no)`?

`building=no` is [used around 10k times](https://taginfo.openstreetmap.org/tags/building=no).

eginhard: Thanks. [At least 300](https://overpass-turbo.eu/s/162f) of those are `leisure=pitch`. Should I update https://github.com/streetcomplete/StreetComplete/blob/a807e450698956ef6bf70eb2d1d376005656c03f/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt#L29 as well?

eginhard: Done.

matkoniecz: > building=no is used around 10k times.

I think that it can be safely ignored, given that it is not increasing ( https://taginfo.openstreetmap.org/tags/building=no#chronology ) and we have over 400 000 000 buildings increasing constantly - https://taginfo.openstreetmap.org/keys/building#chronology

westnordost (approved):

FloEdelmann (approved):

PR #2741, opened 2021-04-13
Userproofing team mode - provide more explicit info

matkoniecz:
ping @FloEdelmann

I think that this extra info is worth more filled screen and extra translating time but I want others to review that change.

Without that I see easily one person enabling team mode and expecting that it will work for others.

screen


app/src/main/res/layout/dialog_team_mode.xml
```diff
@@ -75,5 +75,12 @@
         tools:listitem="@layout/cell_team_mode_color_circle_select"
         tools:visibility="visible" />
 
+    <TextView
```
FloEdelmann, line 78: Why is this an extra text block that is always shown at the very bottom?

To save space, I'd rather …

1. … append it to the text at the top:
	> When mapping the same area together in a group, you can enter team mode to split up quests between all team members. Every team member needs to enable it on their own device. Remember to disable it after shared mapping!
2. … leave it at the bottom, but only show it when no team size is entered yet.
3. … leave it at the bottom, but only show it after a color is selected.

app/src/main/res/values/strings.xml
```diff
@@ -835,6 +835,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="team_mode">Team Mode</string>
     <string name="team_mode_exit">Exit Team Mode</string>
     <string name="team_mode_description">When mapping the same area together in a group, you can enter team mode to split up quests between all team members.</string>
+    <string name="team_mode_description_additional_tutorial">Every team member needs to enable team mode on their device. Remember to disable it after shared mapping!</string>
```
FloEdelmann, line 838: I'm no native speaker, but "shared mapping" sound wrong. @peternewman Can you help here?

```suggestion
    <string name="team_mode_description_additional_tutorial">Every team member needs to enable team mode on their device. Remember to disable it after mapping together!</string>
```

FloEdelmann (commented):
I'm not sure if it's actually needed. But I'm fine with adding it if it clears things up.

app/src/main/res/values/strings.xml
peternewman: Yeah @FloEdelmann 's "mapping together" feels better to me and echos some of the wording above, without the potential for confusion of the social media usage of "share via...".

It's probably neater than stuff I could come up with too like "...after simultaneous mapping!" or collaborative.

app/src/main/res/layout/dialog_team_mode.xml
westnordost: I think option "1 + it hides together with the text at the top after the team size has been entered" or option 3 is best.

FloEdelmann: @matkoniecz What's the current state of this PR?

matkoniecz: > Why is this an extra text block that is always shown at the very bottom?

As it is an extra info that is explaining use, so putting it at the end seemed logical.

> hides together with the text at the top after the team size has been entered

Given situation that triggered this PR - it is likely that it would be missed. Also, someone processing it sequentially: read message on top, answer question would never read it.

> … leave it at the bottom, but only show it after a color is selected.

I will implement this one.

> @matkoniecz What's the current state of this PR?

I am trying to finish it right now to close started, but not completed projects.

matkoniecz: As I worried - it works poorly for large team sizes.
screen05

matkoniecz:
Ready for review, but I know that it likely should be changed.

westnordost:
Yes. I don't understand this PR at all anyway. The extra text doesn't really add clarity IMO.

Would anyone think that team mode must be enabled only on one device and the rest are automagically somehow included? That doesn't make sense. If it is really necessary to point that out, why not change the wording to 

"Now everybody in your group chooses a unique color **on their phone**"

app/src/main/res/values/strings.xml
peternewman: @matkoniecz this seems to have either been reverted or missed at some point, as it still says shared mapping.

matkoniecz:
> Yes. I don't understand this PR at all anyway. The extra text doesn't really add clarity IMO.

In discussion with an actual person they were confused how team mode works - and they were confused because one person enabled it and was unaware that other person also must enable it rather than waiting for a magical communication.

(I encountered this after this PR was opened from what I remember, PR speaks about hypothethical)

FloEdelmann:
> "Now everybody in your group chooses a unique color **on their phone**"

Please also consider the original wording (or a variation thereof):
"Everybody in the group has to choose a unique color (on their own phone) now."

It was changed in [`f79f783` (#2557)](https://github.com/streetcomplete/StreetComplete/pull/2557/commits/f79f783ab63f9b0e218a5b7599ac3fddf3bf75f9#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103L833-R833).

westnordost:
Is this PR superceded by #3170?

matkoniecz:
yes, I will close it now

matkoniecz:
Hopefully #3170 will be sufficient, lets see whether more confused people will be encountered

Issue #2745, opened 2021-04-14
Maxspeed quest should tag a road as lit when that is asked


arrival-spring:
In the UK the speed limit can depend on whether or not there is street lighting, so when a user selects "No sign, defaults apply" and then "Yes, no sign" they are then asked if the road is lit (if that hasn't been set already).

However this does not set `lit` on the road, and it feels like it should have done.

westnordost:
I guess nothing speaks against this.

arrival-spring:
However when working on this I realised that the quest doesn't quite work in the UK in all situations.

Single and dual carriageway roads (`maxspeed:type=nsl_single` or `nsl_dual`) can be lit and should still have that `maxspeed:type` set, at the moment this assumes that any lit road is `nsl_restricted`.

See [maxspeed:type#UK](https://wiki.openstreetmap.org/wiki/Key:maxspeed:type#UK) on the wiki:
`nsl_restricted` means 30 mph for cars
`nsl_single` means 60 mph for cars
`nsl_dual` means 70 mph for cars

So it's quite important to get the difference between these right.

At the moment, if I was asked for the maxspeed of a single carriageway national speed limit road I would want the app to set at least `maxspeed:type=nsl_single` and probably also `maxspeed=60 mph`, whether or not the road is lit.

I don't have an option that leads to that though. If I say that the speed limit is defined by a sign then I have to enter a numeric value for the sign, but the national speed limit sign is white with a black diagonal line (main picture on the  [maxspeed:type wiki page](https://wiki.openstreetmap.org/wiki/Key:maxspeed:type)) and this wouldn't add the correct `maxspeed:type`.

If there is a single carriageway lit road that is national speed limit then if I say that there is no sign and defaults apply then the app will add `maxspeed:type=nsl_restricted` which implies 30 mph, but it should be `nsl_single` and 60 mph.

Perhaps the solution to this is another top level option in the UK which is "National Speed Limit" (all UK drivers will know what this means), this should then ask if it's single or dual carriageway and add `nsl_single` or `nsl_dual` as appropriate, and would ideally have an image of the sign to confirm.

I would also suggest that (in the UK) when implicit maxspeeds are set that the appropriate maxspeed is also set.

peternewman:
> However this does not set `lit` on the road, and it feels like it should have done.

[...]

PR #2747, opened 2021-04-15
Add lit tag in UK for implicit maxspeed

arrival-spring:
#2745 (what it was originally opened for)

Comments in the file explain why it's done like this.

A later PR with national speed limit option will change how this works when there are other ways of adding `nsl_single` or `nsl_dual`.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeed.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeed.kt
@@ -56,6 +56,16 @@ class AddMaxSpeed : OsmFilterQuestType<MaxSpeedAnswer>() {
                 changes.modify("highway", "living_street")
             }
             is ImplicitMaxSpeed -> {
+                if (answer.countryCode == "GB") {
+                    // nsl_restricted only gets added if the road is lit (whether already set or answered by the user in this quest)
+                    if (answer.roadType == "nsl_restricted") {
+                        changes.add("lit", "yes")
+                    }
+                    // The other answers, nsl_single and nsl_dual, only get added if the road is not lit
+                    else {
+                        changes.add("lit", "no")
+                    }
+                }
                 changes.add("maxspeed:type", answer.countryCode + ":" + answer.roadType)
             }
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeed.kt
```diff
@@ -56,6 +56,16 @@ class AddMaxSpeed : OsmFilterQuestType<MaxSpeedAnswer>() {
                 changes.modify("highway", "living_street")
             }
             is ImplicitMaxSpeed -> {
+                if (answer.countryCode == "GB") {
+                    // nsl_restricted only gets added if the road is lit (whether already set or answered by the user in this quest)
+                    if (answer.roadType == "nsl_restricted") {
+                        changes.add("lit", "yes")
```
matkoniecz, line 62: what is happening in case where lit tag is present already? (maybe `addOrModify` should be used or maybe not)

arrival-spring: Oh yes, I realised that later, but was in the branch for adding national speed limit

app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeed.kt
```diff
@@ -56,6 +56,16 @@ class AddMaxSpeed : OsmFilterQuestType<MaxSpeedAnswer>() {
...
+                    }
+                    // The other answers, nsl_single and nsl_dual, only get added if the road is not lit
+                    else {
+                        changes.add("lit", "no")
+                    }
```
westnordost, line 67: but couldn't an `nsl_dual` be lit?

arrival-spring: Yes, and that's what I'm working on now for the National Speed Limit option, but at the moment `nsl_dual` only gets added if you say the road is not lit:
[AddMaxSpeedForm.kt#216](https://github.com/streetcomplete/StreetComplete/blob/3645254bea414b0b9275b68203fa96ee2fce8f57/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt#L216-L221)

westnordost:
Does the PR still make sense after the changes you made in the other PR`?

arrival-spring:
Not really, there are various issues with how it is now, and this doesn't actually fix anything, it just makes it so the user doesn't have to answer whether the road is lit or not twice.

westnordost:
So should this be closed, or shall it be included at least in the next minor update?

arrival-spring:
I'm happy for this to be closed

PR #2748, opened 2021-04-15
National Speed Limit option for maxspeed quest in GB

### LINKED ISSUE #2745: Maxspeed quest should tag a road as lit when that is asked

arrival-spring:
Technically all seems to be working fine, but I'm probably missing something obvious and can't work out why the sign isn't at the top like the living street sign.

Maxspeed1
Maxspeed2
Maxspeed3

There doesn't seem to be attribution for other speed signs, and the NSL sign [is public domain](https://commons.wikimedia.org/wiki/File:UK_traffic_sign_671.svg) but I can add something if needed.

Closes #2745


westnordost:
Hm, how is the user supposed to know if the "National Speed Limit" applies to a road or "defaults apply"? The two options seem contradictory, as the national speed limit is kind of the "default speed limit".

arrival-spring:
> Hm, how is the user supposed to know if the "National Speed Limit" applies to a road or "defaults apply"? The two options seem contradictory, as the national speed limit is kind of the "default speed limit".

Well, yes.

In many places, there are no signs at all (so defaults apply).

- On single carriageway roads without street lighting (normally in rural areas), where there are no other signs, the speed limit is 60 mph (`nsl_single`) and there will be no signs except for where there is a change between speed limits (so you could be miles from the nearest sign).
- On single carriageway roads with street lighting and no other signs, the speed limit is 30mph (`nsl_restricted`) and again there are only signs where there is a change, so if you're in the centre of town then there are no speed limit signs, but you know it's 30mph.
- On single (and some dual) carriageway roads with street lighting and which are national speed limit (`nsl_single`) there will be repeater signs of the nsl sign (white with black stripe) every few lamp posts.
- Dual carriageway roads are always 70mph (`nsl_dual`) unless there are other signs, and will only have the nsl sign where there is a change in speed limit (or sometimes repeated on street lights if it's lit).

Most people do not consider that the 30mph speed limit because of street lights is "National Speed Limit", even if it technically is a form of that - most people consider "National Speed Limit" to be the same as the nsl sign, meaning 60mph on single carriageway and 70mph on dual carriageway (and different speeds for lorries, coaches and towing vehicles).

So sometimes there are no signs and it's nsl, sometimes there are no signs and it's `nsl_restricted`, sometimes there are 30 mph signs, sometimes there are nsl signs. Hence why it probably makes sense to have both.

And I've just that although motorways are also 70mph for cars, they have different speed limits for other vehicles compared to dual carriageways, so should have `GB:motorway` set. I'll cater for that too.

arrival-spring:
> And I've just that although motorways are also 70mph for cars, they have different speed limits for other vehicles compared to dual carriageways, so should have `GB:motorway` set. I'll cater for that too.

Never mind, realised that there's `ROADS_WITH_DEFINITE_SPEED_LIMIT` that covers motorways.

arrival-spring:
As there can be lit dual carriageways which should be `nsl_dual` I've made it to ask single or dual carriageway first, then whether or not it's lit.

This is now all working as intended and I've tested that all changes are tags are being added as expected.

The only issue I can see now is the NSL image being wrong.

app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeed.kt
```diff
@@ -56,7 +56,31 @@ class AddMaxSpeed : OsmFilterQuestType<MaxSpeedAnswer>() {
...
+                            changes.add("maxspeed", "60 mph")
+                        }
+                        "nsl_dual" -> {
+                            changes.add("maxspeed", "70 mph")
+                        }
+                        "motorway" -> {
+                            changes.add("maxspeed", "70 mph")
+                        }
+                    }
+                }
```
westnordost, line 74: I don't want to have this in the app. I have reasoned about this in the past multiple times. In a nutshell:
- statuary speed limits could change anytime (see f.e. France a few years ago). It then would tag it wrong until an app update is made
- having a list of all default speed limits in the app for every country is unmaintainable. If it is added for one country, people will start asking why it can not also be added for their country etc.
- smart data consumers will need to have the ability to infer actual speed limits from `maxspeed:type` (etc) anyway, and there is a list for that in the wiki, even.

app/src/main/res/drawable/ic_national_speed_limit.xml
```diff
@@ -0,0 +1,17 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="342.05dp"
+    android:height="342.05dp"
```
westnordost, line 3: Looks a bit big to me. Best look at the size of the other signs and adapt it.

Anyway, I really wonder why the path data is so big for that sign, isn't it just a circle with a stroke through it?

westnordost (changes requested):
Since the maxspeed quest is so important, could you please add unit tests for the changed functionality as well?

app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeed.kt
arrival-spring: That makes sense, I'll take it out

app/src/main/res/values/strings.xml
```diff
@@ -231,6 +231,7 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_maxspeed_answer_sign">"A sign"</string>
     <string name="quest_maxspeed_answer_zone2">"It's within a (slow) zone"</string>
     <string name="quest_maxspeed_answer_noSign2">"No sign, defaults apply"</string>
+    <string name="quest_maxspeed_answer_nsl">"National Speed Limit"</string>
```
westnordost, line 234: Should this be translated at all? F.e. if I am on a vacation in England, should it say "Nationale Höchstgeschwindigkeit" for me or should it stay untranslated? I think maybe it should stay untranslated, in this case it should go into `untranslatableStrings.xml`

arrival-spring: Hmm, possibly untranslated. You'll never see "National Speed Limit" on a sign anywhere, you'll just see the sign. But it is also a technical, official term.

arrival-spring:
> Since the maxspeed quest is so important, could you please add unit tests for the changed functionality as well?

For the tests, I can't see an easy way of emulating the user input (when the user selects single/dual carriageway and then lit/not lit), so should I add a new class for NSL?

arrival-spring:
Not sure if it needed that many tests, but they're added now.
I think this is all done, other than the open question about translating the string or not.

westnordost:
Then let's not translate it

arrival-spring:
Done

westnordost:
Nice! Then I can merge it now, thank you!

PR #2751, opened 2021-04-16
remove crossings with cycle-only ways from sound and vibration quests

Helium314:
For the tactile pavings quest, crossings with cycle-only ways are already excluded, so I simply copied and slightly adjusted the relevant parts to the sound and vibration quests.

Tested and works as expected: The sound signal quest does indeed not show up on crossings with cycleways with foot=no (without this PR it does for the same corssing)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_sound/AddTrafficSignalsSound.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_sound/AddTrafficSignalsSound.kt
@@ -1,28 +1,48 @@
 package de.westnordost.streetcomplete.quests.traffic_signals_sound
 
+import de.westnordost.osmapi.map.MapDataWithGeometry
+import de.westnordost.osmapi.map.data.Element
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
 import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
-import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmElementQuestType
 import de.westnordost.streetcomplete.ktx.toYesNo
 import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
 
-class AddTrafficSignalsSound : OsmFilterQuestType<Boolean>() {
+class AddTrafficSignalsSound : OsmElementQuestType<Boolean> {
 
-    override val elementFilter = """
+    private val crossingFilter by lazy { """
         nodes with crossing = traffic_signals and highway ~ crossing|traffic_signals and foot != no
         and (
           !$SOUND_SIGNALS
           or $SOUND_SIGNALS = no and $SOUND_SIGNALS older today -4 years
           or $SOUND_SIGNALS older today -8 years
         )
-    """
+    """.toElementFilterExpression() }
+
+    private val excludedWaysFilter by lazy { """
+        ways with
+          highway = cycleway and foot !~ yes|designated
+    """.toElementFilterExpression() }
 
     override val commitMessage = "Add whether traffic signals have sound signals"
     override val wikiLink = "Key:$SOUND_SIGNALS"
     override val icon = R.drawable.ic_quest_blind_traffic_lights_sound
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_traffic_signals_sound_title
+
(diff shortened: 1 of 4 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_vibrate/AddTrafficSignalsVibration.kt
```diff
@@ -1,28 +1,48 @@
...
+        val excludedWayNodeIds = mutableSetOf<Long>()
+        mapData.ways
+            .filter { excludedWaysFilter.matches(it) }
+            .flatMapTo(excludedWayNodeIds) { it.nodeIds }
+
+        return mapData.nodes
+            .filter { crossingFilter.matches(it) && it.id !in excludedWayNodeIds }
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? = null
```
westnordost, line 44: This could return false at least if the crossingFilter does not match. This improves performance when unlocking quests after solving one.

app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_sound/AddTrafficSignalsSound.kt
```diff
@@ -1,29 +1,49 @@
...
+        val excludedWayNodeIds = mutableSetOf<Long>()
+        mapData.ways
+            .filter { excludedWaysFilter.matches(it) }
+            .flatMapTo(excludedWayNodeIds) { it.nodeIds }
+
+        return mapData.nodes
+            .filter { crossingFilter.matches(it) && it.id !in excludedWayNodeIds }
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? = null
```
westnordost, line 45: here too

app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_vibrate/AddTrafficSignalsVibration.kt
```diff
@@ -1,28 +1,48 @@
...
           !$VIBRATING_BUTTON
           or $VIBRATING_BUTTON = no and $VIBRATING_BUTTON older today -4 years
           or $VIBRATING_BUTTON older today -8 years
         )
-    """
+    """.toElementFilterExpression() }
+
+    private val excludedWaysFilter by lazy { """
+        ways with
+          highway = cycleway and foot !~ yes|designated
```
westnordost, line 25: There are many other combinations with foot - permissive etc.
Does it make sense to include those? 🤷

westnordost (changes requested):
Nice, you even wrote tests. Top!

Two comments though:

app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_vibrate/AddTrafficSignalsVibration.kt
Helium314: Hmm... I guess a crossing with foot=permissive cycleway is unlikely to have sound signals as it's mainly for cyclists
plus, this way it's constistent with the tactile paving quest: https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt#L30

westnordost: ok

app/src/main/java/de/westnordost/streetcomplete/quests/traffic_signals_vibrate/AddTrafficSignalsVibration.kt
Helium314: you mean something like
```
override fun isApplicableTo(element: Element): Boolean? = if (crossingFilter.matches(element))
    false
else
  null
```
?

Helium314: well, actually it should be `if (!crossingFilter.matches(element))`

westnordost: yes

Helium314: Alright, I will add it.
Same could als be done for tactile pavings and crossing island. Should I add it there as well?

westnordost: No, it is already done in the dev branch

westnordost (approved):

westnordost:
Looks good, thank you!

Issue #2752, opened 2021-04-16
Answering that traffic calming is gone should not delete nodes


matkoniecz:
**How to Reproduce**
I enabled short resurvey interval (may be not needed) and scanned around https://www.openstreetmap.org/node/5254567356

I answered "it is gone"

> D/QuestController: Deleted NODE #5254567356 in frame of quest CheckExistence

It should not happen as it may change road geometry.

**Versions affected**
9d8b615164700455a6762f2341fb7fa1523556ac (Android should not be relevant here)

matkoniecz:
not checked: maybe nodes are deleted only when road is straight and node is i the middle of ways (without crossing between various ways) and deletion would not affect anything? That is unlikely, as it seems to be not worth extra complexity.

westnordost:
did you check that it actually deleted the node and not just stripped it off its tags?

matkoniecz:
> did you check that it actually deleted the node and not just stripped it off its tags?

No, I just looked into log and have not investigated further. Is such deletion converted later into tag blanking?

westnordost:
yes, see `DeleteSingleOsmElementUploader` on master or `DeletePoiNodeAction` on the dev branch

matkoniecz:
OK, in such case I got scared by this log entry and things are working as expected (maybe except log message)

PR #2753, opened 2021-04-16
Barrier, fixes #2739

### LINKED ISSUE #2739: Change barrier=yes on nodes into more specific values
### General
Affected tag(s) to be modified/added: [barrier](https://wiki.openstreetmap.org/wiki/Key:barrier)
Question asked: **What kind of barrier is here?**

The wiki says on [barrier=yes](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dyes) "Unspecific barrier value, it is always better to use more specific one if possible. Still, sometimes when mapping from aerial images or surveying something in distance it may be necessary to use it."

### Checklist
crucial for routing
armchair mappers may add barrier=yes. iD even adds this automatically if a user opens the "barriers" menu, but then doesn't choose a specific type.

### Ideas for implementation

**Element selection:** 
nodes with `barrier=yes|unknown`
Maybe even limit it to nodes which are part of a `highway` way, because they're the ones important for routing.

**Metadata needed:** 
Metadata is not necessary. But, like in the orchard quest, a pre-selection of common values would be nice. There are a lot of possible values. Data may be derived from current use of the tags. Is there a source to get such tag usage statistics per country? Or a way to create the statistics with a script?

**Proposed GUI:**
list/grid to select from. Either with symbols (like the building type quest) or with photos (like the crop quest). I don't know what is the better UI design.
On the wiki there are already photos for nearly all of the values.

Answers:
[bollard](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbollard)
[chain](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dchain)
[jersey barrier](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Djersey_barrier)
([kerb](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dkerb)) is it imaginable that someone can see a kerb from aerial imagery but cannot determine that it is a kerb?
[log](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dlog)
[rope](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Drope)

[block](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dblock)
([bump gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbump_gate))
([bus trap](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbus_trap))
[cattle grid](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcattle_grid)
[cycle barrier](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcycle_barrier)
[debris](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Ddebris)
[entrace](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dentrance)
[HEET](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dfull-height_turnstile)
[gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dgate)
[sliding gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dsliding_gate)
[hampshire gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dhampshire_gate)
[height restrictor](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dheight_restrictor)
([horse stile](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dhorse_stile))
([kent carriage gap](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dkent_carriage_gap)) UK-specific according to the wiki
[kissing gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dkissing_gate)
[lift gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dlift_gate)
([motorcycle barrier](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dmotorcycle_barrier))
[sally port](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dsally_port)
[spikes](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dspikes)
[stile](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dstile)
[sump buster](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dsump_buster)
([swing gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dswing_gate)) (isn't that just a gate?)
[turnstile](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dturnstile)
[wicket gate](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dwicket_gate)

A similar quest was once suggested for ways (#2100). It wasn't followed up on because some linear barriers require a specific direction of the way. But this problem doesn't affect nodes, which this quest suggestion is about.

matkoniecz:
## After merge

- [ ] after merging translate as curb in UK English ("In American and Canadian English, the noun meaning the edge of a sidewalk or roadway is spelled curb. In varieties of English from outside North America, the word is spelled kerb. ")

## WONTFIXed unless someone has a good idea
- [ ] chain on bollards can be confusing to people who may select bollard instead of a chain - no idea how to deal with it
- [ ] is there a way to use some very short descriptive term instead of "kissing gate" that is unclear even with an image for many people?  - no idea how to deal with it

## Done
- [x] icon (@westnordost is there a chance that you will make one?)
- [x] review of selected images and labels (comments welcomed!)
- [x] go/no go for `barrier=curb` (either deletion or a proper crop of a current is needed) (kept, can be easily convinced to delete it)
- [x] get stepover stile image with margin at the bottom, so stile itself will not be obstructed by text
- [x] ladder stile with a clearly visible wall/fence and a squeezer stile with a clearly visible path also would be nice
- [x] attempt to get better photos for some (all except this one are at least kind of acceptable)
- [x] specifying licences/authors of files (and removal of notes about sources from source code) - after possible image replacements
- [x] verify that sources and licences match
- [x] selecting it is not present anymore deletes node, even if it is a part of way geometry (variant of #2752 - so maybe it is not actually present problem)
- [x] merge history to avoid having 10 variants of each image present in the history, some of them without attribution
- [x] check is it necessary to exclude some `barrier=yes` nodes
- [x] Is "kissing gate" a clear term or something deeply technical and not clear to a typical person? If it should be replaced - how to describe it more clearly?
- [x] review #2739 comments have I missed something
- [x] add to achievements (I can probably go ahead and add to rare ones, I am betting that this `barrier=yes` are HOT tagged somewhere and real occurrences are really rare)
- [x] <s>using squeezer stile image from https://www.geograph.org.uk/photo/4255454 rather https://www.geograph.org.uk/photo/110502</s> - rejected for now

And before people will get too excited: note that this quest will be asked rarely, only on `barrier=yes` nodes (due to importance of that information I still think it worth implementing, despite that it adds just about 5-10 quests in the entire city).

screen01
screen02


DerDings:
What about [putting the entrance at first place](https://github.com/streetcomplete/StreetComplete/issues/2739#issuecomment-820701183)?

HolgerJeromin:
I do not like the image of 

* entrance (too much focus on the behind and not the "opening/surrounding barrier") and 
* cycle barrier (confusing 10 white bars).
 
On both wiki pages there is a clearer setup of the feature on the images.

matkoniecz:
in general: if someone has a better image available on an open license - feel free to mention it!

For example I know that hampshire gate (wire gate) image is substandard, but I failed to find anything better and this is acceptable.

Especially something better for stepover stile would be welcomed. I will search again, ping someone who mentioned that they have plenty of stile images and likely ask talk-gb for images if that will fail.

> cycle barrier (confusing 10 white bars).

I can try again, but this was really best one that I found (trying to find something that would have decent contrast, clearly distinct from stile, without level crossing on image and on open license and with place at bottom for text turned out to be challenging and I am relatively happy about this one)

> entrance (too much focus on the behind and not the "opening/surrounding barrier") and

I hoped for fitting into a pretty image, but that may be detracting overall. I will try to search for replacement, but not for hours.

matkoniecz:
> On both wiki pages there is a clearer setup of the feature on the images.

Not after you crop https://wiki.openstreetmap.org/wiki/File:Archway_between_walled_gardens,_Greenway_-_geograph.org.uk_-_191221.jpg into a square

I will reconsider https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcycle_barrier

DerDings:
For the cycle_barrier one, maybe I'm able to take some photos this weekend.
I also know a place where a plastic jersey barrier was joined to a concrete one, which would make the perfect generic picture. But I don't know if it's still there.

matkoniecz:
@DerDings Photos would be welcomed! Note that here photos are cropped into square and bottom is used for name - I needed to reject some images due to lacking bottom margin (see stepover stile case) 

Even in the worst case they still can be added to available selection of photos and be available to others.

matkoniecz:
And before people will get too excited: note that this quest will be asked rarely, only on `barrier=yes` nodes (due to importance of that information I still think it worth implementing, despite that it adds just about 5-10 quests in the entire city).

HolgerJeromin:
>> On both wiki pages there is a clearer setup of the feature on the images.

> Not after you crop https://wiki.openstreetmap.org/wiki/File:Archway_between_walled_gardens,_Greenway_-_geograph.org.uk_-_191221.jpg into a square

Sorry, should have been more precise. I meant:
https://wiki.openstreetmap.org/wiki/File:Entrance.png
But you are right, this has no (proper) licence and the user was last active 5 years ago. 8-/

matkoniecz:
> What about putting the entrance at first place?

Done, why not. I have no strong preferences for the initial sorting.

matkoniecz:
>     cycle barrier (confusing 10 white bars).
> On both wiki pages there is a clearer setup of the feature on the images.

I am testing it right now, if I will not consider result noticeably worse I will keep it.

screen04

I think that it is a bit worse, but of people like it I can keep it.

I am also thinking about using squeezer stile image from https://www.geograph.org.uk/photo/4255454 rather https://www.geograph.org.uk/photo/110502 but not convinced that improvement is significant so I will keep it for now

westnordost:
fyi I added the icon to master

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -702,6 +707,7 @@ object AchievementsModule {
                     "AddCyclewayPartSurface",
                     "AddFootwayPartSurface",
                     "AddBenchBackrest",
+                    "AddBarrierType",
```
westnordost, line 710: IMO that's a bit much. I think "for cars" and "outdoors" could be removed.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierType.kt
```diff
@@ -0,0 +1,46 @@
...
+                changes.addOrModify("stile", "squeezer")
+            }
+            BarrierType.STILE_LADDER -> {
+                changes.addOrModify("stile", "ladder")
+            }
+            BarrierType.STILE_STEPOVER -> {
+                changes.addOrModify("stile", "stepover")
+            }
+            else -> {
+            };
```
westnordost, line 42:
```suggestion
```

westnordost: Also, a simple unit test for the logic here would be good but not required. (Just a few lines, to test the stile-stuff.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierTypeForm.kt
```diff
@@ -0,0 +1,40 @@
...
+        Item(LOG, R.drawable.barrier_log, R.string.quest_barrier_type_log),
+        Item(KERB, R.drawable.kerb_height_raised, R.string.quest_barrier_type_kerb), // TODO: translate it as curb in UK!
+        Item(HEIGHT_RESTRICTOR, R.drawable.barrier_height_restrictor, R.string.quest_barrier_type_height_restrictor),
+        Item(FULL_HEIGHT_TURNSTILE, R.drawable.barrier_full_height_turnstile, R.string.quest_barrier_full_height_turnstile),
+        Item(TURNSTILE, R.drawable.barrier_turnstile, R.string.quest_barrier_type_turnstile),
+        Item(DEBRIS_PILE, R.drawable.barrier_debris_pile, R.string.quest_barrier_type_debris_pile),
+        Item(STILE_SQUEEZER, R.drawable.barrier_stile_squeezer, R.string.quest_barrier_type_stile_squeezer),
+        Item(STILE_LADDER, R.drawable.barrier_stile_ladder, R.string.quest_barrier_type_stile_ladder),
+        Item(STILE_STEPOVER, R.drawable.barrier_stile_stepover, R.string.quest_barrier_type_stepover),
+        Item(KISSING_GATE, R.drawable.barrier_kissing_gate, R.string.quest_barrier_type_kissing_gate), //Is "kissing gate" a clear term or something deeply technical? If it should be replaced - how to describe it more clearly?
```
westnordost, line 31: 💋

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
matkoniecz: I think that for both it actually matters and matches - for example for cars betwen `barrier=height_restrictor` vs `barrier=gate`. And even between different `barrier=stile` variants there is a difference.

But I can remove if you want.

matkoniecz:
> fyi I added the icon to master

Noticed, thanks! Already pulled it and used. Much better than sad fence that I considered, despite not really fitting for nodes.

I think that I already said it, but I like your art style :)

westnordost (approved):
The code is clean, the pictures are all good! Surely a lot of time went into researching which values to display and to find suitably licensed and good pictures for it. 

Great work!

westnordost:
> I think that I already said it, but I like your art style :)

Ha, I just lazily copied from the construction site icon and changed the colors! But thanks anyway

matkoniecz:
> Ha, I just lazily copied from the construction site icon and changed the colors! But thanks anyway

Well, if you did something that worked well and it was easy for someone - that is just more impressive :)

> The code is clean, the pictures are all good! Surely a lot of time went into researching which values to display and to find suitably licensed and good pictures for it.

Yeah, proportions were: 1h coding. 0.5h testing, 1h selecting values, 7.5h selecting images, 1h responding on GitHub, 2h doing various related things like research for the new stile quest or `barrier`-related clean-up of wiki documentation.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierTypeForm.kt
peternewman: > Is "kissing gate" a clear term or something deeply technical? If it should be replaced - how to describe it more clearly?

From my experience it's used in fairly common parlance within the UK (or at least within outdoorsy people). Probably the name helps it be easily remembered compared to some of the other more niche things on that list.

peternewman: I've also seen an inverted kissing gate once in Luxembourg (and I think I found the name for it at the time but can't now). The gate bit is actually a V, which swings either side of a central fixed post, so the means of getting through the gate was the same idea as a kissing gate, but the part that moved was reversed.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierType.kt
```diff
@@ -0,0 +1,44 @@
+package de.westnordost.streetcomplete.quests.bike_parking_type
```
FloEdelmann, line 1:
```suggestion
package de.westnordost.streetcomplete.quests.barrier_type
```

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
```diff
@@ -54,6 +54,7 @@ import de.westnordost.streetcomplete.quests.oneway_suspects.data.TrafficFlowSegm
 import de.westnordost.streetcomplete.quests.oneway_suspects.data.WayTrafficFlowDao
 import de.westnordost.streetcomplete.quests.opening_hours.AddOpeningHours
 import de.westnordost.streetcomplete.quests.atm_operator.AddAtmOperator
+import de.westnordost.streetcomplete.quests.bike_parking_type.AddBarrierType
```
FloEdelmann, line 57:
```suggestion
import de.westnordost.streetcomplete.quests.barrier_type.AddBarrierType
```

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierTypeForm.kt
```diff
@@ -0,0 +1,40 @@
+package de.westnordost.streetcomplete.quests.barrier_type
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.quests.AImageListQuestAnswerFragment
+import de.westnordost.streetcomplete.quests.bike_parking_type.BarrierType.*
```
FloEdelmann, line 5:
```suggestion
import de.westnordost.streetcomplete.quests.barrier_type.BarrierType.*
```

FloEdelmann (changes requested):
Two more references to the old package

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
matkoniecz: Thanks! I already commited it (not pushed), together with one more reference.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierTypeForm.kt
matkoniecz: Thanks! I already commited it (not pushed), together with one more reference.

matkoniecz:
> Two more references to the old package 

Thanks! I already committed it (not pushed), together with two additional references.

DerDings:
> 
> I do not like the image of
> 
>     * entrance (too much focus on the behind and not the "opening/surrounding barrier") and
> 
>     * cycle barrier (confusing 10 white bars).

I know a location of an straightforward entrance. But it's currently not accessible due to construction works. Maybe next week.

As said before, I went to take a photo of a simple but contrasted cycle barrier. On my way there, I encountered some even more beautiful cycle barriers, and other barriers. Number 1 is my personal favourite. Here they are:
1. DSC_3094 - Kopie
⬆ Double, angled cycle barrier, used to reduce speed rather than preventing access

2. DSC_3105 - Kopie
⬆ The double cycle barrier I was originally looking for

3. DSC_3113 - Kopie
⬆ Single cycle barrier. Not my favourite, because all those fences are confusing.

4. DSC_3118 - Kopie
⬆ Another single cycle barrier. Better than the one above, but sun was already setting 🌇.

5. DSC_3089
⬆ A swing gate. This matches the colour of the cycle barriers. However, the current picture is probably better because there's less around.

6. DSC_3091_e1
⬆ Two swing gates with some offset can make a cycle barrier. Just to show what variants can be found out there, but probably not to be included here.

7. DSC_3086 - Kopie
⬆ Some bollards. I found bollards everywhere. Those had the least distracting background. However, the current picture is already quite nice. But in case you want some without adjacent kerb, you may take them.

> I also know a place where a plastic jersey barrier was joined to a concrete one, which would make the perfect generic picture. But I don't know if it's still there.

Sadly, those jersey barriers have been removed in the meantime

I also found a lot of ladder stiles, but daylight had vanished. They're probably not better than the current picture anyway.
If you like one of the barriers shown above, but dislike the perspective, I've got some other perspectives as well.

(All pictures included in this comment taken by me. License is CC0/public domain/whatever you like)

matkoniecz:
In general - big thanks for images! And note that after this PR will be merged new images can still be applied, just open a new issue if you have a better one than currently used one.

Plenty of effort was taken to select existing ones but in some cases there was nothing really good available. For example wire gate is bad, noone took photos of wire gates. Do not remember specific right now but some quests also could be benefit a bit from better images.

And debris pile would be more clear if it would be pile of contrasting sand etc (I cheated anyway by taking landslide, as at this size you do not see that it is a clearly temporary obstruction).

> The double cycle barrier I was originally looking for

Much simpler/better contrasting than alternatives! I was confused that it only slows down cyclists, not blocks them. But description on wiki has "slows or prevents access for bicycle users" so it should be fine. I will try to crop something.

> However, the current picture is probably better because there's less around.

Yes, especially as guard rail is also a barrier. And hopefully noone will be confused by trees in the current one :)

> Some bollards. I found bollards everywhere. Those had the least distracting background. However, the current picture is already quite nice. But in case you want some without adjacent kerb, you may take them.

Would it be possible to take photo so just paving stones are in the backround? Current one is also quite noisy

> Sadly, those jersey barriers have been removed in the meantime

:(

> I also found a lot of ladder stiles, but daylight had vanished. They're probably not better than the current picture anyway.

I quite like current ladder stile. Old one was bad because on a tiny image it looked like a ladder placed randomly in field as wired become hidden.

Right now stepover stile is the worst one.

matkoniecz:
@DerDings - if you want and if you rememeber: do you want to share locations where photos were taken? Especially bicycle barriers and swing gates?

I uploaded bicycle barriers to https://commons.wikimedia.org/wiki/File:Bicycle_barrier.JPG https://commons.wikimedia.org/wiki/File:Bicycle_barrier_2.JPG

matkoniecz:
screen08
I will take the double cycle barrier that you planned originally, the (1) seems missing context a bit - that seems still more clear in (2).

Maze of white bars will be eliminated.

FloEdelmann (approved):

matkoniecz:
Version after changes/fixes:
screen15
screen16

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
westnordost: No strong feeling about this. Achievements are 🤷  anyway

app/src/main/res/values/strings.xml
```diff
@@ -991,4 +991,27 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
     <string name="quest_drinking_water_not_potable_signed">Not safe to drink (signed)</string>
     <string name="quest_drinking_water_not_potable_unsigned">Not safe to drink (unsigned)</string>
+    <string name="quest_barrier_type_title">What type of barrier is this?</string>
+    <string name="quest_barrier_type_gate">Gate</string>
+    <string name="quest_barrier_type_lift_gate">Boom barrier</string>
+    <string name="quest_barrier_type_bollard">Bollard(s)</string>
+    <string name="quest_barrier_type_chain">Chain</string>
+    <string name="quest_barrier_type_rope">Rope</string>
+    <string name="quest_barrier_type_wire_gate">Removable wire section</string>
+    <string name="quest_barrier_type_cattle_grid">Grid blocking cattle from walking over</string>
```
peternewman, line 1001: Is there any reason this doesn't just say "cattle grid" or is the idea to try and be more descriptive?

If so, it should probably be this (or some simplified word for livestock, animals, your dog wouldn't like it either?):
```suggestion
    <string name="quest_barrier_type_cattle_grid">Grid blocking livestock from walking over</string>
```

As it will stop sheep (and also horses) too.

matkoniecz: > Is there any reason this doesn't just say "cattle grid" or is the idea to try and be more descriptive?

At least I would have no idea what "cattle grid" is, or even that it applies to something road surface.

I tried similar explanation with kissing gate (failed) and did with wire gate/Hampshire gate.

peternewman: > > Is there any reason this doesn't just say "cattle grid" or is the idea to try and be more descriptive?
> 
> At least I would have no idea what "cattle grid" is, or even that it applies to something on a road surface (rather than vertical object).

Okay fair enough. I don't know if that's a language thing, or just down to the sort of environment/activies I was in growing up, or just being someone who walks a bit in the countryside.

Are you happy livestock isn't too "technical" then?

> I did similar explanation with wire gate/Hampshire gate. Also tried and failed with kissing gate.

Yeah while I might be able to summarise a kissing gate, I doubt I could get it down to six words, and I suspect certainly not without confusing some people who'd know what a kissing gate is, but possibly not a description (which I suspect would be angles and movement and stuff).

matkoniecz:
I got "gate with a buffer" as a suggestion for descriptive text for kissing gate.

Not entirely happy about it (so not applying) but the first viable one.

westnordost:
Yeah, that would be confusing I think

peternewman:
Being pedantic, it's got two buffers for starters, maybe "gate between two buffers", but I'd really want to see kissing gate in there too personally.

westnordost:
> chain on bollards can be confusing to people who may select bollard instead of a chain - no idea how to deal with it

Find a picture for chains that shows a bollard with chains

westnordost:
I'll merge it as it is, we can find a better image for chain (later)

app/src/main/res/values/strings.xml
```diff
@@ -991,4 +991,27 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="quest_barrier_type_block">Big immobile object</string>
+    <string name="quest_barrier_jersey_barrier">Jersey barrier</string>
+    <string name="quest_barrier_type_log">Log</string>
+    <string name="quest_barrier_type_kerb">Curb</string>
+    <string name="quest_barrier_type_height_restrictor">Height barrier</string>
+    <string name="quest_barrier_full_height_turnstile">Full height turnstile</string>
+    <string name="quest_barrier_type_turnstile">Turnstile</string>
+    <string name="quest_barrier_type_passage">Passage through wall/fence/etc</string>
+    <string name="quest_barrier_type_debris_pile">Pile of debris</string>
+    <string name="quest_barrier_type_stile_squeezer">Squeezer stile</string>
```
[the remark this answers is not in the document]
matkoniecz: Looking at it (not into result number) it seems that "squeeze" is both shorter and more popular. Changed.

app/src/main/res/values/strings.xml
```diff
@@ -991,4 +991,27 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="quest_barrier_type_wire_gate">Removable wire section</string>
+    <string name="quest_barrier_type_cattle_grid">Grid blocking livestock from walking over</string>
+    <string name="quest_barrier_type_block">Big immobile object</string>
+    <string name="quest_barrier_jersey_barrier">Jersey barrier</string>
+    <string name="quest_barrier_type_log">Log</string>
+    <string name="quest_barrier_type_kerb">Curb</string>
+    <string name="quest_barrier_type_height_restrictor">Height barrier</string>
+    <string name="quest_barrier_full_height_turnstile">Full height turnstile</string>
+    <string name="quest_barrier_type_turnstile">Turnstile</string>
+    <string name="quest_barrier_type_passage">Passage through wall/fence/etc</string>
```
FloEdelmann, line 1009: @matkoniecz `etc.` should end with a period: https://en.wikipedia.org/wiki/Etc.

matkoniecz: Thanks, fixed in code in https://github.com/streetcomplete/StreetComplete/commit/f0621e6200179694fb3eb43ca62df96b18e877b6 and checked in POEditor.

cyclingcat:
> Find a picture for chains that shows a bollard with chains

Do you main a "chain on bollards" photo like this?

chain

The cycling cat

matkoniecz:
No, row of bollards connected by chain

matkoniecz:
I just reviewed notes created by people encountering this quest (as I promised here at some point), I see no systematic problems.

In general note count is very low - probably because use of that quest is also really small.

See https://ent8r.github.io/NotesReview/?view=map&map=2%2F3.5134%2F0&query=%22What%20type%20of%20barrier%20is%20this%3F%22&limit=100&status=open

See https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#solving-notes in general (looking at your country is in general very helpful and many notes can be remotely fixed)

PR #2754, opened 2021-04-16
Initial start on auto-generating taginfo JSON

peternewman:
I'm not expecting any serious code reviews yet, more just sort of getting it out there a bit, and I guess starting to look at workflows.

TODO:

- [ ] Combine the function names (e.g. add versus updateCheckDateForKey) - "Whether add or modify doesn't really matter, if it tags it, it tags it."


westnordost:
Good start! Though, I can't say much about this yet because I haven't looked closer at how the file would need to look like for taginfo to work with it properly.

I'd just say that this should go into an own script.

I wonder if it makes more sense to generate this file once every release or as a hook like @FloEdelmann 's wiki-script. If the former, the proper location would be `buildSrc/src/main/java/TaginfoJasonTask.kt` and then register a task in the project's `build.gradle.kts`

The old taginfo json from @goldfndr listed the "quest fill" and "quest filter" separately. For a new version, I think it is better to only keep the "quest fill" (=how the app tags something) part, because obviously, for the filter, the app sometimes interprets deprecated, outdated, ambiguous, not-really-used and "wrong" tags, just so that the quest is not shown for too many elements. These kind of tags should then not appear in taginfo as "StreetComplete uses these tags" because this paints a false picture of the situation:
- It may make it look like StreetComplete is propagating the use of this tag
- It may look like a tag is "accepted and used" because StreetComplete uses it

matkoniecz:
If you reach stage when it starts generating things and you are unaware about bugs... You can publish output as gist and I can look at some quests that I know well. Maybe I will spot some bugs this way.

And it is really cool that you are doing this!

peternewman:
> Good start! Though, I can't say much about this yet because I haven't looked closer at how the file would need to look like for taginfo to work with it properly.

Thanks. It's just the same as the existing one, I was initially trying to just automate what @goldfndr had done manually. Most of the fields are just free text, so obviously his hand-crafted explanations are likely to be better, but only if it's kept up to date all the time, hopefully we can be good enough via some automation:
https://wiki.openstreetmap.org/wiki/Taginfo/Projects

> I'd just say that this should go into an own script.

Yes, although it shares lots of common requirements with @FloEdelmann 's code, such as extracting quest names and lists of quests, so unless we're going to have a helper library too... I was going to add some command line arguments (if Kotlin does them, so you would decide which output to produce).

> I wonder if it makes more sense to generate this file once every release or as a hook like @FloEdelmann 's wiki-script. If the former, the proper location would be `buildSrc/src/main/java/TaginfoJasonTask.kt` and then register a task in the project's `build.gradle.kts`

Possibly, we can automate the generation like @FloEdelmann 's so it just gets built automagically each time there is a release, the benefit then is you don't need to do anything. I guess with both there's an argument that it's beneficial to have the data available sooner (i.e. as soon as a quest is released), either because people might be running their own builds of the code, either forks or nightly releases or whatever, and also because as soon as its been committed it means it's going to appear at some point soon, so in the wiki case it would mean the page could be updated before it's released, so it's immediately current when people checkout the new version, and in the Taginfo case, an accessible map designer for example might choose to prioritise adding a feature to their app knowing SC supports it and therefore data will exist and grow. Essentially once the code has been written, it's going to appear so having a little head start can't hurt.

> The old taginfo json from @goldfndr listed the "quest fill" and "quest filter" separately. For a new version, I think it is better to only keep the "quest fill" (=how the app tags something) part, because obviously, for the filter, the app sometimes interprets deprecated, outdated, ambiguous, not-really-used and "wrong" tags, just so that the quest is not shown for too many elements. These kind of tags should then not appear in taginfo as "StreetComplete uses these tags" because this paints a false picture of the situation:

I guess it depends what people are doing with the data. Avoiding the query part is certainly likely to make the code far simpler, as I don't have to parse and understand the Overpass type stuff, just some basic Kotlin methods. Presumably it's mostly irrelevant as for the majority of current tags and ones we're "filling", we're likely to be filtering on them too, to ensure they aren't already set. Are there actually many which we only filter on?

>     * It may make it look like StreetComplete is propagating the use of this tag
> 
>     * It may look like a tag is "accepted and used" because StreetComplete uses it

Yep, understood. Although we could add some descriptions about those usages.

peternewman:
> If you reach stage when it starts generating things and you are unaware about bugs... You can publish output as gist and I can look at some quests that I know well. Maybe I will spot some bugs this way.

Still on my list is to actually write the JSON file and commit it to the repo, it would be in the runner output, but the runner doesn't currently run apart from on a release due to the workflow config. I might tweak that to aid development. Thanks for the offer anyway. As long as a quest isn't setting things in multiple different ways (which I think a few are), it's likely most bugs will be fairly obvious as it will complain it can't process the data.

> And it is really cool that you are doing this!

Thanks. I thought it seemed like an interesting little challenge with a useful outcome, without me having to get into proper Android and GUI development! I'm still a bit worried it might end up rather brittle and fragile, but probably less hassle than using a full Kotlin lexer to do it properly!

matkoniecz:
> Are there actually many which we only filter on?

From just implemented https://github.com/streetcomplete/StreetComplete/pull/2753/files 

sets `barrier` (several values), filters on `barrier=yes` and:

```
         and !man_made
         and !historic
         and !military
         and !power
         and !tourism
         and !attraction
         and !amenity
         and !leisure
```

(as there are rare tagging mistakes of `leisure=playground barrier=yes` type that could result in mapper selecting "not existing" option for question about specifying barrier type. This filtering should reduce this anyway rare cases to something basically not happening)

Or building levels quest with

```
!building:levels and !height and !building:height
         and !man_made and location != underground and ruins != yes
```

filter (though here only `building:height` is a bad value, other would be kind of useful)

> I was going to add some command line arguments (if Kotlin does them, so you would decide which output to produce).

It may be preferable to have only actually used code (as unused code triggered by arguments is by definition not used but still requires maintenance/review etc)

> probably less hassle than using a full Kotlin lexer to do it properly!

Though after some times of doing this I am nowadays really preferring starting from a proper parser if at all available :)

westnordost:
Are you still working on this?

peternewman:
> Are you still working on this?

I had mostly forgotten, or been doing other things.

I've now picked it up and it covers most stuff I think. There are a few fiddly edge cases, but I wonder if it's worth merging as is/after a small tidy up, and perhaps having both JSON files in parallel. E.g. @goldfndr manual one still covers all the query stuff, which this doesn't touch at all yet, but equally this has already picked up say the bollard and CCTV camera quests with no input from anyone, so it's doing useful things already. I'm pretty confident it's not lying, and while it's incomplete in some ways, so is the existing one. I believe TagInfo is quite happy with having multiple JSON files for the same project, so there's no major downside there.

peternewman:
Remember to update https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#improving-documentation when this is in!

.github/generate-quest-metadata.main.kts
```diff
@@ -0,0 +1,505 @@
...
+//              // TODO: Fix me!
+//              println("Skipping, complex enum " + it + "...")
+//              allChanges.add(TaginfoChange(key, "", change))
+//              continue
+            } else if (it == "BusStopShelterAnswer" || it == "StepsRampAnswer") {
+              // TODO: Fix me!
+              println("Skipping, simple enum " + it + "...")
+              allChanges.add(TaginfoChange(key, "", change))
+              continue
+            } else if (it == "RoadNameAnswer" || it == "PlaceNameAnswer" || it == "BusStopRefAnswer" || it == "HousenumberAnswer" || it == "PostboxRefAnswer") {
```
westnordost, line 316: Just randomly looking through this PR...

```suggestion
            } else if (it in setOf("RoadNameAnswer","PlaceNameAnswer","BusStopRefAnswer","HousenumberAnswer","PostboxRefAnswer") {
```

Though that the script relies on hardcoded quest names is maybe a red flag anyway

peternewman:
How do you feel about my suggestion @westnordost ? :

> I've now picked it up and it covers most stuff I think. There are a few fiddly edge cases, but I wonder if it's worth merging as is/after a small tidy up, and perhaps having both JSON files in parallel. E.g. @goldfndr manual one still covers all the query stuff, which this doesn't touch at all yet, but equally this has already picked up say the bollard and CCTV camera quests with no input from anyone, so it's doing useful things already. I'm pretty confident it's not lying, and while it's incomplete in some ways, so is the existing one. I believe TagInfo is quite happy with having multiple JSON files for the same project, so there's no major downside there.

Likewise having updated my branch, it's picked up the stile ones, including the materials.

Essentially aside from a bit of workflow tweaking to write the JSON to the repo, we're probably there in terms of minimum viable product, which would work if it ran in parallel as a separate file to the manual one, and those hard-coded quests it currently ignores could be manually fixed and then removed from that skip list, thereby improving it's coverage and JSON generation.

.github/generate-quest-metadata.main.kts
peternewman: Thanks for the hint, I've applied it there and in a few other places.

> Though that the script relies on hardcoded quest names is maybe a red flag anyway

See discussion here as to why they're (currently) hardcoded:
https://github.com/streetcomplete/StreetComplete/pull/2754#issuecomment-901544052

westnordost:
The goal would be to completely replace the old JSON. I wouldn't like a provisionary solution where both JSONs exist alongside each other (if that is even possible).

westnordost:
In regards to "what is in the quest filter syntax is currently ignored" - good! I think it should stay that way.

The quest filter also includes all kind of deprecated, old, ambigious and mostly unused tagging - it should not convey the impression that StreetComplete is really using these (i.e. it might appear, since the app is an editor, that the app is tagging this or somehow encouraging to use this).

In regards to it being a .github task. I think it would be better as a gradle task in /buildSrc. Advantages:
- gradle tasks can add to and update files in the repository. I think it makes sense if the generated JSON is somewhere in the repository. 
- it would usually be updated on release, which makes more sense as if it is update on master
- you can share code, use proper library dependencies etc. like in a normal application. Right now, you stuffed to completely different tasks into one file to achieve that.

peternewman:
> The goal would be to completely replace the old JSON. I wouldn't like a provisionary solution where both JSONs exist alongside each other (if that is even possible).

It's possible:
https://wiki.openstreetmap.org/wiki/Taginfo/Projects#Project_Files

Otherwise how would you feel about just storing some of the static fill data from the existing JSON file and populating it for those quests for the moment (with automating them a future enhancement in a separate PR). There are a few name ones, which need some special work but would at least apply to a number of them, but the worst case is probably the cycleway quest, because of the processing of left/right/both and also the partial values which aren't tagged, essentially it needs its own dedicated parsing code that is unlikely to be any use for anything else (apart from maybe the pavements quest).

> In regards to "what is in the quest filter syntax is currently ignored" - good! I think it should stay that way.

Great, that's one less bit of work to do then! I wonder if there's an argument for having some hand-curated stuff, possibly separately, for primary filter tags (e.g. `highway=bus_stop`), as some of those are never tagged. Also looking at the current JSON, there are the floor/level/tunnel/underground stuff which we use but never write back.

> The quest filter also includes all kind of deprecated, old, ambigious and mostly unused tagging - it should not convey the impression that StreetComplete is really using these (i.e. it might appear, since the app is an editor, that the app is tagging this or somehow encouraging to use this).

Having separate projects for read and write might help to clear up some of that confusion (even more easily than text in a wordy description), but there are probably still downsides for showing the "bad" tagging we read.

> In regards to it being a .github task. I think it would be better as a gradle task in /buildSrc.

Fair enough, presumably you mean similar to #2815 ? What about the Wiki page task, does that need moving too. They share a lot of the same code, so I think they need to share some common "library" files at least.

> Advantages:
> 
>     * gradle tasks can add to and update files in the repository. I think it makes sense if the generated JSON is somewhere in the repository.

GitHub can do this too, although I guess if we can do it in Gradle it means that works regardless of who or what is calling it, whereas GitHub style you'd need to do something else when doing it yourself.

>     * it would usually be updated on release, which makes more sense as if it is update on master

As I mentioned above, I still feel it should probably be updated as soon as something have been merged to master. While the releases are pretty regular, it would still give a few weeks heads up of the intention that SC is supporting a particular tag.

>     * you can share code, use proper library dependencies etc. like in a normal application. Right now, you stuffed to completely different tasks into one file to achieve that.

I wasn't really clear if I could do that with scripts too, or can they not load local external files, just full libraries? Again presumably the wiki task needs migrating too?

westnordost:
> Otherwise how would you feel about just storing some of the static fill data from the existing JSON file and populating it for those quests for the moment (with automating them a future enhancement in a separate PR). There are a few name ones, which need some special work but would at least apply to a number of them, but the worst case is probably the cycleway quest, because of the processing of left/right/both and also the partial values which aren't tagged, essentially it needs its own dedicated parsing code that is unlikely to be any use for anything else (apart from maybe the pavements quest).

Don't really understand this. What I want to avoid is to merge something that is not completely finished and in effect, requires maintenance over time (because hardcoded stuff needs to be updated).
I didn't have a closer look at the code, but wouldn't looking for `changes\.(add|addOrModifiy|modify|.....etc)` in the source files of quests be enough to catch it all?

peternewman:
> Don't really understand this. What I want to avoid is to merge something that is not completely finished and in effect, requires maintenance over time (because hardcoded stuff needs to be updated).

Yes, I'd broadly agree, although how much is the cycleway quest likely to get updated, or at least the left/right/both tagging of it. It's likely if it got updated significantly the parsing code would break anyway so would need updating again.

There are currently two or more different types of enums used, and lists and boolean as well as a separate yes/no style. If people add further ways of representing the state data internal to the quest, this will again either need extending again or that quest ignoring and not being represented.

I would counter that we currently have the situation where the existing data is over two years out of date, hence my comments that even in it's current state this is probably better than that.

> I didn't have a closer look at the code, but wouldn't looking for `changes\.(add|addOrModifiy|modify|.....etc)` in the source files of quests be enough to catch it all?

That's essentially what all the code does, but to capture and interpret the key and value it needs to do a certain amount of processing. The cycleway quest makes this even more involved than most:
https://github.com/streetcomplete/StreetComplete/blob/acfac0b7a3fa6cc2b34743e8955ba09db3f0737b/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt#L177-L185

So for the value, it needs to lookup elsewhere that the values are `yes` or `-1`. For the key it's worse, it needs to substitute the `$cyclewayKey` variable, which itself is a variable:
https://github.com/streetcomplete/StreetComplete/blob/acfac0b7a3fa6cc2b34743e8955ba09db3f0737b/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt#L177

From another variable:
https://github.com/streetcomplete/StreetComplete/blob/acfac0b7a3fa6cc2b34743e8955ba09db3f0737b/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt#L165-L167

So I think that's about three layers of indirection it needs to handle for that one quest alone and that's before it deals with the when assignments of the values or the if statements buried in there too. Even just getting to the minimum viable for that quest of the tag names will be enough of a challenge.

westnordost:
> I would counter that we currently have the situation where the existing data is over two years out of date, hence my comments that even in it's current state this is probably better than that.

True. Okay. I understand the difficulties.
How about the algorithm would ragequit if it finds anything unexpected and requires the usages for that quest type be hardcoded? Then, if a new quest is added whose syntax is not detected automatically by the script, we will know of it.

peternewman:
> > I would counter that we currently have the situation where the existing data is over two years out of date, hence my comments that even in it's current state this is probably better than that.
> 
> True. Okay. I understand the difficulties.
> How about the algorithm would ragequit if it finds anything unexpected and requires the usages for that quest type be hardcoded? Then, if a new quest is added whose syntax is not detected automatically by the script, we will know of it.

Yeah that's fine, its essentially doing that already with the logging, but currently it's just writing a message and moving on, it could certainly serve up hard-coded data then if it hits something it doesn't understand and doesn't have hard-coded data for it could die.

This for me would also be a strong reason for running it on every commit to master as part of the CI (even if we don't publish the JSON at this point (although I personally still think we should)) as it could then flag up early that it didn't work and we could add the necessary hard-coded info before the release.

Alternatively I was thinking it would be fairly easy to merge two JSON files, so we could also just merge what's coming out of here into the existing manual JSON if that was preferred (either permanently or in the interim).

On a similar note, we could drop a local taginfo.json file in each quest folder for the hard-coded ones and it could scoop those up if they exist, rather than putting the desired tags for complicated quests in the generating code itself to keep data and logic separate.

westnordost:
> Yeah that's fine, its essentially doing that already with the logging

Ah I see. But the logging is somewhere in some github task right? Or where is it?

Is it possible for a github task to add files to the repository? If yes, sounds like a security risk. But at the same time if no, it's too bad, because it would be nice to have a json automatically updated (so f.e. it could be fetched by taginfo directly from the repos).
This is why I mentioned a custom gradle task, because gradle tasks can of course modify files (which I then can commit, like for pulling the translations, generating the metadata, pulling the presets and so forth).

In any case, can I see somewhere what exactly this script produces (log + result)?

> On a similar note, we could drop a local taginfo.json

I wouldn't like that, because it would be useless files **in** the app source directory that would need to be filtered out when building the app. But surely, hard-coded quest answers could reside in a different file than the script itself, some kind of JSON like 
```js
{
  "AddStreetName": {
    //...
  },
  //...
}
```
maybe f.e. just next to the script itself

peternewman:
> > Yeah that's fine, its essentially doing that already with the logging
> 
> Ah I see. But the logging is somewhere in some github task right? Or where is it?

Yeah, it's visible here (under the relevant action workflow and task step):
https://github.com/streetcomplete/StreetComplete/pull/2754/checks

If it was a Gradle task run via GitHub actions which output log info it would show in the same place. It could also be gathered and added as an asset, but I'm not sure why you'd bother as it would be more faff to view.

> Is it possible for a github task to add files to the repository? If yes, sounds like a security risk. But at the same time if no, it's too bad, because it would be nice to have a json automatically updated (so f.e. it could be fetched by taginfo directly from the repos).

It has commit rights, but only in the local repo, forks only have read rights. You also have to approve running new actions and obviously review changes to the action, so I can only add an action to make changes to the repo if you merge said action:
https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

> This is why I mentioned a custom gradle task, because gradle tasks can of course modify files (which I then can commit, like for pulling the translations, generating the metadata, pulling the presets and so forth).

Yeah again we can do it directly, or as a gradle task which we run so it's less reliant on GitHub and would translate more easily to other systems.

We could also run the other tasks at sequential periods to both keep the data current and detect failures/changes in their APIs promptly, as well as invalid data added to the YAML metadata files.

> In any case, can I see somewhere what exactly this script produces (log + result)?

As above:
https://github.com/streetcomplete/StreetComplete/pull/2754/checks
The artifacts are linked at the top (currently here):
https://github.com/streetcomplete/StreetComplete/suites/3536820625/artifacts/84633906

Which is a zip file with the JSON in.

> > On a similar note, we could drop a local taginfo.json
> 
> I wouldn't like that, because it would be useless files **in** the app source directory that would need to be filtered out when building the app.

Ah yes, good point.

> But surely, hard-coded quest answers could reside in a different file than the script itself, some kind of JSON like
> maybe f.e. just next to the script itself

They could indeed. I quite like the idea of a file per quest, as you can see at a glance which quests have hard-coded data, but I'm not too wed on it if you'd prefer one file.

westnordost:
> Yeah, it's visible here (under the relevant action workflow and task step):

Oh, that log output is quite verbose. Difficult to spot with which quests exactly the script has issues.

I've looked at the actual output now. I'd like to suggest:
- leave out any deletions of tags. This is not "using" the tag in my opinion, also it is confusing to read. F.e. `AddBuildingType` deletes `building` key, what?
- don't mention quest type in description, it is already mentioned in doc_url
- don't mention `add`/`addOrmodify` etc in description, it's also confusing

> Yeah again we can do it directly, or as a gradle task which we run so it's less reliant on GitHub and would translate more easily to other systems.

I'd really prefer that

> We could also run the other tasks at sequential periods to both keep the data current and detect failures/changes in their APIs promptly, as well as invalid data added to the YAML metadata files.

There is no need, I run all these tasks at least before every release.

peternewman:
> Oh, that log output is quite verbose. Difficult to spot with which quests exactly the script has issues.

Yeah, it's mostly just my debugging for now. It definitely fails with the named ones, well it skips them, and also anything where it says skip. I'd imagine the final script would error if it met an un-handled quest and we'd be able to see the ones that manual JSON was set for.

> I've looked at the actual output now. I'd like to suggest:
> 
>     * leave out any deletions of tags. This is not "using" the tag in my opinion, also it is confusing to read. F.e. `AddBuildingType` deletes `building` key, what?

Fair enough, done.

>     * don't mention quest type in description, it is already mentioned in doc_url

It looks like this only shows as a More... link:
https://github.com/taginfo/taginfo/blob/34676f568d61c1a3a6b8fd965107e04a6b645524/web/public/js/taginfo.js#L727-L739

Although I guess we don't need the quest name AND question, so I'll drop the former.

>     * don't mention `add`/`addOrmodify` etc in description, it's also confusing

So you think drop the type entirely? We've currently got the following:
    239 add 
     63 addOrModify 
      3 applyRampAnswer 
     96 modify 
      3 modifyIfExists 
      2 updateCheckDateForKey 
     85 updateWithCheckDate

Should we really combine all those (e.g. add versus updateCheckDateForKey)?

> I'd really prefer that

Okay, I'll see how I get on with that once the logic is working as intended.

> There is no need, I run all these tasks at least before every release.

I'm not saying you don't, but it would mean that issues such as #3202 and breakages in this script would be caught both sooner (rather than in the days/week before release) and could also potentially be fixed by others. Potentially (e.g. with https://github.com/streetcomplete/StreetComplete/commit/7df21b15e465cce6ec8394be1e4ebba683fc8746 ) it could be done during the PR itself.

matkoniecz:
> but it would mean that issues such as #3202 and breakages in this script would be caught both soone

#3202 was caught, but as it was nothing fatal and obnoxious to fix the relevant scripts were commented out - in hope that Sophox situation will be fixed on Sophox side

westnordost:
> Should we really combine all those (e.g. add versus updateCheckDateForKey)?

Yes, I think yes. Whether add or modify doesn't really matter, if it tags it, it tags it. Regarding check date, it should probably be interpreted like what it tags. F.e. the taginfo JSON should show that the opening hours quest tags `opening_hours` but also `check_date:opening_hours`.

matkoniecz:
@peternewman Are you still working on this one? Or is it waiting for someone to adopt it?

.github/generate-quest-metadata.main.kts
```diff
@@ -0,0 +1,535 @@
...
+    val questFileContent = sourceDirectory.resolve("quests/QuestModule.kt").readText()
+    val questNameRegex = Regex("(?<=^ {8})[A-Z][a-zA-Z]+(?=\\()", RegexOption.MULTILINE)
+    val questNames = listOf(noteQuestName) + questNameRegex.findAll(questFileContent).map { it.value }
+
+    val questFiles = getFilesRecursively(sourceDirectory.resolve("quests/"))
+
+    when (task) {
+      "wiki" -> generateWikiCsv(questNames, questFiles)
+      "json" -> generateTaginfoJson(questNames, questFiles)
+      else -> { // Note the block
```
matkoniecz, line 71: What this comment is referring to? `{` `}`?

.github/generate-quest-metadata.main.kts
```diff
@@ -0,0 +1,535 @@
...
+}
+
+fun generateTaginfoJson(questNames: List<String>, questFiles: List<File>) {
+    val project = Project(
+        "StreetComplete",
+        "Surveyor app for Android",
+        "https://github.com/streetcomplete/StreetComplete",
+        "https://wiki.openstreetmap.org/wiki/StreetComplete",
+        "https://raw.githubusercontent.com/streetcomplete/StreetComplete/master/app/src/main/res/mipmap-xhdpi/ic_launcher.png",
+        "Peter Newman"
```
matkoniecz, line 94: Are you sure that you want to be contacted directly about maintaining this script?

matkoniecz:
General question: would it be easier to parse Kotlin code into some AST?

I did may share of "parsing is complicated, I will solve it bunch of regexps" and after several months I wished that I did it properly with parsing - as I kept encountering one edge case after another and ended with monster regexp which was uneditable nightmare.

End result was dynamically build regexp that was incomprehensible large and really poor in editing (at least I had tests).

If that would be regexp - then it really needs some tests, otherwise (I know from experience) this will be irritating source of regressions if this will be ever changed.

smichel17:
> would it be easier to parse Kotlin code into some AST?

I was going to suggest the same thing. At the very least, we already *have* code for parsing a string to an `ElementFilterExpression` (EFE); it shouldn't be *too* much work\* to have an EFE spit out an equivalent tagInfo for it.

For more complicated parsing, perhaps we could take an approach something like a javadoc: embed a comment above the `getApplicableElements` function with the relevant information. That would still require manual maintenance, but at least it would be right in your face when you're changing the code, rather than off in a separate repo.

\*famous last words, I know

peternewman:
> #3202 was caught, but as it was nothing fatal and obnoxious to fix the relevant scripts were commented out - in hope that Sophox situation will be fixed on Sophox side

Yeah, it just avoids that mad rush finding loads of stuff is broken just as you're about to release.

> > Should we really combine all those (e.g. add versus updateCheckDateForKey)?
> 
> Yes, I think yes. Whether add or modify doesn't really matter, if it tags it, it tags it.

Okay, I'll add to the TODO list, it's vaguely useful during debug to know where it's come from.

> Regarding check date, it should probably be interpreted like what it tags. F.e. the taginfo JSON should show that the opening hours quest tags `opening_hours` but also `check_date:opening_hours`.

I think it already did this before you commented, it definitely does now anyway.

> @peternewman Are you still working on this one? Or is it waiting for someone to adopt it?

On and off, although mostly off unfortunately, too many other bits to do too. I wouldn't hugely object if someone wanted to take it on, I think I've done most of the easy bits! I've just updated it all to cope with all the changes due to #3665 .

> General question: would it be easier to parse Kotlin code into some AST? For example with https://github.com/kotlinx/ast

From a quick look at that, I'm not sure it actually solves the hard part of this, which is dealing with the logic and enums etc. It would avoid the regex, and some potential errors there, but probably in exchange for more complicated/less readable means of processing the results

e.g. given this code:
https://github.com/streetcomplete/StreetComplete/blob/9c6d3e25216d06a2c5afa71086949e1e195de926/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt#L196-L216

We need to produce the JSON version of these statements; that to my mind is the hard bit:
```
tags["left"] = "no"
tags["left"] = "lane"
tags["left:oneway"] = "yes"
tags["left:oneway"] = "-1"
tags["right"] = "no"
tags["right"] = "lane"
tags["right:oneway"] = "yes"
tags["right:oneway"] = "-1"
tags["both"] = "no"
tags["both"] = "lane"
tags["both:oneway"] = "yes"
tags["both:oneway"] = "-1"
```

As I mentioned, I could possibly conceive you might be able to do some preprocessor magic in C/C++, I don't know if Kotlin offers equivalents, or if it would even help.

Maybe if everything could be switched to code similar to these (which in fairness the Cycleway is already):
https://github.com/streetcomplete/StreetComplete/blob/9c6d3e25216d06a2c5afa71086949e1e195de926/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/BarrierType.kt#L31-L48

Then we'd just need to write code which feeds in all combinations of inputs (i.e. Cycleway x Side x dir) and captures all the output; but then in this case, dir is currently an Int so has thousands of values; we'd need to do the Int to enum conversion first to make it manageable.

> I did may share of "parsing is complicated, I will solve it with bunch of simple regexps". And after several months I wished that I did it properly with parsing - as I kept encountering one edge case after another and ended with monster regexp which was uneditable nightmare.

Yeah I've found a few quirks here which I've caught and fixed, some of which is inconsistency in the codebase (e.g. func = func2 versus func { func2 } etc).

> If that would be regexp - then it really needs some tests, otherwise (I know from experience) this will be irritating source of regressions if this will be ever changed.

Agreed, I've started collecting the examples so some tests can be written.

> > would it be easier to parse Kotlin code into some AST?
> 
> I was going to suggest the same thing. At the very least, we already _have_ code for parsing a string to an `ElementFilterExpression` (EFE); it shouldn't be _too_ much work* to have an EFE spit out an equivalent tagInfo for it.

Remember we're only interested in the tagging side, not the selection side (@westnordost preference, but I can see the logic so we don't flag e.g. `roundabout=yes` in our road quest or some other random/obscure tag). So I don't think any EFE stuff will help.

> For more complicated parsing, perhaps we could take an approach something like a javadoc: embed a comment above the `getApplicableElements` function with the relevant information. That would still require manual maintenance, but at least it would be right in your face when you're changing the code, rather than off in a separate repo.

Again, we don't care about `getApplicableElements` just `applyAnswerTo`. I think @westnordost didn't want JSON in the source folders, we talked about some JSON outside of the main source. I'd imagine he may feel the same about Javadoc style commenting.

.github/generate-quest-metadata.main.kts
peternewman: Yeah. I think it's possibly because if I didn't put the block in, it didn't print or something?

.github/generate-quest-metadata.main.kts
peternewman: Possibly not, a bit of a placeholder at least. I think the spec suggested it should be an individual... :shrug:

westnordost:
This has been open as draft since almost a year. Is this still being worked on?

westnordost:
Closing as this seems to be abandoned.

westnordost:
If you intend to finish it, please only open a PR when it is ready for review or if you need help (then as draft).

PR #2755, opened 2021-04-17
Add aerial imagery, fixes #56

### LINKED ISSUE #56: add functionality to show satellite map
This feature would be very helpful for the roof shapes.

matkoniecz:
Notes: 

- it uses Esri aerial as it has a global coverage and - unlike Bing - is available without byzantine setup.
- it works, but I suspect that it is architecturally sad way of getting preferences.
- I worry a bit about this setting encouraging people to start and answering remotely and guessing "gray on aerial"? definitely asphalt! But it would be useful in some cases.
- fixes second-oldest open issue ( https://github.com/streetcomplete/StreetComplete/issues?q=is%3Aissue+is%3Aopen+sort%3Acreated-asc ) and finishes branch that I started a year ago :)


andrewharvey:
Awesome. In the future it would be nice to use the "best" layer from Editor Layer Index, but that would be a fair bit more effort, so glad to see this launch first.

HolgerJeromin:
Please add
fixes #56
To the first comment to be recognised by the PR gui on the right/bottom.

matkoniecz:
>  To the first comment to be recognised by the PR gui on the right/bottom.

It is in the title what should be sufficient.

And likely is in one of commits already anyway.

matkoniecz:
> Awesome. In the future it would be nice to use the "best" layer from Editor Layer Index, but that would be a fair bit more effort, so glad to see this launch first.

I worry a bit about potential misuse. So sinking time into implementing all that seems to not be a good idea if it could be potentially removed anyway.

And many imagery sources require various special support, API keys, workarounds etc so it would be quite time consuming.

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
```diff
@@ -218,7 +221,12 @@ open class MapFragment : Fragment(),
     protected open fun getSceneFilePath(): String {
         val currentNightMode = resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK
         val isNightMode = currentNightMode == Configuration.UI_MODE_NIGHT_YES
-        val scene = if (isNightMode) "scene-dark.yaml" else "scene-light.yaml"
+        val isAerialView = prefsShared.getString(Prefs.THEME_BACKGROUND, "MAP") == "AERIAL_ESRI"
+
+        val scene = if (isAerialView) "scene-satellite.yaml" else {
+            if (isNightMode) "scene-dark.yaml" else "scene-light.yaml"
+        }
```
FloEdelmann, line 228: That would make it a bit easier to read, I guess:

```suggestion
        val scene = when {
            isAerialView -> "scene-satellite.yaml"
            isNightMode -> "scene-dark.yaml"
            else -> "scene-light.yaml"
        }
```

app/src/main/res/values/arrays.xml
```diff
@@ -36,4 +36,13 @@
...
     </string-array>
 
+    <string-array name="pref_entries_background_type_select" translatable="false">
+        <item>@string/background_type_map</item>
+        <item>@string/background_type_aerial_esri</item>
+    </string-array>
+
+    <string-array name="pref_entryvalues_background_type_select" translatable="false">
+        <item>MAP</item>
+        <item>AERIAL_ESRI</item>
```
FloEdelmann, line 46: Why does the option name include "Esri" here? If at some point in the future the tile provider is changed, the option saved in the database will be outdated/wrong.

FloEdelmann (commented):
That's surprisingly little code for such a big feature!

app/src/main/res/values/arrays.xml
matkoniecz: Good question, I assumed that in such case it would mean rewrite to multiple aerials, but either single provider can be exchanged for another or SC will autoselect best imagery.

matkoniecz:
> That's surprisingly little code for such a big feature!

I never though that I will praise Esri for making the simplest product that is available for free. But I must say that from worldwide aerial views without extra complexity - their is the best among ones available for OSM mapping.

Also, nice case of how measuring programming work by lines of code is a bad idea: that was not simple to figure out how to run it (commit history reveals it only partially).

app/src/main/res/values/arrays.xml
matkoniecz: Changed, thanks!

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
matkoniecz: Looks nice - but I will need to check is it working. What if both isAerialView and isNightMode are true?

Yes, it works: https://kotlinlang.org/docs/control-flow.html#when-expression

westnordost:
Doesn't ESRI require attribution in the corner of the map or something?

matkoniecz:
No, see https://github.com/streetcomplete/StreetComplete/issues/56#issuecomment-526805321

app/src/main/res/xml/preferences.xml
```diff
@@ -61,6 +61,16 @@
             android:entryValues="@array/pref_entryvalues_theme_select"
             android:persistent="true"
             />
```
FloEdelmann, line 63: Should the theme selection preference be disabled when aerial imagery is selected?

matkoniecz: It still affects UI - for example menu of quest question and of settings menu itself.

FloEdelmann (approved):

westnordost (approved):

PR #2759, opened 2021-04-17
Ask which specific shop is here on shop=yes

### LINKED ISSUE #2391: Ask which specific shop is here (for shop=yes objects)
### General
Affected tag(s) to be modified/added: [shop](https://wiki.openstreetmap.org/wiki/Key:shop)
Question asked: **What kind of a shop is here?**

### Checklist
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one) - obviously some rare shop types and ones not included in iD presets will require a note
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)  115 490 ( https://overpass-turbo.eu/s/11o2 )

### Ideas for implementation

**Element selection:**  `shop=yes and !amenity and !tourism and name`

**Metadata needed:** Already present presets for shop types

**Proposed GUI:** Reuse interface for quest about disused shops (so most of the work was done already)

matkoniecz:
fixes #2391

screen06


matkoniecz:
@westnordost Are you interested in making icon also for this one? Current one is copied from vacant shop quest and can be probably also used.

After all, it can be explained as shopping list element or something. And quest is basically the same as vacant location resurvey anyway.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
```diff
@@ -0,0 +1,64 @@
...
+        nodes, ways, relations with (
+             shop = yes
+             and !man_made
+             and !historic
+             and !military
+             and !power
+             and !tourism
+             and !attraction
+             and !amenity
+             and !leisure
```
FloEdelmann, line 24: Are these needed? The quest does not enable POI deletion, so this shouldn't be an issue.

matkoniecz: Some people (IMHO badly and pointlessly) used `amenity=fuel` `shop=yes` to tag "this fuel station has shop".

In theory the same mistake can be applied more widely, so this avoids all of that.

(I prefer overkill over whack-a-mole in case of handling tagging schemes, but if someone wants to exclude only `amenity=fuel` I would not protest)

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
```diff
@@ -0,0 +1,64 @@
...
+    override val elementFilter = """
+        nodes, ways, relations with (
+             shop = yes
+             and !man_made
+             and !historic
+             and !military
+             and !power
+             and !tourism
+             and !attraction
+             and !amenity
```
westnordost, line 23:
```suggestion
             and !amenity
             and !craft
             and !tourism
```

For example if someone tagged `craft=shoemaker + shop=yes` to denote that this is a shoemaker you can just walk in as opposed to a workshop of a shoemaker that is not also a shop.
Or for example, a museum and somebody put `shop=yes` on it too, to denote that there is also a museum (gift) shop.

(you mentioned you prefer overkill over whack-a-mole)

westnordost (changes requested):
Hmm, I am still split about this, I think there are lots of shops which do not fit neatly into the OSM categories - even if an effort was made to create values for every type of shop. So I think, a value like "miscellaneous" or something is missing, or `shop:description=...`.

If this is added, someone should monitor the notes being created for this quest and see if there is any need to act.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
```diff
@@ -0,0 +1,64 @@
...
+             and !leisure
+          )
+    """
+    override val commitMessage = "Specify shop type"
+    override val wikiLink = "Key:shop"
+    override val icon = R.drawable.ic_quest_check_shop // TODO nice icon would be nice (reuse graphic from building quest?)
```
westnordost, line 29: I tried to make an icon with that, but I wasn't happy with it. I think this icon is fine. The interface is also identical to that other quest.

matkoniecz:
> If this is added, someone should monitor the notes being created for this quest and see if there is any need to act.

I am fine with that, in the worst case it can be reverted and removed. I am also fine with some work on improving OSM Wiki documentation (and iD preset pull requests, if iD development will restart).

I am occasionally looking at OSM notes produced by SC users (English and Polish ones) and opened some PRs related to that, latest was cycleway related about making more clear how to tag "no cycleway here" case.

westnordost:
Or maybe iD preset development can restart? Could the active contributors to that preset index not be given push access? Maybe you could enquire this.

westnordost (approved):

westnordost (changes requested):
Ah, maybe it should be added to that one achievement

matkoniecz:
> Could the active contributors to that preset index not be given push access? Maybe you could enquire this.

I may be not the best person for that given `highway=track` situation (I still think that iD preset is misleading).

matkoniecz:
> Ah, maybe it should be added to that one achievement

Added.

westnordost:
Well, that's basically a translation issue. IIRC in the German version of that preset, `highway=track` is denoted correctly. But anyway, if this is a potential problem, surely possible contributors with push access can come to an agreement not to change existent presets unilaterally.

westnordost (approved):

matkoniecz:
> Well, that's basically a translation issue. IIRC in the German version of that preset, highway=track is denoted correctly. But anyway, if this is a potential problem, surely possible contributors with push access can come to an agreement not to change existent presets unilaterally.

Yes, and I can restrict myself. But even opening track-adjacent issues was considered negatively so asking for write access (even if promising to not touch track definition) may be also not welcomed.

PR #2766, opened 2021-04-19
Tag stile type, fixes #2749

### LINKED ISSUE #2749: Stile type
### General
Affected tag(s) to be modified/added: [stile](https://wiki.openstreetmap.org/wiki/Key:stile)
Question asked: **What type of stile is this?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose - https://wiki.openstreetmap.org/wiki/Talk:Tag:barrier%3Dstile#Is_stile_type_used_for_anything_or_potentially_useful.3F
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one) - https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dstile#Stile_details

> stile=*, often together with step_count=*, is generally of very high value & use. Different types of stiles can pose difficulties for older people and those with limited mobility. For instance a squeezer with step_count=1 will generally be readily accessible, a ladder stile may be far too difficult. I very recently searched for all stiles on a particular route in the Peak District so that we can try & avoid the difficult ones to enable my father to climb a hill. Material is also useful in some contexts (squeezers, stepovers): a stepover made of stone with a high step count is likely to be beyond someone like my Dad, whereas a wooden one with step count of 2 or less he can manage (albeit with a bit of help)

### Ideas for implementation

**Element selection:** `barrier=stile !stile`

**Metadata needed:** No

**Proposed GUI:** Image selector

matkoniecz:
@westnordost Do you have an idea for an icon?

screen14


app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.osmquest.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.changes.StringMapChangesBuilder
+
+class AddStileType : OsmFilterQuestType<BarrierType>() {
+
+    override val elementFilter = """
+        nodes with barrier=stile and !stile
+    """
```
westnordost, line 11: exclude private barriers?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
```diff
@@ -0,0 +1,46 @@
...
+            BarrierType.STILE_LADDER -> {
+                changes.add("stile", "ladder")
+            }
+            BarrierType.STILE_STEPOVER -> {
+                changes.add("stile", "stepover")
+            }
+            BarrierType.KISSING_GATE -> {
+                changes.modify("barrier", "kissing_gate")
+                changes.deleteIfExists("step_count")
+                changes.deleteIfExists("wheelchair")
```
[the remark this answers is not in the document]
matkoniecz: In thiis case it was tagged as stile, so wheelchair tag - if tagged - described stile accessibility, so if present should be removed.

app/src/main/res/drawable/ic_quest_cow.xml
```diff
@@ -5,20 +5,24 @@
     android:viewportHeight="128">
```
westnordost, line 5: What did you do here?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
westnordost: yes, no, perhaps?

matkoniecz: there are some with access=private/no but with foot=designated - https://www.openstreetmap.org/node/5000261512

I found one without extra tags ( https://www.openstreetmap.org/node/3978247540 ), but looks like a tagging mistake https://www.openstreetmap.org/note/2632723

`foot=yes` `foot=permissive` also sometimes amend `access`

And even for private stiles asking still is not harmful.

Though https://www.openstreetmap.org/node/7450489924 https://www.openstreetmap.org/node/891091396 https://www.openstreetmap.org/node/6034850648 https://www.openstreetmap.org/node/3707227051 seem actually private

But it is just tiny, tiny fraction of all. https://overpass-turbo.eu/s/16rU

------------------

So if added filter would block on `access=private` `access=no` and let `foot=yes` `foot=permissive` `foot=designated` to override it.

westnordost: Actually I meant access on the way that crosses the stile

matkoniecz: I will check but likely the same applies, especially in UK such rights of way are common (path on a private land with right of way for pedestrians)

westnordost: We have `access !~ private|no` for many quests

app/src/main/res/drawable/ic_quest_cow.xml
matkoniecz: no idea at all what happened here, I was importing drawable that I see was apparently already present but not detected - maybe due to some che issues

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
matkoniecz: done, tested

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
```diff
@@ -0,0 +1,69 @@
...
+        val excludedWayNodeIds = mutableSetOf<Long>()
+        mapData.ways
+            .filter { excludedWaysFilter.matches(it) }
+            .flatMapTo(excludedWayNodeIds) { it.nodeIds }
+
+        return mapData.nodes
+            .filter { stileNodes.matches(it) && it.id !in excludedWayNodeIds }
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? = null
```
westnordost, line 33: could be false if `stileNodesFilter` does not match, otherwise null

PR #2768, opened 2021-04-19
Update some links to point directly to the SC org

peternewman:


westnordost:
thanks!

peternewman:
No problem. Am I right in thinking you said there's no sensible way to batch-edit the string in POEditor or by modifying the XML files?

westnordost:
Correct: Of course the XMLs could be batch-edited but then they would all need to be uploaded to POEditor by hand. It is easier to edit them in POEditor directly.

Why, what does need to be changed for strings in POEditor?

peternewman:
> Correct: Of course the XMLs could be batch-edited but then they would all need to be uploaded to POEditor by hand. It is easier to edit them in POEditor directly.

Oh wow, didn't realise there was no way to push them up!

> Why, what does need to be changed for strings in POEditor?

Because translations and functionality haven't been entirely separated in the GUI:

```diff
-    <string name="credits_contributors">"See the &lt;a href=\"https://github.com/westnordost/StreetComplete/graphs/contributors\"&gt;complete list on GitHub&lt;/a&gt;."</string>
+    <string name="credits_contributors">"See the &lt;a href=\"https://github.com/streetcomplete/StreetComplete/graphs/contributors\"&gt;complete list on GitHub&lt;/a&gt;."</string>
```

westnordost:
Ok I changed all the translations in poeditor

peternewman:
> Ok I changed all the translations in poeditor

Amazing, thanks very much!

That looks like aside from authors.txt (which I assume isn't published anywhere), everything else is just in comments, maybe I'll run sed across a git checkout at some point to fix them, but they seemed less important.

Issue #698, opened 2017-11-20
Width of a cycleway/cyclelane


pietervdvn:
### General

Another factor in the comfort of cyclists is the width of a cycleway/cyclelane. A two meter cycleway is way nicer to drive then a 1m one.

**Example question**: `What is the width of this cycleway (in meters)?`

Affected key(s) (or tags) to be modified: [width](https://wiki.openstreetmap.org/wiki/Key:width)

### Checklist
Checklist for quest suggestions (see guidelines):
- [X] 🚧 To be added tag is established and has a useful purpose
- [X] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [ ] 🐿️ Easily answerable by everyone from the outside
- [X] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [X] 🕓 Applies to a reasonable number of elements (Worth the effort)

**This is not an easy question**. Users should have a measuring tape on them. Therefore, I propose that this quest be disabled by default.
Furthermore, measuring widths of cycleways can hinder other cyclists or can be dangerous. A warning message should be included.

### Ideas for implementation

**Element selection:** 

        way["cycleway"]["width"!~".*"]["cycleway:width"!~".*"]({{bbox}});
        way["cycleway:left"]["cycleway:left:width"!~".*"]({{bbox}});
        way["cycleway:right"]["cycleway:right:width"!~".*"]({{bbox}});

**Proposed GUI:** a number field.

It should show a few examples of how to measure, e.g.

        -----------+
                     |                +-----------------------------------------------------------+----+
                     +-------------+          (cyclepath)     (line defining cycleway    ^  )
        (curb)       (gutter)   <-------- Width to measure, including lines ---------->

westnordost:
Perhaps the smartphone could be made into a measuring tape, using the rotation sensors and the camera.

rugk:
…if that works (is accurate enough), that is an awesome idea!

westnordost:
With triangulation:
1. The app must know in which height **h** from the ground the smartphone is. (user must measure that once with a measuring tape and then commit to always holding the smartphone in the same height when measuring)
2. User points through camera to a point, presses screen -> app registers device Z rotation **α**
3. User presses screen again -> app registers device Z rotation **β**
(User sees a line being drawn on top of camera picture in step 2/3 from starting point to end point)

The distance on the ground is then `h * tan (abs(α-β))`, basically.

[...]

PR #2788, opened 2021-04-22
[One-way quest] Also show quest for streets where there are no sidewalks

naposm:
Hi, it's me again. This morning, while out, I noticed that most of the roads which were one way had no sidewalks, indeed they were so narrow that adding sidewalks would have made car access impossible. 
I must specify that in southern Italy (at least in my area) we have a lot of narrow roads like that and that's true almost always (no sidewalks = narrow = one-way road), but that could not be the always true for all other roads in the world.
But I think that this would not cause much spam and could actually be useful if asked after the sidewalk quest (maybe also after the lane quest). 
With the current settings, this quest is almost never shown as I think there's no quest that asks for width (correct me if I'm wrong, they are so many 😂), so adding this other option (and maybe more in the future) could actually help as this quest is maybe the only one which can help mapping one-way roads in areas where there's not much data for the suspect one-way quest.
Let me know what you think about it!

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt
@@ -25,11 +25,8 @@ class AddOneway : OsmElementQuestType<OnewayAnswer> {
         ways with highway ~ living_street|residential|service|tertiary|unclassified
          and !oneway and area != yes and junction != roundabout
          and (access !~ private|no or (foot and foot !~ private|no))
-         and (
-               (lanes <= 1 and width)
-              or
-               (sidewalk ~ none|no)
-             )
+         and lanes <= 1 
+         and (width or sidewalk ~ none|no)
     """.toElementFilterExpression() }
 
     override val commitMessage = "Add whether this road is a one-way road because it is quite slim"
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt
@@ -25,7 +25,11 @@ class AddOneway : OsmElementQuestType<OnewayAnswer> {
         ways with highway ~ living_street|residential|service|tertiary|unclassified
          and !oneway and area != yes and junction != roundabout
          and (access !~ private|no or (foot and foot !~ private|no))
-         and lanes <= 1 and width
+         and (
+               (lanes <= 1 and width)
+              or
+               (sidewalk ~ none|no)
+             )
     """.toElementFilterExpression() }
 
     override val commitMessage = "Add whether this road is a one-way road because it is quite slim"


matkoniecz:
Have you tested this quest? From looking at the code ( https://github.com/streetcomplete/StreetComplete/pull/2788/files#diff-27fbe30eebafd2d2bbefd9efa6d753e351c2813027a658ea63e09c6242b65712R80 ) it seems that your change will not cause this quests to appear.

In general I am not entirely sure. I would at east continue to demand <=1 lane. See https://www.openstreetmap.org/way/19801536 https://www.openstreetmap.org/way/234878375 that IMHO should not trigger this quest.

> With the current settings, this quest is almost never shown as I think there's no quest that asks for width

@westnordost is planning to add one (though it is quite big project so...)

naposm:
I have not tested it, but I think it is the same as 
IF everything before AND ( (has lanes <= 1 and width) or (hasn't sidewalks) ) THEN show quest maybe I got something wrong? I might test it.

Maybe you're right, it should be something like this

IF everything before AND ( (has lanes <= 1 or hasn't sidewalks ) AND width THEN show quest

matkoniecz:
Note linked function: it checks whether `width` is a small value.

It discards any cases where `width` is not tagged.

andrewharvey:
>IF everything before AND ( (has lanes <= 1 and width) OR (hasn't sidewalks) ) THEN show quest maybe I got something wrong? I might test it.

:-1: from me because this would mean every road without sidewalks would be asked if it's oneway, many places have no sidewalks at all but oneways would be <10% of roads. I don't think we need to explicitly mark oneway status just because of lack of sidewalks.

> IF everything before AND has lanes <= 1 AND (hasn't sidewalks OR has width tag) THEN show quest.

:-1: from me, because I think it's helpful to ask if a road is oneway if it only has a single lane regardless of if it has sidewalks. Many ways tagged as a single lane, might be oneway even with sidewalks.

naposm:
> -1 from me because this would mean every road without sidewalks would be asked if it's oneway, many places have no sidewalks at all but oneways would be <10% of roads. I don't think we need to explicitly mark oneway status just because of lack of sidewalks.

You're right, indeed I was wrong and followed @matkoniecz suggestion, which makes way more sense, this is an outdated suggestion, but thanks for pointing that out anyway, it's actually true what you're saying

> -1 from me, because I think it's helpful to ask if a road is oneway if it only has a single lane regardless of if it has sidewalks. Many ways tagged as a single lane, might be oneway even with sidewalks.

So you mean something like this:
IF everything before AND has lanes <= 1 THEN show quest ?

P.S. Note that currently the quest only shows up only if there's the width tag AND the lanes tag, so this quests adds a feature, it doesn't limit it. At the moment it is like this
IF road AND has lanes <= 1 AND has width tag THEN show quest
just saying because as I understand your -1 is based on the fact that you think that checking for sidewalk is useless, but there's already a check for another tag in the current implementation.

naposm:
> Note linked function: it checks whether `width` is a small value.
> 
> It discards any cases where `width` is not tagged.

Maybe I missed something, isn't width this tag? https://wiki.openstreetmap.org/wiki/Key:width

matkoniecz:
> Maybe I missed something, isn't width this tag? https://wiki.openstreetmap.org/wiki/Key:width

Yes. But your change affects only initial query - and further filtering in https://github.com/streetcomplete/StreetComplete/blob/91049d7f0fd8669a4d6659fddf7778439146a01e/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt#L77 seems to be discarding any ways without `width` tagged.

(I have not tested it so far, in general I am quite dubious is adding `oneway=`on all roads with one lane and without sidewalks is useful)

naposm:
> > Maybe I missed something, isn't width this tag? https://wiki.openstreetmap.org/wiki/Key:width
> 
> Yes. But your change affects only initial query - and further filtering in
> 
> https://github.com/streetcomplete/StreetComplete/blob/91049d7f0fd8669a4d6659fddf7778439146a01e/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt#L77
> seems to be discarding any ways without `width` tagged.
> 
> (I have not tested it so far, in general I am quite dubious is adding `oneway=`on all roads with one lane and without sidewalks is useful)

Yes, my bad, I missed everything below, I apologize!

Anyway this is aimed at roads like these, I'll leave a Google StreetView link:
https://www.google.com/maps/@40.8477984,14.263126,3a,75y,261.23h,77.62t/data=!3m6!1e1!3m4!1se9Eqf6kyzok8gmT_J-M6fw!2e0!7i16384!8i8192

westnordost:
If it is aimed at roads like this, then it does make more sense to record the width (first) and only after the width is known, ask for whether it is a oneway or not

naposm:
> If it is aimed at roads like this, then it does make more sense to record the width (first) and only after the width is known, ask for whether it is a oneway or not

If there will be a quest for width, then yes. Indeed mine was a suggestion for width "prediction" without having the width.
If a quest for width is planned, then this won't be useful anymore, so we can close this PR if you want. 😅

matkoniecz:
There is #698 but once implemented it could be also easily applied also for such roads.

westnordost:
Yes, it is planned

naposm:
That would be awesome, thanks!

PR #2794, opened 2021-04-24
Download satellite imagery for offline

smichel17:
The surrounding pieces are in place; the actual download is not implemented yet. I'll probably also do a little more cleanup at the end.

What's the preferred way of getting access to the preferences from within the MapTilesDownloader? Should I add another (injected) argument, or something else?


app/src/main/java/de/westnordost/streetcomplete/map/components/SceneMapComponent.kt
```diff
@@ -62,10 +62,13 @@ class SceneMapComponent (
         getBaseSceneUpdates() + sceneUpdates.map { SceneUpdate(it.key, it.value) }
 
     private fun getBaseSceneUpdates(): List<SceneUpdate> {
+        val activeTileSource = vectorTileProvider.let {
+            if (sceneMapComponent?.isAerialView == true) it.rasterLayerSource else it.baseTileSource
+        }
```
westnordost, line 67: You misunderstood something there. When the aerial view is active, **both** the raster layer source and the vector tile source is used.

app/src/main/java/de/westnordost/streetcomplete/map/VectorTileProvider.kt
```diff
@@ -1,13 +1,18 @@
 package de.westnordost.streetcomplete.map
 
-abstract class VectorTileProvider(
+data class VectorTileProvider(
+    val sceneFilePath: String,
+    val baseTileSource: TileSource,
+    val rasterLayerSource: TileSource,
```
westnordost, line 6: maybe it would be easier if this was simply a list of tile sources, so then it is not necessary to discriminate between "baseTileSource" and "rasterLayerSource" everywhere.

On the other hand, if you plan to only . But as said, maybe first look at what is the actual additional download size of always also downloading the satellite imagery. If it is not much more, it is maybe not worth it to **not** always also download the satellite imagery.

app/src/main/java/de/westnordost/streetcomplete/map/components/SceneMapComponent.kt
smichel17: Ah, you're right, this should always be using the base. That won't work if we ever add another source that needs an api key (thus, two needed), but for now it's fine; I'll revert this part.

app/src/main/java/de/westnordost/streetcomplete/map/VectorTileProvider.kt
smichel17: This way forces me to discriminate, but the other way forces me to use magic numbers to know which source is which; to me it's much clearer that the base tile source is always used and the raster source is sometimes used. 

I imagined that if we add other renderings in the future, the base layer is probably still used and it's just the visual layer that changes, so I was imagining that in this case, the `rasterLayerSource` would be converted to an array. But while I read enough of the tangram docs and skimmed our yaml to get a sense of what a scene is, I didn't learn the details yet, so maybe this is the wrong abstraction.

I think I should probably just rename `raster` to `aerial` here, and I should add parts of the paragraph above to the commit message.

smichel17:
I was planning to conditionally download the aerial tiles; I don't think it will be that much more work. But I'll do it unconditionally first and then add that in a separate commit, to confirm :)

app/src/main/java/de/westnordost/streetcomplete/map/VectorTileProvider.kt
westnordost: > This way forces me to discriminate, but the other way forces me to use magic numbers to know which source is which

Why do you need to know which is which?

smichel17: Because the base layer is always downloaded and the aerial layer isn't (or if it is, different cache rules will be used).

smichel17: I suppose it would be possible that we didn't need to know which is which, if we stick more information into the `VectorTileProvider` (and rename it to `MapScene` as you suggested earlier). E.g. this info should be stored in the `MapScene`:

https://github.com/streetcomplete/StreetComplete/blob/0d70f5261e1e8c198f24671a94b23eb0a68c626e/app/src/main/java/de/westnordost/streetcomplete/map/components/SceneMapComponent.kt#L82-L86

So then we'd have 3 different map scenes — light, dark, and aerial. Light and dark would each have only one TileSource, and Aerial would have two. All sources would always be downloaded unconditionally. Cache rules would be stored in the data source. This has better separation of concerns, and will scale much better if additional scenes are ever added. So *for the moment* I'm convinced this is a better architecture.

On the other hand, it's also an additional layer of abstraction, which cements this interface. So, I'm leaning towards staying with the baseLayer / aerialLayer split *for now*. Then, later, once the feature has been released for a bit and any related feature requests have come in, we can see if this interface still makes sense, and I'll send a refactoring PR if it does.

app/src/main/java/de/westnordost/streetcomplete/map/components/SceneMapComponent.kt
smichel17: Removed this commit.

app/src/main/java/de/westnordost/streetcomplete/map/VectorTileProvider.kt
smichel17: Added a longer explanation to this effect in the first commit

smichel17:
Rebased on master, implementing the download now. Two thoughts:

- Is there ever a minimum zoom? Right now we get the tiles to download with
   ```kotlin
   (vectorTileProvider.baseTileSource.maxZoom downTo 0).asSequence().flatMap { zoom ->
       bbox.enclosingTilesRect(zoom).asTilePosSequence().map { Tile(zoom, it.x, it.y) }
   }
   ```
   For aerial imagery, do we still want to go down to zero?

- It makes me slightly uncomfortable that we list the same url endpoint in two places (the `.yaml` scene files and the `MapModule` / `TileSource#getTileUrl()`). Maybe we either `getTileUrl()` could parse the endpoint from the relevant scene file, or the `SceneMapComponent` could update the scene and set that info, like it does for the api key (and a few others)?

matkoniecz:
> For aerial imagery, do we still want to go down to zero?

Yes, https://server.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/tile/0/0/0 is still showing something. Not entirely sure is it ever used by SC ona typical device, but it is a single tile anyway.

> could parse the endpoint from the relevant scene file

That would be likely as fragile as repeated info in two places, parsing has its own share of fun.

>  or the SceneMapComponent could update the scene and set that info

I feel that adding comment in https://github.com/streetcomplete/StreetComplete/blob/3da55f0bee4b6e7f0edfb3f33ca66dc7110d1635/app/src/main/assets/map_theme/jawg/streetcomplete-satellite-style.yaml#L10 would be good enough - and would be less confusing to check how map style actually works

smichel17:
Got into a nice long rabbit hole reading about coroutines, but made it back and added a simple/naive "always download both" implementation, which is ugly (and I will clean up later) but mostly works. What doesn't work is zooming in past the maxZoom, which is arbitrarily set at 16 because that's what Jawg was at. In this case, rather than enlarging the image, the background disappears entirely.

What's the best way to handle this? Does some configuration in the map scene need to change? Do we need to download tiles at a higher max zoom? I didn't pull out the stats yet so I'm not sure how much of the download size is raster vs vector tiles yet (this will be part of cleaning up).

westnordost:
I backed out hte visibility of the satellite setting on master, so you would need to uncomment this in your PR

smichel17:
> What's the best way to handle this? Does some configuration in the map scene need to change? Do we need to download tiles at a higher max zoom?

I would appreciate if someone knows the answer here. It's the kind of thing that might take me a long time to figure out (with my limited knowledge of osm) but that someone else might be able to answer quickly, or point me in the right direction to look/test.

westnordost:
Maybe it is a setting in the YAML map style? There is a documentation for the tangram map styles.

smichel17:
Fixed the zoom bug, I think. The problem was that Tangram chooses its max_zoom based on the yaml scene file, not the TileSource declared on kotlin-side. It was not specified, and the default max_zoom is 18. So, Tangram was trying to show tiles at zoom levels 17 and 18 -- and failing, of course, since the phone is offline and they were not downloaded ahead of time. Left to do:

- [ ] Refactor `MapTilesDownloader::download` to separate stats for base & satellite layers
- [ ] Use stats to determine whether it's ok to download satellite layers unconditionally or whether we need a smarter download strategy, and implement that strategy, if needed. Options I can think of, if needed:
    - Download aerial tiles at a lower resolution (lower max_zoom)
    - Only download if the map cache is set to a high enough value
    - Only download if the satellite style is enabled at time of download
- [ ] Consider refactoring, as discussed [above](https://github.com/streetcomplete/StreetComplete/pull/2794#discussion_r619664025).
- [ ] Request review / testing

app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
```diff
@@ -30,19 +31,20 @@ class MapTilesDownloader @Inject constructor(
...
-                        downloadTile(tile.zoom, tile.x, tile.y)
-                    } catch (e: Exception) {
-                        DownloadFailure
-                    }
-                    ++tileCount
-                    when (result) {
-                        is DownloadFailure -> ++failureCount
-                        is DownloadSuccess -> {
-                            if (result.alreadyCached) cachedSize += result.size
-                            else downloadedSize += result.size
```
smichel17, line 45: So, there's actually a concurrency bug here: `++` and `+=` are not atomic, so the different coroutines we `launch` here can overwrite each other's values mid-flight. It's the same [shared mutable state problem mentioned in the docs](https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#the-problem). This doesn't happen too often, because we're not downloading thousands of tiles, but it does happen. In a single test of around-75-percent-cached tiles, I got
```
06-19 20:10:40.084  3899  4031 I MapTilesDownload: Caught 163 tiles of 164
```
<details>
<summary>Code</summary>

```diff
diff --git a/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt b/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
index 4ff149169..da95426a2 100644
--- a/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
@@ -12,6 +12,7 @@ import okhttp3.*
 import okhttp3.internal.Version
 import java.io.IOException
 import java.lang.System.currentTimeMillis
+import java.util.concurrent.atomic.AtomicInteger
 import javax.inject.Inject
 import kotlin.coroutines.resume
 
@@ -29,6 +30,7 @@ class MapTilesDownloader @Inject constructor(
         var downloadedSize = 0
         var cachedSize = 0
         val time = currentTimeMillis()
+        val reliableCounter = AtomicInteger()
 
         coroutineScope {
             listOf(
@@ -39,6 +41,7 @@ class MapTilesDownloader @Inject constructor(
                     launch {
                         val result = downloadTile(source, tile.zoom, tile.x, tile.y)
                         ++tileCount
+                        reliableCounter.incrementAndGet()
                         when (result) {
                             is DownloadFailure -> ++failureCount
                             is DownloadSuccess -> {
@@ -52,6 +55,7 @@ class MapTilesDownloader @Inject constructor(
         }
         val seconds = (currentTimeMillis() - time) / 1000.0
         val failureText = if (failureCount > 0) ". $failureCount tiles failed to download" else ""
+        Log.i(TAG, "Caught $tileCount tiles of ${reliableCounter.get()}")
         Log.i(TAG, "Downloaded $tileCount tiles (${downloadedSize / 1000}kB downloaded, ${cachedSize / 1000}kB already cached) in ${seconds.format(1)}s$failureText")
     }
```
</details>

Since the data here are simple, the quickest solution would be to change out the integers for AtomicIntegers (and, I may end up doing this anyway for performance). However, I want to clean up this part of the code anyway, so I'm planning on (read: mostly done with) converting it to use [flows](https://kotlinlang.org/docs/flow.html), which are basically "sequences but for coroutines".

westnordost: But it doesn't matter if they overwrite each other. Whether the log reads that either 163 or 164 tiles have been downloaded hardly matters. I'd prefer not to use `AtomicInteger` because that's Java and I try to not have the core classes depend on Java dependencies.

smichel17: Well, it means that the `downloadedSize` is not accurate; it is a little larger than the reported value.

mnalis:
As noted in https://github.com/streetcomplete/StreetComplete/issues/3137, there is potential that this feature would likely be used for mapping things not surveyed on the ground. While it is is not necessarily  bad in itself, it does heavily influence decisions made by other mapper (so far, StreetComplete changets could've be assumed to be made from the data visible on the ground, with this PR it is no longer the case - eg. roof type and more might actually be from satellite images).

I think it should **definitely be noted** (eg. by changing changeset  tag from `source=survey` to `source=Esri` or `source=Esri;survey` if map provider was changed from default to indicate the fact)

HolgerJeromin:
Many editors uses https://wiki.openstreetmap.org/wiki/Key:imagery_used in the changeset for that.

app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
smichel17: I *would* like to fix this before proceeding, because having an accurate picture of download sizes is important for deciding whether to always download aerial images or only when the setting is enabled.

The idea I'm currently planning to go with is to iterate *indexed* inside a coroutineScope and store the DownloadStats result into an ArrayList by index. Then I'll sum the stats at the end, outside the coroutineScope, where I know all the coroutines have finished.

westnordost: It makes the code more makes though. Before you commit this,  I'd like to ask you to log via both methods ans see by how much the two numbers differ. I would expect that the imprecision through race conditions is miniscule and this doesn't matter.  But i may be wrong.

smichel17:
I removed the flows refactor and added the array version. Then I tested downloads in a few areas.

- To my surprise, at **maximum zoom** level, ESRI imagery is actually ***less* data** than the Jawg tiles, regardless of location. Usually around 1/3 as big, sometimes 1/2.
    - I wonder if this is related in some way to
- The ratio grows as you zoom out (ESRI data is *comparatively* larger). How much? It depends…
- When zoomed out as far as SC will allow, there is quite a large amount of variance depending on location.
    - In the rural area I tested (Northampton, MA, US), ESRI data was 2-8x larger than Jawg data.
    - In urban areas (Boston and Hamburg), ESRI data was sometimes smaller and sometimes larger than the Jawg data— between 0.8x and 3x the size. I could not predict the ratio, as it would change between those limits with not very much scrolling.

I am not sure if the location difference was actually because of rural vs urban, or because the data in those cities is more complete, or something else. I am also confused why the zoom level would have such a large impact on the data ratio. Perhaps it has to do with the zoom level of available tiles?

Anyway, my conclusion is that we probably **need more testing**. The code I tested with is here: https://github.com/smichel17/StreetComplete/commit/30bac73bca62e6bf153e0727f01b9833cde1b0cc

mnalis:
@smichel17 I was under impression that vector tiles are downloaded **once** and work for **all zoom levels** (eg. they are re-rendered locally on the mobile device). 

Bitmapped images (like satellite / aerial imagery) on the other hand need **separate download for each zoom level** so they are bigger (if downloaded for all zoom levels) and have interesting side-effects if not all zoom levels are downloaded (eg. map is shown on one zoom level nicely, but completely disappears on other zoom levels; or even worse it is shown but looks _horrible_ because device tried to resize the bitmap) 

(I do not know about StreetComplete code specifically though, so take that with a grain of salt)

westnordost:
@mnalis No, vector tiles also exist for different zoom levels.

mnalis:
@westnordost thanks, interesting approach! I can see how is is useful bandwidth-wise if one starts zoomed out, and then (occasionally) zooms in. I can't see there however if the jawg is smart enough to download only missing data (as it should be):

- eg. if one starts at `z=15` (and has thus downloaded admin, aeroway, landuse, building etc.) and then zooms in a little more to `z=16`, will jawg be smart enough to **only** download missing `housenum_label` vectors (which are only available on min `z=16`), or will it download again **everything** for that area, including vectors it already had (thus behaving "dumbly" like bitmap tiles, who always have to download full new set of tiles for each new zoom level)?
- conversely if one starts at `z=16` and moves around (eg. download all details for bigger area) and then zooms out a little to `z=15`, jawg should not have to download anything (if it is smart enough)? (whereas bitmapped tiles would have to download completely new set of tiles, as it is  different zoom level)

westnordost:
That has nothing to do with jawg, that's how vector tiles work. Everything will be downloaded "again" but it may very well be that the geometry at lower zoom levels is somewhat simplified.

smichel17:
I'm looking to wrap this up. I mainly volunteered to take this on because it seemed relatively easy (which it has been, research aside), was an opportunity to learn more of SC's code base, and I felt like I hadn't contributed enough code. Now I feel like I've learned what there is to learn and I've helped out with other parts of the code recently.

Current status: we need to decide a download strategy. Always cache aerial tiles, only when aerial view is enabled, or something else? We have some rough statistics; possibly better stats would point to an obvious answer.

I'm thinking perhaps I'll implement unconditional download; remove the commit which re-enables the preference, so this can be merged but still not enabled; and let someone else take it from there.

westnordost:
> I'm thinking perhaps I'll implement unconditional download; remove the commit which re-enables the preference, so this can be merged but still not enabled; and let someone else take it from there.

Implementing unconditional download but somehow disabling the feature would be fine to me. It will then be up to someone else (@matkoniecz , since he started it?) to do the rest, namely to decide whether unconditionally downloading always satellite imagery or to only download if it is turned on at the time the tiles are downloaded.

matkoniecz:
Yeah, I will put in on my TODO list back again.

smichel17:
@matkoniecz Would you rather I merge something with the download disabled, or keep download unconditional and send this PR against a different branch in this repo where you can continue it?

matkoniecz:
> somehow disabling the feature would be fine to me

So it would continue being hidden from user?

> download disabled, or keep download unconditional

If it would be hidden from user then no downloading should happen for now

> namely to decide whether unconditionally downloading always satellite imagery or to only download if it is turned on at the time the tiles are downloaded

Ideally it would be predownloaded on wifi and fetched when needed on data connection - what again is increasing complexity...

mnalis:
> If it would be hidden from user then no downloading should happen for now

I agree. In particular, I think download of satellite imagery should **only happen when user has explicitly selected** that option. So for majority of SC users (the simple ones, eg. who do not go clicking around different settings), no download of satellite imagery should ever happen (in order not to increase their data bills,  nor slow down processing, nor use extra space on their phone, nor cause extra battery drain).

smichel17:
Oi, oi! @mnalis We're not talking about the final behavior, we're talking about the hand-off point where I stop working on this and @matkoniecz picks it up. @matkoniecz the two options I proposed were:

1. I keep the setting hidden, implement the conditional download (which of course means always disabled, since the setting is hidden), and this PR is merged to master.
2. I implement unconditional download, close this PR, and push to a non-master branch in this repo.

westnordost:
@Helium314 wrote in #3306:

> When creating a note via long click with aerial image background enabled the app crashes.
> 
> I know this is a (currently) disabled feature, but I report anyway because I have the impression that the technical part besides download for offline use is considered done.
> 
> Reproduced using my smoothness quest build, so no wacky modifications from my side.
Preferences xml file was manipulated to enable aerial background.
(can't test on current master because it's throwing tons of build errors, sorry if it is already fixed there)
> 
> <details>
>  <summary>Log</summary>
> 
> <code>
> Thread: main
> App version: 34.0
> Device: samsung serranovelte, Android 9
> Locale: en_GB
> Stack trace:
> de.westnordost.streetcomplete.map.tangram.LoadSceneException: scene update path not found
> at de.westnordost.streetcomplete.map.tangram.KtMapControllerKt.toException(KtMapController.kt:362)
> at de.westnordost.streetcomplete.map.tangram.KtMapControllerKt.access$toException(KtMapController.kt:1)
> at de.westnordost.streetcomplete.map.tangram.KtMapController._init_$lambda-0(KtMapController.kt:65)
> at de.westnordost.streetcomplete.map.tangram.KtMapController.lambda$Bq4Dpp7zu-6-cr6v3_0B738-Wm0(Unknown Source:0)
> at de.westnordost.streetcomplete.map.tangram.-$$Lambda$KtMapController$Bq4Dpp7zu-6-cr6v3_0B738-Wm0.onSceneReady(Unknown Source:2)
> at com.mapzen.tangram.MapController$9.run(MapController.java:1214)
> at android.os.Handler.handleCallback(Handler.java:873)
> at android.os.Handler.dispatchMessage(Handler.java:99)
> at android.os.Looper.loop(Looper.java:193)
> at android.app.ActivityThread.main(ActivityThread.java:6718)
> at java.lang.reflect.Method.invoke(Native Method)
> at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:491)
> at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
> </code>
> </details>

smichel17:
@helium314 unless you have merged this PR to your branch, offline download is not enabled for Aerial imagery.

Helium314:
> @Helium314 unless you have merged this PR to your branch, offline download is not enabled for Aerial imagery.

I did not, and I think @westnordost meant to forward the comment to @matkoniecz's merged PR #2755

smichel17:
I'm planning on getting this to the hand-off point soon. @matkoniecz I'd still like to hear which of the following two options you would prefer. If you don't care (or don't reply soon), I'll go with option 2, since it's easier for me :P

> 1. I keep the setting hidden, implement the conditional download (which of course means always disabled, since the setting is hidden), and this PR is merged to master.
> 
> 2. I implement unconditional download, close this PR, and push to a non-master branch in this repo.

smichel17:
As @peternewman [pointed out](https://github.com/streetcomplete/StreetComplete/issues/3355#issuecomment-940929797), I failed the changelog merge :(

https://github.com/streetcomplete/StreetComplete/blob/7519268ad4499ca65037f706b22c732b2316627c/app/src/main/res/raw/changelog.yml#L162

I amended it now and forced pushed to the branches mentioned above. The new merge commit is 2735f3a783e and the new refactor commits are bc7a1f6544 and 61664ffae. I edited the message above with the new commit to avoid accidental confusion.

PR #2796, opened 2021-04-24
Check for container recycling still existing

peternewman:
Untested

This was triggered by some clothes recycling bins which have moved within a few months of being added (and I think initially appeared during the pandemic). I don't know if they are deliberately moving it around or were just trying to find the best spot to place it.

Could narrow it down to just certain types if we don't feel the glass/plastic/paper ones are very likely to move (which I'd probably agree as they're normally council based).

I think the one I saw was more almost charity fundraising like this one:
https://commons.wikimedia.org/wiki/File:Clothing_recycling_bin.JPG

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
@@ -42,6 +42,7 @@ class CheckExistence(
             or advertising ~ column|board|poster_box
             or traffic_calming ~ bump|hump|island|cushion|choker|rumble_strip|chicane|dip
             or traffic_calming = table and !highway and !crossing
+            or amenity = recycling and recycling_type = container
           )
           and (${lastChecked(4.0)})
         )) and access !~ no|private


westnordost:
Hmm, I missed this PR.

But this is not necessary. The question about what is being recycled is already asked again every few years. So if it does not exist anymore, the user will answer that in that quest.

peternewman:
> Hmm, I missed this PR.

No worries.

> But this is not necessary. The question about what is being recycled is already asked again every few years. So if it does not exist anymore, the user will answer that in that quest.

But similar to the post box quest, we check existence and collection times independently. I think the clothing ones specifically deserve a special case, as they are often fundraising, so run by a charity, so they'll want to optimise their income, whereas glass/plastic/paper is likely to be council based so wants to be in one static place so you can always find it (although as kerbside recycling has increased they seem to have mostly gone away aside from big central council sites).

Maybe it's just a pandemic thing, but as I say I've seen quite a few clothing boxes move around relatively frequently, certainly far more than every few years. Also if it's done as part of existence we don't need to bother the user with asking the collection type with that frequency.

westnordost:
Hm then why is the post box even in the filter for the CheckExistence quest. Maybe it should be removed then instead recycling containers be added.

Specifically for clothes recycling containers, there is also the "wat operator is this?" quest. Though this is not asked again, and asking it again probably doesn't make sense.

peternewman:
> Hm then why is the post box even in the filter for the CheckExistence quest.

Because not everywhere has collection times for postboxes but might still remove them occasionally:
https://github.com/streetcomplete/StreetComplete/blob/b7907e2b4ce7d414f8501ccddfc5110eee6e1f5e/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt#L34-L47

> Maybe it should be removed then instead recycling containers be added.
> 
> Specifically for clothes recycling containers, there is also the "wat operator is this?" quest. Though this is not asked again, and asking it again probably doesn't make sense.

Yes, this is sort of the crux of my point and why I think it should be in the existence quest too. If it's still there, it's unlikely the operator or what it collects has changed, or certainly won't change very frequently, plus the "effort" of gathering that info is more than a simple yes/no, particularly typing the operator.

Whereas the simple yes/no is low effort for people to answer, but high benefit as it keeps the map current, hence I think it should be added additionally.

westnordost:
> Because not everywhere has collection times for postboxes but might still remove them occasionally:

Right, ok.

westnordost:
Ok then, I'll merge this.

PR #2805, opened 2021-04-26
Ensure the list of tags checked for to decide the title type matches those used to fetch the name

peternewman:
Fixes bug introduced in 84f39d8124eb7f496992de3caf100a3f7baea069

At least I'm pretty confident it's a bug.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
@@ -67,7 +67,7 @@ class CheckExistence(
     override val icon = R.drawable.ic_quest_check
 
     override fun getTitle(tags: Map<String, String>): Int =
-        if (tags.containsAnyKey("name", "brand", "operator"))
+        if (tags.containsAnyKey("name", "ref", "brand", "operator"))
             R.string.quest_existence_name_title
         else
             R.string.quest_existence_title


app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
```diff
@@ -67,7 +67,7 @@ class CheckExistence(
     override val icon = R.drawable.ic_quest_check
 
     override fun getTitle(tags: Map<String, String>): Int =
-        if (tags.containsAnyKey("name", "brand", "operator"))
+        if (tags.containsAnyKey("name", "ref", "brand", "operator"))
```
peternewman, line 70: Is it worth making this an array stored in a constant, then passing the array here and passing the array through a map and popping the first one off in getTitleArgs, then they're always in sync (although perhaps slightly less performant)?

matkoniecz:
I have not tested it so far, but it looks like a bugfix.

Also, I would put ref after `brand` in https://github.com/streetcomplete/StreetComplete/commit/84f39d8124eb7f496992de3caf100a3f7baea069 (but that is a separate change)

See case of `shop=convenience brand=Orlen ref=82828` (while name is missing here brand would be more descriptive than `ref`).

peternewman:
> I have not tested it so far, but it looks like a bugfix.

Likewise me neither.

> Also, I would put ref after `brand` in [84f39d8](https://github.com/streetcomplete/StreetComplete/commit/84f39d8124eb7f496992de3caf100a3f7baea069) (but that is a separate change)
> 
> See case of `shop=convenience brand=Orlen ref=82828` (while name is missing here brand would be more descriptive than `ref`).

Yeah I'd probably agree with you. There will always be counter-cases, but unless they are very close together (e.g. bus stops) it's probably unlikely to matter.

westnordost:
Ok, I did that

PR #2809, opened 2021-04-26
Replace awaitPreDraw with awaitLayout (fixes #2801)

### LINKED ISSUE #2801: OSM link display sometimes broken(?)
The display of links in your profile was broken and did not display any links.
I found this by swiping through these screens a lot of times (all other ones work, e.g. the country ones or the most solved quests).
I could even close the menu and go back to it, so it was no temporary display bug.

The app is an old installation I just upgraded.

At another try after I locked my screen and when I was just about to reproduce this when writing this issue/for this issue, the links had suddenly appeared again…

**How to Reproduce**

Go to the menu “My profile” at the right panel.

It displayed “You have no links yet” or so.

**Versions affected**

v31.3 from F-Droid

tapetis:
Due to the way the `ViewPager2`'s `RecyclerView` works, the `onPreDraw` callback is not always called. This results in the links list not being displayed correctly. This PR fixes this behavior by instead using the `awaitLayout()` function.

Interestingly, the bug can also be fixed by setting the `ViewPager2`'s [OffscreenPageLimit](https://developer.android.com/reference/androidx/viewpager2/widget/ViewPager2#setOffscreenPageLimit(int)) to the value `1`. Since the bug seems to be related to the preloading of the fragments, I have also adjusted the corresponding place in the `AchievementsFragment`. Otherwise, the problem would probably also occur there if the order of the pages were to be changed.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/user/AchievementsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/user/AchievementsFragment.kt
@@ -15,7 +15,7 @@ import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.user.UserStore
 import de.westnordost.streetcomplete.data.user.achievements.Achievement
 import de.westnordost.streetcomplete.data.user.achievements.UserAchievementsSource
-import de.westnordost.streetcomplete.ktx.awaitPreDraw
+import de.westnordost.streetcomplete.ktx.awaitLayout
 import de.westnordost.streetcomplete.ktx.toPx
 import de.westnordost.streetcomplete.view.GridLayoutSpacingItemDecoration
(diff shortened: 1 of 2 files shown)


westnordost:
Wow, you are cool, @tapetis , thank you for that contribution! I remember spending almost half a day looking for the cause of this bug before

westnordost:
Did you test it though? (I'd like to avoid trying to reproduce a bug that is difficult to reproduce if you already did this)

tapetis:
Before changing the function call, I could reproduce the bug relatively often by opening the links tab very quickly after launching the app. With the `awaitLayout` function I was not able to reproduce this bug once.

westnordost:
Nice, ok!

PR #2811, opened 2021-04-26
Remove trunk from definite maxspeed type in GB

### LINKED ISSUE #2750: Maxspeed type on trunk roads and living street
`highway=trunk` is included in `ROADS_WITH_DEFINITE_SPEED_LIMIT`, meaning that when the maxspeed quest is answered with "No sign, defaults apply" (implicit speed limit) `maxspeed:type` is set to `GB:trunk` (or whichever country code).

This is not an established tag, and a spot check of some ways (in a few different countries) from taginfo/overpass makes it look like most have been added by SC users.

I don't know enough about other countries to know whether this should just be removed from the `ROADS_WITH_DEFINITE_SPEED_LIMIT` or whether other checks are needed.

In the UK at least it could be removed and the user should be asked if it's dual or single carriageway.

I think that the same can be said for `living_street`, there are tags for that, but at the moment it isn't `GB:living_street` (taginfo shows no uses of that, but a few uses for other countries).

(Turns out that speed limits are quite a can of worms!)

arrival-spring:
Tested and working as expected for me.

Fixes #2750

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/max_speed/AddMaxSpeedForm.kt
@@ -215,10 +215,10 @@ class AddMaxSpeedForm : AbstractQuestFormAnswerFragment<MaxSpeedAnswer>() {
 
     private fun determineImplicitMaxspeedType() {
         val highwayTag = osmElement!!.tags["highway"]!!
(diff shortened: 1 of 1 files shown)


westnordost:
Mh ok, I was hoping that the code be like "if GB do special things else do normal things" or something. But if you don't want to change that, that's fine too.

arrival-spring:
I think I was thinking about making minimal changes, this seems cleaner now (and again, tested and working for me).

westnordost:
Nice, thank you!

PR #2814, opened 2021-04-27
Improve check existence quest wording for named things

### LINKED ISSUE #2806: Improve check existence quest wording
**Use case**
Current wording:
https://github.com/streetcomplete/StreetComplete/blob/79d78db7c905d87a9a80260d8bdf2be53a34daa1/app/src/main/res/values/strings.xml#L449-L450

The quest will ask using arguments (name, featureName.value), where name will be the first of ("name", "ref", "brand", "operator"). So if you imagine Google started installing benches everywhere, and they were tagged with some combination of:
```
amenity=bench
name=Sergey's Seat
ref=600913
brand=Google
operator=Google RESTful
```

Giving these possible results:
```
Is this still here? (bench)
Is Sergey's Seat still here? (bench)
Is 600913 still here? (bench)
Is Google still here? (bench)
Is Google RESTful still here? (bench)
```

**Proposed Solution**
IMHO, these work for big brands where the name/brand thing is what they're known for, e.g. McDonald’s, but doesn't where it's not the main thing they are known for as in these examples, or e.g. a BT phonebox or Royal Mail postbox.

I was going to just do a PR, but given there are a few options, some other input might work (and maybe some consideration of how they translate).

I think one of these options would be clearer:
For the unnamed one:
```
Is this %s still here?
Is this bench still here?
```

Or if you really want to keep the brackets:
```
Is this (%s) still here?
Is this (bench) still here?
```
For the named ones:
```
Is %1$s %2$s still here?
Is Sergey's Seat bench still here?
Is 600913 bench still here?
Is Google bench still here?
Is Google RESTful bench still here?
```

Or (slightly longer):
```
Is this %1$s %2$s still here?
Is this Sergey's Seat bench still here?
Is this 600913 bench still here?
Is this Google bench still here?
Is this Google RESTful bench still here?
```

Alternative (I think this is my preferred choice):
```
Is this %2$s (%1$s) still here?
Is this bench (Sergey's Seat) still here?
Is this bench (600913) still here?
Is this bench (Google) still here?
Is this bench (Google RESTful) still here?
```
I'm assuming the substitution would let you just do this and swap the order in the translation?

peternewman:
Closes #2806

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -447,7 +447,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_dietType_explanation">Only answer “yes” if the place provides a comparable choice of eating options for this diet</string>
     <string name="quest_toiletAvailability_name_title">Does %s have a restroom?</string>
     <string name="quest_existence_title">Is this still here? (%s)</string>
-    <string name="quest_existence_name_title">Is %1$s still here? (%2$s)</string>
+    <string name="quest_existence_name_title">Is %1$s (%2$s) still here?</string>
     <string name="quest_powerPolesMaterial_title">What is the material of this power pole?</string>
     <string name="quest_powerPolesMaterial_wood">Wood</string>
     <string name="quest_powerPolesMaterial_metal">Steel</string>


westnordost:
And corrected all the translations.

PR #2770, opened 2021-04-19
Fix credits

arrival-spring:
I don't want to take away from the contributions that @peternewman has made, but he probably only needs to be listed in the credits once.

PR #2815, opened 2021-04-27
Add the git stats task(s)

peternewman:
Progress on discussions triggered in #2770

Potential todo's:

- [ ] Add other filters for strings/graphics/photos etc (the existing used categories are credits_art.yml  credits_code.yml  credits_main.yml  credits_projects.yml  credits_translations.yml)
- [ ] Work out what other totals are needed for the other systems (e.g. do we score lines for an SVG, or just one "point" for each SVG change commit)
- [ ] Possibly add a list of suspect words for the start or within a commit message and then an allow/deny list when they've been checked off if they appear.
- [ ] Use https://api.github.com/repos/streetcomplete/StreetComplete/commits/84ad625510223900a104b399807f4d26eea3daa5 to find people's usernames (for potential links to GitHub) and avoid the splitting of e.g. PanierAvide/Adrien Pavie
- [ ] Above possibly combined with https://api.github.com/repos/streetcomplete/StreetComplete/contributors or similar to avoid looking up rugk when it's a match to a contributor login already
- [ ] Add some sort of check that the full history has been checked out

See some sample output here (or just run the task):
https://github.com/peternewman/StreetComplete/runs/2451053449?check_suite_focus=true


peternewman:
File type stats (from https://unix.stackexchange.com/questions/18506/recursive-statistics-on-file-types-in-directory):
`find . -type f -printf '%f\n' | sed -r -n 's/.+(\..*)$/\1/p' | sort | uniq -c | sort -n -r`
```
   1177 .xml
    813 .kt
    582 .jpg
    475 .yml
    402 .svg
    204 .png
     97 .txt
     41 .json
     10 .yaml
     10 .java
      9 .sample
      7 .kts
      7 .bin
      6 .wav
      6 .md
      5 .properties
      3 .xcf
      3 .pack
      3 .lock
      3 .idx
      2 .jar
      1 .ser
      1 .pro
      1 .MockMaker
      1 .MF
      1 .drawio
      1 .csv
      1 .bat
```

So jpg/png covers photos.

buildSrc/src/main/java/GitStatsTask.kt
```diff
@@ -0,0 +1,77 @@
...
+    @TaskAction fun run() {
+        val countsByName = mutableMapOf<String, Int>()
+        val countsByCommit = mutableMapOf<String, Int>()
+        Runtime.getRuntime().exec("git log --no-merges --pretty='%an'%n%H --numstat").inputStream.bufferedReader().useLines { lines ->
+            var name = ""
+            var commit = ""
+            var commitNext = false
+            var skipNext = false
+            for(line in lines) {
+                if (line.startsWith('\'')) {
```
westnordost, line 28: I don't get it, what lines start with `'`? (Maybe add a comment)

build.gradle.kts
```diff
@@ -66,3 +66,26 @@ tasks.register("updateStreetCompleteData") {
...
+    group = "streetcomplete"
+    commitFileFilter = Regex(".*\\.(java|kt|kts)$")
+    commitSkipList = arrayOf("ae7a244dd60ccfc91cf2dc01bf9e60c8d6a81616")
+    targetFile = "$projectDir/app/src/main/res/raw/commits_code.yml"
+}
+
+tasks.register<GitStatsTask>("gitStatsPhotos") {
+    group = "streetcomplete"
+    commitFileFilter = Regex(".*\\.(jpe?g|png)$")
+    binaryScore = 1  // This is actually doubled due to the maths
```
westnordost, line 80: would be better if it is not doubled, then you also don't need a comment explaining this.

buildSrc/src/main/java/GitStatsTask.kt
```diff
@@ -0,0 +1,77 @@
...
+                    if (!splits.last().matches(commitFileFilter)) continue
+                    val additions = splits[0].toIntOrNull() ?: binaryScore
+                    val deletions = splits[1].toIntOrNull() ?: binaryScore
+                    val commitTotal = additions + deletions
+                    if (!skipNext) {
+                        countsByName[name] = commitTotal + countsByName.getOrPut(name, { 0 })
+                    }
+                    if (commitTotal > 400) {
+                        countsByCommit[commit] = commitTotal
+                    }
```
westnordost, line 50: What is this for?

westnordost:
Hmm interesting, I didn't know that you can create a gradle task, but hten also a github workflow to execute that gradle task. Thing is, the plans would be to include these statistics as a resource to be read in the app, so that is something I would update on each update together with translations, metadata etc.

Code looks very clean, also cool that you parametrized the tasks!

Some more general comments:

### Count everything

I was thinking that maybe after all, **all** file changes except those that are automatically generated or imported from another source should be counted. After all, fiddling around with an Android layout resource or creating SVG graphics etc is also work and should not go by the board. At least, we should do a comparison - count only code and count all files except for those exceptions and see what changed.  
Files I can think of right now which should not be counted:
    - every `strings.xml` - is imported from POEditor. The "default" one is not, but that's just a small detail
    - conveniently, currently pretty much all files in `/app/src/main/assets`
    - probably, all files from `/app/src/main/res/drawable*`. Some may be proper contributions, but most are actually either imported SVGs from `res/graphics`, photos, imported from another repos (the flags) or imported from the Android icon repository
    - everything in `/metadata`

### Count other repositories

I don't know about you, but I have all the StreetComplete-related repositories in one place, so I can from StreetComplete `cd ../streetcomplete-mapstyle` for example. As contributing to these repositories should also count, the gradle task should be able to run through the other locally checked out repositories too, or fail/log if they are not available.
And this I guess would also mean that having the script executed as a github workflow would also be out of question.

riQQ:
> And this I guess would also mean that having the script executed as a github workflow would also be out of question.

You can check out as many repositories as you want in a GitHub actions workflow. Just use multiples checkout steps and, if necessary, specify a path for each checkout.

build.gradle.kts
peternewman: Yeah, this was a bit of a quick hack, it was to deal with the `- - filename` binary entries. Need to ponder a little on a neat way to fix it.

buildSrc/src/main/java/GitStatsTask.kt
peternewman: https://github.com/streetcomplete/StreetComplete/blob/1844ef7a17e3ed7c411316f918f1d2d0d304544b/buildSrc/src/main/java/GitStatsTask.kt#L22
The author line does. This was your code, I've fudged it a bit since. I suspect there is probably a better way to handle it like maybe separating with nulls or some sort of FSM per line or something.

buildSrc/src/main/java/GitStatsTask.kt
peternewman: Just for stats, to list any commits with lots of changes to try and find more of those suspect ones. Could be removed in due course once we're happy nothing suspect is getting through.

peternewman:
> Hmm interesting, I didn't know that you can create a gradle task, but then also a github workflow to execute that gradle task.

Yep, the GitHub workflows are just VMs briefly spun up. You can do anything you want on them including installing software. Even including mining bitcoing:
https://www.bleepingcomputer.com/news/security/github-actions-being-actively-abused-to-mine-cryptocurrency-on-github-servers/

> Thing is, the plans would be to include these statistics as a resource to be read in the app, so that is something I would update on each update together with translations, metadata etc.

Yeah I'm only running them in here because my Gradle is too old to play properly.

> Code looks very clean, also cool that you parametrized the tasks!

Thanks, just mostly copied the base code, based on yours, and some tasks, again mostly based on yours.

> ### Count everything
> I was thinking that maybe after all, **all** file changes except those that are automatically generated or imported from another source should be counted. After all, fiddling around with an Android layout resource or creating SVG graphics etc is also work and should not go by the board. At least, we should do a comparison - count only code and count all files except for those exceptions and see what changed.

I'd respectfully disagree, or suggest it's more complicated than that, e.g. if you look at this commit:
https://github.com/streetcomplete/StreetComplete/commit/6e9dafae9d1714da3309dd91f735ffdaed2cb8ab
348 changed files with 49,188 additions and 27,403 deletions.

Or an icon like archery:
https://github.com/streetcomplete/StreetComplete/blob/6e9dafae9d1714da3309dd91f735ffdaed2cb8ab/res/sport%20icons/archery.svg
275 lines

That one commit is more than 10 times Flo's total lines of actual code changes. The archery logo alone is more lines of changes than all but these people committed as code changes:
```
311569	Tobias Zwick
7896	Mateusz Konieczny
6930	ENT8R
4585	Flo Edelmann
1116	PanierAvide
918	Mark Tully
788	David Dean
778	Isira Seneviratne
738	Krzysztof Krason
718	Adrien Pavie
413	L.Lemke
315	arrival-spring
```

I'd agree that "fiddling around with an Android layout resource or creating SVG graphics etc is also work and should not go by the board", but hence my stats comment. IMHO the best ways are either to credit each of those edits with some arbitrary "score" or better have them separately. If you wanted to do your bouncing balls it could show people, and then when you tapped a person show their code/graphics/photos split.

> Files I can think of right now which should not be counted:
> 
> * every `strings.xml` - is imported from POEditor. The "default" one is not, but that's just a small detail

Fair enough, but probably just as easy to ignore all BUT that one.

> * conveniently, currently pretty much all files in `/app/src/main/assets`
> * probably, all files from `/app/src/main/res/drawable*`. Some may be proper contributions, but most are actually either imported SVGs from `res/graphics`, photos, imported from another repos (the flags) or imported from the Android icon repository

Ah cool, I wasn't quite sure where the originals of those lived.

> * everything in `/metadata`

> ### Count other repositories
> I don't know about you, but I have all the StreetComplete-related repositories in one place, so I can from StreetComplete `cd ../streetcomplete-mapstyle` for example.

I've only got one actually checked out, but I'll often have checkouts by branch name or similar, so streetcomplete-gitstats. I've suggest relying on a particular folder pattern is even more brittle than the other stuff we're doing in here.

> As contributing to these repositories should also count, the gradle task should be able to run through the other locally checked out repositories too, or fail/log if they are not available.

Yes, potentially, but see below.

> And this I guess would also mean that having the script executed as a github workflow would also be out of question.

As @riQQ notes, this is more than possible. I'd actually suggest the opposite, it means it's the way to go, it could auto-generate all the stats either periodically or on-demand (sadly it can't do it on release for SC itself, but I guess it could for the other repos). It can even commit the updated stats back into git for you!

I don't know the specifics of your release workflow, but if you wanted you could get GitHub to build the APK or whatever when you generate a release as well as updating all the other stats and committing them if required. You might even be able to deploy to the app stores from there!

westnordost:
> I'd respectfully disagree, or suggest it's more complicated than that, e.g. if you look at this commit:
> 6e9dafa
> 348 changed files with 49,188 additions and 27,403 deletions.

Well, this was still quite some work. But of course not as much work as doing them all anew. I think this is one of those commits that should be excluded.

peternewman:
> > I'd respectfully disagree, or suggest it's more complicated than that, e.g. if you look at this commit:
> > [6e9dafa](https://github.com/streetcomplete/StreetComplete/commit/6e9dafae9d1714da3309dd91f735ffdaed2cb8ab)
> > 348 changed files with 49,188 additions and 27,403 deletions.
> 
> Well, this was still quite some work. But of course not as much work as doing them all anew. I think this is one of those commits that should be excluded.

Yeah, I wasn't entirely meaning that commit itself, but we can certainly exclude it. More about the weighting of SVGs compared to photos compared to normal code.

E.g.
> Or an icon like archery:
> https://github.com/streetcomplete/StreetComplete/blob/6e9dafae9d1714da3309dd91f735ffdaed2cb8ab/res/sport%20icons/archery.svg
> 275 lines

Is that one icon really comparable to all of arrival-spring's code commits for example (an entire quest and tweaks to some others)?

It seems we could either score it directly, have a scaling factor for things like SVGs and GUI elements, then we need some points for photos and any binary blobs, where we have no value at all (unless we measure the size of it).

westnordost:
Hmm, maybe have a fixed score for SVGs then

westnordost:
As I am right now working on a branch where I had to revert much of what has been done, maybe it would be good, if possible, to only count the changes that were actually merged to master. Example:

Person A changes in a branch
```diff
+ abc
- def

+ foo
- bar
```

but then later in another commit on that branch, Person B does
```diff
+ def
- abc

+ lol
```

Then, Person A should get one change attributed (+foo -bar ) and Person B should get also one change attributed (+ lol) but the abc-def change should not count.

westnordost:
I will tae over here

westnordost:
see #3850

PR #2820, opened 2021-04-28
Fix keyboard not appearing in TimeRangePickerDialog (fixes #2799)

### LINKED ISSUE #2799: Keyboard doesn't appear when using the time input fields in the opening hours quest
Hi there,

@TurnrDev's comment (https://github.com/streetcomplete/StreetComplete/issues/2795#issuecomment-826063895) inspired me to try it out myself and found out that this issue doesn't occur with each time picker:

**How to Reproduce**

- Find an **opening hours** quest and open the associated time picker.
- Instead of choosing the time from the time picker, tap on the keyboard icon. A form with two fields (hour/minute) appears.
- Tap into one of the two fields. You would expect the (numeric) keyboard to appear, but nothing happens. It's impossible to enter a time here, you have to go back to the time picker by tapping on the clock icon.

However, this bug does **not** appear with the time picker of the **post box collecting times** quest. Here the keyboard appears as expected.

**Versions affected**

Tested with two devices with Android 8.1 and 9.0 repectively. SC v31.3 on each.

The cycling cat

tapetis:
As explained in the `AlertDialog`'s [Javadoc](https://developer.android.com/reference/android/app/AlertDialog), the class normally automatically takes care of setting the `WindowManager.LayoutParams.FLAG_ALT_FOCUSABLE_IM` window flag by checking if the dialog's view contains an `EditText`. However, this does not seem to work with the `TimePicker`, which displays a clock dial by default. This PR fixes this problem by setting the correct window flags by hand. On my device, the keyboard can now be opened by clicking in the text fields.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/TimeRangePickerDialog.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/TimeRangePickerDialog.kt
@@ -2,6 +2,7 @@ package de.westnordost.streetcomplete.quests.opening_hours
 
 import android.content.Context
 import android.content.DialogInterface
+import android.os.Bundle
 import com.google.android.material.tabs.TabLayout
 import androidx.viewpager.widget.PagerAdapter
 import androidx.viewpager.widget.ViewPager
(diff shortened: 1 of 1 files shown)


matkoniecz:
I have never noticed that part of the interface :)

Tested, confirmed to be fixed, thanks for finding workaround! I would need like a month of work to discover that.

westnordost:
Wow, really cool that you found that!

PR #2821, opened 2021-04-28
Wrap some osmapi exceptions in Kotlin exceptions

### LINKED ISSUE #2818: [v32] Define/wrap exceptions thrown by MapDataApi and NotesApi in Kotlin
- [ ] The exceptions the methods in NotesApi throw must also be defined in Kotlin in StreetComplete. It is ok to simplify/merge them if they are each handled the same by the code that uses the api.
- [ ] The exceptions the methods in MapDataApi throw must also be defined in Kotlin in StreetComplete. It is ok to simplify/merge them if they are each handled the same by the code that uses the api.

FloEdelmann:
Closes #2818.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
@@ -23,7 +23,6 @@ import androidx.localbroadcastmanager.content.LocalBroadcastManager
 import androidx.preference.PreferenceManager
 import de.westnordost.osmapi.common.errors.OsmApiException
 import de.westnordost.osmapi.common.errors.OsmApiReadResponseException
-import de.westnordost.osmapi.common.errors.OsmAuthorizationException
 import de.westnordost.osmapi.common.errors.OsmConnectionException
 import de.westnordost.streetcomplete.Injector.applicationComponent
 import de.westnordost.streetcomplete.controls.NotificationButtonFragment
@@ -37,6 +36,7 @@ import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.data.upload.UploadController
 import de.westnordost.streetcomplete.data.upload.UploadProgressListener
 import de.westnordost.streetcomplete.data.upload.VersionBannedException
+import de.westnordost.streetcomplete.data.user.AuthorizationException
 import de.westnordost.streetcomplete.data.user.UserController
 import de.westnordost.streetcomplete.ktx.toast
 import de.westnordost.streetcomplete.location.LocationRequestFragment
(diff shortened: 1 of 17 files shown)


app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
```diff
@@ -23,7 +23,6 @@ import androidx.localbroadcastmanager.content.LocalBroadcastManager
 import androidx.preference.PreferenceManager
 import de.westnordost.osmapi.common.errors.OsmApiException
 import de.westnordost.osmapi.common.errors.OsmApiReadResponseException
-import de.westnordost.osmapi.common.errors.OsmAuthorizationException
 import de.westnordost.osmapi.common.errors.OsmConnectionException
```
FloEdelmann, line 26: Here

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/changesets/ChangesetAutoCloserWorker.kt
```diff
@@ -6,9 +6,9 @@ import javax.inject.Inject
 
 import androidx.work.Worker
 import androidx.work.WorkerParameters
-import de.westnordost.osmapi.common.errors.OsmAuthorizationException
 import de.westnordost.osmapi.common.errors.OsmConnectionException
```
FloEdelmann, line 9: Here

FloEdelmann (commented):
What should I do about these few missing ones? (`OsmApiException`, `OsmApiReadResponseException` and `OsmConnectionException`)

app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
westnordost: These exceptions are not documented on the `MapDataApi` and `NotesApi` from the de.westnordost.osmapi package because they are like "unchecked exceptions", see line 256ff

I'd say, wrap em all in "ConnectionException" or so. These kind of errors should just lead to a toast message displayed to the user that upload/download is currently not possible.
Since these errors could be thrown on any API call (f.e. if the OpenStreetMap API is down for maintenance, connection is lost due to bad network etc), I fear these need to be caught on every single API function. Maybe write an inline helper function like `inline fun tryConnect(block: () -> T): T` so to not bloat up the code there.
Maybe it also makes sense to not catch the individual erros but have an `catch(e: Exception)` at the end?

westnordost (changes requested):

westnordost:
Two thoughts:
- catching "all" is not viable because f.e. `IllegalArgumentException` should not be caught
- Actually, this proposed `tryConnect` function could maybe even do all/most of the wrapping, f.e. for the authorization exception, so then this will be very little boilerplate code.

westnordost:
Looks good!

westnordost (approved):

westnordost:
I'll merge it right away. And then, with that done, I think I can also release a public alpha version to get some more error reports before launching it as beta in the play store.

Issue #2640, opened 2021-03-06
"Does this still exist? ($BRAND)" is too unspecific


schuetzm:
**How to Reproduce**
Not sure how to reproduce. I was walking around with my mobile phone which has no mobile internet access, so the quests I solved were not being uploaded immediately; now that they've been uploaded at home, the quest has disappeared.

After entering the collection times for [this postbox](https://www.openstreetmap.org/node/366339764), a new quest appeared at the same position asking "Does this still exist? (Deutsche Post)". So, the first problem is that it's not clear what the question is referring to; it would be nice if it asked "Does this *postbox* still exist?" or something similar.

As right next to it there is a phone, I was not sure whether the question was actually about the postbox, or if it was about the phone which someone might have accidentally mistagged. I've now checked this, and the phone is correctly tagged with "Deutsche Telekom AG" as operator, so I assume the quest was indeed referring to the postbox. This then leads to the second bug/problem: As I've just entered the collection times of this postbox, it's pointless to ask whether it still exists...

**Versions affected**
StreetComplete v30.1, German language

TurnrDev:
When online, I've always had answering the times make it skip the "does this still exist?" question. I wonder if this is related to being offline?

smichel17:
Probably yes. https://github.com/streetcomplete/StreetComplete/issues/2506 may change that.

westnordost:
No, it isn't related to being online or offline.

Anyway, @schuetzm , this has been fixed by @FloEdelmann 16 days ago and the fix is part of the v31.0 release that I released a couple of days ago 😄

westnordost:
If you got the app from F-Droid, it may take up to a few days to months till it is updated.

schuetzm:
@westnordost Yes, I'm used to the delays by now :-) Just to be sure, the first issue was fixed in [this PR](https://github.com/streetcomplete/StreetComplete/pull/2594), right? The second (minor) problem of the existance being asked is probably not fixable easily, because quests cannot take previous answers into account before the data is re-downloaded...

Helium314:
Not sure, but it might be fixed as part of #2506

FloEdelmann:
In #2594, it was fixed that the existence quest didn't show the feature name, but only the element name. I.e. it should be now "Does this postbox still exist? (Deutsche Post)" instead of "Does this still exist? (Deutsche Post)".

[...]

PR #2824, opened 2021-04-29
Add test that checks CheckExistence quest title

FloEdelmann:
There were some issues in the past regarding the title of the CheckExistence quest:

* https://github.com/streetcomplete/StreetComplete/issues/2512
* https://github.com/streetcomplete/StreetComplete/issues/2640
* https://github.com/streetcomplete/StreetComplete/issues/2806

I've added regression test cases for those, so that future updates won't break those improvements.

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/existence/CheckExistenceTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/existence/CheckExistenceTest.kt
@@ -1,11 +1,13 @@
 package de.westnordost.streetcomplete.quests.existence
 
+import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.meta.toCheckDateString
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapEntryAdd
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapEntryDelete
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapEntryModify
-import de.westnordost.streetcomplete.testutils.mock
 import de.westnordost.streetcomplete.quests.verifyAnswer
+import de.westnordost.streetcomplete.testutils.mock
+import org.junit.Assert.*
 import org.junit.Test
 import java.time.LocalDate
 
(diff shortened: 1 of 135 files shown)


app/src/test/java/de/westnordost/streetcomplete/quests/existence/CheckExistenceTest.kt
```diff
@@ -36,4 +38,61 @@ class CheckExistenceTest {
...
         )
     }
+
+    @Test fun `title is constructed correctly, issue #2512`() {
+        val newspaperVendingMachineWithName = questType.getQuestTitle(mapOf(
+            "amenity" to "vending_machine",
+            "vending" to "newspapers",
+            "name" to "Bild",
+        ))
+        assertEquals(newspaperVendingMachineWithName, "Is Bild (vending machine) still here?")
```
westnordost, line 48: How can this work? This is a `test`, not an `androidTest` right? So how is it possible that `de.westnordost.streetcomplete.R` is available?

If this works... then I don't know how, and I think it would be better to move these tests to `androidTests`

app/src/test/java/de/westnordost/streetcomplete/quests/existence/CheckExistenceTest.kt
```diff
@@ -36,4 +38,61 @@ class CheckExistenceTest {
...
+        R.string.quest_existence_title -> "Is this still here? (%s)"
+        else -> throw IllegalArgumentException("Unknown title resource")
+    }
+
+    // simplified feature name, instead of looking up iD preset
+    val featureName = tags["amenity"]?.replace('_', ' ')
+
+    val titleArgs = getTitleArgs(tags, lazy { featureName })
+
+    return titleTemplate.format(*titleArgs)
```
westnordost, line 97: Oh I see, you reimplemented that function for the test. Well, I still think it would be better to move the whole test to `androidTests`, then you don't need this function too.

westnordost:
More tests, always a good idea, thank you @FloEdelmann !

westnordost (approved):

FloEdelmann:
TODO: Add a test case here for issue 4 in #2840, regardless whether/how the behavior is changed.

westnordost:
Thanks for the hint! It shouldn't be up to you to adapt a PR that is already approved. So I'll merge your PR now and see if any tests need to be adapted.

Issue #2158, opened 2020-10-12
Bug in Themes


tintinmaster:
When the System Theme differs from the selected them in the App (also in Auto mode) some elements have the theme background, but the wrong text color. So it's hard/impossible to read.
Using theme "System Default" works flawless.

It does not happen while changing the theme and a restart does not help, so it is not directly related to the other issues (which are all closed by now) 

It seems like some backgrounds are always determined by the system theme and not by the theme the app enforces.

**How to Reproduce**
Have System _Dark Mode_ **activated** and in-app _Light theme_ **activated** or
Have System _Dark Mode_ **deactivated** and in-app _Dark theme_ **activated**

**Versions affected**
v24.1 and Android 10
Black System and Light In-App:
![image](https://user-images.githubusercontent.com/36089973/95775815-648c3b80-0cc3-11eb-9264-a6592a4f91b0.png)
![image](https://user-images.githubusercontent.com/36089973/95775820-6950ef80-0cc3-11eb-8fca-a391d45619ab.png)

Light System and Dark In-App:
![image](https://user-images.githubusercontent.com/36089973/95775828-6c4be000-0cc3-11eb-8ff6-ce457a4428df.png)

westnordost:
I don't have a new phone so I do not have dark mode to reproduce it. Help would be appreciated.

ENT8R:
I'm also on Android 10 and just tried whether I could reproduce any similar behaviour, but this is the only thing I noticed which comes close to the issue described above:
<div align="center">
<img src="https://user-images.githubusercontent.com/25306497/95890052-8c42d880-0d83-11eb-8544-53f4628b6dbf.jpg" width="250px"> <img src="https://user-images.githubusercontent.com/25306497/95890048-8a791500-0d83-11eb-90c8-5fd5fea4e7b4.jpg" width="250px">
</div>
Seems weird to me that it differs that much across different devices, maybe it's the vendor's fault because they screwed something up in their dark mode implementation?

matkoniecz:
> I don't have a new phone so I do not have dark mode to reproduce it. Help would be appreciated.

I just (accidentally and with no clear info how to repeat, but may be related to opening phone after it was closed during night) reproduced it in Android 10 emulator

![screen](https://user-images.githubusercontent.com/899988/95956318-660e4e80-0dfe-11eb-9eee-209d929e149a.jpg)

tintinmaster:
[...]

Issue #1959, opened 2020-07-16
Pointer pin sometimes cut off at bottom


Helium314:
Sometimes the pin pointing to my position is cut of when it is at the bottom of the screen.
![pointer_pin](https://user-images.githubusercontent.com/43007630/87634441-8cf4b000-c73d-11ea-9624-08998de37651.png)

This seems to happen when the pin is moving from the left or right edges to the bottom, but not when the pin appears at the bottom first.
I was able to always reproduce it when my initial position (after start or switch to app) was off the visible map, and the pin was initially shown at left, right or top. Then I move/rotate the map so the pin moves to the bottom (without my position ever being on the visible map), and the pin is always cut off.

StreetComplete 21.0 and 21.1 (fdroid)
Android 9 (LineageOS based)

westnordost:
Can't reproduce it, Android 10.

Though, even if I could, quite unlikely I'd fix it to be honest because it is such a minor issue. Though, others are contributors and new contributors who can reproduce it are invited to investigate it. Probably the issue is related to 'padding', ' margin' and 'WindowManager.LayoutParams.FLAG_TRANSLUCENT_NAVIGATION'.

Sequynth:
Reproduced on Android 6.0 running 21.1

cyclingcat:
Can reproduce on Android 5.1 running StreetComplete v21.1.
Can **_not_** reproduce on Android 9.0 running StreetComplete v20.2.

I am wondering if this has something to do with "real" vs. "virtual" navigation keys at the bottom of the phone display - my old Android 5.1 phone has "real" touch keys which have their own place at the bottom and thus are never part of the displayed map while my newer Android 9.0 phone shows the navigation keys as a layer on top of the map.

The cycling cat

Helium314:
There could be a connection... my phone also has "real" buttons.
But when I disable the buttons and enable the navigation bar the pin is still cut off.

westnordost:
What do you mean with "real" buttons by the way? Hardware buttons?

cyclingcat:
Don't know if "hardware keys" is the correct term, but this photo illustrates the difference:

![keys_on_smartphones](https://user-images.githubusercontent.com/32601012/87848563-cd8d2e80-c8e1-11ea-86e9-d6d90a01731b.jpg)

[...]

Issue #1236, opened 2018-10-19
TimePicker width is cut off in landscape mode


lzmartinico:
When the app is in landscape mode, the time picker width is cut off, making it impossible to select a time (similarly to [this](https://stackoverflow.com/questions/35494234/android-timepicker-not-displayed-well-on-landscape-mode) Stackoverflow question)
**How to Reproduce**
Add steps to reproduce this behavior
Turn your phone in Landscape mode, select an Opening Hours quest, Select some weekday, the watch is not displayed correctly in either _opening_ or _closing time_ tabs
**Versions affected**
Android 7.1.1, Street Complete 8.2 and previous

westnordost:
This is an Android bug (imo). I already noticed this months ago and tried to workaround this, also consulting stackoverflow, but did not succeed.

So, it is a known problem but can not be fixed.

lzmartinico:
Yes, it's an Android API bug, but it depends on how you use the TimePicker. The default AOSP calendar, for example, does not get cut off in landscape mode. I know they are not using a tabbed view, but maybe this could be useful? https://android.googlesource.com/platform/packages/apps/Calendar/+/master/src/com/android/calendar/event/EditEventView.java

HolgerJeromin:
screenshot for reference:
![rps20181021_005119_990](https://user-images.githubusercontent.com/2410353/47261266-9e0e7f80-d4cb-11e8-8cee-806d149e3b04.jpg)

smichel17:
I checked if I could reproduce this on a Galaxy S6 running Android 7.0 and found a different but similar issue:

![image](https://user-images.githubusercontent.com/6563664/75038964-9840f980-5485-11ea-9277-88ac1bf7b726.png)

FWIW I ran into the same issue on Red Moon and decided not to fix it. The bug occurs in standard usage and is therefore an Android bug that should be fixed upstream, in my opinion.

If someone wanted to fix it downstream, I think it would be better to write a TimePickerDialog library that works in landscape mode, rather than fixing it in each individual app.

matkoniecz:
add phrase to enable finding this issue: opening hours interface

westnordost:
Would be fixed by #3796

westnordost:
Well I now made that particular view scroll. Not great, but at least it is a mitigation.

PR #2831, opened 2021-05-01
Reset flag offset after onMeasure

### LINKED ISSUE #2819: [already in v31] at least some flags are overly stretched horizontally
screen

**How to Reproduce**
Edit in Israel/Cyprus, login, view flag screen
**Versions affected**
v31, v32

tapetis:
This PR fixes #2819 by resetting the calculated flag bounds offset when the width of the view changes.
Before:
device-2021-05-01-131902
After:
device-2021-05-01-131939

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/user/CircularFlagView.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/user/CircularFlagView.kt
@@ -51,6 +51,7 @@ class CircularFlagView @JvmOverloads constructor(
             if (widthMode == MeasureSpec.EXACTLY) width else height
         } else min(width, height)
         setMeasuredDimension(size, size)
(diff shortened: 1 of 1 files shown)


matkoniecz:
I can confirm that it works! (tested in emulator)

It seems that it could be also placed `onSizeChanged` (I my reading of code and poking at it is correct).

BTW, if for some reason you are interested in working your magic on bizarre bugs - #2158 #1236 #1959 are kind of similar. I would be really thankful if you managed to make any progress on them, I am really completely confused by them.

tapetis:
Yes, you are right. I did not think about the `onSizeChanged` function.

When I have time, I may look at the other bugs you listed.

westnordost:
Great work @tapetis , thank you!

PR #2843, opened 2021-05-04
Clarify that quest selection screen allows reordering

### LINKED ISSUE #2828: Add a re-order handle icon to the quests list to hint that it is user sortable
**Use case**
SC allows you to re-order quests, but this is not obvious at all. I missed this until it was pointed out here and so did someone else https://github.com/streetcomplete/StreetComplete/issues/2802#issuecomment-829609238.

Re-ordering quests makes mappers happier and more likely to continue doing quests, allowing mappers to have easier to answer quests appear first, or quests that they are more interested in appear first.

**Proposed Solution**
Add the standard re-order handle to the left of the quests list, which makes it obvious to users that quests can be re-ordered.

reorder-horizontal

FloEdelmann:
Closes #2828.

* Rename screen from "Quest selection" to "Quest selection and priority". I avoided the word "order", as "priority" is already used in that screen and I think it's better to stick with one word for one concept.

  

* Add a drag handle at the left of each quest selection row. Dragging there is immediately possible (without long-press), dragging anywhere by long-pressing is still allowed.

  

I used the [drag-vertical Material Design Icon](https://materialdesignicons.com/icon/drag-vertical), should I add that to `authors.txt`?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/questselection/QuestSelectionAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/questselection/QuestSelectionAdapter.kt
@@ -5,6 +5,7 @@ import androidx.appcompat.app.AlertDialog
 import androidx.recyclerview.widget.RecyclerView
 import androidx.recyclerview.widget.ItemTouchHelper
 import android.view.LayoutInflater
+import android.view.MotionEvent
 import android.view.View
 import android.view.ViewGroup
 import android.widget.CheckBox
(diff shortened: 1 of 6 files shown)


westnordost:
Nice, also looks more slim than I thought! Though, regarding the size, according to the accessibility guidelines, the minimum size for a tap area should be about 40dp (ideally 48dp), see

https://material.io/design/usability/accessibility.html#layout-and-typography

If this is becoming too wide, one possibility to have it visually still small, but the invisible click area overlaps with the icon.

westnordost:
>  avoided the word "order", as "priority" is already used in that screen and I think it's better to stick with one word for one concept.

👍

FloEdelmann:
I've made the touch area bigger, so it's now 48dp wide (and still fills the available height):

grafik

westnordost:
Nice, thanks!

matkoniecz:
> I used the drag-vertical Material Design Icon, should I add that to authors.txt?

It may be overkill, at that point such work is no longer eligible for copyright anyway. But adding would be still OK.

PR #2844, opened 2021-05-04
ask less often about presence of some objects

matkoniecz:
For this object existence quest is an annoying spam, at least for my location.

I would expect this to change relatively rarely and be present in huge numbers, so I think that asking rarer is justified.

Posting as PR for wider review.


westnordost:
that's reasonable

westnordost (approved):

PR #2847, opened 2021-05-04
Add "X of Y enabled" subtitle to quest list

### LINKED ISSUE #2243: Making disabled quests more discoverable
**Use case**
Users are often unaware that StreetComplete comes with some quests disabled by default. In some cases people are actually interested in quests that were disabled.

As number of quests grows (what is great!) it gets harder and more time consuming to hunt for ones that come disabled.

**Proposed Solution**
Add new "X disabled quest types" entry in settings, below "quest selection", with X being number of disable quests. Opening this would show list of quests similar to one in "Quest selection" but without ability to reorder them and only disabled ones would be listed.

BTW, maybe for "quest selection" show "reorder, enable and disable quests" subtitle?

FloEdelmann:
Closes #2243, follow-up to #2843.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -16,6 +16,7 @@ import androidx.preference.PreferenceManager
 import de.westnordost.streetcomplete.ApplicationConstants.DELETE_OLD_DATA_AFTER
 import de.westnordost.streetcomplete.ApplicationConstants.REFRESH_DATA_AFTER
 import de.westnordost.streetcomplete.BuildConfig
+import de.westnordost.streetcomplete.HasTitle
 import de.westnordost.streetcomplete.Injector
 import de.westnordost.streetcomplete.Prefs
 import de.westnordost.streetcomplete.R
@@ -33,7 +34,7 @@ import java.util.*
 import javax.inject.Inject
 
 /** Shows the settings screen */
-class SettingsFragment : PreferenceFragmentCompat(),
+class SettingsFragment : PreferenceFragmentCompat(), HasTitle,
     SharedPreferences.OnSharedPreferenceChangeListener, VisibleQuestTypeSource.Listener {
 
     @Inject internal lateinit var prefs: SharedPreferences
@@ -53,6 +54,8 @@ class SettingsFragment : PreferenceFragmentCompat(),
     init {
         Injector.applicationComponent.inject(this)
     }
+
+    override val title: String get() = getString(R.string.action_settings)
 
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
         PreferenceManager.setDefaultValues(requireContext(), R.xml.preferences, false)
@@ -99,7 +102,6 @@ class SettingsFragment : PreferenceFragmentCompat(),
 
     override fun onStart() {
         super.onStart()
-        activity?.setTitle(R.string.action_settings)
         visibleQuestTypeSource.addListener(this)
         updateQuestPreferenceSummary()
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -23,8 +23,11 @@ import de.westnordost.streetcomplete.data.download.tiles.DownloadedTilesDao
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataController
 import de.westnordost.streetcomplete.data.osmnotes.NoteController
 import de.westnordost.streetcomplete.data.quest.QuestController
+import de.westnordost.streetcomplete.data.quest.QuestTypeRegistry
+import de.westnordost.streetcomplete.data.visiblequests.VisibleQuestTypeController
 import de.westnordost.streetcomplete.ktx.format
 import de.westnordost.streetcomplete.ktx.toast
+import de.westnordost.streetcomplete.settings.questselection.getSubtitle
 import kotlinx.coroutines.*
 import java.util.*
 import javax.inject.Inject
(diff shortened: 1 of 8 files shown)


matkoniecz:
Why the title is jumping a bit?

"Quest selection and priority" moves a bit too right. Tested on emulator for now - not appearing earlier.

Peek 2021-05-04 22-49

matkoniecz:
And big thanks for implementing it!

FloEdelmann:
I can't reproduce that jumping behavior on the emulator. I haven't tested on a real device yet.

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
```diff
@@ -39,6 +42,8 @@ class SettingsFragment : PreferenceFragmentCompat(),
     @Inject internal lateinit var mapDataController: MapDataController
     @Inject internal lateinit var questController: QuestController
     @Inject internal lateinit var resurveyIntervalsUpdater: ResurveyIntervalsUpdater
+    @Inject internal lateinit var questTypeRegistry: QuestTypeRegistry
+    @Inject internal lateinit var visibleQuestTypeController: VisibleQuestTypeController
```
westnordost, line 46: should use the read-only interface `VisibleQuestTypeSource`

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
```diff
@@ -100,6 +106,7 @@ class SettingsFragment : PreferenceFragmentCompat(),
     override fun onResume() {
         super.onResume()
         prefs.registerOnSharedPreferenceChangeListener(this)
+        updateQuestPreferenceSummary()
```
westnordost, line 109: Theoretically this data could change while the settings fragment did not move into the backround. I am thinking about the tablet-setup where you still see the settings menu on the left and the "submenu" on the right. I think currently this is not shown like that, but it would be more stable if you listened to changes on the VisibleQuestTypeSource on `onStart` and stopped listening on `onStop`. Whenever the visible quest type count changes, update the subtitle.

app/src/main/java/de/westnordost/streetcomplete/settings/questselection/QuestSelectionFragment.kt
```diff
@@ -113,4 +128,16 @@ class QuestSelectionFragment
         questTypeOrderList.sort(questTypes)
         return questTypes.map { QuestVisibility(it, visibleQuestTypeController.isVisible(it)) }.toMutableList()
     }
+
+    private fun updateSubtitle() {
+        val subtitle = getSubtitle(requireContext(), questTypeRegistry, visibleQuestTypeController)
+        (requireActivity() as AppCompatActivity).supportActionBar?.subtitle = subtitle
```
westnordost, line 134: Ah, please don't do that. A fragment should not control its parent activity like that.

There is an interface called `HasTitle`. You could expand it with another field `val subtitle: String?`. The logic should happen in `FragmentContainerActivity`. It is probably necessary to add a public method to `FragmentContainerActivity` called something like `updateTitle` then. Or maybe better: An interface called `ContainsTitle` or similar that is implemented by `FragmentContainerActivity` that has this method. (Since any fragment could also be the child of anoter fragment)

The `QuestSelectionFragment` would then talk to its parent in the same pattern as all other Fragments in StreetComplete do:
```kotlin
private val parentTitleContainer: TitleContainer ? get() = 
    parentFragment as? TitleContainer ?: activity as? TitleContainer
```

app/src/main/java/de/westnordost/streetcomplete/settings/questselection/QuestSelectionFragment.kt
```diff
@@ -113,4 +128,16 @@ class QuestSelectionFragment
...
+    }
+}
+
+fun getSubtitle(context: Context, questTypeRegistry: QuestTypeRegistry,
+                visibleQuestTypeController: VisibleQuestTypeController): String {
+    val enabledCount = questTypeRegistry.all.count { visibleQuestTypeController.isVisible(it) }
+    val totalCount = questTypeRegistry.all.size
+    return context.getString(R.string.pref_subtitle_quests, enabledCount, totalCount)
```
westnordost, line 142: I think it is best to inline that. This function only saves about 1 line of code to be shared and its name is so generic ("get subtitle of what??") that it may be confusing - after all you could import this global function anywhere.

If one wanted to reduce the logic here, I could come up with two things:
1. QuestTypeRegistry could implement the list interface itself instead of exposing `all` ... I think there was a neat syntax trick in Kotlin for that,
```kotlin 
QuestTypeRegistry(private val all: List<QuestType<*>>): List<QuestType<*>> by all
```  
or something like that

2. create extension function `QuestTypeRegistry.getAllVisible(visibleQuestTypeSource: VisibleQuestTypeSource): List<QuestType<*>>`

westnordost (changes requested):
There is just one thing I'd like to nitpick about. The code in general is fine, it's just "mosern auf hohem Niveau":

### Listener / Observer architecture

I've been trying to slowly establish this kind of architecture for a while now:
- the view just displays what it gets from those sources/controllers, i.e. doesn't keep an own state
- the view just tells those controllers what should be changed
- the view does not directly update the view as response to an action being taken in the UI, but because it got notified that the data has changed (as direct result of that action, but that logic is nothing that should be of concern for the UI)

I'd prefer if QuestSelectionFragment would not update the UI as a response to the user tap, but as a response to the data changed. You can register a listener on `VisibleQuestTypeSource` to get notified of changes in that data. 

This point is optional, because this pattern is currently also not followed through in that fragment - otherwise `VisibleQuestTypeSource.Listener` would need to be more fine-grained (i.e. have a method like `onQuestTypeVisibilityChanged(QuestType, Boolean)`)

westnordost (approved):
Nice, all clear. Thank you!

PR #2848, opened 2021-05-04
Fix specifying shop is vacant

peternewman:
So it sets it as vacant, fixing the bug from a7ad093aa80c97c6cda7f0e57a4f5a89f09f8f4d

Untested, but aside from compilation presumably not much different to a7ad093aa80c97c6cda7f0e57a4f5a89f09f8f4d :laughing:


matkoniecz:
Thanks for spotting that!

It would be better to use `disused:shop=yes` here

matkoniecz:
> use disused:shop=yes

is now done, but also not (yet) tested properly

matkoniecz:
now it is tested

peternewman:
Great, thanks for the tweak and merge.

I'd thought it was a bug with the new offline stuff until I started looking at the code! I just got lucky with locations I guess.

Issue #2849, opened 2021-05-05
Some ambiguous quest title wordings in the settings


cyclingcat:
Hi there,

some German quest titles in the settings are ambiguous. At least I (but probably many more users) sometimes don't know what exactly I'm going to activate or deactivate here.

See the following screenshot, where I reordered some quests and marked three pairs of them. I did a reverse lookup at POEditor and found out that the English wording is partly ambiguous as well:

![quests_german](https://user-images.githubusercontent.com/32601012/117102842-6d388200-ad79-11eb-8e53-43bc3a9e70a9.jpg)

**1. The red pair:**
Here the second quest string is a part of the first one. In English it's similar ("Is this a one-way street?" vs. "Is this a one-way street? In which direction?"). What's the difference between these quests? The wording of the first quest suggests that it's in fact two quests in one. (Both seem to be a very rare question, according to my own statistics I got just a handful of "direction" quests and not a single one of the other.)

**2. The green pair:**
This is even more confusing as the German titles are identical. In English they are at least slightly different ("Is this location wheelchair accessible?" vs. "Is this place wheelchair accessible?"), but is this sufficient to distinguish between the two quests?

**3. The blue pair:**
An example for a better wording (in English "path" vs. "piece of road"), just as an illustration how it could be improved.

Maybe I'll be able to make suggestions for better German wordings but first I have to fully understand what these quests are exactly asking for. All tested with v32-alpha1, but I noticed this with earlier version as well.

The cycling cat

matkoniecz:
> In English it's similar ("Is this a one-way street?" vs. "Is this a one-way street? In which direction?"). What's the difference between these quests? 

For user? None? Difference is how roads are selected - one proposed by some external API, one based on tagged `width` tag.

It would be nice to merge them but it would be complicated from technical part.

>  This is even more confusing as the German titles are identical. In English they are at least slightly different ("Is this location wheelchair accessible?" vs. "Is this place wheelchair accessible?"), but is this sufficient to distinguish between the two quests?

Green is "AddWheelchairAccessOutside" - and right now it asks solely about `leisure = dog_park`

Blue is "AddWheelchairAccessPublicTransport" and I see way to fix it - not sure is it worth doing it given translation effort but I will make PR

westnordost:
I decided not to fix this. If there will be presets... one day, users will probably not want this fine control anyway anymore.

PR #2850, opened 2021-05-05
try to make titles more clear in the quest list

matkoniecz:
fixes part of #2849 (remaining part if not fixable in easy or reasonable way)

screen05

Previously this quests had basically the same titles.


westnordost:
Except a clearer message in the quest selection list, is there any added value in this? Because otherwise I'd rather close this as will-not-fix. This wording is really awkward, maybe even confusing. Also of course, it is added effort for translators.

smichel17:
Wouldn't it be better to work out a way for these to be shown as the same quest? I can't imagine people would want to disable one, but not the other, or give them a different priority.

PR #2852, opened 2021-05-05
implements #2774 and allow tagging lane count without markings

### LINKED ISSUE #2774: Add "no markings, clear lane count'
**Use case**
In many cases there are no lane markings, but lane count is very clear. (I can attach Street View examples)

**Proposed Solution**
Add one more option to leave quest. It should not cause invalid data appearing, as there is a clear escape hatch with existing "no markings" option

I will definitely implement it, not sure should I push it to fork only or also open StreetComplete PR

matkoniecz:
fixes #2774

note that per discussion in #2774 it is unclear whether this will be accepted and requires further tweaks if accepted

opening as I tested and reviewed this code

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/lanes/AddLanes.kt
@@ -35,7 +35,7 @@ class AddLanes : OsmFilterQuestType<LanesAnswer>() {
 
         laneCount?.let { changes.addOrModify("lanes", it.toString()) }
 
-        val isMarked = answer !is UnmarkedLanes
+        val isMarked = answer is MarkedLanes || answer is MarkedLanesSides
         // if there is just one lane, the information whether it is marked or not is irrelevant
         // (if there are no more than one lane, there are no markings to separate them)
         when {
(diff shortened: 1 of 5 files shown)


westnordost:
To determine whether it is obvious is itself fuzzy. You posted that picture of some track... someone could also say "well it's 1.5-laned". Cars can still pass each other very cautiously.

But that discussion aside, I can say that I'll not merge it in the state shown in https://github.com/streetcomplete/StreetComplete/issues/2774#issuecomment-824231402 
If it was changed like suggested here https://github.com/streetcomplete/StreetComplete/issues/2774#issuecomment-824245129 to be honest, I'm still not sure if this should be merged, but it would be better at least. Because there is some oppurtunity to explain to the user what this means.

matkoniecz:
For road in https://github.com/streetcomplete/StreetComplete/issues/2774#issuecomment-832599154 there is no chance whatsoever that normal cars would be able to pass each other without leaving asphalt

I think that in this case it is 100% obvious `lanes=1` (if not, then this change is not useful).

Another shot of road of that size.

IMG_20210429_173347

westnordost:
Well, obviously, there are situations in which it is clear an obvious. But presenting to the user the option to first decide if it is obvious is just shifting the problem to deciding whether the lane count is fuzzy or not. Which is in itself already a fuzzy decision.

We have had multiple iterations on the lanes quest already. I think it is best if it remained this way for now.

Also, not that for roads with no marked lanes or lanes<=1, it is planned to ask for the width of the road and let the user measure it with their smartphone. So that solves this use case conclusively and data consumers will/should be happe to accept the width as an alternative to `lanes` too.

PR #2854, opened 2021-05-05
Fix issue where bridge structure quest was too tough to identify

TurnrDev:
I never knew what to pick on this quest as my bridges near me are much shorter and go over railway lol.  so I added titles and descriptions. I tested and it builds :) 

Before:

After: 

Not sure why the descriptions didn't show?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bridge_structure/AddBridgeStructureForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bridge_structure/AddBridgeStructureForm.kt
@@ -9,15 +9,15 @@ class AddBridgeStructureForm : AImageListQuestAnswerFragment<BridgeStructure, Br
 
     // structures sorted highest to lowest amount of values on taginfo, footbridge-types last
     override val items = listOf(
-        Item(BEAM, R.drawable.bridge_structure_beam),
-        Item(SUSPENSION, R.drawable.bridge_structure_suspension),
-        Item(ARCH, R.drawable.bridge_structure_arch),
-        Item(TIED_ARCH, R.drawable.bridge_structure_tied_arch),
-        Item(TRUSS, R.drawable.bridge_structure_truss),
-        Item(CABLE_STAYED, R.drawable.bridge_structure_cablestayed),
-        Item(HUMPBACK, R.drawable.bridge_structure_humpback),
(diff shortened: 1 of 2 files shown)


westnordost:
Because users should not be expected to know the technical terms for the different kinds of bridges. This is deliberate. For roof shapes, the names are also not shown.

westnordost:
So, this will not be merged, but what would be merged is to switch the pictures with some more common bridges (as most bridges are not those huge bridges).

TurnrDev:
I'll take a look at what I can find, I was thinking of using a collage of images of similar bridges for each type so people can clearly see the characteristic that is being pointed out :)

westnordost:
(Side-)graphics would also be possible, if you are gifted

image

TurnrDev:
> (Side-)graphics would also be possible, if you are gifted
> 
> image
> 

I've not tackled vectors yet but I'll try it

TurnrDev:
I just drew a rough plan on paper. For beam, I seen that used a lot in urban environments when a road or street does over another road, street or railway, so that's gonna have an over water and urban style split in the middle to easily show it's common uses. I don't feel like others need this treatments they're often more identifiable.

I got illustrator installed and will give it a tackle 😊

westnordost:
The pictures don't need to be square. For bridges shown from the side, maybe 2:1 would be good as well.

Inkscape is also fine, especially for simple SVGs

TurnrDev:
> The pictures don't need to be square.

I wish I read this before I started, oops!

Anyway, I've been working on the split scenery for beam, thoughts?
image

westnordost:
I see that you put quite some effort into this, with the water, the clouds, the railway and the skyline. 👍  It kinda looks nice, but I have to say that it misses its purpose. The bridge itself is almost not visible on this picture, or at least it is not clear that it is a bridge and the purpose of this picture is to show the structure of the bridge. There is just so much else going on in that image, and the yellow-brown of the bridge silhouette just vanishes in the background.

It is irrelevant whether the bridge goes over water or railway, so there is no need to have this kind of split screen view which IMO is just confusing.

In my opinion, what would work best would be some kind of silhoutte view, black on white. Maybe use a third color for cable or a third color to indicate where is the ground, but that's it. There is no need to have any kind of background or decoration, it all just distracts from what this should be about.
Take the roof shape graphics for example, there is also just the roof and nothing else. No color variation on the roofs, no height variation, no decoration, no shape variation, no scale on the buildings.

Issue #2354, opened 2020-12-08
Show other nearby objects of the same type when answering quests


peternewman:
**Use case**
There have been a number of quests already where it would help (e.g. bench backrests, postboxes etc), but it's become even more vital with "Is x still present?". We need to be able to see all the other nearby x's while answering the quest. But likewise for any other nodes not visible on the map, so when I answer or re-confirm a postbox, it can show the nearby ones so I know I'm actually answering the quest on the right node.

**Proposed Solution**
When you select the bin for example, it should show the other nearby bins in a say 30m radius or whatever as dots on the map (probably in a different colour to the quest dot). Presumably this shouldn't be too hard now you're downloading the whole tile and processing it offline, we'd only need to keep the co-ordinates of them.

Otherwise any time the nodes are added at different times (e.g. if I notice one is missing and add it now), or just if I didn't get a chance to answer both adjacent postboxes in one go, myself or the next person is left with a pin/quest and no idea which of the two it applies to.

This is on v28.0-beta1 onwards

westnordost:
Would surely be a good improvement, but would also be very much work to implement. I do see it less dramatic than you, because when answering that the X is not there, the user is warned that he should look around if he sees the X somewhere near and that in doubt, he should rather leave a note. Leaving a note will also lead to that the maybe wrongly placed X is moved to the correct location.

The other problematic case would be that there are two Xs really close to one another and people might wrongfully answer that both exist even though one is gone (already). I'd expect that the user will notice that something is not right when he is asked for the second X, and will leave a note. But even if he doesn't notice (or the situation is as described in your last paragraph), it would only lead to that a bogus X still continues to be on the map. It's not nice, but not critical.

peternewman:
> Would surely be a good improvement, but would also be very much work to implement.

I thought you're fetching and parsing the tile already to find quests, so storing just the co-ordinates of the nearest `amenity=phone_box` and rendering on the map wouldn't be too complex. Or do you just mean it's a lot more complicated than adding another quest? Rendering them on the map would also work, but I'm assuming they're not in the tile.

> I do see it less dramatic than you, because when answering that the X is not there, the user is warned that he should look around if he sees the X somewhere near and that in doubt, he should rather leave a note. Leaving a note will also lead to that the maybe wrongly placed X is moved to the correct location.

[...]

PR #2855, opened 2021-05-06
Ask for pitch lit for nodes too

peternewman:
To match the base sport quest

Untested, but copied from the base sport one

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/pitch_lit/AddPitchLit.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/pitch_lit/AddPitchLit.kt
@@ -10,7 +10,7 @@ import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
 class AddPitchLit : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
-        ways with (leisure = pitch or leisure = track)
+        nodes, ways with (leisure = pitch or leisure = track)
         and (access !~ private|no)
         and indoor != yes and (!building or building = no)
         and (


peternewman:
CC @eginhard

matkoniecz:
@peternewman Can you test it?

peternewman:
> @peternewman Can you test it?

Sorry I've still not got Android Studio installed and not really ready to start fighting with that yet, think of this more as an issue report with a suggested fix.

matkoniecz:
Still very useful, thanks for submitting it!

westnordost:
Fighting with it? You just install it. Or do you run some obscure operating system for which there is no installer?

westnordost:
Android studio comes packaged with its own gradle

westnordost:
Sorry, gotta revert that after all. Without seeing the outline of a sports pitch, it is often not really possible to see which pitch is meant. This is especially true for table tennis "pitches" but for others too, f.e. within a larger area where there are multiple pitches.

peternewman:
> Sorry, gotta revert that after all.

Fair enough.

> Without seeing the outline of a sports pitch, it is often not really possible to see which pitch is meant. This is especially true for table tennis "pitches" but for others too, f.e. within a larger area where there are multiple pitches.

Doesn't this apply to https://github.com/streetcomplete/StreetComplete/blob/31a51693165996b4f79ff449f5667c36b1493bf9/app/src/main/java/de/westnordost/streetcomplete/quests/sport/AddSport.kt#L17 too then? i.e. shouldn't the type of sport and whether it's lit have the same challenges/limitations?

I guess this would be another good candidate for fixing via #2354 if that gets implemented at some point.

westnordost:
Hmm... yeah, probably yes. I'll also remove `nodes` there.

Helium314:
Would it be possible to check whether there are other `leisure = pitch` elements/nodes nearby, like it is done for the recycling quest? If there is only one within sth like 20-50 m, there should be no problem

peternewman:
> Would it be possible to check whether there are other `leisure = pitch` elements/nodes nearby, like it is done for the recycling quest? If there is only one within sth like 20-50 m, there should be no problem

Sounds good, I'd pondered something similar when @westnordost mentioned it, but I didn't realise code for it already existed. I guess it would need to exclude nodes when there was a nearby node or way of the right type.

westnordost:
I'd like to keep such distance checks to the necessities, because they are very CPU intensive to do (and I would like to not let the code grow too complex)

Issue #2840, opened 2021-05-03
[v32.0-alpha1] Some small bugs


cyclingcat:
Hi there,

I just installed the freshly released v32.0-alpha1 apk on my devices in order to help find bugs before this version with its major changes will be released to the public. Here are my first observations, I collect them in this ticket. Let me know if you'd prefer one ticket for each issue, I will change it then and close this one.

**Preconditions** 
* My usual two Android devices (smartphone with Android 9.0, tablet with Android 8.1)
* SC v32.0-alpha1 apk downloaded from GitHub and updated from v31.3.

**The issues I encountered so far**

1. Directly after the update the quest count showed a value of 0. However, the information on the "My profile" screen was correct (OSM username, quest count, rankings); after going back to the main screen the number after the star symbol was now correct as well.

2. All the quests I have disabled in my settings were reenabled again! The changed order, however, was preserved.

3. At least some (maybe all) quests I have hidden in the past are visible again.

4. In some quests texts a mysterious number appears. See the following screenshot, (it's [node 4810322593](https://www.openstreetmap.org/node/4810322593)):<br/>
![ref](https://user-images.githubusercontent.com/32601012/116851327-79de9e00-abf2-11eb-9387-6b4f31308e22.jpg)<br/>
The number is the content of a "ref" tag, it seems that this tag is now shown in the text at least for the existence quest if no "name" tag is present for the corresponding node. The node in the screenshot above contains an "operator" tag which would be a better choice for the quest text IMHO as the "ref" tag is often just confusing information for the user. (I have never seen this before with earlier versions. so I'd assume that it has been introduced with v32.)

All of these phenomenons are likely to confuse users, so I'd suggest to find a solution for them. If 2. and 3. are consequences of the architectural change of the app, at least a hint should be presented to the user.

The cycling cat

westnordost:
Thank you for the report! Most of this is not really a bug.

> 1, 2, 3

This is because the whole local database was deleted (replaced), as announced. The statistics, what quests are hidden, what quest types are disabled are all saved in the database.
What is not stored in the database (but in another persistent storage) are your OSM login and the quest order.

So yeah, this might be confusing, but it is likely a will not fix / working as intended.

> 4 

Hm true. I fear it depends on the context if rather `ref` or rather `operator` should be shown. For example, for bus stops, ref is certainly more helpful than the operator. (In the US, many bus stops do not have names but some numbers/identifiers).

mnalis:
[...]

PR #2864, opened 2021-05-09
Add some additional notes to the changelog

peternewman:
As mentioned in https://github.com/streetcomplete/StreetComplete/discussions/2834#discussioncomment-688164 and https://github.com/streetcomplete/StreetComplete/issues/2840#issuecomment-831205491


westnordost:
The alpha versions are not upgraded from google play. That starts with the first beta. But ok...

PR #2872, opened 2021-05-11
Only show lit quests during night-time

### LINKED ISSUE #1285: Consider enabling "Is lit" quest only during night
From https://github.com/westnordost/StreetComplete/issues/1281#issuecomment-443621518

> It's often impossible (!) to solve these quests at daylight time when the lights are out. For example: Does a road light reach the (separately mapped) parallel footway as well? (It often does, but you cannot be sure until it's dark!) So I decided for me to disable this quest also and to deliberately reenable it if I go out for an evening walk in the dark

I do exactly the same thing for the same reasons (I even moved lit quest earlier to make enabling/disabling it easier).

Maybe it would be a good idea to do it automatically? Calculating sunset/sunrise based on a location and date should be available in some library.

TurnrDev:
Fixes #1285 by working out the time of civil sunset and sunrise

It currently works via mayCreateQuest the same way blocking quests out of the country works, but that's causing that sync issue when toggling the setting. I need to move the logic to displaying the quest instead of creating the quest, I'm not so sure where that would be in the code base so any help would be appreciated.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -176,6 +176,9 @@ dependencies {
 
     // opening hours parser
     implementation("ch.poole:OpeningHoursParser:0.25.0")
+
+    // sunset-sunrise parser for lit quests
+    implementation("com.luckycatlabs:SunriseSunsetCalculator:1.2")
 }
 
 /** Localizations that should be pulled from POEditor etc. */
--- a/app/src/main/java/de/westnordost/streetcomplete/data/quest/DayNightCycle.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/quest/DayNightCycle.kt
@@ -1,0 +1,3 @@
+package de.westnordost.streetcomplete.data.quest
+
+enum class DayNightCycle { DAY_AND_NIGHT, ONLY_DAY, ONLY_NIGHT }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/DayNightQuestFilter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/DayNightQuestFilter.kt
@@ -1,0 +1,20 @@
+package de.westnordost.streetcomplete.data.visiblequests
+
+import de.westnordost.streetcomplete.data.quest.DayNightCycle.*
+import de.westnordost.streetcomplete.data.quest.Quest
+import de.westnordost.streetcomplete.util.isDay
+import javax.inject.Inject
+
+class DayNightQuestFilter @Inject internal constructor() {
+    /*
+    Might be an idea to add a listener so this is reevaluated occasionally, or something like that.
+    However, I think it's reevaluated everytime the displayed quests are updated?
+     */
+    fun isVisible(quest: Quest): Boolean {
+        return when (quest.type.dayNightCycle) {
+            DAY_AND_NIGHT -> true
+            ONLY_DAY -> isDay(quest.position)
+            ONLY_NIGHT -> !isDay(quest.position)
+        }
+    }
+}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
@@ -128,7 +128,11 @@ import javax.inject.Singleton
         // ↓ 1. notes
         OsmNoteQuestType,
 
-        // ↓ 2. important data that is used by many data consumers
+        // ↓ 2. time-specific quests, such as lit quests which are moved to only show ar night
+        AddWayLit(), //  used by OsmAnd if "Street lighting" is enabled. (Configure map, Map rendering, Details)
+        AddBusStopLit(),
+
+        // ↓ 3. important data that is used by many data consumers
         AddRoadName(),
         AddPlaceName(featureDictionaryFuture),
         AddOneway(),
@@ -152,7 +156,7 @@ import javax.inject.Singleton
         AddReligionToPlaceOfWorship(), // icons on maps are different - OSM Carto, mapy.cz, OsmAnd, Sputnik etc
         AddParkingAccess(), //OSM Carto, mapy.cz, OSMand, Sputnik etc
 
-        // ↓ 3. useful data that is used by some data consumers
+        // ↓ 4. useful data that is used by some data consumers
         AddRecyclingType(),
         AddRecyclingContainerMaterials(),
         AddSport(),
@@ -191,15 +195,14 @@ import javax.inject.Singleton
         AddFerryAccessMotorVehicle(),
         AddAcceptsCash(featureDictionaryFuture),
 
-        // ↓ 4. definitely shown as errors in QA tools
+        // ↓ 5. definitely shown as errors in QA tools
 
-        // ↓ 5. may be shown as missing in QA tools
+        // ↓ 6. may be shown as missing in QA tools
         DetermineRecyclingGlass(), // because most recycling:glass=yes is a tagging mistake
 
-        // ↓ 6. may be shown as possibly missing in QA tools
+        // ↓ 7. may be shown as possibly missing in QA tools
 
-        // ↓ 7. data useful for only a specific use case
-        AddWayLit(), //  used by OsmAnd if "Street lighting" is enabled. (Configure map, Map rendering, Details)
+        // ↓ 8. data useful for only a specific use case
         AddToiletsFee(), // used by OsmAnd in the object description
         AddBabyChangingTable(), // used by OsmAnd in the object description
         AddBikeParkingCover(), // used by OsmAnd in the object description
@@ -237,10 +240,10 @@ import javax.inject.Singleton
         AddBollardType(), // useful for first responders
         AddCameraType(),
 
-        // ↓ 8. defined in the wiki, but not really used by anyone yet. Just collected for
+        // ↓ 9. defined in the wiki, but not really used by anyone yet. Just collected for
         //      the sake of mapping it in case it makes sense later
         AddPitchSurface(),
-        AddPitchLit(),
+        AddPitchLit(), // Not affected by new DayNight cycle because the lights are usually only on during games
         AddIsDefibrillatorIndoor(),
         AddSummitRegister(),
         AddCyclewayPartSurface(),
@@ -256,7 +259,6 @@ import javax.inject.Singleton
         AddCarWashType(),
         AddBenchStatusOnBusStop(),
         AddBinStatusOnBusStop(),
-        AddBusStopLit(),
         AddBenchBackrest(),
         AddTrafficSignalsButton(),
         AddPostboxRoyalCypher()
--- a/app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestType.kt
@@ -19,4 +19,7 @@ interface QuestType<T> {
 
     /** The quest type can clean it's metadata that is older than the given timestamp here, if any  */
     fun deleteMetadataOlderThan(timestamp: Long) {}
+
+    /** if the quest should only be shown during day-light os night-time hours */
+    val dayNightCycle: DayNightCycle get() = DayNightCycle.DAY_AND_NIGHT
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
@@ -5,6 +5,7 @@ import de.westnordost.streetcomplete.data.osm.osmquests.OsmQuest
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmQuestSource
 import de.westnordost.streetcomplete.data.osmnotes.notequests.OsmNoteQuest
 import de.westnordost.streetcomplete.data.osmnotes.notequests.OsmNoteQuestSource
+import de.westnordost.streetcomplete.data.visiblequests.DayNightQuestFilter
 import de.westnordost.streetcomplete.data.visiblequests.TeamModeQuestFilter
 import de.westnordost.streetcomplete.data.visiblequests.VisibleQuestTypeSource
 import java.util.concurrent.CopyOnWriteArrayList
@@ -17,7 +18,8 @@ import javax.inject.Singleton
     private val osmQuestSource: OsmQuestSource,
     private val osmNoteQuestSource: OsmNoteQuestSource,
     private val visibleQuestTypeSource: VisibleQuestTypeSource,
-    private val teamModeQuestFilter: TeamModeQuestFilter
+    private val teamModeQuestFilter: TeamModeQuestFilter,
+    private val dayNightQuestFilter: DayNightQuestFilter,
 ) {
     interface Listener {
         /** Called when given quests in the given group have been added/removed */
@@ -82,7 +84,7 @@ import javax.inject.Singleton
     }
 
     private fun isVisible(quest: Quest): Boolean =
-        visibleQuestTypeSource.isVisible(quest.type) && teamModeQuestFilter.isVisible(quest)
+        visibleQuestTypeSource.isVisible(quest.type) && teamModeQuestFilter.isVisible(quest) && dayNightQuestFilter.isVisible(quest)
 
     fun addListener(listener: Listener) {
         listeners.add(listener)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_lit/AddBusStopLit.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_lit/AddBusStopLit.kt
@@ -4,6 +4,7 @@ import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.DayNightCycle.ONLY_NIGHT
 import de.westnordost.streetcomplete.ktx.arrayOfNotNull
 import de.westnordost.streetcomplete.ktx.containsAnyKey
 import de.westnordost.streetcomplete.ktx.toYesNo
@@ -29,6 +30,7 @@ class AddBusStopLit : OsmFilterQuestType<Boolean>() {
     override val commitMessage = "Add whether a bus stop is lit"
     override val wikiLink = "Key:lit"
     override val icon = R.drawable.ic_quest_bus_stop_lit
+    override val dayNightCycle = ONLY_NIGHT
 
     override fun getTitle(tags: Map<String, String>): Int {
         val hasName = tags.containsAnyKey("name", "ref")
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/way_lit/AddWayLit.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/way_lit/AddWayLit.kt
@@ -3,8 +3,9 @@ package de.westnordost.streetcomplete.quests.way_lit
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.meta.MAXSPEED_TYPE_KEYS
 import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
-import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.DayNightCycle.ONLY_NIGHT
 
 class AddWayLit : OsmFilterQuestType<WayLit>() {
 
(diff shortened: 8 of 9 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/quest/DayNightCycle.kt
```diff
@@ -0,0 +1,7 @@
+package de.westnordost.streetcomplete.data.quest
+
+sealed class DayNightCycle
+
+object DayAndNight : DayNightCycle()
+object OnlyDay : DayNightCycle()
+object OnlyNight : DayNightCycle()
```
westnordost, line 7: Any reason to use a sealed class? An enum would do, too

```suggestion
enum class DayNightCycle { DAY_AND_NIGHT, ONLY_DAY, ONLY_NIGHT }
```

app/build.gradle.kts
```diff
@@ -170,6 +170,9 @@ dependencies {
 
     // opening hours parser
     implementation("ch.poole:OpeningHoursParser:0.23.0")
+
+    // sunset-sunrise parser for lit quests
+    implementation("com.luckycatlabs:SunriseSunsetCalculator:1.2")
```
westnordost, line 175: I am always cautious when adding new dependencies. It looks like the calculation of that is not trivial, but did you review and compare all the different libraries that do implement that?

A short search revealed...
- https://github.com/shred/commons-suncalc
- https://github.com/caarmen/SunriseSunset
- https://github.com/mikereedell/sunrisesunsetlib-java (the one used by you)

The questions one gotta ask is whether they are well tested, whether they have any dependencies (they should not, if possible), whether it is maintained (an algorithm once written is finished, but maybe there are any bugs that are not addressed?) and finally, of course, whether it actually fulfills our needs.
(P.S: Java Date/Time classes from Java 8 **are** available in this app)

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt
```diff
@@ -222,6 +229,18 @@ import javax.inject.Singleton
...
+
+        if (!prefs.getBoolean(Prefs.ALWAYS_SHOW_TIME_DEPENDENT, false)) {
+            if (!when (dayNightCycle) {
+                    is DayAndNight -> true
+                    is OnlyDay -> !isNight(pos)
+                    is OnlyNight -> isNight(pos)
+                }
+            ) return false
+        }
+
```
westnordost, line 243: The logic should go into an own class. For example `DayNightQuestFilter`. This class could have just one method: `isVisible(quest: Quest): Boolean`

Also, the the `dayNightCycle` is not necessarily limited to OSM quests, so the better place for this would be `VisibleQuestsSource`

westnordost (changes requested):
What is the use case for displaying this option at all and not always only show those quests that should be only visible by night at night? Would make implementation a little easier, less code and less clutter in the settings menu.

It would be ideal if the app could register a callback whenever the sunset or sunrise event occurs. Maybe one of the libraries mentioned can do that. Then, the `VisibleQuestsSource` could listen for that and invalidate the quests so that they are refreshed.

TurnrDev:
Thank you for the review, I will take a closer look tomorrow and see what I can do to fix it up!

smichel17:
For reference, other conversations related to displaying quests are collected here: [#2789 (comment)](https://github.com/streetcomplete/StreetComplete/discussions/2789?sort=top#discussioncomment-646707).
It might be worth giving them a skim — or not, some are quite long :P

app/build.gradle.kts
TurnrDev: https://github.com/caarmen/SunriseSunset seems to only include maven as a dependency, has some nice convenience methods and looks to be fairly well maintained, albeit not touched for a while. 

It's used by Telefónica, which gives me confidence in its abilities

https://github.com/shred/commons-suncalc has a few issues related to bugs, which have since been fixed but it doesn't give confidence in their testing. 

Feel free to criticise my judgement, I don't know a lot about Java so I'm not sure I'd be able to review the code for bugs myself.

TL;DR: Gonna switch to https://github.com/caarmen/SunriseSunset

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt
TurnrDev: How can I query OsmElementQuestType variables like the new `dayNightCycle` I declared in [OsmElementQuestType.kt#L29](https://github.com/streetcomplete/StreetComplete/blob/ecb25f986ed93ca8ad4d635a0dd83bda9c7d7185/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt#L29) from the Quest object given to isVisible? 

I have tried `quest.type.dayNightCycle` but Android Studio is saying no 😭

westnordost: Because you added the property to `OsmElementQuestType.kt`, not `QuestType.kt`

TurnrDev: If it's added to QuestType.kt will it still work the same? Are there any quests that aren't OsmElementQuestType's? :)

westnordost: currently, only note quests

app/build.gradle.kts
TurnrDev: caarmen/SunriseSunset had a weird bug, had to not use it. Explain tomorrow :) It's currently working well tho

TurnrDev: Okay so caarmen's library war getting the wrong times for sunsets entirely. It was trying to tell me sunset was 2:02AM lat night for me when it was actually 20:45ish.

I have temporarily reverted back to the original library, but will take a look at the other one later today.

westnordost: Maybe the location entered was wrong? Switched lat and lon?

smichel17: Hey, finally something I have some expertise with!

Red Moon uses `com.luckycatlabs:SunriseSunsetCalculator`, mostly because it is the oldest and the only one available at the time when it was added. At some point, when the other libraries popped up, I did a pretty comprehensive review. Between that & my experience, my takeaways are:

- There is a fair amount of room for variation, because sunrise/sunset happen over a length of time.
- The different libs do not all use the same algorithm. Some are more accurate than others, but they are all approximations (Also, the amount of light hitting you varies depending on whether you're in a valley or mountain, etc, which these libs can't take into account due to technical constraints). Some of them perform better at different latitudes (closer/farther from the poles).
- `com.luckycatlabs:SunriseSunsetCalculator` is not the most accurate (I think — my memory is a little fuzzy). However, I decided that this didn't really matter for me. The inherent variation mentioned above was enough that even if it's off by 10 minutes from some ideal time to swap, it doesn't really matter.
    - The api has been extremely stable. I don't think I've had to modify the [calling code](https://github.com/LibreShift/red-moon/blob/master/app/src/main/java/com/jmstudios/redmoon/model/Config.kt#L173-L175)… ever? Maybe once. I think there were some bugfixes early on (none that affected me), but now it's pretty mature. I wouldn't have a problem with copying it into my code base if I wanted to avoid external dependencies entirely.\*

IIRC the [SunTimes](https://f-droid.org/en/packages/com.forrestguice.suntimeswidget/) app on F-Droid has an option to choose which library you want to use to calculate the time (currently it looks like just two options, luckycatlabs\*\* and caarmen, although I thought I remembered one or two others being included in the past), if you want to play around with it and see the differences.

\*Although I'd probably keep it in a separate library inside the same repo, like [how I have TimePickerPreference in Red Moon](https://github.com/LibreShift/red-moon/tree/master/timepickerpreference). Incidentally, @westnordost I wonder if it makes sense to use that kind of structure for your countryboundaries and osm… libraries (either moving them to this repo, or all to one repo for libraries; in any case, they can still be published as independent packages).

\*\*Technically it is a fork with a bugfix, but the bugfix has since been merged upstream and the dependency just wasn't switched back.

TurnrDev: Hearing good things about `com.luckycatlabs:SunriseSunsetCalculator` has made me feel better about my decision.

I have yet to test what happens if Sunrise or Sunset is before or after midnight, or what happens at the poles.

app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt
```diff
@@ -0,0 +1,36 @@
...
+    val calendar = Calendar.getInstance()
+    calendar.set(localDate.year, localDate.month.value, localDate.dayOfMonth)
+    return calendar
+}
+
+fun isNight(pos: LatLon): Boolean {
+    /* This functions job is to check if it's currently dark out.
+    It will use the location of the node and check the civil sunset time.
+
+    Currently it checks the phone time and returns true if between 7pm and 7am.
```
westnordost, line 23: Is this old?

app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.time.ZonedDateTime
+import java.util.*
+
+fun localDateToCalendar(localDate: LocalDate): Calendar {
+    val calendar = Calendar.getInstance()
+    calendar.set(localDate.year, localDate.month.value, localDate.dayOfMonth)
+    return calendar
+}
+
+fun isNight(pos: LatLon): Boolean {
```
westnordost, line 19: Ok, I now recognize that here is a potential for a performance issue. Let's say you have 1000 quests that shall not be visible during day and they shall now be displayed on the map.

Then, the calculation in `CheckIfNight` is executed 1000 times. I am not sure how expensive the calculation is, but to circumvent that, I suggest the following:

1. `isNight` calculates the sunset/dawn with the given latitude and longitude truncated to two decimals ([precision of ~1km](https://wiki.openstreetmap.org/wiki/Precision_of_coordinates)). That should be enough precision
2. the calculation is cached in a dayTime: Map<LatLon, Pair<LocalTime,LocalTime>>. Only if there is no entry in the map, the calculation is done. LatLon is always the truncated latlon.
3. If `isNight` is going to have a state, it would make sense to make this into a class, or maybe just put the code here in `DayNightQuestFilter`

app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    val timezone = TimeZone.getDefault().id
+    val location = Location(pos.latitude, pos.longitude)
+    val calculator = SolarEventCalculator(location, timezone)
+    val now = ZonedDateTime.now(ZoneId.of(timezone))
+    val today = now.toLocalDate()
+    val tomorrow = today.plusDays(1)
+
+    val time1 = ZonedDateTime.of(today, LocalTime.parse(calculator.computeSunsetTime(Zenith.CIVIL, localDateToCalendar(today))), ZoneId.of(timezone))
+    val time2 = ZonedDateTime.of(tomorrow, LocalTime.parse(calculator.computeSunriseTime(Zenith.CIVIL, localDateToCalendar(tomorrow))), ZoneId.of(timezone))
```
westnordost, line 34: Are you sure that timezones are necessary here at all? If `calculator.computeSunsetTime` returns a string like "18:30", then why'd you need the timezone at all? Because of today / tomorrow? Well then why not turn it around. The function returns the time range from sunrise to sunset (that will be within one day) and it `isNight` when the current time is not within that range.

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/DayNightQuestFilter.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import de.westnordost.streetcomplete.data.quest.DayNightCycle.*
+import de.westnordost.streetcomplete.data.quest.Quest
+import de.westnordost.streetcomplete.util.isNight
+import javax.inject.Inject
+
+class DayNightQuestFilter @Inject internal constructor() {
+    /*
+    Might be an idea to add a listener so this is reevaluated occasionally, or something like that.
+    However, I think it's reevaluated everytime the displayed quests are updated?
```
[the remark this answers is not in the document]
TurnrDev: I did try to work out listeners but it did go over my head, wasn't sure what I was doing, sorry!

app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt
TurnrDev: Yes, oops, I'll fix the comment!

dbdean:
Would it be possible to tag (like `survey:atnight=yes` or something) to indicate that the answer was surveyed during night time?That way we could re-ask this for roads/paths/etc that are already tagged lit but weren't surveyed at night to verify that they are actually lit at night time, and didn't just look like they should be lit.

westnordost:
Well, this is the wrong place to propose a tag like this.

westnordost:
@TurnrDev What would be quite cool would be that if you go into the quest selection settings, it will actually show the quest as disabled and say something like "available in 3 hours" or something

app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt
TurnrDev: Whilst I fully agree with you, I'm not sure how to go about implementing this

smichel17: Hmm… … how should this feature actually work?

1. When it is detected to be after sunset at the user's current position, all lit quests appear.
    - How I imagined it would be when I read the title.
    - Should be synchronized with using the dark theme, if the "Auto" theme is selected.
    - Simple to implement, cheap to compute.
    - What if we don't have the user's current location?
    - What if the user is looking (say, to see what quests are available) at a position different than their current location? Use gps or screen/scroll position?
2. Lit quests appear if it is past sunset at their current location.
    - How it's currently implemented, I think.
    - Potential performance issues, or more complicated implemented to mitigate them.
    - Might be confusing (hard to figure out when a given lit quest will appear), particularly along with the "available in 3 hours" feature suggested below.
        - Might also be *cool* — seeing lit quests gradually appear as the sun crawls across the map. But would definitely work better if the map visualized the current sun line instead of switching all at once.

Personally I'm more inclined toward solution 1 and simply not worrying about making the app work in those edge cases, since they are not the primary use case of StreetComplete (in-person surveying).

westnordost: @smichel17, if you are going to comment on the code, you should read it first before making assumptions.

smichel17: In the diff, I came across this code:

https://github.com/streetcomplete/StreetComplete/blob/e5c6135d0ad1e30e11f6ad25e3ee5474228eeffd/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/DayNightQuestFilter.kt#L13-L19

Which seems to be applied per-quest. I admit I have not traced the call stack more than one step, to here:

https://github.com/streetcomplete/StreetComplete/blob/e5c6135d0ad1e30e11f6ad25e3ee5474228eeffd/app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt#L86-L87

to see if it is called differently, but I don't think my assumption — made *after* reading the code — was unreasonable.

---

It seems like I have missed something that you find obvious, but I'm honestly not sure what it is. Also happy to figure it out myself if you want to point me in the right direction.

westnordost: Alright, maybe a misunderstanding. The way you wrote the other comment sounded like you would comment on the functionality in general as if no code had been written yet.

westnordost: In any case, I'd also prefer 1, but I am not going to reject this PR just because 2 and not 1 has been implemented.

If @TurnrDev or you would like to move to behavior 1 instead, I can give advice how to do that on the technical side.

10992-osm:
> Would it be possible to tag (like `survey:atnight=yes` or something) to indicate that the answer was surveyed during night time?That way we could re-ask this for roads/paths/etc that are already tagged lit but weren't surveyed at night to verify that they are actually lit at night time, and didn't just look like they should be lit.

Could you not infer that from when the changeset was made?  (I suppose that might not work with delayed uploads, whether due to connection issues or deliberate choice on the part of the user).

TurnrDev:
I got in my head with imposter syndrome, I doubt this will be ready in time for v33

Gonna have to break all my to-do up into smaller tasks and focus on one at a time lol

TurnrDev:
To Do:

- [ ] Fix [comment](https://github.com/streetcomplete/StreetComplete/pull/2872#discussion_r632455825) on isNight https://github.com/streetcomplete/StreetComplete/blob/e5c6135d0ad1e30e11f6ad25e3ee5474228eeffd/app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt#L20-L24

- [ ] Cache the ifNight calculation? https://github.com/streetcomplete/StreetComplete/pull/2872#discussion_r632459615 @westnordost said he could potentially help with this.
  - [ ] At the very least, time the impact isNight is adding to performance

- [ ] Find out if timezone is needed https://github.com/streetcomplete/StreetComplete/pull/2872#discussion_r632462027

- [ ] Add countdown to sunset in quest list https://github.com/streetcomplete/StreetComplete/pull/2872#issuecomment-845928473

- [x] Cry

westnordost:
Note: If this is merged, the quest priority of lit-quests should go up, as during night, those should have higher priority (because they are easier to solve)

dbdean:
I was having a thought about this PR recently. What would the plan be for lit on sports pitches? I treat the current lit quest for them as 'can this sporting pitch be lit', rather than is it lit every night like I would treat a similar quest on roads and paths.

So (if I'm treating the sporting lit quest appropriately) the sporting lit quest is actually generally easier to answer in the day time as you can see the lighting poles, which might be harder to see at night if they aren't on that particular night.

TurnrDev:
> So (if I'm treating the sporting lit quest appropriately) the sporting lit quest is actually generally easier to answer in the day time as you can see the lighting poles, which might be harder to see at night if they aren't on that particular night.

Good point, and I agree with you there. I will make changes

app/src/main/java/de/westnordost/streetcomplete/util/CheckIfNight.kt
TurnrDev: I just looked, it's because `SolarEventCalculator` needs it

TurnrDev:
@westnordost As much as I'd like to cache the result, I'm not entirely sure on how to do that! I have changed the function to isDay to resolve a bug with regards to after midnight. I'm testing it soon :) 

I removed AddPitchLit from the PR, @dbdean raised a good point regarding pitches

I also movet the remaining lit quests to the top. Building now :)

TurnrDev:
Just tested in Alert, Nunavut, Canada which currently doesn't experience sundown. Worked fine.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
```diff
@@ -237,10 +240,10 @@ import javax.inject.Singleton
         AddBollardType(), // useful for first responders
         AddCameraType(),
 
-        // ↓ 8. defined in the wiki, but not really used by anyone yet. Just collected for
+        // ↓ 9. defined in the wiki, but not really used by anyone yet. Just collected for
         //      the sake of mapping it in case it makes sense later
         AddPitchSurface(),
-        AddPitchLit(),
+        AddPitchLit(), // Not affected by new DayNight cycle because the lights are usually only on during games
```
westnordost, line 246: I know this has been requested by someone before, but I don't completely follow that logic.

The lights are usually only on during games, but only if it is dark outside, right? So, that should not change the fact that whether it is lit or not will only be possible to see when it is dark.

westnordost (approved):

westnordost:
I'll also probably add a cache after merge, so the `isDay` will not be calculated for every single road.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
TurnrDev: True, but given how infrequent games are- It would be possible to pass by on the street when there isn't a game, see no light on and assume there aren't any.

TurnrDev:
> I'll also probably add a cache after merge, so the `isDay` will not be calculated for every single road.

I was thinking it might be better to calculate for user position, but I don't know what the course of action to work that out is.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
dbdean: And during the daytime, it will be easier to see if there are lights at all.

westnordost: Hm okay, if you agree with each other there..

smichel17:
I have a couple dev tasks I've volunteered to do, so I don't know about my bandwidth here, but besides that I'd be up to help with performance work here (caching and/or using the current map position instead of calculating for each quest individually).
…I'd be more exited about it if someone wanted to do the profiling exploration first (the most boring and most important part :P)

westnordost:
No, I am right now doing it myself but thank you for the offer

TurnrDev:
@westnordost Thank you for your improvements. I'm really not sure I would have ever figured most of that out

PR #2873, opened 2021-05-11
Team mode: show all split sections also in online-mode  (fixes #2868)

### LINKED ISSUE #2868: Splitting ways in team mode
After splitting a way in team mode, only some segments are shown for further quests.

Would it be possible to show all segments when auto-upload is disabled? 
Before uploading it shouldn't have a "real" id from which the team is determined, so that's why I think it could probably be done.

westnordost:
@Helium314  would you test this in the field?

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/created_elements/CreatedElementsDaoTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/created_elements/CreatedElementsDaoTest.kt
@@ -1,0 +1,37 @@
+package de.westnordost.streetcomplete.data.osm.created_elements
+
+import de.westnordost.streetcomplete.data.ApplicationDbTestCase
+import de.westnordost.streetcomplete.data.osm.mapdata.ElementKey
+import de.westnordost.streetcomplete.data.osm.mapdata.ElementType
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Test
+
+class CreatedElementsDaoTest : ApplicationDbTestCase() {
+    private lateinit var dao: CreatedElementsDao
+
+    @Before fun createDao() {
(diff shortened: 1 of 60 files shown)


Helium314:
Great, thanks!
I'll give it a try, but a real field test probably needs to wait until Saturday

FloEdelmann (approved):
The code looks fine, but I haven't field-tested it.

Why is the roof shape icons commit [`df660c4`](https://github.com/streetcomplete/StreetComplete/pull/2873/commits/df660c47b25caab26763b160a511b693a7b5b8a4) in here, too? It's already in the master branch (2e2d32e4e00a5ce871eec2e999db58731723e3e9) and makes the diff quite large (but easy to ignore).

westnordost:
> Why is the roof shape icons commit df660c4 in here, too

Yeah sorry about that. I committed that to the wrong branch, then cherry picket it into master too

Helium314:
Real field test will need to wait until tomorrow, but I did some small test:

* activate team mode (2 phones)
* enable auto-upload and have internet connection active on both
* split way into 2 parts (shown on A only)
  * now both segments are shown on A
  * manual download on B -> one segment of the split way shows up
  * exit team mode on B, enter again -> still both parts show
* split way into 3 parts (shown on A only)
  * all 3 segments are shown on A
  * manual download on B -> one segment of the split way shows up
  * exit team mode -> all parts shown (on both phones)
  * enter team mode -> same as before (all 3 parts shown on A, one part on B)
  * enter team mode, but with A and B switched -> old A shows 2 parts (the one with original id is hidden), old B shows opposite parts

So:
* all newly created segments are shown
* new segments are also shown on other devices if id allows (and new quests are downloaded)
* exit and re-enter team mode does not change which quests are shown

westnordost:
That's great! This is enough of a test so that it can confidently be merged

PR #2875, opened 2021-05-12
put add housenumber quest together with add street name quest

matkoniecz:
recently I keep running into problem of finding a quest among all of them

it fixes one of problems that I have with

opening PR to check is such reorganization in general and specifically about this one welcomed - and also because there is (tiny) chance that I broke something, so I want to not add it on what is in the next release (though I tested the change)


westnordost:
ok

westnordost (approved):

PR #2892, opened 2021-05-19
Minor improvements to changelog wording

peternewman:


app/src/main/res/raw/changelog.yml
```diff
@@ -4,14 +4,14 @@ v32.0-beta1: |
...
-  to do this. Note that this required to drop most local data, such as unsynced edits, downloaded
+  to do this. Note that this required dropping most local data, such as unsynced edits, downloaded
   quests and statistics.<br/>Special thanks to @FloEdelmann for helping with the refactor.
   </p>
   <h3>🔙️ Superpowered Undo</h3>
   <ul>
   <li>
-    You can now <b>undo any edit</b> that hasn't been uploaded yet: <br/>
-    Quest answers, split ways, deleted places, hid quests, created notes and note comments (#1029, #2616)
+    You can now <b>undo any edit</b> that hasn't been uploaded yet:<br/>
```
peternewman, line 13: Given this and 
> You can now undo deletions of places even after you have already uploaded it

Which of those is true? Or can we not undo splitting, hiding quests, creating and commenting on notes after upload?

matkoniecz (approved):
looks OK to me and I guess that it makes sense to change it as many will see it with a regular update

but I want to give chance to look at this for people more competent in English

peternewman:
> looks OK to me and I guess that it makes sense to change it as many will see it with a regular update

I'm assuming the changelog for the final release will be based on this too, so it will feed into that.

> but I want to give chance to look at this for people more competent in English

Fair enough, makes sense.

smichel17:
This is in reply to the file discussion, app is not letting me comment there for some reason :/

Will not wordsmith from phone, but I would do the order like so:

## Superpowered Undo

- view edit history, undo *any* edit in any order..
- ..except a few which can only be undone before upload: x, y, z
- ..You _can_ undo deletions, though, even if you already uploaded!
 - The
- other
- bullets

PR #2898, opened 2021-05-23
Add bus stop bin quest

### LINKED ISSUE #2760: New quest: tag bin=* at bus stops
### General
Add tagging to indicate if a bus stop has a waste basket: [bin=*](https://wiki.openstreetmap.org/wiki/Key:bin)

Affected tag(s) to be modified/added: bin=* on nodes where highway=bus_stop and bin=* is not set
Question asked: **Does this bus stop have a waste basket?**
Valid responses: Yes | No

### Checklist

### Ideas for implementation

**Element selection:** 

**Metadata needed:** 

**Proposed GUI:**

FloEdelmann:
Closes #2760.

The code is basically the same as `AddBenchStatusOnBusStop`.


westnordost (approved):
Looks good!

For the icon, I first thought this is an ice cream. Maybe remove the colorful litter on top?

FloEdelmann:
I changed the icon:

arrival-spring:
There don't seem to be any checks for a node nearby tagged with `amenity=waste_basket`
If there was a bus stop with a bin right next to it then it could be that the bin is already mapped separately, so the information should not be repeated on the bus stop object.

(Although basically the same thing could be said of `AddBenchStatusOnBusStop`)

westnordost:
As far as I know, that's ok. I.e. it is the same as with bus stop benches.

matkoniecz:
>  If there was a bus stop with a bin right next to it then it could be that the bin is already mapped separately, so the information should not be repeated on the bus stop object.

`bin=yes` means "this bus stop has nearby waste basket" not "waste basket is here"

The same for `cycleway=separate` and `sidewalk=separate`, similarly to `lit=yes`

peternewman:
> > If there was a bus stop with a bin right next to it then it could be that the bin is already mapped separately, so the information should not be repeated on the bus stop object.
> 
> `bin=yes` means "this bus stop has nearby waste basket" not "waste basket is here"

Surely if it's not actually the bus stop's bin, but an adjacent one, it should be mapped independently:
https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element

> The same for `cycleway=separate` and `sidewalk=separate`, similarly to `lit=yes`

The wiki (now) features separate for bench and shelter, doing the same for bin doesn't seem a very big stretch:
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_stop#Physical_attributes

I'd say lit is a bit more vague as it's variable, the rest are very binary, hence the lit discussion here:

PR #2914, opened 2021-05-27
"does not exist" enabled for more quests, fixes #2913

### LINKED ISSUE #2913: Fire hydrant type quest has no option for "does not exist"
Some quests, like "does this bench have a backrest?" and "what can be recycled here?" have an option to indicate that the feature doesn't exist. There is no such option for the quest "what kind of fire hydrant is this?" I found myself wanting this option a couple times recently, and had to resort to using another editor to remove the features.

matkoniecz:
I thought about adding it for bus stop quests, but `highway=bus_stop` is fully valid also for unremarkable piece of dirt where people enter into bus - even if not marked in any way whatsoever. So extra step for deletion is useful here.

workaround for github bug: fixes #2913


FloEdelmann (approved):

westnordost (approved):

peternewman:
> OK, reverted. Not entirely sure but it was intended to make obvious changes.

Is it worth adding a comment in that file to "document" that consideration?

PR #2915, opened 2021-05-27
Add bollard type quest

### LINKED ISSUE #2128: 'Type of bollard?'-Quest for first responders
This information is quite useful for routing first responders vehicles, like police, ambulances and fire brigades to a destination. 

### General
Affected tag(s) to be modified/added: [bollard=*](https://wiki.openstreetmap.org/wiki/Key:bollard)
Question asked: **What type of bollard is this?**

### Checklist

FloEdelmann:
Closes #2128.

<details>
<summary>Older screenshots</summary>
<table>
<tr>
<th>Version 1</th>
<th>Version 2</th>
</tr>
<tr>
<td></td>
<td>
</td>
</tr>
</table>
</details>

---

@RubenKelevra Please note that I didn't include any cycle barriers as I wasn't sure what you meant. See the bottom part of https://github.com/streetcomplete/StreetComplete/issues/2128#issuecomment-846586181.


matkoniecz:
Thanks for this pull request!


-----------

I would be likely confused by that and need some time time to realize that "Fixed" is "standard, normal, regular bollard".

Though I have no ideas how to word it better :(

-----------

It is put among quests with 

> useful data that is used by some data consumers

priority. Is it actually used already by anyone?

it seems to rather fit

> 8. defined in the wiki, but not really used by anyone yet.

------------

Thanks for excluding free-floating nodes.

Areas like https://www.openstreetmap.org/node/3122426162#map=19/52.22561/20.99315 would be !!interesting!!

screen04

I would also commit original images into repo if you used crops/reduced quality etc (or at least upload them to Wikimedia Commons). See say `./res/graphics/kerbs/surface_paving_stones_original.jpg`

RubenKelevra:
Really cool! 😎

FloEdelmann:
> I would be likely confused by that and need some time time to realize that "Fixed" is "standard, normal, regular bollard".

Maybe "Permanent" or "Immovable"?

> it seems to rather fit "defined in the wiki, but not really used by anyone yet."

I think it's better suited in the "data useful for only a specific use case" category. I changed it in [`02dd641`](https://github.com/streetcomplete/StreetComplete/pull/2915/commits/02dd64103490640c5cf6b0344730d197a6e3ad8b).

> I would also commit original images into repo if you used crops/reduced quality etc (or at least upload them to Wikimedia Commons).

I uploaded them:

* https://commons.wikimedia.org/wiki/File:Two_Bollards.jpg
* https://commons.wikimedia.org/wiki/File:Two_bollards_removable_without_key.jpg
* https://commons.wikimedia.org/wiki/File:Bollard_removable_without_key_(detail).jpg
* https://commons.wikimedia.org/wiki/File:Bollard_removable_with_key.jpg
* https://commons.wikimedia.org/wiki/File:Flexible_bollard.jpg

matkoniecz:
> Maybe "Permanent" or "Immovable"?

"Usual permanent bollard"?

Not sure is it more clear

FloEdelmann:
Or better "Fixed (not removable)" to avoid confusion whether it is "not removable and not fixed" or "not removable, i.e. fixed".

matkoniecz:
Maybe change also "Flexible" to "Flexible (no removable)"? (not sure here)

goldfndr:
I suggest "constant".

westnordost:
```typescript
const bollard = "fixed"
```

app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/AddBollardType.kt
```diff
@@ -0,0 +1,44 @@
...
+    override val isDeleteElementEnabled = true
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_bollard_type_title
+
+    // exclude free-floating nodes
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
+        val wayNodeIds = mutableSetOf<Long>()
+        mapData.ways
+            .filter { it.tags["area"] != "yes" }
+            .flatMapTo(wayNodeIds) { it.nodeIds }
```
westnordost, line 30: I wonder if for this here and other quests that work similarly, it would make a noticeable performance difference if this was used instead:

```suggestion
        mapData.ways.asSequence()
            .filter { it.tags["area"] != "yes" }
            .flatMapTo(wayNodeIds) { it.nodeIds }
```
Because the result of the `.filter` will be a list of almost all ways in the map data. I.e. almost the whole list of ways is copied. On the other hand, I don't know how big is the overhead of using `asSequence`

Care to do some performance testing?

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/AddBollardType.kt
westnordost: (and if it is a noticeable performance difference, maybe it would make sense if `ways` actually returned a sequence already? Or maybe even that `MapData` didn't implement `Iterable` but `Sequence`?)

FloEdelmann: > Care to do some performance testing?

How would I do it?

westnordost: You could just write a test or a small kotlin program (with `fun main()`) and simulate the situation, measure the time with System.currentTimeMillis().

smichel17: This can also be convenient, sometimes: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/measure-time-millis.html

FloEdelmann: I'm not sure how to simulate the situation in a test or script, because I need real map data there.

I did a small comparison of the processing times of the [OsmQuestController](https://github.com/streetcomplete/StreetComplete/blob/da5cb321546bace9ba39910bad6977051d01c6d1/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt#L145-L156) though (exact same location, always already filled database):

* without `asSequence`:

  ```
  AddBollardType: Found 0 quests in 23ms
  AddBollardType: Found 0 quests in 15ms
  AddBollardType: Found 0 quests in 28ms
  AddBollardType: Found 0 quests in 20ms
  AddBollardType: Found 0 quests in 30ms
  ```

  average: 23.2ms

* with `asSequence`:
  ```
  AddBollardType: Found 0 quests in 53ms
  AddBollardType: Found 0 quests in 29ms
  AddBollardType: Found 0 quests in 23ms
  AddBollardType: Found 0 quests in 21ms
  AddBollardType: Found 0 quests in 17ms
  ```

  average: 28.6ms

RubenKelevra:
> > I would be likely confused by that and need some time time to realize that "Fixed" is "standard, normal, regular bollard".
> 
> Maybe "Permanent" or "Immovable"?

How about `solid`?

app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/AddBollardType.kt
westnordost: Interesting, so the sequence seems to have a considerable overhead and/or copying lists is really fast.

FloEdelmann:
The foldable and removable bollards are also "solid".

PR #2918, opened 2021-05-27
Linkify URLs in note quests

### LINKED ISSUE #2905: Links in notes should be clickable
It would be nice if links in notes were clickable.

I noticed this as I have it set to show all notes and when I added a new note with a photo the link was not clickable.
On that note, it does seem a bit odd to add a note through the app and have it appear as a quest to be solved, but having read the comments where this change was discussed, that does seem to be the desired behaviour when the "show all notes" option is on.

The text in the note popup is selectable and then Android seems to recognise that it's a url and auto selects the whole thing on a long press and adds a tooltip option to open the link, but this isn't very intuitive.

FloEdelmann:
Closes #2905.

See https://developer.android.com/reference/android/widget/TextView#attr_android:autoLink

### DIFF
--- a/app/src/main/res/layout/quest_note_discussion_item.xml
+++ b/app/src/main/res/layout/quest_note_discussion_item.xml
@@ -54,10 +54,11 @@
 
                 <TextView
                     android:id="@+id/commentText"
+                    style="@style/NoteDiscussion.Comment"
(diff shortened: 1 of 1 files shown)


matkoniecz:
Heh, during testing I found defunct note fixed already by @westnordost

PR #2919, opened 2021-05-27
Change bikeway references to cycleway

TurnrDev:
No real reason other than it bothered me that it was bikeway 🤷🏻


TurnrDev:
tested at 8a66df4b9e3c93eb5ee53334c8670e3aaf848fd8 - builds

westnordost:
Why, is bikeway not a word?

TurnrDev:
Never heard it before today, but it might be. I saw the two terms used interchangeably in the same module and just found it a little confusing so thought it would make sense to fit the key?

westnordost:
Okay 🤷

westnordost (approved):

PR #2920, opened 2021-05-27
Add "woodchips" path surface option

### LINKED ISSUE #2889: Woodchip path surface
**Use case**
For the quest "What surface does this path have here?", there is not currently an option for woodchip.  The most recent case of this was in [note 2672366](https://osm.org/note/2672366), the picture of which I attach here, in case it is deleted (as I understand note images are).  I have at least once had to leave a note to myself so I could later tag it with `surface=woodchips`, which is the most used tagging ([taginfo](https://taginfo.openstreetmap.org/keys/surface#values) 3569 (the others being `surface=`: `wood_chips` (12), `wood_chippings` (8), `wood_mulch` (6), `wood_chip` (6), `wooden_chips` (3), `wood_shavings` (2), `Bark_wood_Chip` (1); possibly `bark_wood` (9), `wooden_chaff` (3)).  There is currently no documentation on the wiki for this tag.

Other case in [note 2629714](https://osm.org/note/2629714).

Image of woodchips from note 2672366

**Proposed Solution**
There should, in my view, be an option to choose woodchips, which is not currently present.    It is likely only wanted for paths and tracks, rather than any other ways.  The path should then be tagged with `surface=woodchips`.

FloEdelmann:
Closes #2889.


matkoniecz:
Are there sport fields (`leisure=pitch`) with woodchips as surface? http://overpass-turbo.eu/s/17No says no.

FloEdelmann:
I don't know any. A quick Ecosia and Google search also didn't reveal any. There are just many playgrounds with woodchips as fall protection surface. 

Maybe there are also some ropes courses tagged with [`sport=climbing_adventure`](https://wiki.openstreetmap.org/wiki/Tag:sport%3Dclimbing_adventure), but I guess adding a surface there doesn't make much sense anyway.

matkoniecz:
Thanks for implementing this!

FloEdelmann:
Yay we're below 100 open issues again :tada:

westnordost:
Btw, @matkoniecz , this should not have been merged. The policy is to only include changes in minor updates that are either bugfixes or are small enhancements that contain no new strings. This change was a small enhancement, but added a new string.

This is why I mark some PRs as approved instead of merging them immediately.

matkoniecz:
Sorry! So pull requests adding new strings or making major changes should be just marked approved, and you will merge all of them at the correct time?

westnordost:
Correct

Am 29. Mai 2021 07:52:44 MESZ schrieb Mateusz Konieczny ***@***.***>:
>Sorry! So pull requests adding new strings or making major changes
>should be just marked approved, and you will merge all of them at the
>correct time?
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/2920#issuecomment-850776177

-- 

PR #2583, opened 2021-02-16
Add postbox royal cypher quest (new)

arrival-spring:
Making a new pull request as I wasn't confident fixing the extra commits on the other one (I clearly need to brush up on git).

Almost there now I think, just a couple of things from my side

I couldn't find a source for http://www.bechler.free-online.co.uk/gc/RoyalCyphers2.jpg that said anything about copyright.

Also I'm not quite happy with saying "Just a crown" and it being tagged with "scottish_crown". "crown" has only been used twice per taginfo, so we shouldn't introduce that. Maybe only show that option in Scotland?

Fixes #2563

PR #2922, opened 2021-05-27
Expand cypher locations

TurnrDev:
Roughly based on the list of locations of [pillar boxes on wikipedia](https://en.wikipedia.org/wiki/Pillar_box) crossed with google image searches.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/AddPostboxRoyalCypher.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/postbox_royal_cypher/AddPostboxRoyalCypher.kt
@@ -16,7 +16,10 @@ class AddPostboxRoyalCypher : OsmFilterQuestType<PostboxRoyalCypher>() {
     override val wikiLink = "Key:royal_cypher"
     override val icon = R.drawable.ic_quest_crown
     override val enabledInCountries = NoCountriesExcept(
-        "GB"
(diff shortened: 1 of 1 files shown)


westnordost:
Cool!

arrival-spring:
Do any of these places have different cyphers?
e.g. [#2583 (comment)](https://github.com/streetcomplete/StreetComplete/pull/2583#issuecomment-792274451)

And maybe would some be better to have the images in different colours?

TurnrDev:
Some excellent points raised

westnordost:
Ok so should I revert that merge until that is sorted out?

TurnrDev:
My assumption is that most of these places will have the same set, or less, than the UK set. Maybe we can get an SVG for the Irish one shown [here](https://commons.wikimedia.org/wiki/File:Irish_post_box.jpg)?

TurnrDev:
> The Department of Posts and Telegraphs continued installing similar pillar boxes and wall boxes, but with the initials SÉ (for Saorstát Éireann), a harp or the Department of Posts and Telegraphs P ⁊ T logo, instead of a monarchical cypher.

[source](https://en.wikipedia.org/wiki/Pillar_box#Ireland)


Strictly speaking, these extra markings on Irish boxes aren't royal cyphers, but rather just the initials of the state at the time.

westnordost:
Strictly speaking yes, but what counts is how they are tagged.

PR #2925, opened 2021-05-28
Remember more building level quest answers

### LINKED ISSUE #1772: Make it faster to answer the building levels quest
### Use case

Building levels is a very common quest in residential neighborhoods, that could be faster.

### Proposed Solution

Note: this is a rough mockup -- eg, I was just copy-pasting and the rows are not quite aligned.

mockup-grid

Pressing *any* button switches to the current quest view.

- For the 6 on the left, the behavior is the same as the current one: the fields would be pre-filled and the keyboard is not opened.

mockup-grid-preselect

- For the blank one on the right, the keyboard would be opened like this:
    - This keeps the number of taps for custom building height the same as it currently is.

mockup-grid-select-blank

### Discussion

The slowest part of this quest is when you have to open the keyboard, since that introduces a delay. This solution gives more pre-fill options so you don't have to open the keyboard as often. You still go through the regular entry screen, so new mappers will still see the text, "regular levels (not counting the roof)" and "levels in the roof".

My original idea was to use have the same behavior as above, but show recently used heights instead of a static grid. I opted for this instead because it will be better (consistent positions) in rural, residential environments, where this quest is common. In urban environments with tall buildings, buildings are more varied and also larger, so this quest appears less frequently and recent options are less likely to be useful.

I think there may be room for improvement. For example, we could swap the axes of regular vs roof levels and allow scrolling, if 3 regular levels is not enough for common use cases.

FloEdelmann:
Closes #1772.

### Screenshots

<table>
<tr>
	<th>First time opening the quest:<br>Unchanged interface.
	<th>One saved answer:<br>The button position changed.
	<th>Two saved answers:<br>The latest one is on the left.
<tr>
	<td>
	<td>
	<td>
<tr>
	<td colspan="3">
<tr>
	<th>After selecting a saved answer:<br>The buttons remain there.
	<th colspan="2">When there is too little space:<br>The buttons can scroll horizontally. There are at most 5 saved answers.
<tr>
	<td>
	<td>
    <td>
</table>

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
@@ -1,9 +1,12 @@
 package de.westnordost.streetcomplete.quests.building_levels
 
 import android.os.Bundle
+import android.view.LayoutInflater
 import androidx.appcompat.app.AlertDialog
 import android.view.View
-import androidx.core.view.isGone
+import android.view.ViewGroup
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
 
 import javax.inject.Inject
 
@@ -15,8 +18,6 @@ import de.westnordost.streetcomplete.quests.OtherAnswer
 import de.westnordost.streetcomplete.util.TextChangedWatcher
 
 import kotlinx.android.synthetic.main.quest_building_levels.*
-import kotlinx.android.synthetic.main.quest_building_levels_last_picked_button.lastLevelsLabel
-import kotlinx.android.synthetic.main.quest_building_levels_last_picked_button.lastRoofLevelsLabel
 
 class AddBuildingLevelsForm : AbstractQuestFormAnswerFragment<BuildingLevelsAnswer>() {
 
@@ -29,6 +30,8 @@ class AddBuildingLevelsForm : AbstractQuestFormAnswerFragment<BuildingLevelsAnsw
     private val levels get() = levelsInput?.text?.toString().orEmpty().trim()
     private val roofLevels get() = roofLevelsInput?.text?.toString().orEmpty().trim()
 
+    private val lastPickedStrings by lazy { favs.get(javaClass.simpleName) }
+
     @Inject internal lateinit var favs: LastPickedValuesStore<String>
 
     init {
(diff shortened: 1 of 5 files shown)


westnordost:
Is the form always expanded already? Or do you have to pull it up?

matkoniecz:
Not tested yet, but I really like this idea.

mnalis:
Looks good to me, is there perhaps an `.apk` to try out?

smichel17:
Screenshots look good to me, my only concern is what happens when the OK button is visible and there are a lot of options.

I haven't tried it out yet but here's a debug apk: https://smichel.me/files/streetcomplete/streetcomplete-debug-09629c77.apk

FloEdelmann:
@westnordost The form is always expanded already.

mnalis:
@smichel17 debug version feels pretty nice! 

Only issue I see, is that on my test phone ([Xiaomi Redmi Note 3 Pro](https://www.gsmarena.com/xiaomi_redmi_note_3-7863.php)) there seems to be missing little space to show all 6 single-digit (only 5 and a half fit on screen). 

Purely cosmetic, but maybe icons can be squeezed a little so all 6 last buildings fit on screen (if they are all single-digit levels)? Or if there is not enough space for that, maybe even drop to showing last 5 buildings?

levels1

westnordost:
I think 5 would be enough

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/BuildingLevelsAnswer.kt
```diff
@@ -1,3 +1,3 @@
 package de.westnordost.streetcomplete.quests.building_levels
 
-data class BuildingLevelsAnswer(val levels:Int, val roofLevels:Int?)
+data class BuildingLevelsAnswer(val levels: String, val roofLevels: String?)
```
westnordost, line 3: It would be cleaner if the data structure for the answer remained to be an integer.

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -80,4 +74,47 @@ class AddBuildingLevelsForm : AbstractQuestFormAnswerFragment<BuildingLevelsAnsw
...
+            val lastLevelsLabel: TextView = view.findViewById(R.id.lastLevelsLabel)
+            val lastRoofLevelsLabel: TextView = view.findViewById(R.id.lastRoofLevelsLabel)
+
+            init {
+                itemView.setOnClickListener(this)
+            }
+
+            override fun onClick(view: View) {
+                onItemClicked(bindingAdapterPosition)
+            }
```
westnordost, line 97: Why not simply something like

`itemView.setOnClickListener { _ -> onItemclicked(bindingAdapterPosition) }`

?

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/BuildingLevelsAnswer.kt
FloEdelmann: I noticed during testing that it'll crash when a large number (bigger than MAX INT) is entered, because `string.toInt()` throws a `NumberFormatException` then.

Now that we don't convert the string to an integer, just to convert it back to a string when uploading the changes, the issue can't occur anymore.

I agree that `Int` is nicer in the data class though. Should we continue to ignore the crash?

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
FloEdelmann: Old Java habits...

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/BuildingLevelsAnswer.kt
westnordost: Let's ignore it

matkoniecz:
I tested it today and it works great! 

Maybe making clear that this things at the bottom are clickable buttons may be nice. Borders?

PR #2926, opened 2021-05-28
Use the new Activity Result API in the AttachPhotoFragment

### LINKED ISSUE #2911: Notify user of not working camera app when adding trying to add a photo in note
I want to take a photo at a note. When I press the button "Adding a photo..." nothing happens. Notes without photos can be created. 

**How to Reproduce**
Create new note --> Adding a photo...

**Versions affected**
StreetComplete 31.3 (F-Droid)
Android Version 11 (GrapheneOS RQ2A.210505.003.2021.05.19.06)
Phone: Pixel 4a (5G)
All available permissions have been granted.

tapetis:
This PR replaces the usage of the deprecated `startActivityForResult` method in the `AttachPhotoFragment` with the new [Activity Results API](https://developer.android.com/training/basics/intents/result) introduced in AndroidX. In addition, it fixes #2911 by displaying a `Toast` when no external camera app can be found. I tested both changes on my device and everything works fine.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/AttachPhotoFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/AttachPhotoFragment.kt
@@ -11,7 +11,6 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
-import android.widget.Toast
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.core.content.FileProvider
 import androidx.core.net.toUri
@@ -109,7 +108,7 @@ class AttachPhotoFragment : Fragment() {
             takePhoto.launch(photoUri)
         } catch (e: ActivityNotFoundException) {
             Log.e(TAG, "Could not find a camera app", e)
-            context?.toast(R.string.no_camera_app, Toast.LENGTH_LONG)
+            context?.toast(R.string.no_camera_app)
         } catch (e: IOException) {
             Log.e(TAG, "Unable to create file for photo", e)
             context?.toast(R.string.quest_leave_new_note_create_image_error)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/AttachPhotoFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/AttachPhotoFragment.kt
@@ -1,18 +1,18 @@
 package de.westnordost.streetcomplete.quests.note_discussion
 
-import android.app.Activity.RESULT_OK
-import android.content.Intent
+import android.content.ActivityNotFoundException
 import android.content.pm.PackageManager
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -472,6 +472,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="about_category_feedback">Feedback</string>
     <string name="quest_cycleway_value_sidewalk">explicitly shared sidewalk</string>
     <string name="quest_leave_new_note_create_image_error">Unable to create file for photo</string>
+    <string name="no_camera_app">"No external camera app available. Is your pre-installed camera app disabled?"</string>
```
westnordost, line 475: This is a little bit long for a toast. I'd suggest to just write "No camera app installed"

westnordost (commented):
Nice!

I didn't read through the new API to capture a photo but I trust you tested your change?

app/src/main/res/values/strings.xml
tapetis: I took the idea of a longer explanation from https://github.com/streetcomplete/StreetComplete/issues/2911#issuecomment-850585869. Using a shorter sentence is of course also possible.

matkoniecz: > "No camera app installed"

In #2911 camera app was installed but was not cooperating. If shortening is necessary I would switch to "No camera app available"

tapetis: It would also be possible to replace the `Toast` with a dialog.

westnordost: I'd be for "No camera app available"

app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/AttachPhotoFragment.kt
```diff
@@ -42,6 +41,8 @@ class AttachPhotoFragment : Fragment() {
 
     private lateinit var noteImageAdapter: NoteImageAdapter
 
+    private val takePhoto = registerForActivityResult(ActivityResultContracts.TakePicture(), ::onTakePhotoResult)
```
westnordost, line 44: You tested that, right? I wonder if/how this is possible, because on creation, the fragment is not attached to any Activity yet, but you need the activity to register an activity result contract to, no?

westnordost: Yeah ok, seems to work like that

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/note_discussion/AttachPhotoFragment.kt
tapetis: Yes, I tested the change. I even enabled the "don't keep activities" developer option.

PR #2935, opened 2021-05-31
ask for building:levels where height is known

### LINKED ISSUE #2927: consider asking about building levels when `height` is tagged
original issue title: consider asking about roof shape when height is tagged

**Use case**
1) building levels is not asked if `height` is tagged - https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
2) roof shape is not asked unless building levels are tagged - https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/roof_shape/AddRoofShape.kt
3) as result one of imports is planning to deliberately skip importing available height to avoid disabling roof shape quests. This seems to not be a good way of doing things.

**Proposed Solution**
Ideally SC would not discourage using `height` tag. Maybe just ask for roof shape if `height` is tagged on a building, in area where roofsAreUsuallyFlatAt is not true?

triggered by https://discord.com/channels/413070382636072960/789437210797408277/847899428483629076

matkoniecz:
fixes #2927

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
@@ -6,11 +6,9 @@ import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChanges
 
 class AddBuildingLevels : OsmFilterQuestType<BuildingLevelsAnswer>() {
 
-    // building:height is undocumented, but used the same way as height and currently over 50k times
     override val elementFilter = """
         ways, relations with building ~ ${BUILDINGS_WITH_LEVELS.joinToString("|")}
-         and !building:levels and !height and !building:height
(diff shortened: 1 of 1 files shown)


matkoniecz:
For justification: there are buildings with 3 levels higher than buildings with 4 levels (old style tenement houses vs minimum cost optimization housing), it can help spotting invalid `height` data.

Also, actually useful and used in 3D rendering - is necessary to get info about whether there are building levels in roof (unavailable from height alone) and for rendering floor count. Yes, "for 3D rendering" is not a good enough reason by itself but if actually used already it matters a bit.

FloEdelmann:
@matkoniecz I think you force-pushed the wrong branch. This PR now includes my commits from #2925.

matkoniecz:
@FloEdelmann Yep, fixed. Thanks.

Issue #2194, opened 2020-10-24
Improve data for AtmOperators (and probably other equivalent ones)


peternewman:
**Use case**
There is currently a lot of duplication within AtmOperators. Some of which some rudimentary de-duplication would fix:

```
  - Note Machine # 67
  - NoteMachine # 26
  - Notemachine # 7
```

Also:
```
 - Tesco # 158
  - tesco # 6
```

See also Coop in here (but more complicated):
https://github.com/westnordost/StreetComplete/blob/0a1a722fece9b8706915806ebfebd5cd901e8b5d/res/country_metadata/atmOperators.yml#L1753-L1844

I've not looked or answered those quests yet, but I suspect charging stations and clothes containers would benefit from the same stuff too.

**Proposed Solution**
Simple short-term fix, take each returned name, lower case it and remove any punctuation ([\s,.:;-_], then group by that de-duplicated key. Return the most popular entry within each de-duplicated key. So in my two examples above `Note Machine` and `Tesco`. Optionally bump the count (if used) to be the merged total.

Longer term, switch to https://github.com/osmlab/name-suggestion-index/ when available, e.g. https://github.com/osmlab/name-suggestion-index/issues/2883 for ATMs.

There are already charging stations:
https://nsi.guide/index.html?t=brands&k=amenity&v=charging_station

I'm sure they'd be happy to add recycling container operators too.

westnordost:
StreetComplete just counts the data as it is currently used and sorts it by usage. If you want certain spelling variants to establish, then you have to do this yourself. Long term solution might be NSI, yes. 

> Simple short-term fix, take each returned name, lower case it and remove any punctuation ([\s,.:;-_],

eh no, StreetComplete should not just modify the operator names to its own whim. Who is to say wether it should be "Post", "Deutsche Post" or "Deutsche Post AG"? Not StreetComplete.

westnordost:
So, StreetComplete does only expedite harmonization of this tags here insofar as that often-used values are suggested, but nothing more.

matkoniecz:
Ones that are obvious and appearing just few times are easy to fix - for tesco see on a computer https://overpass-turbo.eu/s/Zmo (overpass turbo has poor mobile interface).

Select run button to run query, select export to export objects to JOSM/level0

Note https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct - manual fixing 6 tesco should be OK, but for NoteMachine I would consult with a local community.

[...]

PR #2941, opened 2021-06-02
Include `building=unclassified` in building type quest

FloEdelmann:
It's used 3425 times and is deprecated: https://taginfo.openstreetmap.org/tags/?key=building&value=unclassified#overview

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
@@ -10,7 +10,7 @@ class AddBuildingType : OsmFilterQuestType<BuildingType>() {
     // information about the purpose of the building, so no need to force asking it
     // same goes (more or less) for tourism, amenity, leisure. See #1854, #1891
     override val elementFilter = """
-        ways, relations with building = yes
+        ways, relations with (building = yes or building = unclassified)
          and !man_made
          and !historic
          and !military


matkoniecz:
With 3245 and a clear duplicate of `yes` I think that retagging values may be smarter.

Total building count is 457 882 386.

westnordost:
What do you mean with retagging might be smarter?

matkoniecz:
> What do you mean with retagging might be smarter?

In general I think that with rare invalid values better strategy would be to fix it in data (or add to JOSM/Osmose validator) rather than starting to handle it in SC.

There are many, many, many such values in OSM. 

For building values: `building=building` - https://taginfo.openstreetmap.org/tags/building=building#chronology

Operator tags have 10+ variants of `operator=Euronet` from what I remember (and I already eliminated some of them).

peternewman:
> For building values: `building=building` - https://taginfo.openstreetmap.org/tags/building=building#chronology

But if you look at that chronology, people are clearly doing that, and the problem is re-appearing, which either means there is a source issue causing it (perhaps a specific app with a problematic preset), or it's people entering free text, in which case aside from hinting to them that tagging `foo=foo` is probably wrong (is that always the case, or is there an edge case key where it's used) you won't ever stop the new ones appearing. Unless someone sets up a bot to fix it monthly, but then you'll still delaying an onsite fix with SC compared to just handling it in the app.

Even unclassified appears to have had one attempt at cleanup in the past:
https://taginfo.openstreetmap.org/tags/building=unclassified#chronology

> Operator tags have 10+ variants of `operator=Euronet` from what I remember (and I already eliminated some of them in OSM data).

That sounds like a duplicate of #2194 if we can improve the data going in, then we reduce the amount of cleanup required.

PR #2952, opened 2021-06-06
Fix a suspected Achievements bug

peternewman:
That was the only place in the codebase that postbox appeared with that case formatting


peternewman:
CC @arrival-spring 

I've no proof, but I noticed the oddity when working on #2754

westnordost:
👍

peternewman:
Thanks, I assume it works! I don't know if we can trap such errors somehow. Could we do a class.name sort of thing?

westnordost:
that would be possible

PR #2965, opened 2021-06-16
add building=shop to buildings with address

matkoniecz:
synonym of building=retail, not very popular but within top100


matkoniecz:
https://taginfo.openstreetmap.org/tags/building=shop

triggered by recent discovery that some buildings nearby are still missing house numbers

not committing directly as it is kind of not very popular, less that 10 000 uses worldwide (but within top 100 building values...)

PR #2986, opened 2021-06-20
Improve resurveying of postbox collection times

### LINKED ISSUE #2105: When resurveying postbox collection times show original times first
**Use case**
To add a collection time during resurvey, you have to tap as follows:
Add collection times
Confirm Mo-Fr
Tap Hour
Tap Minute
Tap Sa
Tap Hour
Tap Minute

This is quite a lot of taps if nothing has changed, compared to say surface or shelter quests.

If https://github.com/westnordost/StreetComplete/issues/1776#issuecomment-619662195 went in, this would reduce it by one tap, but it's still quite a lot.

**Proposed Solution**
Show the original time first, like opening hours and allow you to accept to change.

This seems to match what was suggested on the interim report:
https://wiki.openstreetmap.org/wiki/Microgrants/Microgrants_2020/Proposal/Map_Maintenance_with_StreetComplete/Interim_Report#Detail_Report

In the UK a lot of postboxes have moved to Mo-Fr 09:00, Sa 07:00, so there are changes, but a lot may already be on that, or haven't changed.

eginhard:
This adds a yes/no form to ask whether the previous collection times are still correct like for opening hours and is based on that implementation. Fixes #2105 (sparked by #2963).

The tests in https://github.com/streetcomplete/StreetComplete/blob/master/app/src/test/java/de/westnordost/streetcomplete/quests/AddPostboxCollectionTimesTest.kt are broken because I don't understand the format there in order to upgrade them, e.g. what is `WeekdaysTimes(weekdays(0b10000000), mutableListOf(60))`? In general, I'm not really familiar with Kotlin yet, so let me know if anything can be simplified.

One issue so far is that the resurvey form only displays collection times separated by commas (`Mo-Fr 18:15, Su 10:30`), not those separated by semicolons (`Mo-Fr 18:15; Su 10:30`).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
@@ -7,7 +7,7 @@ import java.util.Locale
 
 /** A time range from [start,end). The times are specified in minutes. */
 @Serializable
-data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = false) : Comparable<TimeRange> {
+data class TimeRange(val start: Int, val end: Int = UNDEFINED_TIME, val isOpenEnded: Boolean = false) : Comparable<TimeRange> {
 
     fun intersects(other: TimeRange): Boolean =
         isOpenEnded && other.start >= start ||
@@ -32,7 +32,7 @@ data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = fa
     fun toStringUsing(locale: Locale, range: String): String {
         val sb = StringBuilder()
         sb.append(timeOfDayToString(locale, start))
-        if (start != end) {
+        if (end != UNDEFINED_TIME && (start != end || !isOpenEnded)) {
             sb.append(range)
             var displayEnd = timeOfDayToString(locale, end)
             if (displayEnd == "00:00") displayEnd = "24:00"
@@ -51,5 +51,9 @@ data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = fa
             .toInstant()
             .toEpochMilli()
         return DateFormat.getTimeInstance(DateFormat.SHORT, locale).format(todayAt)
+    }
+
+    companion object {
+        const val UNDEFINED_TIME = Int.MIN_VALUE
     }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
@@ -6,7 +6,7 @@ import java.util.Locale
 
 /** A time range from [start,end). The times are specified in minutes. */
 @Serializable
-data class TimeRange(val start: Int, val end: Int = UNDEFINED_TIME, val isOpenEnded: Boolean = false) : Comparable<TimeRange> {
+data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = false) : Comparable<TimeRange> {
 
     fun intersects(other: TimeRange): Boolean =
         isOpenEnded && other.start >= start ||
@@ -31,7 +31,7 @@ data class TimeRange(val start: Int, val end: Int = UNDEFINED_TIME, val isOpenEn
     fun toStringUsing(locale: Locale, range: String): String {
         val sb = StringBuilder()
         sb.append(timeOfDayToString(locale, start))
-        if (end != UNDEFINED_TIME && (start != end || !isOpenEnded)) {
+        if (start != end || !isOpenEnded) {
             sb.append(range)
             var displayEnd = timeOfDayToString(locale, end)
             if (displayEnd == "00:00") displayEnd = "24:00"
@@ -42,8 +42,4 @@ data class TimeRange(val start: Int, val end: Int = UNDEFINED_TIME, val isOpenEn
     }
 
     override fun toString() = toStringUsing(Locale.GERMANY, "-")
-
-    companion object {
-        const val UNDEFINED_TIME = Int.MIN_VALUE
-    }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
@@ -32,7 +32,7 @@ data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = fa
     fun toStringUsing(locale: Locale, range: String): String {
         val sb = StringBuilder()
         sb.append(timeOfDayToString(locale, start))
-        if (start != end || !isOpenEnded) {
+        if (start != end) {
             sb.append(range)
             var displayEnd = timeOfDayToString(locale, end)
             if (displayEnd == "00:00") displayEnd = "24:00"
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
@@ -1,8 +1,7 @@
 package de.westnordost.streetcomplete.quests.opening_hours.model
 
+import de.westnordost.streetcomplete.util.timeOfDayToString
 import kotlinx.serialization.Serializable
-import java.text.DateFormat
-import java.time.*
 import java.util.Locale
 
 /** A time range from [start,end). The times are specified in minutes. */
@@ -43,15 +42,6 @@ data class TimeRange(val start: Int, val end: Int = UNDEFINED_TIME, val isOpenEn
     }
 
     override fun toString() = toStringUsing(Locale.GERMANY, "-")
-
-    private fun timeOfDayToString(locale: Locale, minutes: Int): String {
-        val seconds = (minutes % (24*60)) * 60L
-        val todayAt = LocalDateTime.of(LocalDate.now(), LocalTime.ofSecondOfDay(seconds))
-            .atZone(ZoneId.systemDefault())
-            .toInstant()
-            .toEpochMilli()
-        return DateFormat.getTimeInstance(DateFormat.SHORT, locale).format(todayAt)
-    }
 
     companion object {
         const val UNDEFINED_TIME = Int.MIN_VALUE
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
@@ -133,7 +133,8 @@ fun Rule.isSupportedCollectionTimes(): Boolean =
     // all sub-elements must be supported if specified
     holidays?.all { it.isSupported() } ?: true &&
     days?.all { it.isSupported() } ?: true &&
-    // only time points are supported for collection times
+    // times must be defined, and may not be ranges or something that implies ranges
+    !times.isNullOrEmpty() &&
     times?.all { it.isSupportedCollectionTimes() } ?: true &&
     // months not supported
     dates == null
@@ -188,7 +189,8 @@ fun TimeSpan.isSupportedCollectionTimes(): Boolean =
     interval == 0 &&
     start != TimeSpan.UNDEFINED_TIME &&
     // only support time points
-    end == TimeSpan.UNDEFINED_TIME
+    end == TimeSpan.UNDEFINED_TIME &&
+    !isOpenEnded
 
 /* ------------------------------ Collision/Intersection checking ------------------------------- */
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
@@ -63,11 +63,7 @@ fun OpeningHoursRuleList.toCollectionTimesRows(): List<OpeningHoursRow>? {
     val result = mutableListOf<OpeningHoursRow>()
 
     for (rule in rules) {
-        if (rule.modifier != null && rule.modifier!!.isSimpleOff()) {
-            result.add(rule.createOffDays())
-        } else {
-            result.addAll(rule.createOpeningWeekdays())
-        }
+        result.addAll(rule.createOpeningWeekdays())
     }
 
     return result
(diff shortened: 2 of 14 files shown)


westnordost:
Bin grad unterwegs, aber

WeekdaysTimes(...) is a constructor, weekdays(...) is a function that takes an Int and 0b.... is the syntax in kotlin to define an int as a series of booleans.... you know .... "there are 10 types of people, those that understand booleans and those who don't" ;-)

https://kotlinlang.org/docs/basic-types.html#literal-constants

I think weekdays is a utility function in the tests to be able to write 0b1110011 to mean Mo-Fr, Sa, So

westnordost:
So uf that was a source of confusion, maybe add a dicumentation comment to that function

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddCollectionTimesForm.kt
```diff
@@ -69,8 +80,21 @@ class AddCollectionTimesForm : AbstractQuestFormAnswerFragment<CollectionTimesAn
         }
     }
 
-    private fun loadCollectionTimesData(savedInstanceState: Bundle?): List<WeekdaysTimesRow> =
-        savedInstanceState?.let { Json.decodeFromString(it.getString(TIMES_DATA)!!) } ?: listOf()
+    private fun initStateFromTags() {
+        val ct = osmElement!!.tags["collection_times"]
+        val rows = ct?.toOpeningHoursRules()?.toOpeningHoursRows()
```
westnordost, line 85: There should be a function called `OpeningHoursRuleList.toCollectionTimesRows()` that both checks if it is really valid collection times and also if it is supported by streetcomplete. F.e. months, off days etc etc are not supported for collection times.

Also that function should return a data structure that is fitting for collection times, f.e. `WeekdaysTimesRow` As said, those are not time ranges.

westnordost (changes requested):
Alright, I am back and had time to review this PR. Cool that you are doing this, it always felt like this was kinda missing.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
```diff
@@ -32,7 +32,7 @@ data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = fa
     fun toStringUsing(locale: Locale, range: String): String {
         val sb = StringBuilder()
         sb.append(timeOfDayToString(locale, start))
-        if (start != end || !isOpenEnded) {
+        if (start != end) {
```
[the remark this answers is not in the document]
eginhard: I tried to reuse as much as possible from the opening hours logic which is based on TimeRange. Currently I'm using a time range with same start and end to represent the collection time *points*. This line then ensures that "10:00" is shown in the form instead of "10:00 - 10:00". But maybe that doesn't actually work because we could have 24h range like "Mo 10:00-10:00".

eginhard:
> WeekdaysTimes(...) is a constructor, weekdays(...) is a function that takes an Int and 0b.... is the syntax in kotlin to define an int as a series of booleans.... you know .... "there are 10 types of people, those that understand booleans and those who don't" ;-)
> 
> https://kotlinlang.org/docs/basic-types.html#literal-constants
> 
> I think weekdays is a utility function in the tests to be able to write 0b1110011 to mean Mo-Fr, Sa, So

Thanks, I found the relevant functions. For some reason Android Studio doesn't let me do Ctrl+B (go to declaration) in the tests so I had trouble with that.

Thank you for your other comments! I better understand some of the subtleties now and will update the PR accordingly, probably only later this week.

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
```diff
@@ -8,6 +8,7 @@ import de.westnordost.streetcomplete.data.quest.NoCountriesExcept
 import de.westnordost.streetcomplete.ktx.arrayOfNotNull
 import de.westnordost.streetcomplete.ktx.containsAnyKey
 import de.westnordost.streetcomplete.quests.getNameOrBrandOrOperatorOrRef
+import de.westnordost.streetcomplete.quests.opening_hours.parser.toOpeningHoursRules
 
 class AddPostboxCollectionTimes : OsmFilterQuestType<CollectionTimesAnswer>() {
```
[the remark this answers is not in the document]
eginhard: Actually `collection_times` is allowed to have any value supported for opening hours: https://wiki.openstreetmap.org/wiki/Key%3Acollection_times (although there are not many time range values and some of them appear to be 24h time ranges like 09:00-09:00 when probably 09:00 was meant: https://overpass-turbo.eu/s/18Qe).

So I would suggest to use the same opening hours parser to check for validity and then only show quests for the invalid values and ones that we can handle (time points).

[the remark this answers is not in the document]
westnordost: Same parser, yes, but time ranges for collection times are not supported by the app.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/model/TimeRange.kt
```diff
@@ -7,7 +7,7 @@ import java.util.Locale
 
 /** A time range from [start,end). The times are specified in minutes. */
 @Serializable
-data class TimeRange(val start: Int, val end: Int, val isOpenEnded: Boolean = false) : Comparable<TimeRange> {
+data class TimeRange(val start: Int, val end: Int = UNDEFINED_TIME, val isOpenEnded: Boolean = false) : Comparable<TimeRange> {
```
eginhard, line 10: This is how time points are defined in TimeSpan of ch.poole.openinghoursparser. I think it makes sense to adopt the same here because a lot of code is based on that.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
```diff
@@ -53,9 +53,30 @@ fun OpeningHoursRuleList.toOpeningHoursRows(): List<OpeningHoursRow>? {
...
+        }
+    }
+
+    return result
+}
+
 /* ---------------------------------- Checks if it is supported --------------------------------- */
 
 fun OpeningHoursRuleList.isSupported(): Boolean = rules.isSupported()
+fun OpeningHoursRuleList.isSupportedCollectionTimes(): Boolean = rules.isSupportedCollectionTimes()
```
eginhard, line 79: There is some code duplication now. It might be cleaner to have a Boolean parameter in the various isSupported functions to handle the differences?

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
```diff
@@ -49,11 +50,22 @@ class AddPostboxCollectionTimes : OsmFilterQuestType<CollectionTimesAnswer>() {
...
 
-    override fun getTitle(tags: Map<String, String>): Int =
-        if (tags.containsAnyKey("name", "brand", "operator", "ref"))
-            R.string.quest_postboxCollectionTimes_name_title
-        else
-            R.string.quest_postboxCollectionTimes_title
+    override fun getTitle(tags: Map<String, String>): Int {
+        val hasName = tags.containsAnyKey("name", "brand", "operator", "ref")
+        // treat invalid collection times like it is not set at all
+        val hasValidCollectionTimes = tags["collection_times"]?.toOpeningHoursRules() != null
```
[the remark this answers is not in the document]
eginhard: This should remain. Valid opening hours are valid collection times. The valid times we can't handle won't be shown for resurvey anymore.

eginhard (commented):
I've added those refinements, namely checking which valid collection times are supported in the app. Please check whether that is correct.

I'm still using TimeRange for collection times as well, but now with a special end value, to be compatible with ch.poole.openinghoursparser.

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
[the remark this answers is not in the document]
westnordost: Ok

app/src/main/java/de/westnordost/streetcomplete/quests/postbox_collection_times/AddPostboxCollectionTimes.kt
```diff
@@ -49,11 +54,36 @@ class AddPostboxCollectionTimes : OsmFilterQuestType<CollectionTimesAnswer>() {
...
         arrayOfNotNull(getNameOrBrandOrOperatorOrRef(tags))
 
-    override fun getTitle(tags: Map<String, String>): Int =
-        if (tags.containsAnyKey("name", "brand", "operator", "ref"))
-            R.string.quest_postboxCollectionTimes_name_title
-        else
-            R.string.quest_postboxCollectionTimes_title
+    override fun getTitle(tags: Map<String, String>): Int {
+        val hasName = tags.containsAnyKey("name", "brand", "operator", "ref")
+        // treat invalid collection times like it is not set at all
```
westnordost, line 59:
```suggestion
        /* treat invalid collection times like it is not set at all. Any opening hours are
           legal tagging for collection times, even though they are not supported in
           this app, i.e. are never asked again */
```

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
```diff
@@ -71,6 +92,16 @@ fun List<Rule>.isSupported(): Boolean =
...
+ * Returns false otherwise, in cases where it is not directly representable */
+@JvmName("isSupportedCollectionTimesRuleList")
+fun List<Rule>.isSupportedCollectionTimes(): Boolean =
+    // all rules must be supported
+    all { it.isSupportedCollectionTimes() } &&
+        // this kind of collection times specification likely require fix
+        // anyway, it is not representable directly by SC
+        (!weekdaysCollideWithAnother())
+
 fun Rule.isSupported(): Boolean =
```
westnordost, line 105: since they are now two different "isSupported", it'd be better to name the other one accordingly too - i.e. `isSupportedOpeningHours` etc

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
```diff
@@ -95,6 +126,32 @@ fun Rule.isSupported(): Boolean =
...
+    !isFallBack &&
+    // "1995-1998 08:00-11:00" not supported
+    years == null &&
+    // "05-08 08:00-11:00" not supported
+    weeks == null &&
+    (
+        // for normal rules, only "Mo-Fr" not supported. "open" modifier is ok as long as it does not have a comment
+        (modifier == null || modifier!!.isSimpleOpen()) && !times.isNullOrEmpty() ||
+        // "off"/"closed" only compatible without comment and no times
+        (modifier != null && times.isNullOrEmpty() && modifier!!.isSimpleOff())
```
westnordost, line 145: does the collection times form support "off days"? I think not

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
westnordost: No, that's okay. It is for two different UIs which also do not inherit anything from each other.

Except the `OpeningHoursRow` data structure, about which I am not happy about anyway.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
```diff
@@ -53,9 +53,30 @@ fun OpeningHoursRuleList.toOpeningHoursRows(): List<OpeningHoursRow>? {
...
+    }
+
+    val result = mutableListOf<OpeningHoursRow>()
+
+    for (rule in rules) {
+        if (rule.modifier != null && rule.modifier!!.isSimpleOff()) {
+            result.add(rule.createOffDays())
+        } else {
+            result.addAll(rule.createOpeningWeekdays())
+        }
```
westnordost, line 70: 1. I'd very much prefer if the opening hours model used in the view and collection times model used in the view is separate.
2. The collection times form does not support off days, does it?

westnordost:
Are you still working on this and/or intend to finish this PR?

westnordost:
I'd really like to pull your work into the app, it is a nice feature and I am thankful to you for dwelving into it. If you lost interest, that's okay, someone else may finish it and your work on it will still be credited in any case.

eginhard:
Sorry, I've been busy. I'll update the PR next week.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
```diff
@@ -64,14 +81,24 @@ fun OpeningHoursRuleList.isSupported(): Boolean = rules.isSupported()
...
-fun Rule.isSupported(): Boolean =
+/** Returns true if the collection times are supported by StreetComplete
+ * Returns false otherwise, in cases where it is not directly representable */
+@JvmName("isSupportedCollectionTimesRuleList")
+fun List<Rule>.isSupportedCollectionTimes(): Boolean =
+    // all rules must be supported
+    all { it.isSupportedCollectionTimes() } &&
+        // this kind of collection times specification likely require fix
+        // anyway, it is not representable directly by SC
+        (!weekdaysCollideWithAnother())
```
westnordost, line 99:
```suggestion
    // this kind of collection times specification likely require fix
    // anyway, it is not representable directly by SC
    (!weekdaysCollideWithAnother())
```
indentation

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/parser/OpeningHoursParser.kt
eginhard: Ok, let's finish this. I've rebased the PR and made it consistent with recent changes in the opening hours quest. Off days are ignored now for collection times.

But I don't really know how to best use such a CollectionTimesRow data class here. I think it would still have to be a subclass of OpeningWeekdaysRow anyway to avoid a lot of duplication in the opening hours parser...

westnordost:
Congratulations! Glad you had the persistance to see it through. I'll review it again and if I find anything small to critize, I'll just do it myself.

westnordost:
What I changed:
- I used an own data structure for the collection times. As-such it wasn't that much more code.
- but I also added tests for "collection times parsing". With this, I found 2 bugs in the parsing code and fixed them

westnordost (approved):

eginhard:
Thanks a lot for fixing this up, that looks good!

peternewman:
Just to say @eginhard , thanks for this and for getting it in, much appreciated and works perfectly!

PR #2992, opened 2021-06-22
Adding OpenCampingmap to achievements

slackline:
I came across [OpenCampingmap.org](https://opencampingmap.org/) and thought it would be a useful addition to the achievements.

I've virtually no experience with Java but have attempted to add it based on existing entries. Couldn't work out how the icon is included though.

Happy to take feedback on how to improve this Pull Request.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
@@ -154,6 +154,14 @@ object AchievementsModule {
             R.string.link_cyclosm_description
         ),
         Link(
+            "opencampingmap",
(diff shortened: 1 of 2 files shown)


westnordost:
Thanks, slackline. As of now, however, adding links is not open for contribution, I am curating it alone.

But I will add OpenCampingMap to my private "maybe include later" list, I didn't know it before, thank you for the link.

slackline:
Not a problem at all and thanks for the quick feedback and all of the work you've put into StreetComplete, its a great application.

westnordost:
There is also http://unterkunftskarte.de/#14/53.9111/10.8277 btw., though right now I am getting no results. I wonder if it is broken now. That's in general a problem with those... smaller map-mashups that are focussed on just one thing, they tend fall into disrepair quite quickly.

Issue #2692, opened 2021-03-29
Allow user to see/comment on ALL open Notes (even their own)


mnalis:
Related to https://github.com/streetcomplete/StreetComplete/issues/1029, it would be nice if SC user could see all open Notes (including their own Notes).

**Use case**

I would prefer if `Show all notes` preference would in fact be **showing all notes** (and not only notes that the user currently logged in via SC has never interacted with, which seems to be current behavior). Some of the reasons:

- sometimes I mark notes on desktop (JOSM) to mark that I need to (re-)survey something physically on-site (which I can then either solve directly in SC, or can improve the note with comment and picture if it requires editing in full-blown editor).
- sometimes I mark note in SC which I intend so solve in same physical trip on location, but am unable to do from current position (for example I'm on the wrong side of the stream, or I don't want to wait 10 minutes for train to pass to gain access, or current vintage point only shows part of the picture [think quest for long stretches of road], or I want to complete mapping the area in this trip but part of it is inaccessible from current point but will be on return part of trip etc.)
- sometimes I opened the note on earlier day (in SC or other mobile app) as I couldn't answer it fully then, but can now (for example, place wasn't opened at the time, or I only could check part of the way, or the data was about to change [like under (re-)construction] or is only surveyable during daytime [think visibility] or nighttime [like `lit` sometimes] etc.)
- the default preference (showing only notes not related to current user, and ending with question mark) **would still be the default**, so vast majority of users will not be affected - only those that indicated they want to see **all notes** would actually see all notes - so it does not seem to me such a change would cause general user outrage.

Current workaround is using additional app on the mobile and switching between it and SC (which has it's own set of problems - not only is it quite unergonomic and delays the mapping, switching apps also causes SC to lose GPS signal and never acquire it automatically again in some annoyingly high percentage of cases, which requires closing and opening the app again, and is annoying)

**Proposed Solution**

Make `Show all notes` preference setting to do exactly that if enabled - show all Notes (and not filtered subset of Notes).

matkoniecz:
So if note would be present it would never disappear, even once answered? Or would it reappear on rescan.

matkoniecz:
Also, see #569

dbdean:
Maybe notes could show again if there is a new comment since it was hidden in SC. Something like that would be very useful for me too.

mnalis:
[...]

Issue #2641, opened 2021-03-07
Expand note selection


Cj-Malone:
Currently notes are only shown if they contain a question mark. I think that should be expanded to also check for "#suveryme". It's used in the UK when something needs to be check on the ground, we have a QA tool that highlights theses notes, although a lot of them overlap by also containing a question mark not all of them do. I find myself rephrasing some notes to contain a question mark just to get it visible in StreetComplete.

westnordost:
Can you provide some numbers on #surveyme?

westnordost:
Also, do you know that you can change this in the settings? (to show all notes you haven't contributed to yet)

Cj-Malone:
> Can you provide some numbers on #surveyme?

No, I don't know how to query notes.

> Also, do you know that you can change this in the settings? (to show all notes you haven't contributed to yet)

I know that, but default is king. That also  doesn't change that these notes have been tagged as needing on the ground knowledge, and StreetComplete is a on the ground OSM editor.

ENT8R:
> Can you provide some numbers on #surveyme?

[https://ent8r.github.io/NotesReview/?query=#surveyme&limit=1000&closed=true](https://ent8r.github.io/NotesReview/?query=%23surveyme&limit=1000&closed=true) returns 605 notes in total and 411 notes which are still open as of now.

mnalis:
> [https://ent8r.github.io/NotesReview/?query=#surveyme&limit=1000&closed=true](https://ent8r.github.io/NotesReview/?query=%23surveyme&limit=1000&closed=true) returns 605 notes in total and 411 notes which are still open as of now.

@ENT8R  Interestingly, that link just forwards to the main page without doing the query specified (both in Firefox and Chromium). But I can manually enter the query and it works just fine then! Nice tool, BTW

Marc-marc-marc:
there are even more notes with #survey, wouldn't it be relevant to include it ?

westnordost:
Maybe. Would you like to do some research as to whether the #survey hashtag is actually used to denote that?

westnordost:
Tried to do the research, however, when entering #survey, even "#survey", NotesReview searches simply for "survey" (without the hashtag).

Created a bug report here. https://github.com/ENT8R/NotesReview/issues/66

westnordost:
Ok I researched via http://resultmaps.neis-one.org/osm-notes (click on country, then on "only open notes".
[...]

PR #2998, opened 2021-06-26
Add via to note comments

dbdean:
Add a 'via StreetComplete <version>' to note comments added by StreetComplete. This will allow (in a different, future change) the notes quest to differentiate between note comments created in StreetComplete or elsewhere, allowing notes to be reactivated in StreetComplete when commented on elsewhere even by the StreetComplete user.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsController.kt
@@ -30,7 +30,7 @@ import javax.inject.Singleton
             noteId,
             position,
             action,
-            fulltext,
+            fullText,
             imagePaths,
             currentTimeMillis(),
             false,
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsController.kt
@@ -1,5 +1,6 @@
 package de.westnordost.streetcomplete.data.osmnotes.edits
 
+import de.westnordost.streetcomplete.ApplicationConstants
 import de.westnordost.streetcomplete.data.osm.mapdata.BoundingBox
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.data.osmnotes.Note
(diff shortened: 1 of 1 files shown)


matkoniecz:
It feels to me that it would be quite spammy for people processing notes.

dbdean:
It's the same thing we add when we create comments, and the OSM website team are not interested in adding metadata to notes to show the creation app, unfortunately.

dbdean:
And I want to stop notes disappearing in StreetComplete when I comment on them elsewhere, and this is the first step in making that happen.

andrewharvey:
> It feels to me that it would be quite spammy for people processing notes.

It's helpful for people processing notes to understand where the comment came from, it gives more context and aids mappers dealing with notes. A comment from StreetComplete means it was via a ground survey which is much more reliable than someone commenting based on an external dataset.

SC already adds via to the initial note, so adding to comments would complete the coverage.

andrewharvey:
Once the OSM website eventually supports a proper created_by metadata field it can be removed, but until then the next best thing is appending to the comment text.

westnordost:
I'm with @matkoniecz here. It is not really relevant for the one who processes notes to know with which app a note comment has been created. It is only relevant to know whether a comment has been written by someone on-site / survey knowledge or someone else. And that information is usually clear from the context.

It seems that to supply the app name and version in every note comment would only serve for the app StreetComplete itself, to be able to distinguish notes commented with the app and notes commented from home so that the latter could still be shown in SC.

westnordost:
> And I want to stop notes disappearing in StreetComplete when I comment on them elsewhere, and this is the first step in making that happen.

Maybe that could be done some other way. Some other magic keyboard or word that would need to be mentioned so that it is shown anyway?

mnalis:
> Maybe that could be done some other way. Some other magic keyboard or word that would need to be mentioned so that it is shown anyway?

maybe we can reuse `#surveyme`, as added in https://github.com/streetcomplete/StreetComplete/issues/2641? It implies that note needs ground survey, so adding it when desktop editing would force showing it in StreetComplete (and other ground-survey based apps that choose to use it). 

It might also be usable workaround for people unhappy with https://github.com/streetcomplete/StreetComplete/issues/2692 resolution (myself included) - it would require little more typing but would ensure note which has this in last comment was always shown in SC, no matter what.

dbdean:
Ok, clearly this isn't the solution we want - I've created https://github.com/streetcomplete/StreetComplete/issues/3002 to continue discussion.

PR #2999, opened 2021-06-27
Missing highway crossing

### LINKED ISSUE #398: Quest: add highway=crossing
#This is related to the other quest with crossing (#397) but this time we would look for intersections between a street (highway ~ primary|secondary|tertiary|residential|unclassified|road) and a footway that don't have a `highway=crossing`.

If such node doesn't have a `highway=crossing`, we add it, if the crossing is not possible we add `crossing=no`.

The question can be a simple yes/no, e.g.: Can pedestrians cross the street here?

I'm not sure how to built such query in current tag filters, a query from [overpass-turbo](http://overpass-turbo.eu/s/qoV):
```
[bbox:{{bbox}}];
way[highway]["highway"~"primary|secondary|tertiary|residential|unclassified|primary_link|secondary_link|tertiary_link|service"];node(w)->.n1;
way["highway"="footway"];node(w)->.n2;
node.n1.n2["highway"!="crossing"];
out meta;
```

EDIT: Removed `road`, added `primary_link|secondary_link|tertiary_link|service`

westnordost:
Fixes #398

However, this is blocked until a tagging issue is resolved. More about it in the linked ticket.


westnordost:
The block is resolved after some research (see wiki page)

app/src/main/res/values/strings.xml
```diff
@@ -369,6 +369,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="credits_and_more">and more…</string>
     <string name="quest_maxspeed_answer_living_street">It is a living street</string>
     <string name="quest_maxspeed_answer_living_street_confirmation_title">So, there is a sign like this?</string>
+    <string name="quest_crossing_title">Can and may the road be crossed without obstruction here?</string>
```
westnordost, line 372:
```suggestion
    <string name="quest_crossing_title">Is this spot suitable to cross the road and is it legal to do so here?</string>
```

westnordost:
So this is ready to merge (for the next major version) from my view. Anyone wants to review it, as it is a little more complex? I'd be thankful for that, maybe we can spare ourselves some bugfixes on this quest - it is similarily complex as the kerb quest(s).

app/src/test/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingTest.kt
```diff
@@ -0,0 +1,186 @@
...
+        val mapData = TestMapDataWithGeometry(listOf(
+            node(1, p(0.0, -1.0)),
+            shared,
+            node(3, p(0.0, +1.0)),
+            node(4, p(-1.0, 0.0)),
+            way(1, nodes = listOf(1, 2, 3), tags = mapOf("highway" to "unclassified")),
+            way(2, nodes = listOf(4, 2), tags = mapOf("highway" to "footway")),
+        ))
+        assertEquals(0, questType.getApplicableElements(mapData).toList().size)
+    }
```
jyasskin, line 67: This is what it looks like at the boundary between separated sidewalks and shared-way sidewalks, and it represents a crossing at that boundary: https://wiki.openstreetmap.org/wiki/File:Sidewalk_crossing-EN.png. It probably makes sense to present this quest when the `highway=unclassified` here also has a `sidewalk=*` key indicating there's a sidewalk on the opposite side of the road from the separated footway.

app/src/test/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingTest.kt
```diff
@@ -0,0 +1,186 @@
...
+            node(1, p(0.0, -1.0)),
+            shared,
+            node(3, p(0.0, +1.0)),
+            node(4, p(-1.0, 0.0)),
+            node(5, p(+1.0, 0.0)),
+            way(1, nodes = listOf(1, 2, 3), tags = mapOf("highway" to "unclassified")),
+            way(2, nodes = listOf(4, 2, 5), tags = mapOf("highway" to "footway")),
+        ))
+        assertEquals(shared, questType.getApplicableElements(mapData).toList().single())
+    }
```
jyasskin, line 35: A 'no' answer for this quest is going to give a weird result: instead of the footway ending at the edge of the road, where pedestrians actually have to stop, it'll continue into the middle of the road. If this result is acceptable, we could do it for #3106 too, but it's not really the right resulting map.

westnordost:
Thank you for the review, @jyasskin !

app/src/test/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingTest.kt
westnordost: To paraphrase your suggestion...:

Treat a footway segment as **crossing** a road even if it only touches it, **if** the road has a sidewalk on the other side of the road (as seen from that footway segment).

This could work.

However, at least for the first iteration/implementation, I think I would rather be more conservative here because maybe I am missing a possible configuration. After all, we kind of need to keep in mind also kind of wrong-ish taggings that don't follow the ideal tagging strictly.

The crossover between the two tagging schemes are always problematic. In general, we need to be wary to not produce false positives, i.e. the quest definitely should not show if a footway just ends on a road (or touches it) but does not cross it.

So also because it increases the complexity of the implementation, I would not implement that for the time being. After all, if any `highway=crossing` is tagged, we already ask for the properties of a crossing. This is just to find the places where this info is missing with relative certainty.

app/src/test/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingTest.kt
westnordost: Not sure what you mean.

image

jyasskin: Here's a real [intersection ](https://www.openstreetmap.org/edit#map=20/45.51423/-122.62283)with a footway drawn across the street, but no crossings marked:

Footway=crossing across the street

Here's the footway after the user gives this quest a "no":

`crossing=no` node, but still a footway across the street

And here are the ways drawn as I believe is correct:

No more way across the street

westnordost: Hmm, I understand, you are right.

westnordost: I guess the only solution would be:
1. exclude `footway=crossing`. So basically, the quest is only asked for footways with neither `footway=sidewalk` nor `footway=crossing` tagged
2. but what could be done: For the quest that asks for the type of crossing, also ask for nodes at the intersection between `footway=crossing` and a road, i.e. implicitly assume that footways tagged with crossing are indeed crossings. In the case depicted above - where the whole crossing way was added in error, the user can leave a note, explaining that there is indeed no crossing

jyasskin: Yeah, that's similar to what I was doing with #3111: if a footway is drawn to cross a car-way with `sidewalk=separate`, and the surveyor says that it's not actually a crossing, prompt them to leave a note. Even if it doesn't have a `footway=*` key, a 'not a crossing' answer probably means that the ways need to be rearranged.

This stuff is easier on shared-way sidewalks, given StreetComplete's goal to avoid having a quest touch more than one node: separate-way sidewalks just have lots of nodes involved in any road feature.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+
+        val footwaysByNodeId = mapData.ways.asSequence()
+            .filter { footwaysFilter.matches(it) }
+            .groupByNodeIds()
+        /* filter out nodes of footways that are the end of a footway, f.e.
+         * https://www.openstreetmap.org/node/1449039062 or
+         * https://www.openstreetmap.org/node/56606744 */
+        footwaysByNodeId.removeEndNodes()
```
matkoniecz, line 72: this will filter out also cases when footway line is split on the road, right?

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
```diff
@@ -0,0 +1,186 @@
...
+                    if (roadTakesRightTurn) {
+                        if (side1 > 0 && side2 > 0) 1 else -1
+                    } else {
+                        if (side1 > 0 || side2 > 0) 1 else -1
+                    }
+                }
+                val anyAreOnDifferentSidesOfRoad = sides.toSet().size > 1
+                anyAreOnDifferentSidesOfRoad
+            }
+            anyFootwayCrossesAnyRoad
```
matkoniecz, line 143: what kind of syntax is being used here? I am really stumped how it works

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
```diff
@@ -0,0 +1,186 @@
...
+         * https://www.openstreetmap.org/node/280046349 or
+         * https://www.openstreetmap.org/node/56606744 */
+        roadsByNodeId.removeEndNodes()
+
+        /* require all roads at a shared node to either have no sidewalk tagging or all of them to
+         * have sidewalk tagging: If the sidewalk tagging changes at that point, it may be an
+         * indicator that this is the transition point between separate sidewalk mapping and
+         * sidewalk mapping on road-way. F.e.:
+         * https://www.openstreetmap.org/node/1839120490 */
+        val anySidewalk = setOf("both","left","right")
```
matkoniecz, line 57: is sidewalk=separate skipped deliberately?

With `sidewalk=separate` and footway crossing road this quest still makes sense. See https://www.openstreetmap.org/#map=19/50.06687/19.92738

screen06

tagging crossing there would enable say tactile paving quest

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
```diff
@@ -0,0 +1,186 @@
...
+import de.westnordost.streetcomplete.util.initialBearingTo
+import de.westnordost.streetcomplete.util.normalizeDegrees
+
+class AddCrossing : OsmElementQuestType<Boolean> {
+
+    private val roadsFilter by lazy { """
+        ways with
+          highway ~ primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential
+          and area != yes
+          and crossing != no
```
matkoniecz, line 21: `crossing=no` is used on 272 ways worldwide, I would not consider it as a good or common tagging

In such case missing `crossing=no` on nodes should be added

See https://taginfo.openstreetmap.org/tags/crossing=no#overview

matkoniecz (commented):
BTW, I was quite surprised how many missing `highway=crossing` are in my city and were found by this quest

app/src/test/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingTest.kt
matkoniecz: > And here are the ways drawn as I believe is correct:

In such case this quest still resulted in improvement - there will be still a spurious way, but it is not worse than before.

Maybe in such case SC should ask user to produce a note with a photo?

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
westnordost: no

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
westnordost: Quirks of Kotlin syntax I don't like neither. Imagine there is a `return@retainAll` there

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
```diff
@@ -0,0 +1,186 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+import de.westnordost.streetcomplete.util.initialBearingTo
+import de.westnordost.streetcomplete.util.normalizeDegrees
+
+class AddCrossing : OsmElementQuestType<Boolean> {
+
+    private val roadsFilter by lazy { """
+        ways with
+          highway ~ primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential
```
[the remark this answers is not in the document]
westnordost: ok

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
westnordost: ok

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
westnordost: I think you misunderstood that portion of the code. See the explanatory comment above.

But yeah, separate footways are skipped, because of the discussion I had with @jyasskin : If there is indeed no crossing on separately mapped sidewalks, setting `crossing=no` is not enough, but the whole crossing way needs to be deleted which is something that StreetComplete can't do.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt
smichel17: You could pull the lambdas out into a functions:

```kt
fun anyAreOnDifferentSidesOfRoad(i) {
    // ...
    return sides.toSet().size > 1
}

fun anyFootwayCrossesAnyRoad(nodeId, footways) {
    // ...
    return (1 until neighbouringRoadPositions.size).any(::anyAreOnDifferentSidesOfRoad)
}

// ...

footwaysByNodeId.entries.retainAll(::anyFootwayCrossesAnyRoad)
```

westnordost: I will look at that later, thanks

westnordost: Alright, I hope with this commit, the code is easier to read:

https://github.com/streetcomplete/StreetComplete/pull/2999/commits/a510cd525a23962202dd8e61f98fece7cc30ee44

PR #3004, opened 2021-06-28
Change achievement from "Postman" to non-gendered "Postal Worker"

jancborchardt:
"Postal worker" seemed to be the most common fitting word. There’s also "mail carrier" but it’s quite specific and sounds off as an achievement.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -938,8 +938,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="achievement_bicyclist_solved_X">You solved %d quests that help cyclists!</string>
     <string name="achievement_wheelchair_title">Mobility</string>
     <string name="achievement_wheelchair_solved_X">You solved %d quests that help wheelchair users get around town!</string>
-    <string name="achievement_postman_title">Postman</string>
-    <string name="achievement_postman_solved_X">You helped to locate %d addresses!\nThis data is essential not only for the postman but also for any navigation system.</string>
+    <string name="achievement_postman_title">Postal Worker</string>
+    <string name="achievement_postman_solved_X">You helped to locate %d addresses!\nThis data is essential not only for postal workers but also for any navigation system.</string>
     <string name="achievement_building_title">High-Rise</string>
     <string name="achievement_building_solved_X">You solved %d building related quests.\nNote that this app asks for the housenumber of a building only after its type has been determined.</string>
     <string name="achievement_blind_title">Sixth Sense</string>


westnordost:
I like postman better. "Postal worker" sounds stilted.

But since no actual postman is visible on the achievement itself, it doesn't need to be named "postman" at all, could have a more creative and engaging name too, if you have any ideas.

jancborchardt:
Sure, here's some nicer achievement titles:
- House number hunter
- Carrier pigeon
- Number cruncher
- Number navigator
- Address acquirer
- Locator
- Destination discoverer
- Street scout

westnordost:
Hmm, some nice ideas, however, then the association of the name (f.e. number cruncher, carrier pigeon) doesn't really fit the picture. Granted, same with postman. The names for the achievements were added before the picture was made. Maybe this could be an oppurtunity to find a better name for the achievement(s) that also fits better to the picture(s):


For the **Treasure hunt** achievement for example, it is perfect:

matkoniecz:
> Maybe this could be an oppurtunity to find a better name for the achievement(s) that also fits better to the picture(s):

Letters
Post delivery
Signpost
Address hunter
Addressator
Address hunt
Census

westnordost:
Housenumber forest? (pun of Schilderwald -> forest of traffic signs)

riQQ:
How about Guide or Tour Guide? Or something like "Leading the way"?

smichel17:
"[Signed, Sealed, Delivered!](https://www.youtube.com/watch?v=inXC_lab-34)"?

mnalis:
I'm having a hard time thinking of suitable replacement (only one that seems to fit has same problem `mailman`), and the suggested replacements so far don't  sound so good to me...

While I support (and try to use) gender neutrality, is this **really an issue** here?  I'm no native English speaker, but never did it imply to me that term `postman` in any way tied this word only to males of the human species.

So I've looked up https://www.etymonline.com/word/postman and links, at it does **not** seem that `man` specified **male** of the human species in this case. It seems to come from proto-Germanic **_mann_**, meaning _"human being, person (male or female)"_. Interestingly, it seems that only in late 13th century English did the word "**man**" contract it's male connotation (**_wer_** and **_wif_** were used before for such purpose), and there are many words today (with both neutral, positive and negative connotations) which use it in its original non-gender-specific meaning, like for example: _**man**kind_, _**man**slaughter_, _**mann**ers_, _**man**ure_, _**man**ual_, _**man**euver_ and even _post**man**_.

But now that the issue is raised, more problematic might be probably that in all achievements that I have (probably not all, but quite a few of them) all persons seem to be of white skin tones and Caucasian looking (no Asian or African looking people, for example).

westnordost:
Jesus. Let's not spend so much time on this issue:
- If a good replacement that is an improvement of the current name because it is interesting / better fitting / funny  ... can be found, that's fine
- Otherwise, postman is fine too

westnordost:
"You got mail!" ?

westnordost:
Oh that's nice. I've read the book and seen the movie. Of course, puns are always hard to translate into other languages.

Anyway, I've created this discussion in the forum. Everybode is free to contribute his achievement title suggestions:

https://github.com/streetcomplete/StreetComplete/discussions/3009

PR #3007, opened 2021-06-29
Refactor MapTilesDownloader to use flows

smichel17:
ref: https://github.com/streetcomplete/StreetComplete/pull/2794#commitcomment-52700330

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
@@ -14,6 +14,7 @@ import java.io.IOException
 import java.lang.System.currentTimeMillis
 import javax.inject.Inject
 import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
 
 class MapTilesDownloader @Inject constructor(
     private val vectorTileProvider: VectorTileProvider,
@@ -50,48 +51,46 @@ class MapTilesDownloader @Inject constructor(
         (vectorTileProvider.maxZoom downTo 0).forEach { zoom ->
             bbox.enclosingTilesRect(zoom).asTilePosSequence().forEach { (x, y) ->
                 launch {
-                    val result = try {
-                        downloadTile(zoom, x, y)
+                    try {
+                        val result = downloadTile(zoom, x, y)
+                        val logText = if (result.alreadyCached) "in cache" else "downloaded"
+                        Log.v(TAG, "Tile $zoom/$x/$y $logText")
+                        send(result)
                     } catch (e: Exception) {
-                        DownloadFailure
+                        Log.w(TAG, "Error retrieving tile $zoom/$x/$y: ${e.message}")
+                        send(DownloadFailure)
                     }
-                    send(result)
                 }
             }
         }
     }
 
-    private suspend fun downloadTile(zoom: Int, x: Int, y: Int): DownloadResult = suspendCancellableCoroutine { cont ->
+    private suspend fun downloadTile(zoom: Int, x: Int, y: Int): DownloadSuccess = suspendCancellableCoroutine { cont ->
         /* adding trailing "&" because Tangram-ES also puts this at the end and the URL needs to be
            identical in order for the cache to work */
         val url = vectorTileProvider.getTileUrl(zoom, x, y) + "&"
         val httpUrl = HttpUrl.parse(url)
         require(httpUrl != null) { "Invalid URL: $url" }
 
-        val builder = Request.Builder()
+        val request = Request.Builder()
             .url(httpUrl)
             .cacheControl(cacheConfig.cacheControl)
-        builder.header("User-Agent", ApplicationConstants.USER_AGENT + " / " + Version.userAgent())
-        val call = okHttpClient.newCall(builder.build())
+            .header("User-Agent", ApplicationConstants.USER_AGENT + " / " + Version.userAgent())
+            .build()
+        val call = okHttpClient.newCall(request)
 
         /* since we use coroutines and this is in the background anyway, why not use call.execute()?
         *  Because we want to let the OkHttp dispatcher control how many HTTP requests are made in
         *  parallel */
         val callback = object : Callback {
             override fun onFailure(call: Call, e: IOException) {
-                Log.w(TAG, "Error retrieving tile $zoom/$x/$y: ${e.message}")
-                cont.resume(DownloadFailure)
+                cont.resumeWithException(e)
             }
 
             override fun onResponse(call: Call, response: Response) {
-                var size = 0
-                response.body()?.use { body ->
-                    // just get the bytes and let the cache magic do the rest...
-                    size = body.bytes().size
-                }
+                // just get the bytes and let the cache magic do the rest...
+                val size = response.body()?.use { it.bytes().size } ?: 0
                 val alreadyCached = response.cacheResponse() != null
-                val logText = if (alreadyCached) "in cache" else "downloaded"
-                Log.v(TAG, "Tile $zoom/$x/$y $logText")
                 cont.resume(DownloadSuccess(alreadyCached, size))
             }
         }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
@@ -45,6 +45,9 @@ class MapTilesDownloader @Inject constructor(
         Log.i(TAG, "Downloaded $tileCount tiles (${downloadedSize / 1000}kB downloaded, ${cachedSize / 1000}kB already cached) in ${seconds.format(1)}s$failureText")
     }
 
+    // Android Studio incorrectly wants to annotate this with @ExperimentalCoroutinesApi.
+    // The type of the block is ProducerScope<T>.() -> Unit, and we use ProducerScope.send().
+    // ProducerScope is indeed experimental, but `send` is inherited from SendChannel, which isn't.
     private fun downloadTilesFlow(bbox: BoundingBox): Flow<DownloadResult> = channelFlow {
         /* tiles for the highest zoom (=likely current or near current zoom) first,
            because those are the tiles that are likely to be useful first */
--- a/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
@@ -7,6 +7,7 @@ import de.westnordost.streetcomplete.ktx.format
 import de.westnordost.streetcomplete.map.VectorTileProvider
 import de.westnordost.streetcomplete.util.enclosingTilesRect
 import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
 import okhttp3.*
 import okhttp3.internal.Version
 import java.io.IOException
@@ -20,8 +21,6 @@ class MapTilesDownloader @Inject constructor(
 ) {
     private val okHttpClient = OkHttpClient.Builder().cache(cacheConfig.cache).build()
 
-    data class Tile(val zoom: Int, val x: Int, val y: Int)
-
     suspend fun download(bbox: BoundingBox) = withContext(Dispatchers.IO) {
         var tileCount = 0
         var failureCount = 0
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
```diff
@@ -7,90 +7,90 @@ import de.westnordost.streetcomplete.ktx.format
...
         require(httpUrl != null) { "Invalid URL: $url" }
 
-        val builder = Request.Builder()
+        val request = Request.Builder()
             .url(httpUrl)
             .cacheControl(cacheConfig.cacheControl)
-        builder.header("User-Agent", ApplicationConstants.USER_AGENT + " / " + Version.userAgent())
-        val call = okHttpClient.newCall(builder.build())
+            .header("User-Agent", ApplicationConstants.USER_AGENT + " / " + Version.userAgent())
+            .build()
```
smichel17, line 79: This is mostly a matter of style, but I think it is clearer, since it does not requires double-checking what kind of builder it is.

app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
```diff
@@ -7,90 +7,90 @@ import de.westnordost.streetcomplete.ktx.format
...
-                response.body()?.use { body ->
-                    // just get the bytes and let the cache magic do the rest...
-                    size = body.bytes().size
-                }
+                // just get the bytes and let the cache magic do the rest...
+                val size = response.body()?.use { it.bytes().size } ?: 0
                 val alreadyCached = response.cacheResponse() != null
-                val logText = if (alreadyCached) "in cache" else "downloaded"
-                Log.v(TAG, "Tile $zoom/$x/$y $logText")
                 cont.resume(DownloadSuccess(alreadyCached, size))
```
smichel17, line 94: `use` returns the inner return value, so we don't need to use a var here.

I went back and forth quite a bit on whether to replace this whole bit with `cont.resume(response) { response.close() }` and also move the other lines above. I ended up choosing this because it closes the response body as close to the source as possible — lower chance of accidental leaks.

smichel17: I'm also on the fence about whether the `DownloadResult` abstraction is needed at all — kotlin has a built in [`Result`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) class we could use instead, which also includes the exception, in the case of failures, in case we wanted to move the logging all the way up to the top. But this seems good enough; I don't see any practical benefit to that except marginally cleaner code.

app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt
```diff
@@ -7,90 +7,90 @@ import de.westnordost.streetcomplete.ktx.format
...
-        (vectorTileProvider.maxZoom downTo 0).asSequence().flatMap { zoom ->
-            bbox.enclosingTilesRect(zoom).asTilePosSequence().map { Tile(zoom, it.x, it.y) }
+        (vectorTileProvider.maxZoom downTo 0).forEach { zoom ->
+            bbox.enclosingTilesRect(zoom).asTilePosSequence().forEach { (x, y) ->
+                launch {
+                    try {
+                        val result = downloadTile(zoom, x, y)
+                        val logText = if (result.alreadyCached) "in cache" else "downloaded"
+                        Log.v(TAG, "Tile $zoom/$x/$y $logText")
+                        send(result)
```
smichel17, line 58: The `send` call here produces a new warning about not being marked with the `@ExperimentalCoroutinesApi` decorator, which I believe is incorrect.

[`channelFlow`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/channel-flow.html) is not experimental. However, the [`ProducerScope`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-producer-scope/index.html) block that you pass is. As far as I can tell, it's only marked that way because a couple of its functions are experimental — *not* including [`send`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-send-channel/send.html) — but the compiler isn't smart enough to tell that the part of the interface we're using isn't actually experimental.

smichel17: I suppose I should probably include a comment to that effect.

smichel17:
This is ready for review/merging. I have tested it, although not extensively.

westnordost:
Do you have a good link to a guide to understanding flows? I skipped on flows and channels when reading into coroutines because I had issues understanding it. F.e. where is the difference of a flow to a list (or sequence) of `Deferred`s etc

smichel17:
I just read the official docs… several times; I agree they are confusing, especially given that I'm comfortable with concurrency in general.

The **core point** is that flows allow you to abstract the details of how the data is created (sync or async, in one coroutine or many) and process the results in a single coroutine, where you don't need to worry about concurrency. Or in more than one — the point is it depends on the coroutine context of the **call site.** They also provide a bunch of utilities for combining multiple flows.

> difference of a flow to a list (or sequence) of `Deferred`s etc

- Flows are more like a builder, which is evaluated each time you `collect` it, so you can run them multiple times. (Except hot/shared flows, but I didn't look into them yet).
    - Note: It wasn't clear to me from the docs what the full list of functions that would trigger collection are. It's any function [here](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/) that's marked with `suspend`.
- Flows don't *have* to use multiple coroutines, while a list/sequence of `Deferred`s would be one `Job` per item.
- With Channels, Flows have more flexibility in how you build them.
    - Say you wanted to start the Jobs in a certain order, but you wanted to process them in the order that they complete (for example, this PR). This would be annoying with a sequence of `Deferred`s; here we start the downloads in order and `send` back the result (through a `Channel`) when they complete, which is the order we receive them in.
    - If you're familiar with Erlang's message passing system, that's my point of reference for Channels & their usefulness. Maybe there are better docs for it?

Overall, Flows are *almost* a strict improvement over Sequences, except:

1. Flows are newer, so the standard library for them isn't as large (yet?).
2. There is a *slightly* higher overhead to creating a flow than a Sequence. [This post](https://medium.com/mobile-app-development-publication/use-sequence-instead-of-kotlin-flow-when-ad577316ce51) measured 2.5 seconds additional time per 100 million. So not something we have to worry about most of the time, but occasionally it might be worth optimizing.

westnordost:
Hmm, I couldn't completely follow but my brain power is down currently because I got the flu. Thank you for the information, I think I will reread and review it later. This will be something for v34 anyway, so there is no hurry.

smichel17:
Okay, no rush, hope you recover soon :). If you want to chat with a shorter delay, I emailed you some contact info.

Tangent: I wonder how long it will be before GitHub includes a messaging system. I'm honestly a little surprised they didn't buy Gitter.

smichel17:
When you get back to this specific code, I suggest going commit-by-commit.

westnordost:
> Flows are more like a builder, which is evaluated each time you collect it, so you can run them multiple times. 

Okay, this sounds neat. This is one aspect I like about sequence (builders). In `SlippyMapMath.kt`, there is this function

```kotlin
    /** Returns all the individual tiles contained in this tile rect as an iterable sequence */
    fun asTilePosSequence(): Sequence<TilePos> = sequence {
        for (y in top..bottom) {
            for (x in left..right) {
                yield(TilePos(x, y))
            }
        }
    }
```

and I really like the concept of that. 

However, in your code, there is not only a double-for-loop but inside that, there is a `launch`. And this I find confusing. So before the first `send` is reached, which supposedly suspends the code within the `channelFlow` builder, we already launched a number of sub-coroutines. The only point at which these sub-coroutines will suspend - as far as I understand - is at the `send` call, so after they are finished. So, I am thinking... what's the point of this again?

Here are my thoughts on this PR:

<ol>
<li>What does this PR actually solve or improve? It does not introduce a new feature, it does not make the code shorter, it does not make the code better readable and it does not make it more consistent (flows are used nowhere else in the app)

</li>

<li>On the other hand, as mentioned, it introduces another concept from the kotlin coroutines library.  

And to be honest, I am not that satisfied with Kotlin coroutines in general. The easiest stuff (`suspend`, `launch`) looks easy, but the learning curve is steep. It starts with `Context` vs `Scope` vs `Job` that I always tend to confuse with each other, then the thousand different builders and "adapter" functions that help translate to/from Java Concurrency constructs. And then, with structured concurrency and exception handling, it starts to get really complex as exceptions thrown in supervisorScopes/jobs are handled differently than in coroutineScopes/normal jobs and there is some difference in exception handling between `launch` and `async` (deferreds) which I don't remember right now. And this, I'd say, still counts as normal/basic usage of Kotlin coroutines - after all, you **need** structured concurrency, especially in Android context where the app itself has so little control over its own lifecycle, i.e the view might be destroyed any time.  

And here, I am just assuming that I am not a complete blockhead and I can't be the only one who finds Kotlin coroutines really complex to understand and also to recall how things work. So, to try to use only the basic things (in a consistent manner) of the Kotlin coroutines is a deliberate choice I made because I am assuming that if I as the project maintainer already have difficulties to understand this (possible) contributors will have a problem even more. And it is very important to me that the code in this project remains easy to read and to understand (how they work).
</li>
</ol>

<details>
<summary>
P.S: Did you know that I tried in the frame of the big architectural refactor to make "everything" in the app (i.e. all the way up from the data layer) compeletly based on coroutines? 
</summary>
Like...

```kotlin
class MyDao {
  suspend fun getAll(ids: List<Long>): List<Thing>
}
```

and all the way through to controllers, notifying listeners etc.
I was kind of almost through with it, but gave it up. These two discussions on the Kotlin forum were from this time, if you are interested. 

https://discuss.kotlinlang.org/t/coroutines-correct-use-for-class-managing-data/21152
https://discuss.kotlinlang.org/t/suspend-safety-something-to-keep-in-mind/21261

The crux was how to not cancel critical sections of the code that did the data persisting and update (you know, persist nodes, ways, relations -> leads to persisting geometry -> leads to quest generation -> leads to persisting quests etc) because if I wouldn't have done anything, the default behavior would be that this could be cancelled anytime if the view (that triggered this) is being destroyed.
</details>

smichel17:
I agree with your dislike of coroutines in general. Actually, I liked the language as a whole more when it was just "better java" and was not trying to provide an api for ~everything.

That said, each of the commits does have a purpose. The second logs any errors inside the download function, not just connection errors. The first fixes a concurrency bug in recording stats and is half of the reason for using flows. (The other half is making it easier to implement #2794 cleanly.) Yes, it's possible to fix the bug a different way, like using a mutex or an AtomicInteger. But this way allows writing regular, sequential code inside `download`. I think that's nice.

> double-for-loop

Before it was a `flatMap`. I changed it because I thought this was clearer, but I can do it the other way.  I'll share code later when I have more time (probably Saturday).

> supposedly suspends the code within the channelFlow builder

I need to do some re-reading before I respond to the suspending bit.

But we don't really care about *laziness* here. Like before, we queue up all the downloads at once (each in their own coroutine) and let the okHttp client limit the number of concurrent downloads.

smichel17:
Okay, I've confirmed — a **channel**Flow doesn't suspend, anyway. One coroutine (in the example; the PR has more because of the `launch`) is producing events, which get buffered, and the other is consuming them as soon as it's ready. I modified the example [here](https://kotlinlang.org/docs/flow.html#flows) to demonstrate:

<details><summary>Regular flow code & output</summary>

```kt
fun simple(): Flow<Int> = flow { // flow builder
    println("Starting flow")
    for (i in 1..11) {
        delay(10) // pretend we are doing something useful here
        println("Sending $i")
        emit(i) // emit next value
    }
}

fun main() = runBlocking<Unit> {
    // Launch a concurrent coroutine to check if the main thread is blocked
    launch {
        for (k in 1..3) {
            println("I'm not blocked $k")
            delay(100)
        }
    }
    // Collect the flow
    simple().collect { value -> 
        println("Collected $value")
        delay(20)
        println("Processed $value")
    } 
}
```
Output:
```
Starting flow
I'm not blocked 1
Sending 1
Collected 1
Processed 1
Sending 2
Collected 2
Processed 2
Sending 3
Collected 3
Processed 3
I'm not blocked 2
Sending 4
Collected 4
Processed 4
Sending 5
Collected 5
Processed 5
Sending 6
Collected 6
Processed 6
Sending 7
Collected 7
I'm not blocked 3
Processed 7
Sending 8
Collected 8
Processed 8
Sending 9
Collected 9
Processed 9
Sending 10
Collected 10
Processed 10
Sending 11
Collected 11
Processed 11
```
</details>

<details><summary>channelFlow code & output</summary>

```kt
import kotlinx.coroutines.*
import kotlinx.coroutines.flow.*


fun simple(): Flow<Int> = channelFlow { // flow builder
    println("Starting flow")
    for (i in 1..11) {
        delay(10) // pretend we are doing something useful here
        println("Sending $i")
        send(i) // emit next value
    }
}

fun main() = runBlocking<Unit> {
    // Launch a concurrent coroutine to check if the main thread is blocked
    launch {
        for (k in 1..3) {
            println("I'm not blocked $k")
            delay(100)
        }
    }
    // Collect the flow
    simple().collect { value -> 
        println("Collected $value")
        delay(20)
        println("Processed $value")
    } 
}
```
Output:
```
I'm not blocked 1
Starting flow
Sending 1
Collected 1
Sending 2
Sending 3
Processed 1
Collected 2
Sending 4
Sending 5
Processed 2
Collected 3
Sending 6
Processed 3
Collected 4
Sending 7
Sending 8
Processed 4
Collected 5
Sending 9
I'm not blocked 2
Sending 10
Processed 5
Collected 6
Sending 11
Processed 6
Collected 7
Processed 7
Collected 8
Processed 8
Collected 9
Processed 9
Collected 10
I'm not blocked 3
Processed 10
Collected 11
Processed 11
```
</details>

westnordost:
Hmm, okay. I've read your explanations in your second comment about channelFlow and flow. It is interesting, but I don't quite get it, though, to be honest. I don't see a reason at all why the "I am not blocked" loop should be blocked at all by anything, after all, it is launched in a sub-coroutine.

Anyway, at least for now, I will not want anything with `flow`s within the app. I mentioned the reasons above (in _"Here are my thoughts on this PR"_)

PR #3015, opened 2021-07-02
Change "Postman" achievement to "You've got mail!"

jancborchardt:
Based on the discussion at https://github.com/streetcomplete/StreetComplete/pull/3004


westnordost:
See #3009

jancborchardt:
I assumed cause you said

> Let's not spend so much time on this issue

(which I agree with), that a simple pull request to just fix it to that suggestion would be the best way to go forward.

PR #3036, opened 2021-07-08
Change "weekdays" to "days"

### LINKED ISSUE #3035: Use of "weekdays" is not correct in English
In several quests (such as the postbox collection times quest) there are dialogues which refer to "weekdays". These are used to refer to all days of the week, but in English the word "weekday" usually refers only to Monday to Friday.

From [Wikipedia](https://en.wikipedia.org/wiki/Workweek_and_weekend):

    A weekday or workday is any day of the working week.

In usual British English at least, the correct word would simply be "days", e.g.
- "Select weekdays" -> "Select days"
- "Add weekdays" -> "Add days"

**How to Reproduce**
On one of the affected quests, click "Add Opening Hours" and the dialog shows as this
image

**Versions affected**
StreetComplete v33.0

Xcodo:
Fixes #3035


HolgerJeromin:
Issue is closed already

Issue #2145, opened 2020-10-08
Splitting ways corrupts step_count


DerDings:
When a way with a value for `step_count` (e.g. `step_count=5`) is splitted, two ways are created with the same value for `step_count`, which leads to the total steps of both ways being doubled (two ways with `step_count=5` in a row => 10 steps in total) 

**How to Reproduce**
1. Find steps that can be splitted, e.g. by the surface quest: https://overpass-turbo.eu/s/YQR
2. Split the way with StreetComplete

**Versions affected**
v24.0 (probably among others)

**Possible solutions**

1. disable splitting ways which have a numeric value for `step_count`.
2. when splitting a way with 'step_count', immediately ask how many steps there are in each segment.

DerDings:
related to #2144

westnordost:
Oh damn, didn't think of that! Should have waited a day with the update!

Any other tags that come to your mind that should change their value based on the length of way?

westnordost:
Btw might be a problem in every other editor too. Should be checked and a bugticket created for each where the problem exists as well.

Maybe it would be better to delete the offending key rather than trying to "correct" it.

DerDings:
> Any other tags that come to your mind that should change their value based on the length of way?

`duration=*`, but that is very uncommon (and discouraged) with regular highways. Maybe it is used for some escalators or moving walkways.

(In my opinion, escalators shouldn't be splitted at all, because they're not only a chunk of highway, but also ONE machine. But that should be discussed in the community, not here.)

DerDings:
> Btw might be a problem in every other editor too. Should be checked and a bugticket created for each where the problem exists as well.

I just tried Vespucci, same problem.

matkoniecz:
> Maybe it would be better to delete the offending key rather than trying to "correct" it.

possible alternative from https://github.com/openstreetmap/iD/issues/8069#issuecomment-706176064 :

> I think iD can be smart enough here to distribute the step count proportionally to the resultant lengths. Sure it may not be 100% accurate, but it should be pretty close and it preserves the total step count.

DerDings:
> > Maybe it would be better to delete the offending key rather than trying to "correct" it.
> 
> possible alternative from https://github.com/openstreetmap/iD/issues/8069#issuecomment-706176064 :
> 
> > I think iD can be smart enough here to distribute the step count proportionally to the resultant lengths. Sure it may not be 100% accurate, but it should be pretty close and it preserves the total step count.

[...]

PR #3037, opened 2021-07-09
Allow handrail quest to split the way

peternewman:
Because handrails may not exist the whole way along a staircase.

Untested but trivial, I assume this automagically deals with the can't say/other answers switching without additional changes?


HolgerJeromin:
This should/must remove the step count from the steps.
Perhaps this was the reason for deactivating.

peternewman:
> This should/must remove the step count from the steps.

Yes, agreed, it was discussed in #2145 and fixed in 8683f4ccca265a2bd908ca1794eb1b2d3085ea35 .

> Perhaps this was the reason for deactivating.

All the other step related quests already allow splitting (I thought it was surface I was having issues with at first), so I don't believe that's the reason.

peternewman:
Thanks @westnordost .

PR #3039, opened 2021-07-09
Add building=bungalow to buildings with an address

moreati:
In the UK at least a bungalow is most often a residential house, that
happens to be single story.


westnordost:
Then the housenumber quest will also be asked for every single bungalow used for vacation purposes, which seems to be the prevalent use according to the wiki documentation

westnordost:
Using `building=bungalow` for a single-storey detached seems to make no sense to me, after all how many levels a building has is a separate information: `building:levels=1`

westnordost:
So I am closing this one. The wiki also seems to be rather describe the bungalow as a holiday home, not a special type of detached house (even though it also links to the wikipedia article)

westnordost:
But I added some explanatory text in the wiki

PR #3041, opened 2021-07-10
In Croatia, enable cycleways and tactile pavings

mnalis:
In Croatia:

- there are lots of cycleways
- tactile pavings are on the rise (about 18% with non-no answer)
  http://overpass-turbo.eu/s/19iZ


westnordost:
ok, since you are in Croatia, you are the one that can tell

PR #3042, opened 2021-07-10
mismatch of ToC / header names breaks anchor navigation in documentation

mnalis:
It was called `Issues of dependencies` in table of content, but `Issues with dependencies` as an actual section name.


westnordost:
thanks

Issue #2522, opened 2021-01-21
GitHub Actions workflow for build and unit tests


riQQ:
Would you accept a pull request for a GitHub actions workflow that runs build and unit tests?

I saw that there once was continous integration via Travis CI. Any reason why this was abandoned?

FloEdelmann:
See also https://github.com/streetcomplete/StreetComplete/pull/1917

riQQ:
I would start with a simple workflow that builds and runs unit test to have a base level of CI. It can always be expanded upon later.

westnordost:
I don't see the necessity for this. I do run Android lint, the tests etc before each release.

Also, currently some Android tests always fail, see https://github.com/streetcomplete/StreetComplete/issues/2028  so the CI would just report for everything that tests fail. That's not useful. It would be useful to do something about #2028 or its successor-ticket.

peternewman:
> I don't see the necessity for this. I do run Android lint, the tests etc before each release.

Obviously it's up to you, but it means it puts the onus on linting and passing tests on each contributor, rather than you having to rework their contributions during each release. It also means if people choose to run the lint and tests locally, they can be confident they'll pass.

I've ran CI (and added it to lots of repos) with general success I'd say.

I don't really know how the Android tests work, but it also potentially means you can run the test suite against the newest and oldest supported versions of Android that SC should cope with.

> Also, currently some Android tests always fail

Yeah obviously it relies on the tests passing to be useful, or being able to run a subset of them which do pass.

riQQ:
When you have several contributions and one breaks a test, in all the following contributions you don't have 100% certainty that you didn't break anything additionally. Of course this only concerns areas covered by tests.

I don't know how you handle PRs. But if you pull the code and run the tests locally, that is something that could be done by CI.

PR #3046, opened 2021-07-10
Add GitHub workflow to check the project builds

TurnrDev:
Checks on PR and push to `master`

Might be worth adding a workflow to check code-cov and tests too?

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -1,0 +1,16 @@
+name: Build
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


mnalis:
@TurnrDev  See also previous attempts at https://github.com/streetcomplete/StreetComplete/issues/2522 , https://github.com/streetcomplete/StreetComplete/pull/1917

TurnrDev:
Some good points made, closing my PR

PR #3055, opened 2021-07-12
Fix the incorrect ToC, not the corrected heading

peternewman:
Correctly fixes #3042

Sorry, I'm not used to manual ToC given they've been automated in Mediawiki for decades!

PR #3058, opened 2021-07-12
Fix remaining references of #issues-with-dependencies

mnalis:
fixes incomplete fix in https://github.com/streetcomplete/StreetComplete/pull/3055  (not all `#issues-of-dependencies` were changed to `#issues-with-dependencies`, which broke navigation again).


peternewman:
Thanks @mnalis . Perhaps we should have left all the refs as they were for backwards compatibility and just changed the display names.

mnalis:
@peternewman Well, we couldn't have done that, as MD seems to auto-generate those references from the title text automagically (lowercasing, replacing spaces with hyphens, etc.) - anyway it looks good now so it's hopefully OK :+1:

peternewman:
Oh that's odd, I thought we'd manually updated it. Anyway all sorted now as you say.

PR #3073, opened 2021-07-16
Allow Kosher quest to answer it does not exist too

peternewman:
Untested but trivial, from what I could see in 6164ebb940404010d2e6a86e17512ca33e190199, this automagically deals with everything else.


westnordost:
jup, thanks

PR #3077, opened 2021-07-18
new quest to set building flat count

### LINKED ISSUE #1152: building:flats where building=apartments
### General
Affected tag(s) to be modified/added: [building:flats](https://wiki.openstreetmap.org/wiki/Key:building:flats)
Question asked: How many units are there in this apartment building?

### Checklist
Yes has 1.6M usages already and is uselful in understanding the density of the apartment.

Most apartment buildings have letterboxes for each unit at the entrance, a count of these will give the answer.

Yes.

### Ideas for implementation
**Element selection:** `building=apartments AND building:flats IS NULL` (please excuse my SQL and not OverpassQL)

**Metadata needed:** I don't think so

**Proposed GUI:** A simple numeric input.

### LINKED ISSUE #3076: How many units are there in this apartment building?
### General
Affected tag(s) to be modified/added: [building:flats](https://wiki.openstreetmap.org/wiki/Key%3Abuilding%3Aflats)
Question asked: **How many units are there in this apartment building?**
Or in royal English: how many flats in this block geez

### Checklist
Checklist for quest suggestions (see [guidelines](https://github.com/streetcomplete/StreetComplete/wiki/Adding-new-Quests-to-StreetComplete):
Although the key is valid on building=house,  I suggest limiting it to building=apartments to reduce spam.
Usually by a sign or by counting door buzzers

### Ideas for implementation

On building=apartments, ask how many flats/units exist in the apartment block.

andrewharvey:
WIP addr:flats

I know this was rejected, but someone else asked for it recently so I though I should open up this code.

closes #3076
closes #1152

This is untested code, I need to test it more and try it out.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -129,7 +129,7 @@
     <string name="quest_openingHours_comment_description">"Be concise, ideally use the wording from the sign, e.g. “by appointment”. The absence of a sign does not necessarily mean that there are no regular opening hours."</string>
     <string name="quest_buildingLevels_title">"How many levels does this building have?"</string>
     <string name="quest_buildingLevels_title_buildingPart">"How many levels does this building part have?"</string>
-    <string name="quest_buildingFlatCount_title">"How units does this building have?"</string>
+    <string name="quest_buildingFlatCount_title">"How many units does this building have?"</string>
     <string name="quest_roofShape_title">"What basic shape does this building’s roof have?"</string>
     <string name="quest_roofShape_select_one">"Select one:"</string>
     <string name="quest_roofShape_show_more">"Show more…"</string><!--it is used in generic list, not only for roof shape but rename would break translations-->
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -129,6 +129,7 @@
     <string name="quest_openingHours_comment_description">"Be concise, ideally use the wording from the sign, e.g. “by appointment”. The absence of a sign does not necessarily mean that there are no regular opening hours."</string>
     <string name="quest_buildingLevels_title">"How many levels does this building have?"</string>
     <string name="quest_buildingLevels_title_buildingPart">"How many levels does this building part have?"</string>
+    <string name="quest_buildingFlatCount_title">"How units does this building have?"</string>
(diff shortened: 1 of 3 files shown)


westnordost:
#1152 is already closed. Did anything change in the meantime?

andrewharvey:
I took another read of that ticket to try and address the concerns raised, I'm still not seeing anything there as a reason not to include this quest. I think it's 100% fine to decide not to enable this quest by default because of your concerns, but can't see why we could not allow this quest to be included as an opt-in quest.

app/src/main/res/values/strings.xml
```diff
@@ -129,6 +129,7 @@
     <string name="quest_openingHours_comment_description">"Be concise, ideally use the wording from the sign, e.g. “by appointment”. The absence of a sign does not necessarily mean that there are no regular opening hours."</string>
     <string name="quest_buildingLevels_title">"How many levels does this building have?"</string>
     <string name="quest_buildingLevels_title_buildingPart">"How many levels does this building part have?"</string>
+    <string name="quest_buildingFlatCount_title">"How units does this building have?"</string>
```
FloEdelmann, line 132:
```suggestion
    <string name="quest_buildingFlatCount_title">"How many units does this building have?"</string>
```

westnordost:
I'll answer in #1152

westnordost:
I've decided to not have this in the app, sorry. Reasons are stated in #1152. Summary:
- too spammy (asked for every single apartment building) combined with too little information value, also, this information is not used anywhere yet
- (small issue:) unclear what to tag when this is not visible from the outside
- adding flat numbers as mentioned in #3064 for countries in which flat numbers are visible from the outside bears more information value (because it is part of the address) while when this information is provided also making the flat numbers even less useful
- (small issue that however affects #3064 too:) According to my experience, the actual apartment building staircase entrance is often already within private property and the footway leading to the entrances is signposted like "private property, no entry" while at least the housenumbers are visible from the public sidewalk

matkoniecz:
> According to my experience, the actual apartment building staircase entrance is often already within private property and the footway leading to the entrances is signposted like "private property, no entry"

Skip entrances which have footway with `access=private` on it? It will not catch all of them. But many of the and will provide way to stop quest from reappearing in SC.

PR #3081, opened 2021-07-19
Vegetarian: skip if diet:vegan=only set already

### LINKED ISSUE #3072: Do not ask Vegetarian Quest for POI with diet:vegan=only
**Use case**
Right now even if a POI has diet:vegan=only Vegetarian Quest is asked.
According to https://wiki.openstreetmap.org/wiki/Key:diet:vegan diet:vegan=only implies diet:vegetarian=only (makes sense).
Example of POI where Quest is asked: https://www.openstreetmap.org/node/8359986392
Example of POI where I responded: https://www.openstreetmap.org/changeset/104325841

**Proposed Solution**
I would like to replace https://github.com/streetcomplete/StreetComplete/blob/bb71ae91976ce4f16c22c7ce09ecdcc388a649e1/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt#L13 to `(!diet:vegetarian and diet:vegan != only)`

I haven't checked whether the syntax is correct and whether it works yet.
I am happy to make PR if it makes sense

starsep:
Fixes #3072

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
@@ -9,7 +9,7 @@ class AddVegetarian : OsmFilterQuestType<DietAvailability>() {
 
     override val elementFilter = """
         nodes, ways with amenity ~ restaurant|cafe|fast_food
-        and name and (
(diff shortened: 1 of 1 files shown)


westnordost (approved):

westnordost:
Looks fine, thank you!

Issue #2744, opened 2021-04-14
Check for existence of more things


arrival-spring:
I think the following things should also be checked for existence.

These should probably be checked every 2 years:
- Barbecues (`amenity=bbq`) ~25000 uses and is an iD preset
  - There could also be a quest about whether or not the bbq is covered or what fuel it uses, but that would be a separate issue
- Defibrillator (`emergency=defibrillator`) ~37000 uses and is an iD preset
  - There's the quest asking if defibrillators are indoors, but if that had already been set then the object wouldn't otherwise be changed by SC

And these every 4 years:
- Life rings (`emergency=life_ring`) ~4000 uses and is an iD preset
- Emergency phones (`emergency=phone`) ~52000 uses and is an iD preset
- Flagpole (`man_made=flagpole`) ~63000 uses and is an iD preset

I'm not fully sure on which ones should be 2 or 4 years, but I think it would be good to check for the existence of all of these at some interval.

westnordost:
`emergency=defibrillator` will be difficult as they are often in buildings (and small). We wouldn't want a defibrillator be deleted only because a surveyor wasn't able to see it from the outside and/or was looking in the wrong room for it.

For the rest, I'd say that it makes sense, yeah.

arrival-spring:
That makes sense, although if someone can't find a defibrillator when looking for it then that would be a problem.

matkoniecz:
`man_made=flagpole` and `emergency=phone` - what about 6 or 8 years? This things are very stable (at least in my experience).

arrival-spring:
I'd say that emergency phones are probably more stable than flagpoles in my experience.
I'd suggested 2 and 4 years because those are the only current timescales for checking existence.

arrival-spring:
@matkoniecz does this mean you are working on this?

matkoniecz:
@arrival-spring For now I implemented it and will be testing is it a good idea (testing is more time-consuming part).

If I will think that it is a good idea, I will make PR (or commit directly if sure).

But if you are interested - feel free to implement, test and make a PR.

Lukas458:
I think amenity=ticket_validator` might also be worth resurveying. Similar to `amenity=vending_machine` + `vending=public_transport_tickets`, ticket validators get sometimes removed in these days because people are using more eletric/internet-based solutions for buying pt-tickets, so ticket validators aren't needed so much anymore.

I could look for that `amenity=ticket_validator` gets an iD preset

[...]

PR #3089, opened 2021-07-22
Deseason

matkoniecz:
Idea from https://github.com/streetcomplete/StreetComplete/issues/2744#issuecomment-885065544

Tested on https://www.openstreetmap.org/node/2894090978#map=15/50.8365/13.2820 and some season-less nodes


westnordost:
Makes sense

Lukas458:
@matkoniecz @westnordost Thank you very much! :-)

PR #3097, opened 2021-07-26
Ask about vegetarian and vegan food at pubs

peternewman:
Untested but pretty trivial

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
@@ -10,7 +10,7 @@ class AddVegan : OsmFilterQuestType<DietAvailability>() {
     override val elementFilter = """
         nodes, ways with
         (
-          amenity ~ restaurant|cafe|fast_food and diet:vegetarian ~ yes|only
+          amenity ~ restaurant|cafe|fast_food|pub and diet:vegetarian ~ yes|only
           or amenity = ice_cream
         )
         and name and (
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
@@ -8,7 +8,7 @@ import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChanges
 class AddVegetarian : OsmFilterQuestType<DietAvailability>() {
 
     override val elementFilter = """
-        nodes, ways with amenity ~ restaurant|cafe|fast_food
+        nodes, ways with amenity ~ restaurant|cafe|fast_food|pub
         and name and diet:vegan != only and (
           !diet:vegetarian
           or diet:vegetarian != only and diet:vegetarian older today -2 years


westnordost (approved):

westnordost:
makes sense, yeah

peternewman:
Thanks. I did have a minor fear about no food pubs, but that doesn't match the ones I see.

I've not checked about Kosher, but would need to read up on Jews and alcohol to see if it's even relevant.

westnordost:
Oh hmm, right, I didn't think about that. I only thought about "pub", as in Britain. But in Germany, `amenity=pub` are mostly "Kneipen", which are like bars for old people and/or smokers (feel like it). For most, there is no food.

Let's change that to only pubs that have `food=yes`.

westnordost:
Though even though mentioned in the wiki, `food=*` is tagged on less than 0.5% of all pubs (because it doesn't show up in taginfo at all)

peternewman:
> Oh hmm, right, I didn't think about that. I only thought about "pub", as in Britain. But in Germany, `amenity=pub` are mostly "Kneipen", which are like bars for old people and/or smokers (feel like it). For most, there is no food.

Aside from the no food bit, that pretty much matches a lot of British pubs too!

That also explains the contradictions all over the place in the wiki around whether pubs do or don't have food, it presumably depends which nationality wrote the text.

> Let's change that to only pubs that have `food=yes`.
> Though even though mentioned in the wiki, `food=*` is tagged on less than 0.5% of all pubs (because it doesn't show up in taginfo at all)

From memory we can't easily include or exclude particular query terms by country can we? Or with different queries by country?

I could propose the obvious food quest, but the tag seem to have limited usage currently...

westnordost:
> From memory we can't easily include or exclude particular query terms by country can we? Or with different queries by country?

No

westnordost:
The `food` tag is documented and seems to make sense. Maybe a quest could be added for that (I have already an icon for that! The beer icon!) but maybe... would you mind asking on the tagging mailing list if there are any reservations against using this tag? (as it is not really used yet)

westnordost:
Or @matkoniecz, would you? I am not on the list (too much traffic, awkward/obsolete media format)

peternewman:
> The `food` tag is documented and seems to make sense. Maybe a quest could be added for that (I have already an icon for that! The beer icon!)

Shouldn't that be saved for a draught beer quest or something more important? :smile: As well as possibly applying the food quest to other non-pub things in future?

> Or @matkoniecz, would you? I am not on the list (too much traffic, awkward/obsolete media format)

Seconded please, I'm not either.

peternewman:
Do you want me to revert this for now or open a new PR which requires `food!=no` too for pub or wait and see about the tagging list?

matkoniecz:
> Or @matkoniecz, would you? I am not on the list (too much traffic, awkward/obsolete media format)

Done! For "too much traffic" - I have a separate email account for that (used to have on my main one - was far too distracting)

westnordost:
Thanks! Though, more important than what some people consider helpful or not, I'd find if the tag could be viewed in any way problematic. 

@peternewman I think for now it would be good to open a PR to only ask for pubs with food=yes. Would you do that?

peternewman:
> @peternewman I think for now it would be good to open a PR to only ask for pubs with food=yes. Would you do that?

No problem, see #3099

codeinabox:
> The `food` tag is documented and seems to make sense. Maybe a quest could be added for that (I have already an icon for that! The beer icon!)

I would love to see a quest asking if a bar or pub serves food, as in meals not bar snacks. Currently I have to add food=yes manually with another editor.

matkoniecz:
You can propose a new quest ( https://github.com/streetcomplete/StreetComplete/issues/new?assignees=&labels=&template=quest-suggestion.md  via https://github.com/streetcomplete/StreetComplete/issues/ ) if it matches requirements.

(no idea about this specific case)

peternewman:
@codeinabox for some background discussion on that sort of quest, see https://github.com/streetcomplete/StreetComplete/pull/3099 and https://github.com/streetcomplete/StreetComplete/issues/3408 .

The initial first challenge is agreeing what counts as food. But yes I'd certainly agree that this is annoying:

> Currently I have to add food=yes manually with another editor.

PR #3099, opened 2021-07-28
Only ask about vegan and vegetarian diet for pubs serving food, not all pubs

peternewman:
See discussion in https://github.com/streetcomplete/StreetComplete/pull/3097

Untested, but I think I've got the logic right, I'd normally throw more brackets at it to make sure, but I based it off the lit way quest.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
@@ -10,8 +10,12 @@ class AddVegan : OsmFilterQuestType<DietAvailability>() {
     override val elementFilter = """
         nodes, ways with
         (
-          amenity ~ restaurant|cafe|fast_food|pub and diet:vegetarian ~ yes|only
-          or amenity = ice_cream
+          amenity = ice_cream
+          or diet:vegetarian ~ yes|only and
+          (
+            amenity ~ restaurant|cafe|fast_food
+            or amenity = pub and food != no
+          )
         )
         and name and (
           !diet:vegan
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
@@ -14,7 +14,7 @@ class AddVegan : OsmFilterQuestType<DietAvailability>() {
           or diet:vegetarian ~ yes|only and
           (
             amenity ~ restaurant|cafe|fast_food
-            or amenity = pub and food != no
+            or amenity = pub and food = yes
           )
         )
         and name and (
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
@@ -8,7 +8,11 @@ import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChanges
 class AddVegetarian : OsmFilterQuestType<DietAvailability>() {
 
     override val elementFilter = """
-        nodes, ways with amenity ~ restaurant|cafe|fast_food|pub
+        nodes, ways with
+        (
+          amenity ~ restaurant|cafe|fast_food
+          or amenity = pub and food != no
+        )
         and name and diet:vegan != only and (
           !diet:vegetarian
           or diet:vegetarian != only and diet:vegetarian older today -2 years
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
@@ -11,7 +11,7 @@ class AddVegetarian : OsmFilterQuestType<DietAvailability>() {
         nodes, ways with
         (
           amenity ~ restaurant|cafe|fast_food
-          or amenity = pub and food != no
+          or amenity = pub and food = yes
         )
         and name and diet:vegan != only and (
           !diet:vegetarian


peternewman:
Pending discussion in https://lists.openstreetmap.org/pipermail/tagging/2021-July/062106.html

Although seemingly even this is a little controversial for the food=yes/no quest at least...

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
```diff
@@ -10,8 +10,12 @@ class AddVegan : OsmFilterQuestType<DietAvailability>() {
     override val elementFilter = """
         nodes, ways with
         (
-          amenity ~ restaurant|cafe|fast_food|pub and diet:vegetarian ~ yes|only
-          or amenity = ice_cream
+          amenity = ice_cream
+          or diet:vegetarian ~ yes|only and
+          (
+            amenity ~ restaurant|cafe|fast_food
+            or amenity = pub and food != no
```
westnordost, line 17: should be `food=yes`, not `food != no`. Otherwise this is also asked if `food` is missing.

westnordost (changes requested):

westnordost:
> Although seemingly even this is a little controversial for the food=yes/no quest at least...

From what I have read so far, not that much. If has been noted that it should be clear what is food and what is not and so far people said that snacks are not food, so I'd go with that definition, i.e. a quest of that sort should make it clear that crisps and nachos don't count.

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
peternewman: Would `food and food != no` work? There are currently 348 values which aren't no but appear to be positive/true.

peternewman:
> > Although seemingly even this is a little controversial for the food=yes/no quest at least...

I also realised this doesn't block this PR being merged as such, it just makes it a bit ineffective on it's own.
 
> From what I have read so far, not that much. If has been noted that it should be clear what is food and what is not and so far people said that snacks are not food, so I'd go with that definition, i.e. a quest of that sort should make it clear that crisps and nachos don't count.

Personally I'd say nachos, when coated with cheese and guacamole and chilli would, but I guess that's the somewhat subjective bit. I think we're all in agreement about crisps/potato chips and pork scratchings being out anyway.

Incidentally, I think part of the issue is inconsistency like this:
https://wiki.openstreetmap.org/wiki/Key%3Afood#Amenities_typically_providing_food - "Also sells food which also can be eaten on the premises."
https://wiki.openstreetmap.org/wiki/Tag%3Aamenity%3Dpub - "Pubs commonly sell food which also can be eaten on the premises."/"food=yes - If the pub serves food, or type of food thereof"
https://wiki.openstreetmap.org/wiki/Key:amenity#Sustenance - "may also provide food or accommodation (UK)"

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
westnordost: but are you sure that those values are/would be used on `amenity=pub`? Is maybe the `food` key also used in another context?

westnordost:
On what counts as food... maybe it would make sense to let the user answer when asked

proper food | just snacks | nothing

with the first being `food=yes`, the second something like `snacks=yes`?

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegan.kt
peternewman: Pizza, which looked like the most popular likely value, has four hits, when I do a query; restaurant has 7. So aren't we better off including them too? Even if they weren't used on pub now, presumably they'd show up in iD and whatever as suggestions when you used the food tag, so they might get added if they weren't present already.

peternewman:
> On what counts as food... maybe it would make sense to let the user answer when asked

Yeah that would certainly help make clear what they meant.

> with the first being `food=yes`, the second something like `snacks=yes`?

I almost wonder if the second one should be `food=snacks` given if it sells proper food, you could snack on a bowl of chips so they're pretty much mutually exclusive? A handful of people are using it already ( https://taginfo.openstreetmap.org/tags/food=snacks ) including on pubs. Also if there is some snack tagging, we probably still want to ask about vegetarian/vegan aspect of it, although a lot of meat flavour crisps are actually suitable for vegetarians!

westnordost:
> I almost wonder if the second one should be food=snacks 

hmm... there is a problem with that. That would mean that `food=yes` needed to be interpreted as that  is just "unspecified" what food, i.e. could also be "snacks".

peternewman:
> > I almost wonder if the second one should be food=snacks
> 
> hmm... there is a problem with that. That would mean that `food=yes` needed to be interpreted as that is just "unspecified" what food, i.e. could also be "snacks".

Ah yes, you'd then need a whole set of values for levels of food such as nibbles, snacks, full, banquet with all the confusion about where something falls which that would entail.

Although equally it does seem a bit odd to me if we end up with tagging:
```
food=no
snacks=yes
```

Whereas something like this seems better from a logical point of view:
```
food=yes
bountifulness=snacks
```

> The "more specific" ones should actually be covered by different tags...
> Almost 90% of this so far not that much tag has the value of either "no" or "yes".

Yes, I agree they could certainly be tagged in better ways, however none of them (even `designated`), imply the place doesn't serve something edible and therefore shouldn't we still ask if that edible stuff is vegetarian etc?

Then a separate food quest could fix up the cuisine mis-tagging, or just retag as `food=yes`, although I'm not sure exactly how you'd do that in SC without just removing the fact a place sells pizza (but perhaps nothing apart from the website can represent that as the tag won't be exposed in anything more specific).

But I'm rambling, my main point regarding https://github.com/streetcomplete/StreetComplete/pull/3099#discussion_r679384509 is we should therefore accept a broad church of mistagging or creative tagging for *this* particular quest (and fix it in a future food one).

westnordost:
Let's just accept food=yes

westnordost:
because

> Yes, I agree they could certainly be tagged in better ways, however none of them (even designated), imply the place doesn't serve something edible and therefore shouldn't we still ask if that edible stuff is vegetarian etc?

We don't know what these values mean. Not documented. We can only assume. And note, that places where `food=*` is used so far are other things, not `amenity=pub`, so far. So maybe `food=restaurant` is used in a different context. Maybe as an addition to `tourism=hotel`? As said, only 10% of the usages of this tag use not `yes` or `no` and I don't want to let the app speculate what the other values mean when it is not clear and documented.

peternewman:
Thanks @westnordost , sorry I was away when you replied and I've only just got back so thanks for fixing and merging.

PR #3104, opened 2021-07-30
Count a node as a kerb even if it has a kerb= tag.

jyasskin:
Otherwise, if the AddKerbHeight quest is ordered before AddTactilePavingKerb, the latter will never apply.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/kerb_height/KerbUtil.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/kerb_height/KerbUtil.kt
@@ -37,7 +37,7 @@ private val allowedKeysOnKerbNode = setOf(
     "sloped_curb",
     "tactile_paving",
     "surface", "smoothness", "material",
-    "kerb:height", "height",
+    "kerb", "kerb:height", "height",
     // access/eligibility-related
     "wheelchair", "bicycle", "foot", "stroller",
     // misc / meta info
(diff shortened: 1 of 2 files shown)


westnordost:
Hmm, the kerb quest(s) went through many iterations and I am pretty sure everything there is has a good reason, but I'll look into the code, maybe I overlooked this!

westnordost:
Jop, you are right! This could actually be considered a bug, thank you for finding this and offering this fix!

I noticed that there is a nother bug: Both kerb quests set `check_date:kerb` / `check_date:tactile_paving` on the node. This would also be considered as "not allowed on a kerb node". So what I should do is to make the `allowedKeysOnKerbNode` regex-enabled and allow any `check_date:.*` too.

I'll do this after merging your PR.

westnordost:
I should also check if the... `couldBeAKerb()` actually leads to a performance improvement, because this is mostly what this is about, it only pre-filters the candidates

westnordost:
Ah yes, it is definitely a performance improvement. Not on download, but when a node is updated: 

After answering a question, the element is updated. When an element is udpated, it is checked for every quest if it is now eligible (or not eligible any more), which also works offline. This is done first by looking at the element in isolation (i.e., just the tags). This is fast.

However, if it is not possible to say for a quest if it is now eligible or not just by looking at the tags because it depends on the surroundings (=geometry), the app retrieves the map data in a (small) bounding box around the elemet in question from the database and uses the same algorithm as for the download. This takes a long time, because this requires multiple disk drive accesses (varies a lot, often between 50ms and 500ms).

Now, **any** node (f.e. one just specified the opening hours of a hairdresser, thus updated the node) is a potential kerb candidate when looked at isolation, i.e. it is not possible to say if a node could be a kerb by just looking at the node without the surrounding geometry, the retrieving of that bounding box would happen almost always when answering any quest. This is unacceptable, thus, if the updated node has any tag that would not be expected on a kerb, it is filtered out.

Now of course, what I could also do to make the retrieval of the bounding box faster would be to also cache all the downloaded map data in memory, like all other openstreetmap editor apps do most probably, but that would probably use a lot of RAM and add complexity to the `MapDataController` (also to keep it thread-safe, i.e. guard against race-conditions), so I did not try that yet but I might at one point.

I am going to link this in a comment from `KerbUtil.kt` so people (and I) know why that piece of code is there in the future.

PR #3111, opened 2021-08-02
Give users a "this crossing doesn't exist" option in the crossing-type quest.

### LINKED ISSUE #3106: AddCrossingType should have "no crossing" and "forbidden" answers
**Use case**
I've been describing crosswalks in my neighborhood by drawing sidewalks at home using aerial photography, and then walking around with StreetComplete to fill in details. Sometimes the aerial photography is unclear, and I wind up drawing a crossing where there isn't one, but then the StreetComplete quests aren't enough to correct that during my survey. There are two kinds of mistakes:

1) The crossing just doesn't exist. This has 3 sub-cases:
   1. https://www.openstreetmap.org/note/2777232 was a place where one traffic light controls both parts of the street, and pedestrians have to cross on the outside edges of the overall intersection.

   1. https://www.openstreetmap.org/changeset/108846698 is a little more ambiguous: the street geometry is the same, but there's no traffic light. Pedestrians should still cross on the outside edges.

   1. https://www.bing.com/maps?osid=f592579f-02d5-4684-9ace-c3a780f5b2d9&cp=45.514298~-122.622983&lvl=19&dir=62.119&pi=0&style=x&mo=z.0&v=2&sV=2&form=S00027 is even more ambiguous. There's a legal crossing here, but as you can see from the ghost bike, it's a really bad idea, and pedestrians should walk a block to the light-controlled intersections a block away.

I don't think the app needs to have an opinion on which of these should omit the `highway=crossing` tag, although I don't oppose an opinion either. The wiki doesn't express an opinion, and it's worth having the option in order to support the first case above.

2) The crossing has a sign forbidding crossing there: https://www.openstreetmap.org/note/2773674 / Photograph of a no crossing sign. 

**Proposed Solution**

For (1), I'd remove the `highway=crossing` node and `footway=crossing` way, and if there's a short `footway=sidewalk` segment attached to the crossing way whose other end is another sidewalk, ideally that should also be removed.

For (2), I'd do the above, and also add a `crossing=no` tag.

It's also plausible to collapse (1) and (2) and always add the `crossing=no` tag.

jyasskin:
Fixes #3106

This just opens a note with a suggestion of what to write, but it's clearer than the status quo in saying a note is the right way to handle this problem.

When the only aspect of the crossing is a `highway=crossing` node, without a `footway=crossing` way, I think we could implement this without a note by removing the `highway=crossing` tag (and leaving the node), but conversation on the issue seemed unfavorable to being even that aggressive.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -363,8 +363,8 @@ class MainFragment : Fragment(R.layout.fragment_main),
         }
     }
 
-    override fun onComposeNote(questKey: QuestKey, questTitle: String) {
-        showInBottomSheet(LeaveNoteInsteadFragment.create(questKey, questTitle))
+    override fun onComposeNote(questKey: QuestKey, questTitle: String, initialMessage: String?) {
+        showInBottomSheet(LeaveNoteInsteadFragment.create(questKey, questTitle, initialMessage))
     }
 
     override fun onSplitWay(osmQuestKey: OsmQuestKey) {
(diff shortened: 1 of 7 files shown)


westnordost:
I am sorry that you went trhough the effort to implement this. I was on vacation over the weekend so I couldn't (properly) look at the issue earlier. See my comment in #3106

jyasskin:
No worries, if I'd cared about never wasting some programming, I'd have waited for your reply in the issue. :)

PR #3117, opened 2021-08-04
Sort building levels recents by regular, then roof

smichel17:
The new building levels recent values is great. However, I've found that the vast majority of the buildings in the areas I map are 5 values: 1/0, 1/1, 2/0, 2/1, and -- probably to the surprise of no one -- 3/0. With that in mind, a stabler sort order would be much easier for me to scan and find the entry I'm looking for than the current FIFO queue.

~~I haven't tried this yet, going to do it this morning.~~ Tried it, works as intended.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
@@ -31,7 +31,9 @@ class AddBuildingLevelsForm : AbstractQuestFormAnswerFragment<BuildingLevelsAnsw
     private val roofLevels get() = roofLevelsInput?.text?.toString().orEmpty().trim()
 
     private val lastPickedAnswers by lazy {
-        favs.get(javaClass.simpleName).map { it.toBuildingLevelAnswer() }
+        favs.get(javaClass.simpleName).map { it.toBuildingLevelAnswer() }.sortedWith(
+            compareBy<BuildingLevelsAnswer> { it.levels }.thenBy { it.roofLevels }
+        )
     }
 
     @Inject internal lateinit var favs: LastPickedValuesStore<String>


app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -31,7 +31,12 @@ class AddBuildingLevelsForm : AbstractQuestFormAnswerFragment<BuildingLevelsAnsw
     private val roofLevels get() = roofLevelsInput?.text?.toString().orEmpty().trim()
 
     private val lastPickedAnswers by lazy {
-        favs.get(javaClass.simpleName).map { it.toBuildingLevelAnswer() }
+        favs.get(javaClass.simpleName)
+        .map { it.toBuildingLevelAnswer() }
+        .sortedWith(
+            compareBy<BuildingLevelsAnswer> { it.levels }.thenBy { it.roofLevels }
+        )
+
```
smichel17, line 39: oop, extra newline. I'll remove that before merge

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -31,7 +31,12 @@ class AddBuildingLevelsForm : AbstractQuestFormAnswerFragment<BuildingLevelsAnsw
     private val roofLevels get() = roofLevelsInput?.text?.toString().orEmpty().trim()
 
     private val lastPickedAnswers by lazy {
-        favs.get(javaClass.simpleName).map { it.toBuildingLevelAnswer() }
+        favs.get(javaClass.simpleName)
+        .map { it.toBuildingLevelAnswer() }
+        .sortedWith(
+            compareBy<BuildingLevelsAnswer> { it.levels }.thenBy { it.roofLevels }
```
smichel17, line 37: Annoyingly, the compiler can't figure out the type of the `thenBy` without the annotation.

Originally I tried sorting by total levels, then by regular levels (descending), but I found it more confusing to parse than a simple, "What's the first number you're looking for, then what's the second".

matkoniecz:
So new answer can reorder all present ones?

smichel17:
Did a little mapping this morning. Unfortunately, the neighborhood I walked through had only 1, 1|1, and 2 story houses. So while this did feel nicer, it also didn't feel like much of a stress test. Anyone in a varied area want to take it for a spin? I can share a debug apk if needed.

smichel17:
@matkoniecz They are still added and removed in the same order, just the display order is sorted instead of chronological. So it has the disadvantage that you can't be precisely sure which one will disappear next time, when you enter a new answer. In practice, I think it doesn't matter that much.

smichel17:
Updated formatting and added all the commentary here to the commit message. So aside from maybe waiting on testing from people in a different area, I consider this ready for merge.

FloEdelmann (approved):
I didn't get around testing this in the field, but the code looks fine and I imagine it being useful.

westnordost:
I also think the same. I'll merge this before the release of the next version but I'll wait because you said that you wanted to do some field testing.

westnordost (approved):

smichel17:
I've done enough field testing now that I'm confident it works the way I'd like. But merge conflicts are extremely unlikely, so I'm happy to wait to merge.

Issue #3101, opened 2021-07-29
Compass rotation changes position of note draft


ltog:
Opening a draft for a note, then rotating the map, changes the position where the map note would be positioned.

**How to Reproduce**
1. Change the compass direction of the map by about 90 degrees away from north using a two-finger rotational gesture
2. Long press on the top half of the map on a distinctive place, e.g. a crossroad.
3. Choose `Create new note`
4. Notice the position of the `!` marker
5. Press the compass symbol (thereby getting north at the top)
6. Notice how the `!` symbol changed its position :-( I would expect instead that the marker still points to the same location on the map.

**Versions affected**

StreetComplete v33.1 on Android 10

westnordost:
This is intended / will not fix.

It is intended that you can move the map to adjust the position of the note. This is also explained in the text above the message when adding a note.
Rotating the map still falls under adjusting the map position. If you want the marker to stay at the center of the rotation, then you need to place your fingers so that the marker is at the center of rotation. It is a will not fix to change the behavior of the rotation-gesture to automatically center the rotation on the marker (because it is technically probably not possible with tangram-es and also not worth implementing it if it were, IMO)

smichel17:
I think you've misread. This is about auto-rotation back to north when tapping the compass icon.

smichel17:
Okay, I tried it. I think this is solve-able. I'll look into it within the next couple days.

ltog:
Yes @smichel17, this is about auto-rotating using the compass-icon, NOT about manually rotating using a two-finger gesture.

Thanks for looking into it!

westnordost:
Ah I see. Thanks, @smichel17 for giving it a try.

smichel17:
Okay, so this is slightly more complicated than I thought.

We do indeed have all the right information we need.

- The `MapFragment` has utility functions for working with coordinates vs screen offsets
   https://github.com/streetcomplete/StreetComplete/blob/6b86eb09260f06858dd927e56c28939dbfe0961e/app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt#L329-L369

- The `MainFragment` stores the offsets of the current open dialog
   https://github.com/streetcomplete/StreetComplete/blob/6b86eb09260f06858dd927e56c28939dbfe0961e/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L111
   And it has a callback for responding to compass taps (we'd change the last `else`, line `611`)
[...]

PR #3122, opened 2021-08-05
[Exploration] Keep rotation center when note or form is open

smichel17:
This is a quite messy fix for #3101. I don't expect it to be merged as-is; the point is to find out whether @westnordost thinks it's reasonable to clean this up and merge it at all.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/map/tangram/KtMapController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/tangram/KtMapController.kt
@@ -279,6 +279,27 @@ class KtMapController(private val c: MapController, contentResolver: ContentReso
         return position.translate(distanceAfterZoom, angle)
     }
 
+    fun getLatLonThatCentersLatLonAfterRotation(position: LatLon, padding: RectF, zoom: Float = cameraPosition.zoom): LatLon? {
+        val view = glViewHolder?.view ?: return null
+        val w = view.width
+        val h = view.height
+        if (w == 0 || h == 0) return null
+
+        val screenCenter = screenPositionToLatLon(PointF(w / 2f, h / 2f)) ?: return null
+        val offsetScreenCenter = screenPositionToLatLon(
+            PointF(
+                padding.left + (w - padding.left - padding.right) / 2,
+                padding.top + (h - padding.top - padding.bottom) / 2
+            )
+        ) ?: return null
+
+        val zoomDelta = zoom.toDouble() - cameraPosition.zoom
+        val distance = offsetScreenCenter.distanceTo(screenCenter)
+        val angle = 0.0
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/map/tangram/KtMapController.kt
```diff
@@ -279,6 +279,27 @@ class KtMapController(private val c: MapController, contentResolver: ContentReso
...
+                padding.left + (w - padding.left - padding.right) / 2,
+                padding.top + (h - padding.top - padding.bottom) / 2
+            )
+        ) ?: return null
+
+        val zoomDelta = zoom.toDouble() - cameraPosition.zoom
+        val distance = offsetScreenCenter.distanceTo(screenCenter)
+        val angle = 0.0
+        val distanceAfterZoom = distance * (2.0).pow(-zoomDelta)
+        return position.translate(-distanceAfterZoom, angle)
```
smichel17, line 300: I wasn't sure at the beginning how much of `getLatLonThatCentersLatLon` I'd have to change to get the desired behavior, so I opted to add a copy of it to modify instead of trying to add a flag, and figure out what the right interface ought to be, afterward.

In the end, all I ended up changing was the angle (offset from north) and negating `distanceAfterZoom` on the final line (which I probably wouldn't even need if I chose the correct angle).

smichel17:
In addition to the janky code, this is also janky behavior: It doesn't rotate smoothly around the point. Instead, the map does a little boomerang where it zooms away from the point, then quickly returns. It's a little jarring, but I don't think it's worth fixing, unless someone more experienced with camera manipulation knows the right way to do it; better to request "rotate around a point" from Tangram.

That said, even with the awkward transition, it's an improvement over the previous behavior, since at least you end up with the same point centered, rather than some random location.

westnordost:
Well, as said in https://github.com/streetcomplete/StreetComplete/issues/3101#issuecomment-891278049 , I don't really want to incorporate even more workarounds for missing functionality in tangram-es. If you want to explore, what would be more helpful would be to try out maplibre and see if it have all the features we need.

smichel17:
> the tangram wrapper is very tangram specific. There is no sense in keeping if if moving to MapLibre

Hm. What if I refactored this to push all the workarounds behind the wrapper? In other words, the wrapper would expose the interface we'd *like* tangrom to expose, and if they implement this functionality, we'd replace the workarounds with a direct call. This seems like it would limit the maintenance burden and especially it would not make migrating off of tangram (https://github.com/streetcomplete/StreetComplete/issues/3123) any more difficult, since the workaround code would just be thrown away in that case.

westnordost:
This was the reason why the wrapper was created in the first place. But I'd stay with my opinion.

PR #3139, opened 2021-08-08
Filter `bicycle_parking=floor` from AddBikeParkingCapacity Quest

john-h-kastner:
This type of bicycle parking is just "An area designated for the parking of bicycles", so it does not have a well defined capacity and users should not be asked about it.

I don't have a development environment setup, so I can't test at the moment. I could get around to it eventually, but, if the premise of the change is approved, I don't think the implementation is sufficiently complex to require testing on my end.


matkoniecz:
Good point, in such case mapping as an area would be the best way to mark its capacity, estimates with capacity can be dubious, frustrating and tricky.

I think that it is worth adding despite that tag usage is tiny.

I have tested it and confirmed that it works.

PR #3159, opened 2021-08-14
add building=office to civic buildings

### LINKED ISSUE #3156: non-commercial office buildings are confusing
There are also office buildings used by government and municipal companies etc.

Maybe `building=office` should be also listed in "civic buildings" group? Not only in "commercial buildings" group?

Right now it is located in group that has description

> Commercial building
> building where people work, shop or do any other commercial activities

that is not really matching for say office building of police or volunteer NGO (for police at least "work" matches).
Note that it gets even worse in translations - maybe English "commercial building" actually would include office building used by judiciary?

Actual example: https://commons.wikimedia.org/w/index.php?title=File:Budynek_biurowy_filtr%C3%B3w_wodoci%C4%85g%C3%B3w_Krakowskich_-_Krak%C3%B3w_-_Filtrowa_1.jpg (offices of water works municipal company, that would not be considered as commercial by people in Poland - at least with how "commercial" is translated in SC)

Similarly: office buildings of hospital, parliament, military, court, police, border guards, church, associations...

Maybe put this building type also in "civic buildings" group? Maybe Polish translation is faulty?

matkoniecz:
fixes #3156


westnordost:
thanks

PR #3161, opened 2021-08-14
stop asking for roof shape by default

### LINKED ISSUE #3155: Don't ask for roof shapes that likely involve guessing
Problem: Standing right next to a building, being asked for the roof shape is either hard or impossible to answer correctly in a surprising number of cases.

The cases I experienced basically boiled down to standing too close to a building (depends on building height and/or levels) and a rather flat roof shape. This concerns all roof shapes which are not gabled and which can be very flat.

Possible improvements that come to mind:
- Depending on building height/levels, don't ask for roof shapes inside a certain radius of the current position
- Don't ask for all roof shapes, but precede a question like "Does this building have a gabled roof".
- Provide an answer choice like "Cannot see clearly from my current position"/"Must be identified from further away".
- Based on user position, add a "Are you sure you could distinguish all possible roof shapes"-kind of like confirmation if the user is likely to be too close to a high building/difficult roof shape.

These Ideas are only based on the UI, I don't have any idea how feasible this is from a data point of view.

(from https://github.com/streetcomplete/StreetComplete/issues/2609#issuecomment-897872446)

Used Version: StreetComplete 33.2 on Android 10

Edit: added fourth idea for improvement

matkoniecz:
correct answer is tricky, this quest type is flooding most areas, data is of minimal importance, in cities for many buildings it cannot be answered at all by survey from ground, typically answerable from aerial imagery, no important use of this data (even less after you skip entities not crediting OSM)

fixes #3155 as now only people really interested in roof mapping will enable it

For example: potential mappers were wasting enormous amount of effort on categorizing https://www.google.pl/maps/@50.1388172,19.9129097,3a,31.4y,232h,109.1t/data=!3m6!1e1!3m4!1shE0xPjaYZANvFQO1px9E_A!2e0!7i13312!8i6656 (roof style popular in that village)

screen01

When I map with this quest enabled I have niggling thought "WTF I am doing with my time".

I am tempted to propose the same (disabling by default) with building level quest, power pole material and backrest quest.

Obviously, quest should remain as anyone interested in this kind of data can enable roof shape quest and enjoy an excellent SC interface.


app/src/main/res/values/strings.xml
```diff
@@ -497,6 +497,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="enable_quest_confirmation_title">Enable this quest type?</string>
     <string name="default_disabled_msg_go_inside">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type.</string>
     <string name="default_disabled_msg_go_inside_regional_warning">This quest type is disabled by default because it is not useful in some regions and you will likely have to enter the place to be able to answer a quest of this type.</string>
+    <string name="default_disabled_msg_tricky_3d_rendering">This quest type is disabled by default because it is time-consuming, correct answer is not always obvious and data is used only in 3D rendering.</string>
```
westnordost, line 500: How about this? I find it more important to mention that it would be more efficient in most cases to record this information from aerial imagery (=not with this app) than what the usages might be.
```suggestion
    <string name="default_disabled_msg_tricky_3d_rendering">This quest type is disabled by default because the roof shapes are often not well visible from the street. This quest type is also quite time-consuming, while it is in most cases easier and more efficient to map this from aerial imagery at home.</string>
```

westnordost (commented):
I think when the roof shape quest is disabled by default, what we could remove is the "not asking for roof shape if it is not likely visible from below" because by enabling the quest, the user knows what he is doing.

app/src/main/res/values/strings.xml
matkoniecz: It misses the part that convinced me that it is a good idea (roof shapes not really matching to any selection), but maybe it is not needed to mention it...

matkoniecz:
New reasoning misses the part that convinced me that it is a good idea (roof shapes not really matching to any selection), but maybe it is not needed to mention it - user is warned already...

Though with new reasoning generic `default_disabled_msg_tricky_3d_rendering` should be changed as reasoning is not reusable anymore for other values.

app/src/main/res/values/strings.xml
```diff
@@ -497,6 +497,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="enable_quest_confirmation_title">Enable this quest type?</string>
     <string name="default_disabled_msg_go_inside">This quest type is disabled by default because you will likely have to enter the place to be able to answer a quest of this type.</string>
     <string name="default_disabled_msg_go_inside_regional_warning">This quest type is disabled by default because it is not useful in some regions and you will likely have to enter the place to be able to answer a quest of this type.</string>
+    <string name="default_disabled_msg_roofShape">This quest type is disabled by default because the roof shapes are often not well visible from the street. This quest type is also quite time-consuming, while it is in most cases easier and more efficient to map this from aerial imagery at home.</string>
```
peternewman, line 500: I think this improves the English a little.
```suggestion
    <string name="default_disabled_msg_roofShape">This quest type is disabled by default because the roof shapes are often not easily visible from the street. This quest type is also quite time-consuming; in most cases it is easier and more efficient to map this from aerial imagery at home.</string>
```

peternewman:
> typically answerable from aerial imagery

I've not tried doing this from aerial imagery, but when I've been looking at other stuff via imagery, I'd be surprised if you could accurately differentiate a lot of these using aerial imagery, especially e.g. flat versus sloped or say the various Mansard styles versus a normal pitched roof.

> I am tempted to propose the same (disabling by default) with building level quest, power pole material and backrest quest.

At least the last one has a possible medically related benefit (although whether anyone is actually searching based on that is debatable).

westnordost (approved):

PR #3165, opened 2021-08-15
Add "office" to buildings allowed to have a housenumber

yrtimiD:
Minor change. From my experience office buildings always have an address.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
@@ -184,7 +184,7 @@ private val buildingsWithMissingAddressFilter by lazy { """
 private val buildingTypesThatShouldHaveAddresses = listOf(
     "house", "residential", "apartments", "detached", "terrace", "dormitory", "semi",
     "semidetached_house", "farm", "school", "civic", "college", "university", "public", "hospital",
-    "kindergarten", "train_station", "hotel", "retail", "shop", "commercial"
+    "kindergarten", "train_station", "hotel", "retail", "shop", "commercial", "office"
 )
 
 private fun Element.containsAnyNode(nodeIds: Set<Long>, mapData: MapDataWithGeometry): Boolean =


matkoniecz:
Good idea! Have you tested this change? It is trivial, it looks right but...

Just to check: https://github.com/streetcomplete/StreetComplete/pulls?q=office+address - seems to not be rejected earlier

yrtimiD:
> Have you tested this change? It is trivial, it looks right but...
Sincerely saying - no. Overhead of installing android studio for such a minor change is pretty big, and I'll be very surprised if something will fail in this case :) But, of course, I can test it, if you insist. 
Any chance you have any external builder which can produce an apk from specific branch?

matkoniecz:
No, in such case I will simply test it. Sadly setting up Android Studio is quite obnoxious and I am not aware about some easy to use builder

smichel17:
There is an easy-to-use builder named @smichel17 :)

https://smichel.me/files/streetcomplete/streetcomplete-debug-daac41.apk

matkoniecz (approved):
tested (there is always tiny chance of for example invisible character lurking somewhere) - and it works nicely. Thank you for the contribution!

smichel17:
(I'm very sympathetic to the setup issue. I've been putting off https://gitlab.com/fdroid/fdroidserver/-/issues/829 for almost a year now because every month or two when I find myself with both the time and motivation to work on it, getting a good python dev environment set up saps all my motivation, and then by the next time I've forgotten what I needed to do to get there. To the point where I bought some additional ram so I can just do it in a Debian vm and install everything globally so I don't have to worry about virtualenv crap. /rant)

Anyway, if you want to do more SC dev, ping me and I'm happy to walk you through whatever setup is needed (imo Android Studio installation is relatively easy, but if you prefer to work in the terminal, installing the Android SDK via CLI is not easy). One of these days, someone will get the dev tools setup experience right…

yrtimiD:
@smichel17 , thanks! 
Meanwhile I'm trying to setup github actions for the build only (despite some failed tries I found), I believe this way is a most convenient for contributors who want to do a fast fix.

westnordost (approved):

mnalis:
@yrtimiD For me, I just added [.github/workflows/build.yml](https://github.com/mnalis/StreetComplete/blob/mnalis-westnordost-profiles-helium314/.github/workflows/build.yml) to my fork, and then Github builds debug `.apk` for me (you can run it manually in the `Actions` github tab, or automatically on pull), which I just download and install manually to test changes.

yrtimiD:
@mnalis, right, eventually, this was an easies way to get debug apk, just took me some time to prepare it first time.
FYI, now [actions/workflows/build-debug-apk](https://github.com/streetcomplete/StreetComplete/actions/workflows/build-debug-apk.yml) is available on upstream too, so no need to mess with this each time.

smichel17:
I tried it, then canceled it, and it sent me an email about the cancellation. Sorry if I spammed anyone else.

PR #3166, opened 2021-08-15
Fix gps-related bugs

### LINKED ISSUE #3147: (When no GPS signal) clicking GPS button does not return to North, even when follow-me is disabled
If there is **no GPS signal**, clicking Compass button misbehaves, as SC does not seem to disengage follow-me mode in such situation (usually follow-me is enabled by clicking `GPS` icon, and disabled by scrolling the map).

**What I would expect**
- when no GPS signal (and especially if I moved the map manually), follow-me mode should be **disengaged**, so pressing compass icon first time should rotate **North to point up** (it works that way when GPS signal is strong)

**What actually happens**
- when no GPS signal, **even after moving the map manually**, follow-me mode **refuses to disengage**, and pressing compass button only switches between 2D/3D, never rotating the map to north again

**How to Reproduce**
- Start SC in area where there is no GPS signal
- move map to disengage possible [follow-me mode](https://wiki.openstreetmap.org/wiki/StreetComplete/FAQ#How_to_use_2D.2F3D_and_GPS_following.2Frotation.3F)
- click on Compass, expecting it would rotate map to "North is up"
- see it instead engages 3D follow-me mode, and that it will never rotate to "North is up" no matter how many times you press it (until you regain GPS signal strength)

Video example:

https://user-images.githubusercontent.com/156656/128779388-2f41d446-17c9-40cd-aa03-2fa06974ff82.mp4

**Versions affected**
EMUI 11 (Android 10), StreetComplete v33.2

smichel17:
Fixes #3147

This is a draft because I'm going to add a few more commits, and I may want to refactor after that, to avoid making the same fix in many places.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
@@ -601,7 +601,7 @@ class MainFragment : Fragment(R.layout.fragment_main),
         // Camera cannot rotate upside down => full circle check not needed
         val isFlat = mapFragment.cameraPosition?.tilt?.let { it <= margin } ?: false
 
-        if (mapFragment.isFollowingPosition) {
+        if (mapFragment.isFollowingPosition && mapFragment.displayedLocation != null) {
             setIsCompassMode(!mapFragment.isCompassMode)
         } else {
             if (isNorthUp) {


smichel17:
@westnordost The `MainFragment` appears to be getting location updates in two places.

https://github.com/streetcomplete/StreetComplete/blob/d6f31b5a11394e1679932b6ca4a4886ac5eb8cb2/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L278-L282

https://github.com/streetcomplete/StreetComplete/blob/d6f31b5a11394e1679932b6ca4a4886ac5eb8cb2/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L568-L573

Are both of these needed? If not, which would you prefer to be removed? If it's the same to you, I am slightly leaning towards receiving location updates in the `MainFragment` and passing them down into the `MapFragment`. This way, the same flow can be adapted to handle location availability updates— the `MapFragment` currently does not receive them, and this is why the location pointer still displays after gps is turned off: in `MainFragment::updateLocationPointerPin`, it only checks the MapFragment's location, which is never reset to null when location services are disabled.

https://github.com/streetcomplete/StreetComplete/blob/d6f31b5a11394e1679932b6ca4a4886ac5eb8cb2/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L703-L707

westnordost:
Both are needed, because both do different things. Maybe the naming is off:

`onLocationDidChange` is called after the map fragment updated its **displayed** location (scroll around the map)

`onLocationChanged` is called when the user's location changed (GPS location)

westnordost:
Anyway, I'll merge this as it is, the change you did so far fixes the mentioned bug.

smichel17:
Well, ok. I was going to get back to this, but I can continue in another PR :)

> Both are needed, because both do different things.

[written several days ago] Okay, I see: the `LocationAwareMapFragment` only receives location updates (and thus, calls back to the `MainFragment`) when it is tracking the user's location, and the `MainFragment` only receives a single update, when location is first enabled.

> `onLocationDidChange` is called after the map fragment updated its displayed location (scroll around the map)

<details><summary>This is not exactly true, as far as I can tell. In summary:

- The `MainFragment` triggers the `MapFragment` to begin location tracking any time location services are enabled.
- The `MapFragment` does not stop location tracking except when location services are turned off or when the fragment is closed (in `onStop()`).
- The location update is always passed back up to the `MainFragment`, regardless of whether the map location changed.</summary>

> `LocationAwareMapFragment` subscribes to location changes here (and nowhere else):
> 
> https://github.com/streetcomplete/StreetComplete/blob/0ca3031ef7c712450514df394210b3f207224d6f/app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt#L106-L109
> 
> Which in turn is called only here:
> 
> https://github.com/streetcomplete/StreetComplete/blob/0ca3031ef7c712450514df394210b3f207224d6f/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L542-L549
> 
> And handles them like this:
> 
> https://github.com/streetcomplete/StreetComplete/blob/0ca3031ef7c712450514df394210b3f207224d6f/app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt#L136-L142

</details>

…I'd call this, "The MainFragment subscribing to location updates, with extra steps." :smile:

<details><summary>It makes me feel like the abstraction boundary is wrong, somewhere.</summary>

We have some public functions on `MapFragment` that exist only to tell it to start tracking location updates.

https://github.com/streetcomplete/StreetComplete/blob/0ca3031ef7c712450514df394210b3f207224d6f/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L542-L557

I understand the relationship between the MainFragment and GPS button: the button is a dumb view\*, which the fragment updates sometimes. I don't really understand the relationship with the map fragment. It's a 3-class hierarchy, seemingly just to split the logic up in to manageable/logical chunks. It definitely is the source of truth for map state. But the responsibilities for managing state updates and the "business logic" are split up between the different classes and the MainFragment in confusing ways.

Actually, the abstraction boundaries are not *that* confusing. It seems like MainFragment is intended to have the high level logic, for coordinating the various components

\*Actually, there's one place where the fragment uses the state of the view as part of control flow… perhaps we should change this.

https://github.com/streetcomplete/StreetComplete/blob/0ca3031ef7c712450514df394210b3f207224d6f/app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt#L625-L628
</details>

I was originally going to say that I wasn't sure what bothers me about the abstraction boundary, but after writing out the part inside the drop-down above, I have a clearer idea. Still, I think I will proceed with my original plan: finish the bugfixes "locally" however I need to, then experiment with refactoring afterward.

---

Finally, completely unrelated: I've noticed in several places we do something like
`displayedLocation.let { LatLon(it.latitude, it.longitude) }`.

Would you accept a PR adding an extension function, `Location.toLatLon()`?
If so, which file would be the right file to add that? `LocationUtil.java` (with converting to kotlin), a new file, or somewhere else?

westnordost:
> Would you accept a PR adding an extension function, Location.toLatLon()?
> If so, which file would be the right file to add that? LocationUtil.java (with converting to kotlin), a new file, or somewhere else?

Yes, yes

matkoniecz:
@smichel17 Thanks for this fix! I got bitten by it and failed to track down or notice what is causing problems.

PR #1113, opened 2018-06-29
Add .travis.yml

ENT8R:
Fixes #1003 
The Travis configuration runs two scripts after each other:
1. `./gradlew testDebugUnitTest` (works without any problems right now)
2. `./gradlew connectedDebugAndroidTest` (currently fails at a test related to the housenumber quest: https://travis-ci.org/ENT8R/StreetComplete/jobs/398136718#L3469)

There are two separate tests, one for API 22 and one for API 24.
In order to enable Travis for this repository you need to go to https://travis-ci.org/ and enable it for StreetComplete. If you want to, I can also add a badge to the readme which always shows the latest build status.
You won't receive any emails if a build fails but this can be turned on again by setting `email` to `true`
The builds and the emulator start time take always some time, so a normal build time for this repository is at around 30 to 45 minutes...

PR #3167, opened 2021-08-15
Add build debug apk github action

yrtimiD:
This PR adds a zero-effort way to get a fresh apk to test your changes. 
This might be useful for testing small fixes without overhead of setting up dev environment and installing Android Studio.
Triggered manually only.
Result is an app debug apk with different app name and different icon which might be safely pushed to android device with `adb install` command.
Retention for the archived result file is a 1 day (no real reason to keep temp debug package, and full rebuild takes only ~8 min)

### DIFF
--- a/.github/workflows/build-debug-apk.yml
+++ b/.github/workflows/build-debug-apk.yml
@@ -1,0 +1,28 @@
+name: Build debug apk
+
+on:
+  [workflow_dispatch]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - name: Build with Gradle
+      run: ./gradlew assembleDebug
+    - name: Archive APK
+      uses: actions/upload-artifact@v2
+      with: 
+        name: debug-apk
+        path: app/build/outputs/apk/debug/app-debug.apk
+        retention-days: 1


matkoniecz:
This would be great! Thank you for that PR!

Looking at https://github.com/pricing - it is free at least right now, and it would be "nice to have" not some fundamental workflow. So there is no real risk of lock in, I think at least.

Sorry to ask that: have you tested it?

There was a previous attempt with Travis that was not working well - #1113 #1502

Maybe Github actions are better or this is better written?

yrtimiD:
@matkoniecz , AFAIK, it was given free by Microsoft for public repos. 
I did tested this build several times on my fork and successfully got a fully working debug apk with yellow icon just after 8 min.
I found more attempts to have a full builds with tests and linting, I tried this too but failed because of soma failed tests and some strange issue with linting. 
Running such a minimal build It's not perfect, but at least it allows fast feature validation.

westnordost:
Oh nice, thanks!

westnordost:
I don't even get to 8 minutes on my machine.

matkoniecz:
Should I see this action or its effects or its status say in #3191 ?

yrtimiD:
> Should I see this action or its effects or its status say in #3191 ?

@matkoniecz, don't think so. The action is set to be executed manually only.

riQQ:
It can be started manually from https://github.com/streetcomplete/StreetComplete/actions/workflows/build-debug-apk.yml by all collaborators of this repository. But right now, it is only able to build the latest commit of the master branch.

yrtimiD:
> But right now, it is only able to build the latest commit of the master branch.

@riQQ, you should be able to select a branch (assuming branch has this action)
image
(screenshot is from my fork)

riQQ:
You're right, my bad. I didn't remember this, maybe it wasn't available last time I looked into this feature.

PR #3170, opened 2021-08-16
hint that all users should enable team mode

matkoniecz:
some actual people were confused about that
closes #2741 that proposed to add more info about this

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -870,7 +870,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="team_mode_description">When mapping the same area together in a group, you can enter team mode to split up quests between all team members.</string>
     <string name="team_mode_team_size_label">How many people are there?</string>
     <string name="team_mode_team_size_hint">Team mode works for groups of 2 to 12 people.</string>
-    <string name="team_mode_choose_color">Now everybody in your group chooses a unique color:</string>
+    <string name="team_mode_choose_color2">Now everybody in your group chooses a unique color on their phone:</string>
     <string name="team_mode_active">Team mode is active</string>
     <string name="team_mode_deactivated">Team mode deactivated</string>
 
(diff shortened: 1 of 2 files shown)


westnordost:
@FloEdelmann mentioned the original wording before I changed it https://github.com/streetcomplete/StreetComplete/pull/2741#issuecomment-899431919. I don't remember why I changed it and I did not see any post explaining the reason for that, so I guess it wasn't about this PR re-adds but maybe just the wording.

So from my view, this is good to merge.

westnordost (approved):

FloEdelmann (approved):

app/src/main/res/values/strings.xml
```diff
@@ -870,7 +870,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="team_mode_description">When mapping the same area together in a group, you can enter team mode to split up quests between all team members.</string>
     <string name="team_mode_team_size_label">How many people are there?</string>
     <string name="team_mode_team_size_hint">Team mode works for groups of 2 to 12 people.</string>
-    <string name="team_mode_choose_color">Now everybody in your group chooses a unique color:</string>
+    <string name="team_mode_choose_color2">Now everybody in your group chooses a unique color on their phone:</string>
```
smichel17, line 873:
```suggestion
    <string name="team_mode_choose_color2">Now, everybody must pick a different color. Choose yours:</string>
```

smichel17 (changes requested):
Aah, I missed this. But, I would do it like this: the suggested change, and then also change the `team_mode_team_size_label` string to "How many people are there? Each person in the group must enter this number on their phone." That way the concept that everyone is doing it on their own phone is introduced earlier, and stays on screen the whole time. And then it's also clear that the second instruction is just for you.

westnordost:
I can't reopen a PR, could you maybe create a new PR for that? Then we can also see the big picture.

smichel17:
Done, #3172

PR #3175, opened 2021-08-17
make stone stepover stiles less confusing

### LINKED ISSUE #2991: Stile Type Quest improvements
When I was adding the stile type with StreetComplete, I found myself adding a note for stone stepover stiles, since they obviously don't look like wooden ones, as the whole setup is rather different.  [On the wiki](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dstile#Examples), it says that they are, in fact, stepover stiles, and should be tagged as such (with `material=stone`).  Two pictures might help, showing both a wooden and stone stepover stile, or another option could be added which also tags `material=stone`.

There appear to be a handful of examples of stiles tagged `stile=stone_steps` or `stile=stone steps`, which I imagine are the same as the stiles I found that, according to the wiki, are meant to be tagged `stile=stepover`.

As well as this (though I might open another issue if it is deemed appropriate), a new quest could be created asking for the material of stiles.  Currently, [according to Taginfo](https://taginfo.openstreetmap.org/tags/?key=barrier&value=stile#combinations), 16.9% of `barrier=stile` objects are tagged with `material=*`.  58.01% of [objects tagged with `stile=*`](https://taginfo.openstreetmap.org/keys/stile#combinations) are tagged with `material=*`.

An example note that I made is [note 2723843](https://osm.org/note/2723843).

matkoniecz:
fixes #2991

@10992-osm I have good news for you!

screen
screen01


matkoniecz:
I can be easily convinced to use https://www.geograph.org.uk/photo/398328 or https://commons.wikimedia.org/wiki/File:Dry_Stone_Wall_and_Stile,_Yorkshire_Dales_(3875164717).jpg instead - to show that small piece of wood is not changing that it is a stone stile.

westnordost (approved):
Does it make sense to treat `stile=stepover` without `material` tagged as if it is not tagged yet?, i.e. ask again?

westnordost:
🤷

PR #3182, opened 2021-08-18
Auto-mirror arrow for RTL languages

### LINKED ISSUE #3181: When the app is in an RTL language, arrows for list items are not flipped
First off, before I start, I just want to say I am grateful for this app. Aside from very minor issues like this one, everything has been working smoothly. The app is beautiful and generally easy to use, even for OSM newbies like me. Thank you for making this.

**How to Reproduce**
1. Change to an RTL language (e.g. Arabic)
2. Open the settings page
3. The arrow for chosen quests list tile is rotated the wrong way

Here's a screenshot:

screenshot

**Versions affected**

I am using StreetComplete 33.2 on Android 11.

mhmdanas:
Fixes #3181. ~~Draft because I have not tested this yet (it's pretty late here).~~ Tested and confirmed as not working. I will need to try some more to fix this.

Note that the "right" in the drawable name is now misleading, as it is a left arrow on RTL languages now. What to do?

### DIFF
--- a/app/src/main/res/drawable/ic_chevron_right_24dp.xml
+++ b/app/src/main/res/drawable/ic_chevron_right_24dp.xml
@@ -1,4 +1,5 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:autoMirrored="true"
         android:width="24dp"
         android:height="24dp"
         android:viewportWidth="24.0"


westnordost:
Thank you for trying to fix #3181, but now of course, it is the wrong direction for all non RTL languages. Don't worry, I will fix this myself.

mhmdanas:
Thanks in advance, but why would it be in the wrong direction (well, if it actually worked properly for RTL languages)? `autoMirrored` mirrors the icon only when in RTL mode.

westnordost:
`autoMirrored` only works with Android 4.4 onwards, but the app also supports Android 4.2 and Android 4.3.

When lower Android versions support is dropped, this could be used.

PR #3191, opened 2021-08-19
promote information=* and segregated=* quests

matkoniecz:
this quests seem both interesting, and actually used data - worth putting above lit=* quest spam

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
@@ -190,6 +190,8 @@ import javax.inject.Singleton
         AddFerryAccessPedestrian(),
         AddFerryAccessMotorVehicle(),
         AddAcceptsCash(featureDictionaryFuture),
+        AddInformationToTourism(), // OSM Carto
+        AddCyclewaySegregation(), //Cyclosm, Valhalla, Bike Citizens Bicycle Navigation...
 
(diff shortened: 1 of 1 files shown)


westnordost:
No, the quests will be reordered in the next major release anyway, all of them.

matkoniecz:
@westnordost Note that I made some minor changes where I was fully certain about them and commited it directly - including one ordering change in https://github.com/streetcomplete/StreetComplete/commit/ff4c07ee689fe1573226e19a9f708c84cfa11fa8

Should I revert this?

westnordost:
Well it will be overwritten. Best if you revert this, so that I don't have merge conflicts. If you want to hcange the order, you should do it on that branch

matkoniecz:
Is https://github.com/streetcomplete/StreetComplete/tree/quest-profiles the correct target?

I see https://github.com/streetcomplete/StreetComplete/tree/quest-order that initially seemed to match better, but from looking at history seems not the proper one (maybe even this pointer qualifies for deletion as no longer needed).

westnordost:
Ehm, the newer one, don't know right now

FloEdelmann:
The [`quest-profiles`](https://github.com/streetcomplete/StreetComplete/tree/quest-profiles) branch contains all the commits from the [`quest-order`](https://github.com/streetcomplete/StreetComplete/tree/quest-order) branch, so that one can likely be deleted.

PR #3192, opened 2021-08-19
Fix grammar in link description

1ec5:
I saw a missing preposition in [this nice writeup about StreetComplete](https://opensource.com/article/21/8/streetcomplete-quests) and my inner grammarian couldn’t resist. 😁


westnordost:
ok, thanks!

PR #3194, opened 2021-08-20
Change no-sidewalk tagging to =no

ZeLonewolf:
This PR changes the StreetComplete tagging for streets with no sidewalks to `sidewalk=no` instead of `sidewalk=none`.  This is consistent with the tagging used by both iD and JOSM.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
@@ -110,7 +110,7 @@ class AddSidewalk : OsmElementQuestType<SidewalkAnswer> {
                 answer.left && answer.right -> "both"
                 answer.left -> "left"
                 answer.right -> "right"
-                else -> "none"
+                else -> "no"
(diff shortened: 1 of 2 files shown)


riiga:
Consistency is key. I support this change.

FloEdelmann:
Apparently, StreetComplete is one of only two projects that uses `none`: https://taginfo.openstreetmap.org/tags/sidewalk=none#projects

`no` is slightly more often used: https://taginfo.openstreetmap.org/keys/sidewalk#values

Vespucci also uses `no`: https://taginfo.openstreetmap.org/tags/sidewalk=no#projects

ZeLonewolf:
Note that =none has experienced a recent rise of 100K tags since January, which presumably is due to the popularity of StreetComplete.

westnordost:
Haha, StreetComplete vs everyone else, who wins? ;-)

image

westnordost (approved):

westnordost:
`none` will have to be supported by any data consumer pretty sure for forever, so it doesn't really matter which value is used. But it's good if all editors tag the same thing the same, so approved.

Issue #2472, opened 2021-01-09
AddProhibitedForPedestrians quest leads to bad OSM data


riiga:
The AddProhibitedForPedestrians quests asks users to answer which roads are prohibited or not for pedestrians. I've seen other people in my area answering the quest which adds `foot=no` to roads which may be unsuitable to walk on but which you legally are still allowed to walk on. This causes bad data in OSM since I've already added all such roads in my area according to what is legally allowed, but now more roads are getting those tags while they really shouldn't. I feel this is a mostly a user error, but I think it also lies in StreetComplete and there needs to be a clarification that there should be a sign prohibiting pedestrians explicitly, otherwise no access tag should be added.

The OSM wiki clearly says `foot=no` is the *legal* accesibility.

> Access values describe **legal** permissions/restrictions and should follow ground truth e.g. signage or legal ruling and not introduce guesswork. It does not describe common or typical use, even if signage is generally ignored. 
-[Key:access](https://wiki.openstreetmap.org/wiki/Key:access)

Here are a few examples of the data being added:
https://www.openstreetmap.org/changeset/92125582 (Industrigatan: [Location](https://www.google.se/maps/@58.4118217,15.586051,3a,75y,2.02h,86.81t/data=!3m6!1e1!3m4!1s6BxNhHohs9EGjCY8CZXDFw!2e0!7i13312!8i6656), [nearest intersection showing no sign](https://www.google.se/maps/@58.4185072,15.5911612,3a,25y,230.35h,88.43t/data=!3m6!1e1!3m4!1sH4jDcRwUAE8z0PQvQETQ6A!2e0!7i13312!8i6656))
https://www.openstreetmap.org/changeset/96803039 (Haningeleden: [Location](https://www.google.se/maps/@58.389732,15.6244885,3a,33.8y,119.04h,86.26t/data=!3m6!1e1!3m4!1sKfuxs3Y6XHjUAIMnqvc9Ow!2e0!7i13312!8i6656), [nearest intersection showing no sign](https://www.google.se/maps/@58.3873158,15.6287043,3a,66.5y,344.01h,83.4t/data=!3m6!1e1!3m4!1svEroB2SdnQB_luIFEVjJng!2e0!7i13312!8i6656))
[...]

PR #3195, opened 2021-08-20
Disable AddProhibitedForPedestrians quest in Sweden

riiga:
After using StreetComplete for a while now and having already changed the Swedish translation to specify the AddProhibitedForPedestrians only applies to legal prohibitions, this quest still causes issues.

1. There's still `foot=no` being added in error, mostly by people not using the Swedish localisation (Having your phone or computer set to English is common among more tech-savvy people). It's not a worthwhile enterprise to try and contact those users nor is it possible to force them to use a localised version of the app.
2. Roads (in Sweden) where pedestrians are prohibited and which are not already covered by other regulations such as motorways are very uncommon.
3. The biggest issue is not actually adding `foot=no` in error, it's adding `foot=yes` if pedestrians are allowed. By default `foot=yes` is implied for all roads, streets, paths, etc. in Sweden, except motorways, expressways (`motorroad=yes`) and explicit cycleways. Unlike countries like the UK, there is no legal "right of way" that can be restricted and for which adding `foot=yes` is useful.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestrians.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestrians.kt
@@ -34,6 +34,8 @@ class AddProhibitedForPedestrians : OsmFilterQuestType<ProhibitedForPedestriansA
     override val icon = R.drawable.ic_quest_no_pedestrians
     override val isSplitWayEnabled = true
 
+    override val enabledInCountries = AllCountriesExcept("SE")
+
     override fun getTitle(tags: Map<String, String>) = R.string.quest_accessible_for_pedestrians_title_prohibited
 
     override fun createForm() = AddProhibitedForPedestriansForm()


app/src/main/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestrians.kt
```diff
@@ -34,6 +34,8 @@ class AddProhibitedForPedestrians : OsmFilterQuestType<ProhibitedForPedestriansA
     override val icon = R.drawable.ic_quest_no_pedestrians
     override val isSplitWayEnabled = true
 
+    override val enabledInCountries = AllCountriesExcept("SE")
```
Helium314, line 37: you'll need to import `de.westnordost.streetcomplete.data.quest.AllCountriesExcept` for this to work

westnordost:
Translation was changed from
_"Är fotgängare förbjudna att gå på den här vägen?"_
to
_"Finns det ett skyltat förbud att gå på den här vägen?"_
on the 2nd of May 2021.

The issue this PR refers to is #2472.

I closed the issue because there was no reply / suggestion after my [last comment](https://github.com/streetcomplete/StreetComplete/issues/2472#issuecomment-841744824).

At this point, I still think there can't be a Swedish solution for that.

Either we can make this quest not being misunderstood better, or maybe the whole quest should be removed if this is not possible. For the sake of keeping it in order, I will reopen #2472 and ask everyone to contribute to that ticket and close this PR (because a Sweden-only solution will not be implemented, as this is an issue that is independent of the country).

westnordost:
Oh, I never closed #2472. So anyway, please respond there.

PR #3199, opened 2021-08-21
#2643 language switch

### LINKED ISSUE #2643: Add option to change app language independent of phone language
Some people have their phones set to a specif language for learning or other reasons. Because accuracy is an important requirement for StreetComplete, this may cause a problem with users that are not as comfortable with the selected language, having them change the entire phone's language just to use StreetComplete.

It would be interesting to add an option to dissociate SteetComplete's language to that used in the device. This would allow users to have the app in a more familiar and/or convenient language, without needing to change the whole phone environment to said language. This would be an setting in the normal setting panels, allowing you to choose the interface language from available options.

amenk:
This adds the possibility to switch the app language to any of those where a translation exists (fixes #2643).

Known issues also see #2643 

We might improve the way the available language list is build - currently it's hard coded (maybe a YAML file?) Or can we auto detect this? --> it would improve the workflow when adding new languages

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
@@ -46,6 +46,7 @@ import de.westnordost.streetcomplete.map.MainFragment
 import de.westnordost.streetcomplete.notifications.NotificationsContainerFragment
 import de.westnordost.streetcomplete.tutorial.TutorialFragment
 import de.westnordost.streetcomplete.util.CrashReportExceptionHandler
+import de.westnordost.streetcomplete.util.LocaleContextWrapper
 import de.westnordost.streetcomplete.util.parseGeoUri
 import de.westnordost.streetcomplete.view.dialogs.RequestLoginDialog
 import kotlinx.coroutines.launch
@@ -113,6 +114,10 @@ class MainActivity : AppCompatActivity(),
             }
         }
         handleGeoUri()
+    }
+
+    override fun attachBaseContext(newBase: Context?) {
+        super.attachBaseContext(LocaleContextWrapper.wrap(newBase,"am"))
     }
 
     private fun handleGeoUri() {
--- a/app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
@@ -117,7 +117,7 @@ class MainActivity : AppCompatActivity(),
     }
 
     override fun attachBaseContext(newBase: Context?) {
-        super.attachBaseContext(LocaleContextWrapper.wrap(newBase,"am"))
+        super.attachBaseContext(LocaleContextWrapper.wrap(newBase))
     }
 
     private fun handleGeoUri() {
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
@@ -2,6 +2,7 @@ package de.westnordost.streetcomplete.settings
 
 import android.content.Intent
 import android.content.SharedPreferences
+import android.icu.util.ULocale.getDisplayLanguage
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.widget.TextView
@@ -10,6 +11,7 @@ import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.os.bundleOf
 import androidx.lifecycle.lifecycleScope
+import androidx.preference.ListPreference
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.PreferenceManager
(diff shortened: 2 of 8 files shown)


app/src/main/java/de/westnordost/streetcomplete/util/LocaleContextWrapper.java
```diff
@@ -0,0 +1,75 @@
...
+		}
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+			context = context.createConfigurationContext(config);
+		} else {
+			context.getResources().updateConfiguration(config, context.getResources().getDisplayMetrics());
+		}
+		return new LocaleContextWrapper(context);
+	}
+
+	@SuppressWarnings("deprecation")
```
matkoniecz, line 56: Just to confirm: how serious this and other deprecations should be treated? Is there some existing replacement and deprecated things are used to support older phones?

Or is it deprecated without replacement and it is likely that in future the feature will no longer work?

Or is it deprecation that can be ignored?

amenk: I have no clue :-) I am more a beginner Android Developer and it's source is stackoverflow... The good thing is that there is alternative code for newer version in there. Not sure if we could delete one version of the code, based on the build target.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
```diff
@@ -115,6 +116,10 @@ class MainActivity : AppCompatActivity(),
         handleGeoUri()
     }
 
+    override fun attachBaseContext(newBase: Context?) {
+        super.attachBaseContext(LocaleContextWrapper.wrap(newBase))
+    }
+
```
westnordost, line 122: I'd really like to avoid something like this. This is a big potential for bugs because one got to remember to add this to every single Fragment and Activity(?)

There are a lot more Fragments and Activities in the app than for which you added those lines. Also, there may be Fragments from third parties (libraries) etc.

Why is this necessary? The code in `LocaleContextWrapper.wrap` contains code that replaces the default locale. This is a static variable, meaning it should stay this way until the app is restarted, right? So, why set it again and again on attaching the base context of each `Activity` and `Fragment`?

I was hoping it would be possible to for example set the default locale within the app to something else **on app startup before any activity is started** (in the `Application` class). And whenever one changes the language setting, the app would restart. Is this not possible?

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
```diff
@@ -137,6 +205,10 @@ class SettingsFragment : PreferenceFragmentCompat(), HasTitle,
                 AppCompatDelegate.setDefaultNightMode(theme.appCompatNightMode)
                 activity?.recreate()
             }
+            Prefs.LANGUAGE_SELECT -> {
+                val intent = Intent(context?.applicationContext, MainActivity::class.java)
+                this.startActivity(intent)
```
westnordost, line 210: I think what you want to do is
```suggestion
                activity?.recreate()
```

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
```diff
@@ -100,6 +100,74 @@ class SettingsFragment : PreferenceFragmentCompat(), HasTitle,
         }
     }
 
+    private fun buildLanguageSelector() {
+        // FIXME: externalize to YML?
```
westnordost, line 104: Don't worry about this, I will create a gradle task after that creates/updates a list if this PR can be merged

app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
amenk: okay, I believe I did not try it in the application class. I could work.
I believe I tried in the main activity and I believe the locale was reset on rotation or something like this.

Fragments from libraries, were this cannot be added, are a good point.

app/src/main/java/de/westnordost/streetcomplete/settings/SettingsFragment.kt
amenk: That seemed to switch the language only for the settings activity but not for the rest of the app.
Anyways, we don't need to switch it on the fly and it would be okay prompt the user for a full restart of the app, so this could be avoided.

westnordost: Restarting the app would be fine, maybe it could be done automatically

smichel17: As a user, I wouldn't want the app to restart as soon as I picked a language, because that would boot me out of the settings. I'd be happy if the app restarted as soon as I closed the settings screen.

app/src/main/java/de/westnordost/streetcomplete/util/LocaleContextWrapper.java
smichel17: Looks like the legacy methods are only called on older api versions, where they will never be removed, so it should be safe.

westnordost:
Should I mark this as "help requested"?

amenk:
@westnordost yes, that would be nice

westnordost:
FYI, because I had to look deep into the topic for some other issue:

### How to set another locale?

1. The locale is a property of the configuration of the context's resources.
2. It is sufficient to replace the locale for every activity because fragments inherit their locale from the activity they are attached to
3. It is not sufficient to replace the locale of the global application context. It seems that (at least) the locale is overwritten on activity-level contexts
4. There are several methods to replace the locale. There is a lot of hackery, ignorance, outdated information on stackoverflow, aggrevated by the fact that there seemed to have been a bug in the appcompat library not that long ago which required certain workarounds but this bug is solved now. Below, the methods to replace the locale to French (as an example) that work

### Methods that work

In an activity, when it is initialized

#### 1. applyOverrideConfiguration
```kotlin
init {
    val config = Configuration().apply { locale = Locale.FRENCH }
    applyOverrideConfiguration(config)
}
```

- added in API level 17 
- the method may only be called **once** on an activity

#### 2. createConfigurationContext

```kotlin
override fun attachBaseContext(base: Context) {
    val config = Configuration().apply { locale = Locale.FRENCH }
    val configContext = base.createConfigurationContext(config)
    super.attachBaseContext(configContext)
}
```

- added in API level 17 
- `createConfigurationContext` wraps an original context into a context that has the given different configuration

#### 3. updateConfiguration

```kotlin
override fun attachBaseContext(base: Context) {
    super.attachBaseContext(base)
    resources.configuration.locale = Locale.FRENCH
    resources.updateConfiguration(resources.configuration, resources.displayMetrics)
}
```
- deprecated since API 25, though in API 30 still used internally in AppCompatActivity
- does not wrap the resources/context, but modifies the current configuration and re-sets it
- this is the only method that works to modify the current resources to have a different locale without recreating the activity. However, it makes sense to re-start it, or at least recreate the view, because the view has already been inflated with the old locale. Hence, changing the locale while the activity is already displayed changes nothing.

---

It seems to me that Google intended `applyOverrideConfiguration` to be used for this, but this doesn't play well together with inheritance: If a superclass overrides someting in the configuration, any subclass cannot do this anymore, otherwise an exception will be thrown.
The `createConfigurationContext` method is more flexible, especially if one aims at putting the localization-code into a common superclass of all activities in the application.

westnordost:
I implemented this now after all. I redid it completely, see #3512

PR #3201, opened 2021-08-22
put charging station quests in one package

matkoniecz:
I want to do similar merges in other cases - folder listing of quests is getting quite long


westnordost:
I don't know, what exactly is the advantage of this? The reason why there are these packages is so that the quest UI and quest logic are in the same place. Whether they are related through the properties they edit is not really relevant from a code perspective - i.e. they have no code interdependencies, have they?

matkoniecz:
Primarily? Finding correct quest in this window would be easier.

screen

Though maybe I should navigate less by mouse and more by keyboard and/or write code outside Android Studio?

westnordost:
Have you used Ctrl+N before? That's how you find individual quests easily.

matkoniecz:
I guess that I should try that first

westnordost:
And Ctrl+Shift+N for "any file with that name"

PR #3206, opened 2021-08-22
Survey invalid cycleway

### LINKED ISSUE #3148: Immediately resurvey any incomplete cycleway tagging
**Use case**
Off the back of #3129 it's clear some common cycleway tagging is incomplete for example `cycleway=lane` or `cycleway:both=lane`  without a valid `cycleway:both:lane` value, we could fix this immediately, we don't need to wait for a resurvey time period to expire.

**Proposed Solution**
Expand the existing element search logic to use the isAvailableAsSelection function to filter partial tagging and show it for resurvey immediately too.

westnordost:
Fixes #3148

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
@@ -81,7 +81,7 @@ class AddCycleway(private val countryInfos: CountryInfos) : OsmElementQuestType<
             val minAngleToWays = 25.0
 
             if (maybeSeparatelyMappedCyclewayGeometries.isNotEmpty()) {
-                // filter out roads with missing sidewalks that are near footways
+                // filter out roads with missing cycleways that are near footways
                 roadsWithMissingCycleway.removeAll { road ->
                     val minDistToWays = estimatedWidth(road.tags) / 2.0 + 6
                     val roadGeometry = mapData.getWayGeometry(road.id) as? ElementPolylinesGeometry
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
@@ -16,7 +16,7 @@ import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
 import de.westnordost.streetcomplete.quests.cycleway.Cycleway.*
 import de.westnordost.streetcomplete.util.isNearAndAligned
 
-class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
+class AddCycleway(private val countryInfos: CountryInfos) : OsmElementQuestType<CyclewayAnswer> {
 
     override val commitMessage = "Add whether there are cycleways"
     override val wikiLink = "Key:cycleway"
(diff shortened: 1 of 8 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -80,19 +80,28 @@ class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
 
             val minAngleToWays = 25.0
 
-            // filter out roads with missing sidewalks that are near footways
-            roadsWithMissingCycleway.removeAll { road ->
-                val minDistToWays = estimatedWidth(road.tags) / 2.0 + 6
-                val roadGeometry = mapData.getWayGeometry(road.id) as? ElementPolylinesGeometry
-                roadGeometry?.isNearAndAligned(minDistToWays, minAngleToWays, maybeSeparatelyMappedCyclewayGeometries) ?: true
+            if (maybeSeparatelyMappedCyclewayGeometries.isNotEmpty()) {
+                // filter out roads with missing sidewalks that are near footways
```
peternewman, line 84: Shouldn't it be this for this quest?
```suggestion
                // filter out roads with missing cycleways that are near footways
```

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -115,7 +124,7 @@ class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
 
         /* but if already tagged an old, we don't need to look at surrounding geometry to see if
            it applicable or not */
```
peternewman, line 126: Minor grammar nit:
```suggestion
        /* but if already tagged and old, we don't need to look at surrounding geometry to see if
           it is applicable or not */
```

peternewman (commented):
LGTM, just a couple of very minor comments

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
westnordost: yeah, i think you are right

westnordost:
Thank you, @peternewman !

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -293,7 +302,7 @@ class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
                 or highway = residential and (
                   maxspeed > 30
```
peternewman, line 303: It might be good to have a comment that these are in kph, I was a bit confused what was going on at first.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -308,7 +317,7 @@ class AddCycleway : OsmElementQuestType<CyclewayAnswer> {
                 !maxspeed
                 or maxspeed > 20
                 or (maxspeed ~ ".*mph" and maxspeed !~ "([1-9]|1[0-2]) mph")
```
peternewman, line 319: In the UK on a "proper" road it's rare to see < 20mph (which I guess is handled in the case above). But 15, and more often 10 or occasionally 5mph are common on e.g. large private sites/roads. I don't know if it's worth relaxing it to include the likely mph equivalent, rather than a direct conversion of kph to mph?

peternewman (commented):
A couple more minor comments.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
westnordost: nah, that's the default in OSM. The filter syntax just looks up what are inthe OSM tags, and there is something like "30" or "50" etc. in the OSM tags. (Actually, `>` comparisons don't work with "30 mph" because that is not a number but a string). Hence the elaborate regex syntax you also commented on.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
westnordost: That's the "definitely never ask for a cycleway" filter. In roads where there is a speed limit of 20 or less (whatever the reason for this), there is absolutely no sense for a cycleway because that is the speed cyclists go.

westnordost: And to shift this limit solely on the base of the unit used for speed limit doesn't make sense to me. Cyclists don't go faster in the UK than in the rest of the world.

westnordost:
Thanks for the review, @peternewman !

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
peternewman: Actually they do :laughing: (I was just looking for cycling speeds):
https://www.cyclingweekly.com/fitness/training/13-ways-increase-average-cycling-speed-144937

UK Men average 15mph, which means the filter should exclude all roads <= 15mph. FWIW I did find some 12mph tagging (round two UK parks), but 15mph is used 200 times more.

Anyway, I was more meaning that directly converting from kmh to mph is likely to result in unusual/infrequently used speed limits, potentially cutting off the realistic/actual ones, see this table for some more examples (going the other way):
https://www.sharetheroadtozero.com/wp-content/uploads/2016/03/Highway-Code-Speed-Limit-Infographic.png

westnordost: British ride faster because the constant rain cools them down ;-)

As said, this is not about speed limit signs but about some (TBH randomly) chosen speed at which cyclists go.  Look at the females, 20 km/h seems quite reasonable as a rough average. Also, how were the statistics created? Is this only for cycling enthusiasts (on racing bikes) or the average on any bike?

The case you seem to make here is that roads signed with 15 mph will never have a cycleway because cyclists are never expected to go slower than that. Is this the case you want to make?

peternewman: > British ride faster because the constant rain cools them down ;-)

I assume the Dutch win as there are no hills.

> As said, this is not about speed limit signs but about some (TBH randomly) chosen speed at which cyclists go. Look at the females, 20 km/h seems quite reasonable as a rough average. Also, how were the statistics created? Is this only for cycling enthusiasts (on racing bikes) or the average on any bike?

Strava is a tracking app, so it will probably have more enthusiasts, but possibly on mountain bikes and stuff too. People probably aren't using it to track their daily commute.

> The case you seem to make here is that roads signed with 15 mph will never have a cycleway because cyclists are never expected to go slower than that. Is this the case you want to make?

I think more broadly that 15 mph is so slow it's the safety factor, you don't then need to separate them. As this article alludes to, <20 mph is very rare in the UK on a normal road, if you give them a cycle lane too, you're only really protected pedestrians just randomly crossing. I'd imagine the reality here is there isn't the space to add a separate cycleway, so the limit is the alternative option:
https://inews.co.uk/essentials/lifestyle/cars/car-news/uk-set-to-get-its-first-15mph-speed-limits-295471

PR #3208, opened 2021-08-23
improve clarity in privacy info [ready for merging]

matkoniecz:
in typical cases it is only technically implictly and for some cases (way splitting) it is even 100% explicit

This word serves no useful purpose here and I think it can be removed and improving meaning/clarity

Opening PR as I am not fully trusting my English.

### DIFF
--- a/app/src/main/res/values-en/strings.xml
+++ b/app/src/main/res/values-en/strings.xml
@@ -173,7 +173,7 @@ First off, understand that with this app, you make actual and direct contributio
 Anything you contribute in this app is directly added to the map, there is no third party in-between the app and OSM infrastructure.
 &lt;/p&gt;
 &lt;p&gt;
-Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and implicitly the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
+Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
 &lt;/p&gt;
 
 &lt;b&gt;Location&lt;/b&gt;&lt;br/&gt;
--- a/app/src/main/res/values-en/strings.xml
+++ b/app/src/main/res/values-en/strings.xml
@@ -173,7 +173,7 @@ First off, understand that with this app, you make actual and direct contributio
 Anything you contribute in this app is directly added to the map, there is no third party in-between the app and OSM infrastructure.
 &lt;/p&gt;
 &lt;p&gt;
-Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
+Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and implicitly the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
 &lt;/p&gt;
 
 &lt;b&gt;Location&lt;/b&gt;&lt;br/&gt;
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -174,7 +174,7 @@ First off, understand that with this app, you make actual and direct contributio
 Anything you contribute in this app is directly added to the map, there is no third party in-between the app and OSM infrastructure.
 &lt;/p&gt;
 &lt;p&gt;
-Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
+Editing OpenStreetMap anonymously is not possible. Any changes you make (and their date and location) are attributed to your OSM user account and publicly visible on the openstreetmap.org website. Because StreetComplete should only be used for on-site survey, this reveals where and when you have used the app.
 &lt;/p&gt;
 
 &lt;b&gt;Location&lt;/b&gt;&lt;br/&gt;
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -174,7 +174,7 @@ First off, understand that with this app, you make actual and direct contributio
 Anything you contribute in this app is directly added to the map, there is no third party in-between the app and OSM infrastructure.
 &lt;/p&gt;
 &lt;p&gt;
-Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and implicitly the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
+Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
 &lt;/p&gt;
 
 &lt;b&gt;Location&lt;/b&gt;&lt;br/&gt;


smichel17 (approved):
I think this makes sense. If it read, "and implicitly *your* location", then I would prefer to keep "implicitly", but I agree it is not needed since it talks about "the location of your changes".

westnordost:
You need to edit values/strings.xml, not values-en/strings.xml

westnordost:
@smichel17 but it is **implicitly** _your_ location, because the users are using the app on-site, thus the location of their changes == their location. This is why there is "implicitly".

mnalis:
> but it is implicitly your location, because the users are using the app on-site, thus the location of their changes == their location

That is good explanation, and while one would hope users would deduce that themselves, it really is preferable to be explicit in Privacy Policy about such facts. However, original text is somewhat incorrect:

> This means, that the date, content and _implicitly the location of your changes_ are visible publicly [...]

The **_location_ of the user _changes_** is quite **explicit**. It it part of node data, and spelled exactly in OSM database - there is nothing implicit there. The user _action_ that led to the said change might have been explicit or implicit, but the _location_ of change itself is explicit.

Also, I think date/time of the change need not be the date/time of the upload (they are only very close if users chooses to use `auto-upload`), but I do not think such implementation detail is worth mentioning. The purpose of Privacy Policy is to (coarsely - otherwise it would be too big and unreadable) warn users about things they _might not like_ (and not to mention every _exception_ to that list).

So, if we want to be explicit (and correct at the same time), which I think is good idea, maybe something like this:

> This means, that the date, content and location of your changes _(and thus implicitly your location at the time)_ are visible publicly [...]

matkoniecz:
> You need to edit values/strings.xml, not values-en/strings.xml

Done.

> The location of the user changes is quite explicit.

Yes, even when just editing tags the edit is strongly linked to the location of edited objects.

> but I do not think such implementation detail is worth mentioning

Yes, and I would expect that delay between visit and upload is rarely so great to completely remove such leak of private data.

smichel17:
> @smichel17 but it is **implicitly** _your_ location, because the users are using the app on-site, thus the location of their changes == their location. This is why there is "implicitly".

@mnalis has clarified my position well. I did a pass at updating the phrasing.

> Editing OpenStreetMap anonymously is not possible. Any changes you make are publicly visible on the openstreetmap.org website and are attributed to your account, as well as the date and location of those changes. Because StreetComplete should only be used for on-site survey, this implicitly reveals your location.

app/src/main/res/values-en/strings.xml
```diff
@@ -173,7 +173,7 @@ First off, understand that with this app, you make actual and direct contributio
 Anything you contribute in this app is directly added to the map, there is no third party in-between the app and OSM infrastructure.
 &lt;/p&gt;
 &lt;p&gt;
-Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and implicitly the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
+Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.
```
peternewman, line 176: @smichel17 's suggestion for comparison.
```suggestion
Editing OpenStreetMap anonymously is not possible. Any changes you make are publicly visible on the openstreetmap.org website and are attributed to your account. So are the date and location of those changes. Because StreetComplete should only be used for on-site survey, this implicitly reveals your location.
```

peternewman, line 176: @matkoniecz alternative:
Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and the location of your changes are visible publicly on the openstreetmap website and are attributable to your account. It means that also with StreetComplete on-site survey and OpenStreetMap improvement reveals your location.

peternewman: I quite like @matkoniecz suggestion with a few minor tweaks:
```suggestion
Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and the location of your changes are visible publicly on the OpenStreetMap website and are attributable to your account. It means that StreetComplete on-site survey and OpenStreetMap improvements reveal where you have been.```

**Comment by peternewman** [comment]
> @smichel17 but it is **implicitly** _your_ location, because the users are using the app on-site, thus the location of their changes == their location. This is why there is "implicitly".

They **should** be using the app on-site, although aside from a warning there's nothing to stop them doing so. Also with offline workflows and non-automatic upload, it would only show where you'd been. Would it even have the correct timestamp for there, or are the changesets only opened once you go online?

> Yes, and I would expect that delay between visit and upload is rarely so great to completely remove such leak of private data.

If we're trying to stop you flagging up to a burglar that you're not at home, manual upload is probably sufficient. If you're trying to avoid showing you've been to your lover or the hospital, months or years would probably be necessary, which would provide a challenge for the merge algorithm!

Personally I'd say given there is an easy built-in way to avoid leaking data immediately (by switching to manual upload), it feels like this just needs to be a positive warning, not scaring people off. But then maybe I'm not as privacy conscious as others.

File: app/src/main/res/values-en/strings.xml (hunk 1 above)
Comment (matkoniecz) [reply app/src/main/res/values-en/strings.xml:176]:
``` is likely not intended

Comment (peternewman) [reply app/src/main/res/values-en/strings.xml:176]: Indeed, fixed! That's an odd bug if you don't have the line break!

Comment (smichel17) [reply app/src/main/res/values-en/strings.xml:176]: My version was also intended only as cleanup for easier reading.

- Is OSM really referred to as "the OpenStreetMap"? Sounds strange to me. Like using the Git[Hub] for version control.
- "changes you make are made" is awkward, almost redundant. I fixed it with changing the sentence structure. Another alternative is, "so your changes are made".
- I like putting the website url there, gives a hint about where to look for that data.
- I slightly prefer my phrasing of "the changes are public, also this includes the date and location", but no strong feelings.
- I much prefer your phrase, "reveals where you have been".

Sorry for any verbosity, hard to edit on mobile!

Comment (mnalis) [reply app/src/main/res/values-en/strings.xml:176]: > My version was also intended only as cleanup for easier reading.

> * Is OSM really referred to as "the OpenStreetMap"? Sounds strange to me. Like using the Git[Hub] for version control.

Huh? The project is called "OpenStreetMap" (with capitalisation and all, see https://www.openstreetmap.org/about), and "OSM" is just an acronym for that name.

I don't get Git[Hub]/version control reference too - those are three different things:

- [version control](https://en.wikipedia.org/wiki/Version_control) is software which keeps history of editing of the files, so previous revisions can be retrieved, compared, merged etc. There are many different and incompatible types of them: CVS, SVN, Git, Mercurial, Bazaar, ...
- [Git](https://en.wikipedia.org/wiki/Git) is one specific type of version control software, written initially by Linus Torvalds, and gaining huge popularity in recent years.
- GitHub is (currently owned Microsoft) project management suite which incorporates (among other things) indexing engine, issue trackers, wiki software, e-mail interfacing, and both web-based and ssh-based Git version control software. So Git itself is only tiny part of GitHub.

> * "changes you make are made" is awkward, almost redundant. 

Agreed,  I prefer majority of @smichel17 version here.

> * I slightly prefer my phrasing of "the changes are public, also this includes the date and location", but no strong feelings.

Huh, I don't see that exact version anywhere?

> * I much prefer your phrase, "reveals where you have been".

I too like that part from @peternewman version.

So the combined version I'd prefer best:

> Editing OpenStreetMap anonymously is not possible. Any changes you make are publicly visible on the openstreetmap.org website and are attributed to your account. So are the date and location of those changes. Because StreetComplete should only be used for on-site survey, this implicitly reveals where you have been.

> Sorry for any verbosity, hard to edit on mobile!

Those PR nested comments are hard to follow/reply in readable way on desktop too :cry:

**Comment by westnordost** [comment]
@matkoniecz tell me when I should look at/merge it, didn't follow the discussion and don't know if they are resolved yet or not.

File: app/src/main/res/values-en/strings.xml (hunk 1 above)
Comment (peternewman) [reply app/src/main/res/values-en/strings.xml:176]: > > * Is OSM really referred to as "the OpenStreetMap"? Sounds strange to me. Like using the Git[Hub] for version control.
> 
> Huh? The project is called "OpenStreetMap" (with capitalisation and all, see https://www.openstreetmap.org/about), and "OSM" is just an acronym for that name.
> 
> I don't get Git[Hub]/version control reference too - those are three different things:

I think @smichel17 was referring to the seemingly redundant "the" at the start, which you dropped in your reply comment. I think it should probably be "the openstreetmap.org website" or "OpenStreetMap" (but we could probably capitalise in the website name too). They may have a preferred style guide? Or not, they use with and without www on https://www.openstreetmap.org/copyright

> > * I slightly prefer my phrasing of "the changes are public, also this includes the date and location", but no strong feelings.
> 
> Huh, I don't see that exact version anywhere?

It's probably in the discussion thread, but not as a suggestion.
 
> I too like that part from @peternewman version.

I'm not sure I've actually added one have I, just added other people's as suggestions, as I'm about to do for @mnalis as I far prefer to concentrate on what's changed:

```suggestion
Editing OpenStreetMap anonymously is not possible. Any changes you make are publicly visible on the openstreetmap.org website and are attributed to your account. So are the date and location of those changes. Because StreetComplete should only be used for on-site survey, this implicitly reveals where you have been.
```

Finally, some minor tweaks on @mnalis suggestion to give my own one (but I wouldn't object to dropping "therefore" or my final change about are/have been):
```suggestion
Editing OpenStreetMap anonymously is not possible. Therefore any changes you make are publicly visible on the openstreetmap.org website and are attributed to your account; as are the date and location of those changes. Because StreetComplete should only be used for on-site survey, this implicitly reveals where you are or have been.
```

matkoniecz:
current:

> Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and implicitly the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.

minimal, initially proposed rewrite:

> Editing the OpenStreetMap anonymously is not possible, so any changes you make are made with your OSM user account. This means, that the date, content and the location of your changes are visible publicly on the openstreetmap website and are attributable to your account.

rewrite based on feedback collected here:

> Editing OpenStreetMap anonymously is not possible. Any changes you make are publicly visible on the openstreetmap.org website and are attributed to your account. So are the date and location of those changes. Because StreetComplete should only be used for on-site survey, this reveals where you have been.

more radical rewrite:

> On-site survey used to improve OpenStreetMap reveal where you have been, also when done with StreetComplete. Date, content and the location of your changes are visible publicly on the OpenStreetMap website and are attributable to your account.

matkoniecz:
I am pretty sure about doing at least minimal change, but I would appreciate feedback also on stronger edits.

FloEdelmann:
I like the "rewrite based on feedback collected here", but I'd reword it a bit:

> Editing OpenStreetMap anonymously is not possible. Any changes you make (and their date and location) are attributed to your account and publicly visible on the openstreetmap.org website. Because StreetComplete should only be used for on-site survey, this reveals where and when you have used the app.

mnalis:
I too like "rewrite based on feedback collected here", especially @FloEdelmann rewording.

matkoniecz:
One more round

> are attributed to your account

I would restore

> are attributed to your OSM user account

as otherwise it may seem that it is linked to for example Google user account associated with a phone (fairly typical on Android devices), for



> Editing OpenStreetMap anonymously is not possible. Any changes you make (and their date and location) are attributed to your OSM user account and publicly visible on the openstreetmap.org website. Because StreetComplete should only be used for on-site survey, this reveals where and when you have used the app.

smichel17 (approved):

FloEdelmann (approved):

westnordost (approved):

peternewman (approved):
LGTM

Issue #3034, opened 2021-07-08
Allow to save/load current state of enabled quests


mnalis:
**Use case**
As noted in https://github.com/streetcomplete/StreetComplete/issues/1654 and elsewhere, enabling/disabling of quests and their ordering is very time consuming task (and only becoming more time consuming as new quests are added in the future). 

And yet users often have a need  to have separate sets of quests to use in different circumstances - either due to their mode of movement (bus, foot, ...), time of year (snow covering surface quests), time of day (lit), their priorities and available time (entering restaurants) etc.  See https://github.com/streetcomplete/StreetComplete/discussions/2457 for list of use cases.

It would help users a lot, if instead of having to keep rearranging and enabling/disabling quests on regular bases, they could do it only **once**, and later reuse saved settings.

**Proposed Solution**

If full-fledged custom presets linked above are deemed too problematic to implement, perhaps it would be possible to at least allow users to **save** their current quest list (both `on`/`off` status of various quests, **and** their priorities) to file, and later **load** them from (that or some different file). It could be made available from menu in Quests list like this:
![quests_saveload](https://user-images.githubusercontent.com/156656/124989047-07010180-e03f-11eb-83f4-9797e6c70ba6.png)

It would work like one expects from any program working with files:
- `Save...` would ask for filename in which to save this set of quests (and if file with that name already exists, ask if we want to overwrite it)
- `Load...` would present a list of all files (in a SC folder dedicated for saved quests), and after user clicks on one, current list of quests would be replaced by list of quests that was in that file.

That would keep it simple, unobtrusive (to users not interested in it), and yet allow users to avoid manual drudgery when their changed circumstances require different set of quests.

westnordost:
Okay, it is hard to talk about the situation when a new implementation/mitigation is done but not released yet. I think, having quest presets at all is going to be much less used/requested/necessary after the stuff on the reorder-quests branch is merged.

mnalis:
@westnordost well I actually tried it before it was officially released (if people are comfortable with running untrusted APKs, they can download APK from https://github.com/mnalis/StreetComplete/releases/tag/v33.0-mnalis2-questorder and try it themselves), and while it does make defaults better, it does not solve the problem for me - which is about ability to choose from several **different sets** at **different times** (and not having better **one** set, which seems to me to be what `quest-order` branch does?) . 

[...]

PR #3210, opened 2021-08-24
Quest profiles

westnordost:
implements #3034


app/src/main/res/raw/changelog.yml
```diff
@@ -1,4 +1,14 @@
-34.0-beta1: |
+v34.0-beta1: |
+  <h3>🚌🚴 Quest presets</h3>
+  <ul>
+  <li>Show quests that are quick and easy to solve (from a distance) first by default. This makes
+      contributing more efficient and diverse while also easier to do f.e. as a passenger (#2944, ...)</li>
+  <li>You can now save different quest presets for different situations, f.e. one preset when you
```
FloEdelmann, line 6:
```suggestion
      contributing more efficient and diverse while also easier to do as e.g. a passenger (#2944, ...)</li>
  <li>You can now save different quest presets for different situations, e.g. one preset when you
```

I've personally never heard of the abbreviation `f.e.`, but only `e.g.`. Apparently it is a thing though: https://www.quora.com/What-does-the-abbreviation-f-e-mean

What's your opinion on this @peternewman @smichel17?

app/src/main/res/values/strings.xml
```diff
@@ -487,12 +487,12 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="resurvey_intervals_less_often">Ask less often</string>
     <string name="resurvey_intervals_default">Default</string>
     <string name="resurvey_intervals_more_often">Ask more often</string>
-    <string name="pref_title_quests2">Quest selection and priority</string>
+    <string name="pref_title_quests2">Quest selection and display order</string>
     <string name="pref_subtitle_quests">%1$d of %2$d enabled</string>
+    <string name="pref_subtitle_quests_preset_name">Preset: %1$s</string>
```
FloEdelmann, line 492: This should be enough, right?

```suggestion
    <string name="pref_subtitle_quests_preset_name">Preset: %s</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1068,4 +1070,14 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_camera_type_dome">Dome</string>
     <string name="quest_camera_type_fixed">Fixed</string>
     <string name="quest_camera_type_panning">Panning</string>
+    <string name="no_search_results">The what now?</string>
```
FloEdelmann, line 1073: Shouldn't this be a more formal wording?

app/src/main/res/values/strings.xml
```diff
@@ -1068,4 +1070,14 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
+    <string name="quest_selection_hint_title">Too much to do?</string>
+    <string name="quest_selection_hint_message">"Contributing should primarily be fun!
+
+If you are overwhelmed by the number of quests, you can always fine-tune what quests are displayed to you and in what order in the settings.</string>
```
FloEdelmann, line 1077:
```suggestion
If you are overwhelmed by the number of quests, you can always fine-tune which quests are displayed to you and in which order in the settings.</string>
```

CC @peternewman @smichel17 again.

FloEdelmann (commented):
Just a quick (preliminary) code-only review. I don't know if I manage to do a full review, so I just submit this now already.

Thanks for implementing @westnordost! :blush:

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteSQLiteOpenHelper.kt
```diff
@@ -96,8 +101,30 @@ import de.westnordost.streetcomplete.quests.oneway_suspects.data.WayTrafficFlowT
         if (oldVersion == 1 && newVersion > 1) {
             db.execSQL(CreatedElementsTable.CREATE)
         }
+        if (oldVersion == 2 && newVersion > 2) {
```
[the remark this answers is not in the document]
westnordost: Hmm yeah, I think you are right! Thank you!

app/src/main/res/raw/changelog.yml
westnordost: Oh, I didn't know that f.e. was so unknown / unused, so I have done it wrong my whole life. Thanks! Gonna do a string-replace...

app/src/main/res/values/strings.xml
westnordost: yes

app/src/main/res/values/strings.xml
westnordost: I don't know, why should it?

matkoniecz: Informal seems fine and even preferable - as long as it is not some slang/insulting

FloEdelmann: I don't know what "The what now" should mean... Maybe the context where it is used will help with understanding.

smichel17:
Apparently you can only request reviews from folks who have explicitly been granted access to the repo, but @Helium314 might be interested too

app/src/main/res/values/strings.xml
westnordost: Well I hope it means something like "hä, was jetzt?" but maybe i misunderstood is meaning

smichel17: "hä, was jetzt?" auto-translates to "huh, what now?" As in, "I'm not sure what to do now.". I *think* that's more what you meant.

I would interpret "The what now?" more like "Sorry, what?" or "Are you talking nonsense at me?" Or, in gif form, 

white guy blinking meme

app/src/main/res/values/strings.xml
smichel17: I think it should be "which quests… what order"
```suggestion
If you are overwhelmed by the number of quests, you can always fine-tune which quests are displayed to you and in what order in the settings.</string>
```

app/src/main/res/values/strings.xml
westnordost: Well, that would be fine, ain't it?

westnordost: It is shown when the search word(s) entered do not match any quest types, so the list is empty.

peternewman: It's understandable as a native, but it's very informal/casual, almost cryptic, similar to:
https://www.theguardian.com/lifeandstyle/wordofmouth/2014/mar/25/wackaging-trend-food-packaging-innocent-language

I think you may need to add a hint for translators so they get the right idea.

It also seems out of place compared to the rest of the language used in the app (e.g. connection failures) which is generally a bit less "fun".

peternewman: Although "Too much to do?" below seems similar too.

mnalis: I have no idea what "the what now" should mean either... Something like `No matches` would be shorter, more common, and more understandable. But maybe that is the point (like [borken](https://www.landley.net/history/mirror/jargon.html#borken) for broken)

Or, if we're going for such level of borken/informal, we could go with ["What you say !!"](https://knowyourmeme.com/memes/all-your-base-are-belong-to-us) meme at least :smile:

westnordost: Do you have a suggestion for a slightly less "cryptic" wording, @ native speakers?

> Although "Too much to do?"

That's pretty much German ("Zu viel zu tun?") translated into English. If it doesn't work that well in English, please post a suggestion!

peternewman: Well this is equally applicable, but less fun, but should be more understandable for people who might be using it in a non-native language, or translating it.
```suggestion
    <string name="no_search_results">No results found</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1068,4 +1070,14 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_camera_type_dome">Dome</string>
     <string name="quest_camera_type_fixed">Fixed</string>
     <string name="quest_camera_type_panning">Panning</string>
+    <string name="no_search_results">The what now?</string>
+    <string name="quest_selection_hint_title">Too much to do?</string>
```
peternewman, line 1074: > Although "Too much to do?"

That's pretty much German ("Zu viel zu tun?") literally translated into English. If it doesn't work that well in English, please post a suggestion!

_Originally posted by @westnordost in https://github.com/streetcomplete/StreetComplete/pull/3210#discussion_r695644574_

Yeah it works fine in English, I just wonder if there is a bit of a negative connotation about work or it being a job (rather than the gamification side of things). i.e. these are your tasks to complete.

This seems a little softer or just ambiguous (regarding why you're completing the quests) to me:
```suggestion
    <string name="quest_selection_hint_title">Too many quests?</string>
```

Or alternatively to put a more positive spin on it:
```suggestion
    <string name="quest_selection_hint_title">Too much to choose from?</string>
```

westnordost: Hmm well, the point is that it should not feel like work and this very dialog addresses that. No point in denying that it may feel like work if there is too much to do:

> Too much to do?
>
> Contributing should primarily be fun!
> 
> If you are overwhelmed by the number of quests, you can always fine-tune which quests are displayed to you and in what order in the settings.</string>

app/src/main/res/values/strings.xml
westnordost: Meh. People don't do their taxes in StreetComplete.

PR #3213, opened 2021-08-27
Fix location pin showing after location is lost

smichel17:
Steps to reproduce:

- Turn on location services and get a location fix
- Turn off location services
- Scroll so your location is off screen

Before: Location pin points back to last known location
Now: Location pin does not appear

(Also described in https://github.com/streetcomplete/StreetComplete/issues/3147#issuecomment-895581863)


westnordost:
Code looks fine, thank you

PR #3223, opened 2021-08-30
Use LocationManagerCompat.isLocationEnabled().

Isira-Seneviratne:


westnordost:
I looked at the implementation of `LocationManagerCompat.isLocationEnabled`, it is a different implementation than `LocationUtil.isLocationOn`. For example, it does not check if the app has location permissions. But it looks like that is OK, `locationManager.isProviderEnabled` does not have a  location permission requirement since Lollipop.

Though, it would still be better if this change is tested before it is merged (E.g. don't have location permissions and start the app; grant location permissions later; go back into settings while app is running in background and remove location permissions again, return to app). Did you test it?

Also, this can also only be merged after the upgrade-to-minSDK21 branch is merged.

Isira-Seneviratne:
I got a crash on startup with the change, so I added the location permission request.

I also fixed a compilation error.

peternewman:
> Also, this can also only be merged after the upgrade-to-minSDK21 branch is merged.

Would it make sense merging this into that branch instead? Then this PR itself can be merged and a pile of changes can build up which have the same restricted compatibility.

westnordost:
upgrade to minsdk21 branch is already merged to master.

Really odd that you get a crash on startup though, because the documentation comment claims that the permission is only needed before lollipop.

westnordost:
Okay, but I reckon that you tested it, corrected it now and now it works, so it could be merged right?

peternewman:
> upgrade to minsdk21 branch is already merged to master.

:facepalm:

Isira-Seneviratne:
> Really odd that you get a crash on startup though, because the documentation comment claims that the permission is only needed before lollipop.

Yeah, I found it odd too. When I re-added the call to `hasLocationPermission`, it requested permission for location access.

westnordost:
Ok, thank you!

PR #3224, opened 2021-08-30
Use ActivityCompat.recreate().

Isira-Seneviratne:
Use [`ActivityCompat.recreate()`](https://developer.android.com/reference/androidx/core/app/ActivityCompat#recreate(android.app.Activity)) to backport the method's behavior to Android versions below P.


westnordost:
Okay, seems reasonable. Thanks!

PR #3232, opened 2021-08-31
Use NotificationChannelCompat.

Isira-Seneviratne:


app/src/main/java/de/westnordost/streetcomplete/data/download/DownloadNotificationController.kt
```diff
@@ -2,18 +2,17 @@ package de.westnordost.streetcomplete.data.download
...
+import androidx.core.app.NotificationManagerCompat
 import de.westnordost.streetcomplete.ApplicationConstants
 import de.westnordost.streetcomplete.MainActivity
 import de.westnordost.streetcomplete.R
 
 /** Shows the download progress in the Android notifications area */
 class DownloadNotificationController(
     private val service: Service,
-    private val notificationChannelId: String,
+    notificationChannelId: String,
```
westnordost, line 15: should stay private, no?

westnordost (changes requested):
Looks good to me, only that one thing about private field

app/src/main/java/de/westnordost/streetcomplete/data/download/DownloadNotificationController.kt
Isira-Seneviratne: I moved the notification channel creation code to `createNotificationBuilder`, so the IDE suggested removing the `private val` part of the parameter declaration.

westnordost: But why?

Isira-Seneviratne: Because its only use is in that method, which is used to initialize a class variable.

westnordost: ah, now I get it, ok

westnordost (approved):

Issue #3249, opened 2021-09-03
Don't ask for vegetarian meals if it's already marked as vegan


khao-soi:
When optional tasks vegetarian / vegan meals on menu are enabled, and you mark a restaurant as vegan, you are still being asked if it has vegetarian meals on the menu. I here agree with the [OSM Wiki](https://wiki.openstreetmap.org/wiki/Key:diet:vegan) that vegetarianism is implied when the restaurant is already marked as vegan.

Thank you.

Helium314:
If a restaurant is `vegan=only`, the vegetarian quest is not asked.
If a restaurant is `vegan=yes`, the vegetarian quest ist still useful to choose between `vegetarian=only` and `vegetarian=yes`

khao-soi:
You're right. Meals could be all vegetarian and some of them vegan. But it still could lead to many redundant `vegan=yes` && `vegetarian=yes` taggings. I guess I'll just ignore and hide the quest.

10992-osm:
What happens if someone says there are no vegetarian meals when `vegan=yes` is set?  Or is it not currently an option?

mnalis:
> What happens if someone says there are no vegetarian meals when vegan=yes is set?

@10992-osm good question! You should try it out and if it is broken, report a new issue (as closed issues get little love).

With my (quite limited!) understanding of the  [AddVegetarian.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt)  likely nothing special would happen -- so "impossible" combination `vegan=yes` + `vegetarian=no` would be tagged. 

Perhaps such invalid combination should be catched and additional action taken (`are you sure, because it is mapped as Vegan?!`) ?

Or (I don't know if there is capability in SC for that) if `vegan=yes` then disable `No` as an answer for `Vegetarian` (and have users leave a note if diet is wrongly mapped).

Helium314:
Vegan is asked only if vegetarian is yes or only.
So vegan=yes would have to be added with a different editor without adding vegetarian (not sure how often this happens) _and_ vegan=yes has to be wrong. Looking at overpass, I found 2 places with this invalid combination in all Europe, both not tagged with StreetComplete.

And I doubt that restaurants would actually switch from vegan=yes to vegetarian=no.
So I am reasonably certain that no one would ever see such a dialog.

mnalis:
[...]

PR #3253, opened 2021-09-04
Remove diet:vegan if vegetarian quest is answered with no

Helium314:
see #3249 
If a place has `diet:vegan != only` and the user answers the vegetarian quest with _no_, the vegan key is obsolete and thus can be removed.
Additionally this avoids the invalid combination of `diet:vegan=yes` and `diet:vegetarian=no`

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
@@ -31,5 +31,7 @@ class AddVegetarian : OsmFilterQuestType<DietAvailability>() {
 
     override fun applyAnswerTo(answer: DietAvailability, changes: StringMapChangesBuilder) {
         changes.updateWithCheckDate("diet:vegetarian", answer.osmValue)
+        if (answer.osmValue == "no")
+            changes.deleteIfExists("diet:vegan")
     }
 }


westnordost:
Makes sense thanks

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt
```diff
@@ -31,5 +31,7 @@ class AddVegetarian : OsmFilterQuestType<DietAvailability>() {
 
     override fun applyAnswerTo(answer: DietAvailability, changes: StringMapChangesBuilder) {
         changes.updateWithCheckDate("diet:vegetarian", answer.osmValue)
+        if (answer.osmValue == "no")
+            changes.deleteIfExists("diet:vegan")
```
rugk, line 35: I don't know SC exact implementation, but would not this also delete `diet:vegan = no`?

westnordost: yes

rugk: But is not this problematic as it removes correct information (or actually more detailed information in case of `only`)?

westnordost: Can you point me to a restaurant that doesn't have vegetarian food but only has vegan food?

matkoniecz: 1) In case that it is "no" it is just not needed duplicate

2) in case that it is not no it would be wrong (or SC mapper is mistaken)

Helium314: Initial idea of this PR was to remove vegan=yes only if SC mapper answered vegetarian=no, because
* the SC user could not have tagged vegan before getting the vegetarian quest (it's simply not asked)
* assume that SC user does give a correct answer (vegan=yes might be outdated)

I decided to remove vegan tag always if vegetarian=no because it's easier and the tag is not needed as its value must be no.

rugk: Hmm yeah, it all seems to depend on the fact that the user's answer is correct. I guess that's a fair thing to assume, respectively we cannot handle it otherwise.

mnalis: Well yes, because if we assume that users will give incorrect answers, we might as well disable all quests altogether, not just this one :smile:

carstenhag: >Can you point me to a restaurant that doesn't have vegetarian food but only has vegan food?
@westnordost not sure if it's still relevant, but as nobody answered I will :).

There are growing amounts, but probably less than 100 in Germany. Two examples: 
https://www.xn--tomundmx-6za.de/ and http://www.die-gartenkueche.com/ (ate a cake there once, was nice!)

westnordost: image

Daal-soup is vegetarian. And vegan.

carstenhag: No, at the bottom right it says "alles vegan" :)

carstenhag: At [happycow](https://www.happycow.net/searchmap/?s=3&location=munich&veganStores=on&filters=vegan-bakery-bnb-catering-coffee-delivery-farmers-foodtruck-health-icecream-juicebar-marketvendor-organization-other-veganprofessional-spa-vegshop&metric=km&limit=81&order=default&lat=48.134255858400735&lng=11.580030835505575&zoom=15.25&page=1&bb=48.12187655432498%2C11.56270890312349%2C48.146635162476485%2C11.597352767887658) you can also search for all-vegan stores, restaurants and co (does not use OSM as data source though).

mnalis: @carstenhag I though the question _"name a vegan restaurant which is not **also** a vegetarian restaurant"_ was supposed to be rhetoric. As far as I know, `vegan` is subset of `vegetarian`, so by definition, if it is `vegan`, it also **must** be `vegetarian`. (Same for example as every _electric car_ is also **always** a _motor vehicle_)

carstenhag: Not sure if I got you, but yeah, vegan is the most strict, vegetarian is less strict.  
But the tag can also have a diet:vegan=only value, then diet:vegetarian=yes should imo not be set. 

(This was just to try to explain again, I am not sure if it's implemented like this already)

Helium314: vegetarian is not asked is a place is vegan only

PR #3259, opened 2021-09-05
Use WindowInsetsAnimationCompat.

Isira-Seneviratne:
Use `WindowInsetsAnimationCompat` to backport the keyboard show/hide animations to Android versions below R.


app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
```diff
@@ -116,7 +116,7 @@ open class MapFragment : Fragment(),
         mapTileProviderLink.text = vectorTileProvider.copyrightText
         mapTileProviderLink.setOnClickListener { showOpenUrlDialog(vectorTileProvider.copyrightLink) }
 
-        attributionContainer.respectSystemInsets(View::setMargins)
+        attributionContainer.respectSystemInsets { setMargins(it.left, it.top, it.right, it.bottom) }
```
westnordost, line 119: For here and other places, I think it would be more convenient if there was an extension function that accepts an `Insets`. I.e. something like

```kotlin
fun View.setMargins(insets: Insets) = 
  setMargins(insets.left, insets.top, insets.right, insets.bottom)
```
and the same for `setPadding`

westnordost: (There is a View.kt somewhere in utils where you could add that function)

feliperodriguess:
Gjdscbjj

app/src/main/java/de/westnordost/streetcomplete/map/MapFragment.kt
Isira-Seneviratne: Sure.

app/src/main/java/de/westnordost/streetcomplete/ktx/View.kt
```diff
@@ -73,6 +74,16 @@ fun View.showTapHint(initialDelay: Long = 300, pressedDelay: Long = 600) {
...
+@Suppress("NOTHING_TO_INLINE")
+inline fun View.setPadding(insets: Insets) {
+    setPadding(insets.left, insets.top, insets.right, insets.bottom)
+}
+
+@Suppress("NOTHING_TO_INLINE")
+inline fun View.setMargins(insets: Insets) {
+    setMargins(insets.left, insets.top, insets.right, insets.bottom)
+}
+
```
westnordost, line 86:
```suggestion
fun View.setPadding(insets: Insets) {
    setPadding(insets.left, insets.top, insets.right, insets.bottom)
}

fun View.setMargins(insets: Insets) {
    setMargins(insets.left, insets.top, insets.right, insets.bottom)
}

```

As far as I understand, inlining only really makes sense if a function has a function parameter. See also the documentation:

> If an inline function has no inlinable function parameters and no reified type parameters, the compiler will issue a warning, since inlining such functions is very unlikely to be beneficial

https://kotlinlang.org/docs/inline-functions.html#noinline

westnordost: Thanks!

westnordost:
Alright, looks good, thank you!

PR #3272, opened 2021-09-08
Refactor CacheConfig init block

smichel17:
Untested, I just did it in the GitHub interface when I was looking at the file for unrelated reasons. If you think it looks good, I'll test it.

It could be refactored to get rid of the init block entirely, and just chain the second `let` directly off the first, when assigning `cache`. I didn't find it much more readable that way, although it would semantically make clear that "all the stuff in this block is just about generating the value of `cache`."


westnordost:
No, this style is not homogenous with the rest of the code. `let `is not used like that in the code, the same with `also`.

PR #3273, opened 2021-09-09
Add updateCheckDate and deleteOtherCheckDates functions and switch to them

peternewman:
Makes #2754 easier and also reduces the risk of future bugs.

Entirely untested sorry, although hopefully the new CI will at least have run the tests against it...


app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
```diff
@@ -43,10 +42,7 @@ class SpecifyShopType : OsmFilterQuestType<ShopTypeAnswer>() {
...
 
     override fun applyAnswerTo(answer: ShopTypeAnswer, changes: StringMapChangesBuilder) {
-        val otherCheckDateKeys = LAST_CHECK_DATE_KEYS.filterNot { it == SURVEY_MARK_KEY }
-        for (otherCheckDateKey in otherCheckDateKeys) {
-            changes.deleteIfExists(otherCheckDateKey)
-        }
+        changes.deleteOtherCheckDates()
         when (answer) {
             is IsShopVacant -> {
                 changes.deleteIfExists("shop")
```
peternewman, line 48: Should this not also delete the survey tag, and therefore the whole quest could just remove all survey tags?
```suggestion
                changes.deleteIfExists("shop")
                changes.deleteIfExists(SURVEY_MARK_KEY)
```

westnordost (commented):
Wow cool, that's a nice refactor also independent of a possible purpose to make that taginfo JSON generation easier!

app/src/test/java/de/westnordost/streetcomplete/data/meta/ResurveyUtilsTest.kt
```diff
@@ -184,4 +184,72 @@ class ResurveyUtilsTest {
...
+        )))
+    }
+
+    @Test fun `deleteOtherCheckDates does not add a check date`() {
+        val builder = StringMapChangesBuilder(mapOf())
+        builder.deleteOtherCheckDates()
+        val changes = builder.create().changes
+
+        assertEquals(listOf(
+        ), changes)
```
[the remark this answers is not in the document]
peternewman: Very good point, I was too busy just copying the existing tests.

app/src/main/java/de/westnordost/streetcomplete/data/meta/ResurveyUtils.kt
```diff
@@ -13,6 +13,14 @@ fun getLastCheckDateKeys(key: String): Sequence<String> = sequenceOf(
     "$key:last_checked", "last_checked:$key"
 )
 
+val LAST_CHECK_DATE_KEYS = listOf(
+    "check_date",
+    "lastcheck",
+    "last_checked",
+    "survey:date",
+    "survey_date"
+)
```
peternewman, line 22: I put this next to getLastCheckDateKeys to ensure things would be added to both together in future.

However I see there are a handful of instances of the missing ones:
https://taginfo.openstreetmap.org/search?q=survey_date
https://taginfo.openstreetmap.org/search?q=survey:date

Would it therefore be worth me changing getLastCheckDateKeys so it does $key:i and i:$key for each i in LAST_CHECK_DATE_KEYS and avoid the duplicate lists entirely?

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
peternewman: > Hm, @matkoniecz wrote this quest.

Okay I'll wait for him to respond too.

> But looking at the code now, it makes no sense to first delete all survey keys except `SURVEY_MARK_KEY` and then delete the `SURVEY_MARK_KEY` anyway.

What about CheckShopType; if we do:
```kotlin
changes.deleteIfExists(SURVEY_MARK_KEY)
changes.addOrModify(SURVEY_MARK_KEY, LocalDate.now().toCheckDateString())
```
Does it compress that down to one change, or should we try and avoid writing code like that?

> So in this case, I guess, writing
> at the very start would be correct because the check date(s) are deleted no matter what the user answers.

I'd probably write a deleteCheckDates function to match the others and use that directly.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopType.kt
```diff
@@ -43,10 +43,7 @@ class CheckShopType : OsmElementQuestType<ShopTypeAnswer> {
...
 
     override fun applyAnswerTo(answer: ShopTypeAnswer, changes: StringMapChangesBuilder) {
-        val otherCheckDateKeys = LAST_CHECK_DATE_KEYS.filterNot { it == SURVEY_MARK_KEY }
-        for (otherCheckDateKey in otherCheckDateKeys) {
-            changes.deleteIfExists(otherCheckDateKey)
-        }
+        changes.deleteOtherCheckDates()
         when (answer) {
             is IsShopVacant -> {
                 changes.addOrModify(SURVEY_MARK_KEY, LocalDate.now().toCheckDateString())
```
peternewman, line 49: This could just become:
```suggestion
                changes.updateCheckDate()
```

Although it would then try deleting the other keys twice, although that could be solved by putting a deleteCheckDates in the is ShopType bit.

app/src/main/java/de/westnordost/streetcomplete/data/meta/ResurveyUtils.kt
westnordost: Nah, it's fine

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
westnordost: Well current behavior is to not set the check date if anything else is tagged

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopType.kt
westnordost: No, it's fine like it is now.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
westnordost: Ok I have looked at it again, please just write 
```
        for (otherCheckDateKey in LAST_CHECK_DATE_KEYS) {
            changes.deleteIfExists(otherCheckDateKey)
        }

```
and remove the other check date related stuff.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopType.kt
peternewman: I'd like to create a function that encompasses this bit at least (even if the delete other is kept elsewhere):
`changes.addOrModify(SURVEY_MARK_KEY, LocalDate.now().toCheckDateString())`

Unless you really object, as that will continue to simplify the TagInfo JSON stuff (as I don't need to understand that I have to ignore LocalDate.now()).

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
peternewman: Do you mean for SpecifyShopType or CheckShopType.

I'd still rather write a deleteCheckDates function to drop them all, unless you're strongly in favour of the for loop?

app/src/main/java/de/westnordost/streetcomplete/data/meta/ResurveyUtils.kt
peternewman: I assume you mean not changing the function *and* not deleting those other check date sub-keys?

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopType.kt
westnordost: IIRC that executing this code

```
changes.deleteOtherCheckDates()
changes.updateCheckDate()
```

will result in an Exception, because the changes builder verifies that each change is only done once. This is why I wrote to leave it as it is.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
westnordost: > Do you mean for SpecifyShopType or CheckShopType.

SpecifyShopType 

> I'd still rather write a deleteCheckDates function to drop them all, unless you're strongly in favour of the for loop?

A helper function is fine. I was referring to the fact that independent of the answer provided in that quest, all check dates should be removed.

app/src/main/java/de/westnordost/streetcomplete/data/meta/ResurveyUtils.kt
westnordost: Yes, not changing the function and not deleting those other check date sub keys, i.e. change nothing here.

- it makes reading the code a little more complicated
- but more importantly, it mixes tagging patterns for check dates for the whole object and namespaced tagging patterns for check-dates of single keys. OpenStreetMap tagging is not that "regular". `surveydate` is used (a lot) for the whole object, but not really used at all with namespace. We shouldn't put it in the code if it practically does not exist just to save a few lines of code / because it looks more regular that way.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/SpecifyShopType.kt
peternewman: Done.

app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedBuildingConstruction.kt
```diff
@@ -33,7 +33,7 @@ class MarkCompletedBuildingConstruction : OsmFilterQuestType<CompletedConstructi
                     changes.modify("building", value)
                     deleteTagsDescribingConstruction(changes)
                 } else {
-                    changes.addOrModify(SURVEY_MARK_KEY, LocalDate.now().toCheckDateString())
+                    changes.updateCheckDate()
```
peternewman, line 36: This will remove other non-standard check dates, which I assume is a good thing?

app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedHighwayConstruction.kt
```diff
@@ -48,7 +48,7 @@ class MarkCompletedHighwayConstruction : OsmFilterQuestType<CompletedConstructio
                     changes.modify("highway", value)
                     deleteTagsDescribingConstruction(changes)
                 } else {
-                    changes.addOrModify(SURVEY_MARK_KEY, LocalDate.now().toCheckDateString())
+                    changes.updateCheckDate()
```
peternewman, line 51: Likewise again.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopType.kt
peternewman: I've added a addOrModifyCheckDate function and switched to that to simplify the TagInfo JSON stuff.

peternewman:
I think this is ready for review/merge, it passes CI here:
https://github.com/peternewman/StreetComplete/runs/3593932451?check_suite_focus=true

westnordost (approved):

westnordost:
Cool, thank you!

westnordost:
Did you work on this until 4 in the morning?

peternewman:
> Did you work on this until 4 in the morning?

Not until, but at (although it was only 3am). I did it during my break on some Night shifts (so the code may be of questionable quality :laughing: ).

westnordost:
What are you working as?

peternewman:
> What are you working as?

I'm a broadcast engineer so we cover 24/7/365.

westnordost:
Cool, so you are the guy that says "you are on air in 5, 4, 3, 2, 1..."?

peternewman:
No, that would be a floor manager (although we have computers to do some of that counting automatically now).

I'd be the person repairing their camera or microphone if it broke (we also don't operate the equipment, there are other people who do that).

PR #3274, opened 2021-09-09
Add Location.toLatLon util function

smichel17:
As discussed in https://github.com/streetcomplete/StreetComplete/pull/3166#issuecomment-903337436

As I usually try for, going through this commit-by-commit may give you the most useful diffs, especially since the first 3 here are boilerplate.


app/src/main/java/de/westnordost/streetcomplete/quests/SplitWayFragment.kt
```diff
@@ -76,7 +76,7 @@ class SplitWayFragment : Fragment(R.layout.fragment_split_way),
         osmQuestKey = Json.decodeFromString(args.getString(ARG_OSM_QUEST_KEY)!!)
         way = Json.decodeFromString(args.getString(ARG_WAY)!!)
         val elementGeometry: ElementPolylinesGeometry = Json.decodeFromString(args.getString(ARG_ELEMENT_GEOMETRY)!!)
-        positions = elementGeometry.polylines.single().map { LatLon(it.latitude, it.longitude) }
+        positions = elementGeometry.polylines.single()
```
smichel17, line 79: There's not quite as many usages of `Location.toLatLon()` as I though there would be, because it turns out the list we're mapping over here is already a `List<LatLon>`. So I removed the redundant(?) `map`.

I'm not sure if there's a reason why we want a fresh set of objects here, so if there is, I'll revert this part.

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.kt
```diff
@@ -0,0 +1,29 @@
...
+fun hasLocationPermission(context: Context): Boolean {
+    return ContextCompat.checkSelfPermission(context,
+            Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
+}
+
+fun createLocationAvailabilityIntentFilter(): IntentFilter {
+    return IntentFilter(LocationManager.MODE_CHANGED_ACTION)
+}
+
+fun Location.toLatLon(): LatLon = LatLon(latitude, longitude)
```
westnordost, line 29: If you like, you can also use the `fun xxx() = yyy` pattern (instead of the classic function declaration), it is widely used in the app.

app/src/main/java/de/westnordost/streetcomplete/quests/SplitWayFragment.kt
westnordost: `LatLon` is immutable, so passing it around freely is not a problem. I think there was a don't remember why there was a `.map`, it doesn't make sense (now)

westnordost (approved):
Very cool, looks good!

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.kt
smichel17: Ha, I meant to do that (a reformatting commit), but that was where I stopped last night and I forgot about it when I picked it up this morning :upside_down_face:

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
+
+fun isLocationEnabled(context: Context): Boolean =
+    hasLocationPermission(context) && LocationManagerCompat
+        .isLocationEnabled(ContextCompat.getSystemService(context, LocationManager::class.java)!!)
+
+fun hasLocationPermission(context: Context): Boolean =
+    ContextCompat.checkSelfPermission(context,
+        Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
```
smichel17, line 21: I'm never sure how I want to format extremely long lines like this that Kotlin sometimes produces, so I just left them the same :P

westnordost: 🤷 
Well I guess you could import `android.Manifest.permission.ACCESS_FINE_LOCATION` and `PERMISSION_GRANTED` and then just write 
```kotlin

ContextCompat.checkSelfPermission(context, ACCESS_FINE_LOCATION) == PERMISSION_GRANTED
```
You choice, I don't mind either way.

smichel17: Ok, that works.

app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.kt
```diff
@@ -0,0 +1,26 @@
...
+import android.location.LocationManager
+
+import androidx.core.content.ContextCompat
+import androidx.core.location.LocationManagerCompat
+
+import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
+
+fun isLocationEnabled(context: Context): Boolean =
+    hasLocationPermission(context) && LocationManagerCompat
+        .isLocationEnabled(ContextCompat.getSystemService(context, LocationManager::class.java)!!)
```
smichel17, line 17: This is what I'd rather do here, I think

```suggestion
fun isLocationEnabled(context: Context): Boolean =
    hasLocationPermission(context) && _isLocationEnabled(context)

private fun _isLocationEnabled(context: Context): Boolean {
    val lm = ContextCompat.getSystemService(context, LocationManager::class.java)!!
    return LocationManagerCompat.isLocationEnabled(lm)
}
```

westnordost: meh. Me no like

smichel17: Actually, found something nicer, one moment

smichel17: Same trick works again

westnordost (approved):

westnordost:
thanks!

smichel17:
Inching closer to 100% kotlin :)

Kotlin: 99.3%. Java: 0.7%

PR #3278, opened 2021-09-10
Remove keys whitelist

mnalis:
As discussed in https://github.com/streetcomplete/StreetComplete/issues/3244, done with help of @smootheFiets

When shop becomes vacant/changes, this PR transits from current blacklist (**removing all keys unless known** wanted ones in `KEYS_THAT_SHOULD_NOT_BE_REMOVED_WHEN_SHOP_IS_REPLACED`) to whitelist (**only remove keys explicitly specified** in `KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED`)

The descriptions / reasons why some key should be kept or removed is available at https://github.com/mnalis/StreetComplete-taginfo-categorize/blob/master/keys.txt

This is an initial draft.

Known issues:
- could someone who speaks kotlin (I don't) verify that the program logic is correctly modified?
- test for `KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED` in `CheckShopTypeTest.kt` should probably have more test tags to verify new behaviour is indeed working. I'll try to look into it shortly.

Comments and suggestions are always welcome


app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestController.kt
```diff
@@ -142,7 +142,7 @@ import kotlin.collections.ArrayList
 
         // first remove old tags
         for ((key, value) in previousTags) {
-            val isOkToRemove = KEYS_THAT_SHOULD_NOT_BE_REMOVED_WHEN_SHOP_IS_REPLACED.none { it.matches(key) }
+            val isOkToRemove = KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED.none { !it.matches(key) }
```
smichel17, line 145: For some reason GitHub's "suggest changes" feature doesn't appear to be available right now, but I would suggest this instead:

```
val isOkToRemove = KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED.any { it.matches(key) }
```

westnordost:
I am (positively) surprised that it's not that many tags!

Could you please sort them by type and add comments like for the whitelist? F.e. group the `email`, `contact:.*`, `facebook`, `phone` etc together

app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestController.kt
mnalis: thanks! changed in 1df4132118baf0f263956f99681e6189e06613d6

mnalis:
something like this, @westnordost ?

Note: it is automatically generated from https://github.com/mnalis/StreetComplete-taginfo-categorize/blob/master/keys.txt, so if someone wants to change / modify, please do PR there (syntax is simple - blank line means linebreak, and line starting with `//` will have that comment copied to kotlin code)

mnalis:
Also, if someone familiar with code can check if extended test in `CheckShopTypeTest.kt` makes sense?

app/src/test/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopTypeTest.kt
```diff
@@ -136,20 +136,26 @@ class CheckShopTypeTest {
...
+		"barrier" to "wall",
                 "name" to "Juppiebude"
             ),
             ShopType(mapOf("shop" to "ice_cream")),
             StringMapEntryAdd("shop", "ice_cream"),
             StringMapEntryDelete("disused:amenity", "yes"),
             StringMapEntryDelete("phone", "123456"),
             StringMapEntryDelete("name", "Juppiebude"),
+            StringMapEntryDelete("ref", "1111"),
+            StringMapEntryDelete("fee", "yes"),
```
smichel17, line 158: From a purely Kotlin perspective (ie, I have not checked whether these tags are the correct ones that should be removed in this situation, this appears to correctly add 4 tags to the initial object, and expect that 2 of them (`ref` and `fee`) should be deleted. I believe it also checks that the other two are not deleted, although I'm not familiar with this assertions library and I didn't check the docs/source of `containsExactlyInAnyOrder` to be sure.

https://github.com/streetcomplete/StreetComplete/blob/c463be9ea67fa2153fadc072b740048a87e28641/app/src/test/java/de/westnordost/streetcomplete/quests/OsmElementQuestType.kt#L13

The tags added have a different indentation than the others, though.

smichel17 (changes requested):

app/src/test/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopTypeTest.kt
mnalis: thanks! converted tabs to spaces now

smichel17 (approved):

westnordost:
Hmm, looking at this list, I wonder if it wouldn't be better to use the iD presets to decide what to delete. I.e. remove all that are in `addTags`, `removeTags` and in the fields linked by `fields` and `moreFields` properties? https://github.com/ideditor/schema-builder

matkoniecz:
I am  worried about strictly depending on iD in this way, as if something is missing/should not be present then there would be tricky to solve this.

And we would end with our own list anyway.

mnalis:
While I'm obviously biased by [sunk cost](https://en.wikipedia.org/wiki/Sunk_cost#Fallacy_effect) after spending some time working on this, I agree with @matkoniecz  there might be issues with iD idea (eg. quick look reveals `level`, which should remain when shop is vacant/changed, but is listed in `moreFields`)

westnordost:
Mh okay, fair enough.

app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
```diff
@@ -32,32 +32,35 @@ val MAXSPEED_TYPE_KEYS = setOf(
...
+    // related to repair shops/crafts
+    "service(:.*)?", "motorcycle:.*", "repair", ".*:repair", "electronics_repair(:.*)?",
+    // hairdresser
+    "unisex", "male", "female",
+    // accomodation & layout
+    "rooms", "stars", "accommodation", "beds", "capacity(:persons)?", "tourism", "laundry_service",
+    // misc specific attributes
+    "clothes", "beauty", "tobacco", "carpenter", "furniture", "lottery", "sport", "leisure", "dispensing", "tailor:.*",
+    "books", "agrarian", "musical_instrument", "massage", "parts", "post_office(:.*)?", "religion", "rental", "tickets:.*", "goods_supply",
 ).map { it.toRegex() }
```
westnordost, line 64: add
- wifi
- instagram, other contact-stuff found in contact:.?
- hairdresser, shoes, butcher, electronics_repair, pet, ... and other sub-tags of shops/crafts/amenity that are named like their shop/craft/amenity value.

Also, `isKindOfShopExpression` also includes some `leisure`, some `office` and one `tourism`. So, these need to be removed as well plus their properties.

smichel17:
iD presets might be a good place to look when updating our list, however. Maybe a comment to that effect could be added? It might also be nice to include whatever query you used to generate the list of tags.

mnalis:
@smichel17 I've added `id-tagging-schema` parsing too. It does [add](e8c6b09f342dcb58172e8f6eee7f5818bfbf9c17) `8` (rarely-used) keys, only **_some_** of which are for removing. Well, at least it is more complete now.

BTW it is good we do not use **just** `id-tagging-schema` - it would remove only `67` keys, while this full combined solution (`id-tagging-schema` + real world `>=0.01%` usage) removes `189` keys.

app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
mnalis: >     * wifi
>     * instagram, other contact-stuff found in contact:.?

added.

>     * hairdresser, shoes, butcher, electronics_repair, pet, ... and other sub-tags of shops/crafts/amenity that are named like their shop/craft/amenity value.

I'm not sure I follow. Do you mean eg. `shop=butcher` should also remove `butcher=*` ? I already do that (at least if that subkey was used at least `0.01%` or more times (or if it was mentioned in `id-tagging-schema/data/presets/`). Some do not exist, though (like `hairdresser=*`). Or did I miss some? 

> Also, `isKindOfShopExpression` also includes some `leisure`, some `office` and one `tourism`. So, these need to be removed as well plus their properties.

I've added them as main keys now (their properties were already added). I didn't initially as I (it seems wrongly) assumed main tag (the one the quest is about) is removed automatically, and that those were just extra keys to be removed beside that main tag. (That would be much nicer!)

Removing all `office=*` when `shop=convenience` will work (as in current PR), but is not ideal - for example sometimes there is a building which will be tagged with both `shop=*` and `office=*`, which are unrelated, and marking a shop vacant would remove information about offices (which it should not do). Yes, technically, it is a problem when **_tagging anything but the building_** on building closed-way, but is widely used practice unfortunately.

- Ideally, I would remove that main tags (`amenity`, `shop`, `craft`, ...) from `KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED`, and the kotlin function would remove main tag that triggered the quest separately (in addition to that list). That would be preferable. (I wouldn't know how to do it myself, even if it is possible, unfortunately).
- if that is not possible, can one use `key=value` instead of just `key` in `KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED` ? So I could say (in order to minimize damage) to remove just `office=insurance`, `office=travel_agent`, `office=tax_advisor` etc. instead of all `office=*` ?

smichel17: >  (it seems wrongly) assumed main tag (the one the quest is about) is removed automatically,

This might be a good candidate for a test.

westnordost: > I'm not sure I follow. Do you mean eg. shop=butcher should also remove butcher=* ?

Yes. Looks like I overlooked that. But I also meant generally. Try searching in taginfo for keys named like shop/leisure/craft/... values. I listeed some I found, but I did not look thoroughly.

> I've added them as main keys now (their properties were already added). I didn't initially as I (it seems wrongly) assumed main tag (the one the quest is about) is removed automatically

Well, right now, **everything** except the specified keys is removed. So when you want to turn this around, you also need to mention every tag that can be removed (i.e. is in that "is kind of shop" expression)... maybe worth a comment in the code, that if the "is kind of shop" expression is extended, your whitelist should be checked too.

westnordost: > Ideally, I would remove that main tags (amenity, shop, craft, ...) from KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED, and the kotlin function would remove main tag that triggered the quest separately

Not possible. the filter expression just returns true or false (triggered or not), it doesn't say exactly which part of the expression matched.

> if that is not possible, can one use key=value instead of just key in KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED ?

No

mnalis: > > I'm not sure I follow. Do you mean eg. shop=butcher should also remove butcher=* ?
> 
> Yes. Looks like I overlooked that. But I also meant generally. Try searching in taginfo for keys named like shop/leisure/craft/... values. I listeed some I found, but I did not look thoroughly.

Well, technically, I could make script do that, but I do not think that is such a good idea. For example, just for `craft=*`, there is `2359` different values according to [taginfo](https://taginfo.openstreetmap.org/keys/craft#values).  `shop=*` is even worse, with `11260` [values](https://taginfo.openstreetmap.org/keys/shop#values). So such heuristics would be adding tens of thousands of keys for removal - vast majority of which does not exist at all in the wild.

The script I've written finds all keys used in combination with `craft=*` and categorizes it if it is used `0.01%` or more times; so if it where listed at least that many times, it would be already categorized (either for removal, or for keeping). (For example for `craft=*`, which is not that popular, that `0.01%` works to only `5` uses worldwide.)  

> Well, right now, **everything** except the specified keys is removed. So when you want to turn this around, you also need to mention every tag that can be removed (i.e. is in that "is kind of shop" expression)... maybe worth a comment in the code, that if the "is kind of shop" expression is extended, your whitelist should be checked too.

I've added a comment to that effect now.

mnalis: > > (it seems wrongly) assumed main tag (the one the quest is about) is removed automatically,
> This might be a good candidate for a test.

Added, since it can't be improved.

smichel17: > This might be a good candidate for a test.

Hm, this is hard though, because *really* the thing we want to test is that our list of keys to remove is exhaustive. So it doesn't particularly help to test a few example tags; the important time to test is when a new tag is added. Which we can't do because the filter expression only returns a boolean.

<details><summary>I did a little dive into how filter expressions work (one more piece of SC demystified :smile:). I'd say it's <i>technically</i> possible to refactor them to return figure out <i>which</i> key matched (allowing automatic removal), but not worth the effort for this minor benefit.</summary>

You'd have to add a new method to each of the [filters](https://github.com/streetcomplete/StreetComplete/tree/master/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/filters) to return a list of tags that matched. Then change the default implementation of `matches` to check if that list is empty.

However, since I'm pretty sure this code is performance-critical for filtering elements, you'd likely have to keep the old `matches` implementation for many of the filters, so they can short-circuit in some places, so now you've got other functions you need to manually keep in sync— barely an improvement; only worth it if there's some other use for the "which key matched" functionality.
</details>

westnordost: > The script I've written finds all keys used in combination with craft=* and categorizes it if it is used 0.01% or more times

0.01% times of what? Of the particular craft value?

mnalis: > 0.01% times of what? Of the particular craft value?

For `craft` and `shop` for all of them (ie. `craft=*`). For specific ones (ie. `amenity=restaurant`), just for that specific one. 

In short, what taginfo shows as percentage for:
- https://taginfo.openstreetmap.org/keys/craft#combinations
- https://taginfo.openstreetmap.org/tags/amenity=restaurant#combinations 
- etc.

So for example for keys used in combination with `craft` (of which there are `191 204`) I've checked all keys that are used `10` or more times worldwide (should've been `>19` instead of `>10`, but taginfo rounds in different way, so we got more precision). 

Eg. `studio=*`, which is used `26` times worldwide in combination with some `craft=*` is categorized (to be removed), but `certification=*`, which is used `6` times worldwide in combination with some `craft=*` was not categorized.

In **addition** to that, all `fields` and `moreFields` from `id-tagging-schema` (no matter what the percentage)  which you suggested before have been categorized too.

westnordost: > For craft and shop for all of them (ie. craft=*). For specific ones (ie. amenity=restaurant), just for that specific one.

Hm I didn't quite understand that. What I mean is the following (just an example):

If there was a `shop=screws` which is used for at least `x` shops and `screws=*` is tagged more often than `y` of all `shop=screws`, then it should be included in the list.

Looking at taginfo, `x` should be around a magnitude 100, i.e. that seems to be around the place where the values start to lack wiki description. I feel around that number, we can assume that a tag has some reasonable usage. Could be less of course. Some numbers that are not 100 today may be 100 next year. `y` should maybe be the same number. 

Using percentages is problematic because that would mean that with the same absolute number of usages, e.g. `craft=screws` would be included, but `shop=screws` not just because the total number of shops is much higher than the total number of crafts.

mnalis: > Looking at taginfo, x should be around a magnitude 100, i.e. that seems to be around the place where the values start to lack wiki description

I don't think I can accomplish that with taginfo, it lacks resolution. Eg. when I try to see combination tags for `shop=trailer` with `103` worldwide uses, it simply does not display any combination tags at all: https://taginfo.openstreetmap.org/tags/shop=trailer#combinations. 

It seems its cut-off point is at `10 000`, so even `shop=lottery` (with `9 605` uses worldwide) does not list any key [combinations](https://taginfo.openstreetmap.org/tags/shop=lottery#combinations). That is two order of magnitude less precision that you'd like. Do you know of other API able to provide such "combined tags" data with that precision @westnordost ?

Maybe I could just process all combination tags which are used 100 or more times, would that suffice?

westnordost: Yes, you cannot see the combinations for `shop=lottery`, but you can look how many tags there are with the key `lottery`

mnalis: OK @westnordost, I will add that as it seems there is likely no damage that it can cause (apart from that Kotlin variable spanning many more lines). 

However note that my preliminary research show such stats can be very skewed. For example: there is `113` occurrences of `shop=scooter`, so it passes criteria for inclusion; and there is `2 093` occurrences of `scooter=*`, so it passes too. However in reality there is only `11` occurrences of combined `shop=scooter` + `scooter=*` worldwide.

mnalis: @westnordost I've updated script to add such "popular subkeys" (with `x=100`, `y=90`). 
So this should be resolved now.

westnordost: Cool! So, it's ready for merge / review?

mnalis: @westnordost yes, as far as I can tell, there should be no outstanding issues, so it should be ready for review (and merge if it passes review)!

app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
```diff
@@ -32,32 +32,38 @@ val MAXSPEED_TYPE_KEYS = setOf(
...
-        "geometry", "position", "shape", "outline", "location",
-        // referring to building
-        "building", "heritage", "height", "architect", "ele"
-    ).joinToString("|")})(:.*)?",
+// generated by "make update" from https://github.com/mnalis/StreetComplete-taginfo-categorize/
+val KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED = listOf(
+    "shop_?[1-9]?(:.*)?", "craft_?[1-9]?", "amenity_?[1-9]?", "old_amenity", "old_shop", "information", "leisure", "office", "tourism",
+    // "popular" shop=* / craft=* subkeys
+    "marketplace", "household", "swimming_pool", "telecommunication", "laundry", "golf", "sports", "ice_cream", "scooter", "music", "retail", "yes", "ticket", "health", "newsagent", "lighting", "truck", "car_repair", "video", "fuel", "farm", "car", "ski", "sculptor", "hearing_aids", "surf", "photo", "boat", "gas", "kitchen", "anime", "builder", "hairdresser", "security", "bakery", "fishing", "doors", "kiosk", "market", "bathroom", "lamps", "vacant", "insurance", "caravan", "gift", "bicycle", "insulation", "communication", "mall", "model", "empty", "wood", "hunting", "motorcycle", "trailer", "camera", "water", "fireplace", "outdoor",
```
westnordost, line 39: could you use about ~max 100 characters per line for better readability?

mnalis: Added auto-wrap at 100 chars.

westnordost (approved):
Looks good, thank you for the effort!

matkoniecz:
@mnalis Have you considered `wheelchair`? I think that it also should be removed.

At least in my area it is typical that it will change - either because new tenant failed to provide wheelchair accessibility, or because they cared more. In some cases former space used by customers is turned into internal one and as result `wheelchair` value changes.

mnalis:
> @mnalis Have you considered `wheelchair`? I think that it also should be removed.
> 
> At least in my area it is typical that it will change - either because new tenant failed to provide wheelchair accessibility, or because they cared more. In some cases former space used by customers is turned into internal one and as result `wheelchair` value changes.

It was considered (see https://github.com/mnalis/StreetComplete-taginfo-categorize/blob/master/keys.txt) but remained in `KEYS TO PROBABLY KEEP` section with comment:

> wheelchair(.*)? // eg. wheelchair, wheelchair:description, wheelchair_access etc. Wheelchair information probably stays correct as shop owners change (worth a separate SC quest?)

But sure, if people prefer it should be removed, it's fine with me. I'll wait for few days for possible other comments, and then add it to `KEYS TO REMOVE` section if nobody objects...

Issue #3234, opened 2021-08-31
Long ways aren't showing any quests


dbdean:
**How to Reproduce**
[This track](https://www.openstreetmap.org/way/976570335#map=17/-30.53631/151.64779&layers=N) is asking me for a surface type.

[This track](https://www.openstreetmap.org/way/976570336) is not (in fact, it has no quests at all).

Other than the size/length, I'm not sure what is different in the two tracks that could be causing one to have quests but not the other.

**Versions affected**
StreetComplete v34.0-beta1 on Android 10

smichel17:
I'm not 100% sure how the filters work, but my best guess is that one of its nodes is part of a driveway, and driveways are excluded from the surface type quest:

https://github.com/streetcomplete/StreetComplete/blob/95940485376b08782aa288f6330c8e81632d9f91/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurface.kt#L14

matkoniecz:
Longer ways are not getting quests (actually, it is one of few design decisions when I would definitely do things differently)

westnordost:
See @matkoniecz . Yeah, maybe this can be removed now. IIRC this was added before splitting ways was introduced.

matkoniecz:
Though it may be preferable to keep it for the forest quest (leaf type).

10992-osm:
If long ways are included:

How would it work (with really long ways)?  One would perhaps not see the quest pin at the start, and one may not walk along the complete length (one might turn of down a side path etc).  If one reviews the whole length of the way, one could split the way at a convenient point, but if one does not scroll over the whole way, how would this be dealt with?

matkoniecz:
Yes, for really long ways user would be more likely to miss quest - or split basing on which part was surveyed.

Maybe repeating quest icon would be nice, but may be irritating to implement

westnordost:
Alright, let's do it then. But maybe not for areas, since there is no "split area".

PR #3280, opened 2021-09-11
Allow quests on long ways

dbdean:
Initial attempt at adding enhancement so that quest show on all ways, regardless of length, to fix https://github.com/streetcomplete/StreetComplete/issues/3234.

In order to make the quest easier to spot, the icon is show at the ends, and repeated so that the maximum distance between them is always less than 500 metres.

This is only dealing with linear ways, I think. If there is a restriction on the area of area ways, it must be somewhere else in the code.


dbdean:
Ok, already noticed one potential issue:

- [ ] Zoom to quest doesn't go wide enough to show whole way.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
```diff
@@ -27,16 +28,26 @@ data class OsmQuest(
...
 
     override val markerLocations: Collection<LatLon> get() {
-        if (osmElementQuestType.hasMarkersAtEnds && geometry is ElementPolylinesGeometry) {
+        if (geometry is ElementPolylinesGeometry) {
             val polyline = geometry.polylines[0]
             val length = polyline.measuredLength()
-            if (length > 15 * 4) {
-                return listOf(
+            if (osmElementQuestType.hasMarkersAtEnds && length > 15 * 4) {
+                return kotlin.collections.listOf(
```
westnordost, line 35:
```suggestion
                return listOf(
```

westnordost (commented):
Nice! Did you test it?

westnordost:
> Ok, already noticed one potential issue:

Hmm, sounds like this should not be your problem. But the logic for that is in `KtMapController.getEnclosingCameraPosition`, if you want to look what's up.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: I have tested it in action, but I haven't added any unit tests.

dbdean:
I've tweaked the marker location code a bit, as I found that the best way to deal with the long ways is to walk the whole length, so having markers showing in the middles isn't that great.

Now the markers show at the ends like the speed limit quest, with markers repeated enough that there is no gap bigger than 500 metres.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: suggested change no longer applies with the new position logic.

dbdean:
> Nice! Did you test it?

In action, yes. In unit tests, no. Didn't seem like there were any for this area yet (but I may have missed them), and I wasn't sure where to start.

westnordost:
Testing in action is fine. 

Why 500m though? Seems to be quite the distance. Does anything speak against 250m or even 100m?

dbdean:
> Why 500m though? Seems to be quite the distance. Does anything speak against 250m or even 100m?

No major reason, Seemed like a good round distance close to the previous maximum of 600 metres. It gets a bit more frequent on short ways, as the maximum distance is 500 metres, so the actual distance can get down closer to 250-300 metres for ways just over 500 metres.

For longer ways, I think more often would probably just add clutter to the map, especially as you really need to start and one end, and walk the entire way either confirming it is the same, or splitting as you go along. Seeing a lot of markers in the middle of the way and heading to them just leaves you needing to back track, or (possibly) unnecessarily splitting the way because you chose to survey in only one direction.

I'm pretty happy with it after about 3 hours of mapping in my area.

westnordost:
> It gets a bit more frequent on short ways, as the maximum distance is 500 metres, so the actual distance can get down closer to 250-300 metres for ways just over 500 metres.

I don't understand this.

westnordost:
Maybe it would be easier (to understand, maybe in code) if there was simply a MARKER_DISTANCE.

Markers are distributed along the lenght of the way in MARKER_DISTANCE distances to each other, starting from the center.

If the markers at end flag is true, the algorithm starts from the two ends and if the way is shorter than MARKER_DISTANCE + padding at the start and end, also just a marker in the middle will be shown as if that flag is not set.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
```diff
@@ -27,16 +27,23 @@ data class OsmQuest(
...
+        if (geometry is ElementPolylinesGeometry) {
             val polyline = geometry.polylines[0]
             val length = polyline.measuredLength()
-            if (length > 15 * 4) {
-                return listOf(
-                    polyline.pointOnPolylineFromStart(15.0)!!,
-                    polyline.pointOnPolylineFromEnd(15.0)!!
-                )
+            if ((osmElementQuestType.hasMarkersAtEnds && length > MARKER_FROM_END_DISTANCE * 4)
+                || (length > MAXIMUM_MARKER_DISTANCE + 2 * MARKER_FROM_END_DISTANCE)) {
```
smichel17, line 34: I am also confused about this algorithm, and I think this is where it's coming from. Which quests don't have markers at the ends, but should have multiple markers along the way? ~~Why do quests with markers along the ends only need 4x the distance from the end before they get the same treatment?~~ I realized this is because if they're sufficiently short, they only get one marker, not on the ends. But I still don't understand why some quests would get markers at both ends even if they're short, and others only start getting that treatment if they're long enough to have markers in the middle, too.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
```diff
@@ -27,16 +27,23 @@ data class OsmQuest(
...
+                val markerCount = 2 + (length / MAXIMUM_MARKER_DISTANCE).toInt()
+                val markerDistance = (length - (2 * MARKER_FROM_END_DISTANCE)) / (markerCount - 1)
+                return (0 until markerCount).map{
+                    polyline.pointOnPolylineFromStart(
+                        MARKER_FROM_END_DISTANCE + (it * markerDistance)
+                    )!!
+                }
```
smichel17, line 41: I would expect the algorithm to be something like, for applicable ways:

```suggestion
                val midLength = length - (2 * MARKER_FROM_END_DISTANCE)
                val markerCount = (midLength / MAXIMUM_MARKER_DISTANCE).toInt()
                val markers = mutableListOf(polyline.pointOnPolylineFromStart(MARKER_FROM_END_DISTANCE)!!)
                if (markerCount > 0) {
                    val markerDistance = midLength / markerCount
                    markers.addAll((0 until markerCount).map {
                        polyline.pointOnPolylineFromStart(
                            MARKER_FROM_END_DISTANCE + (it * markerDistance)
                        )!!
                    })
                }
                markers.add(polyline.pointOnPolylineFromEnd(MARKER_FROM_END_DISTANCE)!!)
                return markers.toList()
```

It's a few more lines, but I think the intention is more understandable. It's possible to shorten or avoid mutation, if you're after that.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: Some quests always have markers at their end (speed limits, I think - maybe others) - unless they are very short.

Otherwise, only quests longer than 500 metres will have markers at their end (and then repeated if needed) - so that the mapper is encouraged to map from one end to the other. I found it worked best this way, as only having markers in the middle encouraged me to end up in the middle of the way, and it was harder to map long ways this way without backtracking.

dbdean: Some quests always have markers at their end (speed limits, I think - maybe others) - unless they are very short.

Otherwise, only quests longer than (about) 500 metres will have markers at their end (and then repeated if needed) - so that the mapper is encouraged to map from one end to the other. I found it worked best this way, as only having markers in the middle encouraged me to end up in the middle of the way, and it was harder to map long ways this way without backtracking.

dbdean:
> Maybe it would be easier (to understand, maybe in code) if there was simply a MARKER_DISTANCE.
> 
> Markers are distributed along the lenght of the way in MARKER_DISTANCE distances to each other, starting from the center.
> 
> If the markers at end flag is true, the algorithm starts from the two ends and if the way is shorter than MARKER_DISTANCE + padding at the start and end, also just a marker in the middle will be shown as if that flag is not set.

Having the marker distance vary a little instead of exactly MARKER_DISTANCE helps keep the spacing even. I have decided that all long ways have markers at the end, regardless of the marker-at-end flag, as it helped encourage me to mark the long ways from end to end, and not end up in the middle.

dbdean:
> > It gets a bit more frequent on short ways, as the maximum distance is 500 metres, so the actual distance can get down closer to 250-300 metres for ways just over 500 metres.
> 
> I don't understand this.

If a way is (say) 540 metres long, then in order to keep the maximum distance under 500, you would need 3 markers:

``` 
-M----------M----------M-

^--------- 540 ---------^ (full length of way)

 ^--------- 510 -------^ (too long for only two markers)

 ^--- 255 --^--- 255 --^ (distance between markers)
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: This seems a lot longer, and significantly less clear to me. The markerCount is not counting the end markers, so is wrong by two, and pointOnPolyLine is called three times when it only needs to be called once.

westnordost:
Okay then. Then this can be merged?

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
smichel17: Here's the same code, cleaned up slightly and with a bugfix. Maybe clearer now?

```kt
                val markers = mutableListOf(polyline.pointOnPolylineFromStart(MARKER_FROM_END_DISTANCE)!!)

                val midLength = (length - (2 * MARKER_FROM_END_DISTANCE))
                if (midLength > MAXIMUM_MARKER_DISTANCE) {
                    val midMarkerCount = (midLength / MAXIMUM_MARKER_DISTANCE).toInt()
                    val markerDistance = midLength / midMarkerCount
                    markers.addAll((1..midMarkerCount).map {
                        polyline.pointOnPolylineFromStart(
                            MARKER_FROM_END_DISTANCE + (it * markerDistance)
                        )!!
                    })
                }

                markers.add(polyline.pointOnPolylineFromEnd(MARKER_FROM_END_DISTANCE)!!)

                return markers.toList()

```

dbdean:
I think so.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: I don't see the point, sorry. Why is `markerCount` being calculated without including the end markers. If we were going to do that (and why?) shouldn't the variable be called `nonEndMarkerCount`?

dbdean:
Oh, I haven't worked out what is going on with the zoom-to-quest not zooming out enough. Do you have any ideas what might be going on, @westnordost?

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
smichel17: > I have decided that all long ways have markers at the end, regardless of the marker-at-end flag

This is the key. It's not immediately obvious (to me) that the math works out such that there's always markers at both the start and end. I suppose a comment to that effect would work, too :P

smichel17: > and why?

Because that's how I'd explain it to someone. "On sufficiently long ways, we put a marker at each end, then divide up the rest of the way equally so that there's never more than 500m between markers".

> shouldn't the variable be called `nonEndMarkerCount`?

Sure, that would be fine, I think it's equivalent to `middleMarkerCount`.

Anyway, here's one last try and if this doesn't get anywhere, I don't care to argue further :)

```kt
                val startMarker = MARKER_FROM_END_DISTANCE
                val endMarker = length - MARKER_FROM_END_DISTANCE

                val middleLength = length - (2 * MARKER_FROM_END_DISTANCE)
                val middleMarkers = if (middleLength > MAXIMUM_MARKER_DISTANCE) {
                    val count = (middleLength / MAXIMUM_MARKER_DISTANCE).toInt()
                    val spaceBetween = middleLength / count
                    (1..count).map { MARKER_FROM_END_DISTANCE + (it * spaceBetween) }
                } else {
                    listOf()
                }

                return mutableListOf(startMarker).addAll(middleMarkers).add(endMarker).map {
                    polyline.pointOnPolylineFromStart(it)!!
                }
```

I agree 100% it's less elegant, but I think it captures intent better.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
smichel17: I know about the quests with markers at both ends; I was wondering which quests *don't* have them at both ends, but can go longer than 500m. But now I remember it's the surface and cycleway/sidewalk quests, I think.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
```diff
@@ -8,8 +8,8 @@ import de.westnordost.streetcomplete.data.osm.mapdata.ElementType
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.data.quest.OsmQuestKey
 import de.westnordost.streetcomplete.util.measuredLength
-import de.westnordost.streetcomplete.util.pointOnPolylineFromEnd
```
FloEdelmann, line 11: This util function can also be deleted, since it's only used here.

Then, maybe even remove the `fromEnd` parameter from the private `pointOnPolyline` function there, rename it to `pointOnPolylineFromStart` and make it public?

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
FloEdelmann: I also find @smichel17's latest suggestion (without the `Marker` variable suffix) the clearest to understand.

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
FloEdelmann: I also find @smichel17's latest suggestion (the one without the `Marker` variable suffix) the clearest to understand.

smichel17: Made a minor edit (how `middleLength` is calculated) for more clarity, I think.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
smichel17, line 34:
```suggestion
            val minDistanceForAddingMultipleMarkers =
                if (osmElementQuestType.hasMarkersAtEnds) MARKER_FROM_END_DISTANCE * 4
                else MAXIMUM_MARKER_DISTANCE + 2 * MARKER_FROM_END_DISTANCE
            if (length > minDistanceForAddingMultipleMarkers) {
```
I think this might also help readability here (didn't test if this compiles/runs)

smichel17: If you take my other suggestion, you could pull the `start`, `end`, and `middleLength` calculations outside the `if` statement and simplify this a little more

```suggestion
            val minDistanceForAddingMultipleMarkers =
                if (osmElementQuestType.hasMarkersAtEnds) MARKER_FROM_END_DISTANCE * 2 else MAXIMUM_MARKER_DISTANCE
            if (middleLength > minDistanceForAddingMultipleMarkers) {
```

smichel17:
(Sorry for posting basically every intermediary step that I'd go through when refactoring the code if I were writing it :sweat_smile:)

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
FloEdelmann: @smichel17's [last suggestion here](https://github.com/streetcomplete/StreetComplete/pull/3280#discussion_r708195922) had some type errors, which I could not solve in a clean way. So I tried @smichel17's [other suggestion instead](https://github.com/streetcomplete/StreetComplete/pull/3280#discussion_r708261653) and only added the `middleLength` variable.

I think this is now a reasonably well-readable and concise version:

```kt
    override val markerLocations: Collection<LatLon> get() {
        if (geometry is ElementPolylinesGeometry) {
            val polyline = geometry.polylines[0]
            val middleLength = polyline.measuredLength() - (2 * MARKER_FROM_END_DISTANCE)
            val minDistanceForAddingMultipleMarkers =
                if (osmElementQuestType.hasMarkersAtEnds) MARKER_FROM_END_DISTANCE * 2
                else MAXIMUM_MARKER_DISTANCE
            if (middleLength > minDistanceForAddingMultipleMarkers) {
                val markerCount = 2 + (middleLength / MAXIMUM_MARKER_DISTANCE).toInt()
                val markerDistance = middleLength / (markerCount - 1)
                return (0 until markerCount).map {
                    polyline.pointOnPolylineFromStart(
                        MARKER_FROM_END_DISTANCE + (it * markerDistance)
                    )!!
                }
            }
        }
        return listOf(position)
    }
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
westnordost: > This util function can also be deleted in SphericalEarthMath.kt, since it was only used here.

Rather not, the util functions in SphericalEarthMath are ... well, you never know when you might need them. They are quite generic.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: I have adjusted the code to be similar to @FloEdelmann's last suggestion.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuest.kt
dbdean: I have done something similar to this now.

smichel17: Looks good to me.

smichel17 (approved):

westnordost:
Thank you for taking into account the comments of the other contributors. I didn't read along, so this is now ready for merge/review?

westnordost (approved):

PR #3296, opened 2021-09-18
Convert some LocationUtil functions to extensions.

Isira-Seneviratne:


app/src/main/java/de/westnordost/streetcomplete/location/LocationUtil.kt
```diff
@@ -8,20 +8,17 @@ import android.location.Location
...
 
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 
-fun isLocationEnabled(context: Context): Boolean =
-    hasLocationPermission(context) && isLocationEnabled(getLocationManager(context)!!)
+val Context.isLocationEnabled: Boolean
+    get() = hasLocationPermission && LocationManagerCompat.isLocationEnabled(getSystemService()!!)
 
-private fun getLocationManager(context: Context): LocationManager? =
-    ContextCompat.getSystemService(context, LocationManager::class.java)
```
smichel17, line 19: What api level was [`Context.getSystemService`](https://developer.android.com/reference/kotlin/androidx/core/content/package-summary#getsystemservice) introduced in? It *seems* like an extension function which calls ContextCompat.getSystemService under the hood, but there's nothing in the docs about api levels (The android javadocs used to be some of my favorite, but they've really gone downhill in usability, plus for some reason they've stopped showing up on DuckDuckGo :rage:) Also, how does it know which service?

westnordost: > What api level was Context.getSystemService introduced in?

That's a Kotlin extension function, see `import androidx.core.content.getSystemService`

westnordost (approved):
Looks good, thanks!

PR #3298, opened 2021-09-19
Fix typo in enGB strings.xml

tech189:
Hi,

Just noticed this typo while using StreetComplete:

"It _was_ multiple purposes" which should be "It _has_ multiple purposes" just like "It has multiple house numbers".

This patch should fix it, but I'm a bit new to pull requests - hopefully this works.


FloEdelmann:
Thanks for the PR!

Please note that translations should be updated via POEditor, see the [contributing guidelines](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app):

> Before each release, translations are pulled in from POEditor. Please, use POEditor for translating. Manual changes submitted as Pull Requests will not be merged as they do not help the project.

I just made this change there, so this PR can be closed.

tech189:
Ah sorry, thanks for fixing it though!

matkoniecz:
And thanks for reporting it!

Issue #3294, opened 2021-09-18
Tactile paving spam


smichel17:
Recently (since some time this year, I don't remember exactly when) I've been seeing a lot more tactile paving and curb quests. I'm not sure if this is because of a change to the app, someone in my area doing more satellite mapping, or both. In some cases, though, this quest is asked far too much.

In a (semi-)rural area, there are never going to be tactile pavings where a driveway crosses a sidewalk. Okay, *maybe* if we're talking about a house in/near the center of town, but even then I'd be surprised. Here's one example: https://www.openstreetmap.org/note/2857706. Consider this picture attached to the note (and see #3293 for why it isn't):

![35b75a25-fb7f-4d8a-b295-d042c45e8911](https://user-images.githubusercontent.com/6563664/133868172-7463a099-9a4e-4377-9a0f-8fb47a7ae426.jpg)

This one happens to be mis-mapped: it's a walking path, not a driveway. I just didn't realize that when I was mapping. But if you open this location in SC, you can see a row of quests like it, on correctly-mapped driveways.

**Proposed Solution**
At the very least, we should not ask when a walking path crosses a driveway which is off of a `highway=residential`. Or perhaps we can widen that to exclude any path crossing a driveway (this would simplify the filter).

matkoniecz:
Do you think that excluding crossings on `service=driveway` would be enough?

And yes, it appears that someone mapped driveways, with crossings - see https://www.openstreetmap.org/node/8830730760/history

westnordost:
I think iD suggests adding highway=crossing also for driveways. This,  this tagging is found in the wild.

Am 18. September 2021 08:30:49 MESZ schrieb Mateusz Konieczny ***@***.***>:
>Do you think that excluding crossings on `service=driveway` would be
>enough?
>
>And yes, it appears that someone mapped driveways, with crossings - see
>https://www.openstreetmap.org/node/8830730760/history
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/streetcomplete/StreetComplete/issues/3294#issuecomment-922213436

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

Helium314:
Yes, iD automatically adds crossings when connecting roads and paths via the "connect the features" solution of detected issues

smichel17:
[...]

PR #3301, opened 2021-09-19
Exclude driveways from tactile pavings quest

smichel17:
Done from GitHub UI, untested and I'm not certain if this is even the right thing to do (it's my first time writing a query, mostly written out of curiosity about what excluding driveways would look like). Also, as discussed in #3294, we might not even fix it this way.


westnordost:
I think that makes sense

westnordost (approved):

PR #3311, opened 2021-09-21
Add picnic table cover quest

### LINKED ISSUE #3142: New quest: Is this picnic table covered?
### General
Rationale: When hiking (or just walking around the city/countryside) it is nice to know whether a picnic table is covered, for example on very sunny days or when it's raining.

Affected tag(s) to be modified/added: [Key:covered](https://wiki.openstreetmap.org/wiki/Key:covered)
Question asked: **Is this picnic table covered?**

### Checklist

### Ideas for implementation

**Element selection:** 
All nodes tagged with [leisure=picnic_table](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpicnic_table)

**Proposed UI:**
Simple yes/no UI, like with "Is this bicycle parking covered"?

FloEdelmann:
Closes #3142.


westnordost:
Also add it to the "outdoor" achievement?

(Maybe the achievement stuff should be refactored so that it is defined in the quests themselves to which achievements they should count. Contributors keep forgetting to add it)

westnordost (approved):
Looks good! Thank you for the contribution!

Is the question wording the same as for the bus stop?

By the way, it is mostly just an OCD, but sometimes I also clean the SVG / XML with an editor, e.g. when I see that stroke properties are defined but a path has no stroke, or fill properties are defined but that path doesn't fill

app/src/main/res/values/strings.xml
```diff
@@ -1080,4 +1080,5 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
     <string name="quest_presets_default_name">Default</string>
     <string name="quest_presets_selected">Selected</string>
     <string name="quest_presets_delete_message">"Delete preset \"%s\"? The quest selection and order for it will be irrevocably lost."</string>
+    <string name="quest_picnicTableCover_title">Is this picnic table protected from the rain?</string>
```
FloEdelmann, line 1083: > Is the question wording the same as for the bus stop?

No. There it's "Is there a shelter at this bus stop?", but it adds `shelter=yes`, not `covered=yes`. For bike/motorcycle parking it's "Is this bicycle/motorcycle parking covered (protected from rain)?", so let's update the picnic table wording accordingly:


```suggestion
    <string name="quest_picnicTableCover_title">Is this picnic table covered (protected from rain)?</string>
```

FloEdelmann (commented):
> Also add it to the "outdoor" achievement?

Will do.

> (Maybe the achievement stuff should be refactored so that it is defined in the quests themselves to which achievements they should count. Contributors keep forgetting to add it)

Maybe I'll try to refactor this afterwards.

> By the way, it is mostly just an OCD, but sometimes I also clean the SVG / XML with an editor, e.g. when I see that stroke properties are defined but a path has no stroke, or fill properties are defined but that path doesn't fill.

I'll clean them up manually.

westnordost (approved):

PR #3320, opened 2021-09-25
Add building type bridge (skyway)

### LINKED ISSUE #2630: Add skybridges/skyways as an option to the building quest
**Use case**

Enable mapping of such structures
(I mean those enclosed skyways which have a roof and walls like 
http://correctionalnews.com/wp-content/uploads/2018/09/SonomaCounty-800x445.jpg
Or
https://placesjournal.org/wp-content/uploads/2016/05/yoos-james-multilevel-metropolis-15.jpg
**Proposed Solution**

Ideally add a tag for those kinds of structures

Sadly it's a bit unclear how to tag them but I found this
https://wiki.openstreetmap.org/wiki/Item:Q16636

kmpoppe:
Adds new building type `bridge` to the building quest, solves the already closed #2630


matkoniecz:
Note

> I'd say that this building type (if you can say that it is a building) is far too rare to be in that list. Better leave a note in that case.

https://github.com/streetcomplete/StreetComplete/issues/2630#issuecomment-789804431

kmpoppe:
I have noted that comment. I just don't concurr that it's too rare to be part of StreetComplete, especially in urban areas. Additionally, [taginfo](https://taginfo.openstreetmap.org/tags/?key=building&value=bridge#chronology) shows a constant growth in use, so it's not an abandoned way to tag skyways as of right now.

westnordost (commented):
Maybe add "bridge" to commercial category too?

westnordost:
Hmm, thinking about this ... does a skyway actually count as an own building? I'd think that it is just a building part and the whole building would be the part on one side of the street and the part on the other side of the street?

westnordost:
NEvermind, it is documented that way in the wiki.

westnordost (approved):

Issue #3312, opened 2021-09-22
Leaving custom notes location still have random location error sometimes (#2)


mnalis:
Original problem was reported and was supposed to be fixed (thanks @smichel17!) in: https://github.com/streetcomplete/StreetComplete/issues/3284

However, I still (sometimes) experience a problem. 

**How to Reproduce**
This time it seems (at least sometimes) to happen even when "follow gps" is disabled? Maybe related to adding pictures to note?

Example, all three of those notes are supposed to be at the same location (crossing of path and street):
- https://www.openstreetmap.org/note/2861754 (ok)
- https://www.openstreetmap.org/note/2861752 (ok)
- https://www.openstreetmap.org/note/2861753 (bad)

While first two are at the correct location, third one is off.

Another example; all three of those notes were marked at the path crossing where the third one is, but they differ if GPS follow was on, and at what moment was picture added. Only the first one is positioned correctly. 
- https://www.openstreetmap.org/note/2861756 (ok)
- https://www.openstreetmap.org/note/2861757 (bad)
- https://www.openstreetmap.org/note/2861758 (bad)

And yet, sometimes everything looks like it's working perfectly (for example, all of those notes correctly ended up at the same location: [2863118](https://www.openstreetmap.org/note/2863118), [2863119](https://www.openstreetmap.org/note/2863119), [2863120](https://www.openstreetmap.org/note/2863120), [2863113](https://www.openstreetmap.org/note/2863113), [2863116](https://www.openstreetmap.org/note/2863116), [2863117](https://www.openstreetmap.org/note/2863117) - even when I tried different things while taking them)

**Versions affected**
 Android 6.0.1 (MIUI 10.2), SC 34.2

westnordost:
So, you can never reproduce it when you create a note, tap OK. Only if there is something in between, like taking a photo, leaving the form open etc?

westnordost:
I've prepared a special version of StreetComplete that puts all kind of debug data (regardin position) into each created note. Would you try it with that version? Maybe that will shed some light:

https://www.westnordost.de/misc/StreetComplete-v35.0-beta1-create-note-debug.apk

mnalis:
Sure. Does that debug APK contains 


 Do you have that debug code as source patch (or public git commit id) somewhere? I'd like to pull it into my fork, as it seems to be happening much more rarely since #3284. 
I found it 3 times

mnalis:
[...]

PR #3321, opened 2021-09-26
Move freezing into map fragment

smichel17:
Exploration for #3312. This might help prevent more bugs of this sort.

I did this on GitHub and haven't tested whether it compiles. There is certainly more to do (e.g., moving `resetFreezeMap` into the mapfragment and removing the now-redundant `[un]freezeMap` functions). Just opening to check what you think of this approach.


westnordost:
In general: Ok. But the change doesn't work as is. QuestsMapFragment is a subclass of LocationAwareMapFragment.

smichel17:
Ok. I will close this and do it properly at some (undetermined) point in the future.

I also have some vague desire to refactor this whole chain-of-subclasses thing. It's awkward to have three different subclass when, as far as I am aware, only the QuestsMapFragment is actually used anywhere. I.e. the only purpose of the subclasses seems to be code organization. Perhaps it could be just one MapFrahment, with a QuestController and LocationController, or something like that. Or, since it's not actually *that* many lines of code, pulling out a bunch of helper functions might keep the core logic understandable enough.

PR #3325, opened 2021-09-27
Disable note OK button until layout is finished

### LINKED ISSUE #3312: Leaving custom notes location still have random location error sometimes (#2)

smichel17:
Maybe fixes #3312

Done from GitHub web UI, uncompiled & untested. I'll be back to my normal setup tomorrow and can look into this further at that point.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
@@ -37,9 +37,15 @@ abstract class AbstractCreateNoteFragment : AbstractBottomSheetFragment() {
         }
 
         noteInput.addTextChangedListener(TextChangedWatcher { updateOkButtonEnablement() })
+        okButton.isEnabled = false
         okButton.setOnClickListener { onClickOk() }
 
         updateOkButtonEnablement()
+    }
+    
+    override fun onViewStateRestored(savedInstanceState: Bundle?) {
+        super.onViewStateRestored(onViewStateRestored)
+        okButton.isEnabled = true
     }
 
     private fun onClickOk() {
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
@@ -44,7 +44,7 @@ abstract class AbstractCreateNoteFragment : AbstractBottomSheetFragment() {
     }
     
     override fun onViewStateRestored(savedInstanceState: Bundle?) {
-        super.onViewStateRestored(onViewStateRestored)
+        super.onViewStateRestored(savedInstanceState)
         okButton.isEnabled = true
     }
 


app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
```diff
@@ -37,10 +37,16 @@ abstract class AbstractCreateNoteFragment : AbstractBottomSheetFragment() {
         }
 
         noteInput.addTextChangedListener(TextChangedWatcher { updateOkButtonEnablement() })
+        okButton.isEnabled = false
         okButton.setOnClickListener { onClickOk() }
 
         updateOkButtonEnablement()
     }
+    
+    override fun onViewStateRestored(savedInstanceState: Bundle?) {
```
smichel17, line 46: I'm not totally sure that this is the right callback to hook into (whether layout is finished by this point); if it doesn't work we can switch to onStart or onResume instead, or I'll hunt more to see if there's a better callback.

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractCreateNoteFragment.kt
```diff
@@ -37,10 +37,16 @@ abstract class AbstractCreateNoteFragment : AbstractBottomSheetFragment() {
...
+        super.onViewStateRestored(onViewStateRestored)
```
smichel17, line 47:
```suggestion
        super.onViewStateRestored(savedInstanceState)
```

smichel17:
Based on further discovery in #3312, this is not the right approach.

PR #3328, opened 2021-09-28
make text more readable - set style on image descriptions used in some quests [ready]

matkoniecz:
changes ImageSelectTitle ImageSelectDescription - that a bit confusingly affects quest where user selects icon wth title + description - and photo-selecion quests are not affected

this starts explicitly setting colour on text visible during selection
why? because in some cases light gray currently used is not clearly visible and this allows to set it at least in some cases

TODO:
- check is it fine to set such colour manually or is it bad idea for some reason (that is why I will open a draft PR)
- fix it for night mode where this change degraded contrast
- before fixing for dark mode check where else hint_text is used and is it also affected by poor contrast in night mode right now

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -27,7 +27,7 @@
     <color name="indicator_dot_selected">@color/accent</color>
 
     <color name="hint_text">#666</color>
-    <color name="title_text_below_image">#000</color>
+    <color name="label_text_below_image">#000</color>
     <color name="title_text_next_to_image">#000</color>
     <color name="description_text_next_to_image">#000</color>
 
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -26,7 +26,7 @@
     <color name="indicator_dot">#888</color>
     <color name="indicator_dot_selected">@color/accent</color>
 
-    <color name="hint_text">#555</color>
+    <color name="hint_text">#666</color>
     <color name="title_text_next_to_image">#000</color>
     <color name="description_text_next_to_image">#000</color>
 
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -26,7 +26,7 @@
     <color name="indicator_dot">#888</color>
     <color name="indicator_dot_selected">@color/accent</color>
 
-    <color name="hint_text">#666</color>
+    <color name="hint_text">#555</color>
 
     <color name="disabled_text">#6666</color>
 
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -27,6 +27,8 @@
     <color name="indicator_dot_selected">@color/accent</color>
 
     <color name="hint_text">#555</color>
+    <color name="icon_selection_title_text">#000</color>
+    <color name="icon_selection_description_text">#000</color>
 
     <color name="disabled_text">#6666</color>
 
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -27,8 +27,8 @@
     <color name="indicator_dot_selected">@color/accent</color>
 
     <color name="hint_text">#555</color>
-    <color name="icon_selection_title_text">#000</color>
-    <color name="icon_selection_description_text">#000</color>
+    <color name="title_text_next_to_image">#000</color>
+    <color name="description_text_next_to_image">#000</color>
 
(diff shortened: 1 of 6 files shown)


matkoniecz:
to clarify what I meant by "in some cases" - I had situations where either I or someone else had trouble to read/notice smaller gray text

And I see no good reason to keep it so light gray, I think it would be fine a bit darker what would make it more readable

FloEdelmann:
Could you share screenshots please?

matkoniecz:
Should show what I wanted to achieve - a bit darker text.

Note that:
- I plan to test with darker text (what would affect text currently set using this value), this is not a final effect
- in such screen light gray is 100% readable - this does not necessarily apply with smaller screen and in less optimal conditions (or with people that have a worse eyesight than average)
 
https://user-images.githubusercontent.com/899988/135102643-90462e52-e2d5-48fd-92e9-22583ab3c2f2.png
https://user-images.githubusercontent.com/899988/135102650-f013333d-0a9b-42cb-b9da-0be5d06ffdc8.png

westnordost:
The description is not a "hint". Define an own color (which may be the same as current @hint_color) for these.

matkoniecz:
> The title+description could as well also be black.

I actually would prefer this, but thought that not-black is intended. So I am going to use black (in normal mode)

> Define an own color (which may be the same as current @hint_color) for these.

Maybe in such case just applying already defined text color would be fine?

`    <color name="text">#000</color>`

westnordost:
> <color name="text">#000</color>

using @color/text would be fine, but it would be better if there'd be a title text color and a description text color.

matkoniecz:
> it would be better if there'd be a title text color and a description text color

OK, I will change it.

In meantime, here is preview of more radical version (black/white, hint_color also increased contrast). I like effect but I guess that it may be too much for some?

https://user-images.githubusercontent.com/899988/135113981-3473c523-a3a0-4a3c-99aa-fad944de9fe1.png
https://user-images.githubusercontent.com/899988/135113984-786630c1-573b-4b54-b5af-1b152fa4ec83.png

matkoniecz:
I set `icon_selection_description_text` and `icon_selection_title_text` as `title_text` would be quite misleading (title alone would be applying rather to quest titles).

app/src/main/res/values/colors.xml
```diff
@@ -26,7 +26,9 @@
     <color name="indicator_dot">#888</color>
     <color name="indicator_dot_selected">@color/accent</color>
 
-    <color name="hint_text">#666</color>
+    <color name="hint_text">#555</color>
+    <color name="icon_selection_title_text">#000</color>
+    <color name="icon_selection_description_text">#000</color>
```
FloEdelmann, line 31: The styles are called `ImageSelectTitle` and `ImageSelectDescription`, so I'd suggest using

```suggestion
    <color name="image_select_title_text">#000</color>
    <color name="image_select_description_text">#000</color>
```

smichel17:
> here is preview of more radical version (black/white, hint_color also increased contrast). I like effect but I guess that it may be too much for some?

Given that the app is supposed to be used outside, sometimes in direct sunlight, I am all in favor of dialing up the contrast.

app/src/main/res/values/colors.xml
matkoniecz: I even thought about renaming `ImageSelectTitle` `ImageSelectDescription` styles. It seems that this is used only with svg illustrations.

image_select_description_text seems to be about
screen
that is using a separate style (`@style/ImageSelectLabel`)

Though, in the end bridge quest now also have svg illustrations, and `ImageSelectTitle` `ImageSelectDescription` could be used with tiny photos (in theory at least)

------

`title_text_next_to_image` `TitleNextToImage` + `description_text_next_to_image` `TitleNextToImage` ?

FloEdelmann: > `title_text_next_to_image` `TitleNextToImage` + `description_text_next_to_image` `DescriptionNextToImage` ?

That sounds very clear for me, so :+1:

matkoniecz: Done, I will keep unresolved in case @westnordost is not a fan of such change.

matkoniecz:
I think that all three new variables can be combined into something like `answer_selection_text` but three different ones are also not problematic I think.

westnordost:
This is still marked as draft. Is it not finished?

matkoniecz:
I want to check it once more to ensure that I am not missing something (for example I got really confused where `hint_text` is actually being applied - this one is fixed)

westnordost (approved):

PR #3334, opened 2021-10-01
Allow converting ways into steps

### LINKED ISSUE #3288: Allow people to convert paths to steps as an other answer
**Use case**
I agree it's quite common when someone is mapping footways from imagery, they usually can't tell if there are steps or not, so this is well suited for someone doing StreetComplete to pickup on. And knowing if there are steps or not is incredibly important for pedestrian routing.
_Originally posted by @andrewharvey in https://github.com/streetcomplete/StreetComplete/discussions/3088#discussioncomment-1040091_

Isn't the main benefit that you can then go on to answer the remaining step quests about the new steps immediately and accurately (and without having to write more notes)?

In the same way as you can go e.g. diet related quest->shop isn't here anymore->different type of food place, name of shop, diet related quest on new shop etc
_Originally posted by @peternewman in https://github.com/streetcomplete/StreetComplete/discussions/3088#discussioncomment-1044090_

I've now personally seen steps just tagged as a path both in the countryside and in urban environments. Yes, you can leave a note, but either you need to try and populate which way up (which is particularly hard to describe), handrail, surface, count etc (i.e. reinventing the wheel and storing that information in a note). Or you need to wait for it to be re-tagged and then revisit it.

**Proposed Solution**
If I can split the way in streetcomplete already, would it be an issue to have the alternative answer "actually steps here" just go ahead and retag the footway as steps?

_Originally posted by @andrewharvey in https://github.com/streetcomplete/StreetComplete/discussions/3088#discussioncomment-1040349_

i.e. add an other answer to a (foot) way that it is actually steps
<hr>
There is a smaller, but possibly still valid argument for being able to do the reverse if you want to micro-map landings on long staircases you could split it and convert them back to paths. This might be more complicated, although IIRC the splitting steps thing would deal with removing most of any existing tagged info.

FloEdelmann:
Closes #3288.


app/src/main/res/values/strings.xml
```diff
@@ -793,6 +793,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_split_way_tutorial">Tap on the way to split it at the point(s) where the answer differs. Try to be as precise as possible, you can zoom in as usual.</string>
     <string name="quest_split_way_too_imprecise">"Please zoom in further"</string>
     <string name="quest_split_way_many_splits_confirmation_description">These are quite a few splits. You can always further split up the way later.</string>
+    <string name="quest_generic_answer_is_actually_steps">"There are steps here"</string>
```
FloEdelmann, line 796: @peternewman @smichel17 @andrewharvey Any advice on concise wording here?

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
```diff
@@ -324,6 +343,10 @@ abstract class AbstractQuestAnswerFragment<T> :
         }
     }
 
+    private fun onClickConvertToStepsAnswer() {
+        listener?.onConvertToSteps(questKey as OsmQuestKey)
```
FloEdelmann, line 347: Should there be a warning first along the lines of "Please split the way accordingly before converting the whole way into steps."

westnordost:
To be honest, since this is just a re-tag from way to steps, I think it should just be an other-answer like other other answers for some quests, with no special treatment. Splitting a way and replacing a shop comes with a lot of logic attached + extra UI, this one, not.

app/src/main/res/values/strings.xml
matkoniecz: Preferably one encoraging to split first - rather than turning into steps footway which is steps only in a small part.

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
westnordost: It's not really done for other quests, is it? Hmm. @matkoniecz suggested that. Is this based on any user tests / known issues? 

Or rephrase other answer option to "Whole way are steps"?

smichel17:
> since this is just a re-tag from way to steps

If we're going to take @matkoniecz's suggestion and prompt the user to split the way first, does it make sense to ask them whether the split is at the top or bottom while we're at it, and also tag that info?

westnordost:
No. There is a quest for that.

app/src/main/res/values/strings.xml
FloEdelmann: Is that better?

```suggestion
    <string name="quest_generic_answer_is_actually_steps">"It’s a staircase or step"</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
matkoniecz: > Whole way are steps

+1, though is it maybe rather

> Whole way is steps

(way is singular here, not plural)

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
```diff
@@ -222,6 +222,18 @@ abstract class AbstractQuestAnswerFragment<T> :
...
+
+        if (way.isArea()) return null
+
+        if (!way.tags.containsKey("highway") || way.tags["highway"] == "steps") return null
+
+        return AnswerItem(R.string.quest_generic_answer_is_actually_steps) {
+            applyAnswer(answer)
+        }
+    }
+
```
westnordost, line 236: No, I meant not like that, but like any other other answer. Per quest. The only thing that is shared is the same string.

There shoiuld be no special provision in AbstractQuestAnswerFragment.

app/src/main/java/de/westnordost/streetcomplete/quests/way_lit/WayLit.kt
```diff
@@ -5,4 +5,5 @@ enum class WayLit(val osmValue: String) {
     AUTOMATIC("automatic"),
     YES("yes"),
     NO("no"),
+    IS_ACTUALLY_STEPS(""),
```
westnordost, line 8: This looks like a hack. Better use sealed classes?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -48,6 +48,7 @@ class AddCyclewayPartSurface : OsmFilterQuestType<SurfaceAnswer>() {
                 changes.updateWithCheckDate("cycleway:surface", answer.value.osmValue)
                 changes.addOrModify("cycleway:surface:note", answer.note)
             }
+            IsActuallyStepsSurfaceAnswer -> throw IllegalStateException()
```
westnordost, line 51: Sealed classes exist to avoid these kinds of things

```kotlin
sealed class SurfaceOrIsStepsAnswer
object IsActuallyStepsAnswer : SurfaceOrIsStepsAnswer()

sealed class SurfaceAnswer : SurfaceOrIsStepsAnswer()
data class GenericSurfaceAnswer(val value: Surface, val note: String) : SurfaceAnswer()
data class SpecificSurfaceAnswer(val value: Surface) : SurfaceAnswer()

```

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
FloEdelmann: Done in [`5b090e1` (#3334)](https://github.com/streetcomplete/StreetComplete/pull/3334/commits/5b090e1a78335a153a58e7f969185e972953978b)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
FloEdelmann: Fixed in [`6715580` (#3334)](https://github.com/streetcomplete/StreetComplete/pull/3334/commits/6715580eb934a694c922d2e24c2a0d952f0c2343)

app/src/main/java/de/westnordost/streetcomplete/quests/way_lit/WayLit.kt
FloEdelmann: Done in [`65b2244` (#3334)](https://github.com/streetcomplete/StreetComplete/pull/3334/commits/65b2244c4efb11484c8f69de2adb76f903467d66)

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
FloEdelmann: What about my suggestion here?
https://github.com/streetcomplete/StreetComplete/pull/3334#discussion_r720684849

app/src/main/res/values/strings.xml
westnordost: IMO not, because I think staircase is most of all a "Treppenhaus" and step is "eine Stufe".

How about
- "It's actually steps"
- "This part are steps"

app/src/main/java/de/westnordost/streetcomplete/quests/way_lit/WayLitAnswer.kt
```diff
@@ -0,0 +1,6 @@
+package de.westnordost.streetcomplete.quests.way_lit
+
+sealed class WayLitOrIsStepsAnswer
+object IsActuallyStepsAnswer : WayLitOrIsStepsAnswer()
+
+data class WayLitAnswer(val value: WayLit) : WayLitOrIsStepsAnswer()
```
westnordost, line 6: Yeah, it's a shame that an enum cannot be a subclass of a sealed class. However, since Kotlin 1.5, there are also `sealed interface`s. So it should be possible to write

```kotlin
sealed interface WayLitAnswer
object IsActuallyStepsAnswer : WayLitAnswer
enum class WayLit {
   //... etc
}  : WayLitAnswer
```

FloEdelmann: Nice, that avoids much of the overhead that the sealed class produced :)

Done in [`3493bd6` (#3334)](https://github.com/streetcomplete/StreetComplete/pull/3334/commits/3493bd6458d21bb64f0bf111a88c73f34e732666)

app/src/main/res/values/strings.xml
westnordost: Or "these are steps"

German translation would be "Das ist eine Treppe"

FloEdelmann: Hmm, I don't really like the mixture of singular and plural.

What about "Consists only of steps"?

FloEdelmann: I changed it to "These are steps" in [`2a7212a` (#3334)](https://github.com/streetcomplete/StreetComplete/pull/3334/commits/2a7212af4f1e31d91a6219c8ef9df16102aefbc2).

westnordost (approved):

Issue #2191, opened 2020-10-23
Improve camera following position


smichel17:
### Use case

I usually want the position of the map to follow my movement. However, I usually do not want the map to be centered on my position, because I am mapping exclusively things in front of me — I would prefer to have my position fixed at about here:

<img height="400" alt="3/4 of the way from the top of the screen, horizontally centered" src="https://user-images.githubusercontent.com/6563664/97014210-90b48180-1539-11eb-95d8-12c5c5fbbfb1.png">

In theory, this is what the compass mode is for. However, in practice, the direction of the compass changes far too erratically, making the map hard to read — there is some stabilization already that prevents the map from spinning around completely, but it still creates jitter. Also, your position is still in the center of the screen instead of closer to the bottom. So instead, I just keep the map unlocked and pan around as I walk down the street. (I am actually curious whether anybody finds compass mode useful.)

### Proposed Solution

I do not have a particular solution I like enough to propose it. However, I have a couple partial ideas; maybe with a few more brains we can turn them into a solution.

- Modify compass mode to add additional stabilization (eg, slower turning)
    - If needed, add a control for the user to quickly rotate the view 90°
- Some way to unlock the camera from the *exact* center, but still move the map with your position.
    - One option, borrowed from some games have this and call it a "semi-locked" camera: allow panning, but only a little bit, so that your position always stays within a certain area (eg, the black rectangle):
        <img height="400" alt="3/4 of the way from the top of the screen, horizontally centered" src="https://user-images.githubusercontent.com/6563664/97019350-be9cc480-153f-11eb-9a75-e198da61408d.png">
       Of course, right now, panning the map stops following your position, so there would need to be a way to toggle this. Maybe an additional state of the gps button, instead of making it disappear while following.
    - Another option: The gps button does not disappear when following location. Pressing it allows you to pan the view freely, but it still moves when you move (keeps the same position relative to your location). And the icon could change to either something like locked/unlocked, linked/unlinked, or arrows in/out (like this: ![image](https://user-images.githubusercontent.com/6563664/97047043-de47e300-1567-11eb-9f19-1bc6c3c99220.png))
- A mode that locks the camera in the position shown above (horizontally centered, 1/4 from bottom of screen), and changes the effect of dragging with one finger — horizontal movement would rotate, and vertical would tilt.

RubenKelevra:
Hey, this sounds pretty great.

[...]

PR #3335, opened 2021-10-01
Navigation mode

### LINKED ISSUE #1040: Map orientation: to direction of movement
I tested the compass mode in StreetComplete ... does not work for me at all! Probably because my smartphone case has magnets to close/open it.

As I am using OsmAnd+ all the time for navigation, I know (and love) their "Map orientation: to direction of movement", obviously not influenced by nearby magnets. I believe it works by using GPS positions (with Kalman filter for smoothing).

With that one always sees what's ahead.

Please add this feature, it would help a lot e.g answering the cycleway quests. And it would be nice for users without compass.

### LINKED ISSUE #2209: Add GPS trail for paths walked
**Use case**

When mapping cities, especially where there are lots of small streets/ways interconnecting, there are often some confusion where exactly is quest located in relation to user. Rotating the map, and using compass viewfield helps, but is not perfect (especially since compass/orientation sensor is not perfect in mobile phones).

**Proposed Solution**

What would help is (preference-enabled?) GPX-track drawn on the screen, so user can see where s/he's been prior to reaching current location.
(Something similar to OsmAnd "record GPX" mode), or this mockup:
Screenshot_20201027_171847

Also, it would show user what streets have been traveled already and letting him/her decide where to go next. The GPX track could be removed if some amount of time had passed since it was last updated (maybe 6 hours? or 24h?), or could be left to user to clear it (for example, by disabling GPX option, and reenabling it later if s/he still wants it) 

Alternatively, instead of full GPX track, only a last (few) hundred meters could be shown as disappearing trail ([spacewar!-alike](https://en.wikipedia.org/wiki/File:Spacewar_gameplay.jpg)). While a little less useful, it would help with immediate orienting, and would avoid the issue of deciding when to remove displayed GPX track. Something like this mockup:
Screenshot_20201027_171725_de westnordost streetcomplete2

(I'm obviously not artist, the idea is the line is thickest and most opaque at the current location, and thins out and/or gets more transparent the father away it gets, and at far enough point it simply disappears)

smichel17:
Fixes #1040, Fixes #2209

Opening this so I can comment, since apparently GitHub won't let you do inline comments on anything except PRs.


app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
```diff
@@ -23,37 +23,44 @@ class LocationStateButton @JvmOverloads constructor(
...
+
     private val tint: ColorStateList?
 
+    var isNavigation: Boolean = false
+    set(value) {
+        if (field != value) {
+            field = value
+            refreshDrawableState()
+        }
+    }
```
smichel17, line 46: `import kotlin.properties.Delegates`
```suggestion
    var isNavigation: Boolean by Delegates.observable(false) {
        refreshDrawableState()
    }
```

westnordost:
and #2191

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
westnordost: Hmm, this is maybe nice(?). Can you link the documentation?

Swift has this built-in. I always missed this in Kotlin
```swift
var isNavigation: Bool = false {
    didSet {
        refreshDrawableState()
    }
}
```

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
```diff
@@ -106,83 +131,145 @@ open class LocationAwareMapFragment : MapFragment() {
...
         isFollowingPosition = prefs.getBoolean(PREF_FOLLOWING, true)
-        isCompassMode = prefs.getBoolean(PREF_COMPASS_MODE, false)
+        isNavigationMode = prefs.getBoolean(PREF_NAVIGATION_MODE, false)
     }
 
     private fun saveMapState() {
         activity?.getPreferences(Activity.MODE_PRIVATE)?.edit {
             putBoolean(PREF_FOLLOWING, isFollowingPosition)
-            putBoolean(PREF_COMPASS_MODE, isCompassMode)
+            putBoolean(PREF_NAVIGATION_MODE, isNavigationMode)
```
smichel17, line 236: Maybe an int (or string) here, so if there is ever a 3rd mode, you don't need to burn the pref key? Using an enum, it should be ~1 additional line of code.

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
smichel17: I edited in the [link](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.properties/-delegates/observable.html) just after posting. You can also check the documentation for [delegated properties](https://kotlinlang.org/docs/delegated-properties.html) in general, though note that it took me quite a while to wrap my head around the custom ones. We're already using the other standard delegate, `lazy`.

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
```diff
@@ -11,7 +11,7 @@ import android.view.View
 import androidx.annotation.Keep
 import androidx.appcompat.widget.AppCompatImageButton
 import de.westnordost.streetcomplete.R
-import java.util.Arrays
+import de.westnordost.streetcomplete.location.LocationState.*
```
smichel17, line 14: `.*` :cry:

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
smichel17: (e.g. perhaps one of the ideas from https://github.com/streetcomplete/StreetComplete/issues/2191)

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
westnordost: Hmm, good to know. I am ambivalent whether this is more readable than the current version. The clear disadvantage of this is that it can (unlike the swift-variant) only be used if there is no other logic for the variable.

So if you like, you can create a PR for that, but then replace all occurances, not just here.

I do wonder if with the `by` delegates syntax, something similar to swift could be achieved. I.e. maybe

```kotlin
var isNavigation: Boolean by swiftlike(
    set = { value ->
        println("set it!")
        return value
    },
    didSet = {
        refreshDrawableState()
    }
)
```

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
westnordost: OCD?

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
smichel17: Along these lines
```diff
+ enum class NAVIGATION_MODE(val isEnabled: Boolean) { TOP_DOWN(false), ANGLED(true) }
- private var _isNavigationMode: Boolean = false
+ private var navigationMode: NAVIGATION_MODE = TOP_DOWN
  var isNavigationMode: Boolean
      set(value) {
-         if (_isNavigationMode != value && !value) {
+         if (navigationMode.isEnabled != value && !value) {
              updateCameraPosition(300) { tilt = 0f }
          }
-         _isNavigationMode = value
+         navigationMode = if (value) NAVIGATION_MODE.TOP_DOWN else NAVIGATION_MODE.ANGLED
      }
-     get() = _isNavigationMode
+     get() = navigationMode == NAVIGATION_MODE.ANGLED
```
```suggestion
            putInt(PREF_NAVIGATION_MODE, navigationMode.ordinal)
```

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
smichel17: You have access to it in the lambda parameters; I just omitted them here. …which I just checked and apparently is only valid syntax for single-parameter lambdas. Also just tested and I was mistaken about when `observable` is called; it's on every `set`, not just when the value changes. So working code, identical to what it replaces, is:

```suggestion
    var isNavigation: Boolean by Delegates.observable(false) { prop, old, new ->
        if (old != new) refreshDrawableState()
    }
```

If you wanted to split into two callbacks, one which is called unconditionally and one called only when the value changes, you could definitely do that with a custom delegate. I implemented something like that for wrapping Android preferences, at https://gitlab.com/smichel17/android-utils. [Direct source link](https://gitlab.com/smichel17/android-utils/-/blob/master/preferences/src/main/java/me/smichel/android/KPreferences/Preferences.kt#L70-84). Please don't use that library, though; I'm not totally happy with the api and might re-do it some day.

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
smichel17: Context: https://github.com/streetcomplete/StreetComplete/pull/3327#discussion_r717515209

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
westnordost: Yes, I know what you meant. I don't see the sense in implementing something that isn't used though.

What I changed right now is the key name of the pref to the old key, because it is conceptionally the same as the "compass mode".

You seem to hint at that you'd like to a top-down navigation mode. I actually decreased the tilt angle for the "new" navigation mode now a bit. The gps location button is now a two-state button (follow me <-> navigation mode). It would be too confusing to make it a three-state button.

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
westnordost: > https://gitlab.com/smichel17/android-utils. Direct source link.

Nice idea though!

> Also just tested and I was mistaken about when observable is called; it's on every set, not just when the value changes

I'd say it makes sense. But then, it is hardly any less code and IMO less readable. Maybe I even read about `Delegates.observable` some years back (when I got into Kotlin) but made a conscious decision to not use them.

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
westnordost: I know the context. I disagree (, but didn't care enough to comment on that pull request or suggest that Florian should change it).

The achievement constants/enum is exclusively used always in the exact same place, so the context is clear. The context is also clear for the `LocationStateButton` because it is literally a button that shows the `LocationState` and the method that returns a `LocationState` explicitly states that it returns a `LocationState`.

Basically, many things we do in Kotlin as normal will not work well if you are not using an IDE. Starting with lambdas whose parameter types are not annotated.

smichel17: Additionally/especially, wildcard imports make it much harder to do a source dive via GitHub.

I don't mind missing parameter types as much as wildcard imports. Usually they're not hard to figure out from context.

> many things we do in Kotlin as normal will not work well if you are not using an IDE.

It's the cost of using a language made by IDE developers, I guess :pensive:
But I don't have to like it :upside_down_face:

app/src/main/java/de/westnordost/streetcomplete/location/LocationStateButton.kt
smichel17: > https://gitlab.com/smichel17/android-utils

It still works fine for the purpose documented in the readme, but I wanted to make it also handle `onSharedPreferenceChanged` events, and it turns out that's not the kind of thing which can be easily bolted on (unregistering in particular is a mess), so I'm going to need to redo the whole thing to get the ergonomics I want
…if it ever bubbles up my priority list, before I switch to a GNU/Linux phone. So probably never :P

> it is hardly any less code

Okay, but fixable enough. [Working example](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS41LjMxIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiaW1wb3J0IGtvdGxpbi5wcm9wZXJ0aWVzLkRlbGVnYXRlc1xuaW1wb3J0IGtvdGxpbi5yZWZsZWN0LktQcm9wZXJ0eVxuXG5pbXBvcnQga290bGluLnRlc3QuKlxuXG5mdW4gPFQ+aWZDaGFuZ2VkKGJsb2NrOiAoVCkgLT4gVW5pdCkgPSB7IF86IEtQcm9wZXJ0eTwqPiwgb2xkOiBULCBuZXc6IFQgLT5cbiAgICBpZiAob2xkICE9IG5ldykgYmxvY2sobmV3KVxufVxuXG5mdW4gY2VsZWJyYXRlKHg6IEludCkgPSBwcmludGxuKFwiSE9PUkFZISAkeFwiKVxuXG5mdW4gbWFpbihhcmdzOiBBcnJheTxTdHJpbmc+KSB7XG4gICAgdmFyIG1heDogSW50IGJ5IERlbGVnYXRlcy5vYnNlcnZhYmxlKDAsIGlmQ2hhbmdlZCg6OmNlbGVicmF0ZSkpXG5cbiAgICBtYXggPSAwXG4gICAgbWF4ID0gMTAgLy8gSE9PUkFZIVxuICAgIG1heCA9IDEwXG59In0=)

```kt
fun <T>ifChanged(block: () -> Unit) = { _: KProperty<*>, old: T, new: T ->
    if (old != new) block()
}
```

```suggestion
    var isNavigation: Boolean by Delegates.observable(false, ifChanged(::refreshDrawableState))
```

smichel17: Or, if you wanted a less FP style, easy enough to make your own `observable` wrapper

```kt
inline fun <T>observable(default: T, crossinline block: () -> Unit) =
    Delegates.observable<T>(default) { _, old, new ->
        if (old != new) block()
    }
```
```suggestion
    var isNavigation: Boolean by observable(false) { refreshDrawableState() }
```

westnordost: Well, I guess LiveData is the correct way to do that,  then

smichel17: Yeah, probably. Although, it can't be used on iOS (#1892). I haven't done any iOS development, so I'm not sure how similar the UI abstractions are / how much code could be shared. I imagine there's some place to draw the line so that any use of LiveData lives inside the Android-only piece; it's just something to keep in mind before depending too heavily on it.

PR #3349, opened 2021-10-08
Remove undo button menu

smichel17:
Following up on https://github.com/streetcomplete/StreetComplete/discussions/3330#discussioncomment-1399451

The undo button popup menu allegedly exists because the history screen makes it slower to undo the most recent entry. However, the menu itself *also* makes it slower to undo the most recent entry, so it barely saves any time. So, we might as well just open the full history directly, and make a few other time-saving changes:

- Pre-select the first entry, to save a tap
- Reverse the direction of the list, so that entry is right\* under where the undo button was
    - Now you basically just need to double tap to undo the most recent edit — it's actually *faster* than the previous menu.

\*Actually it's slightly below, because the jawg/osm credit text pushes up the button but not the history list. So it would be nice to add some additional bottom padding so they line up perfectly. I thing the right place to do this is the line below, but I'm not sure what the best way to get the dimensions is.

https://github.com/streetcomplete/StreetComplete/blob/a8b73d78b2a9c54ab5e67994587c99bbc97d490f/app/src/main/java/de/westnordost/streetcomplete/edithistory/EditHistoryFragment.kt#L76

It might also be nice to make it so that, if you immediately undo the most recent edit without changing the selection, the edit history doesn't stay open. But that could also be unwanted, so I've left it out initially.

Tested on an emulator but not a real devices.

/cc @mnalis, maybe you want to pull this into your fork and take it for a spin?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/edithistory/EditHistoryFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/edithistory/EditHistoryFragment.kt
@@ -78,7 +78,6 @@ class EditHistoryFragment : Fragment(R.layout.fragment_edit_history_list) {
         binding.editHistoryList.respectSystemInsets {
             updatePadding(left = it.left, top = it.top, bottom = it.bottom + initialPaddingBottom)
         }
-        (binding.editHistoryList.layoutManager as LinearLayoutManager).reverseLayout = true
         viewLifecycleScope.launch {
             val edits = withContext(Dispatchers.IO) { editHistorySource.getAll() }
             adapter.setEdits(edits)
--- a/app/src/main/res/layout/fragment_edit_history_list.xml
+++ b/app/src/main/res/layout/fragment_edit_history_list.xml
@@ -16,6 +16,7 @@
         android:paddingBottom="22dp"
         android:clipToPadding="false"
         app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+        app:reverseLayout="true"
         android:orientation="vertical"
         tools:listitem="@layout/row_edit_item" />
 


smichel17:
I believe the header should still work that way. I was pleasantly surprised at how easy this was; I just swapped which element was being passed as `editAbove`, to be the next entry in the list instead of the previous.

westnordost:
Can you provide a screenshot?

smichel17:
Okay, so this is not a great quality solution, but it works…

Screenshot_1633711309

smichel17:
It was a *little* strange to me, too, but then I thought, "it's like a calendar" and it was fine.

smichel17:
Yes, merging the lines would be trivial.

smichel17:
Oh! It's just wider. One vs two digits in the day of the month.

smichel17:
> It was a little strange to me, too, but then I thought, "it's like a calendar" and it was fine.

Update: after using it a little (only in the emulator while testing!) it already felt strange when I went back to the previous direction.

smichel17:
Oop, missed your comment.

> There is also the list row that has "synced" written on it with arrows. The arrows must now point up instead of down.

Let me see if I can get that to appear without pushing some bogus edit to OSM.

smichel17:
The wrapping of "synced" looks awkward to me. What about something more like this? It takes more vertical space (we can reduce it a little), but I think it looks much cleaner.

image |||||||||||||||||||||||| image

smichel17:
Alternatively, we could use the upload icon; I think that would be nice. Any preference for left or above?
I guess it depends on the length of the translations, although we could ellipsize instead of wrapping.
(Left is a smaller version of the icon, to match the text height).

image |||||||||||||||| image

The view has `android:layout_width="match_parent"`, so I'm not sure why it is not expanding to the full width. (The same problem is present for the arrows-on-top version, I'm just hacking around it for the screenshot by manually setting the width to `80dp`)

PR #3352, opened 2021-10-08
rerun generation of ATM operator list

matkoniecz:
note that countries like Australia and USA have no longer lists of ATM operators split by state - I think that it is fine. Autocomplete works and it seems that benefit of such splits is not huge.

But I can add also that - at cost of some coding.

BTW, if someone is interested in cleanup of `operator` tags on ATM in their area I can share Python code that I created to support this (usable in manually reviewed edits and in fully automatic bot edits).

If someone cleaned `operator` tags in their area - let me know and I can run metadata update in their country.

### DIFF
--- a/app/src/main/assets/country_metadata/BA.yml
+++ b/app/src/main/assets/country_metadata/BA.yml
@@ -1,4 +1,5 @@
 # Do not edit. Source files are in /res/country_metadata
+atmOperators: [UniCredit Bank, Raiffeisen Bank, Raiffeisen, Nova Banka, NLB Banka, Sberbank, Unicredit, Unicredit bank, Raifefisen Bank dd BIH, NLB, Raiffeisen bank, Unicredit Bank, Raiffeisenbank, Sparkasse, Addiko Bank, Komercijalna Banka, Raiffeisen banka, Unicredit banka, UniCredit, Ziraat Bank, Sparkasse Bank, Raiffeisen Bank BiH, Raiffeisen Bank dd BIH, Bosna Bank International, BBI, Komercijalna banka, NLB banka, Intesa San Paolo]
 chargingStationOperators: [City]
 isLivingStreetKnown: true
 isSlowZoneKnown: true
--- a/app/src/main/assets/country_metadata/BT.yml
+++ b/app/src/main/assets/country_metadata/BT.yml
@@ -1,4 +1,5 @@
 # Do not edit. Source files are in /res/country_metadata
+atmOperators: [Bank of Bhutan, BOB, Bhutan National Bank, Druk PNB, Bank Of Bhutan, BNB, Tashi Bank, BOBL, BDBL]
 isLeftHandTraffic: true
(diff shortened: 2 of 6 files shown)


matkoniecz:
`AW`, `AX`, `BA` - I need to investigate why atm data disappeared for them.

kmpoppe:
I mean, the best way to fix duplicates would be to use wikidata for the operators, but looking at the absolute bucketload of entries for `DE` this is going to be a crazy task to fix ... I think I'll get some peeps together to fix that 😉

matkoniecz:
> What does your python code do to help with cleanup, @matkoniecz? I might give it a try, if it does not generate too much work.

It makes such retagging much easier to do by automating most of it while still allowing review. Let me know if something can be improved.

It is a template that allows specifying:

- unwanted `operator` values
- wanted operator value
- where cache file should be stored

Then it runs an Overpass query and for each element it 
- opens given element in browser - for review
- lists tags of the object, lists what wold be edited
- asks user whether edit should be made or not
- on accepting the edit program makes the edit, it is possible to skip given edit. Typical edit looks like 

It handles some things like retagging also `name` - but please do not use it blindly. Let me know if something is buggy, but potential users take a full responsibility. It is best to not use without looking through the code (it is relatively simple).

It is possible to configure it to perform edits fully automatically, but that would require permission from local community (see https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct and https://wiki.openstreetmap.org/wiki/Mechanical_Edits/Mateusz_Konieczny_-_bot_account/fix_overly_verbose_Euronet_Sp._z_o.o. for a template of documentation page).

Practical hint: keep script window open on one part of the screen, browser on another so opening objects for review will not require you to switch windows.

Example edit: https://www.openstreetmap.org/changeset/112288731

Code: https://gitlab.com/matkoniecz/osm-automation/-/blob/master/script_assisted_cleanup/tag_migrator_normalize_bank_polskiej_sp%C3%B3%C5%82dzielczo%C5%9Bci.py

If you open issue/PR on github feel free to ping me here - not sure are their email notifications actually working.

> I mean, the best way to fix duplicates would be to use wikidata for the operators

https://github.com/osmlab/name-suggestion-index is trying to do this, not sure about progress with ATMs. In principle their name matching could be used for eliminating unwanted names and/or providing suggestions "did you mean XYZ". But as far as I know the ATM data is not sufficient and not worth using. Let me know if it is an outdated info!

matkoniecz:
I think that missing special support for Aruba (32 mapped ATMs in total - http://overpass-turbo.eu/s/1bTy ) and similar `admin_level=3` areas is not worth effort.

Though I will not merge without @westnordost review and I can add this ATM data back (via expanding code that generates it to support also `admin_level=3` states)

See https://www.openstreetmap.org/relation/1231749 https://en.wikipedia.org/wiki/Aruba

Though Bosnia and Herzegovina should have this data...

PR #3353, opened 2021-10-09
clarify building=bridge - that it should not be used for where man_made=bridge should be used

matkoniecz:
`building=*` was very often used for bridge outlines. Introduction of `man_made=bridge` reduced this, but there are still places where it happens.

This extra info tries to direct user to leaving note in case where `bridge=yes` area that should be `man_made=bridge` is encountered by SC mapper

screen

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
@@ -62,7 +62,7 @@ fun BuildingType.asItem(): Item<BuildingType?> = when(this) {
     BOATHOUSE -> Item(this, R.drawable.ic_building_boathouse, R.string.quest_buildingType_boathouse)
     ALLOTMENT_HOUSE -> Item(this, R.drawable.ic_building_allotment_house, R.string.quest_buildingType_allotment_house)
     ROOF -> Item(this, R.drawable.ic_building_roof, R.string.quest_buildingType_roof)
-    BRIDGE -> Item(this, R.drawable.ic_building_bridge, R.string.quest_buildingType_bridge)
+    BRIDGE -> Item(this, R.drawable.ic_building_bridge, R.string.quest_buildingType_bridge, R.string.quest_buildingType_bridge_description)
     TOILETS -> Item(this, R.drawable.ic_building_toilets, R.string.quest_buildingType_toilets)
     SERVICE -> Item(this, R.drawable.ic_building_service, R.string.quest_buildingType_service, R.string.quest_buildingType_service_description)
     HANGAR -> Item(this, R.drawable.ic_building_hangar, R.string.quest_buildingType_hangar, R.string.quest_buildingType_hangar_description)
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -687,6 +687,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_buildingType_hut_description">small, simple dwelling or shelter</string>
     <string name="quest_buildingType_roof">Roof</string>
     <string name="quest_buildingType_bridge">Bridge between buildings (skyway)</string>
+    <string name="quest_buildingType_bridge_description">Only for bridges that are also a building or building part.</string>
     <string name="quest_buildingType_service">Service building</string>
     <string name="quest_buildingType_service_description">building with machinery like pumps or transformers</string>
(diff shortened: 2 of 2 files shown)


matkoniecz:
> plus users of the app do not see the actual tagging.

But they see object being classified.

> But the icon already shows such bridge

Yes, but it is not 100% clearly excluding such case.

So it clarifies that

screen01
https://www.google.pl/maps/@50.0648041,19.9401325,3a,75y,86.07h,113.7t/data=!3m6!1e1!3m4!1sGi4D2CDqXQwGFMIca9rSEw!2e0!7i13312!8i6656

fits, while
screen02
https://www.google.pl/maps/@50.0488372,19.9345845,3a,77.8y,340.48h,118.48t/data=!3m7!1e1!3m5!1sgkrNBOxi4t_ST_idmTHuag!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fpanoid%3DgkrNBOxi4t_ST_idmTHuag%26cb_client%3Dmaps_sv.tactile.gps%26w%3D203%26h%3D100%26yaw%3D68.29812%26pitch%3D0%26thumbfov%3D100!7i13312!8i6656

does not fit

westnordost:
> But they see object being classified.

I don't understand, what do you mean?

matkoniecz:
User mapping and answering question is standing next to the `building=yes` object that generated this quest.

In some cases it will be modern skyway matching picture (no problem at all). But in some cases it will be outline of a bridge or [historic skyway](https://user-images.githubusercontent.com/899988/136663055-43ec03fe-3642-4aeb-be03-78b101c8afa4.png). In such cases I think that extra hint can improve quality of SC answers, and prevent [bridges](https://user-images.githubusercontent.com/899988/136663102-a74b585a-b0e9-4648-a60a-95c9724b542b.png) being classified as `building=bridge`, as in such case note creation is strictly preferable.

> But the icon already shows such bridge

Some are really differing in style

> plus users of the app do not see the actual tagging

in this case not being aware about tagging is not really changing things, I think

matkoniecz:
Patch, just in case it would be turned out to be a good idea:

```
diff --git c/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt w/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
index d4ede3b53..0edf68562 100644
--- c/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
+++ w/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
@@ -62,7 +62,7 @@ fun BuildingType.asItem(): Item<BuildingType?> = when(this) {
     BOATHOUSE -> Item(this, R.drawable.ic_building_boathouse, R.string.quest_buildingType_boathouse)
     ALLOTMENT_HOUSE -> Item(this, R.drawable.ic_building_allotment_house, R.string.quest_buildingType_allotment_house)
     ROOF -> Item(this, R.drawable.ic_building_roof, R.string.quest_buildingType_roof)
-    BRIDGE -> Item(this, R.drawable.ic_building_bridge, R.string.quest_buildingType_bridge)
+    BRIDGE -> Item(this, R.drawable.ic_building_bridge, R.string.quest_buildingType_bridge, R.string.quest_buildingType_bridge_description)
     TOILETS -> Item(this, R.drawable.ic_building_toilets, R.string.quest_buildingType_toilets)
     SERVICE -> Item(this, R.drawable.ic_building_service, R.string.quest_buildingType_service, R.string.quest_buildingType_service_description)
     HANGAR -> Item(this, R.drawable.ic_building_hangar, R.string.quest_buildingType_hangar, R.string.quest_buildingType_hangar_description)
diff --git c/app/src/main/res/values/strings.xml w/app/src/main/res/values/strings.xml
index f5561bc22..0e718d800 100644
--- c/app/src/main/res/values/strings.xml
+++ w/app/src/main/res/values/strings.xml
@@ -687,6 +687,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_buildingType_hut_description">small, simple dwelling or shelter</string>
     <string name="quest_buildingType_roof">Roof</string>
     <string name="quest_buildingType_bridge">Bridge between buildings (skyway)</string>
+    <string name="quest_buildingType_bridge_description">Only for bridges that are also a building or building part.</string>
     <string name="quest_buildingType_service">Service building</string>
     <string name="quest_buildingType_service_description">building with machinery like pumps or transformers</string>
     <string name="quest_buildingType_hangar">Hangar</string>
```

PR #3356, opened 2021-10-10
Don't ask for crossing type when the highway is private

dbdean:
If a highway=crossing is on a road with access=private or access=no, we should not be encouraging the mapper to ascertain what sort of crossing it is.


matkoniecz:

dbdean:

PR #3357, opened 2021-10-10
Dont ask wheelchair questions on things that are marked as private

dbdean:
Don't ask for wheelchair access on things that are marked as access=no or access=private.

In my case, it is to stop a private school library from asking me this question.

Previously only toilets were restricting access based on access !~ private|customers. I changed that to !~ no|private to match the other wheelchair quests changes in this PR. If we do want to keep not asking for toilets only for customers, I'm happy to change it back to do so.


matkoniecz:

dbdean:

PR #3358, opened 2021-10-10
Don't ask for internet access on things that are marked as private

dbdean:
If a thing is marked as private, we shouldn't be asking mappers to go and work out if it provides internet access or not.

In my case, it is to stop a private school library from asking me this question.


matkoniecz:

dbdean:

PR #3367, opened 2021-10-11
Add fire_hydrant:position Quest

thefeiter:
So far it should be already quite usable. I used the fire_hydrant:type quest as a template and changed imaged and strings.

I just need an image for the "parking_lot" position.


westnordost:
Not convinced this information is reasonable to collect. Better create an issue for that before working on it to see if this would be added at all or not.

PR #3368, opened 2021-10-11
add fire hydrant position quest

thefeiter:
So far it should be already quite usable. I used the fire_hydrant:type quest as a template and changed imaged and strings.

The tag used by this quest is `fire_hydrant:position` ([wiki, under optional tags](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_hydrant))
This Could be useful to find a hydrant faster and is a fast to answer quest.

I just need an image for the "parking_lot" position.


matkoniecz:
It seems to me `fire_hydrant:position` should be mapped by locating hydrant node at the exact position where it is located, like any other objects. And mapping surrounding of fire hydrant is best done by, well, mapping surroundings of hydrant rather describing them as a special tag.

westnordost:
Hmm, is there no dedicated documentation for the key? Where did you get the pictures from? Did you make them yourself? There is a authors.txt also in the /res/ directory, you can add the author + license there.

Maybe if you created them yourself, you also want to add them to the wiki and properly document the tag as it is used?

westnordost:
ping

thefeiter:
Yep, still on my list, but no time for now. Could be closed and reopened later.

17.11.2021 16:59:45 Tobias Zwick ***@***.***>:

thefeiter:
Now the Tags are properly documented in the wiki along with pictures

thefeiter:
photo_2021-12-28_21-07-46

One could try to find a better image for the lane option, but this is not easy since if you crop to the hydrant, it is no longer recognizable as a lane.

smichel17:
Nice!

I think that type of image is fine for the lane. However, it should be more centered-- that *particular* image has the hydrant so close to the bottom that it is partially covered by the text!

For phrasing:

- "On the side of a road lane" -- is *side* relevant for the tagging? If not, I'd shorten it to one of: "In the lane of a road" or "In a travel lane" (depending on how obvious the road part is from the picture).

- "In a parking" -- needs a word added to the end. Probably one of these, depending on your desired meaning: spot, space, lane, area.

- The other two are good.

thefeiter:
> * "On the side of a road lane" -- is _side_ relevant for the tagging? If not, I'd shorten it to one of: "In the lane of a road" or "In a travel lane" (depending on how obvious the road part is from the picture).

What about just "On a road" ? 
I just mapped some hydrants and this seems to be just fine.

westnordost:
Better wordings are IMO:
- On the roadway
- On the sidewalk
- In grass or dirt
- In a parking

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant/AddFireHydrantPosition.kt
```diff
@@ -0,0 +1,30 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
+
+class AddFireHydrantPosition : OsmFilterQuestType<FireHydrantPosition>() {
+
+    override val elementFilter = """
+        nodes with
+         emergency = fire_hydrant and
+         !fire_hydrant:position and
+         fire_hydrant:type and
+         (fire_hydrant:type = pillar or fire_hydrant:type = underground)
```
westnordost, line 15: Not sure if it really should be asked for pillars. First of all, for pillars, the pictures might be confusing. Second, pillars shouldn't be that hard to spot.

thefeiter:
I also took some new pictures:

https://commons.wikimedia.org/wiki/File:Fire_hydrant_on_road_side.jpg
https://commons.wikimedia.org/wiki/File:Fire_hydrant_on_sidewalk_2.jpg

What do you think about using these?

thefeiter:
> * In a parking

Are you sure, that "In a parking area" wouldn't be better?

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant/AddFireHydrantPosition.kt
thefeiter: But the given locations can be applied to this type of hydrant and this data wouldn't be collected otherwise.

westnordost:
I don't know

smichel17:
"In a parking" sounds awkward. I am not sure exactly how the grammar works here :grimacing:, but here is an example that feels the same way:

:heavy_check_mark: "It is for parking"
:x: "It is a for parking".


Going by "it sounds right", the only two situations that I can think of off hand where "parking" stands alone as a noun are "What type of parking is this?" and "Where is the parking [for this store]?" (bracketed part sometimes implied rather than spoken). Otherwise, it is an adjective and thus needs a noun to go with it (spot, space, lane, area…). "Area" is the most broad. I would prefer a more specific word, but only if it is correct.

thefeiter:
Then instead of "parking area" one could go with "parking spot" since it is closer to the reason that one would tag it as such because the hydrant might be blocked by a parked car.

smichel17:
That would be my first choice, as long as it covers all cases where this tag should be applied.

thefeiter:
Well the wiki says 

> The hydrant is in a parking. 

But thats not very helpful here...

thefeiter:
Maybe one could adjust the wiki description as well...

thefeiter:
The key for the tag is actually "parking_lot" so one could actually go with that.

thefeiter:
And it is only used as value only for 2.25% of all occurances.

smichel17:
How should it be tagged if it is in a parking space on the side of a road?

- If it should be `parking_lot`, then "In a parking lot" is not appropriate, because I would not pick it in this case.
- If it should be `lane`, then "In a parking lot" is good, and further the lane answer can just be "In a road".

thefeiter:
Yea, I would go with that. `parking_lot` only when it is a dedicated parking spot, otherwise I would tag as `lane`.

thefeiter:
I would be tagging a hydrant in a specifically marked spot on the side of a road as `parking_lot` since cars would be frequently parked there. So the description "In a parking spot" would be better.

westnordost:
> because the hydrant might be blocked by a parked car.

Maybe write "In a parking spot (dude!)" ;-)
At least that was my first reaction to that picture. Seriously, who puts a parking spot over a fire hydrant, is that even legal?

thefeiter:
> Seriously, who puts a parking spot over a fire hydrant, is that even legal?

Yep, I wondered as well. Sounds pretty dumb.

smichel17:
From that perspective, 2.25% seems ridiculously high. I don't think I've ever seen a column hydrant with a parking spot in front of it (and additionally it's illegal to park in front of one).

thefeiter:
So this seems to be ready for merge from my side...
Screenshot_20211229-202236_Street­Complete_Dev

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant/AddFireHydrantPosition.kt
westnordost: Can a pillar hydrant be on a parking?

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant/AddFireHydrantPosition.kt
```diff
@@ -0,0 +1,30 @@
...
+import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
+
+class AddFireHydrantPosition : OsmFilterQuestType<FireHydrantPosition>() {
+
+    override val elementFilter = """
+        nodes with
+         emergency = fire_hydrant and
+         !fire_hydrant:position and
```
westnordost, line 13:
```suggestion
         (!fire_hydrant:position or fire_hydrant:position ~ "\?|fixme")
```

westnordost (approved):
Very good! I have nothing to complain, just one suggestion

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant/AddFireHydrantPosition.kt
```diff
@@ -0,0 +1,30 @@
...
+import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
+
+class AddFireHydrantPosition : OsmFilterQuestType<FireHydrantPosition>() {
+
+    override val elementFilter = """
+        nodes with
+         emergency = fire_hydrant and
+         (!fire_hydrant:position or fire_hydrant:position ~ "\?|fixme")
```
FloEdelmann, line 13: Now an `and` is missing:

```suggestion
         (!fire_hydrant:position or fire_hydrant:position ~ "\?|fixme") and
```

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant/AddFireHydrantPosition.kt
```diff
@@ -0,0 +1,30 @@
...
+         fire_hydrant:type and
+         (fire_hydrant:type = pillar or fire_hydrant:type = underground)
```
FloEdelmann, line 15: This can probably be simplified:

```suggestion
         (fire_hydrant:type = pillar or fire_hydrant:type = underground)
```

FloEdelmann (changes requested):
I have just two minor nits below, and one suggestion: Maybe the contrast of the "lane" image can be increased a bit, so that the hydrants get more visible against the asphalt?

But that is not merge-blocking from my side (only the missing `and` in the element filter is). Good job!

westnordost:
I just added an icon for the quest - ic_quest_fire_hydrant_grass , nothing special, but at least a different one than for the fire hydrant type quest.

FloEdelmann (approved):
Nice!

westnordost (commented):
Hmm, one more thing: I noticed that the pictures seem to be very large, both in file size and in pixel size.

The pictures have a size of 384x384 dp. That is as large as they'd need to be if each picture should cover the whole form in width. So actually, they can be half the size:

mdpi: 192x192
hdpi: 288x288
xhdpi: 384x384
xxhdpi: 576x576

Also note that the JPEG quality doesn't need to be 90, depending on the image displayed, it can be lower without visible artifacts. (When saving with GIMP, you can compare how the image would look by adjusting the slider.)

westnordost:
👍

westnordost:
I wonder, would it make sense to do the same improvement for this quest as done in #3608 ?

thefeiter:
That's a goo Idea. I would say this is another task for @tiptoptom...

westnordost:
Why tiptoptom?

thefeiter:
Because he was visiting me and did all the work, but I commited it.

31.12.2021 01:18:36 Tobias Zwick ***@***.***>:

tiptoptom:
But I had a lot of help from @thefeiter

matkoniecz:
It would be preferable to mention authors in ` res/graphics/authors.txt`

Note that link is present but file may be deleted/moved/renamed in the future and the license requires attribution.

Also, the license info likely should be in https://github.com/streetcomplete/StreetComplete/blob/bb71ae91976ce4f16c22c7ce09ecdcc388a649e1/app/src/main/res/authors.txt

thefeiter:
Do you mean the authors of the images?
I took 3/4 of them, so I could modify `res/graphics/authors.txt` to:
```
fire hydrant position graphics/
    Fire_hydrant_in_grass.jpg                   Marvin Feiter https://upload.wikimedia.org/wikipedia/commons/2/2d/Fire_hydrant_in_grass.jpg (CC BY-SA 4.0)
    Fire_hydrant_on_sidewalk.jpg                Marvin Feiter https://upload.wikimedia.org/wikipedia/commons/3/32/Fire_hydrant_on_sidewalk.jpg (CC BY-SA 4.0)
    Fire_hydrant_lane.jpg                       Marvin Feiter modified from https://upload.wikimedia.org/wikipedia/commons/d/d0/Fire_hydrant_on_road_side.jpg (CC BY-SA 4.0)
    Fire_hydrant_parking.jpg                    centauri modified from https://upload.wikimedia.org/wikipedia/commons/3/33/Fire-fighting-facility_node-4279866277.jpg (CC BY 2.0)
```

And also add this to the file you mentioned? Is that what you mean?

westnordost:
Already moved the attribution to the other authors.txt

tiptoptom:
Am I right that the Position quest does not currently show neighboring hydrants like the other hydrant quests?

matkoniecz:
@tiptoptom Good point, changed: https://github.com/streetcomplete/StreetComplete/commit/1d809c1b03c6d714e1c585663b0b1a0de52ee905

mpawelski:
This is great addition. I just notices this new quest and went on to fill it in my neighborhood and want to share my experience. Luckily my thirst three hydrants were in three different places (`green`, `sidewalk`, `parking_lot`) and they were `fire_hydrant:type=underground` so the experience was great.

But then I found some hydrants of type `fire_hydrant:type=pillar` and got confused. Because what I saw in front of me was this:
image
but in app I saw this:
image

And I just decided to skip this quest because of confusion, I saw different type of hydrant in real life and different in app's images.

I know that the question is about hydrant position, and I later checked that this hydrant was indeed tagged correctly as `fire_hydrant:type=pillar` so I should just choose `grass`. But maybe we can think about showing different pictures if hydrant is indeed tagged as `fire_hydrant:type=pillar`? StreetComplete is great because of its UX and use of good pictures to ask question. I think this would be small nice UX improvement for average user.

westnordost:
Hm, I suppose this would be an improvement. Pillar hydrants are probably also never in parkings and never on the street surface, so for pillar hydrants, it could be just two images.
Would you create a new issue with this request? Maybe you could also look for freely licensed (or make yourself) pictures of pillar hydrants in the grass/dirt and pillar hydrants on the sidewalk.

FloEdelmann:
> never on the street surface

Maybe on streets that don't have a sidewalk, a pillar hydrant may be located on one side of the road?

westnordost:
Maybe... would be hard to find an example photo i imagine

On 28 January 2022 14:24:19 CET, Flo Edelmann ***@***.***> wrote:
>> never on the street surface
>
>Maybe on streets that don't have a sidewalk, a pillar hydrant may be
>located on one side of the road?
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/3368#issuecomment-1024222956
>You are receiving this because you modified the open/close state.
>
>Message ID:
***@***.***>

-- 

smichel17:
Maybe overpass to find hydrants tagged the way @floedelman describes; 
then Google Street View to check if they are in fact on the street; 
then someone could visit to take a freely-licensed photo. If we can't 
find one within travel distance of one of us, then maybe it's rare 
enough that it doesn't matter.

…not sure if it is worth the effort, though.

FloEdelmann:
@smichel17 See #3686: I already found one (via Wikimedia Commons): https://commons.wikimedia.org/wiki/File:Fire-fighting-facility_node-6464454770.jpg

The OSM node (https://www.openstreetmap.org/node/6464454770) is tagged `fire_hydrant:position=lane`.

mnalis:
> Hm, I suppose this would be an improvement. Pillar hydrants are probably also never in parkings and never on the street surface, so for pillar hydrants, it could be just two images.

Actually surprising percentage of pillars so far (I've just started mapping those, though) for me seems to be on parking lots (although I wholeheartedly agree that is probably not very smart idea - but it does not look like particularly good idea for underground ones either). Eg. https://www.openstreetmap.org/node/3423554061 and https://www.openstreetmap.org/node/3423477857 just the other day. (Obviously it is at the far end of parking or between parking places). I can take picture of that one if wanted.

Didn't see any on the side of road yet IIRC. I'll take notice if I see one.

PR #3370, opened 2021-10-11
stile resurvey, fixes #3187

### LINKED ISSUE #3187: Resurveying of stiles
See also: #2991, #3141, #3188.

I think it would probably be advantageous to re-survey stiles.

I said [here](https://github.com/streetcomplete/StreetComplete/issues/2991#issuecomment-900486882 "link to comment in issue #2991") (clarified and expanded for this issue):

> As far as I can tell, stiles are never resurveyed.  I agree with what Philip Barnes' said (summarised in [this comment](https://github.com/streetcomplete/StreetComplete/issues/2991#issuecomment-899841013)) about resurveying stiles at regular intervals.  I'm not really sure what the interval should be, but my starting suggestion would be 5 years.  In the UK, stiles are gradually being replaced in many areas, and they are not supposed to be installed where one did not exist before (I'm not sure at this precise moment whether this is a legal requirement, or just guidance, it might say in one of the links in my comment on #3141).  I'm not sure whether it would be worth asking the whole stile type quest again, or not.

An alternative could be to just show the stile type it is tagged as, along with possible replacements (see #3188), and if it's tagged `stile=stepover` without a `material=*` tag, then the two options for the step-over stile could be displayed (as discussed in #2991 and added in #3175).  Under `Other answers…`, one could have `It's another type of stile`, which would remove the `stile=*` tag, and any associated tags (`material`, `height`, `width`, `step_count`, `dog_gate`, plus any others currently removed when a `barrier=stile` is changed to a `barrier=kissing_gate`)

If the whole stile type quest is asked again, I think that the current value could be put first in the list, I'm not sure if you already do this for other resurveying quests (e.g. surface of ways), if not, I don't think it's crucial, just helpful.

I'm going to post a message on [Talk-GB] for opinions specifically about re-surveying stiles (combined with #3188).
I will ask about:
- how often stiles should be re-surveyed, and
- which tags should be removed if a stile type is changed (or if the stile is removed entirely).

I will post updates:
- when I've sent the message to the list, and
- when I seem to have received all the responses I'm going to get.

matkoniecz:
resurvey stiles, fixes #3187

resurvey happens after 8 years, so only small pool of old stiles with stile=* becomes qualified
    stiles without stile=* tag would get this quest anyway
note that in rare cases, where stile type is unchanged and material tag is added - then not strictly necessary stile:check_date will be added

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileTypeTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileTypeTest.kt
@@ -59,25 +59,24 @@ class AddStileTypeTest {
             ),
             BarrierType.STILE_SQUEEZER,
             StringMapEntryAdd("stile", "squeezer"),
-            StringMapEntryAdd("material", "stone"),
         )
     }
 
     @Test
-    fun `don't reset tags when adding a stile type tag to an existing enhanced metadata stile`() {
+    fun `handle adding stile tag where other values are present`() {
         questType.verifyAnswer(
             mapOf(
                 "barrier" to "stile",
+                "material" to "stone",
                 "steps" to "3",
                 "tag_not_in_list_for_removal" to "dummy_value",
             ),
-            BarrierType.STILE_STEPOVER_STONE,
-            StringMapEntryAdd("stile", "stepover"),
-            StringMapEntryAdd("material", "stone"),
+            BarrierType.STILE_SQUEEZER,
+            StringMapEntryAdd("stile", "squeezer"),
         )
     }
     @Test
-    fun `handle unmodified, fully tagged stepover`() {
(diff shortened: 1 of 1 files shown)


matkoniecz:
Given that `material` is not really used anywhere despite being tagged quite frequently - I think that I will keep it as is.

westnordost:
ok so err... should I review this again?

matkoniecz:
There were no significant architectural changes, though a serious bug was fixed. Oh, and tests were added.

And I still need to handle https://github.com/streetcomplete/StreetComplete/pull/3370#pullrequestreview-780016908

peternewman:
> Given that `material` is not really used anywhere despite being tagged quite frequently - I think that I will keep it as is.

Vespucci uses it for stile related presets, although maybe just to tag them, I don't know if it displays the info too:
https://taginfo.openstreetmap.org/projects/vespucci#tags

PR #3373, opened 2021-10-12
Sort building type by recently used frequency

### LINKED ISSUE #1771: Smart sort order for recent houses quest
**Proposed Solution**

In https://github.com/westnordost/StreetComplete/issues/1692#issuecomment-578857439, @westnordost proposed sorting like this:

> Per session, count the number of times a certain option has been selected. Then, each time the dialog is opened, the first row of items are sorted by the previous selection count.
> 
> Then, for persisting, the collected data is reduced like this: Let's say the user selected 27x gabled, 10x hipped, 3x dome and 1xround. Only the order is saved, so gabled, hipped, dome, round. When the data is reloaded into memory (=new app start), it is interpreted like this: 4x gabled, 3x hipped, 2x dome, 1x round. So, the ordering is still the same, but will be changed easier if the user selects different roof sizes where he is now.

I'd love to see that applied to house type.

**Use case**

I've been mapping a lot of residential areas lately. There is a very common ratio of building types (detached house > shed > garage), so the sort order would remain stable, which would make selection much faster since I wouldn't have to verify the order before tapping. At the same time, I wouldn't want this to carry over between sessions, if I'm doing a different area later, so the storage method also works well.

Currently my workaround to guarantee a stable sort order as often as possible is:

- While walking slowly down a street…
- Select the house type for 4 detached houses (two on either side of the street, in front of me).
    - For 3 of these selections, detached house is always first, I don't have to look.
- Select the number of levels for all 4 houses. This also allows me to do houses with the same number of stories one after the other, to re-use the last-picked height when possible… another issue about this coming soon, but I need a mockup, so it'll be a few minutes.
- Select the roof type for the houses, if applicable.
- Select the building type for any sheds or garages next to the houses

This works pretty well, but is not always possible depending on the layout of the street. It's also *barely* fast enough to keep up with my (intentionally slow) walking pace, so occasionally I need to stop and finish filling out quests. So, this issue would be a good QoL improvement.

smichel17:
Fixes #1771

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/LastPickedValuesStoreTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/LastPickedValuesStoreTest.kt
@@ -1,6 +1,6 @@
 package de.westnordost.streetcomplete.quests
 
-import de.westnordost.streetcomplete.quests.Letter.* // For concisision (defined in this file)
+import de.westnordost.streetcomplete.quests.Letter.*
 import de.westnordost.streetcomplete.testutils.mock
 import de.westnordost.streetcomplete.testutils.on
 import de.westnordost.streetcomplete.view.image_select.Item


smichel17:
Okay, nothing more I want to do here, ready for merge after review.

westnordost:
In general, would it make sense to apply this sort order to all quests that have their sort order re-shuffled to have recently selected items first?

smichel17:
I'm not sure. I think it *may* be different quest-by-quest, so I'd like to at least do some spot-checks. In any case, I'll go send the follow-up clean-up PR now :)

smichel17:
I have thought most about applying this order to the **building levels** quest. For it:

- I **would** like to use this strategy for deciding **which** answers are shown.
  - Right now, when you encounter one-odd building level (in my neighborhood, that's usually a 3:0 or 1:2), it bumps some more-common value out of your recents. Then, when you go to answer it, it bumps a *different* more-common value out, instead of the uncommon one. GRR. With this strategy, the uncommon value would be dropped quickly, and I probably wouldn't need to open the keyboard at all.
- I am **not sure** if I would like to use this order for **sorting** the answers.
  - Since the levels have a natural sort order, that's pretty convenient for locating the right one to tap. I think with the recently used order, they might shift around too much.
  - But maybe it'll be more stable than I expect, and it'll be okay. I want to test with the building type quest (and possibly make behavior tweaks, like how far back in the history to look; 30 values is arbitrary) before deciding how it should work for the other quests.

westnordost:
Yeah let's use it for the **which** but then **sort** it by how it is currently sorted.

pkoby:
I'm not sure if this is the right place to comment on this, but I didn't want to open an issue for it.

I like the idea of this change, but in practice I will admit it was confusing (coming from the expectations and experience with the old configuration, especially).

Yesterday, I went to an area with a lot of different building types (many industrial and shops). When I got to an area with just rowhouses (by far the most common type around here) I had to reselect it from the full menu (not a problem), but then on the second building, the recent selection of =terrace was not in the top three (which is all I can see on my phone without scrolling) and kept changing places as I kept mapping.

If the last selection was always at the top, then when you keep selecting the same type, it's always in the same spot (whereas otherwise it might change places as it gets more "popular"). When I'm mapping terraces, there's always going to be a bunch in a row.

So essentially, my preference would be latest in spot #1 _always_, then #2-6 sorted by popularity. Thus if there are a lot of the same type (even if it's a new type), you'll find it at the top, but if you select one of #2-6, the previous #1 drops to its spot in recent popularity.

smichel17:
A [new discussion](https://github.com/streetcomplete/StreetComplete/discussions/new) would be ideal.

PR #3375, opened 2021-10-12
Refactor user, statistics and achievement data management

westnordost:
...to be consistent with how data is managed in other places in the app. I.e.
- clear separation between pure data storage (access to database), data management (managing possible multiple data storages, observer pattern,...) and "actors" 
- work with observer pattern, i.e. statistics change, achievement controller listens to that change and updates achievements etc

overview_data_flow1


westnordost:
Added about 250 lines of test code, so the new code is about 150 lines longer than the old implementation.

To do: Test all the user related stuff - login, logout, update statistics, achievements etc...

westnordost:
I figure it **would** be possible to further seperate achievements and links (and even new user achievements). However, I'll leave it like this until links can be unlocked by something else than achievements.

PR #3379, opened 2021-10-14
Remove ability to upload (for testing :rofl:)

smichel17:
Right now the upload button has an indeterminate progress spinner. I think it'd be nice if it were a progress bar (err, circle), which should treat every edit as the same size.

I think this should be pretty easy to implement, but I don't want to have to upload actual answers in order to test it. So, the first thing I did was try to make the uploader totally non-functional— just **discard every edit** as though it failed, after an artificial delay.

It seems to work alright. Would you be interested in a PR that adds a build flavor with similar changes, or something like that?


smichel17:
…where did those other commits come from? I just made a fresh branch off of master…

smichel17:
Whatever. Rebased, forced pushed, it's gone :shrug:

PR #3380, opened 2021-10-14
Traffic calming, fixes #2742

### LINKED ISSUE #2742: What sort of traffic calming is here? (for nodes, ways, areas tagged traffic_calming=yes)
### General
Affected tag(s) to be modified/added: [traffic_calming](https://wiki.openstreetmap.org/wiki/Key:traffic_calming)
Question asked: **What sort of traffic calming device is here?**

### Checklist

There are currently around 14k nodes tagged as traffic_calming = yes: https://taginfo.openstreetmap.org/tags/traffic_calming=yes

### Ideas for implementation

Replace traffic_calming=yes with the common values from https://wiki.openstreetmap.org/wiki/Key:traffic_calming

matkoniecz:
fixes #2742

screen08

Still not ideal, but I think that it is a good enough - https://taginfo.openstreetmap.org/tags/traffic_calming=yes has 14k elements, so not worth as extreme effort on tweaking as `surface` that covers very big part of all edits made with SC.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1084,8 +1084,8 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
     <string name="quest_presets_delete_message">"Delete preset “%s”? The quest selection and order for it will be irrevocably lost."</string>
     <string name="quest_picnicTableCover_title">Is this picnic table covered (protected from rain)?</string>
     <string name="quest_traffic_calming_type_title">What type of traffic calming is this?</string>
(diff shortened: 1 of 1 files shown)


andrewharvey:
Text is a welcome addition.

matkoniecz:
Looking at patch once again - I need to get an icon. Maybe I am capable of stealing bump sign from some traffic sign?

matkoniecz:
And I "made" an icon.

screen02

matkoniecz:
I made one more text change for cushions: https://github.com/streetcomplete/StreetComplete/commit/a9ed1aca9618586d7c868ff71de77304e1eb7036

screen12

PR #3384, opened 2021-10-15
Fix stile resurvey

westnordost:
- the resurvey is basically a check for existence not a check for a particular property, as can be seen by the behavior to remove all properties of the stile if the stile type changed. I.e. `barrier=stile + stile=stepover + material=wooden` is in itself a **primary map feature**. Thus, the quest should work the same (tagging-wise) as the check existence quest: Use `check_date`, not `check_date:stile`. Setting `check_date:stile` would also be incompatible with a possible "recheck barrier type" quest enhancement because each and every barrier is also a primary map feature on its own (i.e. **not** `check_date:barrier` is used)
- I noticed a bug in the code:  `stile:check_date` (and other check dates referring to stile) were not removed when selecting that it is a kissing gate now. I added tests for the kissing gate and also simplified the complex logic in the function (by...:)
- I refactored the code (following @peternewman 's advice) so that the osm values used are properties of the answer enum. I did this mostly because there was a warning in intellij that the `when` is not exhaustive because there are other `BarrierType`s. It turned out that the other barrier types are not selectable in the form, but still I do not think this was a clean (and confusing) solution

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileTypeTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileTypeTest.kt
@@ -85,7 +85,7 @@ class AddStileTypeTest {
         questType.verifyAnswer(
             mapOf(
                 "barrier" to "stile",
-                "material" to "wooden",
+                "material" to "wood",
                 "tag_not_in_list_for_removal" to "dummy_value",
             ),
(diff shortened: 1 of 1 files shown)


westnordost:
I'll wait if Peter Newman has anything to add and then merge it.

peternewman:
Oops, thanks for fixing my broken suggestion!

PR #3389, opened 2021-10-15
Allow to hide zoom buttons

starsep:
## Change description
I don't use zoom in/out buttons, I use pinch gestures instead.
Therefore I think there should be an option to hide them.

Change has been inspired by [Organic Maps](https://github.com/organicmaps/organicmaps) which has the same option available.
I could add their [translations](https://github.com/organicmaps/organicmaps/search?l=XML&q=pref_zoom_title) as they use Apache 2.0 license, right?

## Screenshots:
Settings             |  Zoom buttons displayed | Zoom buttons hidden
:-------------------------:|:-------------------------:|:-------------------------:
  |  |

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/Prefs.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/Prefs.kt
@@ -11,6 +11,7 @@ object Prefs {
     const val SHOW_NOTES_NOT_PHRASED_AS_QUESTIONS = "display.nonQuestionNotes"
     const val AUTOSYNC = "autosync"
     const val KEEP_SCREEN_ON = "display.keepScreenOn"
+    const val SHOW_ZOOM_BUTTONS = "display.showZoomButtons"
     const val UNGLUE_HINT_TIMES_SHOWN = "unglueHint.shown"
     const val THEME_SELECT = "theme.select"
     const val THEME_BACKGROUND = "theme.background_type"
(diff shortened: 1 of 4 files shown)


westnordost:
Who doesn't need the zoom buttons and who needs them? Someone who uses StreetComplete with two hands (one hand holding the phone, one hand using the touch screen) doesn't need them, because that person can use the finger gestures to zoom. 
Who needs them, are those people that use StreetComplete with one hand (hold the phone with one hand and use the thumb to navigate). This only works well on small phones.
So, ironically, those people who need to make full use of the screen space because their screen is small also tend to need these buttons.

So, the setting doesn't make much sense. What would make sense is to have a method of zooming that works with one thumb and doesn't take up much space.

So what I thought about to do (instead) is to remove the +/- buttons alltogether and replace it with a zoom-slider similar to what can be found in the app "mapy.cz". The (visible) size of the zoom slider would be about half of a button. 

Would you try it out in the mapy.cz app and tell me what you think? If people find this slider overall a good idea, we can create a new feature request in this tracker for that.

westnordost:
Can you name an example app?

smichel17:
Well, it works in StreetComplete :grin:

westnordost:
But it works not very good. Half the time I accidentally open the long-press menu, the other times I accidentally zoom out so far that I can see the whole of Europe

westnordost:
Actually, I found that this is a bug. It is still quite a hidden behavior, so I guess this is why all/most of the apps also have a +/- button

https://github.com/tangrams/tangram-es/issues/2295

westnordost:
Did you try the mapy.cz interface?

PR #3390, opened 2021-10-16
minor adjustment to setting building properties

### LINKED ISSUE #3386: StreetComplete tagging disused buildings as abandoned
# What happened

I encountered this issue today while surveying.

- Prior to my survey, the building was tagged (by me in June, I might add) as `building=yes` + `disused=yes`.
- Today, when I was doing a survey in the area, I was asked what the type of the building was.
- I remember thinking that I had tagged it before, but I didn't check.
- I also remembered about the abandoned option, which seemed to be the best fit.
- I therefore selected *Abandoned*, and the `abandoned=yes` tag was added.

# What I think should happen

StreetComplete should not add the `abandoned=yes` tag to buildings with a `disused=yes` tag.

I do not necessarily think that the quest should not be asked, since it may be possible to more precisely determine the building's type, even if it is disused.

# Situational information

Just for some background, I'm not 100% sure how the building ([way 954753584](https://osm.org/way/954753584/history)) concerned should be tagged.  I have, however, found a photo on Geograph which more or less depicts the current state, and also provides some context as to what the building actually is.  However, regardless of how this building should be tagged (I am leaving both in place for now), the issue still (as far as I am concerned) remains.

[Image showing what is labelled as The West Lodge, in Leazes Park, Newcastle upon Tyne.  It is a building built of red brick, with two storeys.  The building is of a somewhat ornamental nature.  All its windows and doors have been boarded up with metal boarding, and graffiti is present on some parts of the walls/boarding.  Surrounding the building is some metal temporary fencing of the standard design.  To the left we can see grass and trees, and also a building in the distant background.  To the right, we can see a trees and a low fence, beyond which there is grass.](https://www.geograph.org.uk/photo/5693609 "Geograph: The West Lodge, Leazes Park cc-by-sa/2.0 - © Mike Quinn - geograph.org.uk/p/5693609")

matkoniecz:
fixes #3386

Not commiting directly as I am still unsure is it worth extra complexity

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/AddBuildingType.kt
@@ -52,7 +52,7 @@ class AddBuildingType : OsmFilterQuestType<BuildingType>() {
                 changes.deleteIfExists("disused")
             }
             if(answer == BuildingType.RUINS && changes.getPreviousValue("abandoned") == "no") {
-                changes.deleteIfExists("disused")
+                changes.deleteIfExists("abandoned")
             }
         } else {
             changes.modify("building", answer.osmValue)


westnordost:
I don't get it. Is `disused` considered to be synonymous to `abandoned`? Do you have a link to the wiki that explains this?

Issue #3188, opened 2021-08-18
Stile replacements and other edge cases


10992-osm:
See also: #2991, #3141, \<issue to be submitted\>.

- Since the stile quest has been changed such that what was `stile=step_over` now must necessarily have a `material=stone` or `material=wood`, it might be helpful to provide a method of tagging the (very few) step-over stiles which are made of another material (metal, probably).  One could add an option for metal stiles on the main screen, but I think this won't be too helpful, as they are pretty rare ([only 207 worldwide](https://overpass-turbo.eu/s/1apE "Overpass Turbo query for metal step-over stiles"), compared to [16659 total step-over stiles](https://taginfo.openstreetmap.org/tags/stile=stepover "step-over stiles on taginfo)).  ([originally in this comment ](https://github.com/streetcomplete/StreetComplete/issues/2991#issuecomment-900432638) in #2991)
- If stiles are re-surveyed (see \<issue to be submitted\>), then it might be useful to provide more options for their replacements.  Currently, only `barrier=kissing_gate` is provided.  It [was suggested](https://github.com/streetcomplete/StreetComplete/issues/2991#issuecomment-899841013) that a stile might be replaced with a gap, a kissing gate, or another type of gate.  A kissing gate is probably the most common replacement for a stile, so the others could be hidden in the `Other answers…` menu (`The stile has been replaced by another sort of gate`, `The stile is no longer here`/`There is a gap`, `Can't say`).  Any report of a gap probably needs a note with a photo, and remote editing, but I'm not 100% sure.

I'm going to post a message on [Talk-GB] for opinions specifically about re-surveying stiles (combined with \<issue to be submitted\>.
I will ask about:
- what stiles are likely to be replaced with, and
- how a 'gap' should be mapped / responded to by StreetComplete.

I will post updates:
- when I've sent the message to the list, and
- when I seem to have received all the responses I'm going to get.

10992-osm:
The message has now been posted on [Talk-GB]:
[[Talk-GB] Removal of and changes to stiles](https://lists.openstreetmap.org/pipermail/talk-gb/2021-August/027488.html)

westnordost:
To the first point: Note that Streetcomplete does not try to cover ALL possible options but only those that are reasonably common. Let's not lose our way here.  The rest is captured via notes.

10992-osm:
[...]

PR #3391, opened 2021-10-16
change entrace image, new answer in a stile quest

matkoniecz:
partially handles #3188

screen06

screen07


westnordost:
Nice! I am happy that I did refactor that code, now this change fit in there like a glove :-)

westnordost:
btw I signed up to this hacktober event too. Instead of a t-shirt, you can also have one tree planted, so why not if you go beyond 5 PRs anyway

PR #3394, opened 2021-10-17
stop asking crossing detail for unmarked crossings

matkoniecz:
this type of object is considered as crossing in OSM tagging, but not by many regular people
as result it can be a very confusing quest, see #3341

alternative solutions:
- consider confusion as acceptable and require people to be aware about OSM classification (contrary to SC purpose and design)
- explain somehow that it may be not exactly crossing as regularly understood
- reword questions somehow to reduce confusion

Main negative effect here is that unmarked crossings may have a pedestrian island and tactile_paving data is useful also on unmarked crossings

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/crossing_island/AddCrossingIsland.kt
@@ -18,6 +18,7 @@ class AddCrossingIsland : OsmElementQuestType<Boolean> {
           highway = crossing
           and foot != no
           and crossing
+          and crossing != unmarked
           and crossing != island
           and !crossing:island
     """.toElementFilterExpression()}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt
@@ -16,7 +16,7 @@ class AddTactilePavingCrosswalk : OsmElementQuestType<Boolean> {
         nodes with
           (
             highway = traffic_signals and crossing = traffic_signals and foot != no
-            or highway = crossing and foot != no
+            or highway = crossing and foot != no and crossing and crossing != unmarked
           )
           and (
             !tactile_paving


andrewharvey:
> Main negative effect here is that unmarked crossings may have a pedestrian island and tactile_paving data is useful also on unmarked crossings

Exactly, according to https://taginfo.openstreetmap.org/tags/crossing=unmarked#combinations, 2% of `crossing=unmarked` have `crossing:island=yes` compared to 6% which have `crossing:island=no`.

Have we noticed if the confusion is coming more from some regions than others? Perhaps it could be disabled for problematic regions only?

matkoniecz:
> Have we noticed if the confusion is coming more from some regions than others? Perhaps it could be disabled for problematic regions only?

It is 100% definitely confusing in Poland (heavily confirmed). I have seen notes indicating similar confusion in UK and USA where I am also monitoring StreetComplete notes. In other English-speaking countries note value is much lower, as result I am less sure.

Given repeated discussions/confusion among OSM mappers what is counted as crossing and what not - I am expecting this confusion to be widespread across regions. But I am not sure about this.

westnordost:
I am with  @andrewharvey here. I actually planned to create a PR to exclude `highway=crossing`s that are on `highway=service` (except `service=alley` or more?) so it should exclude the worst offenders, i.e. those "crossings" where usually a driveway just crosses the sidewalk.

westnordost:
The other common type of maybe-non-crossings are just intersections in a residential area where the sidewalk has been drawn as a separate way, so the sidewalk `highway=footway` crosses the street and often a `highway=crossing` is tagged on this node. IMO this is still a crossing through but opinions differ. At least what is useful to record is whether the kerb here is lowered or not. If it is lowered, it is worth asking if there is tactile paving and if there is an island.

So _maybe maybe maybe_, we could
1. only ask for tactile paving and island on `crossing=unmarked` if it also has `kerb=lowered` (or similar) tagged
2. start asking for the **kerb height** (`kerb=*`) at all `highway=crossing`, i.e. not only [these](https://www.openstreetmap.org/node/8479758949) (vertex of a road, sidewalks are tagged on the road) but also [those](https://www.openstreetmap.org/node/718941176) (vertex of an intersection of a road and a crossing footway). Currently, StreetComplete does not set the kerb height property on the crossing node, but at the location of the kerb (`barrier=kerb`)

matkoniecz:
> IMO this is still a crossing through but opinions differ.

I agree that it is crossing - the problem is that people consistently got confused when asked about crossing detail there.

Maybe for `crossing=unmarked` show extra info about this?

arrival-spring:
> 2. start asking for the **kerb height** (`kerb=*`) at all `highway=crossing`, i.e. not only [these](https://www.openstreetmap.org/node/8479758949) (vertex of a road, sidewalks are tagged on the road) but also [those](https://www.openstreetmap.org/node/718941176) (vertex of an intersection of a road and a crossing footway). Currently, StreetComplete does not set the kerb height property on the crossing node, but at the location of the kerb (`barrier=kerb`)

The problem with this is it can sort of make it look like there's a kerb on a road. i.e. from the perspective of the road there is a node tagged with `kerb=*`. Where there is also `highway=crossing` data consumers could/would assume that the kerb property belongs to the crossing, but I'm not sure if they do.

In fact, this is what the add crossing quest does already (add `kerb=*` on a node in the middle of a road), and I'm not so sure about that, but it may well have been discussed somewhere and I missed it. I don't think it matches accepted usage of the `kerb` tag. See [the wiki](https://wiki.openstreetmap.org/wiki/Key:kerb).

westnordost:
> The problem with this is it can sort of make it look like there's a kerb on a road

`kerb=*` is not a primary map feature but a property - of either a `highway=crossing` or a `barrier=kerb`. Data consumers that are looking for barriers (should) look for `barrier=*`. As far as this tag is defined, there is no issue, it is quite clear.

> I don't think it matches accepted usage of the kerb tag. See the wiki.

Yes it does. Citation from the wiki page you linked: _"[...] If the kerb is identical on both sides of a crossing, it is possible to add the kerb=* tag to the highway=crossing node, which sacrifices accuracy for simplicity [...]"_

The (only) issue there is with this is that it is essentially duplicate information where it is tagged like this:
image
On the other hand, if this is done, then we can avoid tagging `tactile_paving`, `crossing:island` (and `crossing=unmarked` depending on what exactly we filter) on every single of those "non-crossings".

matkoniecz:
> then we can avoid tagging tactile_paving, crossing:island (and crossing=unmarked depending on what exactly we filter) on every single of those "non-crossings" because we can bail on asking any more questions if kerb=raised/none or similar.

It is worse. Most of "it is not a crossing at all" that confused people had lowered curb.

matkoniecz:
> do not show tactile paving + island + ... quests for highway=crossing if it is crossing = unmarked ( and (!kerb or kerb~no|flush) and (!tactile_paving or tactile_paving=no)) and a vertex of a highway=service

So crossings in `highway=service` with `kerb=no/flush` or `crossing and crossing!=unmarked` would be shown?

And `crossing=unmarked` on `highway=residential` questions would be shown, but with some different wording?

PR #3419, opened 2021-10-21
Refactor LastUsedValuesStore to use sequences instead of LinkedList

smichel17:
Three main benefits (in order):

1. Simpler implementation
2. More efficient, I think
3. Switches a Java dependency to Kotlin — some small progress towards #1892


westnordost:
I don't know, must be a relict from before I used injection consistently. If you remove it, kindly check if everything is still working as intended

smichel17:
Okay, this is harder than it looks. The problem is that Dagger doesn't much like generics, and `LastPickedValuesStore<T>` is one. Probably someone who knows Dagger better could get it to work, but I'm at the limit of how much time I want to spend on it right now. Anyway, it's not *really* related to this PR, just something tangential I noticed & wanted to ask about. So I'll leave it for later.

PR #3457, opened 2021-10-26
Add housename to street quest

arrival-spring:
I can't see that this has been discussed anywhere, I did see https://github.com/streetcomplete/StreetComplete/issues/213#issuecomment-406603714
> * [x]  research what should be done with house names (exclude addr:housename?) - no, if `addr:streetnumber` is present then `addr:housename` also present is not changing anything at all

But this seems to be about excluding objects with both housenumber and housename

At least in the UK, houses with housename belong to a street/place in the same way that houses with housenumber do, so it feels odd when answering quests that only with housenumber do you get the street quest after answering.

I don't know if that's different in different parts of the world though.

(This is untested)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
@@ -37,7 +37,7 @@ class AddAddressStreet : OsmElementQuestType<AddressStreetAnswer> {
     override fun getTitle(tags: Map<String, String>) = R.string.quest_address_street_title
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> =
-        arrayOfNotNull(tags["addr:streetnumber"] ?: tags["addr:housenumber"])
+        arrayOfNotNull(tags["addr:streetnumber"] ?: tags["addr:housenumber"] ?: tags["addr:housename"])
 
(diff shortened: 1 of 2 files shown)


westnordost:
@arrival-spring let's use 

`<string name="quest_address_street_title_housename">What street does (house) %s belong to?</string>`

then

arrival-spring:
There we go, I think that's everything, I haven't made it a new string as it's a minor change (in terms of translations).

Works as expected.

When an object has both housename and housenumber tagged it only states the number in the title (which seems fine), but in the subtitle above it the housename is shown (this was also true before these changes). Not sure if that's an issue though.

westnordost:
@arrival-spring I actually meant adding another string. So there are two strings, one for number, one for name.

I guess ... your current solution would be fine for me too.

arrival-spring:
Ah yes, I now see that, I'm happy either way

westnordost (approved):

matkoniecz (commented):
Have not tested the code (has anyone tested?) but issues that I noticed are now fixed.

westnordost:
Would you mark it as approve then? I will only merge it later (goes into next major).

matkoniecz (approved):
erasing "request changes" state, though I had not done testing

arrival-spring:
I tested with these same changes made locally and it seemed to work as expected, looking at objects with housename and housenumber and with just housename.

PR #3473, opened 2021-10-30
reword parking access - #3429

matkoniecz:
primary problem was that publicly owned parking may have private access (say, parking where just president of city is allowed to park) and private parking may have public access (say, paid parking near airport on private grounds)

previous wording was not fully solving it and was still confusing

this rewording tries to clarify what exactly is being asked

Note that especially wording about customers is not 100% fully succesful, but seems the best found so far

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -561,8 +561,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
     <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
     <string name="quest_access_yes">Open to general public</string>
-    <string name="quest_access_private">Not open to general public</string>
     <string name="quest_access_customers">Open only to people visiting a specific place (e.g. customers)</string>
+    <string name="quest_access_private">Not open to general public</string>
     <string name="quest_streetName_pedestrian_title">What is the name of this pedestrian street?</string>
     <string name="quest_fee_answer_hours">Depends on time and day…</string>
     <string name="quest_fee_add_times">"Add times"</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -559,10 +559,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_religion_for_wayside_shrine_title">What religion is represented at this shrine?</string>
     <string name="quest_parking_fee_title">Do you have to pay to park here?</string>
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
-    <string name="quest_parking_access_title">Is it restricted who may park here?</string>
-    <string name="quest_parking_access_private">It is private</string>
-    <string name="quest_parking_access_yes">It is open to the public (with or without fee)</string>
-    <string name="quest_parking_access_customers">It is only for customers</string>
+    <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
+    <string name="quest_parking_access_yes2">Open to general public</string>
+    <string name="quest_parking_access_private2">Not open to general public</string>
+    <string name="quest_parking_access_customers2">Open only to people visiting a specific place (e.g. customers)</string>
     <string name="quest_streetName_pedestrian_title">What is the name of this pedestrian street?</string>
     <string name="quest_fee_answer_hours">Depends on time and day…</string>
     <string name="quest_fee_add_times">"Add times"</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -560,9 +560,9 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_parking_fee_title">Do you have to pay to park here?</string>
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
     <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
-    <string name="quest_parking_access_yes2">Open to general public</string>
-    <string name="quest_parking_access_private2">Not open to general public</string>
-    <string name="quest_parking_access_customers2">Open only to people visiting a specific place (e.g. customers)</string>
+    <string name="quest_access_yes">Open to general public</string>
+    <string name="quest_access_private">Not open to general public</string>
+    <string name="quest_access_customers">Open only to people visiting a specific place (e.g. customers)</string>
     <string name="quest_streetName_pedestrian_title">What is the name of this pedestrian street?</string>
     <string name="quest_fee_answer_hours">Depends on time and day…</string>
     <string name="quest_fee_add_times">"Add times"</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -559,10 +559,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_religion_for_wayside_shrine_title">What religion is represented at this shrine?</string>
     <string name="quest_parking_fee_title">Do you have to pay to park here?</string>
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
-    <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
-    <string name="quest_access_yes">Open to general public</string>
-    <string name="quest_access_customers">Open only to people visiting a specific place (e.g. customers)</string>
-    <string name="quest_access_private">Not open to general public</string>
+    <string name="quest_parking_access_title2">Who is allowed to park here? Parking may be free or paid.</string>
+    <string name="quest_access_yes">Any member of the general public</string>
+    <string name="quest_access_customers">Any person visiting a specific place (e.g. customers only)</string>
+    <string name="quest_access_private">Only individual(s) with permission</string>
     <string name="quest_streetName_pedestrian_title">What is the name of this pedestrian street?</string>
     <string name="quest_fee_answer_hours">Depends on time and day…</string>
     <string name="quest_fee_add_times">"Add times"</string>
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddParkingAccess.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddParkingAccess.kt
@@ -29,7 +29,7 @@ class AddParkingAccess : OsmFilterQuestType<ParkingAccess>() {
 
     override val questTypeAchievements = listOf(CAR)
 
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_parking_access_title
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_parking_access_title2
 
     override fun createForm() = AddParkingAccessForm()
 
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -559,10 +559,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
     <string name="quest_parking_fee_title">Do you have to pay to park here?</string>
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
-    <string name="quest_parking_access_title">Is it restricted who may park here?</string>
-    <string name="quest_parking_access_private">It is private</string>
-    <string name="quest_parking_access_yes">It is open to the public (with or without fee)</string>
-    <string name="quest_parking_access_customers">It is only for customers</string>
+    <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
+    <string name="quest_parking_access_yes2">Open to general public</string>
+    <string name="quest_parking_access_private2">Not open to general public</string>
+    <string name="quest_parking_access_customers2">Open only to people visiting a specific place (e.g. customers)</string>
```
smichel17, line 565: One or both of these two should include the keyword *reserved*.

smichel17 (changes requested):
I have been confused before, when answering this quest. However, it's mostly not the fault of the quest— in many situations (when it is not signed), I just don't know what the rules actually are (who maintains the parking lot & whether non-customers are allowed to park there).

I am also not sure I fully understand the distinction between `private` and `customers`. Perhaps you could share a more technical definition, and I can try to come up with a good phrasing for that?

app/src/main/res/values/strings.xml
matkoniecz: > I am also not sure I fully understand the distinction between private and customers. Perhaps you could share a more technical definition, and I can try to come up with a good phrasing for that?

https://wiki.openstreetmap.org/wiki/Tag:access%3Dcustomers

https://wiki.openstreetmap.org/wiki/Tag:access%3Dprivate

`amenity=parking access=private` - "only for residents"

`amenity=parking access=customers` - "open only to people who are shopping at specific location (or meeting lawyer or picking up children from school or leaving shoes for repair)"

westnordost: "private" can also mean that they are rented, doesn't need to be residents. But there will be a sign. In any case, not relevant for the public.

smichel17: > private parking may have public access (say, paid parking near airport on private grounds)

I think you mean "privately-owned parking", otherwise this seems like a contradiction. So, to clarify, paid parking near an airport could be `access=public` _if_ you are allowed to park there even if you are not flying anywhere.

<details><summary>By the way, this is also a good example of much the most confusing situation, which I mentioned above.</summary>

- On one hand, the parking is clearly *primarily* for people using the airport.
- On the other, it *may* be available for public use.

*Sometimes* "for customers only" is signed, which makes it clear. When there is no sign, I usually skip the question because I'm not sure what the laws are.

Aside: while I don't feel comfortable tagging these places as `access=public`, I would usually not hesitate to park there myself, even if I was not a customer, since that restriction is rarely enforced, especially when not signed.
</details>

> publicly owned parking may have private access (parking where just president of city is allowed to park)

So, correct me if I am wrong— it seems like the dividing line is:

- `access=private` → It is reserved for some specific persons
- `access=customers` → It is open to any person, but only under some conditions (e.g. intent to shop at a certain place)

smichel17: Assuming ^that understanding is correct, I think we should also swap the order in the form, so it's

- public
- customers
- private

That way it's in order from least to most restrictive.

matkoniecz: > That way it's in order from least to most restrictive.

Done

> access=customers → It is open to any person, but only under some conditions (e.g. intent to shop at a certain place)

Yes, though I would also it describe as in general possible to general public (so "only for president of a country" would be `access=private`)

smichel17: > "only for president of a country" would be `access=private`

Right, because one can't simply decide to become the president. Same thing would apply to "only for employees" → `access=private`; a random member of the public can't simply decide to become an employee; it requires an agreement *in advance*.

Okay, so I think I understand now. I still need to think more about how to phrase this clearly and succinctly.

matkoniecz: Though note that for example school parking that can be used by parents picking up children I would mark as `access=customers` Despite that you cannot decide one day to have child and go sign up for specific school.

>I still need to think more about how to phrase this clearly and succinctly.

Let me know if you have any better ideas! I am not entirely happy about wording here but existing has bigger problems and this seems to be a clear improvement to me.

But likely it can be even better.

mnalis:
I'm OK with `customers` => `Open only to people visiting a specific place (e.g. customers)`

For `private`, it is currently too similar to `yes` text, so maybe `Forbidden to general public` would fit better? (or is it `Forbidden for general public`?)

Also of note, `yes` => `Open to general public` in this PR seems to have lost the explanation `(with or without fee)`, which I remember I had found useful when I was starting with SC for parkings, so others might have too, so I'd vote for reinstating that part if possible.

matkoniecz:
> seems to have lost the explanation (with or without fee), 

It was moved to the quest title.

app/src/main/res/values/strings.xml
smichel17: I think that's a great edge case. You could argue that a friend of the family might pick the child up one day, but I think that's still similar to `private` (more like a driveway— the parent would give permission in advance).

I can think of other reasons to tag `customers`. In my area, at least, schools are often used as voting buildings (in the gymnaseum or cafeteria). Also, I'm pretty sure some of the facilities (pool, tennis court, track, soccer field) are open to other people at least some of the time.

But, as a thought experiment, let's say they weren't and the only use for the school parking lot is parking for teachers or a parent picking up their child. Would you still tag it `access=customers`? If so, why?

app/src/main/res/values/strings.xml
```diff
@@ -559,10 +559,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
...
     <string name="quest_parking_fee_title">Do you have to pay to park here?</string>
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
-    <string name="quest_parking_access_title">Is it restricted who may park here?</string>
-    <string name="quest_parking_access_private">It is private</string>
-    <string name="quest_parking_access_yes">It is open to the public (with or without fee)</string>
-    <string name="quest_parking_access_customers">It is only for customers</string>
+    <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
+    <string name="quest_access_yes">Open to general public</string>
+    <string name="quest_access_customers">Open only to people visiting a specific place (e.g. customers)</string>
+    <string name="quest_access_private">Not open to general public</string>
```
smichel17, line 565:
```suggestion
    <string name="quest_access_private">Reserved for a specific person(s) (e.g. employees) </string>
```
Here's a start

westnordost: A bit long and awkward because it mixes singular and plural ("for a specific person(s)"). Also, it sounds it doesn't really fit to the classic "really" private parking on private property because "reserved" kind of implies that it is located on publicly accesible space but that parking in particular is reserved for xyz, e.g. a parking space being reserved for handicapped persons.

IMO "it is private" is better. But maybe "Reserved / Rented / Private" is fine too?

smichel17: Hm, good point about handicapped spots.

> private parking on private property

Is the parking quest asked for these?

westnordost: > Is the parking quest asked for these?

Wat? To find this out is the point of this quest 🤣

smichel17: I guess I haven't come across anything that is `access=private` besides parking lots marked "for employees only", so I wasn't sure if there was some additional filtering which happened to filter out driveways, etc.

smichel17: I guess in a city there's "parking along the street, by permit only"

app/src/main/res/values/strings.xml
matkoniecz: > Would you still tag it access=customers? If so, why?

Mostly because it fits well "in general not allowed to park there, but general public using this place would be allowed to use it".

Though I admit that I would mark "employee-only" parking near factory as `access=private`

It may be worth discussing on tagging mailing list or OSM wiki and maybe inventing new values. (if anyone wants to do this: feel free, I am unlikely to do this)

app/src/main/res/values/strings.xml
smichel17: Closing this thread because I want to make modifications to multiple strings

app/src/main/res/values/strings.xml
```diff
@@ -559,10 +559,10 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_religion_for_wayside_shrine_title">What religion is represented at this shrine?</string>
     <string name="quest_parking_fee_title">Do you have to pay to park here?</string>
     <string name="quest_surface_value_unhewn_cobblestone">Unhewn cobblestone</string>
-    <string name="quest_parking_access_title">Is it restricted who may park here?</string>
-    <string name="quest_parking_access_private">It is private</string>
-    <string name="quest_parking_access_yes">It is open to the public (with or without fee)</string>
-    <string name="quest_parking_access_customers">It is only for customers</string>
+    <string name="quest_parking_access_title2">Who may park here? Parking may be free or paid.</string>
```
smichel17, line 562: I noticed the [wiki `key:access` page](https://wiki.openstreetmap.org/wiki/Key:access) specifies,

> Access values describe **legal** permissions/restrictions […] It does not describe common or typical use, even if signage is generally ignored.

I think this phrasing conveys that more clearly:

```suggestion
    <string name="quest_parking_access_title2">Who is allowed to park here? Parking may be free or paid.</string>
```

app/src/main/res/values/strings.xml
smichel17, line 565: - Changed all the answers to remove the verbs ("Open to"), to match the updated question phrasing.
    - This gives us some extra characters to work with
- Adapted `=private` phrasing from the [wiki `key:access` page](https://wiki.openstreetmap.org/wiki/Key:access#List_of_possible_values) ("Only with individual permission"), which I like.
- Use the word *Any* to describe `=customers`; move "only" into the parentheses, so it's still present in the answer. That's the language we've been using to describe the tag here ("Anyone, but only if they're using the associated facility"); no need to complicate things.

#### Who is allowed to park here? Parking may be free or paid.

```suggestion
    <string name="quest_access_yes">Any member of the general public</string>
    <string name="quest_access_customers">Any person visiting a specific place (e.g. customers only)</string>
    <string name="quest_access_private">Only individual(s) with permission</string>
```

smichel17 (changes requested):
There may be some additional room for improvements, but I think this is pretty good (better than my previous suggestion).

app/src/main/res/values/strings.xml
smichel17: I am also unlikely :D. From the [wiki `key:access` page](https://wiki.openstreetmap.org/wiki/Key:access), it looks like this tag has a history of being confusing (many deprecated values).

In any case, I think I understand the tag better now, and have made updated suggestions below, so I'll resolve this.

westnordost:
What is the status of this ticket?

matkoniecz:
I was unsure which wording would be better and hoped for consensus of English speakers.

On review @smichel17 version seems nicer than mine so I applied it.

Noone protested that it is a poor idea so it should be ready for merging? I would welcome at least one person reviewing it though.

FloEdelmann (approved):
LGTM (but I'm not a native speaker)

smichel17:
> I was unsure which wording would be better and hoped for consensus of English speakers.

For the future, it would be good to request a review. Even when watching everything, it's sometimes easy to miss that your input would be useful, or required… even on my [own PRs](https://github.com/streetcomplete/StreetComplete/pull/3425#discussion_r744305287)!

PR #3474, opened 2021-10-30
Fix note form positioning bug

### LINKED ISSUE #3405: Adding a Note with picture breaks positions of buttons
**How to Reproduce**
- long press on a screen to add a OSM Note
- add picture to note
- confirm `OK`
- see that `undo`, `gps`, `+`, `-` buttons, as well as contributors note is move to the middle of the screen  (instead of remaining at the bottom of the screen)

https://user-images.githubusercontent.com/156656/137830092-8b556533-cdaf-47a4-8e3a-86a009818c4b.mp4

Workaround: Entering the `Undo` / `show edit history` and exiting restores normal positions of buttons. 

**Versions affected**
Android 10, SC `v36.0` (f-droid upstream).

smichel17:
Fixes #3405. Also renames a `isAnimating` to `isKeyboardAnimating`, since that's what it actually tracks.

Context in the commit, mostly copied from https://github.com/streetcomplete/StreetComplete/issues/3405#issuecomment-955124634


westnordost:
Very very cool! Thank you for this effort and the fix! I noticed the forth and back between mnalis and you, it looked very productive. Without this, who knows how long this bug would have gone unfixed!

PR #3477, opened 2021-10-31
remove wheelchair/toilets on shop owner change; add new keys

mnalis:
Wheelchair, toilets and related keys might change when shop owners change (usually by adding missing features). 

As requested on https://github.com/streetcomplete/StreetComplete/pull/3278#issuecomment-947618283

Also, few new keys have appeared at Taginfo, so they were categorized.


matkoniecz (approved):

FloEdelmann:
For reference, these were the upstream commits: https://github.com/mnalis/StreetComplete-taginfo-categorize/commit/93e652a0b868d6db745d37d50c34c07368900d74 and https://github.com/mnalis/StreetComplete-taginfo-categorize/commit/6818593ca0e8ea7467985d2aee15c33752a23f4b

PR #3480, opened 2021-11-02
Show nearby related POIs

### LINKED ISSUE #2354: Show other nearby objects of the same type when answering quests

### LINKED ISSUE #3338: Render power lines on the map
**Use case**
When doing the power pole material quest, the quest icons have no reference, they just float in a void, often in a forest or grass landuse. The lack of spatial awareness becomes an issue when some poles are already tagged (could be earlier quest actions), multiple power lines are close to one another and/or other spatial features are missing. Then you become unsure if you are tagging the right pole, or the next one in line.

To illustrate, this is my current view:
image

While this can actually be this situation in practice:
image

**Proposed Solution**
Render the power lines and power poles on the map.
Similar to how other quests relate to rendered objects, the power pole material quest would benefit if the objects are visualized.
Power lines and power poles are already mapped on the default Openstreetmap.org website, using simple lines and small squares/circles.

image

More general, I think it is valuable to render the nodes and ways to which the quests relate.

westnordost:
Fixes #3338, fixes #2354

This PR adds functionality that quests can highlight map data in its vicinity that is helpful / necessary for context of the quest.

Examples:
- bicycle parking capacity: It would be helpful to know if the bicycle parking just a few meters beside the shown point is mapped as an own bicycle parking or there is just one POI for all of it
- operator of an ATM: If there are several right next to each other from different operators, it would be nice to see which one of those is meant
- openiing hours, shop vacancies: If the shop doesn't exist anymore, it would be helpful to know which shop is left and right of the shop for which the quest is asked so that one knows which one exactly is meant
- power poles: if there is more than one power line, it would be good to see the actual power lines so it is easier to locate which one is meant


westnordost:
Code review and UX testing is welcome

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -819,6 +831,29 @@ class MainFragment : Fragment(R.layout.fragment_main),
...
+        }
+
+        viewLifecycleScope.launch {
+            val elements = withContext(Dispatchers.IO) { questType.getHighlightedElements(element, ::getMapData) }
+            for (e in elements) {
+                // don't highlight "this" element
+                if (element == e) continue
+                val geometry = mapData?.getGeometry(e.type, e.id) ?: continue
+                val icon = getPinIcon(e.tags)
+                putMarkerForCurrentQuest(geometry, icon, e.tags["name"] ?: e.tags["brand"])
```
FloEdelmann, line 852: Maybe use the `getNameOrBrandOrOperatorOrRef` function from `QuestUtil.kt`?

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -113,6 +115,9 @@ class AddPlaceName(
     override fun isApplicableTo(element: Element): Boolean =
         filter.matches(element) && hasFeatureName(element.tags)
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways, relations with " + isKindOfShopExpression())
```
FloEdelmann, line 119: Isn't this confusing when e.g. the name quest is open for a cemetery and all nearby shops are highlighted? (same for the opening hours quest)

app/src/main/java/de/westnordost/streetcomplete/quests/powerpoles_material/AddPowerPolesMaterial.kt
```diff
@@ -20,6 +23,15 @@ class AddPowerPolesMaterial : OsmFilterQuestType<PowerPolesMaterial>() {
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_powerPolesMaterial_title
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes with power = pole or man_made = utility_pole") +
+        // and also show the (power) lines themselves
+        getMapData().filter("ways with power ~ line|minor_line or communication = line or telecom = line")
```
FloEdelmann, line 29: This loops the map data twice, but I guess `nodes with … or ways with …` is not valid ElementFilterExpression syntax?

app/src/main/java/de/westnordost/streetcomplete/quests/clothing_bin_operator/AddClothingBinOperator.kt
```diff
@@ -43,6 +44,9 @@ class AddClothingBinOperator : OsmElementQuestType<String> {
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_clothes_container_operator_title
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes with amenity = recycling and recycling_type = container")
```
FloEdelmann, line 48: To stay consistent with the recycling quests:

```suggestion
        getMapData().filter("nodes with amenity = recycling")
```

I'm not sure about this though.

app/src/main/java/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterials.kt
```diff
@@ -11,11 +11,9 @@ import de.westnordost.streetcomplete.data.meta.hasCheckDateForKey
...
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapEntryModify
 import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
 import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.CITIZEN
-import de.westnordost.streetcomplete.util.LatLonRaster
-import de.westnordost.streetcomplete.util.distanceTo
-import de.westnordost.streetcomplete.util.enclosingBoundingBox
 
 class AddRecyclingContainerMaterials : OsmElementQuestType<RecyclingContainerMaterialsAnswer> {
```
FloEdelmann, line 18: Can this now be converted to a `OsmFilterQuestType`?

res/graphics/pins/bollard.svg
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
FloEdelmann, line 1: This SVG needs optimizing

FloEdelmann (changes requested):
The code looks good, apart from the few minor nits below. I haven't tested the UX yet.

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
westnordost: That would be shops that are as close as 30m from the center of the cemetary though. I thought about it. The name quest basically contains shops + larger things like parks, cemeteries etc.
For shops (etc) we want to have this in any case, for the larger things there is not really any need, but I think it will also not confuse because in most cases, there is no shop or anything anyway that close by.

app/src/main/java/de/westnordost/streetcomplete/quests/clothing_bin_operator/AddClothingBinOperator.kt
westnordost: 🤷  ok

app/src/main/java/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterials.kt
westnordost: No because of `Element.hasAnyRecyclingMaterials`

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
westnordost: We **could** do something like `return if(element "is a kind of shop") mapData."elements that are a kind of shop" else null` but not sure if it is worth it

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapData+ElementFilterExpression=❤.kt
```diff
@@ -0,0 +1,26 @@
+package de.westnordost.streetcomplete.data.osm.mapdata
```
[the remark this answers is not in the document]
westnordost: 🤷 I'll rename it

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
westnordost: Where would that be an improvement? And would it potentially be confusing for some elements?

westnordost:
Thanks for the review!

westnordost:
https://user-images.githubusercontent.com/4661658/139920925-025238f0-0ca2-4e43-a16f-46e1d8a09430.mp4

https://user-images.githubusercontent.com/4661658/139921186-c4d16ded-3999-4c67-8e8b-ab62dfbb2542.mp4

https://user-images.githubusercontent.com/4661658/139921424-c99872e9-8bf7-4752-84a5-326865c0f865.mp4

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
FloEdelmann: I'm not sure where that title is displayed at all :sweat_smile:

FloEdelmann:
In your first video, when opening the quest for Kiosk 104, it looks like the quest is actually for SuperBioMarkt, because its whole geometry is highlighted.

Is it possible to hide all other quest pins while the quest answer sheet is open? Then all highlighted geometries look completely different (violet background + icon) compared to the current quest geometry (orange background + quest marker pin).

That'd also make the example in your third video look better: There, a quest pin with a bike parking icon is shown above an area with a bike parking icon. If the quest pin is hidden, that eliminates the duplication.

westnordost:
Yes it is possible. If you want to test it yourself, add

```diff
  fun startFocusQuest(quest: Quest, offset: RectF) {
+ pinMode = QuestsMapFragment.PinMode.NONE`
  geometryMapComponent?.beginFocusGeometry(quest.geometry, offset)
```

to `QuestsMapFragment`. (Probably not the proper place, but quick and dirty for testing)

I agree that it is much more clear/clean this way, but it results in a visual jerk every time the quest form is open and closed. It is probably possible to reduce this jerk, **maybe** even remove it altogether, not sure. It will be some effort though to cleanly refactor the code so that hide/show is faster.

westnordost:
I checked - hiding the quest pins can not be faster. And it's actually not **that** slow. Showing the quest pins again can be faster but only with more effort, i.e. more working around missing features in tangram-es.

The main issue is that it is not possible in tangram-es to switch the visibility of a layer. You can only clear it. So what is done when "hiding" any layer is to actually clear it. When it should be shown again, it is built up anew from data fetched from the database.

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
peternewman: > I'm not sure where that title is displayed at all sweat_smile

Presumably on the shops as in the first example video?

> Where would that be an improvement?

Postboxes for example, would get no name from the current code, but operator + ref from:
https://github.com/streetcomplete/StreetComplete/blob/0857679a146abaf474f051f8e8d12225d9a10ff3/app/src/main/java/de/westnordost/streetcomplete/quests/QuestUtil.kt#L68-L83

> And would it potentially be confusing for some elements?

Theoretically maybe, but then it gives just one place to fine-tune that logic, rather than having different behaviour in here compared to in a quest.

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapData+ElementFilterExpression=❤.kt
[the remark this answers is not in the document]
peternewman: This looks to still have the heart to me...

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
westnordost: > Postboxes for example, would get no name from the current code, but operator + ref from

Not sure if this would be an improvement though. Then every single post box in Germany potentially has something like "Deutsche Post AG Huberstraße Mitte" displayed. (The "refs" if, stated at all, are actually streets/addresses/address descriptions)

If there are two post boxes directly next to each other, one of "xyz mail" and the other of "royal mail" or whatever, it should already be clear from the location which is which (which is right of which)

westnordost:
Will look on how it will be possible to change the "view mode" when the quest form is open.

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
peternewman: > Not sure if this would be an improvement though. Then every single post box in Germany potentially has something like "Deutsche Post AG Huberstraße Mitte" displayed. (The "refs" if, stated at all, are actually streets/addresses/address descriptions)

Ah, in the UK they are fewer than 10 characters for the ref as a genuine reference number not a location.

What about if we add a boolean brief option to getNameOrBrandOrOperatorOrRef, for things like this, you'd set brief=true and it might only add the ref if it's less than 10 characters, and otherwise just show the operator, whereas for the quest itself, you clearly want it regardless so you'd set brief=false and it would always show it.

> If there are two post boxes directly next to each other, one of "xyz mail" and the other of "royal mail" or whatever, it should already be clear from the location which is which (which is right of which)

We don't get that specific issue in the UK as the postboxes are essentially all Royal Mail, but you'll often get two Royal Mail postboxes next to each other (with different refs). At least once, myself or someone else has got the refs offset and tagged against the wrong boxes, unpicking that in general is a real pain to do easily, but could be quickly spotted using this.

More generally in the UK I've seen a lot of stuff like benches and bins mapped on the wrong side of a path, or the wrong way round. If a post box has been removed, or two phone boxes have become one (i.e. one has gone), the more info referencing which one is which the better and the greater the chance of me making the right choice. I don't want to delete the left one if the remaining one is actually BT and the one I've deleted is the BT one (if that makes sense).

TL;DR, you should at least add operator to your chained selection list, at which point you may as well just use the existing function.

westnordost: I leave it as is for now

westnordost:
Notes:
- also for areas, should show outline so it is clear where one feature begins and ends
- should only show nearby objects on same level (especially for shops) - if specified

westnordost:
so, left to do:

- hide/show other quest pins while quest is open
- optionally can remove logic handling clicking on other quests while this quest form is open
- optionally can remove the extra highlighting of the tapped marker (the blue circle around the highlighted pin)

westnordost:
Note: 

> should only show nearby objects on same level (especially for shops) - if specified

maybe also check for `level:ref` and/or `addr:floor`

westnordost:
https://user-images.githubusercontent.com/4661658/141798205-6269c30f-d8e8-4727-ac94-8a589411144e.mp4

westnordost:
Shout out to @matteblair: Thank for enabling this fast-switching between displayed layers in the new version of tangram-es!

app/src/main/java/de/westnordost/streetcomplete/map/components/PinsMapComponent.kt
```diff
@@ -1,50 +1,23 @@
...
-    fun clear() {
-        clearPins()
-        clearSelectedPins()
-    }
-
-    /* ------------------------------------------  Pins ----------------------------------------- */
+    /** Shows/hids the pins */
+    var isVisible: Boolean
+        get() = pinsLayer.visible
+        set(value) { pinsLayer.visible = value }
```
smichel17, line 17: Note: you could also do this (https://kotlinlang.org/docs/delegated-properties.html#delegating-to-another-property):

```suggestion
    var isVisible: Boolean by pinsLayer::visible
```
I'm not doing a full review, just a quick skim of a couple commits.

westnordost: Interesting, didn't know that. But it should be consistent among the codebase, so this should be done in a separate PR. there are more places other than that where it is done

PR #3482, opened 2021-11-03
allow to mark that shop was replaced by dentist/doctor

matkoniecz:
PR as it redesigns what @westnordost created.

Change is motivated by https://www.openstreetmap.org/changeset/113334033 (POI at ground level of apartment tower - pharmacy replaced by doctor).

While doctors/dentists are relatively rare in malls and dedicated retail areas (but they are sometimes appearing there), but sometimes shop needs to be retagged to dentist/doctor.

As I understand there are two effects:

- "What's here now instead?" will now include dentist/doctor
- dentist/doctor objects in quest with `isReplaceShopEnabled` will now get this "What's here now instead?" menu

Both seem desirable to me. Even outside retail area it would be nice to retag doctor/dentist into another doctor/dentist.

Though I am not getting why it is also excluding `craft` and offices in general.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
@@ -95,7 +95,9 @@ val KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED = listOf(
  *  - no larger or purpose-built things like malls, cinemas, theatres, car washes, fuel stations,
  *    museums, galleries, zoos, aquariums, bowling alleys...
  *  - no things that are usually not found in normal retail shop areas but in offices:
- *    clinics, doctors, fitness centers, dental technicians...
+ *    clinics, fitness centers
+ *  - dentists, doctors are also included as they may appear in such places, depending
+ *    on location
  *  - nothing that is rather located in an industrial estate like car repair and other types
  *    of workshops (most craft=* other than those where people go to have something repaired or so)
  *
@@ -116,6 +118,7 @@ fun isKindOfShopExpression(prefix: String? = null): String {
                 "bank", "bureau_de_change", "money_transfer", "post_office", "internet_cafe",
                 "pharmacy",
                 "driving_school",
+                "doctors", "dentist",
             ),
             "leisure" to arrayOf(
                 "amusement_arcade", "adult_gaming_centre", "tanning_salon",
@@ -126,6 +129,9 @@ fun isKindOfShopExpression(prefix: String? = null): String {
             "craft" to arrayOf(
                 "shoemaker", "tailor", "photographer", "watchmaker", "optician",
                 "electronics_repair", "key_cutter",
+            ),
+            "healthcare" to arrayOf(
+                "doctor",
             )
         ).map { p + it.key + " ~ " + it.value.joinToString("|") }.joinToString("\n  or ") + "\n"
         ).trimIndent()


westnordost:
Well the reason why dentists/doctors etc were excluded is that they often (in my area: Almost exclusively) are **not** located in storefronts which could at will be redesigned to house another shop but in the same kind of spaces where offices are located. 

I.e. usually not on ground floor but renting an office (or even a usually residential) space somewhere in an (office) building.

And since it makes sense to do that instead of renting a storefront (cheaper by m², no walk-in customers but based on appointments), I don't think this observation is exclusive to the areas I know.

westnordost:
So, as a consequence, if a doctor moves out from such a space, the chance that the next tenant will be some kind of shop is low. It may be some company office, or converted into a flat.

IMO marking a (shop) as vacant is only valid for kind of shops (storefronts), not for office spaces.

matkoniecz:
> I.e. usually not on ground floor but renting an office (or even a usually residential) space somewhere in an (office) building.

All dentists that I ever visited in Poland had office at ground level - two in places that could be easily replaced by shops, one in place housing multiple doctor offices and would be replaced by another dentist/doctor, one had office in his house that on stopping service ceased to be a retail location.

For example here "Stomatologia" is the dentist sign, toward left one may see shops in an equivalent retail space.
screen08
https://www.openstreetmap.org/#map=19/50.09098/19.97566
screen09

> And since it makes sense to do that instead of renting a storefront (cheaper by m², no walk-in customers but based on appointments), I don't think this observation is exclusive to the areas I know.

No idea about economics and root cause for that situation.

smichel17:
In my somewhat limited experience, @westnordost's observation is correct in cities, but not in more rural areas. Some of my extended family are doctors, I'll ask them.

westnordost:
The issue I raised is not about how common it is. It is about that there is a reasonable chance that a doctor/dentist is not in a storefront. Meaning that in these cases, tagging `shop=vacant` and similar would be considered wrong. Furthermore, many such places that are not on the ground floor do not have a level tagged because it is not a common thing to do outside of malls etc.
So, users asked about whether X is still vacant, and that X was actually a doctors office on the 2nd floor, he will most likely give a wrong answer.

matkoniecz:
And tagging `disued:office=yes` would break for ones in retail locations...

Maybe separate list for things that can be given this menu and separate list of things selectable as option in this menu would be OK?

westnordost:
> And tagging disued:office=yes would break for ones in retail locations...

I don't understand

matkoniecz:
> I don't understand

In office spaces replacing vacant `amenity=doctor` with `disused:shop=yes` - as you mentioned - would not work well.

Using `disued:office=yes` would be fine, but that in turn would fail for retail locations...

Maybe deleting nodes would be fine in such cases? And request note creation for areas?

Either way as it stands this PR is not viable.

peternewman:
Are there some compromises around this that would fit, or will they all have the same issue (e.g. beauty clinic, (pop-up) Covid testing etc)?

westnordost:
I can't think of any.

Anyway, in case the "replace shop" functionality is not available, the user can leave a note with a photo. I.e. "Dentist XYZ does not exist, but there is... [photo]".

This should in any case be most helpful, because the rest of the businesses within the building could then also be checked if they are still current.

peternewman:
Thanks, I didn't explain very well, but mine was just tagging of a "shop", rather than replacing it. Again it was a standalone unit on a high street rather than in an office block or other building.

PR #3483, opened 2021-11-03
Better hint (review of English welcomed!)

### LINKED ISSUE #3227: Provide feedback to user if no object type matches in "what is here now" dialog
**Use case**
Would fix root cause of confusion from https://github.com/streetcomplete/StreetComplete/issues/3180

**Proposed Solution**
Show something like

> enter shop type, not shop name - so "bakery", not "Ancient Oak"

if no object type matches

matkoniecz:
Fixes #3227 in a modified way.

Showing this error/hint while typing would be annoying. And user who is new to this menu and types things like "Jolly Good" or "Polska Chata" will not be expecting dropdown to appear - they will type and press OK button.

Also, I have not figured out how to do this anyway, but I am pretty sure that showing error in normal operation - as I proposed - would be really annoying.

screen02

original text:

> Don’t know what this is. Consider leaving a note.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error2">"Not recognized. Try entering the type of place, for example 'bakery'. Otherwise, consider leaving a note."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Not recognized. Try entering the type of place, for example “bakery”. Otherwise, consider leaving a note."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What’s here now?</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway at all</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure_explanation">As usual, tap on one side of the illustration, select a proper answer and do the same for the other side.\n\nNote all the available options: some cases are not obvious, like a oneway road that allows cyclists in both directions!</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error2">"Enter the type of object, for example 'bakery', not the name.\nLeave note if that fails."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Enter only the type of object, for example 'bakery', not the name.\nLeave note if that fails."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What’s here now?</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway at all</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure_explanation">As usual, tap on one side of the illustration, select a proper answer and do the same for the other side.\n\nNote all the available options: some cases are not obvious, like a oneway road that allows cyclists in both directions!</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error2">"Enter only the type of object, for example 'bakery', not the name.\nLeave note if that fails."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Enter only the type of object, for example 'bakery', not the name.\nLeave a note if that fails."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What’s here now?</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway at all</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure_explanation">As usual, tap on one side of the illustration, select a proper answer and do the same for the other side.\n\nNote all the available options: some cases are not obvious, like a oneway road that allows cyclists in both directions!</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error2">"Enter only the type of object, for example 'bakery', not the name.\nLeave a note if that fails."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Not recognized. Try entering the type of place, for example 'bakery'. Otherwise, consider leaving a note."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What’s here now?</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway at all</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure_explanation">As usual, tap on one side of the illustration, select a proper answer and do the same for the other side.\n\nNote all the available options: some cases are not obvious, like a oneway road that allows cyclists in both directions!</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error">"Don’t know what this is. Consider leaving a note."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Enter the type of object, for example 'bakery', not the name.\nLeave note if that fails."</string>
     <string name="quest_shop_vacant_type_title">This shop has been vacant. What’s here now?</string>
     <string name="quest_cycleway_answer_no_bicycle_infrastructure">No cycleway at all</string>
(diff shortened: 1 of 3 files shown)


mnalis:
Bot an native speaker, but I think that should be _"[...]\nLeave **a** note if that fails"_

smichel17:
I responded to the approach overall back in #3227. But, to give feedback on this solution…

> I think that should be _"[...]\nLeave **a** note if that fails"_

+1. Also, it's unclear whether it's an instruction to the user, or a typo of "[The app] Leaves a note if that fails" Or "[You should] leave a note if that fails'. Since I'm quite sure it's the latter, I would phrase it as "If that fails, [please] leave a note [instead]." [Bracket words can be left out for brevity. If only one can be kept, keep *please*].

Even better would be, "If StreetComplete cannot validate it, please leave a note instead." (Or, if you follow my suggestion in #3227 to move "type" into the question, the whole thing could be something like "Unknown shop type. Please [try a different variation or] leave a note instead.")

app/src/main/res/values/strings.xml
```diff
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error">"Don’t know what this is. Consider leaving a note."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Enter only the type of object, for example 'bakery', not the name.\nLeave note if that fails."</string>
```
smichel17, line 1013: I just noticed the bit about what it was before. So I think all you need is

```suggestion
    <string name="quest_shop_gone_replaced_answer_error2">"Unknown shop type. Consider leaving a note."</string>
```
I'm not sure if this is different enough to warrant a new string, or just marking the translation as fuzzy in POEditor.

mnalis: I think one of the main ideas was to explicitly **emphasize** to the user that they should  **not** enter the shop **name** in that field, **and** to give helpful **example** of what they need to enter. 

IMHO reducing all that to `Unlisted shop type` kinda misses those points (by making the explanation only implicit again, and also losing the example)...

matkoniecz: Yes, main problem that I noticed is that people try to enter given name ("Jolly good time" / "Lidl" / "Żabka" - not shop type such as "convenience" or "supermarket").

Old feedback was not mentioning this problem specifically and hint in background of the text field was not sufficient.

matkoniecz:
> Not a native speaker, but I think that should be "[...]\nLeave a note if that fails"

Done

BalooUriza:
Perhaps this could be approached by the way the brand presets in iD and
JOSM do it?  Like, if you know the name of a common brand, you can fill
that in and it fills out the rest?

matkoniecz:
> Like, if you know the name of a common brand, you can fill that in and it fills out the rest?

Maybe this is worth adding but it would not solve this as it would only capture brands with multiple shops.

BalooUriza:
On Thu, Nov 4, 2021 at 10:27 AM Mateusz Konieczny ***@***.***>
wrote:

> Like, if you know the name of a common brand, you can fill that in and it
> fills out the rest?
>
> Maybe this is worth adding but it would not solve this as it would only
> capture brands with multiple shops.
>
I feel like that's solving 90% of this problem?  Odds are someone familiar
with Safeway or Lidl isn't going to go "hey, that's a supermarket" but
rather "hey, that's a Safeway", and someone entering that as an answer
expects that the appropriate metadata will automagically happen.

smichel17:
> I feel like that's solving 90% of this problem?

Only if 90% of the shops are chains. It doesn't help at all when the store is something local that iD doesn't know about.
If I answer, "This isn't *Berkshire Yogurt* any more, now it's [*Nourish*](https://www.openstreetmap.org/node/4239455771)", StreetComplete is not going to have the slightest clue that means it changed from a frozen yogurt shop to a cafe.

app/src/main/res/values/strings.xml
smichel17: I am aware that's the main problem. Compare my suggestion to the original, before this PR:

```diff
- Don’t know what this is. Consider leaving a note
+ Unlisted shop type. Consider leaving a note.
```

I think they both fix the problem, which is making it obvious what the problem is. Mine actually gives more detail, because it hints that StreetComplete only recognizes a fixed list of shops types (it's possible that the problem is because the user entered something overly specific, like "Wellness bar and cafe", which isn't in the list. @matkoniecz's suggestion wouldn't make it clear why that is failing.

I agree that we could add a few more words to mine in order to make it clearer, but I think the content is better.

mnalis: I'm afraid users will miss it in just `Unlisted shop type`. 
Example is crucial in my opinion, as is explicit instruction not to enter "name"... 
Maybe something along the lines of 
`Unlisted shop type; please enter a type like "bakery" and not a shop name, or leave a note if unsure.`

westnordost:
I find this PR is better than anything proposed in #3227. 

1. Let's focus on making the string displayed as clear as possible.
2. Add support for making it possible to specify brand names. I am pretty sure it is already, there is no good reason to not do this.

westnordost:
> Enter only the type of object, for example 'bakery', not the name.\nLeave note if that fails

My try:

_"Try enter the kind of place instead of its name, for example 'bakery'. If it is still not found, consider leaving a note."_

smichel17:
I like @westnordost's suggestion best so far. Minor tweaks:

*Not recognized. Try entering the type of place, for example 'bakery'. Otherwise, consider leaving a note.*

The punctuation (comma placement) around `place, for example 'bakery'` seems wrong, but I'm not quite sure how to fix it. @peternewman Do you agree? If so, any ideas to fix it?

@mnalis Do you think `place` causes confusion in this context? If so, then it can be swapped for `establishment`. If it's just hard to translate, perhaps a comment could be left for translators that "place" should be translated like "establishment" here. Or better(?): we could use `establishment` as the base string and `place` as an `en-US` localization.

peternewman:
> I like @westnordost's suggestion best so far. Minor tweaks:
> 
> _Not recognized. Try entering the type of place, for example 'bakery'. Otherwise, consider leaving a note._

LGTM!

> The punctuation (comma placement) around `place, for example 'bakery'` seems wrong, but I'm not quite sure how to fix it. @peternewman Do you agree? If so, any ideas to fix it?

It doesn't really to me, but I can't see a fix option aside from dropping the comma anyway.

matkoniecz:
I also like "Not recognized." as it seems better at expressing that app failed to recognize the type, while "Nothing found" seems less clear.

matkoniecz:
screen11

matkoniecz:
> Add support for making it possible to specify brand names. I am pretty sure it is already, there is no good reason to not do this.

User may be more confused if for example Lidl works and independent shop does not

But if in dropdown shop type (not shop name) would be shown, then it would likely work fine?

smichel17:
I don't have a strong opinion about whether brand names should be allowed. However, if they are allowed it should be done **before merging** this PR, because it should use **different phrasing**. I agree it's a good idea to do a **separate PR**, since this discussion is long enough already.

If brand names are allowed, the current phrasing is good. If not, then the question title should be changed to `What type of establishment is here now?`, and the warning text should `s/Try entering/Please enter/`.

<details><summary>If we allowed brand names I have another probably-too-clever idea</summary>

What if we searched the local database (OSM cache) for shops in the area and allowed autocompleting to those, copying their tags? This would allow catching local chains, *if* the user has downloaded that area.

The downsides are that (1) it's inconsistent, since it depends on what data is cached locally, and (2) I'm not totally sure if it is safe, or if we'd need to confirm the type with the user.
</details>

app/src/main/res/values/strings.xml
```diff
@@ -1010,7 +1010,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_shop_gone_vacant_answer">It’s vacant</string>
     <string name="quest_shop_gone_replaced_answer">A(n)…</string>
     <string name="quest_shop_gone_replaced_answer_hint">Type kind of place here</string>
-    <string name="quest_shop_gone_replaced_answer_error">"Don’t know what this is. Consider leaving a note."</string>
+    <string name="quest_shop_gone_replaced_answer_error2">"Not recognized. Try entering the type of place, for example 'bakery'. Otherwise, consider leaving a note."</string>
```
FloEdelmann, line 1013: To stay consistent with quotation marks in existing strings:

```suggestion
    <string name="quest_shop_gone_replaced_answer_error2">"Not recognized. Try entering the type of place, for example “bakery”. Otherwise, consider leaving a note."</string>
```

matkoniecz:
> However, if they are allowed it should be done before merging this PR, because it should use different phrasing.

I would keep the same - and treat brand names as synonyms. User should not be instructed to try using using brand name as NSI is quite incomplete (especially outside USA/Europe) and likely always will be.

FloEdelmann (changes requested):

FloEdelmann (approved):

westnordost:
The wording is fine.

"A McDonalds" works.
"A bakery" works.
"A Bäckerei Klausen" does not work (unless that would be a locally known **chain**)

I worded the hint deliberately to not say something like "do **not** input a name" but "try inputting a name" so to not exclude brand names.

westnordost:
According to the code, brand names should be included. Right now, I seem unable to input "McDonalds" or anything like that. Can anyone reproduce this?

peternewman:
> According to the code, brand names should be included. Right now, I seem unable to input "McDonalds" or anything like that. Can anyone reproduce this?

Agreed, both on this it's changed, and on the it's vacant quest, I can only find types of shops, not brands. I'm not sure I've ever seen them, but I guess you're saying it worked at some point? It also doesn't work on what is the name of this shop...

> Also, user may be confused when it immediately asks about name in the next quest.

Surely if I select Lidl it will at least tag the shop name from that (although personally I'd like it to tag the brand and the other stuff in NSI.

westnordost:
Ok, I will look into that then. As far as this PR is concerned, it's approved.

westnordost (approved):

matkoniecz:
> According to the code, brand names should be included. Right now, I seem unable to input "McDonalds" or anything like that. Can anyone reproduce this?

Yes, in fact reproduction with Lidl is used as an example ( https://github.com/streetcomplete/StreetComplete/pull/3483#issue-1043808327 ). 

And curiously, disabling filter in `getFeatures` have not allowed to find say Lidl (while enabled things like "Power pole").

It looks like brand data is not loaded at all (???).

smichel17:
> I worded the hint deliberately to not say something like "do not input a name" but "try inputting a name" so to not exclude brand names.

I noticed this, that's why I said the current phrasing is good **if** we include brand names. My suggestions in https://github.com/streetcomplete/StreetComplete/pull/3483#issuecomment-961784441 are only if we do not include brand names.

mnalis:
> I like @westnordost's suggestion best so far. Minor tweaks:
> 
> _Not recognized. Try entering the type of place, for example 'bakery'. Otherwise, consider leaving a note._
> 
> The punctuation (comma placement) around `place, for example 'bakery'` seems wrong, but I'm not quite sure how to 
fix it. @peternewman Do you agree? If so, any ideas to fix it?

`Try entering the type of place, for example 'bakery'` looks mostly OK to me . Usually I might consider using parentheses instead - ie. `Try entering the type of place (for example 'bakery')` - but it does not really look like an improvement to me in this case.

> @mnalis Do you think `place` causes confusion in this context? If so, then it can be swapped for `establishment`. If it's just hard to translate, perhaps a comment could be left for translators that "place" should be translated like "establishment" here. Or better(?): we could use `establishment` as the base string and `place` as an `en-US` localization.

Establishment looks like nice more precise tag (and would be more clear when translating, even if I would have same problem as finding adequate translation in Croatian -- but at least I'd know exactly what I was searching for!), but in this specific case (as it is followed immediately by `bakery` example) `type of place` is also good.

westnordost:
That brand names are not displayed is fixed. Root cause was that more recent versions of the name suggestion index stopped adding the property "suggestion: true" to each and every brand name preset. The osmfeatures library however was built on that, so I had to change a lot of things in that library.

PR #3487, opened 2021-11-04
Bicycle barrier

### LINKED ISSUE #3361: What kind of cycle barrier is this?
### General
Affected tag(s) to be modified/added: [cycle_barrier](https://wiki.openstreetmap.org/wiki/Key:cycle_barrier) 
Question asked: **What kind of cycle barrier is this?**

Basically similar to the question about bollard type.

### Checklist
  - Sometimes answerable from aerial imagery as well, but that doesn't necessarily preclude asking here.
  - `double` is perhaps the most common one, but there are still other options
- [?] 🕓 Applies to a reasonable number of map data (Worth the effort)
  - about 100k instances of `barrier=cycle_barrier`. Is that enough?

### Ideas for implementation

**Element selection:** Nodes on ways with `barrier=cycle_barrier`, but without `cycle_barrier=*`. Possibly similar to how bollards are currently selected for the »What kind of bollard is this?« quest.

**Metadata needed:** Unknown.

**Proposed UI:** Similar to the bollard quest, a selection of options, each with an image (the tag Wiki page has images).

matkoniecz:
fixes #3361

screen06

I am still planning to improve images but that can be done also after merging.


app/src/main/res/values/strings.xml
```diff
@@ -1088,6 +1094,7 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
     <string name="quest_traffic_calming_type_island">Island</string>
     <string name="quest_traffic_calming_type_chicane">Deliberate curve(s)</string>
     <string name="quest_traffic_calming_type_rumble_strip">Rumble strip(s)</string>
+>>>>>>> master
```
riQQ, line 1097: Left-over from merging

matkoniecz:
> Perhaps »diagonal« could use a different text. Something like »at an angle to the path« to emphasize what differentiates it from »double«.

> Why not "diagonal to the path" / "diagonal to the way"? It's more succinct.

Were proposed, not entirely sure about them.

app/src/main/res/values/strings.xml
matkoniecz: Thanks you so much!

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierType.kt
```diff
@@ -0,0 +1,31 @@
...
+    override val isDeleteElementEnabled = true
+
+    override val questTypeAchievements = listOf(CAR, PEDESTRIAN, BLIND, WHEELCHAIR, BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_bicycle_barrier_type_title
+
+    override fun createForm() = AddBicycleBarrierTypeForm()
+
+    override fun applyAnswerTo(answer: BicycleBarrierType, changes: StringMapChangesBuilder) {
+        changes.modify("cycle_barrier", answer.osmValue)
```
FloEdelmann, line 29: Does `modify` work here, when the element filter explicitly excludes elements with `cycle_barrier`?

```suggestion
        changes.add("cycle_barrier", answer.osmValue)
```

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierType.kt
```diff
@@ -0,0 +1,31 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.WHEELCHAIR
+
+class AddBicycleBarrierType : OsmFilterQuestType<BicycleBarrierType>() {
+
+    override val elementFilter = """
+        nodes with barrier=cycle_barrier and !cycle_barrier
+    """
+    override val commitMessage = "Add specific cycle barrier type"
+    override val wikiLink = "Key:cycle_barrier"
+    override val icon = R.drawable.ic_quest_barrier // TODO
```
matkoniecz, line 19: @westnordost Do you have an idea or energy for making an icon?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierTypeForm.kt
```diff
@@ -0,0 +1,23 @@
...
+import de.westnordost.streetcomplete.view.image_select.Item
+
+class AddBicycleBarrierTypeForm : AImageListQuestAnswerFragment<BicycleBarrierType, BicycleBarrierType>() {
+
+    override val items = listOf(
+        Item(SINGLE, R.drawable.barrier_bicycle_barrier_single, R.string.quest_barrier_bicycle_type_single),
+        Item(DOUBLE, R.drawable.barrier_bicycle_barrier_double, R.string.quest_barrier_bicycle_type_double),
+        Item(TRIPLE, R.drawable.barrier_bicycle_barrier_triple, R.string.quest_barrier_bicycle_type_multiple),
+        Item(DIAGONAL, R.drawable.barrier_bicycle_barrier_diagonal, R.string.quest_barrier_bicycle_type_diagonal),
+        Item(TILTED, R.drawable.barrier_bicycle_barrier_tilted, R.string.quest_barrier_bicycle_type_tilted),
```
FloEdelmann, line 15: I think that the images of `single`, `triple` and `diagonal` are a bit faint compared to `double` and `tilted`. Maybe adjust the contrast a bit for them?

app/src/main/res/values/strings.xml
```diff
@@ -1079,6 +1079,12 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
     <string name="quest_presets_selected">Selected</string>
     <string name="quest_presets_delete_message">"Delete preset “%s”? The quest selection and order for it will be irrevocably lost."</string>
     <string name="quest_picnicTableCover_title">Is this picnic table covered (protected from rain)?</string>
+    <string name="quest_bicycle_barrier_type_title">What type of bicycle barrier is this?</string>
+    <string name="quest_barrier_bicycle_type_diagonal">Diagonal</string>
```
FloEdelmann, line 1083: I agree with @riQQ (see https://github.com/streetcomplete/StreetComplete/issues/3361#issuecomment-961432327) that "Diagonal to the way"/"Diagonal to the path" would be better, as the `tilted` barrier is also "diagonal".

*(commenting here, so the discussion can be resolved)*

FloEdelmann (changes requested):
Looks good! Just some smaller issues.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierTypeForm.kt
matkoniecz: went through tweaking round - do you think it is OK now?
screen12

FloEdelmann: It's a lot better. IMO the contrast of `diagonal` could be increased a little more, and the contrast of `third` could still be increased by a lot – but maybe that distorts colors too much there, so a different image would be needed. In any case, this is not merge-blocking.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierType.kt
westnordost: Same as bollard but with a bike?

Or a front-view of a bicycle barrier like "|==|"?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierType.kt
```diff
@@ -0,0 +1,31 @@
...
+    override val isDeleteElementEnabled = true
+
+    override val questTypeAchievements = listOf(CAR, PEDESTRIAN, BLIND, WHEELCHAIR, BICYCLIST)
```
westnordost, line 22: How about just bicyclist? These are bicycle barriers after all.

matkoniecz: car, pedestrian seems pointless and I will remove it

but it is also useful for wheelchair/blind users (tripple chicane is annoying also for people on wheelchair and I suspect that it could be useful info for blind)

westnordost: ok

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierType.kt
westnordost: I added bicycle_barrier.svg

I was lazy and just used barrier.svg

I wanted to do something like 🚳 but it did not look good.

PR #3495, opened 2021-11-08
Request capitalisation in housename and notes

arrival-spring:
To save having to tap shift all the time

Housename: textCapWords - "request capitalization of the first character of every word"

Notes: textCapSentences - "request capitalization of the first character of every sentence"

(From https://developer.android.com/reference/android/widget/TextView)


westnordost:
Sound useful

PR #3515, opened 2021-11-18
ask about barrier on intersection of barrier line and path [ready for review, again]

### LINKED ISSUE #3372: What type of Stile/Gate/Gap is here?
### General
Affected tag(s) to be modified/added: [stile](https://wiki.openstreetmap.org/wiki/Key:stile)
Question asked: **What type of Stile/Gate/Gap is here?**

### Checklist

### Ideas for implementation

**Element selection:** 
Anywhere where a `highway=path/footway` intersects/crosses a `barrier=wall/fence/hedge/guard_rail`

For example this node (and the one above it):
https://www.openstreetmap.org/node/7287751590

**Metadata needed:** 
None

**Proposed UI:**
It can probably just chuck you into the existing stile quest, especially when gaps are added as mentioned in https://github.com/streetcomplete/StreetComplete/issues/3188#issue-973861400

matkoniecz:
fixes #3372

--------------------

What would be the good quest question in this quest?

both 

> What is where path goes through a barrier?

> What type of Stile/Gate/Gap is here?

> What is at intersection of path and barrier?

are unfortunate and seem clunky to me.

--------------------

Below are things that I implemented to my satisfaction (more or less) but I can be easily convinced to change that.

Is it a good idea to keep complex "path must actually cross barrier" logic? It was implemented already and will keep out some rare confusing cases.

Is it fine to duplicate this logic and helper function from crossing quest? If no where would be the best place to put it?

I plan to have both path quest and road quest here. For now just path is implemented, with shared code in a separate class.

(why separate quest for roads? Because it is typical to have gate for road and separately mapped path without a barrier along it, and separate quest logos and titles would be an useful help here)


matkoniecz:
Huh, I just got - immediately after I tried to test also language switching:

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.westnordost.streetcomplete.debug, PID: 6713
    java.lang.RuntimeException: Tried to perform an operation on an invalid pointer! This means you may have used a MapData that has already been removed.
        at com.mapzen.tangram.MapData.checkPointer(MapData.java:116)
        at com.mapzen.tangram.MapData.clear(MapData.java:108)
        at de.westnordost.streetcomplete.map.components.TracksMapComponent.putAllTracksInOldLayer(TracksMapComponent.kt:58)
        at de.westnordost.streetcomplete.map.components.TracksMapComponent.startNewTrack(TracksMapComponent.kt:47)
        at de.westnordost.streetcomplete.map.LocationAwareMapFragment.addTrackLocation(LocationAwareMapFragment.kt:203)
        at de.westnordost.streetcomplete.map.LocationAwareMapFragment.onLocationChanged(LocationAwareMapFragment.kt:188)
        at de.westnordost.streetcomplete.map.LocationAwareMapFragment.access$onLocationChanged(LocationAwareMapFragment.kt:26)
        at de.westnordost.streetcomplete.map.LocationAwareMapFragment$onAttach$2.invoke(LocationAwareMapFragment.kt:85)
        at de.westnordost.streetcomplete.map.LocationAwareMapFragment$onAttach$2.invoke(LocationAwareMapFragment.kt:85)
        at de.westnordost.streetcomplete.location.FineLocationManager$locationListener$1.onLocationChanged(FineLocationManager.kt:37)
        at android.location.LocationManager$ListenerTransport._handleMessage(LocationManager.java:352)
        at android.location.LocationManager$ListenerTransport.access$000(LocationManager.java:275)
        at android.location.LocationManager$ListenerTransport$2.handleMessage(LocationManager.java:298)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

no idea is it fault of this quest, language switching or something else.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestModule.kt
```diff
@@ -249,6 +250,7 @@ import javax.inject.Singleton
         AddBoardType(),
 
         AddBarrierType(), // basically any more detailed rendering and routing: OSM Carto, mapy.cz, OSMand for start
+        AddWhatIsOnBarrierAndPathIntersection(),
```
westnordost, line 253: `AddBarrier`?

westnordost (commented):
I'd very much prefer if there was no class hierarchy to share code. There are other ways to share code. (See e.g. KerbUtil)

Part or most of the filter logic could certainly go into a common utils file. The function that is called should have a clear general interface though. The primitives could move to SphericalEarthMath if they are primitive/general enough

matkoniecz:
Some new ideas for the question:

> What type of stile/gate/gap is here?

> How does this path go through this barrier?

> What is present where this barrier intersects the path?

sourced from [Discord OSM Server](https://discord.com/channels/413070382636072960/413070502580453387/912373224845754449)

westnordost:
Or

"What kind of barrier is here?"

It doesn't really matter for the user where the app sourced the information from that at this point, there is (very probably) some kind of barrier.

matkoniecz:
> It doesn't really matter for the user where the app sourced the information from that at this point, there is (very probably) some kind of barrier.

It is very useful when path and road close to each other have different barriers

matkoniecz:
I suspect that @westnordost is not entirely convinced about splitting this quest - but I would go in opposite direction, merge in here asking about `barrier=yes` located on paths and roads.

In meantime I tried making icons.

screen02

westnordost:
What is the status of this PR, is it ready to review / merge? Any open questions?

matkoniecz:
It is ready to merge but 

- not sure are you happy about icons (I am)
- you were dubious about splitting it into two quests

Why two quests?

It is very useful when path and road close to each other have different barriers.

I think that in this case this info can be easily made available and is likely to be useful so I want to make it available. I even thought about merging in question about `barrier=yes` into this quest for `barrier=yes` within `highway=` lines representing roads and paths.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierOnRoad.kt
```diff
@@ -0,0 +1,69 @@
...
+            }
+            BarrierType.STILE_STEPOVER_WOODEN -> {
+                changes.addOrModify("stile", "stepover")
+                changes.addOrModify("material", "wood")
+            }
+            BarrierType.STILE_STEPOVER_STONE -> {
+                changes.addOrModify("stile", "stepover")
+                changes.addOrModify("material", "stone")
+            }
+        }
```
westnordost, line 67: put this into a `fun BarrierType.applyTo(changes: StringMapChangesBuilder)`, see e.g. `Fee.kt`

Other barrier-type-quests can profit from this too.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/DetectWayBarrierIntersectionUtil.kt
```diff
@@ -0,0 +1,136 @@
+package de.westnordost.streetcomplete.quests.barrier_type
+
+import de.westnordost.streetcomplete.data.elementfilter.ElementFilterExpression
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.Way
+import de.westnordost.streetcomplete.util.isRightOf
+
```
westnordost, line 9: There is a lot of copy-pasta here from the `AddCrossing` quest. You should define this function generically like this

```kotlin
/** Return all nodes that are at crossing points between waysA and waysB.
 *
 *  In this context, two ways cross if
 *  1. they intersect at a shared node
 *  2. the node is not the end node of any of the two ways
 *  3. the ways actually cross each other, not use touch each other, i.e. the way A
 *     intersects, then continues on the other side of way B
 */
fun findNodesAtCrossingsOf(waysA: Iterable<Way>, waysB: Iterable<Way>, mapData: MapData): Iterable<Node>
````

and then make the `AddCrossing` code use that too.

Not sure about the special handling of sidewalk stuff (line 57-69 in `AddCrossing`) and tunnel-bridge stuff. The cleanest solution I can think of is to have the function not return just the node at the crossing, but the node+ways that are part of the crossing so that it can be filtered further.
I.e.
```kotlin
data class WaysCrossing(val node: Node, ways: List<Way>)

fun findNodesAtCrossingsOf(waysA: Iterable<Way>, waysB: Iterable<Way>, mapData: MapData): Iterable<Crossing>
```

westnordost (changes requested):
Regarding the picture, how about this
wall

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/DetectWayBarrierIntersectionUtil.kt
matkoniecz: > you should define this function generically

I tried doing this, but everything felt incredibly generically what made any reasoning about code more complicated.

I will try once more.

smichel17: IMO, some incidental duplication is less bad than coupling things via the wrong abstraction. If it's only duplicated between *two* quests, I wouldn't try too hard to de-duplicate, unless we are very sure that they are *actually* the same exact logic, and not something which just looks similar.

westnordost: The logic to determine if two ways actually cross at a node is both complex and generic enough that it is worth generalizing.

smichel17: A middle ground would be a comment in both places which warns that if you feel a need to copy-paste again, it should be pulled out into a helper function first.

app/src/main/res/values/strings.xml
```diff
@@ -1104,6 +1104,8 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
     <string name="quest_fireHydrant_confirmation_title">"Are you sure? Sometimes the diameter can be found on the hydrant itself."</string>
     <string name="short_no_answer_on_button">No</string>
     <string name="quest_level_title">On which level is %s located?</string>
+    <string name="quest_barrier_path_intersection">How does this path go through this barrier?</string>
+    <string name="quest_barrier_road_intersection">How does this road go through this barrier?</string>
```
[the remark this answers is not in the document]
smichel17: It's hard to say exactly what phrasing is best without seeing the full form. I don't remember why we can't re-use "What type of barrier is this?" (but I do remember it being discussed; I just wasn't paying attention).

In general, I like "pass through" (or maybe "cross") better than "go through".
"Here" is redundant with "this". It should either be "…through this barrier?" or "…through the barrier here?"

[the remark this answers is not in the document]
smichel17: I took more time to read/think about the context, the reason we can't re-use the other version is because this is for when you have something like a road crossing a fence or rock wall. So, I think this is my favorite so far:

```suggestion
    <string name="quest_barrier_path_intersection">How does this path cross the barrier here?</string>
    <string name="quest_barrier_road_intersection">How does this road cross the barrier here?</string>
```

I still don't really like the term *barrier* here. It's probably technically correct, but it makes me thing of something which is constructed specifically to **block** the road, whereas in this case the wall is blocking a big section, and the gate is constructed specifically to allow the road through. I'm not sure what the right word is, though.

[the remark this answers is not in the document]
smichel17: Maybe something like "What allows this road through the barrier here?" @peternewman any ideas?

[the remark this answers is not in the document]
matkoniecz: > I don't remember why we can't re-use "What type of barrier is this?" 

see https://github.com/streetcomplete/StreetComplete/pull/3515#issuecomment-996073809 for my reasons (though @westnordost  also was not a fan so maybe I should drop this idea)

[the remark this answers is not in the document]
westnordost: Alright, then "cross the barrier here" is best because it clearly refers to the crossing (which is highlighted), not ot the road and way

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/DetectWayBarrierIntersectionUtil.kt
matkoniecz: > The cleanest solution I can think of is to have the function not return just the node at the crossing, but the node+ways that are part of the crossing so that it can be filtered further.

Is it not going to be performance problematic due to generating and passing even more data? Though likely in both cases number of objects should be limited, without noticeable impact.

I also thought about passing callback function instead - that would allow further filtering without passing this data.

matkoniecz:
> Regarding the picture, how about this

In this case, probably for the first time, I prefer mine. But I can also this one if mine is actually worse/not really clear.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/DetectWayBarrierIntersectionUtil.kt
westnordost: > Though likely in both cases number of objects should be limited, without noticeable impact.

yes

Yes, I also thought about callback function instead. But I find this solution quite dirty.

westnordost:
> In this case, probably for the first time, I prefer mine.

Ok nvm then.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/DetectWayBarrierIntersectionUtil.kt
matkoniecz: `waysA` `waysB` - I think that I will name them `movingWays` and `barrierWays` at least while coding.

Otherwise it is too easy to confuse them with each other with generic `A` `B`.

I can refactor it to more generic names at the end or keep that. Or maybe there would be a better naming?

matkoniecz:
I am really unhappy about code here (ugly in general, can be probably be far more functional).

But at least it seems to be sort of working at least (no obvious bugs in the initial testing).

FloEdelmann (changes requested):
A lot of my comments are just code style or readability improvements. But there are also some potential mistakes.

The quest seems to work fine, and the tests confirm that. Good job!

app/src/main/res/values/strings.xml
```diff
@@ -1156,7 +1156,9 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
     <string name="quest_fireHydrant_diameter_unusualInput_confirmation_description">This diameter looks implausible.</string>
     <string name="quest_fireHydrant_confirmation_title">"Are you sure? Sometimes the diameter can be found on the hydrant itself."</string>
     <string name="short_no_answer_on_button">No</string>
-    <string name="quest_level_title">On which level number is %s located?</string>
+    <string name="quest_level_title">On which level is %s located?</string>
```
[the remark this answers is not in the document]
matkoniecz: huh, how it even worked

matkoniecz:
> By the way, I like the icons!

Thanks, nice to hear it (especially as it is likely the first real icon that I made on my own and actually liked)

FloEdelmann (changes requested):

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
peternewman: Sorry for being late to the party, I'd favour @smichel17 's "allows this road through the barrier" over "cross the barrier"; a bridge crosses a river, but a road/path doesn't cross a wall it goes through it (or they intersect, but that sounds a bit clunky).

Pedantically I'd argue the wall is bisected/cut by the road, i.e. technically there is no wall at the road, but something else (what we're trying to find out) at that point.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierOnPath.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    private val pathsFilter by lazy {
+        """
+        ways with
+          (highway ~ ${ALL_PATHS.joinToString("|")} and area != yes)
+          and (access !~ private|no or (foot and foot !~ private|no))
+    """.toElementFilterExpression()
+    }
+
+    override val commitMessage = "Add what is on intersection of path and barrier"
```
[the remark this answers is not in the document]
peternewman: Again, @FloEdelmann is better than mine!

FloEdelmann (changes requested):
Please see the three unresolved comments above:

1. https://github.com/streetcomplete/StreetComplete/pull/3515#discussion_r784754264 (the bottom part; unfortunately this is too large for a suggested change on GitHub)
2. https://github.com/streetcomplete/StreetComplete/pull/3515#discussion_r784765317
3. https://github.com/streetcomplete/StreetComplete/pull/3515#discussion_r784767248

After those are resolved, I'll approve this MR.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/BarrierType.kt
```diff
@@ -25,3 +27,23 @@ enum class BarrierType(val osmValue: String) {
     KISSING_GATE("kissing_gate"),
     BICYCLE_BARRIER("cycle_barrier")
 }
+
+fun BarrierType.applyTo(changes: StringMapChangesBuilder) {
+    changes.add("barrier", this.osmValue)
```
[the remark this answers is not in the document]
matkoniecz: This should never happen as only nodes without tags are taken.

matkoniecz:
> #3515 (comment) (the bottom part; unfortunately this is too large for a suggested change on GitHub)

Hmm, I do not really find it better this additional bonus change.

Too many things happening at all for me.

> #3515 (comment)

This should never happen as only nodes without tags are taken.

Though "Error: this should never happen, it is impossible" is not unusual. Not sure.

Even if it crashes, then crash and clear info that things went horribly wrong would be better than silent data corruption.

> #3515 (comment)

Done.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/BarrierType.kt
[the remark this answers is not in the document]
FloEdelmann: In `AddBarrierType`, `nodes with barrier = yes` are selected.

peternewman:
> > #3515 (comment)
> 
> This should never happen as only nodes without tags are taken.
> 
> Though "Error: this should never happen, it is impossible" is not unusual. Not sure.
> 
> Even if it crashes, then crash and clear info that things went horribly wrong would be better than silent data corruption.

Even if it's not currently, there's no reason that function couldn't be used for other barrier related quests which may have barrier set initially (or simply if you add barrier=yes), at which point it would break, so I'd say it's worth changing.

matkoniecz:
Oh right, that can happen even now as it is now a generic code.

> Though "Error: this should never happen, it is impossible" is not unusual.

demonstrated quite well

I would take down entire quest - that I implemented recently, nice one.

Thanks for spotting!

FloEdelmann (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/findNodesAtCrossingsOf.kt
```diff
@@ -0,0 +1,145 @@
...
+            mutableListOf(),
+            it.value
+        )
+    }
+    barriersByNodeId.forEach { (nodeId, barrierWays) ->
+        if (resultMap.contains(nodeId)) {
+            resultMap[nodeId]!!.barrierWays += barrierWays
+        }
+    }
+    return resultMap.values;
```
westnordost, line 106:
```suggestion
    return waysByNodeId.map { (nodeId, ways) ->
        WaysCrossing(
            mapData.getNode(nodeId)!!,
            barriersByNodeId.getValue(nodeId),
            ways
        )
    }
```

Shouldn't this do the same and be shorter and easier to read?

app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddDetectBarrierIntersectionTest.kt
```diff
@@ -0,0 +1,210 @@
...
+
+import de.westnordost.streetcomplete.quests.TestMapDataWithGeometry
+import de.westnordost.streetcomplete.quests.crossing.AddCrossing
+import de.westnordost.streetcomplete.testutils.node
+import de.westnordost.streetcomplete.testutils.p
+import de.westnordost.streetcomplete.testutils.way
+import org.junit.Assert
+import org.junit.Test
+
+class AddDetectBarrierIntersectionTest {
```
westnordost, line 11: No point in testing the same algorithm twice. Instead of testing applicability of a certain quest type, you should just test the findNodesAtCrossingsOf algorithm separately. The only thing then left to test for `AddBarrierOnRoad` (etc.) are any specialities like the thing with the tunnel etc.

So in the end, this will be less test code and easier to maintain (i.e. most of `AddCrossingTest` can be deleted too)

westnordost (changes requested):
Very clean implementation, thank you! Just two small comments:

app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddDetectBarrierIntersectionTest.kt
matkoniecz: The benefit of such testing is that bugs in use of the algorithm will be found (even if algorithm itself would be fine) and bugs in tests are less likely due to testing on something more specific.

It would also catch at least some cases of bad filters and so on. 

I would keep it as is.

app/src/main/java/de/westnordost/streetcomplete/quests/findNodesAtCrossingsOf.kt
matkoniecz: yes, and it - as far as I can see - should return the same. Still, I would test once more.

app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddDetectBarrierIntersectionTest.kt
westnordost: alright

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/findNodesAtCrossingsOf.kt
smichel17: Wouldn't `barriersByNodeId.getValue(nodeId),` return null if the id is not present, instead of an empty list? Unless it has a default value for the getter.
So I think to keep the same semantics, you need `barriersByNodeId.getValue(nodeId) ?: listOf()`.
Not sure if it matters in practice as I didn't look outside this snippet.

matkoniecz: At this point we filter only intersections that are defined - among others - by having connection with both barrier and moving way.

smichel17: Okay. In that case, maybe add a `!!`, so we will get crash reports from a NPE instead of surprising behavior, in case the assumption is wrong?

PR #3518, opened 2021-11-20
document that inheritance and class hierarchy should be avoided

matkoniecz:
see say https://github.com/streetcomplete/StreetComplete/pull/3515#pullrequestreview-811763382

I half-remembered this but I was unsure (again), I even checked CONTRIBUTING file


FloEdelmann (approved):

matkoniecz:
Not sure how to refer to this. "package" has already a distinct meaning, it is not in a class. Maybe I will just use "file"?

PR #3521, opened 2021-11-21
Also ask for backrest on leisure=picnic_table

mnalis:
[leisure=picnic_table](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpicnic_table) contain benches; ask if those benches have [backrest](https://wiki.openstreetmap.org/wiki/Key:backrest) too (as we do already ask for [amenity=bench](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbench))

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt
@@ -11,7 +11,7 @@ class AddBenchBackrest : OsmFilterQuestType<BenchBackrestAnswer>() {
 
     override val elementFilter = """
         nodes, ways with
-          amenity = bench
+          (amenity = bench or leisure = picnic_table)
           and (!area or area = no)
           and !backrest
           and !bench:type


smichel17:
Do picnic tables with backrests exist? I can't remember ever seeing one.

andrewharvey:
@mnalis do you have a picture of one you could upload to https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpicnic_table ?

matkoniecz:
> Do picnic tables with backrests exist? I can't remember ever seeing one.

https://www.e-zslp.pl/pl/p/Altana-lesna-drewniana-3%2C5-x-4-m/479

https://www.drewno.pl/gieldadrzewna/160-Altany-207652-Altana,+Sza%C5%82as,+wiata,+wiatrak+drewniany,meble+ogrodowe.html

https://sprzedajemy.pl/altana-ogrodowa-biesiadna-10-os-duzy-stol-2-lawki-dach-poznan-2-a1d299-nr54891315

From `altana piknikowa` image search.

app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt
```diff
@@ -11,7 +11,7 @@ class AddBenchBackrest : OsmFilterQuestType<BenchBackrestAnswer>() {
 
     override val elementFilter = """
         nodes, ways with
-          amenity = bench
+          (amenity = bench or leisure = picnic_table)
```
matkoniecz, line 14: The question refers to benches. Is it going to be confused where asked for such cases?

Note that there is already answer that it is not a bench but rather a picnic table.

matkoniecz (changes requested):

smichel17:
I wonder if it is a regional thing. At least in the northeast US, I don't think this would pass the "no spam" test.

But wait! Maybe we can do better. These are all of *covered* picnic tables. Is that always the case? Is it tagged in OSM? If yes to both, we could restrict the backrest question to just those, as picnic tables with individual/built in coverings are also rare in my area (thus, asking about any that do exist would not be spammy; you'd *effectively* exclude the region without having to maintain a white/blacklist).

app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt
mnalis: As each picnic table (AFAIK) contains a bench, `Does this bench have a backrest?` sounded OK to me in English and Croatian.  You're correct though that `Other answer` of `It's a picnic table` is superfluous if we've asked the quest for `leisure=picnic_table`, however I wouldn't know how to programmatically disable that.

mnalis:
> @mnalis do you have a picture of one you could upload to https://wiki.openstreetmap.org/wiki/Tag:leisure=picnic_table ?

@andrewharvey Sure. I added https://wiki.openstreetmap.org/wiki/File:Picnic_table_with_backrest_in_Croatia.jpg now.

@smichel17 ah, strange world, I didn't consider that other locations wouldn't have backrests on any benches on picnic table (but presumably still sometimes have on benches without tables?). As for the other ideas: see my picture, there are uncovered picnic tables with benches with backrests too... There are other ideas too to avoid if it is spammy apart from painful country filtering but they are somewhat ugly (like adding `Covered, and it has backrests` in the `Other answers` or similar).

mnalis:
Anyway, as I myself would not be able to implement anything apart from this simple idea, I'd close this PR. If anyone would be interested to contribute something more complex, feel free to link it here so others who are interested may find it.

smichel17:
Strange indeed! Learning the idiosyncrasies of different places is one of my favorite things about hanging around here. Also finding out falsehoods that I believed about my own area, when it turns out I just wasn't paying much attention.

app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt
mnalis: With new SC method of asking questions, this is much clearer though.
```
Does this bench have a backrest?
Picnic table
```

small_Screenshot_20220319_180830_de westnordost streetcomplete mn debug

PR #3523, opened 2021-11-21
Request capitalisation in new notes

arrival-spring:
Follow up to https://github.com/streetcomplete/StreetComplete/pull/3495

I hadn't spotted that there were two separate layouts, one for new notes and one for replying to existing notes, sorry.

This now requests capitalisation in new notes (previous PR was just replying to old notes).


westnordost:
Sounds useful, thanks

PR #3547, opened 2021-11-28
Fix location updates lost when StreetComplete is sent to the background

### LINKED ISSUE #3543: [Bug] User's location becomes frozen after Street Complete sent to background
When first started the user's location dot tracks their position perfectly.
But after SC is sent to the background and brought back again, the location becomes fixed (I believe the user's orientation still changes). 
Using the app switcher screen to close SC and then launching the app again always solves the problem.
Restarted my phone several times and the behaviour is the same.
Other apps (Maps.me) don't seem to have this problem.

May happen when switching apps or when locking phone.

**How to Reproduce**
Load SC, move around to confirm location is working
Press power button to lock screen 
Unlock phone
Location is fixed in SC and doesn't move around.
View open apps, close SC
Re-launch SC
Location tracks fine.

**Versions affected**
Using SC v38  on Android 12 . Pixel 3 phone in the UK.

smichel17:
Fixes #3543

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
@@ -41,8 +41,20 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
         }
     }
 
+    // Both signals are refreshed regardless of whether the device has both providers, because
+    // they are both canceled in removeUpdates and both checked in the locationListener
+    private fun refreshCancellationSignals() {
+        if (gpsCancellationSignal.isCanceled) {
+            gpsCancellationSignal = CancellationSignal()
+        }
+        if (networkCancellationSignal.isCanceled) {
+            networkCancellationSignal = CancellationSignal()
+        }
+    }
+
     @RequiresPermission(ACCESS_FINE_LOCATION)
     fun requestUpdates(minTime: Long, minDistance: Float) {
+        refreshCancellationSignals()
         if (deviceHasGPS)
             locationManager.requestLocationUpdates(GPS_PROVIDER, minTime, minDistance, locationListener, Looper.getMainLooper())
         if (deviceHasNetworkLocationProvider)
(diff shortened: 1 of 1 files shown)


smichel17:
There's still some duplication that I think might be nice to clean up and prevent bugs like this in the future, but unlike https://github.com/streetcomplete/StreetComplete/commit/de1e4aa86b9e0cde7350239750f04321a2bd29e8, this doesn't make things any worse, so the cleanup can happen separately.

app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
```diff
@@ -41,8 +41,20 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
...
+        if (networkCancellationSignal.isCanceled) {
+            networkCancellationSignal = CancellationSignal()
+        }
+    }
+
     @RequiresPermission(ACCESS_FINE_LOCATION)
     fun requestUpdates(minTime: Long, minDistance: Float) {
+        refreshCancellationSignals()
         if (deviceHasGPS)
             locationManager.requestLocationUpdates(GPS_PROVIDER, minTime, minDistance, locationListener, Looper.getMainLooper())
```
smichel17, line 59: I suppose it's a little awkward to refresh the signals here, given that the `requestLocationUpdates` callback doesn't actually *use* the signal, except to check its status in the callback.

app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
```diff
@@ -41,8 +41,20 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
         }
     }
 
+    // Both signals are refreshed regardless of whether the device has both providers, because
+    // they are both canceled in removeUpdates and both checked in the locationListener
```
westnordost, line 45: right

app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
```diff
@@ -41,8 +41,20 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
...
+            gpsCancellationSignal = CancellationSignal()
+        }
+        if (networkCancellationSignal.isCanceled) {
+            networkCancellationSignal = CancellationSignal()
+        }
+    }
+
     @RequiresPermission(ACCESS_FINE_LOCATION)
     fun requestUpdates(minTime: Long, minDistance: Float) {
+        refreshCancellationSignals()
```
westnordost, line 57:
```suggestion
```

westnordost (approved):

westnordost:
Right, another oversight! Thanks!

PR #3549, opened 2021-11-29
Increase highlightedElementsRadius to help ascertain if something is moved/was wrong location or gone

peternewman:


westnordost:
This is not going to be performant.

For a 100m radius, roughly 10 times as many elements need to be read from the database. Reading an area of data from the database is not very performant. The more elements are involved, the slower it gets.

Besides, how would anything move 100 meters? I chose 30m because this is a reasonable distance to make out where exactly a (small) shop is located in relation to other shops. For (previously) very unprecisely mapped things, 30m should be enough to take into account (e.g. recycling containers).

PR #3550, opened 2021-11-29
Improve readability of location betterThan check

smichel17:
Mostly cosmetic.

- Convert early return back to an if statement and convert comparison to an extension function, for better readability at the call site.
- Replace broken link with an archive.org version that works.
- Use consistent brace style (ie, none) for the early `if (...) return` statements
- Rename 'Significantly' to 'Much'. Shorter = easier to parse.
- Rename `currentBestLocation` to `previous`, also shorter, and the previous location isn't *necessarily* the previous best, it could be used to compare any two locations.
    - I'm least sure about this one. Happy to revert it (or anything else) if you don't think it's an improvement.
- Remove comment that's totally redundant with the single line it's commenting (comparing location provider).
- Remove redundant `return` statement (inside and outside the `when`).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
@@ -35,10 +35,10 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
 
     private val locationListener = object : LocationUpdateListener {
         override fun onLocationChanged(location: Location) {
-            if (!isBetterLocation(location, lastLocation)) return
-
-            lastLocation = location
-            locationUpdateCallback(location)
+            if (location.isBetterThan(lastLocation)) {
+                lastLocation = location
+                locationUpdateCallback(location)
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


FloEdelmann (approved):
Looks good to me :)

app/src/main/java/de/westnordost/streetcomplete/location/FineLocationManager.kt
```diff
@@ -83,48 +83,42 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
...
-private fun isBetterLocation(location: Location, currentBestLocation: Location?): Boolean {
-    // check whether this is a valid location at all. Happened once that lat/lon is NaN, maybe issue
-    // of that particular device
-    if (location.longitude.isNaN() || location.latitude.isNaN()) return false
-
-    if (currentBestLocation == null) {
-        // A new location is always better than no location
-        return true
-    }
+// Determines whether this Location reading is better than the current Location fix
```
westnordost, line 90:
```suggestion
/** Determines whether this Location reading is better than the previous Location fix */
```

westnordost (approved):

PR #3555, opened 2021-12-02
Add some metadata about Bulgaria and Greece

Dimitar5555:


res/country_metadata/popularReligions.yml
```diff
@@ -7,6 +7,7 @@ AU: [christian, muslim, buddhist]
 AZ: [muslim]
 BD: [muslim, hindu]
 BF: [muslim]
+BG: [christian, muslim]
```
matkoniecz, line 10: Note that it is default already, maybe it should be mentioned somehow in this file?

Either have `default: [christian, muslim, buddhist, hindu, jewish]` or `# note that adding order here is needed only if default order from AddReligionForm needs to be overriden`?

Dimitar5555: It says that the default is `[]` https://github.com/Dimitar5555/StreetComplete/blob/patch-1/res/country_metadata/popularReligions.yml#L3

matkoniecz: Yes, but it is used solely to change order in https://github.com/streetcomplete/StreetComplete/blob/0857679a146abaf474f051f8e8d12225d9a10ff3/app/src/main/java/de/westnordost/streetcomplete/quests/religion/AddReligionForm.kt#L19 (so default is actually `[christian, muslim, buddhist, hindu, jewish...`)

Dimitar5555: Is it possible to just set the default to `[christian, muslim, buddhist, hindu, jewish...]` in `popularRegligions.yml` or would this break something?

westnordost: it would be possible. I don't think it would break anything.

matkoniecz: It actually breaks something.

screen03

westnordost: Hm but why??

matkoniecz: I found and fixed problem in https://github.com/streetcomplete/StreetComplete/commit/e8c9b295d598991040ff59dc9044b6f48317a9f0 (in way nicer than my initial attempt but maybe still can be refactored)

The problem was that -1 (not found) was sorted before things at 0th, 1st, 2nd.. positions.

matkoniecz:
OK, religion issues are now fixed.

Ordering will now work properly and config file should be less confusing, default is now longer empty.

I applied your commits about road signs - and welcome on https://github.com/streetcomplete/StreetComplete/graphs/contributors

Thanks for this contribution and you are welcome to contribute more if you want!

If you are looking for something relatively simple or without significant programing feel free to ask.

This PR will be marked as closed but changes are applied (done this way as street signs were applied with https://github.com/streetcomplete/StreetComplete/commit/5422234cd17f220101866c636c9c07bcbabc0ee5 - author info preserved and religion part was replaced by changing default)

PR #3564, opened 2021-12-04
Sort building type recents by chronological order

smichel17:
The new "weighted" strategy is good for deciding *which* entries to show, but it does not provide a stable order; sometimes entries jump around, which is confusing/disorienting. This returns the sort order to chronological, while keeping the same strategy for choosing which entries to show. You'll lose the ability to predict *which* entry will disappear from the list (although it's usually nearer to the end), in exchange for the ability able to predict what order they'll be in, which means you can easily tell which one disappeared after it happens.

From https://github.com/streetcomplete/StreetComplete/discussions/3493

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
@@ -34,8 +34,8 @@ private const val MAX_ENTRIES = 100
  *  items of the sequence, in their original order.
  * If there are fewer than `count` unique items, continues counting items
  *  until that many are found, or the end of the sequence is reached.
- * If the first item is not null, it is always included, displacing the
- *  least-common of the other items if necessary.
+ * If the first item (i.e. the most recent one) is not null, it is always
+ *  included, displacing the least-common of the other items if necessary.
  */
 fun <T : Any> Sequence<T?>.mostCommonWithin(count: Int, historyCount: Int): Sequence<T> {
     val counts = this.countUniqueNonNull(historyCount, count)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
@@ -51,14 +51,12 @@ private data class ItemStats(val indexOfFirst: Int, var count: Int = 0)
 private fun <T : Any> Sequence<T?>.countUniqueNonNull(target: Int, minItems: Int): Map<T, ItemStats> {
     val counts = mutableMapOf<T, ItemStats>()
 
-    // Because this is a sequence, the chain of calls is applied to items one at a time.
-    // This means the `getOrPut` for one item affects the `counts.size` for the next item
-    this.withIndex()
-        .filter { it.value != null }
-        .takeWhile { it.index < minItems || counts.size < target }
-        .forEach {
-            counts.getOrPut(it.value!!) { ItemStats(it.index) }.count++
+    for (item in this.withIndex()) {
+        if (item.index >= minItems && counts.size >= target) break
+        item.value?.let { value ->
+            counts.getOrPut(value) { ItemStats(item.index) }.count++
         }
+    }
 
     return counts
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
@@ -30,18 +30,18 @@ class LastPickedValuesStore<T : Any>(
 
 private const val MAX_ENTRIES = 100
 
-/* Returns the `count` most-common non-null items in the first `historyCount`
+/* Returns the `target` most-common non-null items in the first `historyCount`
  *  items of the sequence, in their original order.
- * If there are fewer than `count` unique items, continues counting items
+ * If there are fewer than `target` unique items, continues counting items
  *  until that many are found, or the end of the sequence is reached.
  * If the first item (i.e. the most recent one) is not null, it is always
  *  included, displacing the least-common of the other items if necessary.
  */
-fun <T : Any> Sequence<T?>.mostCommonWithin(count: Int, historyCount: Int): Sequence<T> {
-    val counts = this.countUniqueNonNull(historyCount, count)
-    val top = counts.keys.sortedByDescending { counts[it]!!.count }.take(count)
+fun <T : Any> Sequence<T?>.mostCommonWithin(target: Int, historyCount: Int): Sequence<T> {
+    val counts = this.countUniqueNonNull(historyCount, target)
+    val top = counts.keys.sortedByDescending { counts[it]!!.count }.take(target)
     val latest = this.take(1).filterNotNull()
-    val items = (latest + top).distinct().take(count)
+    val items = (latest + top).distinct().take(target)
     return items.sortedBy { counts[it]!!.indexOfFirst }
 }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
@@ -51,7 +51,8 @@ private data class ItemStats(val indexOfFirst: Int, var count: Int = 0)
 private fun <T : Any> Sequence<T?>.countUniqueNonNull(target: Int, minItems: Int): Map<T, ItemStats> {
     val counts = mutableMapOf<T, ItemStats>()
 
-    // Sequences are evaluated lazily, so the `forEach` affects each `counts.size` check
+    // Because this is a sequence, the chain of calls is applied to items one at a time.
+    // This means the `getOrPut` for one item affects the `counts.size` for the next item
     this.withIndex()
         .filter { it.value != null }
         .takeWhile { it.index < minItems || counts.size < target }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
@@ -38,7 +38,7 @@ private const val MAX_ENTRIES = 100
  *  included, displacing the least-common of the other items if necessary.
  */
 fun <T : Any> Sequence<T?>.mostCommonWithin(target: Int, historyCount: Int): Sequence<T> {
-    val counts = this.countUniqueNonNull(historyCount, target)
+    val counts = this.countUniqueNonNull(target, historyCount)
     val top = counts.keys.sortedByDescending { counts[it]!!.count }.take(target)
     val latest = this.take(1).filterNotNull()
     val items = (latest + top).distinct().take(target)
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -30,29 +30,34 @@ class LastPickedValuesStore<T : Any>(
...
+ * or the end of the sequence is reached. If the first item is not null, it is always included.
  */
 fun <T : Any> Sequence<T?>.mostCommonWithin(count: Int, historyCount: Int): Sequence<T> {
     val counts = this.countUniqueNonNull(historyCount, count)
-    val top = counts.keys.sortedByDescending { counts.get(it) }
+    val top = counts.keys.sortedByDescending { counts[it]!!.count }.take(count)
     val latest = this.take(1).filterNotNull()
     val items = (latest + top).distinct().take(count)
-    return items.sortedByDescending { counts.get(it) }
+    return items.sortedBy { counts[it]!!.indexOfFirst }
```
smichel17, line 42: It's vaguely annoying that the compiler doesn't know that the keys are in the map. It is *possible* to do this without the map, using `counts.entries` to pair the keys and values, but it makes the code more verbose & (imo) confusing, especially around ensuring the latest item is always present.

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -30,29 +30,34 @@ class LastPickedValuesStore<T : Any>(
 
 private const val MAX_ENTRIES = 100
 
-/* In the first `historyCount` items, return the `count` most-common non-null items, in order.
- * If the first item is not included (and is not null), it replaces the last of the common items.
- * If fewer than `count` unique items are found, continue counting items until that many are found,
- * or the end of the sequence is reached.
+/* Returns the `count` most-common non-null items in the first `historyCount` items of the sequence.
```
matkoniecz, line 33: maybe mention what decides about order?

matkoniecz:
Just to check: have you tested it in action? Not sure how much I should test it while surveying

smichel17:
I have not yet tested it in action.

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
smichel17: Done

westnordost:
What is the status of this PR? Did you test it? Is it otherwise ready for final review/merge?

smichel17:
I tested it a few days ago. I liked it. I think it's ready for review.

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -30,29 +30,37 @@ class LastPickedValuesStore<T : Any>(
...
-/* In the first `historyCount` items, return the `count` most-common non-null items, in order.
- * If the first item is not included (and is not null), it replaces the last of the common items.
- * If fewer than `count` unique items are found, continue counting items until that many are found,
- * or the end of the sequence is reached.
+/* Returns the `count` most-common non-null items in the first `historyCount`
+ *  items of the sequence, in their original order.
+ * If there are fewer than `count` unique items, continues counting items
+ *  until that many are found, or the end of the sequence is reached.
+ * If the first item is not null, it is always included, displacing the
+ *  least-common of the other items if necessary.
```
FloEdelmann, line 38:
```suggestion
 * If the first item (i.e. the most recent one) is not null, it is always
 *  included, displacing the least-common of the other items if necessary.
```

app/src/test/java/de/westnordost/streetcomplete/quests/LastPickedValuesStoreTest.kt
```diff
@@ -7,14 +7,14 @@ import org.junit.Test
 
 class LastPickedValuesStoreTest {
 
-    @Test fun `mostCommonWithin sorts by frequency first, then recency`() {
+    @Test fun `mostCommonWithin sorts by recency`() {
         val items = sequenceOf(A, C, B, B, C, D)
-        assertEquals(items.mostCommonWithin(4, 99).toList(), listOf(C, B, A, D))
+        assertEquals(items.mostCommonWithin(4, 99).toList(), listOf(A, C, B, D))
     }
```
FloEdelmann, line 14: I think a new test should be added here:

```suggestion

@Test fun `mostCommonWithin includes the most common items in the original order`() {
    val items = sequenceOf(A, A, B, B, B, C)
    assertEquals(items.mostCommonWithin(2, 99).toList(), listOf(A, B))
}

```

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -30,29 +30,37 @@ class LastPickedValuesStore<T : Any>(
...
-    val items = takeAtLeastWhile(minItems) { counts.size < target }.filterNotNull()
-    return items.groupingBy { it }.eachCountTo(counts)
-}
+private fun <T : Any> Sequence<T?>.countUniqueNonNull(minItems: Int, target: Int): Map<T, ItemStats> {
+    val counts = mutableMapOf<T, ItemStats>()
 
-// Take at least `count` elements, then continue until `predicate` returns false
-private fun <T> Sequence<T>.takeAtLeastWhile(count: Int, predicate: (T) -> Boolean): Sequence<T> =
-    withIndex().takeWhile{ (i, t) -> i < count || predicate(t) }.map { it.value }
+    // Sequences are evaluated lazily, so the `forEach` affects each `counts.size` check
```
FloEdelmann, line 54: I don't get this comment. Does `forEach` cause all items to be evaluated? Then I'd suggest:

```suggestion
    // Sequences are evaluated lazily, but `forEach` causes all items to be evaluated
```

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -30,29 +30,37 @@ class LastPickedValuesStore<T : Any>(
...
- * If fewer than `count` unique items are found, continue counting items until that many are found,
- * or the end of the sequence is reached.
+/* Returns the `count` most-common non-null items in the first `historyCount`
+ *  items of the sequence, in their original order.
+ * If there are fewer than `count` unique items, continues counting items
+ *  until that many are found, or the end of the sequence is reached.
+ * If the first item is not null, it is always included, displacing the
+ *  least-common of the other items if necessary.
  */
 fun <T : Any> Sequence<T?>.mostCommonWithin(count: Int, historyCount: Int): Sequence<T> {
```
FloEdelmann, line 40: Please rename `count` to `target` because

1. it's consistent with `countUniqueNonNull` then
2. it prevents confusion between `counts` and `count`

FloEdelmann (changes requested):
The logic in `mostCommonWithin` seems a bit complicated but I don't know how to achieve the same result with a simpler algorithm. That said, I have a few nitpicks, but otherwise I think it's fine then.

smichel17 (commented):
> The logic in `mostCommonWithin` seems a bit complicated but I don't know how to achieve the same result with a simpler algorithm

Especially for "always includes the first item," I kept wanting to do it with an `if` statement to make the intent more clear (matching the comment above). But it turns out there's always a bunch of edge cases if you do it that way (first item is null or already in common items; not enough target items), and handling those introduces enough mess to make it less readable than before :(

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
FloEdelmann: I like `target` (and even `count`) way better than `n`.

app/src/test/java/de/westnordost/streetcomplete/quests/LastPickedValuesStoreTest.kt
FloEdelmann: It's not really covered, because in the `sorts by recency` test, all items are returned. I think the intent comes better across in my suggestion.

Merging some of the tests is not a bad idea, I guess.

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
smichel17: Does the `filter` (which has no special interaction with sequences here) get in the way of that point being obvious? I could pull it into the `forEach`
```kt
    this.withIndex()
        .takeWhile { it.index < minToCount || counts.size < target }
        .forEach {
            if (it.value != null) {
                counts.getOrPut(it.value!!) { ItemStats(it.index) }.count++
            }
        }
```
Or up onto the previous line
```kt
    this.withIndex().filter { it.value != null }
        .takeWhile { it.index < minItems || counts.size < target }
        .forEach {
            counts.getOrPut(it.value!!) { ItemStats(it.index) }.count++
        }
```

smichel17: I updated the comment. If the new one does not clarify well enough or you like either of those formattings, you can un-resolve this or add a new comment :)

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
smichel17: Done

app/src/test/java/de/westnordost/streetcomplete/quests/LastPickedValuesStoreTest.kt
smichel17: Okay, done.

As a consequence, the tests are now much more readable. I almost considered replacing the comment above `mostCommonWithin`with "see the tests for this function" since the assertion messages read like almost a direct copy-paste.

app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -30,29 +30,38 @@ class LastPickedValuesStore<T : Any>(
...
+    // This means the `getOrPut` for one item affects the `counts.size` for the next item
+    this.withIndex()
+        .filter { it.value != null }
+        .takeWhile { it.index < minItems || counts.size < target }
+        .forEach {
+            counts.getOrPut(it.value!!) { ItemStats(it.index) }.count++
+        }
+
+    return counts
+}
```
FloEdelmann, line 64: Reply to https://github.com/streetcomplete/StreetComplete/pull/3564#discussion_r770959686:

Why not use a `for` loop then? We're not really making use of the functional style in this case, and it's looking rather imperative anyway. A `for` loop is easily understandable, even if you don't know the advantages and intricacies of sequences.

```kt
val counts = mutableMapOf<T, ItemStats>()

// maybe there is a nicer indexed for...in loop style
for (var index = 0; index < this.size; index++) {
    if (index >= minItems && counts.size >= target) break

    if (this[index] != null) {
        counts.getOrPut(this[index]) { ItemStats(index) }.count++
    }
}

return counts
```

smichel17: 1. You can't index into a Sequence, so this exact version wouldn't work.
2. Sequences are important because it's sequences the whole way down. Values are stored as a comma-separated string in the SharedPreferences. We retrieve that, and then split directly to a sequence— laziness here means that if we only end up checking the first 30 values, we'll only do 30 splits. Admittedly, I didn't profile here to make sure this optimization is necessary.

I think I may be able to achieve the same effect using a while loop and an iterator. I'll try that.

westnordost: I interpret the 👍 by Flo as "Ok, resolved then".

Does this mean that Flo's code review is done and resolved?

smichel17: I interpreted it as "this will be resolved once you make the change".

smichel17: Turns out a `for` loop actually works great.

FloEdelmann (approved):
Looks good to me now. Thanks @smichel17!

westnordost:
Ok, thanks for the implementation and thanks for the review!

PR #3571, opened 2021-12-07
Fix quest list generation

### LINKED ISSUE #3569: Quest list started failing on CI
https://github.com/streetcomplete/StreetComplete/blob/master/.github/generate-quest-list.main.kts started failing, what I noticed in https://github.com/streetcomplete/StreetComplete/runs/4431608400?check_suite_focus=true

I have not investigated exactly what went wrong

Pinging @FloEdelmann as author of that script and let know recent editor of https://wiki.openstreetmap.org/w/index.php?title=StreetComplete/Quests&action=history ( https://wiki.openstreetmap.org/wiki/User_talk:MalgiK#SC_quest_list )

FloEdelmann:
Fixes #3569.

The script misinterpreted the `|-` in `{{Value|-1}}` as the start of a new table row.

### DIFF
--- a/.github/generate-quest-list.main.kts
+++ b/.github/generate-quest-list.main.kts
@@ -194,7 +194,7 @@ fun getWikiTableContent(): String {
 }
 
 fun parseWikiTable(wikiPageContent: String): List<WikiQuest> {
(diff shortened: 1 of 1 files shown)


westnordost:
Thanks!

Issue #3466, opened 2021-10-28
Google Play Compliance: Precise location supposedly sent off device


westnordost:
[Google Play introduced a new data safety policy](https://android-developers.googleblog.com/2021/07/new-google-play-safety-section.html). For app developers, this is basically a form they need to fill out and declare what user data exactly is shared with a third party (sent off the device).

This information will then be shown on the app page on Google Play.

The data types [as documented here](https://support.google.com/googleplay/android-developer/answer/10787469?hl=en-GB#types&zippy=%2Cdata-types) need to be declared.

I filled out the form and declared the following:
- Email address (but only processed ephemerally) - i.e. for OSM account registration
- Personal identifiers - i.e. OSM user name and oauth token

After filling out the form, I got this message:

![image](https://user-images.githubusercontent.com/4661658/139345176-284c964a-3e23-4920-8a4f-2c2ad4f586ee.png)

TLDR: Google Play claims that StreetComplete sends off the precise user's location somewhere which was apparently automatically determined by some AI. Does anyone have any idea what could cause this AI to think that?

Issue #2909, opened 2021-05-25
App currently not available on Google Play


westnordost:
### StreetComplete has been removed from Google Play

The message:

![image](https://user-images.githubusercontent.com/4661658/119580075-0595b580-bdc0-11eb-850a-94ba2c7a3aae.png)

---

I couldn't find anything in the linked articles

- [Permissions](https://play.google.com/about/privacy-security-deception/permissions/)
- [Location Permissions](https://support.google.com/googleplay/android-developer/answer/9888170#location-permissions)
- [Help Center](https://support.google.com/googleplay/android-developer/answer/9799150)

...that would explain in what way the app is not compliant with the Developer Program Policies. Apparently it is about background location permissions. And so, the link to "Developer Permission Declaration" leads to a page that looks like that...
![image](https://user-images.githubusercontent.com/4661658/119580447-b4d28c80-bdc0-11eb-9f92-62d3254ede66.png)
...which makes sense, because the app does not use any background location permissions.

What doesn't make sense is that the app got removed from Google Play.

westnordost:
I already wrote a message to the policy support team. Content:

> It is claimed that the app does not comply with the Location Permissions policy and it has been removed from Google Play. I believe this is a mistake:
> 
> The app DOES NOT request location in the background or foreground service
> The app DOES NOT display any ads or track its users in any way, i.e. transmit their location anywhere
> The app DOES request location when the app is in foreground. The location is used to
> 1. download data in the vicinity of the user. This data is indispensable for the core functionality of the app
> 2. display the user's position for orientation on the map, like any map-based app does
> 
> The purpose of the location permission is clearly communicated to the user.
> 
> PLEASE NOTE: 
> In the Google Play Console -> App content -> Sensitive app permissions
> I am UNABLE to make any statement because there is only the text "Your app doesn't request sensitive permissions, so you don't need to do anything here" and nothing else. A bug in the Play Console?

westnordost:
It is possible that even if the app does not have the `ACCESS_BACKGROUND_LOCATION` permission in its manifest, a dependency has it so that in the merged manifest, the app will request that permission.

Android Studio has this handy feature where one can directly view the merged manifest:

![image](https://user-images.githubusercontent.com/4661658/119581267-82c22a00-bdc2-11eb-8173-af50289f8f99.png)

**Result: Nope, contrary to the claim, no background permission is requested.**

[...]

PR #3573, opened 2021-12-08
Recording and Uploading of GPX Traces

### LINKED ISSUE #2327: Add an ability to record a track
**Use case**

When I discover a missing road/path etc during my questing, I'm not able to add it immediately, for two reasons: it would take a ton of time to do so right, and StreetComplete does not appear to support adding roads or paths anyway.

**Proposed Solution**

I would love an ability to quickly record a GPS track so that I could add the missing road/path once I'm back home and have access to tools like JOSM. IMO it makes sense to add this functionality to StreetComplete because switching to a different app to record a track is inconvenient and further detracts from questing.

### LINKED ISSUE #2871: "Add missing way" tool
**Use case**
As it stands, the "Create new note" feature is awesome, especially that fact you can upload pictures. However, I've often stumbled upon missing paths and streets in my surveys, and I struggle to adequately convey their location and shape via either text or pictures. Some things just aren't visible on available imagery, so having this knowledge come from people on location would be useful. GPS traces alone often aren't accurate and having people walk down paths would make it cumbersome. I'd therefore propose an option to add a small drawing _over_ the map, attached in the same way a picture is. Let me be clear, this would *not* be directly editing the map, only giving a more useful visual aid to editors.

**Proposed Solution**
My proposal would, in essence, be a new option between "Create new note" and "Open location in another app" called "Add missing way". Clicking it, StreetComplete would give you a draw tool, allowing you to create lines _over_ the map. 
I'm nothing close to an artist, so, a quick impression in paint, the red line drawn manually:
streetcomplete_566113_full

For convenience, an undo option and (possibly) an eraser would complete that feature. When you've drawn the line(s), you click "ok" at the bottom of the screen and a note is created at the location you originally clicked "Add missing way". Your drawing is attached just like with a note with a picture (so, uploaded to the server), and you get to comment on what you just drew (e.g. "Missing footway, asphalt" or "Example Street, residential"). 

If we wanted to get more advanced/useful, StreetComplete could then also offer a selection of possible tags from the highway key, as well as the other questions usually asked for ways (lit, surface, etc.) that are then appended to the comment before upload.

I'm not sure having just an attached drawing would be the ideal way of communicating the way's shape to editors, ideally it'd be something they can pull into iD or JOSM directly, but it'd certainly be better than having to interpret comments and pictures on the ground.

goldbattle:
Hi, thanks again for the project.
This is trying to address the following issues:
- #2209
- fixes #2327
- fixes #2871

Here is an example user experience idea of how to start recording a track, and then composing a note. Right now I am able to record the history of gps points, and have it stored in the NoteEdits database object for later uploading to the server. I also changed to visualization to highlight these recorded segments in red.

https://user-images.githubusercontent.com/2222562/145152485-8515dfe5-b2c2-480f-a2de-b8187caa7c7d.mp4

The idea is that the GPX track will be linked to the note (same as how images are currently) so this can provide additional information for editors and can be a part of the online "Public GPS Traces" overlay. The next step is for me to implement uploading to the gps trace server (I was looking at using some of [this code](https://github.com/westnordost/osmapi/blob/master/libs/traces/src/main/java/de/westnordost/osmapi/traces/GpsTracesApi.java)), so any feedback is much appreciated.


goldbattle:
I have completed the initial uploader.
Here is an example upload I have created:
https://www.openstreetmap.org/note/2967923
https://www.openstreetmap.org/user/Goldbattle/traces/3974369
image

I am not sure why the filename is not what is specified, but I think what it is now is pretty good.
The editor can download the GPX trace, drag and drop into the editor, and then see the path:
image


- - -

It needs to have additional permissions requested from the OSM api so that it can upload and get info about the GPX track.

gpx_sc_example_03

Here is an application I created to test with (`de/westnordost/streetcomplete/data/user/UserModule.kt`):
```
    private const val CONSUMER_KEY = "tbHqDSVojwRZT4hQ2jMsOzdN4i1YsTQqTvAVeFcD"
    private const val CONSUMER_SECRET = "X3tc6G6aD4DPWVOhmvQowFdvPzwXvebSYnJStHWL"
```

westnordost:
First off, thank you for the effort so far and the PR!

There are some points to clarify first, as if I remember correctly, some problems where talked about in prior issues. If I recall correctly, the main points were:
- **Google Play compliance:** As the track should probably continue to be recorded while the app is (shortly) in the background, location permission for the  background would be required. Additionally, a foreground notification must be shown that allows to stop recording the track. Google Play has very strict rules regarding use of background location and precise location, so strict in fact that their compliance bot marked the app as non-compliant and removed it. It was a false-positive, but there was no getting through (to a human) at Google. I do not want to go through this again. See #2909 and then #3466
- **Privacy**: Due to privacy concerns, this may never be on by default. The user should also not be able to accidentally turn it on because he does not know what it implies
- **Feature for advanced users**: This feature is likely only interesting for advanced users, maybe those who would use that GPS trace later at the desktop to add some ways or whatever with JOSM. Hence, as the app is primarily for non-technical people, the addition of such a feature should not be in the way in the regular UI, not take up any space in the normal UI.

Further, the question is what (additional) use cases the recording of tracks should cover. Two come to my mind:
- Generally contributing (public, anonymized) GPS traces for better alignment of imagery
- Directly referencing a (short) trace in a note so that it can be added and/or corrected via JOSM, maybe by the note creator himself

What use case should this feature cover?

smichel17:
> This feature is likely only interesting for advanced users, maybe those who would use that GPS trace later at the desktop to add some ways or whatever with JOSM

I am not sure I agree. Maybe I qualify as "advanced" but I have no interest in using iD or JOSM to add new ways. However, I would be happy to walk down a path that is not mapped and upload a trace so that some armchair mapper can add it later. That said,

1. I'm not sure what the advantage is of having this function in SC instead of a separate app.
2. I agree that this use case is rare enough that it should not take (much) space in the GUI. Nothing which is always visible, at least. I think a menu option would be ok.

goldbattle:
@westnordost  I think the purpose is very specific to SC and is what you mentioned:
> Directly referencing a (short) trace in a note so that it can be added and/or corrected via JOSM, maybe by the note creator himself

SC also allows for uploading of a photo along to provide more context also, which is a big advantage compared to other thirdparty apps. I personally have not used another app, and would rather just use SC instead of having to switch between apps. Some example use cases I can think of if you need motivation for this feature:

- A sidewalk connection between buildings is missing. I can record a small track to attach to the note
- Trail under a tree is missing since it is not visible from satellite imagery. I can record this to provide a trail record for them.
- A trashcan has moved. I can record a stationary trace of my location to provide information on where the trashcan should be.
- A road / walkway should be shifted. I can record a small trace to provide as a reference.

Re: Only for advance users. I am not sure I agree with this. I don't see why adding more information to the note would not be appreciated by someone online. Might me just being naïve here since I have not really used the online editor at all.

@smichel17 I agree that adding more to the UI would be a bad idea, thus I hid this away in the long-press menu where a user can create a note already. Also the user can only record one trace at a time (and can still answer quests while recording), so this simplifies and only adds a single button to the screen to stop the recording.
image

- - -


I understand the google play issues, and will read up on this, but right now if the user leaves the app, no location information is recorded in the track. This causes a "jump" in the recording, but I think it is a bit of work to require background location access and since the traces are so short I am not sure if it is worth it.

I think maybe adding a popup when the user first clicks could help inform the user about where their location information is being sent? Should I add this?

HolgerJeromin:
I was very sceptic about this feature. 
But as it is only active on request of the mapper I am very happy with the presented video. I would explicit vote for no location updates in the background.
The track is be an additional way to provide context for one specific missing OSM data. (continuous gps tracks with mapping stops on a side walk are not so useful)

smichel17:
> I would explicit vote for no location updates in the background.

We might need some text to explain this. And/or, maybe putting the app into the background (for more than a few seconds) should automatically end the trace, since at this point it is no longer accurate. There are probably other options, too, if I thought out this more.

matkoniecz (changes requested):
My fundamental problem with this is:

- this is not a critical feature

- in my opinion (which is a guess) there is decent chance that Googlebot will go berserk and ban StreetComplete again once this is added. Lets say 10% chance for immediate ban. It can also cause ban in the future.

Given that I am scared and would prefer to avoid adding this.

westnordost:
IIRC Google bot removed the app because it claimed that the app used background location and hence, stricter policies applied to the app that the app did not comply with. The solution, in the end, was to comply with the stricter policies even though the app does not use background location.

The other Google bot issue I mentioned was related to "sharing user's location", which the app also doesn't do, but as we all agreed, does so indirectly by 1. downloading map tiles at the position of the user and 2. inherently by requiring that solving quests is done on-site and the user's OSM history is public. We solved this by simply stating that "yes it does share location" in the questionaire. It was just a questionaire, apparently this information is shown to the user somewhere on google play or whatever.

So, as long as a GPX track feature works the way as @goldbattle described, I think nothing should change from the viewpoint of Google.

Anyway, I can agree with the others here that it sounds like a really useful feature. Especially outside cities where footpaths are not well recorded yet, this can be a really handy method to let other users know of an unmapped footpath and conveniently already supply the (exact) path with it. I'll review the code now, maybe I have some more questions then

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
```diff
@@ -22,46 +22,52 @@ class TracksMapComponent(ctrl: KtMapController) {
 
     private var index = 0
     private var tracks: MutableList<MutableList<LngLat>>
+    private var tracksRecording: MutableList<Boolean>
```
westnordost, line 25: I'd prefer something like this. Put into one data structure what belongs together.

```suggestion
    private class Track(val trackpoints: MutableList<LngLat>, isRecording: Boolean)

    private var tracks: MutableList<Track>
```

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/Note.kt
```diff
@@ -35,3 +35,11 @@ data class NoteComment(
         OPENED, COMMENTED, CLOSED, REOPENED, HIDDEN
     }
 }
+
+@Serializable
+data class NoteGPXTrack(
```
westnordost, line 40:
```suggestion
data class Trackpoint(
```

Also, a trackpoint per se has nothing to do with `Note`s, so it shouldn't be in the same file. Actually, it shouldn't even be in the same package. It should probably be in `data/osmtracks` or similar.

westnordost (changes requested):
Overall, looks quite good! I didn't do a complete review yet because I found a few things on the way that should be changed.

In general, stuff that is exclusively related to GPS tracks (API wrapper, data classes) should go into an own package. Also, the naming is off (as explained in one comment).

I did not read the code about the upload yet in detail but the "upload data map" data structure  strikes me as odd. But as I wrote, I did not look into this in detail yet. 
Also, it may be preferrable to have the GPS traces upload in a separate class, like it is done for the image upload.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/Note.kt
FloEdelmann: Another thing to keep in mind after renaming: For acronyms in PascalCase or camelCase, please follow the convention we used elsewhere: e.g. `OsmApi` or `osmApi`, not `osmAPI` or `OSMApi`.

goldbattle:
Great, thanks for the specific feedback, was a bit unfamiliar so wasn't sure where best put things in some case (just created a proof of concept). I will find some time later this week or weekend to make the requested changes, thanks!

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesApi.kt
```diff
@@ -65,4 +66,9 @@ interface NotesApi {
      * @return the incoming notes
      */
     fun getAll(bounds: BoundingBox, limit: Int, hideClosedNoteAfter: Int): List<Note>
+
+    /**
+     * Return the GPS api endpoint
+     */
+    fun getGPXApi() : GpsTracesApi
```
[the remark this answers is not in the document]
goldbattle: Move this, and the `create()` logic into `data.osmtracks.TracksApi`

app/src/main/java/de/westnordost/streetcomplete/user/LoginFragment.kt
```diff
@@ -137,7 +137,9 @@ class LoginFragment : Fragment(R.layout.fragment_login),
         private val REQUIRED_OSM_PERMISSIONS = listOf(
             Permission.READ_PREFERENCES_AND_USER_DETAILS,
             Permission.MODIFY_MAP,
-            Permission.WRITE_NOTES
+            Permission.WRITE_NOTES,
+            Permission.READ_GPS_TRACES,
+            Permission.WRITE_GPS_TRACES
```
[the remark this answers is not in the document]
goldbattle: Currently if a user is already login they can upload existing quests as is, nothing changes here.

(after the below committed refactor):
If a login user creates a track and tries to sync it, it will have a popup saying they need to re-authorize due to throwing `OsmAuthorizationException`. This means that login users will only need to "upgrade" their permissions after they create one of these. New users should already have this permission requested (since there is only one app / client id...).
image

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
goldbattle: I wanted to explicitly enforce that there is only one historical recording at a time. This is stored in the `tracksRecording`. There can be many previous segments of recording tracks, but only the most recent one should be uploaded. I guess with this data structure I would loop through and find the last one which has `isRecording` set to true and this is the track that is uploaded?

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
```diff
@@ -35,9 +37,15 @@ open class LocationAwareMapFragment : MapFragment() {
     var displayedLocation: Location? = null
         private set
 
+    /** If we are actively performing GPX tracking */
```
matkoniecz: Maybe just rename variable to `gpxTrackingActive` and drop this comment?

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
```diff
@@ -130,11 +139,36 @@ open class LocationAwareMapFragment : MapFragment() {
...
         locationMapComponent?.isVisible = false
         locationManager.removeUpdates()
     }
 
+    fun stopPositionTrackingGPX() {
+        gpxTracking = false
+        tracksRecorded.clear()
+        tracks.last().forEach {
+            // time here is in milliseconds
+            // TODO: why is altitude zero in emulation?
```
matkoniecz, line 171: Seems general issue, see https://stackoverflow.com/questions/65325665/simulate-altitude-position-in-android-emulator

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
westnordost: I don't quite understand. With this data structure, you don't enforce that there is only one recording. E.g.
```kotlin
tracks = listOf(listOf(p1,p2,p3), listOf(p4,p5))
tracksRecording = listOf(true, true)
```
-> both are "isRecording==true" tracks.

Also, you write 
> There can be many previous segments of recording tracks, but only the most recent one should be uploaded

but the `TracksMapComponent` has nothing to do with upload. And it has also nothing to do with storing the track temporarily before upload. It exclusively displays the track(s), nothing else.

app/src/main/java/de/westnordost/streetcomplete/user/LoginFragment.kt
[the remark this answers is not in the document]
westnordost: Hmm, so this forcibly logs out the user when he firet uses this feature. This behavior will look as if it is a bug.

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+
+private inline fun <T> wrapExceptions(block: () -> T): T =
+    try {
+        block()
+    } catch (e: OsmAuthorizationException) {
+        throw AuthorizationException(e.message, e)
+    } catch (e: OsmConflictException) {
+        throw ConflictException(e.message, e)
+    } catch (e: OsmQueryTooBigException) {
+        throw QueryTooBigException(e.message, e)
```
westnordost, line 72:
```suggestion
```
will never be thrown by this API call

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+import de.westnordost.streetcomplete.data.upload.ConflictException
+import de.westnordost.streetcomplete.data.user.AuthorizationException
+import java.time.Instant
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+class TracksApiImpl(osm: OsmConnection) : TracksApi {
+    private val api: GpsTracesApi = GpsTracesApi(osm)
+
+    override fun create(tracks: List<Trackpoint>): Track = wrapExceptions {
```
westnordost, line 26: Should name this parameter `trackpoints` or `track`

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+            trackpoints.add(
+                GpsTrackpoint(
+                    OsmLatLon(it.position.latitude, it.position.longitude),
+                    Instant.ofEpochMilli(it.time),
+                    idx == 0,
+                    it.horizontalDilutionOfPrecision,
+                    it.elevation
+                )
+            )
+        }
```
westnordost, line 53:
```suggestion
        val trackpoints = tracks.mapIndexed { idx, it -> 
            GpsTrackpoint(
                OsmLatLon(it.position.latitude, it.position.longitude),
                Instant.ofEpochMilli(it.time),
                idx == 0,
                it.horizontalDilutionOfPrecision,
                it.elevation
            )
        }
```

You can use `mapIndexed`, it's shorter

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+        // https://stackoverflow.com/a/49862573/7718197
+        val name = DateTimeFormatter
+            .ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS")
+            .withZone(ZoneOffset.UTC)
+            .format(Instant.ofEpochSecond(tracks[0].time))
+            .replace("-", "_")
+            .replace(":", "_")
+            .replace(" ", "_") + ".gpx"
+        val visibility = GpsTraceDetails.Visibility.IDENTIFIABLE
+        val description = ApplicationConstants.USER_AGENT
```
westnordost, line 38: Maybe it would make sense to have the description as a parameter so that it can be 1:1 the text the user wrote in the note itself?

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -0,0 +1,82 @@
...
+
+        // Filename is just the start of the track
+        // https://stackoverflow.com/a/49862573/7718197
+        val name = DateTimeFormatter
+            .ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS")
+            .withZone(ZoneOffset.UTC)
+            .format(Instant.ofEpochSecond(tracks[0].time))
+            .replace("-", "_")
+            .replace(":", "_")
+            .replace(" ", "_") + ".gpx"
```
westnordost, line 36:
```suggestion
            .ofPattern("yyyy_MM_dd'T'HH_mm_ss.SSSSSS'Z'")
            .withZone(ZoneOffset.UTC)
            .format(Instant.ofEpochSecond(tracks[0].time)) + ".gpx"
```

shorter and same format as those "dragonpilot" traces on https://www.openstreetmap.org/traces

westnordost (commented):
Here's another small review of the things you moved to the new package

app/src/main/java/de/westnordost/streetcomplete/user/LoginFragment.kt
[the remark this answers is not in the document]
goldbattle: I added a custom dialog which informs the user why they are being logged out.
Not sure how to handle it without logging out the user back out so that they can re-request with all the correct permissions.
image

This will only get shown if a current user uploads a trace, otherwise they won't see this.
New users will always have all correct permissions after their first login.

westnordost:
You made some changes. Let me know when it is ready for another review.

goldbattle:
I need to make a few more based on the first set of comments. Will request a review once I am able to make the changes, thanks!

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
goldbattle: You are completely right here, apologies.

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+        val name = DateTimeFormatter
+            .ofPattern("yyyy_MM_dd'T'HH_mm_ss.SSSSSS'Z'")
+            .withZone(ZoneOffset.UTC)
+            .format(Instant.ofEpochSecond(trackpoints[0].time)) + ".gpx"
+        val visibility = GpsTraceDetails.Visibility.IDENTIFIABLE
+        val description = if (noteText != null && noteText.isNotBlank()) {
+            noteText
+        } else {
+            "Uploaded via " + ApplicationConstants.USER_AGENT
+        }
```
FloEdelmann, line 35:
```suggestion
        val description = if (noteText.isNullOrEmpty()) "Uploaded via ${ApplicationConstants.USER_AGENT}" else noteText
```

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -607,6 +611,25 @@ class MainFragment : Fragment(R.layout.fragment_main),
         mapFragment?.updateCameraPosition(300) { zoomBy = +1f }
     }
 
+    private fun onClickTracksStop() {
+
+        // hide the track information
+        binding.stopTracksButton.visibility = View.INVISIBLE
```
FloEdelmann, line 617: Why not `View.GONE`?

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
```diff
@@ -21,47 +21,49 @@ class TracksMapComponent(ctrl: KtMapController) {
     private val layer2 = ctrl.addDataLayer(LAYER2)
 
     private var index = 0
-    private var tracks: MutableList<MutableList<LngLat>>
+    private class Track(val trackpoints: MutableList<LngLat>, val isRecording: Boolean)
```
FloEdelmann, line 24: Could be a `data class` right? Also, maybe another name would be better to avoid confusion with the `Track` class from `app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/Track.kt`.

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
smichel17: Note: This changes the semantics, because `blank` and `empty` are different. (blank also includes any whitespace)

Here is the shortest/"cleanest" version I came up with
```suggestion
        val description = noteText.orEmpty().ifBlank { "Uploaded via ${ApplicationConstants.USER_AGENT}" }
```

Other options:
```kt
val description = noteText?.takeIf { !it.isBlank() } ?: "Uploaded via ${ApplicationConstants.USER_AGENT}"
val description = noteText?.takeUnless { it.isBlank() } ?: "Uploaded via ${ApplicationConstants.USER_AGENT}"
```

…Or don't change anything. Tbh, recently I've been falling out of love with kotlin's stdlib. I like the readability of the code it produces in any given spot, but it is bloat which really increases the mental overhead of the language in general. :shrug:

app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestController.kt
```diff
@@ -64,10 +65,11 @@ import kotlin.collections.ArrayList
     suspend fun createNote(
         text: String,
         imagePaths: List<String>,
-        position: LatLon
+        position: LatLon,
+        tracks: ArrayList<Trackpoint>
```
smichel17, line 69:
```suggestion
        tracks: ArrayList<Trackpoint>,
```

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -607,6 +611,25 @@ class MainFragment : Fragment(R.layout.fragment_main),
         mapFragment?.updateCameraPosition(300) { zoomBy = +1f }
     }
 
+    private fun onClickTracksStop() {
+
+        // hide the track information
```
smichel17, line 616: This is kind of pedantic, but elsewhere in the app, we don't have empty lines at the beginning and end of functions.
```suggestion
    private fun onClickTracksStop() {
        // hide the track information
```

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -607,6 +611,25 @@ class MainFragment : Fragment(R.layout.fragment_main),
...
+        mapFragment.show3DBuildings = false
+        val location = mapFragment.displayedLocation ?: return
+        val pos = LatLon(location.latitude, location.longitude)
+        val offsetPos = mapFragment.getPositionThatCentersPosition(pos, mapOffsetWithOpenBottomSheet)
+        mapFragment.updateCameraPosition { position = offsetPos }
+
+        freezeMap()
+        showInBottomSheet(CreateNoteFragment())
+
+    }
```
smichel17, line 631:
```suggestion
        showInBottomSheet(CreateNoteFragment())
    }
```

app/src/main/res/layout/fragment_main.xml
```diff
@@ -86,6 +86,16 @@
...
 
+            <ImageButton
+                android:id="@+id/stop_tracks_button"
+                style="@style/RoundAccentButton"
+                android:layout_width="match_parent"
+                android:layout_height="@dimen/map_button_size"
+                android:contentDescription="@string/map_btn_stop_track"
+                android:scaleType="center"
+                android:src="@android:drawable/ic_menu_close_clear_cancel"
+                android:visibility="invisible" />
```
smichel17, line 97: Mirroring @FloEdelmann's comment above
```suggestion
                android:visibility="gone" />
```

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
```diff
@@ -21,47 +21,49 @@ class TracksMapComponent(ctrl: KtMapController) {
...
+    private var tracks: MutableList<Track>
 
     init {
         tracks = ArrayList()
-        tracks.add(ArrayList())
+        tracks.add(Track(ArrayList(), false))
     }
```
smichel17, line 30: I think you could collapse the init block to
```kt
    private var tracks: MutableList<Track> = arrayListOf(Track(ArrayList(), false))
```

smichel17 (commented):
I've reviewed this for coding style and for bugs (didn't find any :tada:), but only at the surface level (each change locally within that file).

That is, I haven't considered how these changes interact with each other or the rest of the app, since I think westnordost already did that. Also, I did this review entirely here on GitHub, so my bug-hunting was limited :).

There's a bunch of comments but they're all both pretty minor & easy to address— overall, it looks good!

app/src/main/assets/map_theme/jawg/streetcomplete.yaml
```diff
@@ -146,20 +156,29 @@ layers:
...
-                    join: round
-                    order: 1000
+          filter: { old: [ false ], record: [ false ] }
+          draw:
+            track-lines:
+              color: global.track_color
+              width: [ [ 14, 6px ],[ 18, 12px ] ]
+              collide: false
+              join: round
+              order: 1000
```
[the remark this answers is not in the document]
westnordost: No, not really possible

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
```diff
@@ -26,15 +29,17 @@ class NoteEditsUploader @Inject constructor(
...
-        uploadEdits()
-    } }
+    suspend fun upload() = mutex.withLock {
+        withContext(Dispatchers.IO) {
+            // first look if any images have not been activated yet
+            uploadMissedImageActivations()
+            // then do the usual stuff
+            uploadEdits()
+        }
+    }
```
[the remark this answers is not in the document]
goldbattle: I think this is from my running format from android studio on the whole file. Should I try to remove all these changes? or leave as is?

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
```diff
@@ -55,17 +60,34 @@ class NoteEditsUploader @Inject constructor(
     }
 
     private fun uploadEdit(edit: NoteEdit) {
-        val text = edit.text.orEmpty() + uploadAndGetAttachedPhotosText(edit.imagePaths)
 
+        // try to upload the image
+        val imageText = edit.uploadedDataMap.getOrDefault(
+            "images",
+            uploadAndGetAttachedPhotosText(edit.imagePaths)
```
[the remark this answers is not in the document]
goldbattle: You are correct, I didn't understand the `getOrDefault` function which will always evaluate the default variable. I converted this to an explicitly check if no data has been uploaded yet.
```kotlin
        // try to upload the image
        var imageText = edit.uploadedDataMap.getOrDefault("images", "")
        if(imageText.isEmpty()) {
            imageText = uploadAndGetAttachedPhotosText(edit.imagePaths)
            noteEditsController.updateData(edit, "images", imageText)
        }

        // try to upload the tracks
        var tracksText = edit.uploadedDataMap.getOrDefault("tracks", "")
        if(tracksText.isEmpty()) {
            tracksText = uploadAndGetAttachedTracksText(edit.tracks, edit.text)
            noteEditsController.updateData(edit, "tracks", tracksText)
        }
```

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
goldbattle: wasn't sure of the difference, changed it to GONE.

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
goldbattle: I changed it to a `data class`  not sure of a better name currently.

goldbattle:
Thanks all for the detailed comments, very much appreciated.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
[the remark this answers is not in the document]
smichel17: If you prefer the previous style, I just checked and it looks like there is a [`getOrElse`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-or-else.html#kotlin.collections$getOrElse(kotlin.collections.Map((kotlin.collections.getOrElse.K,%20kotlin.collections.getOrElse.V)),%20kotlin.collections.getOrElse.K,%20kotlin.Function0((kotlin.collections.getOrElse.V)))/defaultValue) function which has the semantics you expected. That would be one of these:

- Unconditional upload, like it originally was:
   ```kt
   val imageText = edit.uploadedDataMap.getOrElse("images") {
       uploadAndGetAttachedPhotosText(edit.imagePaths)
   }
   noteEditsController.updateData(edit, "images", imageText)
   ```
- Only upload when it hasn't been already, like it is now:
   ```kt
   val imageText = edit.uploadedDataMap.getOrElse("images") {
       uploadAndGetAttachedPhotosText(edit.imagePaths).also {
           noteEditsController.updateData(edit, "images", it)
       }
   }
   ```
    - Less kotlin-ized version (without the `also`):
       ```kt
       val imageText = edit.uploadedDataMap.getOrElse("images") {
           val newText = uploadAndGetAttachedPhotosText(edit.imagePaths)
           noteEditsController.updateData(edit, "images", newText)
           newText
       }
       ```

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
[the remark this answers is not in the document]
smichel17: Personally:

- If it were *my* PR, I probably would, because I'm comfortable mucking around with git; it's (usually) not that much effort for me to pull out changes into a separate commit, which can go in a separate PR.
- If it were my *repository*, I would not insist on you doing it; I would merge as-is. I would also suggest that next time you commit logical pieces separately (e.g. "reformat with Android Studio"), because it makes it easier to revert parts individually.
    - <details><summary>Note on git tooling</summary>
       Android Studio's git plugin makes this quite difficult, because it automatically stages every change you make. If you use the git command-line (or most other git tools), you can pick and choose which of your current changes to add to the commit. This way, if you've done a some changes both before and after reformatting, you don't need to revert it all in order to commit. I use `git add --patch` (`-p` for short) to stage just the changes I care about.
       </details>

In this case— @westnordost do you care about this reformat being mixed with the rest of the PR?

westnordost:
A note to reviewers: Please try to be brief and to keep trivial discussions about code style to a minimum. E.g. requesting to add trailing commas is borderline useless. We should value contributors work and that includes to not let them jump through extra hoops. Contributors don't know which review comments must be solved for the PR to be merged and which are optional and/or a matter of taste, so they are generally forced to heed every single comment which just creates more work for them.

The consequence for those of us who do the review work is to omit (such) optional stuff from the review completely, clearly mark it as optional or just mention it in the concluding post ("that it would be nice IMO if...").

I haven't looked at this review again yet because I think it is too much if three reviewers trip over the code all at once. @goldbattle let me know when the points of the previous reviewers have been addressed and I'll have a (final) look.

goldbattle:
I think they were addressed to be best of my ability. Feel free to look, thanks!

FloEdelmann (approved):
The code looks fine for me now. I haven't tested the feature in the emulator or on a device though.

app/src/main/java/de/westnordost/streetcomplete/map/LocationAwareMapFragment.kt
```diff
@@ -38,6 +39,12 @@ open class LocationAwareMapFragment : MapFragment() {
     /** The GPS trackpoints the user has walked */
     private var tracks: MutableList<ArrayList<Location>>
 
+    /** If we are actively recording track history */
+    var tracksRecording = false;
```
westnordost, line 43:
```suggestion
    var tracksRecording = false
```

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
```diff
@@ -55,17 +60,33 @@ class NoteEditsUploader @Inject constructor(
...
+            imageText = uploadAndGetAttachedPhotosText(edit.imagePaths)
+            noteEditsController.updateData(edit, "images", imageText)
+        }
+
+        // try to upload the tracks
+        var tracksText = edit.uploadedDataMap.getOrDefault("tracks", "")
+        if(tracksText.isEmpty()) {
+            tracksText = uploadAndGetAttachedTracksText(edit.tracks, edit.text)
+            noteEditsController.updateData(edit, "tracks", tracksText)
+        }
```
westnordost, line 76: Okay, I understand now what this "uploadedDataMap" is for. It enables that if one step of the possibly two steps prior to actually creating the note (due to a network problem) fails, the same images / tracks are not uploaded again on next try

That's fine, though not really important or actually a little problematic because
1. uploaded photos not referenced by any notes will be cleaned up quite quickly
2. a few duplicately uploaded traces in case of rare upload connectivity problems don't really make a difference. GPX tracks are tiny in size
3. current implementation is not perfect anyway, because there may be several photos for upload and the upload may fail anytime while uploading the multiple photos
4. Because of point 1, upload of photos and creation of note must happen promptly after one another. Otherwise, the final note will just contain dead links (or worse, image activation fails, not sure what happens then)
So question is if it is really worth the added complexity.

However, if you opt to keep this logic:
1. I'd very much prefer if this is not a "nameless" JSON map but just two columns in the table: `uploadedImagesText` and `uploadedTracksText`
2. What if there are no images to attach / no tracks to attach? Then, the upload functions and the updateData function will be called regardless

westnordost (changes requested):
So, there are 2 important things (permission request handling, uploadedDataMap) to do and the rest are just a few uninmportant but easy to change suggestions for improvement.

app/src/main/java/de/westnordost/streetcomplete/view/dialogs/RequestPermissionUpgradeDialog.kt
```diff
@@ -0,0 +1,24 @@
...
+class RequestPermissionUpgradeDialog(context: Context) : AlertDialog(context, R.style.Theme_Bubble_Dialog) {
+    init {
+        val view = LayoutInflater.from(context).inflate(R.layout.dialog_permission_upgrade, null, false)
+        setView(view)
+        setButton(BUTTON_POSITIVE, context.getString(android.R.string.ok)) { _, _ ->
+            val intent = Intent(context, UserActivity::class.java)
+            intent.putExtra(UserActivity.EXTRA_LAUNCH_AUTH, true)
+            context.startActivity(intent)
+        }
+        setButton(BUTTON_NEGATIVE, context.getString(R.string.later)) { _, _ -> }
```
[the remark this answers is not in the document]
goldbattle: This gives me this:
```
Overload resolution ambiguity.

All these functions match.
public open fun setButton(whichButton: Int, text: CharSequence!, listener: DialogInterface.OnClickListener!): Unit defined in de.westnordost.streetcomplete.view.dialogs.RequestPermissionUpgradeDialog
public open fun setButton(whichButton: Int, text: CharSequence!, msg: Message!): Unit defined in de.westnordost.streetcomplete.view.dialogs.RequestPermissionUpgradeDialog
```

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
```diff
@@ -21,47 +21,44 @@ class TracksMapComponent(ctrl: KtMapController) {
...
-        tracks.add(ArrayList())
+    fun startNewTrack(record: Boolean) {
+        tracks.add(Track(ArrayList(), record))
         putAllTracksInOldLayer()
     }
 
     /** Set all the tracks (when re-initializing) */
     fun setTracks(tracks: List<List<Location>>) {
-        this.tracks = tracks.map { track -> track.map { it.toLngLat() }.toMutableList() }.toMutableList()
+        this.tracks = tracks.map { track -> Track(track.map { it.toLngLat() }.toMutableList(), false) }.toMutableList()
```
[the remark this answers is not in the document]
goldbattle: Yes, as to my understanding this reloading is only when the app is re-opened. In this case, no recording should be active, and I am not sure how useful old recording segments will be compared to just showing were you have been (basically I didn't feel it was worth adding this).

app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
```diff
@@ -235,6 +237,16 @@ class MainActivity : BaseActivity(),
...
+                } else if (e is AuthorizationTracesException) {
+                    // user has not upgraded their permissions to have GPS trace (or is not logged in)
+                    // if the user is not logged in, then we should just ask them to login which will provide all permissions
+                    val isLoggedIn = userLoginStatusController.isLoggedIn
+                    userLoginStatusController.logOut()
+                    if(isLoggedIn) {
+                        RequestPermissionUpgradeDialog(this@MainActivity).show()
+                    } else {
+                        RequestLoginDialog(this@MainActivity).show()
+                    }
```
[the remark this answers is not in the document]
goldbattle: How about this logic:

**Current users**: Has `HAS_UPLOAD_TRACES_PERMISSION` by default set to false. Will see a popup when they try to create a new track that asks if they want to login to enable the feature. Once they login they can then actual start recording tracks.

**New users**: All required permissions are requested, and `HAS_UPLOAD_TRACES_PERMISSION` is set to true on login. They can then upload traces are they see fit.


Comments:
- Current users that do not have internet will be unable to record traces
- Users are not logged out till they say yes to the dialog

app/src/main/java/de/westnordost/streetcomplete/view/dialogs/RequestPermissionUpgradeDialog.kt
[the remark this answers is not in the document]
westnordost: Oh okay, so this only works with the builder

app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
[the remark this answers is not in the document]
westnordost: Where is the difference to what I wrote?

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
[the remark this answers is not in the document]
westnordost: It's when the view is recreated. This can happen any time when a user tabs out and then in again.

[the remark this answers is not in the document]
goldbattle: Oh, let me test and get back to you on this then.

app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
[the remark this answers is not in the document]
goldbattle: I think it is the same, just making sure I understand the logic so I can implement.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
goldbattle: After pondering over the weekend, I have to agree with you that this isn't that important / necessary.  If there isn't any issue with duplicate images then my original concern is invalid. I think just removing the logic and making things simpler is the way to go.

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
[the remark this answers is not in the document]
westnordost: **can** - it will be hard to reproduce because the Android system decides when to do that and when to not do it (based on memory constraints or whatever). I know one way to "force" this - remove from the AndroidManifest that the MainActivity handles `orientation` changes itself for testing. If it does not handle it itself, rotating the phone will recreate the view.

goldbattle:
Not sure if I did the best implementation (not that familiar with injections), but as per https://github.com/streetcomplete/StreetComplete/pull/3573#discussion_r778908679 the logic is to have a preference variable that has recorded if a user has upgraded their permissions to include GPX tracks. If so, then the user can record a track, otherwise the user will be asked to upgrade to the newer permissions.

app/src/main/java/de/westnordost/streetcomplete/map/components/TracksMapComponent.kt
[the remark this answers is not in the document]
goldbattle: It looks like this also stops the recording since this isn't something that is stored in memory. It took a few turns, but eventually it recreated the view and you are correct that it looses the track history of recordings, but it also stops the recording. I am not sure how best to fix, as I think I need to record someplace that the user has started recording a track? or maybe it isn't worth fixing this? thoughts?

https://user-images.githubusercontent.com/2222562/149715792-f86758ac-65b7-4539-95e4-46bc25df4542.mp4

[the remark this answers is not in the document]
westnordost: Hmm,

for the use case that one just wants to add a GPX track to a note showing a short driveway/connecting path, I think it is okay if the recording is cancelled when the user changes apps and changes back after some time because the record feature can be understood to be "in the process of creating a note". Tabbing out in-between can be interpreted as that the user doesn't want to create a note after all.

On the other hand, users may want to add longer records as well, to create notes like "I walked every path in this forest/park, some paths seem to be missing!" or whatever. So in this case, the record would go on for a while and it is normal during a mapping session to sometimes tab out to some other app. Every tabbing out would then endanger that the whole track is being discarded (depending on how aggressive the Android system will destroy apps not in the foreground).

So, in conclusion, I would say that the recording track should indeed be saved into the instance state.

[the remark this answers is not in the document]
goldbattle: I think the user would still expect that the track keeps recording if they switch apps for a few seconds. How do you recommend I record the state in the fragment? I am not sure how to even approach in implementation.

*Side comment: There is the additional case that switching apps will stop getting location information since the app is in the background now and we only have location info when in the forground.*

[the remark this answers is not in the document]
westnordost: > here is the additional case that switching apps will stop getting location information since the app is in the background now and we only have location info when in the forground.

Yes I know, this is deliberate though.

FloEdelmann:
@goldbattle I've updated this branch and fixed the merge conflicts after #3687 was merged.

westnordost:
@goldbattle it looksl ike this PR is stuck. Should I mark this as "help requested" so maybe other people interested in getting this feature live could try to finish it?

goldbattle:
Yeah that would be great. I can fix any merge conflicts, but adding and
debugging the instance state was a bit above me.

FloEdelmann (changes requested):
@goldbattle Unfortunately, I can't make it to the point where I can test saving the instance state:

After login (via the dialog after trying to start a track, and via the button in the empty profile) and granting all asked permissions, an error is thrown:

 

Did I mess something up during the merge? Can you have a look at this?

goldbattle:
You will need to use some test perms that have auth to upload traces (will also need to be done with the main creds before merging). See https://github.com/streetcomplete/StreetComplete/pull/3573#issuecomment-991797076

Here is an application I created to test with (de/westnordost/streetcomplete/data/user/UserModule.kt):

```
    private const val CONSUMER_KEY = "tbHqDSVojwRZT4hQ2jMsOzdN4i1YsTQqTvAVeFcD"
    private const val CONSUMER_SECRET = "X3tc6G6aD4DPWVOhmvQowFdvPzwXvebSYnJStHWL"
```

westnordost:
@FloEdelmann note this was another "must" feature that hasn't been implemented in this PR yet:

When successfully authenticating with the new requested permissions (+ upload traces), the app should persist that it has this permission now. When the user is logged in, he **doesn't** have that permission and tries to use this feature, a dialog will pop up that explains that the permission is missing and asks the user to login again.
The reason for this is so that all current logged in users (who do not have the permission) are not kicked out as long as they don't yet use that feature.

goldbattle:
That logic should be there. If the user doesn't have the permissions yet, then the user will have a dialog asking if they want to relogin to get them. They can decline and thus stay login (but can't record any traces until they relogin with the new perms). The preference variable `OSM_HAS_UPLOAD_TRACES_PERMISSION` is set to true on login which should be persistent.


See commit https://github.com/streetcomplete/StreetComplete/pull/3573/commits/6988216a2e5c9cd192b31637cc39003f77e5cd8e

Dialog has this text:

> To upload a GPS trace, you need to authorize with your OSM user account to publish traces. This will log you out and ask you to re-login with an internet connection. Authorize now?

app/src/main/res/values/strings.xml
```diff
@@ -228,6 +231,7 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
 
     <string name="confirmation_authorize_now">"You need to authorize with your OSM user account to publish your answers. Authorize now?"</string>
     <string name="confirmation_authorize_now_note2">You can also do this later on the profile screen.</string>
+    <string name="confirmation_permission_upgrade">"To upload a GPS trace, you need to authorize with your OSM user account to publish traces. This will log you out and ask you to re-login with an internet connection. Authorize now?"</string>
```
rugk, line 234:
```suggestion
    <string name="confirmation_permission_upgrade">"To upload a GPS trace, you need to authorize with your OpenStreetMap user account to publish traces. This will log you out of the app and ask you to re-login. Proceed now?"</string>
```

* removed internet connection: it's quite obvious for anyone who uses a smartphone that a login requires an internet connection
* do not duplicate the word "authorize" so much -> replace one with "proceed" (IMHO BTW, even better a button named "Proceed" could be that is indicate that the action is somewhat more critical – after all, you could get kicked out of the app if you carelessly tap on "Ok" or whatever is currently shown as the button name. However, I'd still say it's not _that_ criticial after all, as you can use SC without being logged in, for some time at least.)
*  "logged out of the app" sounds better IMHO as it makes clear where you're logged out of…
* removed the abbrevation OSM (quite a technical slang/term) and replaced by full name

Disclaimer: I'm no native English speaker.

app/src/main/assets/map_theme/jawg/streetcomplete.yaml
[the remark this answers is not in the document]
rugk: Oh actually YAML has a way of "re-use groups of values". I would not say it's good :upside_down_face: , after all it's kinda strange/crazy syntax, but well… that's upon you to decide, after all. (it's also a hard thing to implement and one may have to get accustomed to it first)

It's called "YAML anchors": https://yaml.org/spec/1.2-old/spec.html#id2785586
If that's what you want/helps here…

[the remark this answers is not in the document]
westnordost: Uh yeah, don't use that. Not even sure if tangram-es (which has to parse that YAML) is able to parse that.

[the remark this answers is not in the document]
rugk: AFAIK every standard-compliant parser can and should be able to parse this. If it cannot then well… it's kinda broken (never had this case though)

So is not it worth a try? Or is there any other argument not to use it?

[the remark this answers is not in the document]
westnordost: For me: no. You can try it. If it works, you can create a PR and we can discuss it.

app/src/main/res/values/strings.xml
rugk: I'd more agree with the latter part, an audience that does not know that an OSM login requires and internet connection would hardly know what a "GPS trace" is… :upside_down_face:

FloEdelmann:
I'm sorry, but after looking at the code for like an hour, I can't figure out what to actually save in the current instance state. `TracksMapComponent` and `LocationAwareMapFragment` both seem to record the current track? Currently, I have no time to dig deeper. @goldbattle maybe you want to give it another go?

westnordost:
@FloEdelmann The issue is that the gps track that is being recorded is kept in the fragment, and since we don't use ViewModels to keep the data, it will be gone if the fragment is recreated because the view is recreated unless the data of the recording is put into the instance state of that fragment and then restored when recreated.

So, what do you mean exactly? Don't you know how to do it? Or is this already done according to your reading?

goldbattle:
Sorry about the slow response, but I think I implemented it to perform as expected. It already stored the tracks to the instance, so I just needed to add the boolean if recording was active (and make the button show up if it was on re-creation).

There was a small issue with that tracks that were saved using the `tracks.flattenToNullTerminated()` have an empty list appended to them which was preventing me from appending tracks to the current track recording. I removed this in the case when recording, as it is nice when you re-enter the app to see that a new track was started (e.g. there is no big jump).
https://github.com/streetcomplete/StreetComplete/pull/3573/commits/7a04dea42fef72d95049d5ef6209cf3c3117d56c#diff-eaf051bafe9d96639e9e82c9ff06787a6a6fd00f9f7a6ac972ef35ce992c0e2aR113-R122

I will test on my phone and let you know how it goes.

app/src/main/java/de/westnordost/streetcomplete/screens/main/MainFragment.kt
```diff
@@ -265,6 +270,7 @@ class MainFragment :
     override fun onMapInitialized() {
         binding.gpsTrackingButton.isActivated = mapFragment?.isFollowingPosition ?: false
         binding.gpsTrackingButton.isNavigation = mapFragment?.isNavigationMode ?: false
+        binding.stopTracksButton.isVisible = mapFragment?.tracksRecording ?: false
```
goldbattle, line 273: This ensures that if the user exits and the app view is re-created, the recording button will show if the recording is still active.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
```diff
@@ -93,10 +105,22 @@ open class LocationAwareMapFragment : MapFragment() {
...
+            with(savedInstanceState) {
+                displayedLocation = getParcelable(DISPLAYED_LOCATION)
+                tracksRecording = getBoolean(TRACKS_IS_RECORDING)
+                // It seems that the last list element will be an empty one normally
+                // If it is an empty one we can just remove it so we can keep on recording
+                val nullTerminatedTracks =
+                    getParcelableArrayList<Location?>(TRACKS) as ArrayList<Location?>?
+                if (nullTerminatedTracks != null) {
+                    tracks = nullTerminatedTracks.unflattenNullTerminated()
+                    if (tracksRecording && tracks.last().isEmpty()) {
```
goldbattle, line 119: I am not sure if this is since a `unflattenNullTerminated` is used to encode / decode, but if you encode 2 elements in an array and then decode it, it will have a size of 3 with the last one being empty. If recording, we want to append locations to the last track, thus we can't have this new array be appended....

goldbattle:
I just tested on my phone and it worked as expected (albeit on a fresh dev install, so no upgrade perm testing).

westnordost:
@FloEdelmann since you already touched the code here, it may be more efficient if you reviewed the finished PR. Are you available for that?

westnordost:
@goldbattle when you enter the developer options on your phone, there should be an option "Don't keep activities". If you turn this on, does the feature work as expected when you e.g. switch to another app and then back some time later?

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
```diff
@@ -44,6 +45,15 @@ open class LocationAwareMapFragment : MapFragment() {
...
     private var tracks: MutableList<ArrayList<Location>>
 
+    /** If we are actively recording track history */
+    var tracksRecording = false
+        private set
+
+    /** The GPS trackpoints the user has recorded */
+    private var _tracksRecorded: ArrayList<Trackpoint>
+
+    val tracksRecorded: List<Trackpoint> get() = _tracksRecorded
```
FloEdelmann, line 55: I think it would be helpful to rename these variables to make differentiating between `Boolean` and `List<Trackpoint>` easier:

```diff
- tracksRecording
+ isRecordingTracks

- _tracksRecorded
+ _recordedTracks

- tracksRecorded
+ recordedTracks
```

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
FloEdelmann: Better add a comment in the code instead of in this PR only:

```suggestion
                    // unflattenNullTerminated creates an empty list item at the end
                    if (tracksRecording && tracks.last().isEmpty()) {
```

And why is the condition `tracksRecording && …`? Shouldn't the last empty list item always be removed?

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
```diff
@@ -159,6 +183,33 @@ open class LocationAwareMapFragment : MapFragment() {
...
+        _tracksRecorded.clear()
+        tracks.last().forEach {
+            // Emulator has zero altitude:
+            // https://stackoverflow.com/q/65325665/7718197
+            // Time here is in milliseconds
+            _tracksRecorded.add(
+                Trackpoint(
+                    LatLon(it.latitude, it.longitude), it.time, it.accuracy, it.altitude.toFloat()
+                )
+            )
```
FloEdelmann, line 207:
```suggestion
            _tracksRecorded.add(
                Trackpoint(
                    LatLon(it.latitude, it.longitude),
                    it.time, // in milliseconds
                    it.accuracy,
                    it.altitude.toFloat() // always zero in emulator: https://stackoverflow.com/q/65325665
                )
            )
```

app/src/main/assets/map_theme/jawg/streetcomplete.yaml
```diff
@@ -146,20 +156,29 @@ layers:
...
         current:
-            filter: { old: [false] }
-            draw:
-                track-lines:
-                    color: global.track_color
-                    width: [[14, 6px],[18, 12px]]
-                    collide: false
-                    join: round
-                    order: 1000
+          filter: { old: [ false ], record: [ false ] }
```
FloEdelmann, line 159: Can you restore the previous indentation (4 spaces) for all your changes in this file please?

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEdit.kt
```diff
@@ -31,8 +32,11 @@ data class NoteEdit(
     /** whether this edit has been uploaded already */
     override val isSynced: Boolean,
 
-    /** Whether the images attached still need activation. Already true if imagePaths is empty */
-    val imagesNeedActivation: Boolean
+    /** whether the images attached still need activation. Already true if imagePaths is empty */
+    val imagesNeedActivation: Boolean,
+
+    /** attached GPS location tracks */
+    val tracks: List<Trackpoint>,
```
FloEdelmann, line 39: This only stores a single track, right? (One track = list of trackpoints)

Then please rename this variable from `track` to `track` and adjust calling functions, the database column name and code comments accordingly.

I think it was this missing distinction that confused me a lot when I tried to implement saving to the instance.

app/src/main/res/values/strings.xml
FloEdelmann: I like @rugk's suggestion and think it's an improvement.

@goldbattle Feel free to explicitly mention "internet connection" if you think it's necessary; I have no strong opinion on that.

FloEdelmann (changes requested):
> Sorry about the slow response

Don't worry! Better late than never :wink: 
Thanks for implementing saving to the instance now!

I don't have time for a full review with testing. But here is a code review at least :slightly_smiling_face:

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
goldbattle: It should always be removed, if recording, yes.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEdit.kt
goldbattle: Renamed everything to `track` and left the existing `trackpoints` names as is.

app/src/main/assets/map_theme/jawg/streetcomplete.yaml
goldbattle: Not sure why it auto formats to 2spaces. I have made it all 4space hopefully.

goldbattle:
> @goldbattle when you enter the developer options on your phone, there should be an option "Don't keep activities". If you turn this on, does the feature work as expected when you e.g. switch to another app and then back some time later?

It still works as expected when I have this enabled.

@FloEdelmann I have made the requested changes, let me know if anything else.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
```diff
@@ -93,10 +105,23 @@ open class LocationAwareMapFragment : MapFragment() {
...
+                displayedLocation = getParcelable(DISPLAYED_LOCATION)
+                isRecordingTracks = getBoolean(TRACKS_IS_RECORDING)
+                // It seems that the last list element will be an empty one normally
+                // If it is an empty one we can just remove it so we can keep on recording
+                val nullTerminatedTracks =
+                    getParcelableArrayList<Location?>(TRACKS) as ArrayList<Location?>?
+                if (nullTerminatedTracks != null) {
+                    tracks = nullTerminatedTracks.unflattenNullTerminated()
+                    // unflattenNullTerminated creates an empty list item at the end
+                    if (isRecordingTracks) {
```
FloEdelmann, line 120: I suggested to change the condition `tracksRecording && tracks.last().isEmpty()` to `tracks.last().isEmpty()`, not to `tracksRecording`. Sorry if I wasn't clear.

I don't see why the last list item shouldn't be removed when not recording tracks.

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
goldbattle: This might be a preference thing, but if I leave the app and re-enter, it is nice to *not* have the big "jump" in the trajectory and just have the track start recording again since GPS doesn't record in the background. This is why I left it in the non-recording the case. I can change to your suggestion as that makes it more consistent for both cases if you think that is better.

FloEdelmann: Ah I see, so there is a wanted behavior difference. Let's add that to the comment then.

```diff
- // unflattenNullTerminated creates an empty list item at the end
+ // unflattenNullTerminated creates an empty list item (i.e. a new track) at the end.
+ // This is fine if the track is not being recorded.
```

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
```diff
@@ -93,10 +105,23 @@ open class LocationAwareMapFragment : MapFragment() {
...
+            with(savedInstanceState) {
+                displayedLocation = getParcelable(DISPLAYED_LOCATION)
+                isRecordingTracks = getBoolean(TRACKS_IS_RECORDING)
+                // It seems that the last list element will be an empty one normally
+                // If it is an empty one we can just remove it so we can keep on recording
+                val nullTerminatedTracks =
+                    getParcelableArrayList<Location?>(TRACKS) as ArrayList<Location?>?
+                if (nullTerminatedTracks != null) {
+                    tracks = nullTerminatedTracks.unflattenNullTerminated()
+                    // unflattenNullTerminated creates an empty list item at the end
```
FloEdelmann, line 119:
```suggestion
                    // unflattenNullTerminated creates an empty list item (i.e. a new track) at the end.
                    // This is fine if the track is not being recorded.
```

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/LocationAwareMapFragment.kt
```diff
@@ -93,10 +105,23 @@ open class LocationAwareMapFragment : MapFragment() {
...
-        val nullTerminatedTracks = savedInstanceState?.getParcelableArrayList<Location?>(TRACKS) as ArrayList<Location?>?
-        if (nullTerminatedTracks != null) {
-            tracks = nullTerminatedTracks.unflattenNullTerminated()
+        // Restore value of members from saved state
+        if (savedInstanceState != null) {
+            with(savedInstanceState) {
+                displayedLocation = getParcelable(DISPLAYED_LOCATION)
+                isRecordingTracks = getBoolean(TRACKS_IS_RECORDING)
+                // It seems that the last list element will be an empty one normally
+                // If it is an empty one we can just remove it so we can keep on recording
```
FloEdelmann, line 114: This comment can be removed then.

```suggestion
```

FloEdelmann (approved):
Looks good to me now (still only code review, not an extensive testing) :)

app/src/main/res/values/strings.xml
```diff
@@ -228,6 +231,7 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
 
     <string name="confirmation_authorize_now">"You need to authorize with your OSM user account to publish your answers. Authorize now?"</string>
     <string name="confirmation_authorize_now_note2">You can also do this later on the profile screen.</string>
+    <string name="confirmation_permission_upgrade">"To upload a GPS trace, you need to authorize with your OpenStreetMap user account to publish traces. This will log you out of the app and ask you to re-login. Proceed now?"</string>
```
westnordost, line 234:
```suggestion
    <string name="confirmation_permission_upgrade">"You need to authorize this app to publish track recordings on OpenStreetMap. This requires to log in again. Proceed?"</string>
```

I found "will log you out of the app" to be misleading. Also made it a bit shorter.

westnordost (approved):
Changed just a few tiny things

westnordost:
Going to merge this now. The PR is really of excellent quality, great work!

After merge, I am going to add/change a few things:
- adapt the tests for NoteEditsDao to also include testing for persisting the trackpoints
- try if the different color for recording-tracks can be achieved with less YAML tangram-styling code
- remove the usage of "with" in LocationAwareMapFragment because (personally) I think "with" is not a good idea and it is not used anywhere else in the code

goldbattle:
Great, thanks so much, sorry it took so long! Make sure the OSM API permissions get updated for the application before you release otherwise it will get into an infinite loop asking for permissions.

westnordost:
Right, thanks for the hint!

goldbattle:
See the bottom of this comment:
https://github.com/streetcomplete/StreetComplete/pull/3573#issuecomment-991797076

---

It needs to have additional permissions requested from the OSM api so that it can upload and get info about the GPX track.

gpx_sc_example_03

Here is an application I created to test with (`de/westnordost/streetcomplete/data/user/UserModule.kt`):
```
    private const val CONSUMER_KEY = "tbHqDSVojwRZT4hQ2jMsOzdN4i1YsTQqTvAVeFcD"
    private const val CONSUMER_SECRET = "X3tc6G6aD4DPWVOhmvQowFdvPzwXvebSYnJStHWL"
```

westnordost:
Fuck...

I accidentally deleted the OAuth1 profile because I first created a new profile under the same name with the new permissions but only then noticed I could also just edit the old profile to include the new permissions. So then I deleted the "old" profile (the second in the list). But it looks like the newest profile is displayed first in the list :-|

So, for damage control, gotta create a update for v43 now that uses the new profile because the old profile is irrevocably gone, looks like.

westnordost:
Though before I hastily do this, better test what exactly happens code-/api-wise in the app if it uses an unknown oauth key...

westnordost:
Okay, the hotfix v43.2 is deployed. It logs out the user automatically on first start so that the user needs to login again with the new OAuth profile. 😥

On the plus side, I could remove the permission upgrade request dialog, check and strings, see https://github.com/streetcomplete/StreetComplete/commit/31ebb42cb670738d943a299b88bb3e5ec6def361 becaue this upgrade plan obviously failed (my fault).

goldbattle:
Yeah that seems the cleanest now since all users will have the new oauth (with track perms) so no need for the added complexity. Sorry for the all the trouble :(

westnordost:
Thanks, but I caused the trouble onto myself, should have been more cautious when editing/deleting the profile.

matkoniecz:
I committed https://github.com/matkoniecz/Zazolc/commit/06704ea87c52ca981ad4186ed895265c41734a95 to unbreak tests, hopefully without breaking anything new

PR #3257, opened 2021-09-04
Add smoothness quest

Helium314:
This is a very first draft of the smoothness quest, fixes #1630.

What is done so far are the basics, i.e. element selection and generating the answer form from surface and highway tags.
Adding images and descriptions for different smoothnesses is easy and can depend on `highway` and `surface` tags.

There are two _other answers_
* one is supposed to tell the user to ignore small obstacles
* one is to report that the images are shown for the wrong surface, and once implemented should remove the surface tag if if's wrong (like a similar _other answer_ for the housenumber quest that changes the building to `building=yes`)

What is missing: basically all images and descriptions for answers. There are some in this first version, but it's just poor and barely suitable placeholders that need to be replaced.

Adding good images and text is a lot of work, so any help is appreciated (maybe by @rhhsm who did a lot on this topic on OSM wiki?).

Regarding the images I have some questions @westnordost 
* I started using `surface_<type>_<smoothness>`, to be in line with the existing `surface_<type>` from the surface quest. Is this ok, or do you prefer a separate naming scheme?
* how do you generate the density versions? I didn't find such functionality in Android Studio
* should the original images go somewhere in _/res/graphics_?
* I can't decide between 2 or 3 images per row in the answer form. With 3, I fear it will be easier to miss relevant details. With 2, the answer form is really tall, and the existing surface image can't be used (just get padded with empty space). Do you have some recommendation on this?

Once the questions about the images have been clarified, I'll start adding images from links in #1630.

For the texts I still need need to do some more live testing to see which style is actually helping.
I tried imagining how it would be to go on specific parts of a road on a racing bike or inline skates, but I didn't find it very helpful...

PR #3576, opened 2021-12-08
make rejecting cancellation more clear

matkoniecz:
There is a "Discard input?" menu appearing after user fills answer, at least partially and tries to leave answer window.

It has options "cancel" and "discard" with "cancel" being not entirely clear - as user is cancelling filling of the quest. And "cancel cancelling" is tricky.

Situation is even worse in some translations where even the best found text is more confusing than English version.

see https://github.com/streetcomplete/StreetComplete/discussions/3396 where it was discussed

screen02

Known negative effects: translation required again

I am aware of `<string name="quest_generic_hasFeature_no">"No"</string>` but in some languages this fields may require different translations.

### DIFF
--- a/app/src/main/res/layout/quest_buttonpanel_done_cancel.xml
+++ b/app/src/main/res/layout/quest_buttonpanel_done_cancel.xml
@@ -10,7 +10,7 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         style="@style/BottomSheetButtonBarItem"
-        android:text="@string/confirmation_discard_negative" />
+        android:text="@string/confirmation_discard_negative2" />
 
     <Button
         android:id="@+id/doneButton"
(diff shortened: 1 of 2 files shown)


FloEdelmann (approved):

smichel17 (approved):
> Known negative effects: translation required again

I am not sure if this is a good practice, but one option would be to pre-seed the value of `confirmation_discard_negative2` with whatever is translated for `quest_generic_hasFeature_no`, and then mark `confirmation_discard_negative2` as fuzzy / needs-review. I did something like this once when I was changing strings in Red Moon.

matkoniecz:
Sadly there was some accident with translations strings and nearly everything got marked as fuzzy - which is still not entirely repaired, what makes it less useful :/

mnalis:
Speaking with my translator hat on, I support that new answer which should be translated from scratch. Perhaps it should be bundled with some other new translation strings (like Smoothness quest https://github.com/streetcomplete/StreetComplete/pull/3257 ?) and translators be notified some time before so they can update it, then it should work well (for languages which are kept up-to-date at least). And beta release with new translation strings would be good idea before pushing an official one...

Stuff being marked as fuzzy is better than nothing, but for example I have several dozen translations I've myself marked as fuzzy as I'm not completely happy with them... I most probably won't see a new one pushed by SC there for quite some time, unless I'm specifically warned to go look for it.

matkoniecz:
> And beta release with new translation strings would be good idea before pushing an official one...

That is happening already in every release cycle.

Issue #2461, opened 2021-01-06
Layers - Pitch


westnordost:
# Idea

The idea is to make available a tool that makes contributing and maintaining data easy *and efficient* to do on a survey, focussing on one aspect of the map at a time.

# Consideration

On the smartphone, there are three issues that further challenge usability:
- a small screen severely limits the number of things that can be displayed at once
- a fat finger as pointing device is not only much less precise than a computer mouse, the same finger obscures part of the screen, most importantly the part where the tap (or drag) happens
- the user is currently out and about

StreetComplete is so easy and quick to use, because amongst other things, it severely limits what data the user can see and change.
The usability of "fully-featured" editors like iD, JOSM and Vespucci all suffer from that the data they work with is so complex, as they show all of it at the same time. New mappers, especially from GIS backgrounds are often surprised that there is no inherent concept of layers in OpenStreetMap.

For mappers who want to map/update "everything" on a survey, there is and there will probably never be a more efficient method than taking a lot of photos (or for mapillary) and then adding all the info visible on the photos to the map via JOSM.

The **target audience** for this idea are mappers who want to focus on **completing and keeping up-to-date one aspect of OpenStreetMap efficiently** at a time, i.e. per survey.

This requires the app to be able to 
- show the current data in an uncluttered and clear form to the user, only that "layer"
- and the ability to add/update this data quickly and easily so that it can be done on the go without stopping and so that subsequent data entry at the computer at home is not necessary

# Rough UI Concept

The UI / and input form can and should be customized for each layer as long as it enhances the overview and makes the input easier.

![Untitled](https://user-images.githubusercontent.com/4661658/103790465-86d2f280-5041-11eb-9f33-0c6d7b06aa02.png)
![Untitled3](https://user-images.githubusercontent.com/4661658/103790470-889cb600-5041-11eb-84ed-daa3b3a98287.png)
![Untitled2](https://user-images.githubusercontent.com/4661658/103790479-8aff1000-5041-11eb-8242-64f6e23483f2.png)

1. the background map should probably be desaturated. No quests are displayed. At least, the data for the layer must stand out in (flashy) colors.
2. the legend lists what each color stands for. Ideally, this should be retractable into a tappable icon, maybe have it expanded every time one enters this layer
3. selection is highlighted in some way. Maybe like in iD (yellow glow) or maybe as if it is floating above the map (shadow). 
[...]

PR #3582, opened 2021-12-09
house name and house number answer at once

### LINKED ISSUE #2540: Add another answer "House number and a name" which gives a two field form, for house name and house number.
**Use case**
As per:
https://wiki.openstreetmap.org/wiki/Key:addr#Tags_for_individual_houses

> addr:housename - The name of a house. This is sometimes used in some countries like England instead of (or in addition to) a house number.

There's no way to do a sensible quest, but the existing AddHousenumber one could be extended if one was spotted while surveying.

This would cover both single family homes "Pretty Name Cottage, 5 Small Close" and blocks of flats "Accommodation Towers, 20-25 Busy Road".

**Proposed Solution**
Add an other answer "House number and a name" which gives a two field form, for house name and house number.

Users can pick this option if both are present rather than having to choose one or the other and lose the bit of data they didn't enter, or add a note and then someone have to manually copy it into the object.

matkoniecz:
Add another answer "House number and a name" which gives a two field form, for house name and house number.

fixes #2540

Tested case where housenumber is entered, form switched to housename - it record just housename as it should

screen
screen02
screen03

NOTE: PR code needs one filler string in unrelated quest to compile


andrewharvey:
Awesome, any thoughts on how SC could help capture those housenames not captured by SC originally because the mapper could only enter the number?

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -49,13 +51,25 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
...
+            isHousename -> {
+                setLayout(R.layout.quest_housename)
+            }
+            isHousenameAndHousenumber -> {
+                setLayout(R.layout.quest_housename_and_housenumber)
+            }
+            else -> {
+                setLayout(R.layout.quest_housenumber)
+            }
+        }
```
smichel17, line 65: On mobile so can't use the "suggest" feature, but I might pull out the `setlayout`

```kt
val layout = when {
  isHousename -> R.layout.quest_housename
  isHousenameAndHousenumber -> R.layout.quest_housename_and_housenumber
  else -> R.layout.quest_housenumber
}
setLayout(layout)
```

smichel17:
Currently: only solution is to wait for resurvey.

I can't think of anything which could be done for this *specifically*, but in general, either the layers feature or my proposed alternative, "bulk edit mode" (https://github.com/streetcomplete/StreetComplete/issues/2461#issuecomment-755638075) would allow editing at times when a quest would not normally be shown.

matkoniecz:
> Awesome, any thoughts on how SC could help capture those housenames not captured by SC originally because the mapper could only enter the number?

This is likely not feasible to do as a quest. Asking for housenumber for all buildings with house name where someone answered with partial info instead of making note is not a good idea as vast majority has just housename. Unless in some regions it is safe to assume that building always has house number and house name is a possible addition.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -41,6 +42,7 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
     private var subtractButton: View? = null
 
     private var isHousename: Boolean = false
+    private var isHousenameAndHousenumber: Boolean = false
```
FloEdelmann, line 45: Is it possible to use a single enum variable?

Now there are four possible options (false+false, false+true, true+false, true+true), while only three are used (false+false, false+true, true+false).

An enum variable would make that more clear and also allow Kotlin / the IDE to suggest exhaustive `when` branches.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
FloEdelmann: That'd increase the readability a lot! But I think there is no need for the intermediate variable, so better:

```kt
setLayout(when {
  isHousename -> R.layout.quest_housename
  isHousenameAndHousenumber -> R.layout.quest_housename_and_housenumber
  else -> R.layout.quest_housenumber
})
```

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -241,7 +261,14 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
     }
 
     private fun createAnswer(): HousenumberAnswer? =
```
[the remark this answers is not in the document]
FloEdelmann: Or maybe even simpler (but I don't know if the meaning/functionality remains the same then):

```kt
    private fun createAnswer(): HousenumberAnswer? {
        val houseName = houseNameInput?.nonEmptyInput
        val houseNumber = houseNumberInput?.nonEmptyInput
        val conscriptionNumber = conscriptionNumberInput?.nonEmptyInput
        val streetNumber = streetNumberInput?.nonEmptyInput
        val blockNumber = blockNumberInput?.nonEmptyInput

        return when {
            houseName != null && houseNumber != null   -> HouseNameAndHouseNumber(houseName, houseNumber)
            houseName != null                          -> HouseName(houseName)
            conscriptionNumber != null                 -> ConscriptionNumber(conscriptionNumber, streetNumber) // streetNumber is optional
            blockNumber != null && houseNumber != null -> HouseAndBlockNumber(houseNumber, blockNumber)
            houseNumber != null                        -> HouseNumber(houseNumber)
            else                                       -> null
        }
    }
```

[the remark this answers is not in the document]
matkoniecz: It should stay the same but I will test it. It seems much better to me.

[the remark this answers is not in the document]
smichel17: I am personally not a fan of aligning the `->` arrows. It sometimes is more readable, but there are other times when it is less readable, and I think a consistent style is more important. And it requires adjusting multiple lines when one line changes (which hurts the `git blame`).

<details><summary>Example of when aligned is worse</summary>

```kt
when {
    shortCondition                       -> veryExtremelyLongResultThatTakesLotsOfSpace
    suchALongConditionTakesUpLotsOfSpace -> shortResult
}
```
The problem is that now your long result exceeds the line length limit. So your choices become either:
- Keep a really long line there (hard to read)
- Revert to no alignment in just this location (inconsistent)
- Wrap every line:
```kt
when {
    shortCondition ->
        veryExtremelyLongResultThatTakesLotsOfSpace
    suchALongConditionTakesUpLotsOfSpace ->
        shortResult
}
```
Wrapping is *okay*, but I'm always vaguely uncomfortable without braces, and braces add a *lot* of useless vertical lines.
</details>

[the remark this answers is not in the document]
FloEdelmann: Totally agree @smichel17! Also, I can never remember whether to align the arrows or the statements after the arrows. But I wanted to conform to the currently used style when suggesting this refactoring.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -25,6 +25,7 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
 
     override val otherAnswers = listOf(
         AnswerItem(R.string.quest_address_answer_no_housenumber) { onNoHouseNumber() },
+        AnswerItem(R.string.quest_address_answer_house_name_and_housenumer) { switchToHouseNameAndHouseNumber() },
         AnswerItem(R.string.quest_address_answer_house_name) { switchToHouseName() },
         AnswerItem(R.string.quest_housenumber_multiple_numbers) { showMultipleNumbersHint() }
```
smichel17, line 30: Not directly related to this PR, but we should probably standardize on whether to use lambdas or function references
```kt
AnswerItem(R.string.quest_address_answer_house_name_and_housenumer, ::switchToHouseNameAndHouseNumber)
```
Currently they are both used. I see four reasonable options:

- Prefer function references whenever possible
- Prefer lambdas for the last argument (outside the parentheses) and function references otherwise
- Prefer function references when registering callbacks and lambdas otherwise— _less sure about this one_
- Always prefer lambdas



@westnordost Any opinion?

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -25,6 +25,7 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
 
     override val otherAnswers = listOf(
         AnswerItem(R.string.quest_address_answer_no_housenumber) { onNoHouseNumber() },
+        AnswerItem(R.string.quest_address_answer_house_name_and_housenumer) { switchToHouseNameAndHouseNumber() },
```
smichel17, line 28:
```suggestion
        AnswerItem(R.string.quest_address_answer_house_name_and_housenumber) { switchToHouseNameAndHouseNumber() },
```

app/src/main/res/values/strings.xml
```diff
@@ -451,7 +451,9 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_powerPolesMaterial_concrete">Concrete</string>
     <string name="quest_maxspeed_answer_advisory_speed_limit">Just an advisory speed limit…</string>
     <string name="quest_address_answer_house_name">No house number but a name…</string>
+    <string name="quest_address_answer_house_name_and_housenumer">Both house number and a name…</string>
```
smichel17, line 454:
```suggestion
    <string name="quest_address_answer_house_name_and_housenumber">Both house number and a name…</string>
```

smichel17 (changes requested):
Typo

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: Good idea, though `savedInstanceState` will be trickier (I need to find out how to save that - probably as a string?)

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: numer VS number

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
smichel17: String or Int. Both have:
```kt
enum class HouseAddressType { NUMBER, NAME, NAME_AND_NUMBER }
private var addressType: HouseAddressType = HouseAddressType.NUMBER
```
Then String looks something like:
```kt
// serialize
outState.putString(ADDRESS_TYPE, addressType.name)
// deserialize
addressType = try {
    HouseAddressType.valueOf(savedInstanceState?.getString(ADDRESS_TYPE))
} catch (e: IllegalArgumentException) {
    HouseAddressType.NUMBER
}
```
And Int looks something like:
```kt
// serialize
outState.putString(ADDRESS_TYPE, addressType.name)
// deserialize
val ordinal = savedInstanceState?.getInt(ADDRESS_TYPE) ?: HouseAddressType.NUMBER.ordinal
addressType = HouseAddressType.values().getOrElse(ordinal, HouseAddressType.NUMBER)
```

ref: https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants

peternewman:
> Currently: only solution is to wait for resurvey.

I don't believe there is a resurvey and don't see any code for one either. But I can think of very few cases when a house number or name would change either.

> I can't think of anything which could be done for this _specifically_, but in general, either the layers feature or my proposed alternative, "bulk edit mode" ([#2461 (comment)](https://github.com/streetcomplete/StreetComplete/issues/2461#issuecomment-755638075)) would allow editing at times when a quest would not normally be shown.

There is some potentially valid tagging for it:
https://wiki.openstreetmap.org/wiki/Tag:noname%3Dyes

But aside from asking for every house as others have said I can't think of a sensible way to filter it.

app/src/main/res/values/strings.xml
peternewman, line 454:
```suggestion
    <string name="quest_address_answer_house_name_and_housenumber">Both a house number and a name…</string>
```

Or slightly more succinct and I think still good English (@smichel17 )
```suggestion
    <string name="quest_address_answer_house_name_and_housenumber">Both a house number and name…</string>
```

peternewman (commented):
Just an observation, this quest includes:
Housenumber
HouseNumber
HOUSENUMBER
House number

And similar for house name.

I appreciate the OSM key is addr:housenumber and the English is "House Number", but would it be worth being consistent and either using housenumber/HOUSENUMBER/Housenumber throughout, or treating it as two words everywhere apart from when it's the actual OSM key (i.e. houseNumber/HOUSE_NUMBER/HouseNumber).

Likewise, but not as bad, for house name.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: Not entirely happy about savedInstanceState - maybe can be improved further - but done.

matkoniecz:
@smichel17 @FloEdelmann @peternewman  Thanks for reviews here and elsewhere!

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
westnordost: 👍

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
westnordost: I slightly prefer lambdas, but I don't see the necessity that it is standardized in the codebase

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -40,22 +41,45 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
     private var addButton: View? = null
     private var subtractButton: View? = null
 
-    private var isHousename: Boolean = false
+    enum class IntefaceMode {
```
westnordost, line 44:
```suggestion
    enum class InterfaceMode {
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -40,22 +41,45 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
...
         super.onSaveInstanceState(outState)
-        outState.putBoolean(IS_HOUSENAME, isHousename)
+        when(interfaceMode) {
+            IntefaceMode.HOUSENUMBER -> outState.putBoolean(IS_HOUSENUMBER, true)
+            IntefaceMode.HOUSENAME -> outState.putBoolean(IS_HOUSENAME, true)
+            IntefaceMode.HOUSENUMBER_AND_HOUSENAME -> {
+                outState.putBoolean(IS_HOUSENUMBER, true)
+                outState.putBoolean(IS_HOUSENAME, true)
+            }
+        }
```
westnordost, line 82:
```suggestion
        outState.putString(INTERFACE_MODE, interfaceMode.name)
```

and in `onCreateView`
```kotlin
interfaceMode = savedInstanceState?.getString(INTERFACE_MODE, null)?.let { InterfaceMode.valueOf(it) }
```
(or)
```kotlin
interfaceMode = if (savedInstanceState != null) {
  InterfaceMode.valueOf(savedInstanceState.getString(INTERFACE_MODE, null)!!)
} else InterfaceMode.HOUSENUMBER
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -240,27 +270,38 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
...
-            }
+    override fun isRejectingClose(): Boolean {
+        val houseName = houseNameInput?.nonEmptyInput
+        val houseNumber = houseNumberInput?.nonEmptyInput
+        val conscriptionNumber = conscriptionNumberInput?.nonEmptyInput
+        val streetNumber = streetNumberInput?.nonEmptyInput
+        val blockNumber = blockNumberInput?.nonEmptyInput
+
+        if(houseName != null || houseNumber != null) {
+            return true
```
westnordost, line 281: So, user can answer "it has both number and name" but then don't fill out both the fields and still OK it?

I think this is confusing.

That means that the "OK" button will appear as soon as he inputs the first letter for the housename.

The "OK" button should only appear after there is one character at least in **both** input fields, since this is what the user stated that he wanted to input.

westnordost (changes requested):
Alright, also review by me. A few things could be simplified. There is one UI issue I see (IMO).

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: That is in `isRejectingClose` - so it is checking whether user should confirm discarding partially filled info.

`isFormComplete` is higher.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
westnordost: serializing enums according to their ordinal is general bad practice. In this case, it doesn't matter, but one should not get into the habit of doing it. 

I added a new comment regarding this, resolving this one.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
westnordost: I had to put the comment somewhere. But If I misunderstood the behavior from reading the code and it is already the way I described, then you can simply reoslve this conversation.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
smichel17: I think the first option is missing a `?: InterfaceMode.HOUSENUMBER`. I was playing around with it and think I like
```kt
val prevMode = savedInstanceState?.getInt(INTERFACE_MODE)?.let { InterfaceMode.valueOf(it) }
interfaceMode = prevMode ?: InterfaceMode.HOUSENUMBER
```

westnordost: You are right

smichel17: You can stay under the 100 char limit using a function reference :laughing: 
```kt
        val prevMode = savedInstanceState?.getString(INTERFACE_MODE)?.let(InterfaceMode::valueOf)
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: oh right, it was actually broken

smichel17 (approved):
Looks good to me.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -67,16 +78,22 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
         }
     }
 
-    override fun isFormComplete() = (!isShowingHouseNumberHint || isHousename) && createAnswer() != null
+    override fun isFormComplete() = (!isShowingHouseNumberHint || interfaceMode == InterfaceMode.HOUSENAME ) && createAnswer() != null
```
FloEdelmann, line 81:
```suggestion
    override fun isFormComplete() = (!isShowingHouseNumberHint || interfaceMode == InterfaceMode.HOUSENAME) && createAnswer() != null
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -40,22 +41,32 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
...
-        setLayout(if(isHousename) R.layout.quest_housename else R.layout.quest_housenumber)
+        val prevMode = savedInstanceState?.getString(INTERFACE_MODE)?.let { InterfaceMode.valueOf(it) }
+        interfaceMode = prevMode ?: InterfaceMode.HOUSENUMBER
+        setLayout(
+            when(interfaceMode) {
+                InterfaceMode.HOUSENUMBER -> R.layout.quest_housenumber
+                InterfaceMode.HOUSENAME -> R.layout.quest_housename
+                InterfaceMode.HOUSENUMBER_AND_HOUSENAME -> R.layout.quest_housename_and_housenumber
+            }
+        )
```
FloEdelmann, line 63: That could be separated into a function that is also called from `switchToHouseName` and `switchToHouseNameAndHouseNumber`, so the layouts don't have to be repeated there. But I guess it's fine either way.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
```diff
@@ -240,27 +257,38 @@ class AddHousenumberForm : AbstractQuestFormAnswerFragment<HousenumberAnswer>()
...
-                houseNumberInput?.nonEmptyInput?.let { houseNumber ->
-                    HouseAndBlockNumber(houseNumber, blockNumber)
-                }
-            }
+        if(conscriptionNumber != null || streetNumber != null || blockNumber != null) {
+            return true
         }
-        else if (houseNumberInput != null) {
-            houseNumberInput?.nonEmptyInput?.let { HouseNumber(it) }
+        return false
```
FloEdelmann, line 273: You could replace this with `return houseName != null || houseNumber != null || conscriptionNumber != null || streetNumber != null || blockNumber != null`. No need for separate `if` statements that return a boolean.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
FloEdelmann: Also, this will still show the OK button in the `HOUSENUMBER_AND_HOUSENAME` when only one input field contains text, right?

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: I forgot to push :(

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: I know but it seemed better than super long line

FloEdelmann: Conceptually, I find it harder to understand though. It seems that there are three different conditions, when really there is just one.

> super long line

Why not like this then?

```kt
return houseName != null || houseNumber != null || conscriptionNumber != null
    || streetNumber != null || blockNumber != null
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
FloEdelmann: Or this could be done in `setLayout`, which then not accepts a layout ID as parameter, but an `InterfaceMode`.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: Somehow I prefer multiple ifs over splitting line this way, but maybe it is just me.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
smichel17: <details><summary>I had the idea to suggest something similar. I decided not to because I don't think it's really necessary from a cost:benefit perspective.</summary>

- This is isolated code, so a little messiness is not going to spread and become a worse problem later; there's no harm in cleaning up this piece later, before making another change.

- This is user interface code, and I've learned to accept that it's OK for UI code to be a little messy. Same reason I've embraced the ideas behind tailwind css for websites. A lot of what you write really isn't reusable, and half the time it's going to be thrown away and rewritten later, anyway.
    - Admittedly, this is more "back of the front end" (managing state and data), where it's more important to write clean code.
</details>

To be clear, I'm not opposed to it. I just don't consider it a blocker for merging.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
smichel17: How about
```kt
return listOf(houseName, houseNumber, conscriptionNumber, streetNumber, blockNumber).any { it != null }
```

smichel17: I agree that @FloEdelmann's version has better semantics.

I suspect that @matkoniecz likes the current version because of the grouping of the values.

You can get both using intermediary values
```kt
val hasNameOrNumber = houseName != null || houseNumber != null
val hasWhateverTheseAreCalled = conscriptionNumber != null || streetNumber != null || blockNumber != null
return hasNameOrNumber || hasWhateverTheseAreCalled
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumberForm.kt
matkoniecz: I created new `setLayoutBasedOnInterfaceMode` function as `setMode` appears in other quests and it should be consistent

FloEdelmann (approved):
Looks good to me now. Thanks @matkoniecz for implementing the feedback!

westnordost (approved):

PR #3589, opened 2021-12-13
stop encouraging to edit a dead repository

matkoniecz:
https://github.com/goldfndr/StreetCompleteJSON appears to be dead - last commit in 2019, no response even to archive request at https://github.com/goldfndr/StreetCompleteJSON/issues/20

I think that it would be better to remove it, and definitely remove it from listing of improvable things


smichel17:
Is https://github.com/streetcomplete/StreetComplete/pull/2754 a replacement for it?

riQQ:
How about copying the existing one to the StreetComplete organization and updating the the links at taginfo.openstreetmap.org? This can already be done without the #2754 being finished.

matkoniecz:
> How about copying the existing one to the StreetComplete organization

The problem is that it is extremely outdated and misleading at this point.

riQQ:
Sure, but it doesn't get worse because it is the same data. And everything is ready when the updated taginfo json eventually becomes available.

matkoniecz:
The difference is that right now it is unofficial and not supoorted.

If merged into this repository it would become officially supported and every mistake and outdated info would become a SC bug.

PR #3601, opened 2021-12-18
United Kingdom: automatically infer hydrant diameter unit

westnordost:
ping @timothywashere @thefeiter


timothywashere:
The values you've used look good to me.

Issue #3593, opened 2021-12-17
Building levels list removes new entries too quickly


Helium314:
The new buildings level list with remembering answer count is useful if you have a lot of houses with similar levels. But it is quite annoying if you move from an area with 1-2 level houses to an area with higher houses, e.g. 4-6 levels.
If you do this, you will see your last answer, and a bunch of 1-2 levels answers. This gets better only after solving many building level quests, and I foud it rather irritating.

I don't know how the selection of answers works exactly, but some ideas for improvement:
* show the last 2 or 3 answers instead of only the last one
* remember fewer previous answers
* forget (some) previous answers after a while, or upon restart

Note that the ideas are for this specific use case, so they may make things worse in different cases. Not sure what should be considered here.

westnordost:
Currently the last answers are persisted (forever). I am also thinking about whether it wouldn't make sense to completely drop the functionality that it is persisted but instead only memorize it for this session (i.e.  until the app is ended)

matkoniecz:
For me it is very useful as in my area buildings are almost always 1, 2, 2+1, 1+1, 4, 5, 10, 12 levels.

So losing history on closing app would be frustrating.

smichel17:
> I don't know how the selection of answers works exactly

- Check the last 30\* answers in the history and get the 6 most common ones.
   \*Sometimes more, but not for the building levels quest.
- If the most recent answer is not included, then add it.
   - If there were fewer than 6, it's just added. If there were 6, it replaces the least-common answer.

As I said back when I [added the feature](https://github.com/streetcomplete/StreetComplete/pull/3373#discussion_r730184583), I'm pretty confident that the algorithm is right, but the numbers I chose were totally off-the-cuff estimates.

> Currently the last answers are persisted (forever).

Technically they are never cleared, but only the last 100 entries are persisted.

> some ideas for improvement

Those are pretty much the solutions I came up with in https://github.com/streetcomplete/StreetComplete/discussions/3493#discussioncomment-1602090. That discussion was about building **type**, so my [conclusion](https://github.com/streetcomplete/StreetComplete/discussions/3493#discussioncomment-1745118) was, "I suspect it's a good idea to show more recent entries, but I want to change only one thing at a time, so I'll change just the sort order first." That's in https://github.com/streetcomplete/StreetComplete/pull/3564.

[...]

PR #3604, opened 2021-12-23
Make number of recent values a parameter

smichel17:
Allows easier experimentation for #3593. Also, I suspect that the right values are different between the building type and levels quest, so we'll want to do this later, anyway.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
@@ -34,13 +34,13 @@ private const val MAX_ENTRIES = 100
  *  items of the sequence, in their original order.
  * If there are fewer than `target` unique items, continues counting items
  *  until that many are found, or the end of the sequence is reached.
- * If the first item (i.e. the most recent one) is not null, it is always
- *  included, displacing the least-common of the other items if necessary.
+ * If the `first` most recent items are not null, they are always included,
+ *  displacing the least-common of the other items if necessary.
  */
-fun <T : Any> Sequence<T?>.mostCommonWithin(target: Int, historyCount: Int): Sequence<T> {
+fun <T : Any> Sequence<T?>.mostCommonWithin(target: Int, historyCount: Int, first: Int): Sequence<T> {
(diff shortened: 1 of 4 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt
```diff
@@ -34,13 +34,13 @@ private const val MAX_ENTRIES = 100
  *  items of the sequence, in their original order.
  * If there are fewer than `target` unique items, continues counting items
  *  until that many are found, or the end of the sequence is reached.
- * If the first item (i.e. the most recent one) is not null, it is always
- *  included, displacing the least-common of the other items if necessary.
+ * If the `first` most recent items are not null, they are always included,
+ *  displacing the least-common of the other items if necessary.
  */
-fun <T : Any> Sequence<T?>.mostCommonWithin(target: Int, historyCount: Int): Sequence<T> {
+fun <T : Any> Sequence<T?>.mostCommonWithin(target: Int, historyCount: Int, first: Int): Sequence<T> {
```
smichel17, line 40: I wonder if this signature would be more understandable as
```kt
fun <T : Any> Sequence<T?>.mostCommonWithin(first: Int, top: Int, within: Int): Sequence<T> {
```
So instead of `mostCommonWithin(6, 30, 1)` you'd do `mostCommonWithin(1, 5, 30)`.
I guess it's unintuitive in the edge case of "what if the first entry is null and we return 6 top values instead".

smichel17:
This only needs a cursory review, I considered just merging it immediately, but I had the idea I commented above, and @FloEdelmann you seem to like bikeshedding with me about code style :). If you don't have an opinion or don't care, feel free to just merge.

matkoniecz:
I am not convinced that it should be configurable in a normal release.

smichel17:
Maybe I used the wrong word here. It's not end-user configurable, I just added a parameter to the `mostCommonWithin` function instead of hard coding it to the 1 most recent value. Still requires a code change, but now it's just one line of code and we can show different numbers of most-recent values for the building type quest vs building levels.

smichel17:
On second thought, while I remain not-totally-happy with the interface here, I also don't want to spend any more time on this. It's definitely "good enough".

Issue #3612, opened 2021-12-30
Railway electrification system


dlichtistw:
### General

Affected tag(s) to be modified/added: [electrified=*](https://wiki.openstreetmap.org/wiki/Key:electrified)
Question asked: **Is this railway section electrified?**

### Checklist
Checklist for quest suggestions (see [guidelines](https://github.com/streetcomplete/StreetComplete/wiki/Adding-new-Quests-to-StreetComplete)):
- [x] 🚧 To be added tag is established and has a useful purpose
Usefulness depends on the use case, but it is definitely established.
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
If there is some kind of electrification, it can be tagged. If it is not electrified, then `electrified=no` applies.
- [ ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary
Overhead contact lines (`electrified=contact_line`) should be very easy to spot for everyone, 3rd rail (`electrified=rail`), 4th rail systems (`electrified=4th_rail`) and APS systems (`electrified=ground-level_power_supply`) may need a closer look, but should be distinguishible with the help of images.
- [x] 💤 Not an overwhelming percentage of quests have the same answer (No spam)
For my country, the ratio of electrified track kilometers is about 0.5. For other countries, this number may be higher or lower.
- [ ] 🕓 Applies to a reasonable number of map data (Worth the effort)
The number of railway elements is not as large as for highways, but (at least for most of Europe) still considerable. Based on a short overpass query (see below), there are probably in the order of 10<sup>5</sup> fixable railway elements in Germany.

### Ideas for implementation

**Element selection:**
Pretty much every railway track in active use should have an `electrified=*` tag.
Funiculars should be excluded, maybe monorails too.

Something along the line of the following overpass query:
```
(
  way["railway"="rail"]({{bbox}});
  way["railway"="light_rail"]({{bbox}});
  way["railway"="tram"]({{bbox}});
  way["railway"="subway"]({{bbox}});
  way["railway"="narrow_gauge"]({{bbox}});
  way["railway"="preserved"]({{bbox}});
)->.tracks;
way.tracks[!"electrified"];
```

It does also apply to a small portion of `highway=*` elements. But it may be hard to find a good subset, where this quest would be reasonable.

**Metadata needed:**
No metadata necessary.

**Proposed UI:**
Choice illustrated by images:
 * not electrified (`electrified=no`)
[...]

PR #3614, opened 2021-12-31
Add electrification system quest.

dlichtistw:
As beeing discussed in #3612 .

Working draft containing:
- [x] OSM element filter
- [x] OSM update commands
- [x] quest form
- [ ] quest icon
- [ ] answer illustrations

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/railway_electrification/AddRailwayElectrificationSystem.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/railway_electrification/AddRailwayElectrificationSystem.kt
@@ -1,0 +1,47 @@
+package de.westnordost.streetcomplete.quests.railway_electrification
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
+import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
+import de.westnordost.streetcomplete.quests.AbstractQuestAnswerFragment
+
+class AddRailwayElectrificationSystem : OsmElementQuestType<RailwayElectrificationSystem> {
+
+    private val tracksFilter by lazy { """
+        ways with
+            railway ~ light_rail|narrow_gauge|preserved|rail|subway|tram
+            and !electrified
+    """.toElementFilterExpression() }
+
+    override val commitMessage = "Add electrification system for railway track."
+    override val wikiLink = "Key:electrified"
+    //TODO: Create and use proper quest icon.
+    override val icon = R.drawable.ic_quest_railway
+
+    override val isSplitWayEnabled: Boolean = true
+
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/railway_electrification/AddRailwayElectrificationSystem.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    override val questTypeAchievements: List<QuestTypeAchievement> = listOf()
+
+    override fun getTitle(tags: Map<String, String>): Int =
+        R.string.quest_railway_electrification_system_title
+
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> =
+        mapData.ways.filter{ isApplicableTo(it) }
+
+    override fun isApplicableTo(element: Element): Boolean = tracksFilter.matches(element)
```
FloEdelmann, line 35: To simplify the code: You can extend from `ElementFilterQuest` (or similar), remove these functions and override `elementFilter` instead.

app/src/main/java/de/westnordost/streetcomplete/quests/railway_electrification/RailwayElectrificationSystem.kt
```diff
@@ -0,0 +1,9 @@
+package de.westnordost.streetcomplete.quests.railway_electrification
+
+enum class RailwayElectrificationSystem( val osmValue: String ) {
```
FloEdelmann, line 3:
```suggestion
enum class RailwayElectrificationSystem(val osmValue: String) {
```

app/src/main/res/values/strings.xml
```diff
@@ -1138,4 +1138,5 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
 Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_shoulder_title">"Does this road have a shoulder?"</string>
     <string name="quest_fireHydrant_diameter_ref_title">What diameter is specified on the sign for this fire hydrant? (reference number: %s)</string>
+    <string name="quest_railway_electrification_system_title">Is this track section electrified?</string>
```
FloEdelmann, line 1141:
```suggestion
    <string name="quest_railway_electrification_system_title">Is this railway track section electrified?</string>
```

Should we explicitly mention "railway" to avoid confusion with `highway=track`?

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/railway_electrification/AddRailwayElectrificationSystem.kt
dlichtistw: Much simpler, indeed. Thanks.

app/src/main/res/values/strings.xml
dlichtistw: Would probably be a good idea.

matkoniecz:
I would exclude trams (as spammy) and tunnels (as impossible or dangerous to survey)

matkoniecz:
And thanks for your pull requests!

I wonder how it works in practice, I worry that it may be very repetitive and spammy but I will check how it actually feels in practice

dlichtistw:
> I would exclude trams (as spammy)

Yes, this would also avoid the `electrified=ground-level_power_supply` tag, which would probably be difficult.

> and tunnels (as impossible or dangerous to survey)

Not sure about that. Surveying a road tunnel by car is probably more dangerous than surveying a rail tunnel by train.

westnordost:
So, as written in #3612, I'll close this PR

PR #3616, opened 2022-01-02
Add "Lifesaver" achievement

FloEdelmann:
### Quests

All quests without an achievement were suitable for this achievement: `AddIsDefibrillatorIndoor`, `AddFireHydrantPosition`, `AddFireHydrantType` and `AddFireHydrantDiameter`. Additionally, `AddBollardType` should count towards this achievement, because of #2128:

> This information is quite useful for routing first responders vehicles, like police, ambulances and fire brigades to a destination.

### Links

There were no existing links that were suitable for the achievement.

I added [OsmHydrant](https://www.osmhydrant.org), a multilingual viewer of fire hydrants which highlights their coverage. It also allows adding/editing hydrants.

Other links I considered:

* http://www.openfiremap.de/ – Germany only
* https://defikarte.ch/ – Switzerland only
* https://aed.openstreetmap.org.pl/ – Poland only
* https://tappenbeck.net/osm/maps/deu/index.php?id=1029 – Outdated
* https://umap.openstreetmap.de/de/map/evak-karte_11188
* https://umap.openstreetmap.fr/en/map/defi-standorte_72716

### Image

It already had a nice image, so there was nothing to do for me :)

### Title

See also #3009. I just chose the one from [the crowdsourced names](https://docs.google.com/spreadsheets/d/1u5Gtgv8UxsbGDMxM-oCzLfQE7I3cOinEE6OK-mhpydY/edit#gid=0) that I liked the most. But I'm open to other opinions. Should we create a survey – maybe here on GitHub with emoji reactions?


westnordost:
`openinframap` could be unlocked with this achievement too

westnordost:
Could it be that osmhydrant is broken? It looks like this for me, all is red and there are weirdly highlighted red circles:

image

Sooner or later, I think it may make sense to link to thematic layers of mapcontrib (https://www.mapcontrib.xyz/t/s8c2d9-Les_bornes_a_incendie#position/15/44.8356887/-0.5805661) or mapcomplete (https://mapcomplete.osm.be/hailhydrant.html?z=18&lat=53.56075&lon=9.962547&language=en#way/99834104) (once it looks less ugly).

westnordost (approved):

FloEdelmann:
> Could it be that osmhydrant is broken? It looks like this for me, all is red and there are weirdly highlighted red circles

That's good hydrant coverage in that particular area. A circle with 200m radius (configurable with the cog icon in the top right) is drawn around every hydrant. White spots show uncovered area.

Try looking at a more rural area with less mapped hydrants.

matkoniecz:
> mapcomplete (https://mapcomplete.osm.be/hailhydrant.html?z=18&lat=53.56075&lon=9.962547&language=en#way/99834104) (once it looks less ugly).

Is there an issue tracking this or at least one of subproblems? Maybe there are unaware that problem exists.

(personally I have problems with confusing UX rather than ugly looks - and I opened two more issues)

app/src/main/res/values/strings.xml
```diff
@@ -1138,4 +1138,7 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
 Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_shoulder_title">"Does this road have a shoulder?"</string>
     <string name="quest_fireHydrant_diameter_ref_title">What diameter is specified on the sign for this fire hydrant? (reference number: %s)</string>
+    <string name="link_osmhydrant_description">"Lookup, add and edit fire stations and fire hydrants"</string>
+    <string name="achievement_emergency_title">Life saver</string>
```
peternewman, line 1142: I think it shouldn't have a space:
```suggestion
    <string name="achievement_emergency_title">Lifesaver</string>
```

But maybe that's just a British thing; this website seems particularly unsure:
https://life-saver.org.uk/

@smichel17 or other natives?

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -585,16 +594,23 @@ object AchievementsModule {
...
+        ),
+
+        Achievement(
+            QuestTypeAchievement.EMERGENCY.id,
+            R.drawable.ic_achievement_emergency,
+            R.string.achievement_emergency_title,
+            R.string.achievement_emergency_solved_X,
+            SolvedQuestsOfTypes,
+            // levels: 10, 30, 60, 100, 150, 210, 280, 360, 450, 550, 660, 780, 910, 1050, ...
+            { lvl -> (lvl + 1)*10 },
```
peternewman, line 606: Should it have an easier level ranking as they are perhaps rarer/more work/more valuable? e.g.
```suggestion
            // levels: 5, 10, 15, ...
            { lvl -> (lvl + 1)*5 },
```

app/src/main/res/values/strings.xml
matkoniecz: https://en.wiktionary.org/wiki/lifesaver was also confused ( https://en.wiktionary.org/w/index.php?title=lifesaver&diff=65161919&oldid=64363252 ) - it seems that both versions are OK. Maybe one is archaic/formal?

peternewman (commented):
I assume we can't pull out the emergency related completions of existence too (e.g. life ring), just the whole quest or not?

matkoniecz:
> 

I assume we can't pull out the emergency related completions of existence too (e.g. life ring), just the whole quest or not?


not right now and likely it is not worth implementing

app/src/main/res/values/strings.xml
matkoniecz: One person from OSM Discord recommended lifeguard for human keeping guard and lifebuoy for device.

Another "if you mean the floaty rings and stuff, 'lifesaver' is better, but both are used"

peternewman: > One person from OSM Discord recommended lifeguard for human keeping guard and lifebuoy for device.

I think they're both very specific to water/beach.

> Another "if you mean the floaty rings and stuff, 'lifesaver' is better, but both are used"

Yeah lifesaver/life saver is far more generic, a lot of the examples include paramedics, or just the general phrase "you're a lifesaver" when someone does something useful (e.g. has a spare charging cable).

smichel17: I prefer with the space.

I think @matkoniecz is on the money here— with the space is more formal and especially more serious. I agree with @peternewman that I'd probably omit the space in the phrase, "you're a lifesaver", but I don't think I'd use that phrase if someone literally saved my life. Also, "Lifesavers" reminds me of these:
lifesaver candy roll

I don't like lifeguard or similar because they're too water/beach specific.

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
westnordost: No, let's not start fine-tuning here, that's a can of worms. The only achievement that has a different progression is the "rare" achievement.

westnordost:
So, @FloEdelmann , would you:
- change wording to "Lifesaver" (without space)
- add openinframap as unlockable through this achievement too (and remove that TODO)

FloEdelmann:
Why should the lifesaver achievement unlock the [Open Infrastructure Map](https://openinframap.org/) link? It displays neither fire stations, nor fire hydrants, nor AEDs:

grafik

FloEdelmann:
I renamed the achievement in the code now, too, as `emergency` is not any better than `lifesaver` (both are the same length and don't need a `_` separator because they're only one word). But having different names might lead to confusion that we can avoid right from the start.

westnordost:
Hmm, you are right regarding openinframap

westnordost (approved):

peternewman:
> > mapcomplete (https://mapcomplete.osm.be/hailhydrant.html?z=18&lat=53.56075&lon=9.962547&language=en#way/99834104) (once it looks less ugly).
> 
> Is there an issue tracking this or at least one of subproblems? Maybe there are unaware that problem exists.
> 
> (personally I have problems with confusing UX rather than ugly looks - and I opened two more issues)

What was the reason for not adding MapComplete, it didn't look very ugly to me?

timothywashere:
Does SC prompt for whether Lifesaver-rings still exist or not?

matkoniecz:
https://wiki.openstreetmap.org/wiki/StreetComplete/Quests -> search for "still" or "life ring" there -> yes, it does

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt#L44

PR #3624, opened 2022-01-08
Add colon to string

FloEdelmann:
To match the style of `quest_address_answer_no_housenumber_message1`:

https://github.com/streetcomplete/StreetComplete/blob/0d9ee29360a77a14b78e35d8dbb7d7475e65e67b/app/src/main/res/values/strings.xml#L703


westnordost:
OK, and reexported to poeditor

PR #3630, opened 2022-01-10
Add smoothness emojis as icons

FloEdelmann:
Following https://github.com/streetcomplete/StreetComplete/pull/3257#issuecomment-926012458, https://github.com/streetcomplete/StreetComplete/pull/3617#discussion_r779741946 and most recently https://github.com/streetcomplete/StreetComplete/discussions/3626#discussioncomment-1935103, I decided to try replacing the smoothness emojis with custom icons.

For now, I used the Twemoji emojis as is, but they're easy to tweak now (and they will work and look the same on every phone). They are displayed inline (as [`ImageSpan`](https://developer.android.com/reference/android/text/style/ImageSpan)s) like the emojis were before.

## Open issues

* [x] ~The explanation text has the emojis hardcoded (and localized), and only the string resource ID is defined in the quest form. So that'd need a specialized form that allows setting a custom text instead of the string resource ID, and manually replacing the three emojis with `ImageSpan`s.~ fixed in [`0908472`](https://github.com/streetcomplete/StreetComplete/pull/3630/commits/090847269340e18a28a5f92e7705804b34d4b9d0)
* [x] ~The Twemoji emojis are not well visible in the dark mode.~ fixed in [`dfb49da`](https://github.com/streetcomplete/StreetComplete/pull/3630/commits/dfb49dac1702bc829a4226f6ba96c84f10cfb476)

What do you say @westnordost, is this the right direction?


westnordost:
ok. This allows us to tweak the pickup icon to look more like a Jeep/Landrover Defender with bigger tyres, too.

FloEdelmann:
The emojis in the explanation text are now also replaced with the new icons.

In dark mode, I've made the tyre color a lot lighter to increase the contrast against the background. Maybe it should be increased even more, but we probably want completely different colors / icons then, so I didn't put too much effort into this.

Same goes for the pickup truck icon; I increased the tyre size, but didn't completely change the icon.

michaelblyons:
Good idea on bigger tyres for the pickup truck. After looking at the SVG, I think a good way to supplement the picture would be to add roll bars to the back. Something like this: `<polyline points="22,22 22,13 26,13 31,22" style="fill:none;stroke:#99AAB5;stroke-width:2" />`

michaelblyons (approved):

app/src/main/res/values-night/colors.xml
```diff
@@ -16,6 +16,8 @@
     <color name="title_text_next_to_image">#fff</color>
     <color name="description_text_next_to_image">#fff</color>
 
+    <color name="smoothness_vehicle_tyre">#70808b</color>
```
westnordost, line 19: whoa, ... man, okay...

Why not simply make the tyre color darker than the dark mode background?

westnordost (approved):

westnordost:
Anyway, can be merged as is, thank you!

I used this PR to also make some changes to the emojis, in particular make the bike look more like a city bike because this is what it is supposed to represent. See https://github.com/streetcomplete/StreetComplete/discussions/3626#discussioncomment-1955238

app/src/main/res/values-night/colors.xml
FloEdelmann: I tried that first, but I couldn't find a color that looks nice on both bright and dark backgrounds.

PR #3631, opened 2022-01-11
replace LocationRequestFragment with suspending api

westnordost:
This PR replaces the LocationRequestFragment with a suspending API. I think this is the swifter and more readable solution, even though it is only ~40 lines shorter.

Opinions, @smichel17 @FloEdelmann ?

(Main source file is `LocationRequester.kt`)


app/src/main/java/de/westnordost/streetcomplete/MainActivity.kt
```diff
@@ -64,17 +62,18 @@ class MainActivity : BaseActivity(),
 	@Inject lateinit var userLoginStatusController: UserLoginStatusController
 	@Inject lateinit var userUpdater: UserUpdater
 
+	private val requestLocation = LocationRequester(this, this)
```
FloEdelmann, line 65: Indentation should be 4 spaces here, the same as in the rest of the file.

FloEdelmann (commented):
I don't know much about all the location handling, so I can't comment much on the implementation. After a quick look, it seems more readable though.

I just noticed one place where indentation was tabs instead of spaces. However, that's not new.

smichel17:
I am moving tomorrow, and have an all-day commitment Thursday. So I may not be able to look at this until Friday.

smichel17:
Quick review first impressions: looks good

PR #3633, opened 2022-01-11
Allow converting ways into steps in smoothness quest

FloEdelmann:
See https://github.com/streetcomplete/StreetComplete/issues/3288#issuecomment-919634275 and the following comments and https://github.com/streetcomplete/StreetComplete/pull/3257#issuecomment-939955997.

CC @Helium314

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddRoadSmoothness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddRoadSmoothness.kt
@@ -58,12 +58,7 @@ class AddRoadSmoothness : OsmFilterQuestType<SmoothnessAnswer>() {
                 changes.deleteIfExists("smoothness:date")
                 changes.deleteCheckDatesForKey("smoothness")
             }
-            is IsActuallyStepsAnswer -> {
-                changes.modify("highway", "steps")
-                changes.deleteIfExists("smoothness")
-                changes.deleteIfExists("smoothness:date")
-                changes.deleteCheckDatesForKey("smoothness")
-            }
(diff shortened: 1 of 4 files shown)


matkoniecz:
Not sure about AddRoadSmoothness

It may be worrying too much - but what about road that has sidewalks which are steps?

https://www.google.pl/maps/@50.1281274,19.9042143,3a,75y,248.56h,95.24t/data=!3m6!1e1!3m4!1sky4yhxnYWZHPo1qCnowyuQ!2e0!7i13312!8i6656

screen

I worry that mapper may convert road into steps and significantly break data.

Helium314:
> Not sure about AddRoadSmoothness

It's not available for AddRoadSurface, so I think this should be consistent and thus not available for AddRoadSmoothness

app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddRoadSmoothness.kt
```diff
@@ -58,6 +58,12 @@ class AddRoadSmoothness : OsmFilterQuestType<SmoothnessAnswer>() {
                 changes.deleteIfExists("smoothness:date")
                 changes.deleteCheckDatesForKey("smoothness")
             }
+            is IsActuallyStepsAnswer -> {
```
westnordost, line 61: a road is actually steps? I don't think we want to enable such an answer for `AddRoadSmoothness`

mnalis: I agree, it should have been caught much before one got to smoothness quest, or an OSM note left instead if that is really the case.

PR #3641, opened 2022-01-15
Add air conditioning task

### LINKED ISSUE #3127: [Quest] Ask if a restaurant/business has air conditioning
### General
Question asked: **Does X have air conditioning?**

### Checklist
- [X] 🐿️ Easily answerable by everyone from the outside but a survey is necessary (*by looking at the "box" outside of the building*)

### Ideas for implementation

**Element selection:** 
Overpass query for restaurants, bars, libraries and apartments, where the answer could be more balanced between yes and no, almost every hotel has air conditioning and the same goes for malls, supermarkets etc.:
```
/*
This has been generated by the overpass-turbo wizard.
The original search was:
“(type:way or type:node) and (amenity=restaurant or amenity=bar or amenity=library or tourism=apartment) and air_conditioning!=*”
*/
[out:json][timeout:25];
// gather results
(
  // query part for: “amenity=restaurant and air_conditioning!=*”
  way["amenity"="restaurant"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “amenity=bar and air_conditioning!=*”
  way["amenity"="bar"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “amenity=library and air_conditioning!=*”
  way["amenity"="library"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “tourism=apartment and air_conditioning!=*”
  way["tourism"="apartment"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “amenity=restaurant and air_conditioning!=*”
  node["amenity"="restaurant"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “amenity=bar and air_conditioning!=*”
  node["amenity"="bar"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “amenity=library and air_conditioning!=*”
  node["amenity"="library"]["air_conditioning"!~".*"]({{bbox}});
  // query part for: “tourism=apartment and air_conditioning!=*”
  node["tourism"="apartment"]["air_conditioning"!~".*"]({{bbox}});
);
// print results
out body;
>;
out skel qt;
```

**Metadata needed:** None, but maybe there are some laws for this

**Proposed UI:** Simple Yes/No quest like the lit way one.

Wiki page for the tag: https://wiki.openstreetmap.org/wiki/Key:air_conditioning
Currently used on 14 594 nodes and 7 693 areas (mainly amenities, so I think restaurants and bars),
Maybe it should be disabled by default settings if implemented.

Some other businesses where this could be useful, based on the amount of time spent there by customers and the fact that they are usually local businesses:
- Convenience stores
- Bakeries
- Variety stores
- Charity stores
- Second hand stores
- Hairdressers (even if almost all of them have it, might still be useful in some cases)
- Lottery shops (lottery tickets shops)
- Stationery shops
- Bookstores
- Gift shops

What do you think? Could this be useful in your opinion?

P.S. Sorry for bringing the issue count to 100 :cry:

coolultra1:
Really just the implementation of issue #3127. Still needs an icon and the elementFilter might be adjusted. fixes #3127


matkoniecz:
I think that either `isDeleteElementEnabled` or `isReplaceShopEnabled` is worth enabling. Though I need to test which one would fit better.

matkoniecz:
And congratulations on submitting a new quest!

matkoniecz:
screen01

matkoniecz (approved):

FloEdelmann (approved):
Really cool! (Sorry, pun intended :wink:)

@coolultra1 Just a mini request: Please see https://github.com/streetcomplete/StreetComplete/discussions/2917 and update the PR description accordingly.

coolultra1:
@FloEdelmann Thanks for telling me, I didn't know this GitHub feature.

westnordost:
Nice work! Though please make the quest only available at all in countries that are hot enough that ACs are found. E.g. it doesn't make sense in Sweden. 

I propose to only show it in countries/states that have some red, orange, yellow or green. I.e. Tropical, Subtropical and Dry 
<a href="https://en.wikipedia.org/wiki/Trewartha_climate_classification#/media/File:Trewartha_climate_classification_world_map.png">
</a>

You can use this tool to get all the country / state codes by clicking on them.
Given that temperate/cold climate is somewhat of a rarity on this planet, maybe it is shorter to have a list of countries where it is **not** shown.

westnordost:
This tool: https://ent8r.github.io/blacklistr/

westnordost:
I guess it would be more precise to base this on the average summer temperature, but I couldn't find such a map.

westnordost:
Maybe http://www.mappedplanet.com/tuebersicht_en.php?subthema=Temperature

matkoniecz:
> I propose to only show it in countries/states that have considerate red, orange, yellow or green

Poland has plenty of air conditioned places and is blue on this map.

This is enabled manually, so I think this is sufficient. In fact, in could be nicely used when someone spots air conditioned place quite far toward Noth - they enable quest, tag it, disable.

> average summer temperature

peaks are likely more relevant

matkoniecz:
And search `Finland air conditioning` reveals that also there you can find air conditioned places. I think that it is good without such change.

westnordost:
Okay, but then please use ...

app/src/main/java/de/westnordost/streetcomplete/quests/air_conditioning/AddAirConditioning.kt
```diff
@@ -0,0 +1,48 @@
...
+          amenity ~ restaurant|bar|library
+          or tourism ~ apartment|hotel
+        )
+        and !air_conditioning
+    """
+    override val commitMessage = "Add air conditioning"
+    override val wikiLink = "Key:air_conditioning"
+    override val icon = R.drawable.ic_quest_snow_poi
+    override val isReplaceShopEnabled = true
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
```
westnordost, line 25:
```suggestion
    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside_regional_warning
```

westnordost (changes requested):

coolultra1:
@westnordost Good compromise, got it commited.

westnordost (approved):

smichel17:
What about something more like this (very rough sketch):

image Or more likely in blue: image

Inspired by https://duckduckgo.com/?q=air+conditioning+icon&t=ffab&iar=images&iax=images&ia=images

I'm just thinking, the snowflake works, but it's a pretty generic icon to "burn" on this quest. Are there any snow related quests that we might want to use it for in the future?

westnordost:
If there are any in the future, the icon can be changed then.

coolultra1:
@smichel17 I too think, the snowflake is ok. Your sketch: 1) puts focus to
the waves, which themselves don't really mean anything and 2) looks
complicated, making it hard to parse in an area with very many quest
bubbles.

smichel17 ***@***.***> schrieb am Mo., 17. Jan. 2022, 01:01:

mcliquid:
Can't an air conditioner also be used for heating? At least when I think of hotels or vacation apartments, this is also possible there with a normal air conditioner. This information should also be considered in the quest or not?

matkoniecz:
This quests asks about air conditioning and is not asking about heating.

If asking about heating would make sense, and passes quest guidelines ( https://github.com/streetcomplete/StreetComplete/wiki/Adding-new-Quests-to-StreetComplete ) you can open a new issue for a heating quest.

jens1o:
Another question: Why is this quest not enabled for e.g. libraries and cinemas?

matkoniecz:
It is enabled for libraries: https://github.com/streetcomplete/StreetComplete/pull/3641/files#diff-e087c854c1df8561fd0d667718ab0e8e6873114c507c396c7f5c23c020f92f73R16

For cinemas... No idea, feel free to open a PR adding more items where it makes sense or open a new issue for that.

coolultra1:
@mcliquid In my experience, heating is way more common than AC: Where it gets cold enough (basically everywhere that's not like the sun directly), there is heating. They don't want you to freeze but would rather have you burn, apparently.

PR #3658, opened 2022-01-19
Resurvey unsigned hydrant diameter

arrival-spring:
In the UK at least (and I think also all other countries where the fire hydrant diameter quest is enabled) fire hydrants are meant to have a sign. However some are missing or broken. Therefore it may be that a hydrant is surveyed as having no sign, but that sign is later fixed/replaced.

Happy to change the amount of time before resurvey.

Tested that it compiles and runs, but I'm not sure how I can test that it does what I expect without affecting live data. Does it look ok, or should I write a test file?

(Also happy to close if not wanted)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_diameter/AddFireHydrantDiameter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_diameter/AddFireHydrantDiameter.kt
@@ -2,6 +2,7 @@ package de.westnordost.streetcomplete.quests.fire_hydrant_diameter
 
 
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
 import de.westnordost.streetcomplete.data.osm.mapdata.Element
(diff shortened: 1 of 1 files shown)


matkoniecz:
> I'm not sure how I can test that it does what I expect without affecting live data

Log out while testing or switch off automatic uploads.

Trickier part would be finding any object where it would be asked already.

arrival-spring:
> Trickier part would be finding any object where it would be asked already.

That's what I meant, yes, sorry.

It still works to answer (while logged out), with correct console messages for an element without `fire_hydrant:diameter`.

arrival-spring:
Not sure why I didn't think of it before, I temporarily changed it to `older today -2 weeks` and tested with a suitable node.

Everything worked as expected for both adding a diameter and saying that it wasn't signed.

westnordost:
Hm well, I was under the impression that every hydrant in Germany ought to have a plate with the diameter. But as it turned out, just about half of the hydrants I have surveyed so far actually have signs. Most of these were obviously not new, so I won't expect a sign anytime soon. It looks like plates were only used consistently in cases where the hydrant was not in plain sight on the sidewalk.

FloEdelmann:
For me (also in Germany), it's around 80% of the hydrants I surveyed that have a sign. The hydrants that have no sign are most often the pillar type, so the diameter is written directly on them. Also, I noticed some signs that were up to 20 meters away from the hydrant. And I once saw a sign without a hydrant :sweat_smile: 

I don't know if the signs are actually regularly checked/replaced. But I think it's worth a try; if we all notice in 2 years that no signs were updated, we can remove the option then.

westnordost:
I'd prefer to actually go the other way: Let this quest run for a few years, look at the data situation then.

arrival-spring:
> Hm well, I was under the impression that every hydrant in Germany ought to have a plate with the diameter. But as it turned out, just about half of the hydrants I have surveyed so far actually have signs. Most of those that didn't have a sign were obviously not new, so I won't expect a sign anytime soon. It looks like plates are only used consistently in cases where the hydrant was not in plain sight on the sidewalk.

Whereas for me, this was triggered because the hydrants I saw without signs clearly had a place where the sign should go, but it was missing.

All the ones I've surveyed have been underground, and probably about 90% have signs.

arrival-spring:
Closing this, and I'll see if I remember to bring up the issue in a couple of years' time.

smichel17:
I added the "blocked" label as a reminder.

PR #3660, opened 2022-01-20
Include package name in generated quest list CSV

FloEdelmann:
This lets us compare the source location linked from the wiki with the actual source location.

E.g. this [quest-list.csv](https://github.com/streetcomplete/StreetComplete/files/7907257/quest-list.csv) was generated just before opening this PR. It shows 5 quests where the package name in the wiki is wrong:

image

Note that I [fixed](https://wiki.openstreetmap.org/w/index.php?title=StreetComplete/Quests&diff=2249065&oldid=2249046) the inconsistencies now in the wiki.

### DIFF
--- a/.github/generate-quest-list.main.kts
+++ b/.github/generate-quest-list.main.kts
@@ -95,8 +95,7 @@ class WikiQuest(rowCells: List<String>, rowIndex: Int) {
 
             if (it.startsWith(wikiRowSpan2)) {
                 cellContent = it.substring(wikiRowSpan2.length)
-            }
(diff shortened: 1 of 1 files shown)


westnordost:
Sounds reasonable. 

If you need a review, you'd best request it from @peternewman because other than you he was the only one who has been working on this script

FloEdelmann:
Pinging @rugk, too, since they gave a lot of feedback in #2523.

rugk (commented):
Well… I don't know Kotlin so unfortunately there is nothing language-specific I could comment on, but from what I read/what you say it does, it looks all reasonably good. :upside_down_face:

PR #3665, opened 2022-01-21
Make interface of StringMapChangesBuilder work like a Map<String,String>

westnordost:
Before I try to wrap this up by writing tests etc, I'd like to know if you think this is an improvement at all.

Also, right now StringMapChangesBuilder implements Map<String,String> but not MutableMap<String,String> because this would be more complex to do. But if there are clear benefits, I could do that too.

### What I did

- Replaced `changes.addOrModify(k,v)`, `changes.modify(k,v)` and `changes.add(k,v)` with `tags[k] = v`. I don't think the added strict-ness of requiring the code to be aware whether there is already a value or not was worth it.
- Remove `getPrevious*` methods of StringMapChangesBuilder with that it implements `Map<String, String>`, i.e. normal map access functions can be used. This however **also** means that the behacior changed: `changes.getPrevious` returned always the value of the given key regardless of any changes added so far. `tags[key]` returns now the aggregate of the original tag value and/or any changes made. E.g. 
```
tags["highway"] == "residential" // true
tags["highway"] = "unclassified"
tags["highway"] == "residential" // false
```


app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
```diff
@@ -35,7 +35,7 @@ const val SURVEY_MARK_KEY = "check_date"
 // generated by "make update" from https://github.com/mnalis/StreetComplete-taginfo-categorize/
 val KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED = listOf(
     "shop_?[1-9]?(:.*)?", "craft_?[1-9]?", "amenity_?[1-9]?", "old_amenity", "old_shop",
-    "information", "leisure", "office", "tourism",
+    "information", "leisure", "office", "tourism", "craft", "shop", "amenity",
```
westnordost, line 38: Actually I think this is a bug on current master, no? `amenity=cafe`  should certainly be removed when a café is replaced with a florist or the other way round??

mnalis: @westnordost Aren't those matched by regexps above? ie. `craft_?[1-9]?` should match  `craft` as well as `craft_1` and `craft2` (same for `shop` /  `amenity`). At least if kotlin regexps work like I'm used to in `grep(1)` etc. ?

westnordost: Oh, right. It seems I overlooked the `? `

westnordost, line 38:
```suggestion
    "information", "leisure", "office", "tourism"
```

matkoniecz:
I prefer style of the new code in quests

app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
```diff
@@ -35,7 +35,7 @@ const val SURVEY_MARK_KEY = "check_date"
 // generated by "make update" from https://github.com/mnalis/StreetComplete-taginfo-categorize/
 val KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED = listOf(
     "shop_?[1-9]?(:.*)?", "craft_?[1-9]?", "amenity_?[1-9]?", "old_amenity", "old_shop",
-    "information", "leisure", "office", "tourism",
+    "information", "leisure", "office", "tourism"
```
smichel17, line 38: Lost the trailing comma?

westnordost, line 38:
```suggestion
    "information", "leisure", "office", "tourism",
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/update_tags/StringMapChangesBuilder.kt
```diff
@@ -1,68 +1,81 @@
...
-        val change = StringMapEntryDelete(key, valueBefore)
-        if (changes[key] == change) return
-        checkDuplicate(key)
-        changes[key] = change
-    }
-
-    fun deleteIfExists(key: String) {
-        if (source[key] != null) {
-            delete(key)
+    /** Remove the given key from the map */
```
FloEdelmann, line 6:
```suggestion
    /** Remove the given [key] from the map */
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/update_tags/StringMapChangesBuilder.kt
```diff
@@ -1,68 +1,81 @@
...
+        val valueBefore = source[key]
+        if (valueBefore != null) {
+            addChange(StringMapEntryDelete(key, valueBefore))
         }
     }
 
-    fun deleteIfPreviously(key: String, valueBefore: String) {
-        if (source[key] == valueBefore) {
-            delete(key)
+    /** put the given value for the given key */
```
FloEdelmann, line 14:
```suggestion
    /** Add or modify the given [value] for the given [key] */
```

FloEdelmann (changes requested):
> if you think this is an improvement at all

Yes, I think the syntax is much easier to read and write.

app/src/test/java/de/westnordost/streetcomplete/data/osm/edits/update_tags/StringMapChangesBuilderTest.kt
```diff
@@ -17,12 +17,12 @@ class StringMapChangesBuilderTest {
     @Test(expected = IllegalArgumentException::class)
     fun `delete non-existing fails`() {
         val builder = StringMapChangesBuilder(mapOf("exists" to "like this"))
-        builder.delete("does not exist")
+        builder.remove("does not exist")
     }
```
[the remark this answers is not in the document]
westnordost: I didn't adapt the tests to the new code yet.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/update_tags/StringMapChangesBuilder.kt
westnordost: Is this javadoc syntax? This is not used (yet) in the codebase for documentation comments. But if it works like that, it may be worth adding it everywhere for better code navigation and highlighting. But then it should be a separate PR so that it is done consistently

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/update_tags/StringMapChangesBuilder.kt
westnordost: Put like in a map.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/update_tags/StringMapChangesBuilder.kt
FloEdelmann: No, it's Kotlin's equivalent, KDoc:

> The language used to document Kotlin code (the equivalent of Java's JavaDoc) is called KDoc. In its essence, KDoc combines JavaDoc's syntax for block tags (extended to support Kotlin's specific constructs) and Markdown for inline markup.

The syntax being used here is documented here: https://kotlinlang.org/docs/kotlin-doc.html#inline-markup
The style guide also mentions it: https://kotlinlang.org/docs/coding-conventions.html#documentation-comments

I thought I had seen it in the repo recently, but maybe I'm wrong. Postponing this for another PR is a good option then.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopType.kt
```diff
@@ -48,31 +48,26 @@ class CheckShopType : OsmElementQuestType<ShopTypeAnswer> {
...
-                    changes.deleteIfExists("shop")
-                }
-
-                for ((key, _) in changes.getPreviousEntries()) {
+                for (key in tags.keys) {
                     // also deletes all "disused:" keys
-                    val isOkToRemove =
-                        KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED.any { it.matches(key) }
+                    val isOkToRemove = KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED.any { it.matches(key) }
                     if (isOkToRemove && !answer.tags.containsKey(key)) {
```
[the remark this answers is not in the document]
westnordost: No, this will spawn an IllegalStateException in the StringMapChangesBuilder. Currently, the StringMapChangesBuilder checks that each key is only modified once. E.g.
```
tags["a"] = "b"
tags["a"] = "c"
```
currently throws. Sure, I could remove this check and instead overwrite the change made, but I was thinking that with this check, maybe unintended behavior in the code could be spotted.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -194,54 +193,55 @@ class AddCycleway(private val countryInfos: CountryInfos) : OsmElementQuestType<
...
-                changes.addOrModify(cyclewayKey, "track")
+                tags[cyclewayKey] = "track"
                 if (directionValue != null) {
-                    changes.addOrModify("$cyclewayKey:oneway", directionValue)
+                    tags["$cyclewayKey:oneway"] = directionValue
                 }
-                if (changes.getPreviousValue("$cyclewayKey:segregated") == "no") {
-                    changes.modify("$cyclewayKey:segregated", "yes")
+                if (tags.containsKey("$cyclewayKey:segregated")) {
+                    tags["$cyclewayKey:segregated"] = "yes"
```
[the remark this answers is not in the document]
westnordost: Yes, the tag should be changed to what the user answered, even if the key was previously a trash tag value. Though, I primarily changed this because I merged `modify`, `addOrModify` and `add` into one function (`operator fun set`) and it is OK if the statement in line 215 is executed when cyclewayKey:segregated is already "yes". It then simply does nothing.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_type/AddCrossingType.kt
```diff
@@ -60,16 +60,16 @@ class AddCrossingType : OsmElementQuestType<CrossingType> {
...
 
-    override fun applyAnswerTo(answer: CrossingType, changes: StringMapChangesBuilder) {
-        val previous = changes.getPreviousValue("crossing")
-        if(previous == "island") {
-            changes.modify("crossing", answer.osmValue)
-            changes.addOrModify("crossing:island", "yes")
+    override fun applyAnswerTo(answer: CrossingType, tags: StringMapChangesBuilder) {
+        val crossingValue = tags["crossing"]
+        if(crossingValue == "island") {
+            tags.updateWithCheckDate("crossing", answer.osmValue)
```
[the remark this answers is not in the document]
westnordost: Yes, I find it reads more consistent like this though

app/src/main/java/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterials.kt
```diff
@@ -51,88 +51,72 @@ class AddRecyclingContainerMaterials : OsmElementQuestType<RecyclingContainerMat
...
 
-    private fun applyRecyclingMaterialsAnswer(materials: List<RecyclingMaterial>, changes: StringMapChangesBuilder) {
+    private fun applyRecyclingMaterialsAnswer(materials: List<RecyclingMaterial>, tags: StringMapChangesBuilder) {
+        // first clear recycling:* taggings previously "yes"
+        val previousMaterials = tags.entries.filter { (key, value) ->
+            key.startsWith("recycling:") && value == "yes"
+        }.map { it.key }
+        for (material in previousMaterials) {
+            tags.remove(material)
+        }
```
FloEdelmann, line 69: I really like the simplification in this function!

For this part, this would be a bit shorter and saves two loop iterations:

```suggestion
        for ((key, value) in tags.entries) {
            if (key.startsWith("recycling:") && value == "yes") {
                tags.remove(material)
            }
        }
```

FloEdelmann (approved):
Looks good to me. I have only one (not merge-blocking) suggestion.

westnordost:
Thank you for the review!

PR #3649, opened 2022-01-18
Add `ktlint` Kotlin linter

FloEdelmann:
*Before reading on: Yes, there are an awful lot of changes in this PR. But I think it's worth it.*

### About ktlint

[ktlint](https://github.com/pinterest/ktlint) is an easy-to-use linter for Kotlin. Most of the reported issues can also be fixed automatically. It tries to implement the [official Kotlin code style](https://kotlinlang.org/docs/coding-conventions.html) without having to configure anything.

It is configurable though: Rules we don't like [can be disabled](https://github.com/pinterest/ktlint#custom-ktlint-specific-editorconfig-properties) through `.editorconfig`.

And it is extensible: We could write a [custom ruleset](https://github.com/pinterest/ktlint#creating-a-ruleset) someday that e.g. checks the property order in quest type classes (see #3526 and https://github.com/streetcomplete/StreetComplete/commit/2fcfd101cbe01c84f6e9b7549c0ed887e015c6e2) – with a real AST, not with hacky regexes.

### How to run ktlint

* The ktlint binary can be installed and run manually as a CLI tool. That is how I did it locally.
	* Then just run `ktlint` in the repo root to check all Kotlin files, or `ktlint --format` to fix most reported issues automatically.
	* You can run `ktlint --apply-to-idea-project` to adjust the formatting settings of Android Studio to match what `ktlint` would produce.
	* You can run `ktlint --install-git-pre-commit-hook` or `ktlint --install-git-pre-push-hook` to automatically run `ktlint` before `git commit`ing / `git push`ing, so lint issues don't find their way into the repo at all.
* There are [Maven and Gradle integrations](https://github.com/pinterest/ktlint#integration). I haven't checked those, as I'm not that familiar with those tools.
* On GitHub Actions, we can run ktlint on certain events with the [ktlint-check](https://github.com/musichin/ktlint-check) action. I added it to this PR.
	* Please see [my demo PR](https://github.com/FloEdelmann/StreetComplete/pull/508) for how that looks: The check fails because there are lint issues; a GitHub comment with an explanation is created automatically in this case. See [here](https://github.com/FloEdelmann/StreetComplete/runs/4855642710?check_suite_focus=true) for the workflow output. Go to the [files tab](https://github.com/FloEdelmann/StreetComplete/pull/508/files) to see the lint issues inline as annotations directly in the code changes. Also, there are a few unchanged files with check annotations at the bottom – that shouldn't happen if lint issues are regularly fixed/prevented.
[...]

PR #3674, opened 2022-01-24
Simplify shared complex element filters

FloEdelmann:
Instead of keeping a large chunk of copy-pasted (and then adjusted) code in sync between three quests, create the filtered lists programmatically from one single source of truth.

I noticed this in #3649 because the auto-formatting really messed up the indentation there.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -14,15 +14,14 @@ import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
 import de.westnordost.streetcomplete.ktx.containsAny
 import de.westnordost.streetcomplete.osm.opening_hours.parser.isSupportedOpeningHours
 import de.westnordost.streetcomplete.osm.opening_hours.parser.toOpeningHoursRules
+import de.westnordost.streetcomplete.quests.place_name.PlaceFilterQuestType.OPENING_HOURS_QUEST
+import de.westnordost.streetcomplete.quests.place_name.getPlaceElementFilterString
 import java.util.concurrent.FutureTask
 
 class AddOpeningHours (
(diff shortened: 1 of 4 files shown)


westnordost:
Uhm, this is kind of a dejavú. I thought I remember rejecting a very similar pull request some time ago. Yes, it is shorter (though just 50 lines), but this is where the advantages end:

1. The filter logic is split up in several files. More difficult to navigate.

2. Not the common filter logic is put into that separate file but all of it. So, after putting all the filter logic in one file, you need to separate it again with this unusual construct
 ```
    val common = setOf(AddOpeningHours::class, AddPlaceName::class, AddWheelchairAccessBusiness::class)
    val nameAndOpeningHours = setOf(AddPlaceName::class, AddOpeningHours::class)
    val nameAndWheelchair = setOf(AddPlaceName::class, AddWheelchairAccessBusiness::class)
 ```  
which leads to the filter logic for each individual quest being scattered all over the place. I am surprised you find this more readable.

FloEdelmann:
> Uhm, this is kind of a dejavú. I thought I remember rejecting a very similar pull request some time ago.

Hmm, I didn't know/remember that, otherwise I wouldn't have opened this PR.

> The filter logic is split up in several files.

Yeah, but it kinda is right now, too, because you should check all three files when making a change in any one of them.

---

I'm fine with closing this if it doesn't improve (or actually harms) the readability for others. Do you think there is another way it could be improved?

westnordost:
Probably not, what would be permissable is to outsource filter logic for certain properties of things. E.g. `fun Element.isWalkIn()` or similar for places where you can usually just walk in, the same way the `isAKindOfShop` function exists for things thta usually exist in a Ladenfläche. Though I am really not sure if something like that could be found, because in the end it maybe boils down to `usuallyHasOpeningHours`, `usuallyHasName` etc. which would be 1:1 what the quests filter now.

mnalis:
@FloEdelmann maybe related to this https://github.com/streetcomplete/StreetComplete/discussions/3302 ?

PR #3680, opened 2022-01-27
Fix the naming (should be plural) of deleteIfExistList

peternewman:
Also standardise on remove at the same time to match #3665


FloEdelmann (approved):

westnordost:
I actually removed this function in a branch I will merge in an hour or so

peternewman:
Ah okay, I'll watch this space then...

Hopefully it won't upset too much of the other stuff I've been working on today. :crossed_fingers:

PR #3682, opened 2022-01-27
Resurvey for `opening_hours:signed=no` (fixes #3130)

### LINKED ISSUE #3130: Ask if the opening hours here still not signed
### General
Affected tag(s) to be modified/added: [opening_hours:signed](https://wiki.openstreetmap.org/wiki/Key:opening_hours:signed)
Question asked: **Are the opening hours of Rhababerbarbarabar still not signed?**

### Checklist

I noticed that some shops may have opening hours signed at one time, but later they don't. Even later, they might have opening hours signed again. (Smaller shops, shops where opening hours are hand-written our printed out and taped to the door)

Since the opening hours quest also serves as a quest to check if a shop (etc) is still there, I think it would make sense to also ask recurringly about whether non-signed opening hours are signed now.

It should be a simple yes/no quest. However, if the user answers "yes", the UI to enter/display opening hours should be displayed (depending on if there are already opening hours recorded or not). After setting the opening hours, also `opening_hours:signed=yes` should be tagged.
For the check date, simply `check_date:opening_hours` could be used.

Maybe it could be re-asked less often. On the other hand, answering "no" is really fast.

westnordost:
Ask every year whether places that have been tagged with having no opening hours sign (now) have one. This way, implictly, POIs with no (signed) opening hours are checked every year whether they are still there. Also, maybe they have an opening hours sign now. Fixes #3130.
Direct link to the new quest source code: https://github.com/streetcomplete/StreetComplete/compare/opening-hours-signed-resurvey?expand=1#diff-b05845227ebb46097644bc5eab15e1b36ece352cb19d1bc7489cf0f124950c08

Furthermore, typealiased `StringMapChangesBuilder` to `Tags`, simply because it is shorter.

This PR was mentioned in #3680

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt
@@ -75,12 +75,14 @@ interface OsmElementQuestType<T> : QuestType<T> {
      *  any mix-ups that far apart */
     val highlightedElementsRadius: Double get() = 30.0
 
-    /** applies the data from [answer] to the given element. The element is not directly modified,
-     *  instead, a map of [tags] is built */
-    fun applyAnswerTo(answer: T, tags: StringMapChangesBuilder)
+    /** applies the data from [answer] to the element that has last been edited at [timestamp].
+     * The element is not directly modified, instead, a map of [tags] is built */
+    fun applyAnswerTo(answer: T, tags: Tags, timestampEdited: Long)
 
     @Suppress("UNCHECKED_CAST")
-    fun applyAnswerToUnsafe(answer: Any, tags: StringMapChangesBuilder) {
-        applyAnswerTo(answer as T, tags)
+    fun applyAnswerToUnsafe(answer: Any, tags: Tags, timestampEdited: Long) {
+        applyAnswerTo(answer as T, tags, timestampEdited)
     }
 }
+
+typealias Tags = StringMapChangesBuilder
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt
@@ -75,7 +75,7 @@ interface OsmElementQuestType<T> : QuestType<T> {
      *  any mix-ups that far apart */
     val highlightedElementsRadius: Double get() = 30.0
 
-    /** applies the data from [answer] to the element that has last been edited at [timestamp].
+    /** applies the data from [answer] to the element that has last been edited at [timestampEdited].
      * The element is not directly modified, instead, a map of [tags] is built */
     fun applyAnswerTo(answer: T, tags: Tags, timestampEdited: Long)
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
@@ -4,8 +4,7 @@ import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
 import de.westnordost.streetcomplete.data.meta.hasCheckDate
 import de.westnordost.streetcomplete.data.meta.updateCheckDate
-import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
-import de.westnordost.streetcomplete.data.osm.edits.update_tags.deleteIfExistList
+import de.westnordost.streetcomplete.data.osm.osmquests.Tags
 import de.westnordost.streetcomplete.data.osm.mapdata.Element
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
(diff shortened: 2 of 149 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
```diff
@@ -0,0 +1,115 @@
...
+    private val filter by lazy { """
+        nodes, ways, relations with
+          opening_hours:signed = no
+          and (
+            $hasOldOpeningHoursCheckDateFilter
+            or older today -1 years
+          )
+          and access !~ private|no
+          and (name or brand or noname = yes or name:signed = no or amenity = recycling)
+    """.trimIndent().toElementFilterExpression() }
```
FloEdelmann, line 37: I think that `trimIndent` is unnecessary for element filters, as whitespace is ignored anyway?

```suggestion
    """.toElementFilterExpression() }
```

FloEdelmann (commented):
I've only looked at the code, but that looks good. I have a few suggestions.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
```diff
@@ -62,7 +61,7 @@ class AddStileType : OsmElementQuestType<StileTypeAnswer> {
 
                 if (stileWasRebuilt) {
                     // => properties that refer to the old replaced stile should be removed
-                    tags.deleteIfExistList(STILE_PROPERTIES - "material")
+                    STILE_PROPERTIES.forEach { tags.remove(it) }
                     if(newMaterial != null) {
                         tags["material"] = newMaterial
                     } else {
```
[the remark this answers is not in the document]
westnordost: right, thanks

westnordost:
Thank you for the review!

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileType.kt
```diff
@@ -60,25 +59,17 @@ class AddStileType : OsmElementQuestType<StileTypeAnswer> {
...
 
+                // => properties that refer to the old replaced stile should be removed
                 if (stileWasRebuilt) {
-                    // => properties that refer to the old replaced stile should be removed
-                    tags.deleteIfExistList(STILE_PROPERTIES - "material")
-                    if(newMaterial != null) {
-                        tags["material"] = newMaterial
-                    } else {
-                        tags.remove("material")
-                    }
```
peternewman, line 70: Cool, I was going to say this looked like it could be simplified. :partying_face:

app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileTypeTest.kt
```diff
@@ -31,6 +31,7 @@ class AddStileTypeTest {
             ),
             StileType.SQUEEZER,
             StringMapEntryAdd("check_date", LocalDate.now().toCheckDateString()),
+            StringMapEntryModify("stile", "squeezer", "squeezer"),
```
peternewman, line 34: I meant to flag this with the previous changes, but should this really generate a modify when nothing has actually changed? It should be pretty easy to fix and seems like it would be a good thing to me...

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt
```diff
@@ -75,12 +75,14 @@ interface OsmElementQuestType<T> : QuestType<T> {
      *  any mix-ups that far apart */
     val highlightedElementsRadius: Double get() = 30.0
 
-    /** applies the data from [answer] to the given element. The element is not directly modified,
-     *  instead, a map of [tags] is built */
-    fun applyAnswerTo(answer: T, tags: StringMapChangesBuilder)
+    /** applies the data from [answer] to the element that has last been edited at [timestampEdited].
+     * The element is not directly modified, instead, a map of [tags] is built */
+    fun applyAnswerTo(answer: T, tags: Tags, timestampEdited: Long)
```
peternewman, line 80: Would `lastEditTimestamp` be clearer and only a few characters longer? Or `editedTimestamp` or maybe `editTimestamp` is probably slightly better English (in terms of allowing a native speaker to understand what the variable does).

app/src/test/java/de/westnordost/streetcomplete/quests/barrier_type/AddStileTypeTest.kt
westnordost: Yes, this is deliberate. The user answered in this case that there is **still a squeezer**. Now imagine the element changed since then - someone changed `stile=stepover` or whatever. So the change is checked for conflicts. If this entry wouldn't be in the edit made by the user, StreetComplete would just add the check date and be done with it, while it would be preferrable that the app detected a conflict here.

app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmElementQuestType.kt
westnordost: I think it is fine. It is the same name as the field with the same name in `Element` class

smichel17: <details><summary> In general, I do not think "better English" is a great reason for renaming variables.</summary>

Code is not English. Programming languages may borrow English words (much like English borrows words from many other languages), but they have a different **grammar**, based more in logic and math. I think that grammar is a much better fit for writing unambiguous, maintainable programs than English.
</details>

So, while `lastEditTimestamp` is better English, I don't think it matters.

However, I still prefer `lastEditTimestamp`, because `timestamp` is almost type information here. You could define a `type Timestamp = Long` and then write `lastEdit: Timestamp` and it would be ~equivalent.

PR #3692, opened 2022-01-29
Migrate from Dagger2 to Koin

### LINKED ISSUE #3488: Migrate from Dagger2 to Koin
I used Koin in another small project (Luftlinie) and I like it because it 

- is less boilerplate code
- is easier readable, i.e. less complex. Integrates well with kotlin as a DSL. Easier to set up
- is not a Java dependency, e.g. works with multiplatform
- I read that it is supposed to be faster regarding incremental builds and leads to less generated code. Did not check if this is true though
- IIRC assisted injection is possible. I.e. injection where not all constructor parameters injected but only some. Some classes in this project could profit from this.
- I read that scoped injection (e.g. scoped to an activity or whatever) is not possible or problematic. But we do not use this feature of Dagger2 anyway. We only have one scope, the application.

adrianclay:
Fixes #3488

A few pieces of context, which mean I've likely made a few mistakes along the way.
1. My first contribution on this repo
2. My first bit of mobile development
3. My Kotlin knowledge is very basic

A few things I need to do before becoming confident in the quality of this change.
- [x] Understanding what the `ApplicationComponent.kt` file was doing, and whether I need to replicate that behaviour

[Reading the Dagger docs](https://developer.android.com/training/dependency-injection/dagger-android#injecting-activities) explains this is a standard part of their approach to DI.  Koin does not have a similar concept, so will not replicate.

- [x] Dig deeper into the `Provider<OAuthConsumer>`, I skipped over that for now

I've decided to not dig too deep into this [guided by the advice from](https://github.com/streetcomplete/StreetComplete/pull/3692#issuecomment-1027278840) @westnordost.  I'm going to replace the `Provider<OAuthConsumer>.get()` with a call to a `OAuthConsumer` factory method.

- [x] Understand the purpose of `AbstractQuestAnswerFragment`.`InjectedFields` and make sure that I haven't broken any behaviour.

@smichel17 has provided [a very helpful explanation below](https://github.com/streetcomplete/StreetComplete/pull/3692#issuecomment-1025062676) of the purpose.  Given this understanding and some brief testing, Koin appears to happily inject `AbstractQuestAnswerFragment` without the need for `InjectedFields`.

- [x] Learn how to run the automated tests, and then check nothing is broken
- [ ] Perform some more thorough manual testing of the application
- [x] Do some more research on how Koin should integrate with Workers, there was conflicting advice online.

[Koin WorkManager](https://insert-koin.io/docs/reference/koin-android/workmanager/) appears to be the recommended approach.

- [ ] Learn how to prettify/reformat the code.
- [x] Get a better understanding of `io.insert-koin:koin-android-compat` vs `io.insert-koin:koin-android` and confirm that the `compat` version is necessary.


smichel17:
Welcome, and thank you for the effort! Off hand, I can answer this one at least:

> Understand the purpose of `AbstractQuestAnswerFragment.InjectedFields`

It's a workaround for Dagger not being able to instantiate a generic. However, none of the injected fields are actually generic, so we create an instance of `InjectedFields` and inject it instead, then copy the results out to the object we care about.

westnordost:
Great, thank you for digging into and tackling this! :-D

> Dig deeper into the Provider<OAuthConsumer>, I skipped over that for now

Yeah, I think you can skip over this for now, `Provider<T>` is not used a lot in the codebase. It is just a simple interface 
```java
public interface Provider<T> {
    T get();
}
```

Usually, a class just needs one instance of each of its dependencies. For example, the `MapDataDownloader` needs a `MapDataApi` to communicate with and a `MapDataController` to store the downloaded data in. It just needs one instance of each.
But some classes need instead a supply of dependencies. E.g., dumb example: A `CarAssemblyLine` needs just one `RobotArm` but a `Provider<Motor>` so it can put each one instance of a motor in a car that it assembles.

Why exactly the `OAuthFragment` needs a `Provider<OAuthConsumer>` and not just a `OAuthConsumer` I don't know right now, I'd also need to dig deeper into that.

> Learn how to prettify/reformat the code.

One thing I took away last time I looked at Koin was that Koin supports assisted/partical injection or how you call it. E.g. if you have a class that has 4 constructor parameters of which 3 can be injected and one will be determined dynamically in code, there was something you could use.

westnordost:
Let me know if/when I should look at (parts of) the code. Right now I'd defer it because I don't know if/where something will change.

app/src/main/java/de/westnordost/streetcomplete/util/SoundFx.kt
```diff
@@ -9,13 +9,11 @@ import androidx.annotation.RawRes
...
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withContext
-import javax.inject.Inject
-import javax.inject.Singleton
 import kotlin.coroutines.Continuation
 import kotlin.coroutines.resume
 
 /** Simple wrapper to enable just playing a sound effect from raw resources */
-@Singleton class SoundFx @Inject constructor(private val context: Context) {
+class SoundFx constructor(private val context: Context) {
```
westnordost, line 16: FYI you can just write

```suggestion
class SoundFx(private val context: Context) {
```

app/src/main/java/de/westnordost/streetcomplete/map/MainFragment.kt
```diff
@@ -125,12 +124,12 @@ class MainFragment : Fragment(R.layout.fragment_main),
...
-    @Inject internal lateinit var visibleQuestsSource: VisibleQuestsSource
-    @Inject internal lateinit var mapDataWithEditsSource: MapDataWithEditsSource
-    @Inject internal lateinit var soundFx: SoundFx
-    @Inject internal lateinit var prefs: SharedPreferences
+    internal val questController: QuestController by inject()
+    internal val isSurveyChecker: QuestSourceIsSurveyChecker by inject()
+    internal val visibleQuestsSource: VisibleQuestsSource by inject()
+    internal val mapDataWithEditsSource: MapDataWithEditsSource by inject()
+    internal val soundFx: SoundFx by inject()
+    internal val prefs: SharedPreferences by inject()
```
westnordost, line 132: Is `internal` necessary when using `by inject()`? Couldn't it be `private`?

FloEdelmann:
@adrianclay I've updated this branch and fixed the merge conflicts after #3687 was merged.

adrianclay:
Thank you to all who have reviewed so far, and for @FloEdelmann for fixing up the conflicts.  I believe the PR is ready for review now.

app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.kt
```diff
@@ -21,18 +46,23 @@ import kotlinx.coroutines.CoroutineScope
...
-    @Inject lateinit var resurveyIntervalsUpdater: ResurveyIntervalsUpdater
-    @Inject lateinit var downloadedTilesDao: DownloadedTilesDao
-    @Inject lateinit var prefs: SharedPreferences
-    @Inject lateinit var editHistoryController: EditHistoryController
+    private val preloader: Preloader by inject()
+    private val crashReportExceptionHandler: CrashReportExceptionHandler by inject()
+    private val resurveyIntervalsUpdater: ResurveyIntervalsUpdater by inject()
+    private val downloadedTilesDao: DownloadedTilesDao by inject()
+    private val prefs: SharedPreferences by inject()
+    private val editHistoryController: EditHistoryController by inject()
```
westnordost, line 65: And this works? Because `startKoin` is only called in `onCreate` but these fields are created when the application is created.

app/src/main/AndroidManifest.xml
```diff
@@ -83,6 +83,20 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/file_paths" />
         </provider>
+
+        <!-- Disable the Andriod WorkManager in favour of the org.koin.androidx.workmanager.koin.workManagerFactory -->
+        <!-- As per the Koin Work Manager instructions https://insert-koin.io/docs/reference/koin-android/workmanager -->
```
westnordost, line 88: Hmm, if WorkManager is an issue with Koin and needs an extra dependency to work, maybe `WorkManager` could be replaced with `JobScheduler`. If I remember correctly, `WorkManager` was introduced to have an API for deferred jobs work with Android API levels below 21. For Android API 21+, there is the `JobScheduler` API. Since this app now supports only API 21+, the `WorkManager` could be dropped completely. What do you think?

app/build.gradle.kts
```diff
@@ -123,8 +123,8 @@ dependencies {
     androidTestImplementation("org.assertj:assertj-core:2.8.0")
 
     // dependency injection
-    implementation("com.google.dagger:dagger:$daggerVersion")
-    kapt("com.google.dagger:dagger-compiler:$daggerVersion")
+    implementation("io.insert-koin:koin-android-compat:$koinVersion")
```
westnordost, line 126: `compat` because there are is still Java-code in this repository, right?

Maybe this is a good time to convert the remaining Java source code files to Kotlin then
https://github.com/streetcomplete/StreetComplete/search?l=Java

(No need for it to be in this PR)

westnordost (commented):
Review done by ~70% so far. Looks really good, it's great to see how much more expressive Koin is over Dagger2.

---

In Dagger2, declaring a bean as a singleton (`@Provides @Singleton`) was an extra annotation, the default being that it is not a singleton. (The source of some bugs in the past, by the way)

So it's good that in Koin, the declaration, i.e. decision whether something is a factory or a singleton is explicit. This means though that for all those classes that were not annotated because they were kind of implicitly singletons (e.g. had no state, only used singleton dependencies or were only used by one class and thus were effectively singletons), we need to decide whether to explicitly declare them as `single` or `factory`.
For example:
- A *Dao (data access object) class. It has no internal state and the dependency (the database connection) it uses is a singleton
- The Downloader. Only used by the DownloadService, so effectively a singleton.
- QuestPresetsSoure is an interface which is implemented by QuestPresetsController which is a singleton, thus this is also effectively a singleton

I lean towards being explicit here and also declare them as singletons. Having read deeper into Koin and maybe having a better grasp of best practices there, what's your opinion on this?

app/build.gradle.kts
FloEdelmann: In https://github.com/streetcomplete/StreetComplete/pull/2493#issuecomment-758299236, you said:

> The reason why `CountryInfos.java` and `Abbreviations` has not been converted yet is because I am playing with the thought or at least keep open the option to put each into a small separate library. For this reason, it should remain as .java.

Has your opinion changed here? If so, I'd be happy to submit a PR for that.

westnordost: Yes, my opinion changed. It's really not a lot of code, so outsourcing this to a separately maintained library is almost not worth the effort even if it was used by anyone else. And then, it would be just Java. So it would only be usable by other Java projects (Vespucci, JOSM). It doesn't look like those projects are interested in using that data.

On the other hand, a/b street expressed slight interest in using the country metadata in the future. That would be Rust. So if this data is used by others, I think it is enough to just outsource the metadata into a separate repository, not also the implementation to read that data. As said, it is not really a lot of code anyway.

app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.kt
adrianclay: I was also concerned about this, but looking into the `inject` method, it's [implemented as a lazy delegate](https://github.com/InsertKoinIO/koin/blob/main/android/koin-android/src/main/java/org/koin/android/ext/android/ComponentCallbackExt.kt#L40-L44) for Koins `get`.

I wonder if there's a way to make this less confusing. 🤔

app/src/main/AndroidManifest.xml
adrianclay: I've had to read up about WorkManager a bit more to get a better understanding of what its strengths are.

This [Guide to background work](https://developer.android.com/guide/background/) being the most useful resource I've found so far.

What I understand is that `WorkManager` is a more recent than `JobScheduler`, with there being more guidance/support around using Koin with `WorkManager` than Koin with `JobScheduler`.

adrianclay:
> I lean towards being explicit here and also declare them as singletons. Having read deeper into Koin and maybe having a better grasp of best practices there, what's your opinion on this?

I haven't come across "best practices" around this in Koin, and can't think of any good reasons to default either way.

app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.kt
westnordost: Ah okay. Well, a comment would resolve it, but nevermind. That class is touched not very often and if it works, it works.

app/src/main/AndroidManifest.xml
westnordost: Hmm okay. I may be revisiting this in the future though, WorkManager seems to be a pretty heavy dependency - over 1MB

app/build.gradle.kts
westnordost: Anyway, will resolve this conversation, could be done after this is merged.

westnordost:
Then I think I will use `single` for those things that should not be explicitly be `factory` based on the line of thought that it might be confusing for the reader of the code that if he sees a `factory` somewhere, he assumes there will be many instances of it.

But I can do this after this is merged.

Will continue the review now.

app/src/main/java/de/westnordost/streetcomplete/data/meta/MetadataModule.kt
```diff
@@ -1,23 +1,16 @@
...
-        CountryInfos(assetManager, countryBoundaries)
-
-    @Provides @Singleton fun countryBoundariesFuture(assetManager: AssetManager): FutureTask<CountryBoundaries> =
-        FutureTask { CountryBoundaries.load(assetManager.open("boundaries.ser")) }
-
-    @Provides @Singleton fun featureDictionaryFuture(assetManager: AssetManager): FutureTask<FeatureDictionary> =
-        FutureTask { AndroidFeatureDictionary.create(assetManager, "osmfeatures/default", "osmfeatures/brands") }
+    single { CountryInfos(get(), get<FutureTask<CountryBoundaries>>(named("CountryBoundariesFuture"))) }
+    single(named("CountryBoundariesFuture")) { FutureTask { CountryBoundaries.load(get<AssetManager>().open("boundaries.ser")) } }
+    single(named("FeatureDictionaryFuture")) { FutureTask { AndroidFeatureDictionary.create(get(), "osmfeatures/default", "osmfeatures/brands") } }
```
westnordost, line 15: Why is a name necessary? Can Koin not handle generics? (Would it be different if it was not a `FutureTask` but a `Deferred`?)

app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
```diff
@@ -11,26 +9,25 @@ import de.westnordost.streetcomplete.data.osmnotes.NotesApi
...
-
-    @Provides fun userApi(osm: OsmConnection): UserApi = UserApi(osm)
-
-    @Provides fun notesApi(osm: OsmConnection): NotesApi = NotesApiImpl(osm)
+    worker { CleanerWorker(get(), get(), get()) }
+}
 
-    @Provides fun mapDataApi(osm: OsmConnection): MapDataApi = MapDataApiImpl(osm)
+private const val OSM_API_URL = "https://api.openstreetmap.org/api/0.6/"
+/** Returns an osm connection with the supplied consumer (note the difference to the above function)  */
```
westnordost, line 30:
```suggestion
/** Returns an osm connection with the supplied consumer (i.e. not the one from the OAuthStore) */
```

app/src/main/java/de/westnordost/streetcomplete/user/LoginFragment.kt
```diff
@@ -110,7 +105,7 @@ class LoginFragment :
     suspend fun hasRequiredPermissions(consumer: OAuthConsumer): Boolean {
         return withContext(Dispatchers.IO) {
             try {
-                val permissionsApi = PermissionsApi(OsmApiModule.osmConnection(consumer))
+                val permissionsApi = PermissionsApi(osmConnection(consumer))
```
westnordost, line 108:
```suggestion
                /* we didn't save the new OAuthConsumer yet but we want to make an API call with it
                   to check if the user granted all required permissions, this is why we need to 
                   create a new OsmConnection with the supplied consumer instead of using an 
                   injected one */
                val permissionsApi = PermissionsApi(osmConnection(consumer))
```

Had to look up again why this is done. Might as well add a comment here to explain why.

app/src/main/java/de/westnordost/streetcomplete/data/user/OAuthStore.kt
```diff
@@ -4,17 +4,14 @@ import android.content.SharedPreferences
...
 /** Manages saving and loading OAuthConsumer persistently  */
-class OAuthStore @Inject constructor(
+class OAuthStore(
     private val prefs: SharedPreferences,
-    private val oAuthConsumerProvider: Provider<OAuthConsumer>
 ) {
     var oAuthConsumer: OAuthConsumer?
         get() {
-            val result = oAuthConsumerProvider.get()
+            val result = okHttpOAuthConsumer()
```
westnordost, line 14: Hmm, you are not using Koin at all here to "inject" this. I can't imagine such a case (replacement for `Provider<T>`) hasn't been considered in Koin.

There should be a `factory<OAuthConsumer> { OkHttpOAuthConsumer(CONSUMER_KEY, CONSUMER_SECRET) }` in the `userModule`, right?
Isn't there a public method on the interface of `userModule` that you can call to get an instance of `OAuthConsumer`? Something like `userModule.get<OAuthConsumer>()`?

westnordost (changes requested):
Nice, you even found those places where Dagger2 wasn't working ideally and fixed those with the constructs available in Koin!

One thing you should definitely test before this can be merged is logging out, logging in. Though, see the other comments:

app/src/main/java/de/westnordost/streetcomplete/data/meta/MetadataModule.kt
adrianclay: >Can Koin not handle generics

Correct, Koin saw both as FutureTask.  See also [Dealing with generics](https://insert-koin.io/docs/reference/koin-core/definitions/#dealing-with-generics)

app/src/main/java/de/westnordost/streetcomplete/data/user/OAuthStore.kt
adrianclay: Resolved in [2324a3e](https://github.com/streetcomplete/StreetComplete/pull/3692/commits/2324a3e9bc18e6f6928a9c3e80ae34712c580db5) via the use of [KoinComponent](https://insert-koin.io/docs/reference/koin-core/koin-component/).

adrianclay:
I've manually tested the following scenarios without problem:

```
Given I am not logged in.
When I login
Then I can publish a change to OSM.org

Given I am logged in
When I log out
And I solve a quest
Then I have 1 pending unpublished change
```

app/src/main/java/de/westnordost/streetcomplete/settings/OAuthFragment.kt
```diff
@@ -66,7 +66,7 @@ class OAuthFragment : Fragment(R.layout.fragment_oauth), BackPressedListener, Ha
             authorizeUrl = inState.getString(AUTHORIZE_URL)
             oAuthVerifier = inState.getString(OAUTH_VERIFIER)
         } else {
-            consumer = okHttpOAuthConsumer()
+            consumer = get()
```
westnordost, line 69: I'd prefer it to be explicit/expressive even if it is not necessary

```suggestion
            consumer = get<OAuthConsumer>()
```

westnordost (approved):

westnordost:
Looks good now, then it is ready to merge right?

westnordost:
Untitled

westnordost:
Thank you again, this has been an important step to remove java dependencies from the project - one step closer to multiplatform

FloEdelmann:

PR #3694, opened 2022-01-30
Add missing assertFalse import

adrianclay:
Couldn't get the tests to run without this

Appears to be a mistake from 9c6d3e25216d06a2c5afa71086949e1e195de926

Found while working on #3692


FloEdelmann (approved):
Oh, good catch! Seems like I haven't run the tests after rebasing in that branch.

Thanks for fixing!

Issue #879, opened 2018-02-19
Augmented reality support


rugk:
Yeah, I know, and kinda big and heavy thing and a crazy idea. The explanation is simple though: May we add a virtual reality mode, where the quests are shown on the screen, i.e. not on a map, but on a live photo stream from the camera?

I still see the success of Pokémon Go, and while StreetComplete is, of course, no game, but a "serious app", it may still benefit from such a hyped feature as augmented/virtual reality. I really think that has a future, and it would really contribute to the gamification. I mean, you likely won't use it all the time (especially as it is heavy for your battery), but I have seen places, where many quests are crowded and it makes sense. Also, it may be really more for showing StreetComplete to your friends or so and making a tendentious task more fun. As such, it would of course more be a feature for young people, I'd say if we need a target group here - a cool thing. In contrast to Pokémon (Go), it also serves a good cause, so we can also take this as an advantage. 😉 
Also there is no other OSM app with virtual reality integration yet (AFAIK), although it basically makes sense to combine map data (OSM, or, as here, quests for this data) with the reality.

It would also be great for showing what StreetComplete does/can do (i.e. when you show it to others, there is a wow effect [😮] 😉), as it is basically the realisation of the feature graphic in the Readme. 
And for the ones who care about the safety of their users, it also helps as they are more likely to look on the street 😉).

And yeah, in any case, this is a very low-prio thing and may want for v10.0 or so. But at least it is independent from the map library. 😉 

And BTW, this is not the [only](https://github.com/westnordost/StreetComplete/issues/9#issuecomment-270203516) [similarity](https://github.com/westnordost/StreetComplete/issues/183#issuecomment-299334841) StreetComplete has (or, aims at) with Pokemon Go. 😉

rugk:
Summary of pros and cons:

Pro:
* gamification
* Pokémon Go-like
* hyped feature/future-proof
* a thing to show for friends/inspire users to get to StreetComplete
* target group: young people
* independent of any map (offline support)
* first OSM app to have this
* user-safety, i.e. users do not look down on their phone (IMHO, not a valid argument as one should always pay attention)

Con:
* resource/battery heavy, not usable for long mapping
* maybe not too useful when we have few quests, although it could display the ones far away, as it is just a direction indicator)
* maybe not too useful for using in the long-term, not only because of battery, but just because the map may be more effective and this just a gimmick

westnordost:
[...]

PR #3709, opened 2022-02-01
add ability to measure distances with AR

### LINKED ISSUE #698: Width of a cycleway/cyclelane

westnordost:
Just this so far, this is not used in any quest yet. I think it is maybe a good idea to have this merged first (to master, after v40 is released) so that other people usually working on/using master can test this feature early as I expect the AR experience will be very different depending on the device.

Particularly, I am interested in comparisons how precise is the measurement on your device (compare with measure tape).

Fixes #698.


app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
```diff
@@ -0,0 +1,418 @@
...
+    private fun getCursorNode(): AnchorNode {
+        var node = cursorNode
+        if (node == null)  {
+            node = AnchorNode().apply {
+                renderable = cursorRenderable
+                setParent(arSceneView!!.scene)
+            }
+            cursorNode = node
+        }
+        return node
```
FloEdelmann, line 335: I think it should be equivalent and a bit more readable without the local variable:

```suggestion
        if (cursorNode == null)  {
            cursorNode = AnchorNode().apply {
                renderable = cursorRenderable
                setParent(arSceneView!!.scene)
            }
        }
        return cursorNode
```

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
```diff
@@ -0,0 +1,418 @@
...
+    private fun getLineNode(): Node {
+        var node = lineNode
+        if (node == null)  {
+            node = Node().apply {
+                renderable = lineRenderable
+                setParent(arSceneView!!.scene)
+            }
+            lineNode = node
+        }
+        return node
```
FloEdelmann, line 347: Again

```suggestion
        if (lineNode == null)  {
            lineNode = Node().apply {
                renderable = lineRenderable
                setParent(arSceneView!!.scene)
            }
        }
        return lineNode
```

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
```diff
@@ -0,0 +1,418 @@
...
+            val dlg = AlertDialog.Builder(this)
+                .setTitle(R.string.no_camera_permission_warning_title)
+                .setMessage(R.string.no_camera_permission_warning)
+                .setPositiveButton(android.R.string.ok) { _, _ -> cont.resume(true) }
+                .setNegativeButton(android.R.string.cancel) { _, _ -> cont.resume(false) }
+                .setOnCancelListener { cont.resume(false) }
+                .create()
+            cont.invokeOnCancellation { dlg.cancel() }
+            dlg.show()
+        }
```
FloEdelmann, line 366: Why not more speaking variable names?

```suggestion
        suspendCancellableCoroutine { coroutine ->
            val dialog = AlertDialog.Builder(this)
                .setTitle(R.string.no_camera_permission_warning_title)
                .setMessage(R.string.no_camera_permission_warning)
                .setPositiveButton(android.R.string.ok) { _, _ -> coroutine.resume(true) }
                .setNegativeButton(android.R.string.cancel) { _, _ -> coroutine.resume(false) }
                .setOnCancelListener { coroutine.resume(false) }
                .create()
            coroutine.invokeOnCancellation { dialog.cancel() }
            dialog.show()
        }
```

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
```diff
@@ -0,0 +1,418 @@
...
+                is MeasureDisplayUnitFeetInch -> {
+                    intent.putExtra(PARAM_DISPLAY_UNIT, FOOT_AND_INCH)
+                    intent.putExtra(PARAM_DISPLAY_UNIT_PRECISION, unit.inchStep)
+                }
+                is MeasureDisplayUnitMeter -> {
+                    intent.putExtra(PARAM_DISPLAY_UNIT, METER)
+                    intent.putExtra(PARAM_DISPLAY_UNIT_PRECISION, unit.decimals)
+                }
+                null -> {}
+            }
```
FloEdelmann, line 410: Would it make sense to make the `MeasureDisplayUnit` interface serializable? Then its (de)serialization code would be defined there, where it's more relevant.

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
westnordost: `cursorNode` is a nullable field, so this is not equivalent (function returns non-nullable)

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
westnordost: it's not a coroutine, it's a continuation. `cont` is used as an abbreviation of this throughout much of the codebase, same with `dlg` for dialog, so it is consistent.

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
westnordost: Hmm, good idea

westnordost:
The measure AR view only works if it is in the list of supported devices:
https://developers.google.com/ar/devices#google_play_devices

Additionally, it only works
- on Android 7 and above
- the device supports OpenGL ES 3.1

matkoniecz:
Then sadly I will be unable to test it.

app/build.gradle.kts
```diff
@@ -183,6 +184,10 @@ dependencies {
 
     // opening hours parser
     implementation("ch.poole:OpeningHoursParser:0.26.0")
+
+    // measuring distance with AR
+    implementation("com.google.ar:core:1.29.0")
+    implementation("com.google.ar.sceneform:core:1.17.1")
```
FloEdelmann, line 190: Will StreetComplete still be eligible for F-Droid when this is included? Or will it need another build flavor?

matkoniecz: For some apps F-Droid maintains patches applied before build.

westnordost: I don't know, we will see what happens. Maybe it will just get a flag "contains non-free software" or something(?)

In any case, shouldn't be hard (for them) to remove this functionality with patches.

westnordost:
In the settings screen (only in debug builds), there is a new entry at the very bottom. Other than if it works at all, I am very interested to see how precise it is on different devices.

For a fair test setup and comparison between devices, let's presume: daytime, outside but cloudy/rainy, measurement of sidewalk width (sidewalk pattern will be similar everywhere)

westnordost:
@FloEdelmann Could you set a breakpoint to see what `ArCoreSessionCreator` returns? (line 169 in the `MeasureActivity`). I wonder what error is thrown when the Play Store is not available. 
What you could also try is to sideload the ARCore from somewhere else. I.e. download the ARCore APK from somewhere else than the play store, from some mirror or whatever.

app/build.gradle.kts
smichel17: details><summary>GRR GITHUB</summary>

I had written a whole comment here but not yet posted, then clicked "unresolve conversation", and GitHub decided to delete my comment. But anyway.
</details>

My understanding is that F-Droid does not knowingly distribute non-free apps, period. So, once someone noticed, they would probably remove F-Droid, unless someone cares enough to write a patch.

I care enough to write a patch. Actually, I've previously considered volunteering to be the "SC F-Droid [maintainer](http://kmkeen.com/maintainers-matter/)" if the need should arise. It seems that time may be here. However, I must first raise an objection to adding a dependency on Google Play Services, and especially to including a non-free library in the app. Yes, we're already using Android and thus, trusting Google, but I think proprietary dependencies and shipping proprietary code are both meaningful lines that I'd rather not cross. Of course, I know it isn't my decision to make.

If you decide to continue down this route, I'd rather keep the patches in this repo than in the F-Droid repo. That way, you only need to get the necessary code from one location. I'm not sure if it would be best to maintain a patchset, use a separate `foss`/`fdroid` branch, or create a new build flavor.

app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt
```diff
@@ -0,0 +1,406 @@
...
+        val result = createArCoreSession()
+        if (result is ArCoreSessionCreator.Success) {
+            val session = result.session
+            val config = Config(session)
+            configureSession(config)
+            session.configure(config)
+            addArSceneView(session)
+        } else if (result is ArCoreSessionCreator.Failure) {
+            val reason = result.reason
+            if (reason == ArNotAvailableReason.AR_CORE_SDK_TOO_OLD) {
```
FloEdelmann, line 170: For me, the reason is `DEVICE_COMPATIBILITY_CHECK_FAILURE`.

And in `ArCoreSessionCreator`, `availability` is `UNKNOWN_ERROR`:
https://github.com/streetcomplete/StreetComplete/blob/13502d623f0aaf94a12efb001879a0e913c274a1/app/src/main/java/de/westnordost/streetcomplete/measure/ArCoreSessionCreator.kt#L52

After sideloading ARCore from [APKMirror](https://www.apkmirror.com/apk/google-inc/arcore/arcore-1-29-2132102-release/google-play-services-for-ar-1-29-213210293-2-android-apk-download/), the reason here is instead `DEVICE_NOT_COMPATIBLE`.

peternewman:
I've got a compatible phone, can someone run the debug build or upload an APK or something and I'll see how I get on:
https://github.com/streetcomplete/StreetComplete/actions/workflows/build-debug-apk.yml

FloEdelmann:
Workflow's running: https://github.com/streetcomplete/StreetComplete/actions/runs/1783714234

HolgerJeromin:
Crashed at the first call:

```
Thread: main
App version: 40.0-beta1
Device: motorola  ocean, Android 10
Locale: de_DE
Stack trace:
com.google.ar.core.exceptions.UnsupportedConfigurationException
	at java.lang.reflect.Constructor.newInstance0(Native Method)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
	at com.google.ar.core.Session.throwExceptionFromArStatus(Session.java:16)
	at com.google.ar.core.Session.nativeConfigure(Native Method)
	at com.google.ar.core.Session.configure(Session.java:1)
	at de.westnordost.streetcomplete.measure.MeasureActivity.initializeSession(MeasureActivity.kt:166)
	at de.westnordost.streetcomplete.measure.MeasureActivity.access$initializeSession(MeasureActivity.kt:37)
	at de.westnordost.streetcomplete.measure.MeasureActivity$initializeSession$1.invokeSuspend(Unknown Source:11)
	at kotlinx.coroutines.DispatchedTaskKt.resume(DispatchedTask.kt:256)
	at kotlinx.coroutines.DispatchedTaskKt.resumeUnconfined(DispatchedTask.kt:190)
	at kotlinx.coroutines.DispatchedTaskKt.dispatch(DispatchedTask.kt:161)
	at kotlinx.coroutines.CancellableContinuationImpl.dispatchResume(CancellableContinuationImpl.kt:397)
	at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl(CancellableContinuationImpl.kt:431)
	at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl$default(CancellableContinuationImpl.kt:420)
	at kotlinx.coroutines.CancellableContinuationImpl.resumeWith(CancellableContinuationImpl.kt:328)
	at de.westnordost.streetcomplete.util.ActivityForResultLauncher.launcher$lambda-0(ActivityForResultLauncher.kt:15)
	at de.westnordost.streetcomplete.util.ActivityForResultLauncher.$r8$lambda$22ew8YISY4HdjsRYdAbnAyWz-Wk(Unknown Source:0)
	at de.westnordost.streetcomplete.util.ActivityForResultLauncher$$ExternalSyntheticLambda0.onActivityResult(Unknown Source:2)
	at androidx.activity.result.ActivityResultRegistry.doDispatch(ActivityResultRegistry.java:392)
	at androidx.activity.result.ActivityResultRegistry.dispatchResult(ActivityResultRegistry.java:351)
	at androidx.activity.ComponentActivity.onRequestPermissionsResult(ComponentActivity.java:667)
	at androidx.fragment.app.FragmentActivity.onRequestPermissionsResult(FragmentActivity.java:612)
	at android.app.Activity.dispatchRequestPermissionsResult(Activity.java:8290)
	at android.app.Activity.dispatchActivityResult(Activity.java:8140)
	at android.app.ActivityThread.deliverResults(ActivityThread.java:5073)
	at android.app.ActivityThread.handleSendResult(ActivityThread.java:5121)
	at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:51)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2183)
	at android.os.Handler.dispatchMessage(Handler.java:107)
	at android.os.Looper.loop(Looper.java:241)
	at android.app.ActivityThread.main(ActivityThread.java:7604)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:941)

File: app/build.gradle.kts (hunk 5 above)
Comment (westnordost) [reply app/build.gradle.kts:190]: Well, mainline (`master`) will be the release on github and on google play, signed with my private key.

Maybe it is possible to exclude dependencies via a gradle build config constant. E.g. the `AddCyclewayWidth` quest could exist without AR, just the button to start the AR activity would be missing.

File: app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt (hunk 6 above)
Comment (westnordost) [reply app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt:170]: Ah I see, makes sense. Thanks!

Comment (westnordost) [reply app/src/main/java/de/westnordost/streetcomplete/measure/MeasureActivity.kt:170]: Then the checks at least work as intended.

**Comment by westnordost** [comment]
Thanks @HolgerJeromin , I'll look into it!

File: app/build.gradle.kts (hunk 5 above)
Comment (westnordost) [reply app/build.gradle.kts:190]: Actually, all of [Google Filament](https://github.com/google/filament) (rendering engine), [Google Sceneform](https://github.com/google-ar/sceneform-android-sdk) (connecting ARCore/Filament easily) and the [AR Core SDK](https://github.com/google-ar/arcore-android-sdk) are open source. 

What is not open source is the ARCore (aka "Google Play Services for AR") app. That app does all the heavy lifiting, image processing etc.
The AR Core SDK just communicates with this app and makes sure that that app is installed and up-to-date before doing anything useful. But **communicating** with a closed source app doesn't sound like it would violate against F-Droid guidelines.
If the user cannot install the app because the Google Play Services are not installed at all (such in the case of @FloEdelmann ) or the user doesn't want to install the app, it will simply do nothing. 🤷

Comment (smichel17) [reply app/build.gradle.kts:190]: Ah, cool! I didn't realize— usually it's Android components which are Open Source and everything Google-branded is proprietary, so I didn't look further.

Yes, that should be fine for F-Droid. There might be an antifeature tag (similar to `NonFreeNet` for apps which use non-free network services), but I'm not sure. A 'wait and see' approach should be fine, in that case.

**Comment by westnordost** [comment]
Here is the updated APK
https://github.com/streetcomplete/StreetComplete/actions/runs/1784019575

**Comment by mcliquid** [comment]
Thank you @westnordost for the APK. I just installed that version, but the app is crashing when calling the AR feature. I've sent you the error reports by mail. If I can further assist, please give me a note :)

**Comment by westnordost** [comment]
Uhm, should have tested it

**Comment by westnordost** [comment]
Here's another one: https://github.com/streetcomplete/StreetComplete/actions/runs/1786937861

**Comment by peternewman** [comment]
> Here's another one: https://github.com/streetcomplete/StreetComplete/actions/runs/1786937861

Well this one works for me (after I updated the AR tools which it prompted for). On Sony Xperia XZ1 Compact with Android 9.

Measuring two carpet tiles, which appear to actually be 1m, it came up with 1.03m, which could be my use of the arrow as much as anything else. I'll try and do some real world tests soon.

If this gets merged, it could do with some guidance/UI around it, it was a bit random out of the box!

**Comment by andrewharvey** [comment]
Tested on a Samsung Galaxy S20+, it works well. Very much looking forward to some width quests.

**Comment by HolgerJeromin** [comment]
You have to get used to it. There is explanation needed when integrating the feature.
I could only measure stuff on a flat surface with a bit of a distance (no height). But measuring on the floor is working quite good.

* Measuring a 30cm ruler results in 29-31 cm.
* Measuring my 60x60 cm² floor tiles was pretty good, too.
* But measuring in the region of the surface of a table was a disaster because the AR was confused by the table and floor surface. By moving the phone up and down I could change the measurement (tried to measure the distance of the big legs on the floor or the wood surface) from 1 meter to 3 meters.

**Comment by westnordost** [comment]
Yeah as said, please test it outside on the sidewalk. I don't plan to provide an explanation other than what is already there, users will just to try it out. They can't break anything, as the result is not applied immediately but the user will be led back to the then-filled-out width form.

> For a fair test setup and comparison between devices, let's presume: daytime, outside but cloudy/rainy, measurement of sidewalk width (sidewalk pattern will be similar everywhere)

It would be great if you could do this and state your device, so we have a few random samples on how well/precise different smartphones are able to perform in AR.

**Comment by westnordost** [comment]
I am also thinking about removing this visualization of the detected plane (those dots), as the arrow as a cursor should already be enough. Also, the plane render doesn't render vertical planes. But vertical planes should be detected too. Try to measure the height of a house or something. It will likely not work on surfaces that are just one color (like room walls).

**Comment by HolgerJeromin** [comment]
Measurement on the ground are working. Width of a bike parking for example. 
But I could not measure any height at various buildings.
The arrow did not start / stopped an the bottom of the wall..
Tested with a very structured brick wall.

Motorola g7 power, 15:00 center of germany, cloudy but not dark.

**Comment by peternewman** [comment]
> For a fair test setup and comparison between devices, let's presume: daytime, outside but cloudy/rainy, measurement of sidewalk width **(sidewalk pattern will be similar everywhere)**

The surface quest says otherwise! Paving stones or asphalt for starters? 😄

Do you have intentions of what to measure with it (both in terms of surface and scale); I'd imagine different results measuring a road width versus a pavement width, and likewise heights too; even measuring a building is going to be very different to measuring the empty space under a bridge for example.

**Comment by westnordost** [comment]
This is my experience as well, ARCore really has some problems to detect vertical planes. But it is enabled, maybe give it some time. And how precise is it, when compared to a measure tape @HolgerJeromin ? Because I plan to reduce displayed and tagged precision to possibly even 0.1m.

What I imagine what be measured with it:
- cycleway width
- footway/path width
- width of streets assumed to be narrow
- height of tunnels that are unsigned, same with below bridges

I am going to start with cycleway width.

**Comment by mcliquid** [comment]
The new version works with my Google Pixel Phone perfectly. I tried it several times indoor and outdoor and in my measurements I got a stable detection with an tolerance of 5cm plus minus in different ranges. In my opinion this is quite accurate for the distances OSM asked for.

**Comment by westnordost** [comment]
plus minus 5cm expected imprecision would mean that it would be best to restrict precision when tagging to one decimal, e.g. `width=1.2` or 4 inches, e.g. `width=4'0''`

File: app/src/main/res/values/strings.xml [hunk 7]
```diff
@@ -1204,14 +1204,15 @@ Alternatively, you can leave a note (with a photo)."</string>
...
-    <string name="no_camera_permission_warning">The permission is necessary to be able to measure distances with augmented reality.</string>
+    <string name="no_camera_permission_warning">The permission is necessary to be able to measure distances with the camera.</string>
     <string name="ar_measure">Measure</string>
     <string name="ar_core_error_sdk_too_old">This app is too old, try updating.</string>
     <string name="ar_core_tracking_error_bad_state">Something went wrong. Try restarting the measuring.</string>
     <string name="ar_core_tracking_error_insufficient_light">Too dark here</string>
     <string name="ar_core_tracking_error_excessive_motion">Too much movement</string>
     <string name="ar_core_tracking_error_insufficient_features">Insufficient visual features here</string>
     <string name="ar_core_tracking_error_camera_unavailable">Camera is used by another app</string>
+    <string name="default_disabled_msg_no_ar">You need a measure tape to solve this quest because measuring with the camera is not supported on this device.</string>
```
FloEdelmann, line 1215: According to [Wikipedia], it should be "tape measure" or "measuring tape", but not "measure tape". @smichel17 @peternewman which one sounds better?

smichel17: I **slightly** prefer "measuring tape", but I suspect it is a regional preference. In the line above, I would pick "tape measure" to avoid using the word "measuring" later in the sentence.

I also think it would be best to make the second part active voice ("this device does not support measuring with the camera"). However, it's a pet peeve of mine when apps tell me what "my device" cannot do when it's actually a software limitation (Google Play Services not installed), so I would prefer a phrasing like this:

```suggestion
    <string name="default_disabled_msg_no_ar">You need a measuring tape to solve this quest because StreetComplete is unable to measure with the camera on this device.</string>
```

Or a more awkward but more technically accurate version:

```suggestion
    <string name="default_disabled_msg_no_ar">You need a tape measure to solve this quest because this device seems not to support measuring with the camera on this device.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1202,4 +1202,19 @@ Alternatively, you can leave a note (with a photo)."</string>
...
+    <string name="ar_core_error_sdk_too_old">This app is too old, try updating.</string>
+    <string name="ar_core_tracking_error_bad_state">Something went wrong. Try restarting the measuring.</string>
+    <string name="ar_core_tracking_error_insufficient_light">Too dark here</string>
+    <string name="ar_core_tracking_error_excessive_motion">Too much movement</string>
+    <string name="ar_core_tracking_error_insufficient_features">Insufficient visual features here</string>
+    <string name="ar_core_tracking_error_camera_unavailable">Camera is used by another app</string>
+    <string name="default_disabled_msg_no_ar">You need a tape measure to solve this quest because measuring with the camera is not supported on this device.</string>
+
+    <string name="quest_cycleway_width_title">"What’s the width of this cycleway here?"</string>
+    <string name="quest_road_width_title">"What’s the width of this road have?"</string>
```
FloEdelmann, line 1218:
```suggestion
    <string name="quest_road_width_title">"What’s the width of this road here?"</string>
```

westnordost:
Success!

Screenshot_20220210-164115_StreetComplete Dev

michaelblyons:
Very cool. So it would work for round-top tunnels, even if you need to measure the edge?

```
   _________
  ╱         ╲ ↙ here to ground
 ╱╭───╮ ╭───╮╲
╷ │CAR│ │CAR│ ╷
│ └┰─┰┘ └┰─┰┘ │
```

westnordost:
From my end ready to merge. Anyone wants to look through the code or test it?

Here's a debug APK:
https://github.com/streetcomplete/StreetComplete/actions/runs/1825403355

FloEdelmann (approved):
That were quite some changes to go through :sweat_smile:
As I can't test the AR functionality on my phone or on the emulator, I only had a look at the code and the quest forms without AR. It seems that (at least on my device, OnePlus One, Android 10, LineageOS 17.1, no Google Play Services, microG 0.2.211515) opening the new quests is a lot laggier than opening other quests. Does it check for AR capabilities every time? Maybe it can be cached for the session?

Other than that, it looks fine from my side. Good job!

westnordost:
Hmm, it does check every time, but the check should be in the background. That's what the documentation claims. Plus, I assume after the first time the ArCore SDK downloaded the list of supported devices, it would keep it local somewhere. Not sure where I can check this, i.e. what HTTPS calls the device makes during this check

westnordost:
> I was able to enter 99999 meters... :-)
> Entering ".0002" is shown in undo screen as "2.0E-4"

I was thinking about adding sanity checks to the input but kind of decided against it. Though what I could add is a check whether it has more than 2 decimal places (less than one 1cm) and simply don't allow this input.

michaelblyons:
> &hellip; I was able to enter 99999 meters... :-)

212 storey building

https://www.engadget.com/flight-simulator-open-street-map-building-205545509.html

rugk:
I did not check the details here, but [as far as I see ARCore is not fully open-source and ships with proprietary binaries](https://github.com/google-ar/arcore-android-sdk/blob/6b01498cbd1e334178dabb5ade9f29e8350139ee/LICENSE#L2-L15) under some Google ToS.

How do you envision the F-Droid build version to work here? Just strip that feature? Or do we have an alternative lib?

Also, did any anyone already link the related issues https://github.com/streetcomplete/StreetComplete/issues/879 and e.g. the one possibly use-case https://github.com/streetcomplete/StreetComplete/issues/698? Well, aparently no, so I just did so… (I know, bad rhetorical questions here… :upside_down_face: )

westnordost:
The F-Droid version will work just like the normal version **IF** the user has the google play services installed and chooses to install ARCore (if it is not already installed). If not, the button to start the AR view will simply not display.

rugk:
Ah okay, hmm, maybe if one installs that proprietary app it could also be made to work with microG or so at leats. In any case for all those affacted, I have opened https://github.com/microg/GmsCore/issues/1672 already.

ale5000-git:
@westnordost
Is there an apk with this change included to test?

westnordost:
not yet, I want to wait till Google Play reviewed the v40.1 update before I post another beta. There is only the debug APK:

https://github.com/streetcomplete/StreetComplete/actions/runs/1825403355

ale5000-git:
> not yet, I want to wait till Google Play reviewed the v40.1 update before I post another beta. There is only the debug APK:
> 
> https://github.com/streetcomplete/StreetComplete/actions/runs/1825403355

It say it is expired and cannot be downloaded.

mcliquid:
I tried the last debug APK a bit and could see quite good results up to about 2 meters.
All distances above 2 meters bring problems. Especially the range around 3 and 4 meters hardly works.
If I stand on a sidewalk next to a street and want to measure the street, the AR measurement hangs completely. The line can not be drawn over 3 meters even by swiveling the device. I have tried it from several places and always had the same problem. Even if I stand in the middle of the road does not help.
Has anyone been able to find similar problems?

licaon-kter:
@westnordost 
> The F-Droid version will work just like the normal version

It's not about it working, if it's not FOSS it will not be build on F-Droid.

FloEdelmann:
@licaon-kter See https://github.com/streetcomplete/StreetComplete/pull/3709#discussion_r797575254

The newest version is already built and released on F-Droid: https://f-droid.org/de/packages/de.westnordost.streetcomplete/

licaon-kter:
Does the app show this https://github.com/google-ar/arcore-android-sdk#user-privacy-requirements when AR is available and used? There's a Yes/Cancel thing?

xax:
> I don't plan to provide an explanation other than what is already there, users will just need to try it out.

Trying the feature out – I didn't come to a conclusion how it is supposed to work. I think, it *needs* explanation.

<details>
Sometimes an icon showing a hand on a phone circles cryptically across the live camera image shown. Then eventually sometimes an arrow pointing down appears at seemingly random locations (if it does not appear: good luck guessing why and what to do about it), drifting erratically when tilting or turning the phone. Upon a touch of the screen, tilting/turning does indeed display a measuring line, the arrow now at the other end of that line; turning the phone the arrow moves against the cam image but ever so often stops before "reaching" the other side of the road (in the instance of trying to measure the road width); a bubble note gives the mysterious advice "to stay near the arrow (and mind the traffic)" – but trying to walk with the arrow makes you move the entire measurement line or "overtaking" the arrow.
</details>

A brief explanation should tell the user:
1. how to start
     - where to point initially,
     - how to register start point: touch arrow, screen, what?
     - what feedback to expect,
2. how to proceed
    - point by turning/tilting? walking?
3. how to conclude measurement procedure
   - use touch (– touch what: arrow, measurement value bubble, screen?)? back key?

mnalis:
Shouldn't description on how ARCore app works be described somewhere on the internet? If so, perhaps a link on how to use ARcore might be provided in StreetComplete [FAQ](https://wiki.openstreetmap.org/wiki/StreetComplete/FAQ) (by someone who has it working and can verify that it the instructions are correct; i.e. not me). Or _shortly_ described directly there...

HolgerJeromin:
Short one:
* If you see the wavy hand: point to the bottom of the street and show the AI more of the surface and do not try in your living room wall or furniture
* stand at the side of the street
* move the phone so the arrow is exact at the border 
* push the screen (anywhere)
* walk across the street. Line should be drawn
* if the arrow is at the final position touch the screen again > line is locked

endim8:
Support for Huawei's AR Engine would be really great too :)

https://developer.huawei.com/consumer/en/hms/huawei-arengine/

some Huawei phones (for example my p40 pro) have a time-of-flight sensor which are used for AR, and would probably give very accurate measurements in streetcomplete 😄

mnalis:
Regarding Huawei phones and AR not working there: https://github.com/streetcomplete/StreetComplete/issues/3917

mnalis:
@rugk although adding **Google** app on **_Google-free_** device kind of defeats the point I would think :laughing:

rugk:
I don't think it necessarily does. If your aim is to have a 100% FLOSS system, sure. However, if your aim is just to have a privacy-respecting, secure system installing a sandbkxed app and preventing internet access seems fine.
I think everyone can decide for oneselve whether to do this or not.

PR #3710, opened 2022-02-01
Ask for name, opening hours, wheelcair access for some `healthcare` tags

FloEdelmann:
All of them are included in https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/assets/osmfeatures/default/de.json, so translations should be fine. I haven't checked other languages though.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -86,9 +86,10 @@ class AddOpeningHours (
             "healthcare" to arrayOf(
                 // common
                 "audiologist", "optometrist", "counselling", "speech_therapist",
+                "sample_collection", "blood_donation",
 
                 // name & opening hours
-                "sample_collection", "laboratory", "blood_donation", "physiotherapist", "podiatrist",
+                "physiotherapist", "podiatrist",
             ),
         ).map { it.key + " ~ " + it.value.joinToString("|") }.joinToString("\n or ") + "\n" + """
             )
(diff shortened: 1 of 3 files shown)


matkoniecz:
Any reason why `"sample_collection", "blood_donation", "physiotherapist", "podiatrist",` are excluded from wheelchair quest?

I would skip  `"laboratory"` - it can be used according to its definition to internal features not accessible at all (often `healthcare=sample_collection` are marked as laboratories - or to be more specific there was not even a tagging for them before https://wiki.openstreetmap.org/wiki/Proposed_features/healthcare%3Dsample_collection ).

`healthcare=laboratory` + `sample collection=yes` indicates that it is publicly accessible but it is barely used.

Hmmmm. Maybe quest for `healthcare=laboratory` which would add `sample collection=yes` or `sample collection=no` or retag to `healthcare=sample_collection` would be a good idea? (disclaimer: I invented this tags and wrote most of https://wiki.openstreetmap.org/wiki/Proposed_features/healthcare%3Dsample_collection )

FloEdelmann:
For `physiotherapist` and `podiatrist`, it might be difficult to determine if they can actually treat wheelchair patients.

For `sample_collection` and `blood_donation`, there is no real reason. I'll add them.

> often `healthcare=sample_collection` are marked as laboratories

That's why I added them. If you think it will be difficult to determine there, I'll remove it though.

> Maybe quest for `healthcare=laboratory` which would add `sample collection=yes` or `sample collection=no` or retag to `healthcare=sample_collection` would be a good idea?

If they can be "internal features not accessible at all", then it's probably not a good idea. Additionally, I think that determining "yes" is probably mostly easy, but "no" will often be much harder.

matkoniecz:
Looks great to me, though I have not tested code in action so far.

westnordost (approved):
Looks like a good addition, thank you!

PR #3695, opened 2022-01-30
Show pillar hydrant images in hydrant position quest

FloEdelmann:
Fixes #3686.

PR #3711, opened 2022-02-01
Fix size of fire hydrant images

FloEdelmann:
Follow-up to https://github.com/streetcomplete/StreetComplete/pull/3695#issuecomment-1027255554.


westnordost:
Thank you!

PR #3712, opened 2022-02-02
make bicycle parking and playground access quests less confusing

### LINKED ISSUE #3693: Text for "access" quests sometimes confusing (e.g. for playground)
- Due to changes in texts, some quests now sound very strange, eg. `Is this playground publicly accessible` : `Any member of the general public` does not make much sense as an answer
- While in others it sounds normal `Who is allowed to park here?` : `Any member of the general public` 

(might be related to changes in https://github.com/streetcomplete/StreetComplete/discussions/3429, https://github.com/streetcomplete/StreetComplete/pull/3473)

small_Screenshot_20220130_002947_de westnordost streetcomplete small_Screenshot_20220130_001844_de westnordost streetcomplete small_Screenshot_20220130_001824_de westnordost streetcomplete

I propose they all (I found those quests, don't know if there are more of them?) be changed to sound similar:
- car parking remains the same:  `Who is allowed to park here? Parking may be free or paid.`
- bicycle parking changes to: `Who is allowed to park a bicycle here? Parking may be free or paid.` (I've also changed `bike` to `bicycle`, as [bike](https://en.wikipedia.org/wiki/Bike) might refer to motorcycle/motorbike too)
- playground: `Who is allowed to use this playground?`

matkoniecz:
fixes #3693

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddBikeParkingAccess.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/parking_access/AddBikeParkingAccess.kt
@@ -21,7 +21,7 @@ class AddBikeParkingAccess : OsmFilterQuestType<ParkingAccess>() {
 
     override val questTypeAchievements = listOf(BICYCLIST)
 
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_bicycle_parking_access_title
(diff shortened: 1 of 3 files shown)


FloEdelmann (approved):

peternewman (approved):
LGTM

PR #3715, opened 2022-02-03
Convert remaining Java files to Kotlin, switch to kaml for YAML parsing

FloEdelmann:
References https://github.com/streetcomplete/StreetComplete/pull/3692#discussion_r798379994.

~The only Java file that is still missing after this PR is [`CountryInfo.java`](https://github.com/streetcomplete/StreetComplete/blob/6aca0dc9476034466d04949ffe870075932edb96/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.java). I left it out for now because `CountryInfos.kt` [still uses reflection there](https://github.com/streetcomplete/StreetComplete/blob/1c0f7346de18db7068b019448d3f3db698995308/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt#L73-L99), which breaks after converting it to Kotlin.~

~I'll try making `CountryInfo` a Kotlin data class that can be serialized with Kotlin Serialization, and then switch from [YamlBeans](https://github.com/EsotericSoftware/yamlbeans/) to [kaml](https://github.com/charleskorn/kaml) for parsing.~

**EDIT:** Done now, see the following comments.


westnordost:
Hmm, would you be up for adding in some architectural improvements in this PR? Those files are one of the oldest in the project and today I'd do it a bit differently (separating platform-code from application logic). A short list:

- separate `CountryInfos` logic from loading the file from the assets. E.g. maybe have a `CountryInfoParser` that creates `CountryInfo` from an input string and a `AssetsCountryInfoLoader` or something to get those from the assets. Then, the tests for `CountryInfos` can also go into `test` and not `androidTest`
- Same for `Abbreviations`. The `Abbreviations` class should not also contain the code to load/parse it but just the abbreviation logic
- weird, there are no tests for `Abbreviations`. Maybe because it has been too tightly coupled with resource loading for me to think it is worth it to write tests? 

kaml, hmm, this is not an official dependency though. And it only targets the JVM anyway, i.e. is not multiplatform. Writing a "manual" parser (after parsing the yaml to a map<string,any?>) should be just as many lines of code as there are properties, so about ~40. I think this'd be fine.

FloEdelmann:
> kaml, hmm, this is not an official dependency though

Ahh, I've seen this too late. Now I managed to do this. But we use YAML in a few different places with different data structures, so writing a parser ourselves would be a bit more involved.

> weird, there are no tests for `Abbreviations`.

There are: https://github.com/streetcomplete/StreetComplete/blob/9c6d3e25216d06a2c5afa71086949e1e195de926/app/src/test/java/de/westnordost/streetcomplete/data/meta/AbbreviationsTest.kt

> would you be up for adding in some architectural improvements in this PR?

If it can wait a few days, sure.

westnordost:
> Ahh, I've seen this too late. Now I managed to do this. But we use YAML in a few different places with different data structures, so writing a parser ourselves would be a bit more involved.

Misunderstanding. I don't mean that you should write a YAML parser but let YamlBeans parse it to a Map<String, String> and put the data from the map into the constructor of a `CountryInfo`.

FloEdelmann:
> let YamlBeans parse it to a Map<String, String> and put the data from the map into the constructor of a `CountryInfo`

I tried that now, but it seems that YamlBeans parses numbers and booleans as strings if it doesn't have a class definition acting as a type schema. Still worth continuing this way?

westnordost:
Hm and kaml would do all that automatigally like kotlinx-serialization does it? I didn't look at the code - would it be easily exchangable if kotlinx-serialization ever had an official yaml implementation?

westnordost:
I have seen the manual filling of `CountryInfo` to do some more parsing right when parsing it, e.g. convert `meter` and `foot and inch` in `lengthUnits.yml` (etc) right into an enum so the users of `CountryInfo` don't need to parse it

westnordost:
Hmm, what about

```kotlin

@Serializable
internal class IncompleteCountryInfo(
// all is nullable
)

interface CountryInfo {
// with all the properties that should not be nullable not nullable
}

@Serializable
internal class DefaultInfo : CountryInfo {
// anything that is not null in the default.yml is also not nullable here
}

internal class MergedCountryInfo(val info: CountryInfo, val default: DefaultInfo) : CountryInfo {
  val aProperty get() = info ?: default
}

```

FloEdelmann:
I've rebased the branch onto `master` now, so the commits are clean and easier to review.

> separate CountryInfos logic from loading the file from the assets

I implemented your idea with the two data classes `IncompleteCountryInfo` and `CountryInfo`; the latter takes a list of `IncompleteCountryInfo` objects, and for each property returns the first non-null (or non-empty for lists) value. It works great, and doesn't need reflection anymore :tada: 

So most logic is already in `CountryInfo`. Should `CountryInfos` be separated further?

> Then, the tests for CountryInfos can also go into test and not androidTest

The tests currently mostly check if the YAML files are valid (or can be parsed correctly), so they need all the resource loading stuff from androidTest.

> manual filling of `CountryInfo` as an opportunity to do some more parsing right when parsing it

`lengthUnits`, `speedUnits` and `weightLimitUnits` are parsed now as `LengthUnit`s, `SpeedMeasurementUnit`s and `WeightMeasurementUnit`s, respectively. That doesn't even require writing custom serializers, just some annotations for the values in the YAML files. I really like how concise that got!

> Same for `Abbreviations`

I'll wait with that until you reviewed the rest of the code, so any adjustments don't need to be done twice.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
```diff
@@ -0,0 +1,116 @@
...
+    val officialLanguages: List<String> = emptyList(),
+    val orchardProduces: List<String> = emptyList(),
+    val popularReligions: List<String> = emptyList(),
+    val popularSports: List<String> = emptyList(),
+    val regularShoppingDays: Int? = null,
+    val roofsAreUsuallyFlat: Boolean? = null,
+    val slowZoneLabelPosition: String? = null,
+    val slowZoneLabelText: String? = null,
+    val speedUnits: List<SpeedMeasurementUnit> = emptyList(),
+    val weightLimitUnits: List<WeightMeasurementUnit> = emptyList(),
```
westnordost, line 52: Any types used in CountryInfo should probably be moved to this directory or even into this file.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Locale
+
+@Serializable
+data class IncompleteCountryInfo(
+    // this value is not defined in the yaml file but it is the ISO language code part of the file name!
+    // i.e. US for US-TX.yml
+    val countryCode: String,
+
+    // sorted alphabetically for better overview
+    val additionalStreetsignLanguages: List<String> = emptyList(),
```
westnordost, line 16: Wouldn't it be easier to make all fields in `IncompleteCountryInfo` nullable instead of assigning some an empty list? Then, in `CountryInfo` below, you can just use 

```kotlin
additionalStreetsignLanguages: List<String> get() = 
    infos.firstNotNullOf { it.additionalStreetsignLanguages } ?: emptyList()
```

etc.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    val locale: Locale
+        get() = if (officialLanguages.isEmpty()) {
+            Locale.getDefault()
+        } else {
+            Locale(officialLanguages[0], countryCode)
+        }
+}
+
+private fun <T> List<IncompleteCountryInfo>.getFirstNonEmpty(getValue: (IncompleteCountryInfo) -> T): T {
```
westnordost, line 109: Can this function really return `null`. The comment in line 84 claims it does.

westnordost:
🥳 

> So most logic is already in CountryInfo. Should CountryInfos be separated further?

I think this is not necessary. Forcross platform, it would probably be better to reimplement that class. 

> I really like how concise that got!

Me too!

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val countryBoundaries: Future<CountryBoundaries>?,
+) {
+    private val countryInfoMap = HashMap<String, IncompleteCountryInfo?>()
+    private val defaultCountryInfo: IncompleteCountryInfo by lazy {
+        try {
+            loadCountryInfo("default")
+        } catch (e: Exception) {
+            // this should be in any case a programming error
+            throw RuntimeException(e)
+        }
```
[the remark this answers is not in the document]
FloEdelmann: I didn't know about checked exceptions in Java and looked it up just now. Actually, I did wonder when reading this part, why any exception is caught and then re-thrown as a `RuntimeException`. Now I know :sweat_smile: 

In Kotlin, this is explicitly not needed: https://kotlinlang.org/docs/exceptions.html#checked-exceptions

So thanks for letting me know; I'll remove it.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
FloEdelmann: Yes, `T` can also be substituted with a nullable type.
If `getValue` returns null for all country infos, then the return type will be null.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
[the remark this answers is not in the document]
FloEdelmann: Fixed in [`9784201`](https://github.com/streetcomplete/StreetComplete/pull/3715/commits/9784201632ecc0dc5d9b4327cbd448a8dde9235e).

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    /** Get the info by location */
+    fun get(longitude: Double, latitude: Double): CountryInfo {
+        try {
+            val countryCodesIso3166 = countryBoundaries?.get()?.getIds(longitude, latitude)
+            return get(countryCodesIso3166 ?: emptyList())
+        } catch (e: Exception) {
+            throw RuntimeException(e)
+        }
+    }
```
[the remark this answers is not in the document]
FloEdelmann: Fixed in [`855ccf2`](https://github.com/streetcomplete/StreetComplete/pull/3715/commits/855ccf2a2658b5adba2b08f566906a312d25cf17).

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    /** Get the info by a list of country codes sorted by size. I.e. DE-NI,DE,EU gets the info
+     * for Niedersachsen in Germany and uses defaults from Germany and from the European Union */
+    fun get(countryCodesIso3166: List<String>): CountryInfo =
+        CountryInfo(countryCodesIso3166.mapNotNull { get(it) } + defaultCountryInfo)
+
+    private operator fun get(countryCodeIso3166: String): IncompleteCountryInfo? {
```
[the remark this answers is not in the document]
FloEdelmann: Fixed in [`4587515`](https://github.com/streetcomplete/StreetComplete/pull/3715/commits/458751526c753641a19d44a3902782eb026c6801).

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
FloEdelmann: Fixed in [`3a7a88a`](https://github.com/streetcomplete/StreetComplete/pull/3715/commits/3a7a88a248ffe708666c263e708baf6a333a25f6).

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
FloEdelmann: Though this is not relevant anymore with [`8f740dc`](https://github.com/streetcomplete/StreetComplete/pull/3715/commits/8f740dc081a50b52fe7150a7f9e762765c416deb).

app/src/main/java/de/westnordost/streetcomplete/ktx/Resources.kt
```diff
@@ -6,15 +6,23 @@ import android.graphics.drawable.BitmapDrawable
...
-inline fun <reified T> Resources.getYamlObject(@RawRes id: Int): T =
-    YamlReader(BufferedReader(InputStreamReader(openRawResource(id)))).read(T::class.java)
+inline fun <reified T> Resources.getYamlObject(serializer: KSerializer<T>, @RawRes id: Int): T =
+    Yaml.default.decodeFromStream(serializer, openRawResource(id))
+
+fun Resources.getYamlStringMap(@RawRes id: Int): Map<String, String> =
+    this.getYamlObject(MapSerializer(String.serializer(), String.serializer()), id)
+
+fun Resources.getYamlStringList(@RawRes id: Int): List<String> =
+    this.getYamlObject(ListSerializer(String.serializer()), id)
```
westnordost, line 25: It is weird that kaml/kotlinx.serialization doesn't work with generics. For kotlinx.serialization JSON this was no issue. Is this a kaml limitation?

There is also 
https://github.com/Kotlin/kotlinx.serialization/tree/master/formats#yaml-multiplatform
which we might want to use some day anyway (because multiplatform) if not all places where YAML is parsed is in the same place as accessing Android resources.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
```diff
@@ -0,0 +1,56 @@
...
+        private const val BASEPATH = "country_metadata"
+    }
+}
+
+fun CountryInfos.getByLocation(
+    countryBoundaries: CountryBoundaries,
+    longitude: Double,
+    latitude: Double,
+): CountryInfo =
+    get(countryBoundaries.getIds(longitude, latitude) ?: emptyList())
```
westnordost, line 56: Shouldn't be in the same file, IMO, so that CountryInfos does not have an (implicit) dependency on CountryBoundaries. Mabye in `CountryInfosXt.kt` or in the /ktx/ directory? (Not sure if it would be cleaner to only have extensions of things outside this project in that directory or if any utility extensions should go there.)

Also, could just be named `get`

app/src/main/java/de/westnordost/streetcomplete/data/meta/Abbreviations.kt
```diff
@@ -0,0 +1,86 @@
+package de.westnordost.streetcomplete.data.meta
+
+import java.util.Locale
+import java.util.regex.Matcher
+import java.util.regex.Pattern
```
westnordost, line 5: Hmm, is there any reason why to use the Java Regex stuff and not the Kotlin regex stuff?

westnordost (approved):
Regarding the `FutureTask<*>`, this'd actually be the next thing on my list of java deps to get rid of. I think the appropriate replacement is `Deferred`? On the other hand, we probably don't want all code that uses country boundaries (etc.) to be suspending.

Anyway, getting rid of `FutureTask` that would be something for a future task (haha).

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
westnordost: Oh well nevermind. 

I think the best would really be if the extension functions were not all collected in one place but placed each where they belong functionally. So I guess this place is as good as any other.

app/src/main/java/de/westnordost/streetcomplete/ktx/Resources.kt
FloEdelmann: > It is weird that kaml/kotlinx.serialization doesn't work with generics

It should work fine (e.g. `List<LengthUnit>` in `CountryInfo` does serialize just fine), but we need a top-level declaration here. And since there is no `data class` for which it can generate a serializer at compile time, we have to manually construct the serializer ourselves: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#constructing-collection-serializers

This would be the same for JSON, so it's not a kaml limitation.

FloEdelmann: > There is also
[Kotlin/kotlinx.serialization@master/formats#yaml-multiplatform](https://github.com/Kotlin/kotlinx.serialization/tree/master/formats?rgh-link-date=2022-02-19T13%3A29%3A04Z#yaml-multiplatform) […]
> Edit: But it doesn't support multiline strings yet and at least for the changelog, we use them.

So I'd say let's stick to kaml for now; it should be quite easy to switch to another kotlinx.serialization YAML plugin, because the serializers could stay the same.

app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfos.kt
FloEdelmann: We already have `CountryInfo.kt` in the `ktx` directory. Can it go in there, or should I create a new `CountryInfos.kt` or `CountryInfosXt.kt` there?

app/src/main/java/de/westnordost/streetcomplete/data/meta/Abbreviations.kt
FloEdelmann: Ooops, I didn't know there was a difference. Android Studio's automatic Java to Kotlin conversion did that and I didn't question it.

FloEdelmann: I've refactored it to use Kotlin `Regex`es in [`4086e04`](https://github.com/streetcomplete/StreetComplete/pull/3715/commits/4086e0438db5b2bf76567e94fe973b5bbaa871a2).

westnordost: Nice, you even made it shorter (and more performant)!

westnordost (approved):

westnordost:
Thanks, to you

image

🥳

app/src/main/java/de/westnordost/streetcomplete/ktx/Resources.kt
FloEdelmann: I found a way to simplify this, see https://github.com/streetcomplete/StreetComplete/commit/adf9a7890076bfcf12862b2ad87e182a8ba11722 :tada:

westnordost:
Nice!

PR #3719, opened 2022-02-04
better deal with bad OSM data for OH

### LINKED ISSUE #2712: Don't check opening hours of disused shops or amenities
see [here](https://www.openstreetmap.org/note/2602788)

### LINKED ISSUE #3718: SC asks for opening_hours check for disused:shop
SC asks for opening_hours check for vacant shops. This makes no sense, because the shop is closed forever.

**How to Reproduce**

- Have a former shop node with disused:shop key and also the tag opening_hours with a meaningful value (the value of the former shop)
- SC asks to check the opening_hours although the node is already marked as vacant

I am not sure about the desired behavior of the app. Maybe it should not even ask for a opening hours check or it could help to remove that tagging.

**Versions affected**

v40.0-beta1

matkoniecz:
fixes #3718 and fixes #2712

now old opening hours will be asked only when on objects eligible for new opening hours
and for some additional limited cases when opening hours are tagged already and become old such as parks, barriers, toilets

opening PR as I have feeling that it could be implemented in a bit better way


app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -22,7 +22,7 @@ class AddOpeningHours(
 
     /* See also AddWheelchairAccessBusiness and AddPlaceName, which has a similar list and is/should
        be ordered in the same way for better overview */
-    private val filter by lazy { ("""
```
matkoniecz, line 25: dropping private here seems smelly to me, but was needed for tests

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -93,12 +93,25 @@ class AddOpeningHours(
...
             )
-            and !opening_hours
+            and (!opening_hours or opening_hours older today -1 years)
+          )
+          or
+          (
+          opening_hours older today -1 years
+          and
+            (
+            leisure=park
```
matkoniecz, line 103: I think that it is worth supporting (17 000 `leisure=park` with OH data, see http://overpass-turbo.eu/?q=LyoKVGhpcyBoYcSGYmVlbiBnxI1lcmF0ZWQgYnkgdGhlIG92xJJwxIlzLXR1cmJvIHdpemFyZC7EgsSdxJ9yaWdpbmFsIHNlxLBjaMSsxIk6CsOiwoDCnG9wxI3EumdfaG_EqHM9KiBhbsSXbGXEhcSoZT3Eo3JrxI9sb2LEvMWIwp0KKi8KW8WSdDpqc29uXVt0aW1lxbA6MjUwMF07Ci8vxI_ElMSdciByZXN1bHRzCigKICDGhiBxdcSSxJrFoXQgZm9yOiDFiMWKxYxuxY7FkMWScsWUxZbFmMWaxZzGj8aNxaDEsGvFqcaWIG5vZGVbIsWLxY1uxY_FkcWTIsW3IsWbxZ3GjSI9IsWha8eIxoTGl3dheceAx4LGqseExqzHh8eJx4vGtGXHjseQxrfHk8a6xo1sxJRpxbXHmcapxqvHhsaux4jHgMegxZ7Ho8eRx6YpxoTGmXDEt27GoMaNxo_GkcaTxbAgY8WSx747&c=BOyB35NeIQ

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt
```diff
@@ -160,6 +163,60 @@ class AddOpeningHoursTest {
...
     }
 
+    @Test fun `isApplicableTo returns false for known places with recently edited opening hours`() {
+        assertFalse(questType.isApplicableTo(
+            node(tags = mapOf("shop" to "sports", "name" to "Atze's Angelladen", "opening_hours" to "Mo-Fr 10:00-20:00"), timestamp = currentTimeMillis())
+        ))
+    }
+
+    @Test fun `isApplicableTo returns true for known places with old opening hours`() {
+        val milisecondsFor400Days : Long =  34560000000 // 1000 * 60 * 60 * 24 * 400
```
matkoniecz, line 173: otherwise it overflows into a different value

there is likely a better way to handle this

peternewman (commented):
Sticking my comment in a more useful place, I think this wants applying to CheckOpeningHoursSigned too, which could have the same issues.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
peternewman: Is it worth adding that to the CheckOpeningHoursSigned quest too (possibly one for another PR)?

Although would it make sense to extract the list of things which might have opening hours into a common list?

matkoniecz:
https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt has far less complex filtering

Not sure is such elaborate filter useful there.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -93,12 +93,25 @@ class AddOpeningHours(
...
+            or
+            barrier
+            or
+            amenity=toilets
+            or
+            amenity=bicycle_rental
+            )
```
westnordost, line 110:
```suggestion
          or (
            opening_hours older today -1 years
            and (
              leisure = park
              or barrier
              or amenity ~ toilets|bicycle_rental
            )
```

indentation and formatting

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -93,12 +93,25 @@ class AddOpeningHours(
...
           )
-          or opening_hours older today -1 years
         )
         and access !~ private|no
-        and (name or brand or noname = yes or name:signed = no or amenity=recycling)
+        and (name or brand or noname = yes or name:signed = no or amenity=recycling or amenity=toilets)
```
westnordost, line 114:
```suggestion
        and (
          name or brand or noname = yes or name:signed = no
          or amenity = recycling or amenity = toilets or barrier
        )
```

too long line, formatting, barriers also usually have no name

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
westnordost: Yes, please don't.

Why don't you use `isApplicableTo`?

westnordost (changes requested):

westnordost:
Overall, I am somewhat "meh" about this PR. In #3718 @Strubbl clearly stated that the opening hours quest would only be shown in error if the map data is actually wrong.

So if the quest asks if the opening hours for that disused shop are still correct, the user can simply answer "doesn't exist anymore" and StreetComplete will just fix the problem by tagging the disused shop as a disused shop, removing opening hours etc.
So, I don't really see the issue.

matkoniecz:
> clearly stated that the opening hours quest would only be shown in error if the map data is actually wrong

While I agree that in general StreetComplete is allowed to fail in case of invalid data as it is impossible to handle all possible cases of how data can be broken, I think it is fine to filter out some of such cases. And this case the change seems to be worth it.

In such cases survey with SC is a pure waste of time - of SC mapper and by whoever is processing notes.

As person who is really involved in processing notes - such notes are quite irritating.

Technically there is some minimal benefit as it results in removal of `opening_hours` from a closed shop, but benefit of that is really minimal.

And time of SC mapper is definitely wasted as this problem can be trivially detected automatically.

> So if the quest asks if the opening hours for that disused shop are still correct, the user can simply answer "doesn't exist anymore" and StreetComplete will just fix the problem by tagging the disused shop as a disused shop, removing opening hours etc.

I just tested with https://www.openstreetmap.org/node/3087666570 and SC simply started opening a note rather than retag anything (which may be a separate issue)

Note that this problem happens for disused shops when user has "is this shop still vacant" below opening hours or when it is disabled. So it is not happening on default configuration.

westnordost:
> I just tested with https://www.openstreetmap.org/node/3087666570 and SC simply started opening a note rather than retag anything (which may be a separate issue)

Oh right, because the app first checks if the node in question is some kind of shop, and a `disused:shop` is rightly not detected as "some kind of shop". Makes sense, we don't want to potentially retag things that are clearly not a shop as another shop or vacant shop, plus that "some kind of shop" check is used in other places too.

matkoniecz:
> Overall, I am somewhat "meh" about this PR.

But would you be acceptable to merge it (as long as ugly code issues are solved)?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
matkoniecz: 1)

I wanted to test only filter here (as it is part that I modified). But likely testing entire actually run function makes more sense

2)

`isApplicableTo` is also checking `hasName` which calls `hasFeatureName`

`hasFeatureName` relies on `featureDictionaryFuture` which is just `mock()` - and as result tests were crashing with some esoteric stacktrace . Though there are some exsisting similar tests that actually work, I will look into it (if you are not fundamentally opposed to basic idea of this quest)

westnordost:
Yes, because https://github.com/streetcomplete/StreetComplete/pull/3719#issuecomment-1030827413

westnordost:
Hmm, unless we treat disused shops and amenities as "some kind of shop" only for the intent of replacing a shop. Would make sense too

westnordost:
`isKindOfShopExpression` already allows a prefix to be added, so the replace-shop functionality could look for isKindOfShopExpression() + "shop = vacant" + isKindOfShopExpression("disused"), the same way as `CheckShopType` does. I think this is the better solution at least to solve this case.

matkoniecz:
I think it may be useful in general, maybe with some exceptions. 

Correctly tagged `disused:shop=*` will be visible and useful for orientation, so it would make sense also in other contexts.

peternewman:
> https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt has far less complex filtering

Yes, although the flip side of that is it's currently only showing the quest for:
https://github.com/streetcomplete/StreetComplete/blob/6aa3fb16b18f74955ce79a19b201d5bb60707616/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt#L36

It would be nice to also cover the leisure and other amenities from here too.

> Not sure is such elaborate filter useful there.

The more practical thing may be that there only seem to be 50 odd disused things tagged with opening_hours:signed:
https://taginfo.openstreetmap.org/keys/opening_hours:signed#combinations

westnordost:
Alright, let's merge the original PR, though note that CheckOpeningHoursSigned should at least get `or amenity = toilets or barrier` too

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
westnordost: No I am not opposed. You can configure a mock to return what you want. Anyway, I think `hasFeatureName` is only called if the feature itself has no name, so you can just work around it by adding `name` as a tag.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -93,12 +93,22 @@ class AddOpeningHours(
...
+              or amenity ~ toilets|bicycle_rental
+            )
           )
-          or opening_hours older today -1 years
         )
         and access !~ private|no
-        and (name or brand or noname = yes or name:signed = no or amenity=recycling)
+        and (
+          name or brand or noname = yes or name:signed = no
+          or amenity = recycling or amenity = toilets or barrier
```
peternewman, line 110: Match the same logic as above to reduce the cognitive load when looking at the code:
```suggestion
          or barrier
          or amenity ~ toilets|bicycle_rental
```

peternewman (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
matkoniecz: Well, I wanted to test that it will work also for some specific unnamed features.

I will configure mock then instead of visibility hacks.

matkoniecz: Though looking at

```
    private fun hasFeatureName(tags: Map<String, String>): Boolean =
        featureDictionaryFuture.get().byTags(tags).isSuggestion(false).find().isNotEmpty()
```

it seems that I would need mock returning in `get()` a mock that would be returning another mock in `isSuggestion` and so on.

Building own `FeatureDictionary` for this single test also seems supremely annoying.

I think I would rather drop that test.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
```diff
@@ -33,7 +33,8 @@ class CheckOpeningHoursSigned (
             or older today -1 years
           )
           and access !~ private|no
-          and (name or brand or noname = yes or name:signed = no or amenity = recycling)
+          and (name or brand or noname = yes or name:signed = no
+          or amenity = recycling|toilets|bicycle_rental or leisure=park or barrier)
```
westnordost, line 37:
```suggestion
          and (
            name or brand or noname = yes or name:signed = no
            or amenity = recycling|toilets|bicycle_rental or leisure=park or barrier
          )
```

westnordost (approved):

westnordost:
Done, thanks!

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
```diff
@@ -33,7 +33,10 @@ class CheckOpeningHoursSigned (
             or older today -1 years
           )
           and access !~ private|no
-          and (name or brand or noname = yes or name:signed = no or amenity = recycling)
+          and (
+            name or brand or noname = yes or name:signed = no
+            or amenity = recycling|toilets|bicycle_rental or leisure=park or barrier
```
peternewman, line 38: Shouldn't we drop recycling from here too?

Also should this either be `~`, or can the one in AddOpeningHours be `=` too?

matkoniecz: Good catches!

Fixed `=` vs `~`

`amenity = recycling and recycling_type = centre` should be either added to name quest or allowed to have no name in opening hours quest, not checked which would be preferable

matkoniecz: https://wiki.openstreetmap.org/wiki/Tag:recycling_type%3Dcentre

> a public recycling centre

So asking for opening hours is fine

PR #3720, opened 2022-02-04
Resolve CI failures on master branch following on from #3692

adrianclay:
Fix some missed refactorings associated with move to Koin

I managed to miss these, including the whole of the androidTest folder, which is a little worrying.

Nevertheless, the android tests pass on my device.

I have been unable to test the generate-quest-list.main.kts change as I get the following error locally:
```
error: backend Internal error: Exception during IR lowering
The root cause java.lang.RuntimeException was thrown at: org.jetbrains.kotlin.backend.jvm.codegen.FunctionCodegen.generate(FunctionCodegen.kt:50) (generate-quest-list.main.kts): org.jetbrains.kotlin.backend.common.BackendException: Backend Internal error: Exception during IR lowering
```

However it has run within GitHub actions on my fork, which suggests it works.

### DIFF
--- a/.github/generate-quest-list.main.kts
+++ b/.github/generate-quest-list.main.kts
@@ -40,7 +40,7 @@ val wikiRowSpan2 = " rowspan=\"2\" |"
 main()
 
 fun main() {
-    val questFileContent = sourceDirectory.resolve("quests/QuestModule.kt").readText()
+    val questFileContent = sourceDirectory.resolve("quests/QuestsModule.kt").readText()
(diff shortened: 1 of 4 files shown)


westnordost:
Thabnk you!

there are not a lot of `androidTest`s though, as the UI of the app is generally not tested. The few things there are are pretty much all tests of `DAO`s. Maybe it would make sense to move all those `DAO` tests to the normal `test` and just add a SQLite database to the test dependencies.

westnordost:
Maybe @FloEdelmann can help regarding that generate-quest-list.main.kts

PR #3722, opened 2022-02-05
Only allow lit quest to change to steps on footway

arrival-spring:
At the moment the lit quest can change any highway to steps.

I have no evidence of this producing bad data, but the steps other answer should only be shown on ways that are likely to be steps.

I think that the only highway that the lit quest is shown for that could be changed to steps is `footway`.

Maybe this code isn't as readable now, it could be changed to more clearly only return the steps answer for `footway` which isn't an area - should I do that?


arrival-spring:
> What about `highway=path`?

Because `highway=path` isn't asked in the lit quest.

I think reason why not is because `path` is thought of as usually more rural, so `lit=no` is assumed?

(Personally I use `path` in urban areas where I'm not sure about allowed usage, so would be happy for it to be included in lit quest, but on the other hand it could produce a lot of spam of quest on rural paths which are all not lit.)

westnordost:
Good find, thank you!

Your assumption about why lit is not asked for paths is correct.

Issue #3061, opened 2021-07-13
Add "handlebar holder" to list of bicycle parking options


efmkoene:
**Use case**
In Switzerland, the "handlebar holder" version of bicycle parking is the most common (https://wiki.openstreetmap.org/wiki/Key:bicycle_parking), but I couldn't find it in StreetComplete (which is so much fun, by the way, thanks!). So I need to leave a comment on the map, each time.

**Proposed Solution**
Unless I'm mistaken and simply didn't see it correctly in the app, adding the "handlebar holder" as on the Wikipedia page to the options for bicycle parking should solve this.

matkoniecz:
https://taginfo.osm.ch/keys/bicycle_parking#values shows 144 handlebar holders across entire Switzerland right now, at 5.09% of tagged values, after `stands` `wall_loops` `shed` `rack`

https://taginfo.openstreetmap.org/keys/bicycle_parking#values shows 214 worldwide, at 0.08% share, at 17th position.

Main work here would be done by translators that could be confused by this quite unusual type.

> In Switzerland, the "handlebar holder" version of bicycle parking is the most common

Maybe it is the most common across not-yet-tagged ones? Or in your area?

efmkoene:
Interesting statistics, more than half of the world's mapped handlebar holders are in Switzerland, then. In my area (Zürich), I'd bet that handlebar holders definitely cover 50%+ of all bicycle parking options. But I remember that in the Netherlands or England, I only rarely saw this type... I leave it up to you to decide what to do, OK? And I'll understand it if it's closed or something, as it's not the most pressing issue on the map.

PR #3724, opened 2022-02-06
Add handlebar holder answer in bike parking quest

sams96:
This adds a new answer to the bicycle parking type quest, for "handlebar holder" type bicycle parking. These are quite common around me so this option will be useful at least to me. I just used the image used on the [osm wiki page](https://wiki.openstreetmap.org/wiki/Key:bicycle_parking), but I could go out and take another photo that might work better in the tiny box at some point.

image


matkoniecz:
This bicycle parking type is really really rare. But it is a distinct value that appears to be well defined and it bothered you enough to create PR. And it may help to avoid mistakes by other people. So I would be support merging it.

https://taginfo.openstreetmap.org/tags/bicycle_parking=handlebar_holder#overview

https://taginfo.openstreetmap.org/keys/bicycle_parking#values

eginhard:
In Switzerland it's **really** common (~70% of the total are mapped here) and I've also seen it elsewhere, so I think it's useful.

matkoniecz:
Was discussed in #3061

Went from 214 in July 2021 to 401 today. Still really low - but no longer so absurdly low - and shows that many were actually not mapped (and presumably many still remain unmapped).

sams96:
Hmm I didn't realise they were so uncommon, I guess they're mostly just a Swiss thing. Still it would be very useful for me mapping them around here.

westnordost:
The issue is - can we find a a clear line which bicycle parking types to include and which not? 

Why is such a line necessary? 

Not only to not overwhelm users with choices. Most of all because `bicycle_parking` is one of these values where everyone seems to just add a value to the documentation whenever one sees that it is visually different from what has already been documented. 

This is an issue for the same reason why adding documentation for e.g. [`highway=soi`](https://en.wikipedia.org/wiki/Soi) would be an issue: It implicitly redefines other values of `highway=*` because those that have been tagged as `highway=residential` (etc.) may actually rather be `soi`s. Yes, a new `highway` value is a crass example, I just want you to get the idea.

Hence, if we want any sort of stability in bicycle parking type tagging, we need to classify it by their function, not their visual appearance, i.e. have some (implicit) categorization like we also have with `building=*` (`yes` -> `residential` -> `house` -> `detached`). And so far, the convention has been that `stands` offer higher security because you can lock your frame to it and `wall_loops` have lower security because you can only lock your tyre to it. Most/all bicycle parkings can easily be sorted into these categories. The handlebar holder by that definition looks like it is a type of `stand` because you can lock your frame to it.

sams96:
> The issue is - can we find a a clear line which bicycle parking types to include and which not?
> 
> Why is such a line necessary?
> 
> Not only to not overwhelm users with choices. Most of all because `bicycle_parking` is one of these values where everyone seems to just add a value to the documentation whenever one sees that it is visually different from what has already been documented.
> 
> This is an issue for the same reason why adding documentation for e.g. [`highway=soi`](https://en.wikipedia.org/wiki/Soi) would be an issue: It implicitly redefines other values of `highway=*` because those that have been tagged as `highway=residential` (etc.) may actually rather be `soi`s. Yes, a new `highway` value is a crass example, I just want you to get the idea.
> 
> Hence, if we want any sort of stability in bicycle parking type tagging, we need to classify it by their function, not their visual appearance, i.e. have some (implicit) categorization like we also have with `building=*` (`yes` -> `residential` -> `house` -> `detached`). And so far, the convention has been that `stands` offer higher security because you can lock your frame to it and `wall_loops` have lower security because you can only lock your tyre to it. Most/all bicycle parkings can easily be sorted into these categories. The handlebar holder by that definition looks like it is a type of `stand` because you can lock your frame to it.

I would argue that there's a couple of reasons for these to be defined separately from stands & wall loops:
 * Stands support the frame, wall loops support the wheel (which is one of the issues with them, because your wheels get bent), these handlebar holders are clearly different in that respect. I can see this causing issues if you have an odd handlebar setup or, since here they are often mounted up against a wall, if you have a longer than usual front.
 * The method by which you actually lock your bike to it is via a cable, which can be of varying thickness or possibly not present at all (such as in the photo I used), so on average I would think the security for these is worse than a stand.

matkoniecz:
> The issue is - can we find a a clear line which bicycle parking types to include and which not?

When it is a separate type not redefining existing ones. For example it would exclude `bicycle_parking=wide_stands`

westnordost:
Alright, let's add it then

westnordost (approved):

sams96:
> Alright, let's add it then

Great thanks, is there anything else I need to do before it can be merged?

westnordost:
No. It contains new strings, this is why it will be going into the next major release for which there is no branch yet. I will merge it when there is.

PR #3731, opened 2022-02-07
Add a regex character class test

peternewman:
I haven't run the tests on this, I thought we had CI but I can't seem to see it/maybe it doesn't run on all PRs?


westnordost:
Looks good, I'll test it locally. I think CI currently doesn't work, see #3720

peternewman:
> Looks good, I'll test it locally.

Thanks, I was expecting the 45/135 ones to incorrectly pass, as I can't see a ^ at the start of the regex, but maybe that's just a Kotlin difference given your rresidential presumably correctly doesn't work.

> I think CI currently doesn't work, see #3720

Ah okay, and has presumably been disabled?

westnordost:
> ^ at the start of the regex

It's a match, not a find

> Ah okay, and has presumably been disabled?

No

peternewman:
> > ^ at the start of the regex
> 
> It's a match, not a find

And that's exactly the difference I was talking about; perl doesn't have that subtlety, you use `^match$` if you want that.

Issue #2714, opened 2021-04-02
Add surface to cycleways which are tagged as cycleway:right:surface or cycleway:left:surface


ohrie:
### General
Affected tag(s) to be modified/added: [cycleway:left:surface / cycleway:right:surface](https://wiki.openstreetmap.org/wiki/Key:cycleway:surface)
Question asked: **Which surface has the cycleway?**

Currently it's only possible to add the surface of a separate cycleway. Many cycleways are located on-street or some tracks are mapped not as extra way. So this would add the possibility to add the surface also on those types of cycleway.

If the cycleway has the same surface as street, this value should be adopted.

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt could be extended for handling also the right or left lane.

**Element selection:** 
```
way["cycleway:right"][!"cycleway:right:surface"]({{bbox}});
way["cycleway:left"][!"cycleway:left:surface"]({{bbox}});
```

**Proposed GUI:**
Analogous to the selection of cycleway type, first the correct side of the cycleway on the street should be select. Afterwards the existing surface dialog could be used.

matkoniecz:
In areas familiar to me vast majority cycleway lanes would have exactly the same surface as road, so I would ask for ones that are a separate way tagged on the road `cycleway:both=track` and similar.

DerDings:
Also think of sidewalk:(left|right|both):surface. It is also useful and probably only little extra work to implement it together with the cycle way surface quest requested here.

arrival-spring:
n.b. comment in https://github.com/streetcomplete/StreetComplete/pull/3735#pullrequestreview-891355704

> Let's add this quest but let's not add any further detailing quests for street-sides for now.

Issue #1978, opened 2020-07-21
Contributing quests needed for data related to wheelchair routing 


svstoll:
Hi,

For a university project, I'm looking for ways to improve OSM coverage regarding data that is needed for wheelchair routing algorithms such as the one used by [Openrouteservice](https://openrouteservice.org/). Given that the solution should be easy to use and does not require users to have previous OSM knowledge, I think that StreetComplete would be a great place to include such crowdsourcing tasks.

As of now, StreetComplete already has some quests that are relevant for wheelchair routing such as the surface type quests. On top of that, I would like to implement quests for the following things:

* Information about **curb types** on pedestrian crossings (see also #1305)
* **Incline** of ways that are relevant for wheelchair routing (using the smartphone's gyroscope to measure exact values)
* **Width** of sidewalks and other paths that are relevant for wheelchair routing (using Google's ARCore SDK for measuring as recently also suggested in #698)
* **Smoothness** of sidewalks and other paths that are relevant for wheelchair routing (see also #1630)
* Potential **barriers** on sidewalks and other paths that are relevant for wheelchair routing

While all of the involved OSM tags are already well-established, they are not used very often, yet. Hence, there may be a large amount of quests that will be generated until the coverage (hopefully) improves over time. Obviously, there are still other difficulties related to each of the proposed quest types, but I don't think that it makes sense to discuss them in great detail using this general issue.

I mainly opened this thread because I'm interested to know if you have some general feedback regarding the idea and to let you know that I would like to work on these things if you are willing to include such accessibility related quest types in StreetComplete (given proper testing etc.). If this is the case, I will create some more detailed quest proposals according to the contribution guidelines over the next couple of weeks (or continue the discussion on already existing ones).

westnordost:
Hi  @svstoll , that sounds interesting! People contributing (new quests) are always welcome. Thank you also to pitch it first here before opening PRs directly so I and other community members have a chance to comment on your ideas whether it is worth creating an issue/PR for that in the first place.

My general feedback is that of course, quests to improve wheelchair accessibility are very welcome in StreetComplete, because the data collected is inherently useful for at least one purpose (wheelchair accessibility).

> Information about curb types on pedestrian crossings (see also #1305)

Can't say much about it right now, not sure what the state of the ticket #1305 is. If I remember correctly there was some problem with the definition/standardization of kerbs (where they have to be tagged). But it could be that I remember wrong

[...]

PR #3735, opened 2022-02-07
Sidewalk surface quest

### LINKED ISSUE #1593: Ask about surface of sidewalks marked as tag on the road
### General
Affected tag(s) to be modified/added: sidewalk:*:surface tag family - see https://wiki.openstreetmap.org/wiki/Key:sidewalk

Question asked: **what is the surface of sidewalks?**

### Checklist
- [ ] 💤 Not an overwhelming percentage of elements have the same answer (No spam) - this may be the largest problem

### Ideas for implementation

**Element selection:** 

**Metadata needed:** 

**Proposed GUI:**

Sides interface, the same as for existence of sidewalks

arrival-spring:
Allow to add the surface of sidewalks tagged on the road, fixes #1593

Shows up for any suitable road with valid and complete sidewalk tagging. If there's no sidewalk or it's separately mapped for one side then that side is not selectable.

Images:

     

Todo/questions:

- [x] It needs a new icon (it's just using the footway part surface icon at the moment)
- [x] I don't like how the images are shown in the rotater view after having been selected as the photos don't tile well.
Two possible solutions:
* Find/create images for all of the surfaces which will tile (vector graphics)
* Use the icon overlay view used in the parking quest and just create circular images for each surface using the current images which would then essentially hover over that side.
* Something else?

- [x] Should it use the same surface form as the other surface quests? I feel like this would need some reworking of that though, and is then less consistent with other quests for things on different sides
- [x] Should there be an explicit option to say that sidewalk is wrong? At the moment the only thing the user can do is the standard leave a note option


westnordost:
I think if this is implemented, using this sidewalk puzzle as UI is fine. Yeah, I'd display the sidewalk drawable in the background and display the selected surface in a circle or something.

However, I am not entirely sure if this should be added at all. On slack US, I just wrote today with regards to if there will be a "measure (cycle track) width" quest for cycle tracks that are mapped on the roadway:

> @1ec5 Yeah, I could implement asking for cycle lanes and tracks tagged on the roadway-way, but this is more tricky because I need an interface for asking separately for the left and for the right side of the road. Also, the user needs to measure one side of the road, then cross the street and measure the other side of the road. Then he realizes that the width changes on one side but not on the other, so he has to split the way and and and... it's really a lot of work in the end for me (in implementaiton) and for the user... and kind of contributes to the fact that roads are split up into ever smaller segments to accomodate all the changing properties of the various things that are tagged on the roadway.
> This is why I rather not implement this for now but only for separately mapped cycleways.
> (For the same reason as stated above, STreetComplete also does not ask for other details on sidewalks/cycle paths tagged on the roadway-way, like surface or smoothness)
> If I implement this one day, then maybe for cycleway=lane only because cycle lanes cannot be mapped separately from the roadway-way.

So, in a nutshell, the issues:
- Laborous for the user
- Contributes to splitting up in even more and more fine-grained segments. At some point, this gets really difficult to manage/maintain. If such level of detail is wanted, I'd rather nudge users to map sidewalks/cycleways separately instead
- More complex UX. Your suggestion is the best so far, so maybe you solved this
- Lastly, the surface for sidewalks will most often be the same and also not matter much (between sett, asphalt, paving stones) as the surface is not used to drive or cycle upon, so the exact smoothness is rather unimportant. This can not be said in general for any `highway=path/footway`

matkoniecz:
If #1593 will be wontfixed then I am really really sorry that it was listed as an open  issue :(

westnordost:
Hmm, that was me who marked it as "new quest".

https://github.com/streetcomplete/StreetComplete/issues/1593#issuecomment-1029465655 indeed is evidence that my statement here that it will most often be the same is not correct (everywhere).

The other issues remain, especially the very same post https://github.com/streetcomplete/StreetComplete/issues/1593#issuecomment-1029465655 would indicate that tagging this everwhere to detail will lead to an almost unmaintainable situation. In certain places in the world, abutters mustto build and maintain 

Yet, the tag exists and is obviously used https://taginfo.openstreetmap.org/keys/sidewalk%3Aright%3Asurface

arrival-spring:
Thanks for the comments, here are my thoughts on some of it

> * Laborous for the user

Hence disabled by default (with the same message as smoothness - hard and time consuming)

I seem to remember @matkoniecz commenting somewhere recently that he'd look into how many users enabled disabled by default quests, is there any chance you've done that?


> * Contributes to splitting up in even more and more fine-grained segments. At some point, this gets really difficult to manage/maintain. If such level of detail is wanted, I'd rather nudge users to map sidewalks/cycleways separately instead

Valid concern. I wonder if the people who would enable a quest like this would be more likely to split up a way a lot (which can be tricky to do accurately in SC) or to just map the sidewalk separately in a desktop editor?

> * Lastly, the surface for sidewalks will most often be the same

Contradicts the second point somewhat about splitting up a way (wouldn't need to be done if the surface was all the same), but as you said in the later comment it seems to very much depend on area.

> and also not matter much (between sett, asphalt, paving stones) as the surface is not used to drive or cycle upon, so the exact smoothness is rather unimportant. This can not be said in general for any `highway=path/footway`

There was a bit of a discussion here: https://github.com/streetcomplete/StreetComplete/issues/1978#issuecomment-662507670 about this quest, related to wheelchair routing.

Personally in my area, most sidewalks (footpaths) are either asphalt or paving stones and this is consistent at least between junctions off that road and so it would be quite reasonable to answer in my area (especially with the last answer button) and ways would not need much splitting (or at least not in unhelpful ways).

If this is to be closed as wnf then #2714 should probably also be closed (or perhaps I change this to make it be for separated cycleways, e.g. `track`).

krisdoodle45:
Hey, I came to github just to suggest this feature, but to my surprise it is already being worked on. Haha

sidewalk_surface

I think something like this could look nice as a quest image (a bit more polished of course). What are your opinions?

arrival-spring:
> Oh no! You don't need a custom view for that nor an external dependency! Making a drawable circular is no black magic. Please check out `view/RotatedCircleDrawable`. You don't even need it rotated, but you can still just use that class.

Thanks! I did think I was probably missing something obvious

matkoniecz:
> Should there be an explicit option to say that sidewalk is wrong? At the moment the only thing the user can do is the standard leave a note option

I think that note is fine. Though after release I would encourage looking at https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#solving-notes and checking is there large volume of notes caused by such issue.

matkoniecz:
If it would do anything, then it could remove sidewalk tags. Though skipping it is fine and leaving regular way of leaving note is likely fine.

krisdoodle45:
Hey, here's a better looking image in SVG format as well:

sidewalk_surface

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,147 @@
...
+
+class AddSidewalkSurface : OsmFilterQuestType<SidewalkSurfaceAnswer>() {
+
+    override val elementFilter = """
+        ways with
+            highway ~ trunk|trunk_link|primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential
+            and area != yes
+            and motorroad != yes
+            and (
+                sidewalk = both or sidewalk = left or sidewalk = right or
```
westnordost, line 24:
```suggestion
                sidewalk ~ both|left|right or
```

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,147 @@
...
+                (sidewalk:left = yes and sidewalk:right ~ yes|no|separate) or
+                (sidewalk:right = yes and sidewalk:left ~ yes|no|separate)
```
westnordost, line 26: why the part behind the `and`? What about `sidewalk:both`?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,147 @@
...
+                or (
+                  sidewalk:left:surface ~ paved|unpaved|cobblestone
+                  and !sidewalk:left:surface:note
+                  and !note:sidewalk:left:surface
+                )
+                or (
+                  sidewalk:right:surface ~ paved|unpaved|cobblestone
+                  and !sidewalk:right:surface:note
+                  and !note:sidewalk:right:surface
+                )
```
westnordost, line 48: That kind of filter was added to the surface quest because at the time the quest was created, there were still (and are still) a lot of ways tagged with these generic/ambiguous values. This doesn't hold true for `sidewalk:*:surface` because it is barely tagged anyway so far - compared to `surface`. So this additional complexity is not worth it, IMO.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,147 @@
...
+            and (
+                sidewalk = both or sidewalk = left or sidewalk = right or
+                (sidewalk:left = yes and sidewalk:right ~ yes|no|separate) or
+                (sidewalk:right = yes and sidewalk:left ~ yes|no|separate)
+            )
+            and (
+                ${SIDEWALK_SURFACE_KEYS.joinToString(" and ") {"!$it"}}
+                or sidewalk:both:surface ~ ${ANYTHING_UNPAVED.joinToString("|")} and sidewalk:both:surface older today -4 years
+                or sidewalk:left:surface ~ ${ANYTHING_UNPAVED.joinToString("|")} and sidewalk:left:surface older today -4 years
+                or sidewalk:right:surface ~ ${ANYTHING_UNPAVED.joinToString("|")} and sidewalk:right:surface older today -4 years
```
westnordost, line 32: `sidewalk:right:surface older today -4 years`

looks for `check_date:sidewalk:right:surface`. The cycleway quest adds only `check_date:cycleway`, it would be consistent if this quest also did this unless there is a good reason not to.
We don't want this quest to add three different check_dates.

Also, given this filter is already so complex, maybe we could just remove these lines. Just check every 8 years always.

westnordost (commented):
Man, ... this is a lot of code, and complexity! Whew!

And that must all be maintained ... I have a bad feeling about if one dwelved even deeper here - e.g. cycle track obligatory or not, cycle track segregated from sidewalk or not, smoothness of sidewalk, smoothness of cycle track, ...

Let's add this quest but let's not add any further detailing quests for street-sides for now.

---

I didn't look through everything now, but here are already some comments:

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,144 @@
...
+            answer.left?.let { applySidewalkSurfaceAnswerTo(it, Side.LEFT, tags) }
+            answer.right?.let { applySidewalkSurfaceAnswerTo(it, Side.RIGHT, tags) }
+            deleteSidewalkSurfaceAnswerIfExists(Side.BOTH, tags)
+        }
+        deleteSidewalkSurfaceAnswerIfExists(null, tags)
+
+        // only set the check date if nothing was changed
+        if (!tags.hasChanges || tags.hasCheckDateForKey("sidewalk:surface")) {
+            tags.updateCheckDateForKey("sidewalk:surface")
+        }
```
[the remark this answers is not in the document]
westnordost: No, it should not be deleted. If it exists, it should be updated instead. Otherwise this would be destructive.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
```diff
@@ -0,0 +1,247 @@
...
+import de.westnordost.streetcomplete.quests.sidewalk.titleResId
+import de.westnordost.streetcomplete.util.normalizeDegrees
+import de.westnordost.streetcomplete.view.DrawableImage
+import de.westnordost.streetcomplete.view.ResImage
+import de.westnordost.streetcomplete.view.ResText
+import de.westnordost.streetcomplete.view.RotatedCircleDrawable
+import de.westnordost.streetcomplete.view.image_select.ImageListPickerDialog
+import kotlin.math.absoluteValue
+
+class AddSidewalkSurfaceForm : AbstractQuestFormAnswerFragment<SidewalkSurfaceAnswer>() {
```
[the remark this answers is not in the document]
westnordost: Hm, that would be good. I don't remember a/the reason why it wasn't used for the cycleway quest. Maybe there was one, or maybe I didn't have the time to refactor it to use that.

Anyway, following[ my own guidelines on PR reviews](https://github.com/streetcomplete/StreetComplete/discussions/3450), a refactor like that should probably go into a separate PR and not be suggested/demanded in this one.
So I'll resolve this conversation.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
westnordost: Right, thanks!

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,144 @@
...
+                  and !note:sidewalk:right:surface
+                )
+            )
+    """
+
+    override val changesetComment = "Add surface of sidewalks"
+    override val wikiLink = "Key:sidewalk"
+    override val icon = R.drawable.ic_quest_sidewalk_surface
+    override val isSplitWayEnabled = true
+    override val questTypeAchievements = listOf(PEDESTRIAN, WHEELCHAIR, OUTDOORS)
```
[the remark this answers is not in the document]
westnordost: Agree with @FloEdelmann

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
```diff
@@ -0,0 +1,245 @@
...
+                    .setNegativeButton(android.R.string.cancel, null)
+                    .show()
+            } else {
+                onSelectedSide(SurfaceAnswer(it.value!!), isRight)
+            }
+        }.show()
+    }
+
+    private fun onSelectedSide(surface: SurfaceAnswer, isRight: Boolean) {
+        val image = DrawableImage(RotatedCircleDrawable(resources.getDrawable(surface.value.asItem().drawableId!!)))
```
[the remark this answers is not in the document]
westnordost: No, it's fine like this

matkoniecz:
> This branch has conflicts that must be resolved

let me know if you have no idea how to solve THAT

arrival-spring:
> let me know if you have no idea how to solve THAT

Thanks for the offer, but I've managed it.

So, I think this is all sorted, it all runs as expected and the tests all pass.

I believe I've fixed everything from reviews (except for things noted as not needed in this PR).

app/src/main/java/de/westnordost/streetcomplete/osm/sidewalk/SidewalkParser.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    return null
+}
+
+private fun createSidewalksDefault(tags: Map<String, String>): LeftAndRightSidewalk? = when(tags["sidewalk"]) {
+    "left" -> LeftAndRightSidewalk(left = YES, right = NO)
+    "right" -> LeftAndRightSidewalk(left = NO, right = YES)
+    "both" -> LeftAndRightSidewalk(left = YES, right = YES)
+    "no" -> LeftAndRightSidewalk(left = NO, right = NO)
+    "none" -> LeftAndRightSidewalk(left = NO, right = NO)
```
westnordost, line 30:
```suggestion
    "no", "none" -> LeftAndRightSidewalk(left = NO, right = NO)
```
shortcut

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private fun sideSurfaceChanged(surface: SurfaceAnswer, side: Side, tags: Tags): Boolean {
+        val previousSideValue = tags["sidewalk:${side.value}:surface"]
+        val previousBothOsmValue = tags["sidewalk:both:surface"]
+        val osmValue = surface.value.osmValue
+
+        return if (previousSideValue != null && previousSideValue != osmValue) {
+            true
+        } else previousBothOsmValue != null && previousBothOsmValue != osmValue
```
westnordost, line 86:
```suggestion
        val previousSideOsmValue = tags["sidewalk:${side.value}:surface"]
        val previousBothOsmValue = tags["sidewalk:both:surface"]
        val osmValue = surface.value.osmValue

        return previousSideOsmValue != null && previousSideOsmValue != osmValue
            || previousBothOsmValue != null && previousBothOsmValue != osmValue
```
consistent naming + simplify return statement

westnordost (approved):

Helium314:
Maybe I'm missing something obvious, but I didn't find any explicit comment on this:
Why is the sidewalk surface only asked for specific roads, i.e. `highway ~ trunk|trunk_link|primary|primary_link|secondary|secondary_link|tertiary|tertiary_link|unclassified|residential`?

matkoniecz:
For which it is missing? It is likely done this way to avoid asking on say `highway=proposed` with `sidewalk` tag.

Helium314:
I don't actually miss the quest anywhere, I was just curious.
Excluding `highway=proposed` and `highway=construction` does make sense, thanks.

PR #3737, opened 2022-02-08
Make quest list generation script a Gradle task

FloEdelmann:
In #2523, I added the quest list generation script as a standalone Kotlin Script file (`*.main.kts`). But that is unnecessarily complex and inconsistent with other Gradle tasks. Also, it seems to cause issues for others trying to run this script (see #3720).

So in this PR, I converted the script to a Gradle task. The GitHub Actions workflow is also updated and should continue to work as expected. It now calls `gradlew` just like the "Build debug APK" workflow does.


westnordost:
This is great, I much appreciate this PR. Will review now.

buildSrc/build.gradle.kts
```diff
@@ -8,8 +8,10 @@ dependencies {
     implementation("de.westnordost:countryboundaries:1.5")
     implementation("com.esotericsoftware.yamlbeans:yamlbeans:1.15")
     implementation("org.jsoup:jsoup:1.14.2")
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.2")
```
westnordost, line 11: Hm now we have two JSON-parsing libraries here. Klaxon and Kotlinx-serialization. Oh well, I guess dependency frugality doesn't matter much for build scripts.

westnordost (approved):

PR #3739, opened 2022-02-09
Make the questName regex more robust

peternewman:
Improve on c1a7d1eea9daabed2185118636851b12e1bbc132

Untested, but based on some Kotlin I've been using in #2754

### DIFF
--- a/buildSrc/src/main/java/GenerateQuestListTask.kt
+++ b/buildSrc/src/main/java/GenerateQuestListTask.kt
@@ -37,9 +37,9 @@ open class GenerateQuestListTask : DefaultTask() {
     @TaskAction
     fun run() {
         val questFileContent = sourceDirectory.resolve("quests/QuestsModule.kt").readText()
-        val questNameRegex = Regex("(?<=^ {4})[A-Z][a-zA-Z]+(?=\\()", RegexOption.MULTILINE)
+        val questNameRegex = Regex("^\s+([A-Z][a-zA-Z]+)\\(", RegexOption.MULTILINE)
         val questNames =
-            listOf(noteQuestName) + questNameRegex.findAll(questFileContent).map { it.value }
+            listOf(noteQuestName) + questNameRegex.findAll(questFileContent).map { it.groupValues[1] }
 
         val questFiles = sourceDirectory.resolve("quests/").listFilesRecursively()
         val strings = getStrings(projectDirectory.resolve("app/src/main/res/values/strings.xml"))


buildSrc/src/main/java/GenerateQuestListTask.kt
```diff
@@ -37,9 +37,9 @@ open class GenerateQuestListTask : DefaultTask() {
     @TaskAction
     fun run() {
         val questFileContent = sourceDirectory.resolve("quests/QuestsModule.kt").readText()
-        val questNameRegex = Regex("(?<=^ {4})[A-Z][a-zA-Z]+(?=\\()", RegexOption.MULTILINE)
+        val questNameRegex = Regex("^\s+([A-Z][a-zA-Z]+)\\(", RegexOption.MULTILINE)
```
peternewman, line 40: This would also allow digits after the first alpha character which I believe is valid, even if we aren't currently using it.
```suggestion
        val questNameRegex = Regex("^\s+([A-Z][a-zA-Z0-9]+)\\(", RegexOption.MULTILINE)
```

FloEdelmann:
See also https://github.com/streetcomplete/StreetComplete/pull/2523#discussion_r575672287.

I'd rather have the regex more strict to avoid false-positives (and let it produce false-negatives instead).

peternewman:
> I'd rather have the regex more strict to avoid false-positives (and let it produce false-negatives instead).

Fair point, although I'd argue there's a difference between strictness and requiring an exact number of spaces beforehand (even if the linter is supposed to enforce that). It just feels like it's generating unnecessary maintenance work if we ever change our linter or style. This isn't Python after all! :laughing: 

The only close line is:
https://github.com/streetcomplete/StreetComplete/blob/1ecb8ca9d36cbd5132e3a2ad76c1b4b01d8065ba/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt#L146

But it doesn't begin with an upper-case character.

If we wanted to be strict, we should match the:
https://github.com/streetcomplete/StreetComplete/blob/1ecb8ca9d36cbd5132e3a2ad76c1b4b01d8065ba/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt#L155
and
https://github.com/streetcomplete/StreetComplete/blob/1ecb8ca9d36cbd5132e3a2ad76c1b4b01d8065ba/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt#L409

First, so we know we're only getting things in the QuestTypeRegistry when we then process it.

FloEdelmann:
> if we ever change our linter or style

Then updating one number will be our least concern :sweat_smile:

> If we wanted to be strict, we should match the […] and […] first, so we know we're only getting things in the QuestTypeRegistry when we then process it.

That'd be fine for me, but I don't think it's worth the effort.

westnordost:
So it looks like this should be closed?

FloEdelmann:
Yes, but @peternewman feel free to reopen/comment or open another PR if you disagree.

PR #3740, opened 2022-02-09
Provide hint for people confused how to mark lack of sidewalks

matkoniecz:
#2405 was the same for cycleways

screen03


westnordost:
Uhm, do we have to have this dialog for every single place, and with custom text, that uses this UI? I think this adds too much text.

matkoniecz:
Oh right, this one is generic and applicable in general and string name can be using some far more generic name.

app/src/main/res/values/strings.xml
```diff
@@ -1202,4 +1202,7 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="achievement_lifesaver_solved_X">You solved %d quests that help emergency services and first responders!</string>
     <string name="quest_airConditioning_no_name_title">"Does this place (%s) have air conditioning?"</string>
     <string name="quest_airConditioning_name_type_title">"Does %1$s (%2$s) have air conditioning?"</string>
+    <string name="quest_sidewalk_answer_none_title">Selecting that sidewalks are missing</string>
+    <string name="quest_sidewalk_answer_none_explanation">As with other quests presented in this way, tap on one side of the illustration with question marks, then select a proper answer.\nDo the same for the other side, if present.</string>
```
peternewman, line 1206: Suggested improvement:
```suggestion
    <string name="quest_sidewalk_answer_none_explanation">As with other quests presented in this way, tap on one side of the illustration with question marks, then select the correct answer.\nDo the same for the other side if present.</string>
```

Or:
```suggestion
    <string name="quest_sidewalk_answer_none_explanation">As with other quests presented in this way, tap on one side of the illustration with question marks, then select a suitable answer.\nDo the same for the other side if present.</string>
```

westnordost:
So, by me, a generic and short hint text like "Tap each side and select the situation for each side side individually" (+better wording) would be ok. This text should then be used everywhere where such an other answer option exists.

peternewman:
> "Tap each side and select the situation for each side side individually"
With one minor fix:
"Tap each side and select the situation for each side individually"

This reads okay to me.
Or:
"Tap each side and select the situation for that side individually"

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalkForm.kt
```diff
@@ -1,11 +1,30 @@
...
+import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.quests.AStreetSideSelectFragment
+import de.westnordost.streetcomplete.quests.AnswerItem
 
 class AddSidewalkForm : AStreetSideSelectFragment<Sidewalk, SidewalkSides>() {
 
     override val items = Sidewalk.values().map { it.asStreetSideItem() }
 
+    override val otherAnswers: List<AnswerItem> get() {
+        val isNoRoundabout = osmElement!!.tags["junction"] != "roundabout" && osmElement!!.tags["junction"] != "circular"
```
riQQ, line 13:
```suggestion
```
Unused copy-paste relict?

FloEdelmann: Or even better, replace the whole function with a static list:

```kt
        AnswerItem(R.string.quest_sidewalk_answer_none) { noSidewalksHereHint() },
    )
```

westnordost:
So what would be preferrable is to reuse that generic string for all quests where such a "usage hint" in this regard is displayed (e.g. the cycleway quest)

matkoniecz:
> So what would be preferrable is to reuse that generic string for all quests where such a "usage hint" in this regard is displayed (e.g. the cycleway quest)

I noticed that people are mostly confused (to the point of creating notes) by null results - no cycleways at all, no sidewalk at all. Presumably filling with answer that nothing is present is a bit confusing.

I plan to reuse this text if there will be similar confusing noticed for other quests - based on notes and looking how users use SC (I plan another session after processing already collected feedback).

But for now I think that it is ready.

Though maybe it can be added with some generic "help, how to use it" tutorial hint in can't say menu?

westnordost (approved):

Issue #3609, opened 2021-12-27
Use R-Tree for DB query performance improvements


smichel17:
Based on preliminary testing in https://github.com/streetcomplete/StreetComplete/discussions/3603#discussioncomment-1856544, this should make queries based on element location 2-5x faster, at the cost of around 15% more disk usage.

westnordost:
[R*-Tree support is documented here](https://www.sqlite.org/rtree.html). It is an optional module that needs to be added per compiler flag at compile time.

@Helium314 has been experimenting with this.

## Research required

### R*-Tree support in platform sqlite?

Does the sqlite implementation provided by the Android platform include the R*-Tree module? [According to Android documentation](https://developer.android.com/reference/android/database/sqlite/package-summary), the app can assume an SQLite version of at least 3.8. Optional R*-Tree support was added in 3.6, auxiliary columns in 3.24. So maybe it works but without auxiliary columns?

### R*-Tree support in official android sqlite?

I noticed that SQLite has official precompiled binaries for Android. Not distributed via maven-repository but just as a downloadable AAR: https://sqlite.org/download.html
Such an official source is of course preferrable to a third party library. Not sure if that library implements the interfaces from androidx.sqlite:sqlite but it doesn't really matter. Those are as official as it can get. However, I did not check whether the binaries have been compiled with the `-DSQLITE_ENABLE_RTREE=1` flag that enable the R-Tree feature.

## Implementation / Benchmarking

If the answer to none of the above is "yes", there is also https://github.com/requery/sqlite-android which has R-Tree support but/and also depends on androidx.sqlite.

Before a complete implementation is attempted, it is worthwile to compare the speed. Maybe easiest would be to keep the current index and do every query twice - one with the current index and one with the R*-Tree index, measure the time for both.

Tables that may profit from an R-Tree-index are the element geometry table, the OSM quests table and the notes quests table.

Helium314:
> However, I did not check whether the binaries have been compiled with the -DSQLITE_ENABLE_RTREE=1 flag that enable the R-Tree feature.

Yes, it's compiled with R-tree: https://www.sqlite.org/android/doc/trunk/www/install.wiki

> R*-Tree support in platform sqlite?

It does not, see flags used in current Android: https://android.googlesource.com/platform/external/sqlite/+/refs/heads/master/dist/Android.bp
(and if I create R-tree tables with the platform sqlite, SC crashes with a log message: "no such module: rtree")

[...]

PR #3741, opened 2022-02-09
Improve database performance

Helium314:
`ElementGeometryDao` is optimized in this PR, see https://github.com/streetcomplete/StreetComplete/issues/3609#issuecomment-1031177576 for performance comparison with current version.

The element geometry table is split into separate tables for ways and relations to avoid slow `queryIn` in `getAllEntries(keys)`.
Nodes do not have a separate geometry table, as nodes have a simple point geometry that can be created from the node position.

A spatial index is added to `NodeTable`, and some functions for querying nodes inside a bbox are added to `NodeDao` (equivalents to functions in `ElementGeometryDao`).

`MapDataController.getMapDataWithGeometry` makes use of the the new table layout by not querying node geometry and data separatly, instead point geometry is created from node positions.

Still a draft, because database upgrade is missing and tests have not been updated yet.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/geometry/ElementGeometryTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/geometry/ElementGeometryTable.kt
@@ -1,11 +1,11 @@
 package de.westnordost.streetcomplete.data.osm.geometry
 
 object ElementGeometryTable {
-    const val NAME = "elements_geometry"
+    const val NAME_RELATIONS = "elements_geometry_relations"
+    const val NAME_WAYS = "elements_geometry_ways"
 
     object Columns {
-        const val ELEMENT_ID = "element_id"
-        const val ELEMENT_TYPE = "element_type"
+        const val ID = "id"
         const val GEOMETRY_POLYGONS = "geometry_polygons"
         const val GEOMETRY_POLYLINES = "geometry_polylines"
         const val CENTER_LATITUDE = "latitude"
@@ -16,10 +16,9 @@ object ElementGeometryTable {
         const val MAX_LONGITUDE = "max_lon"
     }
 
-    const val CREATE = """
-        CREATE TABLE $NAME (
-            ${Columns.ELEMENT_TYPE} varchar(255) NOT NULL,
-            ${Columns.ELEMENT_ID} int NOT NULL,
+    const val CREATE_WAYS = """
+        CREATE TABLE $NAME_WAYS (
+            ${Columns.ID} int PRIMARY KEY,
             ${Columns.GEOMETRY_POLYLINES} blob,
             ${Columns.GEOMETRY_POLYGONS} blob,
             ${Columns.CENTER_LATITUDE} double NOT NULL,
@@ -27,16 +26,33 @@ object ElementGeometryTable {
             ${Columns.MIN_LATITUDE} double NOT NULL,
             ${Columns.MAX_LATITUDE} double NOT NULL,
             ${Columns.MIN_LONGITUDE} double NOT NULL,
-            ${Columns.MAX_LONGITUDE} double NOT NULL,
-            CONSTRAINT primary_key PRIMARY KEY (
-                ${Columns.ELEMENT_TYPE},
-                ${Columns.ELEMENT_ID}
-            )
+            ${Columns.MAX_LONGITUDE} double NOT NULL
+        );"""
+
+    const val CREATE_RELATIONS = """
+        CREATE TABLE $NAME_RELATIONS (
+            ${Columns.ID} int PRIMARY KEY,
+            ${Columns.GEOMETRY_POLYLINES} blob,
+            ${Columns.GEOMETRY_POLYGONS} blob,
+            ${Columns.CENTER_LATITUDE} double NOT NULL,
+            ${Columns.CENTER_LONGITUDE} double NOT NULL,
+            ${Columns.MIN_LATITUDE} double NOT NULL,
+            ${Columns.MAX_LATITUDE} double NOT NULL,
+            ${Columns.MIN_LONGITUDE} double NOT NULL,
+            ${Columns.MAX_LONGITUDE} double NOT NULL
+        );"""
+
+    const val SPATIAL_INDEX_CREATE_WAYS = """
+        CREATE INDEX elements_geometry_bounds_index_ways ON $NAME_WAYS (
+            ${Columns.MIN_LATITUDE},
+            ${Columns.MAX_LATITUDE},
+            ${Columns.MIN_LONGITUDE},
+            ${Columns.MAX_LONGITUDE}
         );
     """
 
-    const val SPATIAL_INDEX_CREATE = """
-        CREATE INDEX elements_geometry_bounds_index ON $NAME (
+    const val SPATIAL_INDEX_CREATE_RELATIONS = """
+        CREATE INDEX elements_geometry_bounds_index_relations ON $NAME_RELATIONS (
             ${Columns.MIN_LATITUDE},
             ${Columns.MAX_LATITUDE},
             ${Columns.MIN_LONGITUDE},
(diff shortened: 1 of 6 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/osm/geometry/ElementGeometryTable.kt
```diff
@@ -1,11 +1,11 @@
 package de.westnordost.streetcomplete.data.osm.geometry
 
 object ElementGeometryTable {
-    const val NAME = "elements_geometry"
+    const val NAME_RELATIONS = "elements_geometry_relations"
+    const val NAME_WAYS = "elements_geometry_ways"
```
westnordost, line 5: Please give if each table its own `*Table` object to be consistent with the rest of the code

Helium314: Is it ok now?

westnordost: Err no, I meant like all the other tables: Two files, one named WayGeometryTable.kt and the other RelationGeometryTable.kt.
Yes, the columns will be duplicted, but that's okay since they are both two completely independent tables. That the columns have the same names is basically only coincidental.

Helium314: Ok, so the different columns now made me add separate daos as well.

westnordost: Very good, I didn't notice that when scanning over it yesterday. It is good that the Daos are separate because this is consistent with the rest of the code.

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/ElementDao.kt
```diff
@@ -78,11 +78,11 @@ class ElementDao(
     }
 }
 
-private data class ElementIds(val nodes: List<Long>, val ways: List<Long>, val relations: List<Long>) {
+data class ElementIds(val nodes: List<Long>, val ways: List<Long>, val relations: List<Long>) {
     val size: Int get() = nodes.size + ways.size + relations.size
 }
 
-private fun Iterable<ElementKey>.toElementIds(): ElementIds {
+fun Iterable<ElementKey>.toElementIds(): ElementIds {
```
westnordost, line 85: These are not meant to be public. Do they need to be? I don't find them referenced anywhere

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/NodeDao.kt
```diff
@@ -79,4 +81,62 @@ class NodeDao(private val db: Database) {
...
+                ElementPointGeometry(
+                    LatLon(cursor.getDouble(LATITUDE), cursor.getDouble(LONGITUDE))
+                )
+            )
+        }
+    }
+
+    fun getAll(bbox: BoundingBox): List<Node> {
+        return db.query(NAME, where = inBoundsSql(bbox)) { cursor ->
+            Node(
```
westnordost, line 125: duplication of creating Node from cursor, same with ElementGeometryEntry

westnordost (approved):
Looks good! I am going to attempt some more optimizations on top of this branch so you can leave the processing of my comments to me

Helium314:
So is there anything left for me, or are you doing the rest?

westnordost:
https://github.com/Helium314/StreetComplete/pull/334

Helium314:
I remember it was faster, but returned fewer elements: https://github.com/streetcomplete/StreetComplete/issues/2803#issuecomment-830445155

Is this still the case? Then I'll try make the testing a bit more exhaustive and compare the number of returned elements as well.

westnordost:
Yes, it is deliberate that it returns less elements. This is how the OSM API works. It returns all elements who have at least one node inside the given bbox. Current StreetComplete behavior is that it returns all elements whose bbox intersect with the bbox given.

E.g. imagine a bicycle route relation that goes from the town center out to the suburbs in the North-West. The bbox of this relation would be quite large, so the route relation would be included in every bbox-query between the center and anywhere in the north or the west or northwest. When querying the OSM API, the route relation would only be included if a node of a way to which it refers is included in the bbox.

westnordost:
Thank you for the link to #2803. I've read it again and see there is the issue with https://github.com/streetcomplete/StreetComplete/issues/2803#issuecomment-827639718

But on the other hand, we already add a padding of 20m to each side of the bbox plus it is only used for fuzzy filters to not ask for certain elements. Fuzzy filters are fuzzy anyways, it's OK if the user instead answers e.g. "cycleway displayed separately on map" because one might argue that this info was missing anyway.

smichel17:
It seems like the main performance problem comes from fetching *relations* that intersect the bbox. However, the use case in https://github.com/streetcomplete/StreetComplete/issues/2803#issuecomment-827639718 seems like it applies primarily to *ways*.

Would it make sense to use the current SC approach (show if intersecting) for ways and the osmapi approach (show only if a node is contained) for relations? Or is that not possible for technical reasons?

westnordost:
Technical reasons that speak against this is that it is more complex (more code, possibly hacky-looking code) and looks inconsistent, plus is inconsistent with the OSM API. For me, this is reason enough to not consider it.

smichel17:
In my opinion, performance optimizations are possibly the best justification for hacky code.

I imagine an interface like

```kt
fun fetchByBbox(bbox: Bbox, includeIntersectingWays: Boolean = false, includeInteresctingRelations: Boolean = false)
```

Under the hood, the initial, cleaner implementation would perform a fetch-by-bbox query that works like the osm api, then separate call(s) to find intersecting ways or relations. I suspect doing separate calls would be a performance hit, although I'm not sure by how much (maybe still better than always looking for relations). Then we could decide whether combining the queries would be worth the messier code.

But anyway, I do not have a very strong opinion. Just wanted to share the idea. If your opinion remains the same, no need to respond.

Helium314:
action / function | this PR (s) | fetch-by-bbox (s)
--- | --- | ---
`ElementDao.putAll`  | 204 | 204
`ElementGeometryDao.putAll`  | 16 | 12
query bench run | 530 | 477
`MapDataController.getMapDataWithGeometry` | 129 | 81
`ElementGeometryDao.getAllEntries(keys)` | 9 | 9
`ElementGeometryDao.getAllKeys` | 4 | does not exist
`NodeDao.get` | 9 | 8
`NodeDao.getAll` | 83 | 83
`WayDao.get` | 122 | 122
`OsmQuestDao.getAllInBBox` | 14 | 14
`OsmQuestDao.getAllForElements` | 108 | 107

By a very short look, `getMapDataWithGeometry` on `this PR` usually returns 10-20 elements more than on `fetch-by-bbox`. This appears to be more accurate than some %, as it holds for 371 vs 359 and for 25 vs 11 elements

westnordost:
Thank you! Nice, so it is not slower, probably actually a little faster (if only because it returns less elements, who knows).

So, can you merge it into this branch and I merge this branch into master?

Also, compared to current situation, what would be the speed gain, overall in %? Would be nice to have a % for the changelog

Helium314:
Comparing to the current state:
* database is ca 30% smaller
* persising elements is ca 30% faster (whole download + persisting + quest creation is ca 20% faster)
* DB access in "average use" (as in the logged queries for the bench) is ca 130% faster

westnordost:
130%?! Wow, I completely underestimated the difference this makes.... how? `ElementGeometryDao` had an index on type+id didn't it?

Helium314:
My test showed that
https://github.com/streetcomplete/StreetComplete/blob/ed59421ed07e572d41fd8c68197ccb4afc946ffc/app/src/main/java/de/westnordost/streetcomplete/data/osm/geometry/ElementGeometryDao.kt#L86-L90
was quite slow (500 ms+).
When I replaced the `queryIn` by querying a single column with my hacky single index, `ElementGeometryDao.getAllEntries(keys)` was a lot faster. That's what started trying the optimizations on the normal database, because in that query there is no R-tree index involved.

westnordost:
Aaahh, right. This is because it is actually not really supported by SQLite, queryIn is a workaround:

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/data/DatabaseXt.kt#L28

Issue #3736, opened 2022-02-07
Image misplaced on compacted option in smoothness quest


arrival-spring:
See image:
<img src="https://user-images.githubusercontent.com/77166354/152874896-be0e313e-d22b-4add-8314-e09f86022a7b.png" width="500">

The "very bumpy and uneven" option has the image in a weird place for me. I guess something to do with the line wrapping playing up. 

What other information is needed to fix this?

peternewman:
> What other information is needed to fix this?

Screen resolution and SC/Android versions probably.

Which quest exactly or what way?

I don't have those exact options so I guess this is 40.0 release?

matkoniecz:
Screen resolution would be useful and maybe sufficient to reproduce it.

Phone model, Android version, SC version are in general useful.

westnordost:
The icon is inserted as an ImageSpan. I think it must be a bug in Android handling ImageSpans in multiline text

arrival-spring:
This is 40.0 beta 1
Motorola moto one
Screen resolution 1520 by 720
`surface=compacted` (on both road and path versions of the quest)

westnordost:
@FloEdelmann any idea? Otherwise I will close this as will not fix. In my eyes this is simply a bug in the text view that cannot handle ImageSpans that would wrap to the next line.

FloEdelmann:
I'll have a look later.

FloEdelmann:
It looks fine in the emulator with Android 11:

<img src="https://user-images.githubusercontent.com/202916/153280290-a0872655-1af3-4aca-88d2-d15cbd833216.png" width="300">

What is you Android version?

Maybe it's an issue for older versions of Android (before 10) only:
https://github.com/streetcomplete/StreetComplete/blob/57034acb370370982ccc80a645bd47f5e0b8be60/app/src/main/java/de/westnordost/streetcomplete/ktx/Drawable.kt#L28

FloEdelmann:
Anyway, I think there's nothing we can do on our side, unfortunately :slightly_frowning_face:

westnordost:
Thanks for investigating!

arrival-spring:
> What is your Android version?

Android 10

Thanks for having a look

mnalis:
I can confirm this looks broken on my Android 10 (Huawei P30 Pro, EMUI 11.0.0) too - and not only `Very bumpy and uneven` but also `almost impassable`:

[...]

PR #3743, opened 2022-02-09
Fix vehicle icon display for surface quality quest on Android 10 phones

mnalis:
Fixes the display issues on my Huawei P30 Pro (Android 10).
(Also does not seem to break my Android 6 Xiaomi Note 3 Pro, which did not have that bug)
 
Closes: #3736


matkoniecz:
Tested by author, see https://github.com/streetcomplete/StreetComplete/issues/3736#issuecomment-1034322884

PR #3744, opened 2022-02-10
Fixed date in changelog

JohnVeness:
I'm assuming the "21" in "Feb 21" was referring to a year, and that you put the wrong year. If actually it was referring to the 21st day of February, please ignore!


westnordost:
Yes, February the 21st of 2022 :-)

westnordost:
I mean, February is already now, but the new version is still in development

PR #3747, opened 2022-02-11
deleting offices with shop replacement enabled

matkoniecz:
I noticed many notes caused by SC mappers unable to delete objects like offices
this patch allows to fallback from shop replacement to deleting objects rather immediately to note creation

------------------

Additionally, maybe it would be a good idea to allow straight delete of shop objects - not only tagging as vacant.

--------------------

As pull request as it is noticeable change to design,and  I am not entirely happy with code but I see no good replacement (maybe trigger deletion from shop replacement function? But it seems to go against its name...)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestAnswerFragment.kt
@@ -239,11 +239,14 @@ abstract class AbstractQuestAnswerFragment<T> :
                 && osmElement?.type == ElementType.NODE
         val isReplaceShopEnabled = (questType as? OsmElementQuestType)?.isReplaceShopEnabled == true
         if (!isDeletePoiEnabled && !isReplaceShopEnabled) return null
-        check(!(isDeletePoiEnabled && isReplaceShopEnabled)) {
-            "Only isDeleteElementEnabled OR isReplaceShopEnabled may be true at the same time"
-        }
 
         return AnswerItem(R.string.quest_generic_answer_does_not_exist) {
+            val element = osmElement
+            if(element != null) {
+                if (element.isSomeKindOfShop()) {
+                    if (isReplaceShopEnabled) replaceShopElement()
+                }
+            }
             if (isDeletePoiEnabled) deletePoiNode()
             else if (isReplaceShopEnabled) replaceShopElement()
         }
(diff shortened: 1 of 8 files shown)


westnordost:
I am not happy with the code either (there is a bug there anyway), plus, it subverts how the `replaceShopElement` function is desigend.
If it should be changed to immediately delete the element, it should go directly in `replaceShopElement` line 337. However, I think it is not okay if the app just deletes the app only because the app does not recognize the current tagging as "some kind of shop". Leaving a note is the safer option. So, I'll close this PR.

Regarding adding the possibility to delete the node from the `ShopGoneDialog` / `ShopTypeForm` is something that could be considered. Though, wording will be difficult, as "its not a shop anymore" would be misleading. "Has been converted to office or private flat" would be possible, but this 
1. eliminates the possibility that the user can specify what kind of office (e.g. https://www.openstreetmap.org/note/3040665#map=19/53.57725/9.95406&layers=N ) and
2. may not cover all reasons why the `shop=vacant` node should be deleted.

All in all, I am for leaving everything as is in this regard.

matkoniecz:
Right now for some objects asked by opening hours quest it is impossible to delete them by SC, you just get redirected to the note creation menu if not recognized as a shop.

I think it would be desirable to allow deletion in such cases (as usual, for nodes).

westnordost:
This should only work for an inclusive list though, i.e. a list of places where we can be sure that we are not destroying data.

mnalis:
If I understand this correctly,  I'd recommend re-tagging the object as vacant or `disused:*` if it is gone, instead of deleting the object. 
Because, if node itself is deleted, when something else pops up there it cannot be answered / fixed again by SC, but requires full-fledged editor to add new object there (and additionally history is lost).  

And in most of the cases, the **_place is still there_**, it is just that is **_not being used_** at that moment. Deletion should be considered an option only if that was for example kiosk that was **_razed to the ground_,** but not if the place has just become empty/unused (and even then, deletion is not _"must"_ but _"may"_/_"should"_).

westnordost:
Or should be ob replaceShopElement 

On February 13, 2022 1:06:06 AM GMT+01:00, Mateusz Konieczny ***@***.***> wrote:
>>  I'd recommend re-tagging the object as vacant or disused:* if it is
>gone, instead of deleting the object.
>
>That is a good idea if there is an empty local or some disused
>remnants. But sometimes it is fully gone (say single shop replaced two
>previous shops - one node can be retagged but one should be deleted) 
>
>> This should only work for an inclusive list though, i.e. a list of
>places where we can be sure that we are not destroying data.
>
>OK, I will make such list. Do you think that it is fine to trigger
>deletion from within `replaceShopElement()`? Or would it be better to
>do outside it, before it gets called, like in this PR?
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/3747#issuecomment-1037596223
>You are receiving this because you modified the open/close state.
>
>Message ID:
***@***.***>

-- 

matkoniecz:
@westnordost 

> Or should be ob replaceShopElement

Sorry, I am confused and not sure how to interpret it.

matkoniecz:
And it is likely irrelevant as during testing I found several ways how this can damage data due to poor tagging schemes (like putting building address and `office` POI on the same node)

PR #3750, opened 2022-02-11
Use more LocationManagerCompat methods.

Isira-Seneviratne:


westnordost:
Hat is the advantage of this?

Isira-Seneviratne:
> Hat is the advantage of this?

`LocationRequestCompat` provides options for customising the location request, e.g. the request quality, duration, etc.

westnordost:
Ok, then I'll close this. We are not using any of these features. And I refactored the class not long ago, I don't remember clearly but maybe I deliberately decided against using LocationManagerCompat in that function because there was some problem with it. If there is no gain, I don't see the reason to invest time into re-testing the location stuff and possibly find and fix possible upcoming problems with the new implementation.
There had been some issues after the class was last refactored.

PR #3751, opened 2022-02-11
Remove ViewCompat.isLaidOut().

Isira-Seneviratne:


westnordost:
Best if you would add a short comment like "unnecessary on API >= 19" or something like that so that I don't need to check in the code what this is about. But thanks anyway!

PR #3757, opened 2022-02-13
Ask oh, name, wheelchair, fee data on additional POIs

matkoniecz:
(based mostly on failed attempt to improve deletion/note handling - I have looked only on some tags from https://github.com/streetcomplete/StreetComplete/pull/3747#issuecomment-1038127669 )

PR, as similar change related to battery recycling turned out to be problematic

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
@@ -11,7 +11,8 @@ class AddGeneralFee : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
         nodes, ways, relations with
-         (tourism = museum or leisure = beach_resort or tourism = gallery)
+         (tourism = museum or leisure = beach_resort or tourism = gallery
+         or office=coworking or amenity=coworking_space)
          and access !~ private|no
          and !fee
          and name
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
@@ -35,7 +35,7 @@ class CheckOpeningHoursSigned (
           and access !~ private|no
           and (
             name or brand or noname = yes or name:signed = no
-            or amenity ~ recycling|toilets|bicycle_rental or leisure=park or barrier
+            or amenity ~ recycling|toilets|bicycle_rental|charging_station|social_facility or leisure=park or barrier
           )
     """.toElementFilterExpression() }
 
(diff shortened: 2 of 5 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
```diff
@@ -35,7 +35,7 @@ class CheckOpeningHoursSigned (
           and access !~ private|no
           and (
             name or brand or noname = yes or name:signed = no
-            or amenity ~ recycling|toilets|bicycle_rental or leisure=park or barrier
+            or amenity ~ recycling|toilets|bicycle_rental|charging_station|social_facility or leisure=park or barrier
```
westnordost, line 38: Should ask for names of certain social facilities (food_bank|clothing_bank|soup_kitchen|dairy_kitchen) too and remove it from here

app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
```diff
@@ -11,7 +11,8 @@ class AddGeneralFee : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
         nodes, ways, relations with
-         (tourism = museum or leisure = beach_resort or tourism = gallery)
+         (tourism = museum or leisure = beach_resort or tourism = gallery
+         or office=coworking or amenity=coworking_space)
```
westnordost, line 15: Formatting. Please always add spaces at the operators and use the parenthesis and indentation like elsewhere in the filters.

On a separate note, I think a coworking space has no place here. All coworking spaces I know are member-only. I.e. comparable to a fitness studio - you pay a monthly fee and then may use it. Maybe there are exceptions. But maybe there are also exceptiosn for fitness studios and fitness studios are also not in this list because most, as written, are membership-based. (Hence, impossible to answer for surveyor)

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
matkoniecz: Hmm, in Poland there are several that are existing to make nearby objects more interesting - for example in malls or operated by a public entity such as library (though I would need to check has it survived COVID)

OSM Wiki mentions that they are typically paid, but still recommended fee tag so I thought that such free ones are occuring also elsewhere ( https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcoworking_space#Additional_tags )

westnordost: Some libraries have rooms where you can work/read quietly, but those are something else than coworking spaces.

matkoniecz: This specific one was different, but seems relatively unique. Removed coworking spaces here, modified wiki pages to make clear that vast majority is paid.

westnordost (approved):

PR #3758, opened 2022-02-13
Add seating quest

### LINKED ISSUE #1176: outdoor_seating
### General
Question asked: **Does this location have outdoor seating?**

https://wiki.openstreetmap.org/wiki/Key:outdoor_seating

### Checklist
It is established, and the usefulness is pretty obvious - it is nice out, and i want to find an eatery (generally) that has outdoor seating. 

fatal69100:
Add a seating quest, fixes #1176.

Screenshot from 2022-02-13 22-21-26


matkoniecz:
I have not checked other things at all, but note that translations are not done in PRs.

See https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app

matkoniecz:
> Matching other icon theme (I don't know how to do a nice circle icon with orange background)

I am not an icon making expert but I always took an existing SVG icon file and modified it, rather creating it from scratch. See https://github.com/streetcomplete/StreetComplete/tree/master/res/graphics/quest

fatal69100:
Thank you, I hope it is prettier

westnordost:
You are working on Qwant maps, very cool! Keep up the good work, I really like it! Did you consider adding reviews to it? I think there have been a few attempts of creating a distributed and/or open source review system anyone can use (I listened to the talks at [SotM2019](https://2019.stateofthemap.org/sessions/LBGPCD/) and at [SotM2020](https://2020.stateofthemap.org/sessions/AA8RXP/))

---

Anyway, regarding your PR: In the original ticket #1176 I mentioned this concern: 

> Though, how would on be able to determine this during times where the outdoor seating is not available? (Winter)

At times where there is no outdoor seating due to the weather or season, the quest is unanswerable and thus breaks the [quest guidelines](https://github.com/streetcomplete/StreetComplete/wiki/Adding-new-Quests-to-StreetComplete). The only way to solve this is to either only show this quest in places where it is always warm (tropics, I guess), or to implement a dynamic quest filter that disables the quest based on the season. (Or a combination of both.)

I can tell you that the lattter is quite complex. Not only complex as in that it is difficult to pinpoint "when is summer?", probably even on a per-country level, but also on the technical side to add such a dynamic filter that automatically disables a quest. It has been done before for asking whether a way is lit (#2872), but it hadn't been received well by some users, so it was reverted.

peternewman:
How about disabled by default with a note about only doing it in the summer. Or even something like the not being near the quest nag screen, with the option to cancel per session, if you answer no? Given the risk is only answering no when the stuff has been tucked away indoors for the winter.

fatal69100:
Unfortunately, we don't schedule to add open source review system in the short term. The priority is currently adding partner data that help use on the business side and improving search results. But I will definitely push open source review system to the roadmap in the next month, which is a great idea :)

About the PR, as a new contributor, I think it will become too complex for me. I was wondering if users were complaining about mapping at night for the way lit quest, I believe they will have less complaining about a quest available during summer because mapping during summer is more fun ;)

matkoniecz:
if variant

> disabled by default with a note about only doing it in the summer

is acceptable, then it should be relatively easy, see https://github.com/streetcomplete/StreetComplete/blob/7b4e2b592c573a5dcaabe16c64eb7e01eaa218be/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt#L58 for how this can be achieved (just define one more property and one more string resource)

westnordost:
See @matkoniecz last comment. This solution would be acceptable.

app/src/main/res/values/strings.xml
```diff
@@ -1227,4 +1227,10 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_road_width_title">"What’s the width of this road here?"</string>
     <string name="quest_road_width_explanation">"The roadway width from curb to curb: This includes anything on the street surface like on-street parking or cycle lanes but excludes anything beyond it like cycle paths or parking beside the roadway."</string>
 
+    <string name="quest_seating_name_title">Does this place have seating?</string>
+    <string name="quest_seating_takeaway">Only takeaway</string>
+    <string name="quest_seating_indoor_only">Only indoor seating available</string>
+    <string name="quest_seating_outdoor_only">Only outdoor seating available</string>
+    <string name="quest_seating_indoor_and_outdoor">Indoor and outdoor seating available</string>
+    <string name="default_disabled_msg_summer_outdoor_seating">"This quest type is disabled by default because this quest can only be resolved in summer season in order to check outdoor seating "</string>
```
FloEdelmann, line 1235:
```suggestion
    <string name="default_disabled_msg_summer_outdoor_seating">"This quest type is disabled by default because it can only be solved in summer season in order to check outdoor seating."</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
+            nodes, ways with
+            amenity = bar
+            or amenity = pub
+            or amenity = food_court
+            or amenity = ice_cream
+            or amenity = restaurant
+            or amenity = cafe
+            or amenity = fast_food
```
FloEdelmann, line 35: Why not use the same filter as above?

```suggestion
            nodes, ways with amenity ~ restaurant|cafe|fast_food|ice_cream|food_court|pub|bar
```

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
```diff
@@ -0,0 +1,61 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.osmquests.Tags
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
+
+class AddSeating : OsmFilterQuestType<Seating>() {
+    override val elementFilter = """
+        nodes, ways with amenity ~ restaurant|cafe|fast_food|ice_cream|food_court|pub|bar
+        and !outdoor_seating
```
FloEdelmann, line 14: I'd exclude places where `takeaway=only` is tagged, because seating is implicitly already set then.

Should it also exclude places where `indoor_seating` is already tagged?

```suggestion
        and takeaway != only
        and !outdoor_seating and !indoor_seating
```

Or ask for all places where either one is missing?

```suggestion
        and takeaway != only
        and (!outdoor_seating or !indoor_seating)
```

app/src/main/res/values-en/strings.xml
```diff
@@ -1202,4 +1202,5 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="achievement_lifesaver_solved_X">You solved %d quests that help emergency services and first responders!</string>
     <string name="quest_airConditioning_no_name_title">"Does this place (%s) have air conditioning?"</string>
     <string name="quest_airConditioning_name_type_title">"Does %1$s (%2$s) have air conditioning?"</string>
+    <string name="quest_seating_name_title">Does this place have seating?</string>
```
FloEdelmann, line 1205: Please revert this change, `values-en/strings.xml` will automatically be overwritten with the contents of `values/strings.xml`.

app/src/main/res/values/strings.xml
```diff
@@ -1227,4 +1227,10 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_road_width_title">"What’s the width of this road here?"</string>
     <string name="quest_road_width_explanation">"The roadway width from curb to curb: This includes anything on the street surface like on-street parking or cycle lanes but excludes anything beyond it like cycle paths or parking beside the roadway."</string>
 
+    <string name="quest_seating_name_title">Does this place have seating?</string>
+    <string name="quest_seating_takeaway">Only takeaway</string>
+    <string name="quest_seating_indoor_only">Only indoor seating available</string>
+    <string name="quest_seating_outdoor_only">Only outdoor seating available</string>
+    <string name="quest_seating_indoor_and_outdoor">Indoor and outdoor seating available</string>
```
FloEdelmann, line 1234: Maybe like this to make it more explicit and easier to parse?

```suggestion
    <string name="quest_seating_takeaway">No seating, only takeaway</string>
    <string name="quest_seating_indoor_only">Yes, only indoors</string>
    <string name="quest_seating_outdoor_only">Yes, only outdoors</string>
    <string name="quest_seating_indoor_and_outdoor">Yes, indoors and outdoors</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/seating/Seating.kt
```diff
@@ -0,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.seating
+
+enum class Seating {
+    NO,
+    ONLY_INDOOR,
+    ONLY_OUTDOOR,
+    INDOOR_AND_OUTDOOR,
+}
```
FloEdelmann, line 8: You can give some extra attributes to enum members, which can simplify the logic in `applyAnswerTo` (see my next comment):

```suggestion
enum class Seating(val hasOutdoorSeating: Boolean, val hasIndoorSeating: Boolean) {
    NO(false, false),
    ONLY_INDOOR(false, true),
    ONLY_OUTDOOR(true, false),
    INDOOR_AND_OUTDOOR(true, true),
}
```

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
```diff
@@ -0,0 +1,61 @@
...
+            }
+            Seating.ONLY_INDOOR -> {
+                tags["outdoor_seating"] = "no"
+                tags["indoor_seating"] = "yes"
+            }
+            Seating.INDOOR_AND_OUTDOOR -> {
+                tags["outdoor_seating"] = "yes"
+                tags["indoor_seating"] = "yes"
+            }
+        }
```
FloEdelmann, line 59: Using the `enum class` extension suggestion from my previous comment, this can be simplified to:

```suggestion
        if (answer == Seating.NO) {
            tags["takeaway"] = "only"
        }
        
        tags["outdoor_seating"] = answer.hasOutdoorSeating.toYesNo()
        tags["indoor_seating"] = answerhasIndoorSeating.toYesNo()
```

(You might need to import the `toYesNo` extension function though.)

FloEdelmann (commented):
Overall this looks good. I have just a few suggestions. Feel free to apply or reject them. From my side, these are not merge-blocking.

andrewharvey:
I find these values clearer and easier to quickly scan as the key distinguishing information is in the first word:

- Takeaway only
- Indoor seating only
- Outdoor seating only
- Both indoor and outdoor seating

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
andrewharvey: Yeah I'd exclude asking for takeaway only, unless you decide the response was too old and it should be re-checked.

I'd then only exclude where both outdoor_seating and indoor_seating are tagged, so your second option, because this quest is getting more complete information about both and if only one is tagged then the untagged one is undetermined.

fatal69100:
Thank you very much for your helpful feedback

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
```diff
@@ -0,0 +1,41 @@
...
+    override val wikiLink = "Key:outdoor_seating"
+    override val icon = R.drawable.ic_seating_quest
+    override val isReplaceShopEnabled = true
+
+    override val questTypeAchievements = listOf(QuestTypeAchievement.CITIZEN)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_seating_name_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter(elementFilter)
```
FloEdelmann, line 29: That's not what I meant. Right now, this will only highlight elements for which this quest is also applicable.

I meant it like this:

```suggestion
        getMapData().filter("""
            nodes, ways with amenity ~ restaurant|cafe|fast_food|ice_cream|food_court|pub|bar
        """)
```

I.e. still a separate filter string, but the same concise syntax from the `elementFilter` above.

FloEdelmann (changes requested):

FloEdelmann (approved):

FloEdelmann:
Thanks for implementing the feedback!

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
```diff
@@ -0,0 +1,43 @@
...
+    override val wikiLink = "Key:outdoor_seating"
+    override val icon = R.drawable.ic_seating_quest
+    override val isReplaceShopEnabled = true
+
+    override val questTypeAchievements = listOf(QuestTypeAchievement.CITIZEN)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_seating_name_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
```
matkoniecz, line 29: I think that are shoplike objects should be displayed here (see how it looks in other shop-related quests) to give a better orientation info.

app/src/main/res/values/strings.xml
```diff
@@ -1229,4 +1229,10 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_road_width_title">"What’s the width of this road here?"</string>
     <string name="quest_road_width_explanation">"The roadway width from curb to curb: This includes anything on the street surface like on-street parking or cycle lanes but excludes anything beyond it like cycle paths or parking beside the roadway."</string>
 
+    <string name="quest_seating_name_title">Does this place have seating?</string>
+    <string name="quest_seating_takeaway">Takeaway only</string>
+    <string name="quest_seating_indoor_only">Indoor seating only</string>
+    <string name="quest_seating_outdoor_only">Outdoor seating only</string>
+    <string name="quest_seating_indoor_and_outdoor">Both indoor and outdoor seating</string>
+    <string name="default_disabled_msg_summer_outdoor_seating">"This quest type is disabled by default because it can only be solved in summer season in order to check outdoor seating."</string>
```
matkoniecz, line 1237:
```suggestion
    <string name="default_disabled_msg_summer_outdoor_seating">"This quest type is disabled by default because it may require checking during specific part of the year in order to check outdoor seating."</string>
```

Not entirely sure how to phrase it better to be clear also in places where outdoor seating is present during other part of the year or all the time.

app/src/main/res/values/strings.xml
```diff
@@ -1229,4 +1229,10 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_road_width_title">"What’s the width of this road here?"</string>
     <string name="quest_road_width_explanation">"The roadway width from curb to curb: This includes anything on the street surface like on-street parking or cycle lanes but excludes anything beyond it like cycle paths or parking beside the roadway."</string>
 
+    <string name="quest_seating_name_title">Does this place have seating?</string>
```
andrewharvey, line 1232:
```suggestion
    <string name="quest_seating_name_title">What kind of seating does this place have?</string>
```

Minor nitpick, my suggestion given the answers are the values like takeaway, indoor, outdoor rather than yes/no values.

app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
fatal69100: I am not sure to understand, I don't think a lot of shops in the list : https://wiki.openstreetmap.org/wiki/Key:shop have seating available. IMO, It may add too many quests

andrewharvey: Yeah I don't think we should ask on shop=* as they are mostly retail stores to purchase goods or services. I can't even think of a single one where we should ask.

westnordost (changes requested):
Did you draw the quest icon yourself? I am kind of unhappy with it, the black line drawing looks quite bland. Would you mind if I replaced that icon? Maybe I'd just use a coffee cup and a sun or something. Or just a sun. Or a wasp on a piece of cake ;-)

westnordost:
Did this myself now and also replaced the icon

fatal69100:
Thank you very much. Sorry for late reply.

PR #3760, opened 2022-02-13
Run `ktlint` on any push

adrianclay:
At the moment it is easy to make some changes and only get feedback that there are linting issues after that change reaching master.

By running ktlint on any push contributors will get feedback sooner. This should hopefully result in fewer linting issues reaching master.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,8 +1,5 @@
 name: "Lint"
-on:
-  push:
-    branches:
-      - "master"
+on: push
 
 jobs:
   ktlint:


westnordost:
I'd really not like ktlint to build a regime here. I.e. consistent code formatting is nice but if it becomes a (perceived) barrier in contributing, then it must go. Because really, whether it's `when(` or `when (` is seriously not going to be disturb anyones flow of reading and understanding the code.

Thus, that ktlint does not nitpick so often / so early is deliberate.

Incidentally I activated the ktlint plugin for Android Studio as it is now recommended in a readme somewhere but I experience so far has been that ktlint to be a bit annoying. So I am thinking about deactivating it again.
- For example while one writes codes, Android Studio will automatically indent the next line when one presses enter. ktlint will immediately mark this as an "error" and this is really distracting. Sometimes I go like "huh, what is wrong here now? Some wrong variable name? Did Android Studio break again and do I need to press the 'sync from gradle' button again? Ah no, it's just ktlint 🙄"
- The other issue I got with ktlint is that it marks everything as "error". So if anything ktlint seems to inherently train me to ignore errors while in the process of writing the code. And this is not a good thing

FloEdelmann:
> By running ktlint on any push contributors will get feedback sooner.

No, `push` would just work for all branches of this repository; so it would only show in PRs from project members. It would work with the `pull_request` event though: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

I initially had it like that ([`207b5e0` (#3649)](https://github.com/streetcomplete/StreetComplete/pull/3649/commits/207b5e0d3d6ef9221f940f03e6a0ff85c3720e20)), and then even a bit fancier after @matkoniecz's feedback (https://github.com/streetcomplete/StreetComplete/pull/3649#issuecomment-1015557563), but @westnordost didn't like it (https://github.com/streetcomplete/StreetComplete/pull/3649#issuecomment-1018701994).

FloEdelmann:
> For example while one writes code, Android Studio will automatically indent the next line when one presses enter. ktlint will immediately mark this as an "error" and this is really distracting.

I agree with that. "Lint on save" instead of "Lint while typing" would help here. I suggested that in a similar issue on the plugin's bug tracker: https://github.com/nbadal/ktlint-intellij-plugin/issues/125#issuecomment-1040246051

> The other issue I got with ktlint is that it marks everything as "error".

You can configure that:

grafik

FloEdelmann:
> it also has a built-in formatter […], so could […] we […] automatically commit and fix and push them?

Yeah that could work. @westnordost would that be okay for you?

westnordost:
Thanks for the hint to the configurations! Is there any possibility to add this as default per .editorconfig or similar?

> Yeah that could work. @westnordost would that be okay for you?

No. A human must have the final say over how things are formatted, ktlint may only advise.

westnordost:
Anyway, I'll close this PR.

westnordost:
Perfect, thanks!

PR #3761, opened 2022-02-14
Use LocationListenerCompat.

Isira-Seneviratne:
Use the [`LocationListenerCompat`](https://developer.android.com/reference/androidx/core/location/LocationListenerCompat) interface.


Isira-Seneviratne:
> Okay, I understand the other methods in the LocationListener interface have only be declared with `default` empty implementations in some later Android version? And because [LocationListenerCompat](https://developer.android.com/reference/androidx/core/location/LocationListenerCompat) does this always, there is only one remaining method not implemented which is why the kotlin SAM constructor works?

Yeah.

westnordost:
I find this a bit difficult to understand, but it looks like this is how `LocationListenerCompat` is intended, so that's fine then.

PR #3762, opened 2022-02-14
Tactile paving on steps quest

### LINKED ISSUE #3534: [Quest] Do these steps have tactile paving at the top and bottom?
Apologies for mistakes, this is my first attempt at this:

### General
Affected tag(s) to be modified/added: [tactile_paving=] (https://wiki.openstreetmap.org/wiki/Key:tactile_paving)
Question asked: Do these steps have tactile paving at the top and bottom?
Answers same as crossing Yes / No / Can't Say

### Checklist
- [ ] 🚧 To be added tag is established and has a useful purpose - highway=steps has 16 913 instances of tactile_paving
- [ ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one) - yes
- [ ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary - yes
- [ ] 💤 Not an overwhelming percentage of quests have the same answer (No spam) - Most are no, but more and more are yes as areas are redeveloped
- [ ] 🕓 Applies to a reasonable number of map data (Worth the effort) - believe so. I am in the UK. Not sure about other countries?

### Ideas for implementation
Copy and paste the crossing quest :-)

highway=steps
tactile_paving is unset

**Metadata needed:**  No meta data needed, but probably enabled/disabled on country by country basis

<!-- If you have any idea on how the UI (the form appearing when the quest is asked) should look like.
I am willing to try and get some photos, if needed, but the same images for Crossing can be used.

arrival-spring:
Fixes #3534 

My first attempt at an icon (combining tactile paving and steps icons)

Despite discussion in above linked issue, convention does appear to be to tag it on the way tagged with `highway=steps`

It was added to the [steps page of the wiki in 2009](https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dsteps&diff=246497&oldid=237564) and to [the tactile paving page in 2010](https://wiki.openstreetmap.org/w/index.php?title=Key:tactile_paving&diff=401293&oldid=246498) and marked as being specifically for ways [later that year](https://wiki.openstreetmap.org/w/index.php?title=Key:tactile_paving&diff=535396&oldid=514981).

There's been no discussion on the wiki talk pages about this and it's been tagged on over 18,000 steps.

In addition, it's not a primary feature and there would likely be no other tags on the nodes at either end of the steps.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -120,6 +120,7 @@ import de.westnordost.streetcomplete.quests.surface.AddRoadSurface
 import de.westnordost.streetcomplete.quests.tactile_paving.AddTactilePavingBusStop
 import de.westnordost.streetcomplete.quests.tactile_paving.AddTactilePavingCrosswalk
 import de.westnordost.streetcomplete.quests.tactile_paving.AddTactilePavingKerb
+import de.westnordost.streetcomplete.quests.tactile_paving.AddTactilePavingSteps
 import de.westnordost.streetcomplete.quests.toilet_availability.AddToiletAvailability
 import de.westnordost.streetcomplete.quests.toilets_fee.AddToiletsFee
 import de.westnordost.streetcomplete.quests.tourism_information.AddInformationToTourism
@@ -221,6 +222,7 @@ fun questTypeRegistry(
     AddHandrail(), // for accessibility of pedestrian routing, can be gathered when walking past
     AddStepsRamp(),
     AddStepsIncline(), // can be gathered while walking perpendicular to the way e.g. the other side of the road or when running/cycling past, confuses some people, so not as high as it theoretically should be
+    AddTactilePavingSteps(), // need to check top and bottom
 
     AddReligionToPlaceOfWorship(), // icons on maps are different - OSM Carto, mapy.cz, OsmAnd, Sputnik etc
     AddReligionToWaysideShrine(),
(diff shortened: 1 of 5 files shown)


matkoniecz:
It affects also crossing ques, but is `no` really correct when only one side has paving?

arrival-spring:
Probably `incorrect` is the right thing there the wiki has "or if only one sloped kerb at a crossing has tactile paving" as an example of `incorrect`.

westnordost:
I think this is wrong. The `tactile_paving` should be one the first and on the last node of `highway=steps`, no? Just like it is for kerbs.

arrival-spring:
> I think this is wrong. The `tactile_paving` should be one the first and on the last node of `highway=steps`, no? Just like it is for kerbs.

Whilst that would make sense I can't see that it's documented anywhere and has barely been used at all like that.

Comparing `highway=steps` and `tactile_paving=*` : https://overpass-turbo.eu/s/1g8b (388 ways in central England)
With nodes which are members of `highway=steps` where node is tagged `tactile_paving=*` : https://overpass-turbo.eu/s/1g8c (23 nodes in the same area)

And most of the latter are in fact places where steps end at a crossing and `tactile_paving` refers to the crossing.

Granted, more detailed mapping would separate the end of the steps from the road, but there are plenty of places where this is not the case.

In that same view and excluding crossings : https://overpass-turbo.eu/s/1g8d I find 4 nodes

Therefore doing the quest like that would both lead to issues with where steps meet a road (or such places would have to be excluded) and is also inventing a new form of tagging.

westnordost:
`tactile_paving` makes the whole thing more complex (to record, to interpret) while less helpful:
- There'd need to be a tag for "on one side only" (as you mentioned)
- What to do if the `highway=steps` way is split up into e.g. 3 segments for whatever reason? Are they then all `incorrect` because there is only tactile paving at the very beginning and the very end of the steps as a whole?
- What about a footway that joins in at the middle of a `highway=steps`? If the `tactile_paving=yes` is on the steps-way, does this mean there is also tactile paving at the point where the footway joins the steps? In any case, it would be useful to record this info, no?

I reckon `tactile_paving` is not really used on steps yet and/or people haven't really been thinking about this use case, so let's delay this ticket after a discussion about this has been held in the tagging mailing list, the wiki or some other linkable medium where people (interested in tactile paving tagging) might tarry.

PR #2596, opened 2021-02-20
Canary Islands has very different climates, biomes and culture

erik55:
Canary Islands has very different climates and biomes compared to
mainland Spain: The vulcanic land surface ranges from sealevel to 3718
meters above sealevel, and is located 1000 km south of mainland Spain,
very close to Sahara, therefore they have (also) different
produces.
And Canary Islands have also quite a different culture, therefore very
common regional sports pitches on all seven islands.

Therefore I added some country_metadata for the country subdivision ES-CN, because the most important (and most common) produces and sports were missing.

PR #3764, opened 2022-02-14
Add Codeowners file, so I get notified about orchard YAML changes

rugk:
It would be great if I get a notification for (manual) updates of the orchard data, as in https://github.com/streetcomplete/StreetComplete/pull/2596, so I can adapt my script or at least take care of that and suggest things/review the PR.

This file could also be used to include or even require other reviewers for specific quests or quest data or whatnot, too, BTW.

Docs: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+res/country_metadata/orchardProduces.yml  @rugk


westnordost:
Ah cool, makes sense

FloEdelmann:
@rugk Did it work with https://github.com/streetcomplete/StreetComplete/commit/d84a077db8129c0db8a2c5cf449111fb2783ecf7?

Because GitHub complains about the file:
Screenshot_2022-03-14_12-41-38

rugk:
Ah it only works when you actually have write access, so well… okay, feel free to remove me there.

PR #3769, opened 2022-02-16
Improve some donation wording

peternewman:
I assume this is the correct branch to target?

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -866,9 +866,9 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="about_description_translate">%1$s is translated %2$d%%</string>
     <string name="about_title_sponsors">Sponsors</string>
     <string name="about_summary_donate">Show your appreciation! ❤️</string>
-    <string name="about_description_donate">Thank you for considering to support this app! To see current donations and to donate yourself, tap on the respective platform:</string>
-    <string name="about_description_donate_google_play">"Thank you for considering to support this app!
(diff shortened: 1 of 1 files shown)


FloEdelmann (changes requested):

FloEdelmann (approved):

westnordost:
Thanks, I will pick some of the changes made here (i.e. not reword the existing string)

app/src/main/res/values/strings.xml
```diff
@@ -866,9 +866,9 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="about_description_translate">%1$s is translated %2$d%%</string>
     <string name="about_title_sponsors">Sponsors</string>
     <string name="about_summary_donate">Show your appreciation! ❤️</string>
-    <string name="about_description_donate">Thank you for considering to support this app! To see current donations and to donate yourself, tap on the respective platform:</string>
-    <string name="about_description_donate_google_play">"Thank you for considering to support this app!
-Due to a Google Play Store policy, the donation links may not be displayed in-app, but you will find any information regarding this project, including how to sponsor this app’s development, in its home:"</string>
+    <string name="about_description_donate">Thank you for considering supporting this app! To see current donations and to donate yourself, tap on the respective platform:</string>
+    <string name="about_description_donate_google_play">"Thank you for considering supporting this app!
```
peternewman, line 870: Any reason not to fix at least this line in the Google version too? https://github.com/streetcomplete/StreetComplete/commit/8b8c300fd70ca512222d19dd3f64f84e3c960739

I'm also a bit confused why you don't want to fix the base one? I appreciate it will prompt translators to retranslate, but it's likely most of them have fixed in their language already, so this only impacts the English versions.

PR #3770, opened 2022-02-16
Add GB and sort the list of subdivisions

peternewman:
I assume there's no reason you don't support e.g. `gb-eng`?:
https://nsi.guide/index.html?t=brands&k=amenity&v=atm&cc=gb-eng


westnordost:
There is a reason. GB subdivisions are not supported by the default data set used by countryboundaries.

PR #3771, opened 2022-02-16
Better explanation

matkoniecz:
spotted by @mnalis in https://poeditor.com/projects/po_edit?id=97843&term=358528113&id_language=3#showComments358528113

> this is confusing; "tap each side" means that you must do it for both sides, and next sentence then says "do the same for other side"? What additional "other side", as I just did EACH (i.e. both) of the sides?

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1209,7 +1209,7 @@ Alternatively, you can leave a note (with a photo)."</string>
 
     <string name="quest_sidewalk_answer_none">No sidewalk at all</string>
     <string name="quest_sidewalk_answer_none_title">Selecting that sidewalks are missing</string>
-    <string name="quest_side_select_interface_explanation">Tap each side and select the situation for that side.\nDo the same for the other side if present.</string>
+    <string name="quest_side_select_interface_explanation">Tap one side and select the matching answer.\nDo the same for the other side if present.</string>
 
     <string name="no_camera_permission_warning_title">Camera permission</string>
     <string name="no_camera_permission_warning">The permission is necessary to be able to measure distances with the camera.</string>


matkoniecz:
opening PR to ensure that this version is final and fine

matkoniecz:
Pinging @peternewman - if you have moment a review would be highly appreciated.

peternewman (approved):
LGTM, but what do you do if there is no other side? I've not looked at the GUI this maps to...

matkoniecz:
That is why I put "if present"

peternewman:
I mean presumably you can get one-way roads here with only one side.

If this is also used for the sidewalk quest, then the "present" bit may be ambiguous, presumably you need to select both sides there, but tag one as no sidewalk, not only select the side(s) where sidewalk is present?

app/src/main/res/values/strings.xml
```diff
@@ -1209,7 +1209,7 @@ Alternatively, you can leave a note (with a photo)."</string>
 
     <string name="quest_sidewalk_answer_none">No sidewalk at all</string>
     <string name="quest_sidewalk_answer_none_title">Selecting that sidewalks are missing</string>
-    <string name="quest_side_select_interface_explanation">Tap each side and select the situation for that side.\nDo the same for the other side if present.</string>
+    <string name="quest_side_select_interface_explanation">Tap one side and select the matching answer.\nDo the same for the other side if present.</string>
```
peternewman, line 1212: Some random alternatives:
```suggestion
    <string name="quest_side_select_interface_explanation">Tap each side shown and select the matching answer.</string>
```
```suggestion
    <string name="quest_side_select_interface_explanation">Tap each side shown and select the appropriate answer.</string>
```
```suggestion
    <string name="quest_side_select_interface_explanation">Tap each side shown and select the most appropriate answer.</string>
```
```suggestion
    <string name="quest_side_select_interface_explanation">Tap one side and select the matching answer.\nDo the same for the other side if shown on screen.</string>
```

Issue #2816, opened 2021-04-27
Surveillance under Surveillance


paradx:
The project Surveillance under Surveillance https://sunders.uber.space/ uses the open street map object 
man_made surveillance to tag positions of cameras and other surveillance equipment on open street map. 

One of the initial goals of the project was, to track whether temporary video surveillance during G20 congress in hamburg, was removed after the conference is over. 

Since some of the camera data is not up to date a challenge to tag existing cameras and update the existing cameras with further tags (fixed camera, dome or panning) 

**Use case**
I think many of the tagged cameras are either not existing anymore or are not completely tagged. 

**Proposed Solution**
Add the option to add more information to existing cameras, verify existing data.

matkoniecz:
See #87

westnordost:
Could this be that this ticket is about *re-checking* if a surveillance camera is still there?

If yes, @paradx , what would be a reasonable filter for that? Only outdoor ones maybe?

paradx:
Re-checking would be a very good start. 
 
I think out door ones are the most popular surveillance object in general and it makes sense to filter for outdoor cameras only. 

In some countries it also is a legal difference if you film public space or inside your own home (germany for example)

westnordost:
ok let's start with that. What would be the filter to include that?

Helium314:
`surveillance:type = camera and surveillance ~ public|outdoor`?

kjonosm:
I quite often see `man_made=surveillance` being wrongly added as a property of a feature, like `shop=*` or `amenity=*` instead of it being a single object representing the actual surveillance camera. See e.g. [this](https://overpass-turbo.eu/s/16SM) overpass query.

Is it possible to extract `man_made=surveillance` together with its subtags from the main feature and create a new object within this check?

The newly created `man_made=surveillance` object could then be placed at the same location if the feature is tagged on a node or at the centre of an area feature.

westnordost:
No.

But this is a good hint, we'd want to exclude places like this from the query. How best to do that?

matkoniecz:
> How best to do that?

Mass exclude `tourism=`, `shop`, `craft=` etc like for `barrier=yes` quest?

chrt223:
> Re-checking would be a very good start.

@westnordost : The following characteristics for this quests would be useful:
- **set camera type** (dome / fixed / panning / ALPR ...)
- **set surveillance kind** (public / outdoor / ...)
- **set direction**
- possibly the height or the mount

[...]

PR #3777, opened 2022-02-18
Add two achievement links, add debug link activity

FloEdelmann:
This PR adds two links that can be unlocked with the "citizen" achievement:

* [OsmAnd](https://osmand.net/) (similar to Organic Maps) – I was surprised that this wasn't included or at least suggested before.
* [Surveillance under Surveillance](https://sunders.uber.space/) (see also #2816)

I also added a debug "Show Achievement Links" activity in the settings that shows all links, without having to login to an OSM account and actually reach all achievement levels:


app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -178,6 +178,14 @@ private val links = listOf(
...
         R.string.link_organic_maps_description
     ),
+    Link(
+        "osmand",
+        "https://osmand.net/",
+        "OsmAnd",
+        LinkCategory.MAPS,
+        R.drawable.ic_link_osmand,
+        R.string.link_osmand_description
+    ),
```
westnordost, line 188: Is OsmAnd really completely open source? https://play.google.com/store/apps/details?id=net.osmand.plus costs something.

matkoniecz: Yes, that is why F-droid has version with all functionality available for free (though if Osmand operators would want this they could limit server access and require people to host own map servers).

westnordost: ah

westnordost (approved):

Issue #539, opened 2017-08-25
New quest: Smoking?


rugk:
For cafes/restaurants/etc the [smoking tag](https://wiki.openstreetmap.org/wiki/Key:smoking) is useful.

Of course, this needs some country-specific research as in Germany, e.g. restaurants are not allowed to have smokers and non-smokers rooms together, but they *must be* split. (So smoking=yes is not really possible.)

Binnette:
In france smoking is not allowed in public areas since 2008. But some restaurant/bar have an 'isolated' room in wich smoking is allowed.

westnordost:
In Germany there are places which explicitly allow smoking. (And if it is not advertised, it usually means smoking=no)

alejandroscf:
In Spain is forbidden to smoke in all restaurants, bars and pubs, also in public/government buildings.

westnordost:
Since when? When I lived there (2011), smoking was forbidden "by default" and every place that allowed smoking had to explicitly have some kind of note/certificate/sign (I don't remember) posted somewhere visible.

majkaz:
The same Czech republic - no smoking in restaurants and pubs at all (new in 2017).

pietervdvn:
In Belgium, smoking is forbidden as well. It is only allowed in a separated room in some restaurants.

rory:
Smoking has been banned in all enclosed workplaces in Ireland since ~2003. So this would always be `smoking=no` in Ireland.

voschix:
In Italy, all indoors public places are no.smoking by law. Very few places have dedicated smoker rooms.

hmartink:
I would support adding such a quest. I'm not sure general rules (like "smoking=yes" not possible in Germany) are sensible/possible in SC. If it is possible, then of course this quest might be superfluous in some areas (if really all options, like smoking in front of the building) are prohibited.

And if a mapper does not like being asked (e.g. if the law does not allow several options), then she could simply use the settings to avoid being asked this quest.
As the comments show, there are many general rules which usually allow some kind of configurability (e.g. separate rooms, smoking outside). Beeing a non-smoker, I would highly appreciate obtaining the information beforehand. But if I were a smoke, this information would be also useful (e.g. if in Italy or Belgium I would look for those rare places with dedicated smoker rooms). 

[...]

PR #3648, opened 2022-01-17
documentation for making a SC quest

matkoniecz:
Step-by-step instruction how a new quest type can be created.

I think that such documentation would be useful for a potential contributors. It is possible to figure its on your own (that is how I started!), but I think that such description has chance to help some people.

I promise that I will maintain it for foreseeable future (I included some "open a new issue if documentation is unclear" there)

Review is really welcomed. Also, about places that can be cut, simplified or shortened. It is more likely to be too long than too short.

Also, it likely has 181818 grammar issues and many inaccuracies.

PR #3784, opened 2022-02-19
Smoking quest

mnalis:
Closes: #539

First draft of my first quest, I'd appreciate comments.

Also, few questions:

- if `getTitle()` OK? I've just copied it, but as I understand it checks for `name` or `brand` tags, and if they exist it uses `R.string.quest_smoking_name_type_title`. But what I'm unsure about is am I guaranteed that `arrayOfNotNull(name, featureName.value)` will always do right thing? Will `featureName.value` always be present, so at least `R.string.quest_smoking_no_name_title` can substitute one string?
- is it possible in `other answers` menu to force positioning? (i.e. I'd like `quest_smoking_isolated` to be show at the top)
- what is the best way to hide some answers in some cases (e.g. I'd like to hide `outside only` answer if it is for `leisure=outdoor_seating`) - separate `elementFilter` and rest of the file, or is there a better way?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -50,7 +50,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     override fun createForm() = SmokingAllowedAnswerForm()
 
     override fun applyAnswerTo(answer: SmokingAllowed, tags: Tags, timestampEdited: Long) {
-	tags.updateWithCheckDate("smoking", answer.osmValue)
+        tags.updateWithCheckDate("smoking", answer.osmValue)
     }
 
     private fun hasProperName(tags: Map<String, String>): Boolean =
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -16,9 +16,11 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     private val elementFilterBasic = """
         nodes, ways with
         (
-          amenity ~ bar|cafe|pub|biergarten|restaurant|nightclub|stripclub
+          amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
           or leisure ~ outdoor_seating
+          or amenity ~ fast_food|ice_cream and (outdoor_seating != no or indoor_seating != no)
         )
+        and takeaway != only
     """
 
     override val elementFilter = elementFilterBasic + """
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -33,7 +33,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     override val questTypeAchievements = listOf(CITIZEN)
 
     override fun getTitle(tags: Map<String, String>) =
-        if (hasFeatureName(tags))
+        if (hasProperName(tags))
             R.string.quest_smoking_name_type_title
         else
             R.string.quest_smoking_no_name_title
@@ -52,6 +52,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
 	tags.updateWithCheckDate("smoking", answer.osmValue)
     }
 
-    private fun hasFeatureName(tags: Map<String, String>): Boolean =
-        tags.containsKey("name") || tags.containsKey("brand") || tags.containsKey("operator")
+    private fun hasProperName(tags: Map<String, String>): Boolean =
+        tags.keys.containsAny(listOf("name", "brand", "operator"))
+
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -18,7 +18,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
         (
           amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
           or leisure ~ outdoor_seating
-          or amenity ~ fast_food|ice_cream and (outdoor_seating != no or indoor_seating != no)
+          or (amenity ~ fast_food|ice_cream or shop ~ ice_cream|deli|bakery|coffee|tea|wine) and (outdoor_seating != no or indoor_seating != no)
         )
         and takeaway != only
     """
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -59,7 +59,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
             (
                 ${elementFilterBasicFragment()} or
                 ${elementFilterBasicFragment("disused")} or
-                ${IS_SHOP_OR_DISUSED_SHOP_EXPRESSION} or
+                ${IS_SHOP_OR_DISUSED_SHOP_EXPRESSION}
             )
         """)
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -13,18 +13,25 @@ import de.westnordost.streetcomplete.ktx.containsAny
 
 class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
 
-    private val elementFilterBasic = """
-        nodes, ways with
-        (
-          amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
-          or leisure ~ outdoor_seating
-          or (amenity ~ fast_food|ice_cream or shop ~ ice_cream|deli|bakery|coffee|tea|wine) and (outdoor_seating != no or indoor_seating != no)
-        )
-        and takeaway != only
-    """
+    private fun elementFilterBasicFragment(prefix: String? = null): String {
+        val p = if (prefix != null) "$prefix:" else ""
+        return """
+              {p}amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
+              or {p}leisure ~ outdoor_seating
+              or ({p}amenity ~ fast_food|ice_cream or {p}shop ~ ice_cream|deli|bakery|coffee|tea|wine)
+        """.trimIndent()
+    }
 
-    override val elementFilter = elementFilterBasic + """
-        and (!smoking or smoking older today -8 years)
+    /* note: outdoor_seating/indoor_seating extra clause ONLY applies to last group in
+       elementFilterBasicFragment(), and not to whole of it */
+
+    override val elementFilter = """
+            nodes, ways, relations with
+            (
+                ${elementFilterBasicFragment()} and (outdoor_seating != no or indoor_seating != no)
+            )
+            and takeaway != only
+            and (!smoking or smoking older today -8 years)
     """
 
     override val changesetComment = "Add smoking status"
@@ -47,7 +54,13 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     }
 
     override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
-        getMapData().filter(elementFilterBasic)
+        getMapData().filter("""
+            nodes, ways, relations with
+            (
+                ${elementFilterBasicFragment()} or
+                ${elementFilterBasicFragment("disused")}
+            )
+        """)
 
     override fun createForm() = SmokingAllowedAnswerForm()
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -53,6 +53,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
             R.string.quest_smoking_no_name_title
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        Log.d("SmokingQuest debug", "elementFilter is: #${elementFilter}")
         val name = tags["name"] ?: tags["brand"] ?: tags["operator"]
         return arrayOfNotNull(name, featureName.value)
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -1,14 +1,14 @@
 package de.westnordost.streetcomplete.quests.smoking
 
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.osm.osmquests.Tags
 import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.CITIZEN
 import de.westnordost.streetcomplete.ktx.arrayOfNotNull
 import de.westnordost.streetcomplete.ktx.toYesNo
-import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
 
-class AddSmoking : OsmFilterQuestType<Boolean>() {
+class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
 
     override val elementFilter = """
         nodes, ways, relations with
@@ -16,7 +16,7 @@ class AddSmoking : OsmFilterQuestType<Boolean>() {
           amenity ~ bar|cafe|pub|biergarten|restaurant|nightclub|stripclub
           or leisure ~ outdoor_seating
         )
-        and !smoking
+	and (!smoking or smoking older today -8 years)
     """
     override val changesetComment = "Add smoking status"
     override val wikiLink = "Key:smoking"
@@ -37,10 +37,10 @@ class AddSmoking : OsmFilterQuestType<Boolean>() {
         return arrayOfNotNull(name, featureName.value)
     }
 
-    override fun createForm() = YesNoQuestAnswerFragment()
+    override fun createForm() = SmokingAllowedAnswerForm()
 
-    override fun applyAnswerTo(answer: Boolean, tags: Tags, timestampEdited: Long) {
-        tags["smoking"] = answer.toYesNo()
+    override fun applyAnswerTo(answer: SmokingAllowed, tags: Tags, timestampEdited: Long) {
+	tags.updateWithCheckDate("smoking", answer.osmValue)
     }
 
     private fun hasFeatureName(tags: Map<String, String>): Boolean =
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -19,7 +19,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
         return """
               ${p}amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
               or ${p}leisure ~ outdoor_seating
-              or (${p}amenity ~ fast_food|ice_cream or ${p}shop ~ ice_cream|deli|bakery|coffee|tea|wine)
+              or ((${p}amenity ~ fast_food|ice_cream) or (${p}shop ~ ice_cream|deli|bakery|coffee|tea|wine))
         """.trimIndent()
     }
 
@@ -28,7 +28,11 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     override val elementFilter = """
             nodes, ways, relations with
             (
-                ${elementFilterBasicFragment()} and (outdoor_seating and outdoor_seating != no or indoor_seating and indoor_seating != no)
+                ${elementFilterBasicFragment()}
+                and (
+                    (outdoor_seating and (outdoor_seating != no)) or
+                    (indoor_seating and (indoor_seating != no))
+                )
             )
             and takeaway != only
             and (!smoking or smoking older today -8 years)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -29,11 +29,13 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     override val elementFilter = """
             nodes, ways, relations with
             (
-                ${elementFilterBasicFragment()}
+                amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
+                or leisure ~ outdoor_seating
+                or (((amenity ~ fast_food|ice_cream) or (shop ~ ice_cream|deli|bakery|coffee|tea|wine))
                 and (
                     (outdoor_seating and (outdoor_seating != no)) or
                     (indoor_seating and (indoor_seating != no))
-                )
+                ))
             )
             and takeaway != only
             and (!smoking or smoking older today -8 years)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -29,10 +29,10 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     """
 
     override val elementFilter = """
-             nodes, ways, relations with
-             ${elementFilterBasic}
-             and takeaway != only
-             and (!smoking or smoking older today -8 years)
+         nodes, ways, relations with
+         ${elementFilterBasic}
+         and takeaway != only
+         and (!smoking or smoking older today -8 years)
     """
 
     override val changesetComment = "Add smoking status"
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -22,7 +22,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     """
 
     override val elementFilter = elementFilterBasic + """
-	and (!smoking or smoking older today -8 years)
+        and (!smoking or smoking older today -8 years)
     """
 
     override val changesetComment = "Add smoking status"
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -39,7 +39,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
             R.string.quest_smoking_no_name_title
 
     override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
-        val name = tags["name"] ?: tags["brand"]
+        val name = tags["name"] ?: tags["brand"] ?: tags["operator"]
         return arrayOfNotNull(name, featureName.value)
     }
 
@@ -53,5 +53,5 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     }
 
     private fun hasFeatureName(tags: Map<String, String>): Boolean =
-        tags.containsKey("name") || tags.containsKey("brand")
+        tags.containsKey("name") || tags.containsKey("brand") || tags.containsKey("operator")
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -1,6 +1,7 @@
 package de.westnordost.streetcomplete.quests.smoking
 
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.meta.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
 import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
 import de.westnordost.streetcomplete.data.osm.mapdata.Element
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -28,7 +28,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     override val elementFilter = """
             nodes, ways, relations with
             (
-                ${elementFilterBasicFragment()} and (outdoor_seating != no or indoor_seating != no)
+                ${elementFilterBasicFragment()} and (outdoor_seating and outdoor_seating != no or indoor_seating and indoor_seating != no)
             )
             and takeaway != only
             and (!smoking or smoking older today -8 years)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -1,7 +1,6 @@
 package de.westnordost.streetcomplete.quests.smoking
 
 import de.westnordost.streetcomplete.R
-import de.westnordost.streetcomplete.data.meta.isShopExpressionFragment
 import de.westnordost.streetcomplete.data.meta.updateWithCheckDate
 import de.westnordost.streetcomplete.data.osm.mapdata.Element
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -14,23 +14,19 @@ import de.westnordost.streetcomplete.ktx.containsAny
 
 class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
 
-    private val elementFilterBasic = """
-        (
-            amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
-            or leisure = outdoor_seating
-            or (
-                (amenity ~ fast_food|ice_cream or shop ~ ice_cream|deli|bakery|coffee|tea|wine)
-                and (
-                    (outdoor_seating and outdoor_seating != no)
-                    or (indoor_seating and indoor_seating != no)
-                )
-            )
-        )
-    """
-
     override val elementFilter = """
          nodes, ways, relations with
-         ${elementFilterBasic}
+         (
+             amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
+             or leisure = outdoor_seating
+             or (
+                 (amenity ~ fast_food|ice_cream or shop ~ ice_cream|deli|bakery|coffee|tea|wine)
+                 and (
+                     (outdoor_seating and outdoor_seating != no)
+                     or (indoor_seating and indoor_seating != no)
+                 )
+             )
+         )
          and takeaway != only
          and (!smoking or smoking older today -8 years)
     """
@@ -54,16 +50,8 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
         return arrayOfNotNull(name, featureName.value)
     }
 
-    /* note: copied from IS_SHOP_OR_DISUSED_SHOP_EXPRESSION as we seemingly can't extend it with extra strings */
     override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
-        getMapData().filter("""
-            nodes, ways, relations with
-              ${isShopExpressionFragment()}
-              or ${isShopExpressionFragment("disused")}
-              or shop = vacant
-              or ${elementFilterBasic}
-        """)
-
+        getMapData().filter(IS_SHOP_OR_DISUSED_SHOP_EXPRESSION)
 
     override fun createForm() = SmokingAllowedAnswerForm()
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -6,7 +6,6 @@ import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.osm.osmquests.Tags
 import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.CITIZEN
 import de.westnordost.streetcomplete.ktx.arrayOfNotNull
-import de.westnordost.streetcomplete.ktx.toYesNo
 
 class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -17,9 +17,9 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
     private fun elementFilterBasicFragment(prefix: String? = null): String {
         val p = if (prefix != null) "$prefix:" else ""
         return """
-              {p}amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
-              or {p}leisure ~ outdoor_seating
-              or ({p}amenity ~ fast_food|ice_cream or {p}shop ~ ice_cream|deli|bakery|coffee|tea|wine)
+              ${p}amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
+              or ${p}leisure ~ outdoor_seating
+              or (${p}amenity ~ fast_food|ice_cream or ${p}shop ~ ice_cream|deli|bakery|coffee|tea|wine)
         """.trimIndent()
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -1,6 +1,9 @@
 package de.westnordost.streetcomplete.quests.smoking
 
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,57 @@
...
+        else
+            R.string.quest_smoking_no_name_title
+
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        val name = tags["name"] ?: tags["brand"]
+        return arrayOfNotNull(name, featureName.value)
+    }
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter(elementFilterBasic)
```
matkoniecz, line 47: I think that it is OK to highlight also shops and other shoplike objects, like in other similar quests

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override fun createForm() = SmokingAllowedAnswerForm()
+
+    override fun applyAnswerTo(answer: SmokingAllowed, tags: Tags, timestampEdited: Long) {
+	tags.updateWithCheckDate("smoking", answer.osmValue)
+    }
+
+    private fun hasFeatureName(tags: Map<String, String>): Boolean =
```
matkoniecz, line 55: See https://github.com/streetcomplete/StreetComplete/blob/12e95b923acc47b29a2097a15c25644c5bd83f22/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt#L115-L120

AFAIK `hasFeatureName` is consistently used "featureDictionaryFuture preset has name for this set of tags" and `hasProperName` is used for "has name tags such as `name`/`brand`/`operator` (which count as naming tags depends on an object!)"

I would change name here to `hasProperName`

Maybe add `hasFeatureName` that would be used for extra check ensuring that things returned by filter all have names in presets, but I am not entirely sure here is it needed.

I need to check my explanation in #3648 is being confused/misleading

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,57 @@
...
+        tags.containsKey("name") || tags.containsKey("brand")
```
matkoniecz, line 56: I am not entirely sure is there a chance that correctly tagged `bar|cafe|pub|biergarten|restaurant|nightclub|stripclub` will have no `name` but will have `brand` tag.

app/src/main/res/values/strings.xml
```diff
@@ -1206,6 +1206,11 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="achievement_lifesaver_solved_X">You solved %d quests that help emergency services and first responders!</string>
     <string name="quest_airConditioning_no_name_title">"Does this place (%s) have air conditioning?"</string>
     <string name="quest_airConditioning_name_type_title">"Does %1$s (%2$s) have air conditioning?"</string>
+    <string name="quest_smoking_no_name_title">"Is smoking allowed anywhere in this place (%s)?"</string>
+    <string name="quest_smoking_name_type_title">"Is smoking allowed anywhere in %1$s (%2$s)?"</string>
+    <string name="quest_smoking_outside">Outside only</string>
+    <string name="quest_smoking_isolated">Prohibited everywhere, except in separate completely isolated smoking room</string>
```
matkoniecz, line 1212:
```suggestion
    <string name="quest_smoking_isolated">Prohibited everywhere, except in separate isolated smoking room</string>
```

in my experience even decent isolated rooms, passing legislation requirements, have smoke leaving at least when someone enters/leaves it (not aware about systems using airlocks :) )

matkoniecz:
> But what I'm unsure about is am I guaranteed that arrayOfNotNull(name, featureName.value) will always do right thing? Will featureName.value always be present, so at least R.string.quest_smoking_no_name_title can substitute one string?

`featureName.value` will be missing in cases where iD presets (via osmfeatures library) has no named match for given set of tags.

Given element selection

```
          amenity ~ bar|cafe|pub|biergarten|restaurant|nightclub|stripclub
          or leisure ~ outdoor_seating
```

it would happen only in case where iD presets or osmfeatures would be buggy. Or in case where someone would modify filter without testing in future. I admit that I am not sure is it recommended to guard against this kind of problem by adding extra checks.

matkoniecz:
> is it possible in other answers menu to force positioning? (i.e. I'd like quest_smoking_isolated to be show at the top)

From what I remember it is list, order matters in it. Changing positions of items there should change displayed order.

matkoniecz:
> what is the best way to hide some answers in some cases (e.g. I'd like to hide outside only answer if it is for leisure=outdoor_seating) - separate elementFilter and rest of the file, or is there a better way?

See https://github.com/streetcomplete/StreetComplete/blob/88d71c4a792376d82d5b1863553cf96f551088f7/app/src/main/java/de/westnordost/streetcomplete/quests/parking_fee/AddParkingFeeForm.kt#L27-L36 https://github.com/streetcomplete/StreetComplete/blob/9c6d3e25216d06a2c5afa71086949e1e195de926/app/src/main/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestriansForm.kt#L24-L33

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
```diff
@@ -0,0 +1,25 @@
...
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.OUTSIDE
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.NO
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.ISOLATED
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.SEPARATED
+
+class SmokingAllowedAnswerForm : AbstractQuestAnswerFragment<SmokingAllowed>() {
+
+    override val buttonPanelAnswers = listOf(
+        AnswerItem(R.string.quest_generic_hasFeature_no) { applyAnswer(NO) },
+        AnswerItem(R.string.quest_generic_hasFeature_yes) { applyAnswer(YES) },
```
matkoniecz, line 16: I need to check how it works in practice, but I suspect that with "Is smoking allowed anywhere in (...)" question people may answer "yes" where ISOLATED/SEPARATED would be better answer.

Maybe replace "Yes" by "Everywhere" option? Or display all answers (like playground access and parking access quests)?

matkoniecz (changes requested):

matkoniecz:
In general - welcome! And thanks for this pull request, I hope that all of above was not too much :)

One more extra question: is there some place where this quest should be disabled as pointless? Either with total ban in public spaces or with effectively no restrictions at all anywhere?

app/src/main/res/values/strings.xml
kmpoppe, line 1212:
```suggestion
    <string name="quest_smoking_isolated">Prohibited everywhere, except in a separate, isolated smoking room</string>
```

Just some grammar shenanigans.

kmpoppe:
> One more extra question: is there some place where this quest should be completely disabled as pointless? Either with total ban in public spaces or with effectively no restrictions at all anywhere?

There's an [extensive list](https://en.m.wikipedia.org/wiki/List_of_smoking_bans) of smoking bans by country, with a plethora of different handlings of the matter. I think apart from "total ban in the country as a whole" and thus disabling the quest in that countries, it would be too time consuming to correctly handle all the different approaches in the code. JMTC of course.

K

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: You mean by using `isKindOfShopExpression()` I've seen some other quest do? I guess I could, but:
- it misses half the tags I need highlighed, so I'd need to extend it (either `isKindOfShopExpression()` itself (which I'm too afraid to do without external guidance), of locally add extra filters. Also then I'd need to watch if `isKindOfShopExpression()` possibly gets modified in a way that requires me to update the filter (e.g. if some of terms get dropped or limited there), and commiting to "eternal maintenance" is a kind of turn-off for me :smiley: 

- Also, I'm not sure if it would really improve the situation? according to the source code comments:
> Elements that should be highlighted on the map alongside the selected one because they provide context for the given element. For example, nearby benches should be shown when answering a question for a bench so the user knows which of the benches is meant.

If I'm reading that correctly, as I'm mapping `smoking/no smoking` places, what I want to accomplish is to avoid is that users mixes up one cafe/pub (where there might be `smoking`) with another. But showing a lot of extra data around (like shops, offices, crafts, banks etc) does not really help much with that in my opinion. Sure, it sometimes perhaps might help somewhat with locating the cafe (_"ok, this cafe is to the left of post office"_), but it also possibly creates a lot of additional (unwanted?) clutter. 

But if you think that would be improvement, I can replace that with e.g.:
```diff
-getMapData().filter(elementFilterBasic)
+getMapData().filter(elementFilterBasic + isKindOfShopExpression() + " or " + isKindOfShopExpression("disused"))
```

If that is the way to go, then it would be good to also mention `isKindOfShopExpression()` in new quest docs.
 
(Or did I completely missed the point here?)

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: Neither did I, but apparently there [are a few](http://overpass-turbo.eu/?q=LyoKVGhpcyBoYcSGYmVlbiBnxI1lcmF0ZWQgYnkgdGhlIG92xJJwxIlzLXR1cmJvIHdpemFyZC7EgsSdxJ9yaWdpbmFsIHNlxLBjaMSsxIk6CsOiwoDCnGFtxI1pdMSafiAiYsSwfGNhZmV8cHVifGJpxJJnxLDElW58cmVzdGHEqGFudHxuxLhodGNsxZ18xarEt3DFtsWdIiDFrsSXxLvFjCDEhSBudWzEvcaBxJ9wxJLElG9yPSrFiMKdCiovCltvdXQ6anNvbl1bdGnFjMacxp4yNV07Ci8vxI_ElMSdciDFqHPGiXRzCigKICDGsCBxdcSSxJrEo3J0IGbGkTogxYjFisWMxbLFj34vxZRyxZbFmMWaxZzFnsWgxaLFpMSNxafFqcWrxa3Fr8WxxbPFtcW3xZ7FumnFvMepxrHGjMaDZSHGk8aAbsSXb8aOxJN0xpHGk8aVxr7Gh3dyWyLFi8WNxY8ifsWTxZXFl8WZxZvFuMedcsWjx4jHoMa3x6PEk8elxbJnxbTFvceqdMW7yJwixqQix7EiIciJLirIociDYsiXZMihKHt7YsSqeH19KcauCsi4xq_GsXDEt8WvxrbFqca5xrvGqcSYb2R5xq4-xq7JhXNrZcS9cXQ7&c=BO_juVVynJ). There is even more of those without `name` but with `operator`, so I've added that as fallback too: 1ae5f9d0c07ce230015d7fa3c5d9d7c334bb0f82

app/src/main/res/values/strings.xml
mnalis: Yeah, I was **_OVEREMPHASIZING_** on purpose there:  People should answer `isolated` **_only if there is foolproof airlock-like system_** (in Croatia, for example, there is no double-door airlock required, but ventilation system is required that sucks air from above with such power to create underpressure in smoking room so even when door opens, the air can move only from outside to inside, not other way around - needless to say very few places choose to actually implement that).

It should be selected only if there is no chance nonsmoker sitting outside of smoking room would be able to be smell the smoke / be bothered about it.  As it happens only very rarely, and introduces possible confusion,  based on those comments I'll likely remove `isolated` option completely. And I'm not sure about `separated` either, see below...

mnalis: Thanks, the language was awkward in my attempt to _**OVEREMPHASIZE**_ that part on purpose. Obviously it does not work as well as I hoped :smile:, see https://github.com/streetcomplete/StreetComplete/pull/3784#discussion_r815350764

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
mnalis: good idea, I've tried "parking access" type now in e0a4b6d4efca2fb0fae7d0f293172153fec6b00b

matkoniecz:
Note that https://github.com/streetcomplete/StreetComplete/pull/3784#issuecomment-1046173805 example, more specifically https://github.com/streetcomplete/StreetComplete/blob/9c6d3e25216d06a2c5afa71086949e1e195de926/app/src/main/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestriansForm.kt#L24-L33

is accessing tags directly within overridden parts, not directly in class

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: done in 3b75cf814a62eebd3eb2893d66bdb5773fd1031c

matkoniecz:
@mnalis Have you got it to work? Or do you need some help (yesterday it was on like 1AM so sorry if my comment was unclear)

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
```diff
@@ -0,0 +1,30 @@
...
+            TextItem(SEPARATED, R.string.quest_smoking_separated),
+            TextItem(YES, R.string.quest_smoking_yes),
+        )
+        else listOf(
+            TextItem(NO, R.string.quest_smoking_no),
+            TextItem(OUTSIDE, R.string.quest_smoking_outside),
+            TextItem(SEPARATED, R.string.quest_smoking_separated),
+            TextItem(YES, R.string.quest_smoking_yes),
+        )
+        return result
```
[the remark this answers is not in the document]
mnalis: thanks, that's what I wanted to do but had no idea how to do in Kotlin!

mnalis:
> @mnalis Have you got it to work? Or do you need some help (yesterday it was on like 1AM so sorry if my comment was unclear)

Yeah, thanks for the help a lot @matkoniecz , even when it did need some more experimentation (might do me good to flex my brain a little on puzzles like that :smile:).

<Details>

I was on the verge of giving up for that night (as  `<TextItem>` has different syntax that `<AnswerItem>` in examples so didn't want to compile, there was no other places in SC code that used `<TextItem>` like that, and StackOverflow etc. answers on compiler errors were way above my head -- when one variant finally worked: https://github.com/streetcomplete/StreetComplete/pull/3784/commits/953787a21c510c55f009504e2e11d2b104727484. At least I hope that is the way it should be written :smiley:; it seems to do what I want it to do.

</details>

Anyway, basically I think it mostly only needs the answer from you on https://github.com/streetcomplete/StreetComplete/pull/3784#discussion_r815347747 and it should be ready to get out of the Draft.

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
matkoniecz: Yes, union here works. Showing all items that would be of interest to this quest AND all shoplike stuff seems optimal.

Maybe also elementFilterBasic which are disused (would require it to support prefixes like say in https://github.com/streetcomplete/StreetComplete/commit/972637797e93a1f96dd8ec3da14e348598f8bfa7 )

matkoniecz:
> Anyway, basically I think it mostly only needs the answer from you on https://github.com/streetcomplete/StreetComplete/pull/3784#discussion_r815347747 and it should be ready to get out of the Draft.

Answered, I think.

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
```diff
@@ -0,0 +1,30 @@
...
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.YES
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.OUTSIDE
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.NO
+import de.westnordost.streetcomplete.quests.smoking.SmokingAllowed.SEPARATED
+
+class SmokingAllowedAnswerForm : AListQuestAnswerFragment<SmokingAllowed>() {
+
+    override val items: List<TextItem<SmokingAllowed>> get() {
+        val tags = osmElement!!.tags
+        val isAlreadyOutside = tags["leisure"] == "outdoor_seating" || tags["amenity"] == "biergarten"
```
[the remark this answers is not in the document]
mnalis: Aren't both true? I mean, if it is `outdoor_seating=yes` + `indoor_seating=no`, then it is also `isAlwaysOutsideOnly=true`, just like if it was `leisure=outdoor_seating`, right? I've added that now.

Also good pointer to #3758, I've incorporated some OSM element selection from it!

mnalis:
Hmmm, it was `Draft` from the beginning; I have no idea how and when it decided not be draft anymore? :thinking:
Certainly not by my will!

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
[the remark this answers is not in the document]
FloEdelmann: In Kotlin, there is likely syntactic sugar for any situation :sweat_smile: 
(Or you can make it yourself with custom extension functions :wink:)

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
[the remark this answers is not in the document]
FloEdelmann: `leisure=outdoor_seating` and `amenity=biergarten` imply `outdoor_seating=yes` + `indoor_seating=no`.

So my question was: Should the `OUTSIDE` option be hidden only for those exact places (where "only outdoor seating" is always implied), or also for elements where the type of place does not imply it, but the seating tags define it explicitly?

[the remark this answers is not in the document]
FloEdelmann: IMO `isAlwaysOutdoorOnly` is a better variable name for the former case, and `isAlreadyOutdoor` for the latter.

Note also that "outside" should probably be replaced with "outdoor" (I didn't notice that until now).

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: implemented

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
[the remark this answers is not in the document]
mnalis: for both, I just want to skip "Outdoors only" answers if the question is being asked for amenity which has only outdoors (regardless is it is defined implicitly or explicitly). So I've renamed now to `isAlreadyOutdoor` (as well as other `outside` -> `outdoor` references/texts)

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: Just my luck; in the meantime `isKindOfShopExpression()` stopped existing in the `master` :smiling_face_with_tear:

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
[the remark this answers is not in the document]
mnalis: Hmmm, GitHub is playing mind games on me... First `draft` status disappeared by itself, and now this diff that I clicked to commit is not actually committed in the git (nor is there any trace of it) - and yet `Commit suggestion` button is disabled indicating I did really press it.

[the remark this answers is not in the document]
mnalis: applied it manually now

matkoniecz:
Feel free to convert into ready for review once ready! (or ask if stuck)

mnalis:
Ok, it compiles and seems to do what is wanted (at least in several cases I tried). Ready for review.

mnalis:
There seems to be a bug with OSM element selection, which I'm not sure why it happens:

https://github.com/streetcomplete/StreetComplete/blob/a36f2b330fb33aa6440fbdf1ab1d22e65b593017/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt#L17-L35

what I want to get is:

```kotlin
(amenity ~ fast_food|ice_cream or shop ~ ice_cream|deli|bakery|coffee|tea|wine) 
and (outdoor_seating and outdoor_seating != no or indoor_seating and indoor_seating != no)
```

which (in my opinion) should show for example `amenity=fast_food` only if it has either outdoor or indoor seating (i.e. tags `outdoor_seating` or `indoor_seating` exist, and have a value different than `no`).

So for example, it **should** match:
- `amenity=fast_food`+`indoor_seating=yes` 
- `amenity=fast_food`+`indoor_seating=no`+`outdoor_seating=terrace`

but it should **not** match:
- `amenity=fast_food`+`outdoor_seating=no`+`indoor_seating=no`
- `amenity=fast_food` only (i.e.without extra tags for `*_seating` )

However, the code in current branch seems to match for example https://www.openstreetmap.org/node/4993598032 (which has only `amenity=fast_food`), which in my opinion should not happen. Help?

matkoniecz:
@mnalis 

> outdoor_seating and outdoor_seating != no or indoor_seating and indoor_seating != no

please add ()

> (outdoor_seating and outdoor_seating != no) or (indoor_seating and indoor_seating != no)

no idea about operator precedence rules applying here and code should never require understanding them, that is asking for a tricky bugs

Second check: are you 100% sure that you are testing this code?

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -19,7 +19,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
         return """
               ${p}amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
               or ${p}leisure ~ outdoor_seating
-              or (${p}amenity ~ fast_food|ice_cream or ${p}shop ~ ice_cream|deli|bakery|coffee|tea|wine)
+              or ((${p}amenity ~ fast_food|ice_cream) or (${p}shop ~ ice_cream|deli|bakery|coffee|tea|wine))
```
matkoniecz, line 22: this one is a bit too much, `key~value|list` is inherently one element

mnalis: yes, I might have overdone it to be absolutely sure for debug... but **_still it does not work_**:

`.apk` from this build  https://github.com/mnalis/StreetComplete/actions/runs/1907292270 which corresponds to this file https://github.com/mnalis/StreetComplete/blob/e2e103dbbbd774ea40a5642a95a81a5b466a95eb/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt#L17-L39

It still activates a quest on this node, and it shouldn't:  https://www.openstreetmap.org/node/4993598032

matkoniecz: @mnalis 

Are you sure that you run latest code and it is pushed? I fetched it to debug what is going on and it failed on compilation error, with `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` not being defined.

https://github.com/streetcomplete/StreetComplete/pull/3784/files appears to be not defined in your branch.

It is defined in latest StreetComplete code, but it is not yet merged in to your branch.

After `git merge master` compilation succeeds.

matkoniecz: @mnalis 

Are you sure that you run latest code and it is pushed? I fetched it to debug what is going on and it failed on compilation error, with `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` not being defined.

https://github.com/streetcomplete/StreetComplete/pull/3784/files appears to be not defined in your branch.

It is defined in latest StreetComplete code, but it is not yet merged in to your branch.

After `git merge master` compilation succeeds.

I build code from such version and https://www.openstreetmap.org/node/9200004349 was ineligible.

matkoniecz: @mnalis 

Are you sure that you run latest code and it is pushed? I fetched it to debug what is going on and it failed on compilation error, with `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` not being defined.

https://github.com/streetcomplete/StreetComplete/pull/3784/files appears to be not defined in your branch.

It is defined in latest StreetComplete code, but it is not yet merged in to your branch.

After `git merge master` compilation succeeds.

I build code from such version and https://www.openstreetmap.org/node/9200004349 was ineligible.

> It still activates a quest on this node, and it shouldn't: https://www.openstreetmap.org/node/4993598032

It does not for me. It activated for nearby https://www.openstreetmap.org/node/1917302043

matkoniecz: It is also not activated for https://www.openstreetmap.org/node/4993597838

(nevertheless, I am not sure about requiring seating to be tagged, but that is a separate topic)

mnalis:
Well, it looks working OK to me now.

However, I'm not terribly happy with copy/pasting of `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` in https://github.com/streetcomplete/StreetComplete/pull/3784/commits/aab19132860f057eed88b4db39a8801ef6fa7ba3. Is there a way to use it as a string to append another string to it? 

i.e. I originally wanted to do `getMapData().filter("${IS_SHOP_OR_DISUSED_SHOP_EXPRESSION} or ${elementFilterBasic}")` - but that fails to compile, and I don't understand kotlin to imagine a better solution

mnalis:
Anyway, apart from the minor code ugliness, it looks good to me now, so ready for review.

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowed.kt
```diff
@@ -0,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.smoking
+
+enum class SmokingAllowed(val osmValue: String) {
+    YES("yes"),
+    OUTDOOR("outdoor"),
```
FloEdelmann, line 5: The OSM tagging is `outside` in this case: https://wiki.openstreetmap.org/wiki/Key:smoking

My comment about renaming outside to outdoor was only targeted at the `isAlreadyOutside` variable, as it refers to `outdoor_seating` there. Sorry for not being clear!

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,77 @@
...
+            )
+        )
+    """
+
+    override val elementFilter = """
+             nodes, ways, relations with
+             ${elementFilterBasic}
+             and takeaway != only
+             and (!smoking or smoking older today -8 years)
+    """
```
FloEdelmann, line 36: Unnecessary indentation:

```suggestion
    override val elementFilter = """
         nodes, ways, relations with
         ${elementFilterBasic}
         and takeaway != only
         and (!smoking or smoking older today -8 years)
    """
```

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowed.kt
mnalis: should be fixed by https://github.com/streetcomplete/StreetComplete/pull/3784/commits/1633c940b7f01570beb4c54addd52ec52df0a3fb

FloEdelmann (approved):
LGTM! :slightly_smiling_face:

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,77 @@
...
+    override val questTypeAchievements = listOf(CITIZEN)
+
+    override fun getTitle(tags: Map<String, String>) =
+        if (hasProperName(tags))
+            R.string.quest_smoking_name_type_title
+        else
+            R.string.quest_smoking_no_name_title
+
+    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> {
+        val name = tags["name"] ?: tags["brand"] ?: tags["operator"]
```
matkoniecz, line 53: I am not convinced that `operator` is valuable here with any conceivable correct tagging,

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    /* note: copied from IS_SHOP_OR_DISUSED_SHOP_EXPRESSION as we seemingly can't extend it with extra strings */
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
+            nodes, ways, relations with
+              ${isShopExpressionFragment()}
+              or ${isShopExpressionFragment("disused")}
+              or shop = vacant
+              or ${elementFilterBasic}
+        """)
```
westnordost, line 65: I do not think this is worth duplicating the code here. What cases are caught with this over `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION`? 

As far as I see, just `amenity ~ food_court|nightclub|stripclub or leisure = outdoor_seating`. I wouldn't expect any of these to be right next to each other. Maybe two nightclubs, but this is a really rare situation.

Actually, I wonder if `nightclub` and `stripclub` could not be added to "isKindOfShop" filter instead. Aren't those typically in the same kind of rented properties as pubs, bars etc. would be?

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -0,0 +1,77 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.ktx.arrayOfNotNull
+import de.westnordost.streetcomplete.ktx.containsAny
+
+class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
+
+    private val elementFilterBasic = """
+        (
+            amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
+            or leisure ~ outdoor_seating
```
westnordost, line 20:
```suggestion
            or leisure = outdoor_seating
```

app/src/main/res/values/strings.xml
```diff
@@ -1206,6 +1206,12 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="achievement_lifesaver_solved_X">You solved %d quests that help emergency services and first responders!</string>
     <string name="quest_airConditioning_no_name_title">"Does this place (%s) have air conditioning?"</string>
     <string name="quest_airConditioning_name_type_title">"Does %1$s (%2$s) have air conditioning?"</string>
+    <string name="quest_smoking_no_name_title">"Is smoking allowed anywhere in this place (%s)?"</string>
+    <string name="quest_smoking_name_type_title">"Is smoking allowed anywhere in %1$s (%2$s)?"</string>
+    <string name="quest_smoking_no">No, smoking is forbidden</string>
+    <string name="quest_smoking_outside">Only allowed outside</string>
+    <string name="quest_smoking_separated">Allowed in some areas, but not in others</string>
+    <string name="quest_smoking_yes">Yes, smoking is allowed everywhere</string>
```
westnordost, line 1214:
```suggestion
    <string name="quest_smoking_no">No, smoking is forbidden</string>
    <string name="quest_smoking_outside">Only allowed outside</string>
    <string name="quest_smoking_separated">Allowed in separated area</string>
    <string name="quest_smoking_yes">Yes, smoking is allowed everywhere</string>
```
I find this clearer to be honest.

Also, I haven't followed the discussion so far, but is there any reason why the answers need to be that wordy? the more words, the less people read it. The current answers kind of repeat the question.
```suggestion
    <string name="quest_smoking_no">No, forbidden</string>
    <string name="quest_smoking_outside">Only outside</string>
    <string name="quest_smoking_separated">Only in separated area</string>
    <string name="quest_smoking_yes">Yes, everywhere</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: Well they **_are_** on the map; check this overpass for example: https://overpass-turbo.eu/s/1gwj

Unfortunately timeouts on bigger areas, but for example in city of Rome, Italy (default overpass-turbo.eu location) it seems to be _at least_ as popular as places without `name` but with `brand` (and in this particular case much more popular; although at other places `brand` might be more popular; see also https://github.com/streetcomplete/StreetComplete/pull/3784#discussion_r815349445)

So I think it would be better to show it to the user, then to hide it (in cases `operator` exists, but `name` and `brand` do not). Or do you disagree?

westnordost: I disagree for two reasons:
1. AFAIR it is inconsistent with the rest of the code. There are many places where you find `tags["name"] ?: tags["brand"]`. If `operator` is going to be added as a second fallback, it should be done consistently everywhere (and not in this PR)
2. In my experience, some shops are unnamed in the sense that there is no proper sign with a name. But sometimes, the  owner of the shop is known. `operator` is in this case used for the owner of the shop. But this is usually not signed. Depending on the country and legislation, shop owners may be required to display their license somewhere visible (a piece of paper) and this is why the name of the owner may even be known. But this is not suitable for orientation and/or referring to the shop because it is a very non-prominent piece of information - if it is displayed at all

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: > As far as I see, just amenity ~ food_court|nightclub|stripclub or leisure = outdoor_seating. I wouldn't expect any of these to be right next to each other. Maybe two nightclubs, but this is a really rare situation.

Well, `leisure=outdoor_seating` especially are quite often near one another, as they are commonly in front of cafes, which are often (at least down here) crowded together esp. in downtowns, eg. [overpass](http://overpass-turbo.eu/?Q=%2F*%0AThis%20has%20been%20generated%20by%20the%20overpass-turbo%20wizard.%0AThe%20original%20search%20was%3A%0A%E2%80%9Cleisure%20%3D%20outdoor_seating%E2%80%9D%0A*%2F%0A%5Bout%3Ajson%5D%5Btimeout%3A25%5D%3B%0A%2F%2F%20gather%20results%0A(%0A%20%20%2F%2F%20query%20part%20for%3A%20%E2%80%9Cleisure%3Doutdoor_seating%E2%80%9D%0A%20%20nwr%5B%22leisure%22%3D%22outdoor_seating%22%5D(%7B%7Bbbox%7D%7D)%3B%0A%0A)%3B%0A%2F%2F%20print%20results%0Aout%20body%3B%0A%3E%3B%0Aout%20skel%20qt%3B&C=45.81196;15.97645;17&R)

So I'd like to **_at least_** _additionally_ display  `leisure=outdoor_seating`. `nightclub` and `stripclub` could  be added to `isKindOfShop` IMO (but I'm not an expert on all places where it is used in code) - but that would not solve my problem by itself, as I want to add at least that one tag more.

Ideas I've pondered (besides current copy/paste implementation which I'm not terribly happy about either):
- just filter on `elementFilterBasic`  (i.e. only highlight places which could be asked by this quest)
- extend  `isKindOfShop` - as you said, it is OK for `nightclub` and `stripclub`, but how about `leisure=outdoor_seating`? Anyway I would not touch code outside `smoking/` directory without your approval
- create `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` class/object (or how is it called in kotlin) from  `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION_STRING` instead; and then I could use simple `${IS_SHOP_OR_DISUSED_SHOP_EXPRESSION_STRING} or ${elementFilterBasic}` (or is there some other way to merge them in kotlin?). Then, this and other quests could extend that default with just a thing or two they need, without this ugly copy/paste. But again, it is outside my code, so I'd need to know if you'd be OK with that change. i.e.
```diff
- val IS_SHOP_OR_DISUSED_SHOP_EXPRESSION = """
+ val IS_SHOP_OR_DISUSED_SHOP_EXPRESSION_STR = """
    nodes, ways, relations with
      ${isShopExpressionFragment()}
      or ${isShopExpressionFragment("disused")}
      or shop = vacant
- """.toElementFilterExpression()
+ """
+ val IS_SHOP_OR_DISUSED_SHOP_EXPRESSION = IS_SHOP_OR_DISUSED_SHOP_EXPRESSION_STR.toElementFilterExpression()
```
- Or maybe just `.toElementFilterExpression()` could be dropped from existing `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` ? Majority of uses seem to be `getMapData().filter` which as I understand can also take plain string as an argument, so they can continue to use it unchanged; and few other places could add `.toElementFilterExpression()` themselves if they need it (excuse my naivete if that is impossible/undesirable)

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: 1. Yes, I think it would be good fallback in the rest of the code too if neither `name` nor `brand` are present, but I do not presume to make a decision about that for whole codebase. Some other pieces of software like OsmAnd will display `operator` if `name` is missing, so it seems to be (at least partly) accepted practice.
2. Use case I've commonly seen is that cafe/fast_food is part of some supermarket, so while it is not named separately, it is operated by that supermarket. Also it might depend on cultural differences; some places especially smaller family restaurants, cafes etc. might take pride / be referred to by its proprietor.

Anyway, let me know of your decision, and I'll abide.

westnordost: Yes, please remove operator

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
westnordost: Hm ok, I'd find best if:
- strip and night club is added to the "shop expression"
- leave the code in `AddSmoking` as is, mainly because of `leisure=outdoor_seating`

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: removed

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
mnalis: OK, added in PR https://github.com/streetcomplete/StreetComplete/pull/3822
Leaving `smoking/` code as-is.

westnordost: Or actually... please just 
- add strip and night club to the "shop expression"
- use IS_SHOP_OR_DISUSED_SHOP_EXPRESSION only in getHighlightedElements

Reason: highlighting outdoor seatings may actually confuse the user, as there seem to be unnamed shops close by in e.g. the example you posted. The user will be used to seeing shops etc. highlighted for these kinds of quests

mnalis: as you wish. done.

westnordost:
Thank you!

westnordost (approved):

app/src/main/res/values/strings.xml
```diff
@@ -1206,6 +1206,12 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="achievement_lifesaver_solved_X">You solved %d quests that help emergency services and first responders!</string>
     <string name="quest_airConditioning_no_name_title">"Does this place (%s) have air conditioning?"</string>
     <string name="quest_airConditioning_name_type_title">"Does %1$s (%2$s) have air conditioning?"</string>
+    <string name="quest_smoking_no_name_title">"Is smoking allowed anywhere in this place (%s)?"</string>
+    <string name="quest_smoking_name_type_title">"Is smoking allowed anywhere in %1$s (%2$s)?"</string>
+    <string name="quest_smoking_no">No, forbidden</string>
+    <string name="quest_smoking_outside">Only outside</string>
+    <string name="quest_smoking_separated">Only in separated area</string>
```
FloEdelmann, line 1213:
```suggestion
    <string name="quest_smoking_separated">Only in separated areas</string>
```

mnalis:
Hopefully that should be it. I don't know why GitHub claims there are still outstanding unresolved changes requested by @matkoniecz , I've looked several times and all seems resolved.

matkoniecz:
> Hopefully that should be it. I don't know why GitHub claims there are still outstanding unresolved changes requested by @matkoniecz , I've looked several times and all seems resolved.

That is because my latest review had status "changes requested" and have not made new one with "approved" status. (@westnordost  and @FloEdelmann reviewed and approved so do not worry about it)

matkoniecz (commented):
looks fine, but I have not tested it

wetneb:
@mnalis thank you so much for implementing this quest, I use it a lot in Germany (Saxony) where a lot of bars still allow smoking.

One thing that is vaguely annoying though, is that it asks whether bakeries are non-smoking or not. To me, it does not feel very useful to tag them because they are basically always non-smoking. According to this [Sophox query](https://tinyurl.com/27udv9z6) we know of 5 bakeries in the entire world which allow some smoking inside the bakery - they feel like rare exceptions that I would not bother with. Do you agree and should I make a PR to change this?

mnalis:
> @mnalis thank you so much for implementing this quest, I use it a lot in Germany (Saxony) where a lot of bars still allow smoking.

I'm glad you like it!

> One thing that is vaguely annoying though, is that it asks whether bakeries are non-smoking or not

Does it happen that often?  It should only ask that quest on bakeries which have tagged that they have (indoor or outdoor) seating. (which, according to taginfo, is only [about 3%](https://taginfo.openstreetmap.org/tags/shop=bakery#combinations) of all bakeries)
 
> According to this [Sophox query](https://tinyurl.com/27udv9z6) we know of 5 bakeries 

Is Sophox working OK those days? I seem to remember it having issues, and [Overpass numbers](https://overpass-turbo.eu/s/1umD) disagree, even if only by 1 order of magnitude.

> in the entire world which allow some smoking **inside** the bakery

Yes, but the quest also asks about smoking **outside** of bakery (i.e. `smoking=outside` answer), see [SmokingAllowedForm.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedForm.kt). If that is taken into account, there is some [~500 bakeries with smoking allowed](https://overpass-turbo.eu/s/1umC) - which is about 40% of all bakeries which have `smoking=*` tagged).

> they feel like rare exceptions that I would not bother with. Do you agree and should I make a PR to change this?

I'd suggest to go with an new SC GitHub issue to decide what way to go, if there is some change that would be useful worldwide. It might be sensible to discuss on [community.openstreetmap.org](community.openstreetmap.org) first for wider feedback perhaps? (please do link here if you open a thread). 
Personally, in Croatia, I'd agree that regarding bakeries smoking inside is quite rare, but is much more popular outside. But it could be different in other countries (e.g. I seem to recall that smoking inside was _much_ more popular in Bosnia and Herzegovina)

IIRC there has been discussion earlier about showing reduced set of answers for some amenities like that (like e.g. offering only `no` and `outside` answers for bakeries, although it was different amenity in that discussion I think), but it was decided that it was needless complication, might be country-dependent, and would confuse users more than it would help (as there would still be same amount of quests shown, just the reduced number of answers per quest).

wetneb:
@mnalis yesterday I had a revelation: I discovered a bakery with a dedicated smoking area in Germany. Crazy.
So I take it back. Long live StreetComplete and its smoking quest!

PR #3786, opened 2022-02-20
Look for kerbs where crossing meets traffic island

arrival-spring:
Not sure how expensive the checks are, in terms of CPU etc., so not sure if it's worth it given that `footway=traffic_island` is fairly rare.

Tested and working

Happy to close if not wanted.


matkoniecz:
It is kind of barely used, 1500 uses worldwide ( https://taginfo.openstreetmap.org/keys/footway#values ).

screen

If added, also tests should be added to confirm that it works well.

I admit that I am not really motivated to analyze that code again given its complexity and how rare this tag is.

westnordost:
Looks ok. The performance should not be any issue here. However, it is necessary to test case(s) for this. I'd suggest...
```
aaa = footway=crossing
bbb = footway=traffic_island
cccc = footway=sidewalk

1. aaaxbbb -> yields a kerb at x
2. bbbxccc -> yields no kerb at x
3. bbbxaaa -> yields no kerb at x
      c
      c
      c
``` 

I share @matkoniecz notion that it is barely used, so barely worth the effort. There is also the danger that we start supporting a tag that is not blessed by community consent - is this documented somewhere, was it discussed anywhere? It could always be that after such a discussion, consent is that this tag should not be used for one reason or the other and rather a certain replacement should be used.

matkoniecz:
It is documented at https://wiki.openstreetmap.org/wiki/Tag%3Afootway%3Dtraffic_island and appears to make sense to me and I see no complaints at Wiki about it.

Also, SC would not be actually adding it.

But... 1 504 uses worldwide...

arrival-spring:
Let's close it then

gdabski:
I have user locally who insists on using this tag; result is missing info on kerbs, which mostly come from SC quests around here.

westnordost:
Hm well, it's up to 6 493 usages now since one year ago and it is documented. Looks like this can be revisited / implemented.

westnordost:
The one thing missing though is test cases.

arrival-spring:
I'll have a look at adding some tests soon

arrival-spring:
Done, quite easy as there are already similar tests for sidewalk/crossing

All the tests pass and I've also manually tested that the quest also now shows up as expected

westnordost:
Great, thank you!

PR #3787, opened 2022-02-20
Fix wrong tags being applied for the AddStreetParking quest

tapetis:
For the following parking orientation and position

the app currently incorrectly generates the following OSM tags even though the parking type (i.e. diagonal and parallel) is not the same for both sides:

This PR fixes this bug and the app now correctly applies the following tags as described in the [wiki](https://wiki.openstreetmap.org/wiki/Key:parking:lane#Parking_position):

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/street_parking/AddStreetParking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/street_parking/AddStreetParking.kt
@@ -53,7 +53,7 @@ class AddStreetParking : OsmFilterQuestType<LeftAndRightStreetParking>() {
        - roundabouts
        - on sections of the roadway marked with arrows (turn lanes)
 
-       There are some more rules which cannot be filtered due to the lack of tags for that that are
+       There are some more rules that cannot be filtered due to the lack of tags that are
        set on the road-way:
        - in front of important signs (STOP, saltires, yield etc)
        - at taxi stands, bus stops, ...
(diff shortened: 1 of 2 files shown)


matkoniecz:
Given that location of parking is the same on both sides - is it really incorrect?

tapetis:
Tagging the parking position for both sides separately can of course be considered redundant. However, when only one tag is used, some users (like me) may think that this is just a tagging mistake and that either
- the `:both` should be `:right` or `:left`
- the type/orientation should be the same for both sides.

Specifying both sides explicitly prevents such a potential confusion.

In addition, I thought that just using one tag was not the intended behavior of the code, because the parking position value was stored in a variable called "orientation".

westnordost:
Oh, right! Thank you for the fix! I wonder, why was this obvious mistake not covered by tests? Do the tests still run through with your change? Would you mind adding a test for this if there isn't any?

westnordost (approved):

tapetis:
I will add a new test case for this parking scenario tomorrow.

westnordost:
This mistake was a blatant error on my part by the way and I am very happy that you found it before this quest was added to any release version. The reason for it not being in a release version is however not that I didn't deem it stable and ready but because I wanted to wait until a "street parking condition" quest has been created (by Mateusz) so that it can go live at the same time.

So, really, thank you, you averted a possible tagging disaster / data corruption!

westnordost (approved):

PR #3794, opened 2022-02-22
improve English in note creation window

### LINKED ISSUE #3792: "Create new note" UI uses unusual English
The "Create new note" UI says "Best write the note in the locally spoken language or otherwise English". Using "best" like this is fairly uncommon, at least where I am (Ohio, so midwestern AmE), usually, an exclamation, and usually quite informal ("Best get going then!"). The "or otherwise English" bit also feels slightly off. I would suggest something like "Ideally, write the note in the local language, or otherwise in English" or "Write the note in the locally spoken language, or alternatively English."

image

**Versions affected**
Using version 40.0 from F-Droid.

matkoniecz:
"Best write the note in the locally spoken language or otherwise English." -> "Ideally, write the note in the local language, or otherwise in English."

fixes #3792

improves English from #1471

Creating PR and pinging @peternewman @smichel17 as I also authored the original and I am not a native speaker (let me know if you prefer to not be pinged as English experts in SC issues/PRs)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -787,7 +787,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_leaf_type_needles">Needles</string>
     <string name="quest_leaf_type_broadleaved">Broadleaved</string>
     <string name="quest_leaf_type_mixed">Both are present</string>
-    <string name="create_new_note_hint">Best write the note in the locally spoken language or otherwise in English.</string>
+    <string name="create_new_note_hint">Ideally, write the note in the local language, or otherwise in English.</string>
     <string name="quest_cyclewayPartSurface_title">What’s the surface of the cycleway here?</string>
     <string name="quest_footwayPartSurface_title">What’s the surface of the footway here?</string>
     <string name="quest_board_type_history">History</string>


westnordost (approved):

app/src/main/res/values/strings.xml
```diff
@@ -787,7 +787,7 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_leaf_type_needles">Needles</string>
     <string name="quest_leaf_type_broadleaved">Broadleaved</string>
     <string name="quest_leaf_type_mixed">Both are present</string>
-    <string name="create_new_note_hint">Best write the note in the locally spoken language or otherwise in English.</string>
+    <string name="create_new_note_hint">Ideally, write the note in the local language, or otherwise in English.</string>
```
peternewman, line 790: Pedantically some places could have more than one local language, see e.g. https://en.wikipedia.org/wiki/Multilingualism_in_Luxembourg . I don't know if we care enough to say?
```suggestion
    <string name="create_new_note_hint">Ideally, write the note in a local language, or otherwise in English.</string>
```

I suspect most French people say in Germany have a good enough grasp of English to write it in English, but they are probably one of the few edge cases.

peternewman (approved):
LGTM.

app/src/main/res/values/strings.xml
peternewman, line 790: I think you could drop the first comma, although the Internet seems conflicted:
```suggestion
    <string name="create_new_note_hint">Ideally write the note in the local language, or otherwise in English.</string>
```

tapetis: I think that one should rather argue about the second comma. I would delete it because "otherwise in English" provides necessary information for the sentence and cannot stand on its own.

matkoniecz: I asked one more extra native speaker and they recommended "the" version

Eiim: Without looking at any prescriptive sources, I agree that the second comma is more debatable than the first, but I don't think either is a big deal either way.

kmpoppe: Using Grammarly for this sentence, my gut feeling was approved: The first comma should be there. For the second comma, I'd also go with @tapetis and remove it.

kmpoppe: > I asked one more extra native speaker and they recommended "the" version

I don't think this is a "native speaker" type of problem, @peternewman's example of Luxembourg is great, the language used is dependent on the situation, not the geographical area. I'd therefore go with "a local language".

matkoniecz: I mentioned to them also issue of multiple local languages.

(I do not really feel qualified to judge here)

westnordost:
Is this now ready to merge?

westnordost:
Anyway, I will merge this now as is. 

We are literally changing two words here, which is already a bit pedantic based on that it was understandable yet maybe "a little bit odd" before. At some point, it really becomes absurd to bicker over "a" or "the" or whatever. With this kind of pedantry, we'd never get anything done.

matkoniecz:
And if some native speaker feels strongly that this the/a is wrong: feel free to open a pull request.

app/src/main/res/values/strings.xml
smichel17: I was debating was whether "otherwise" should be flanked by (appositive) commas. But it doesn't really matter, they're all fine enough.

```suggestion
    <string name="create_new_note_hint">Ideally, write the note in the local language or, otherwise, in English.</string>
```

PR #3799, opened 2022-02-22
Prevent Dialog Cancel

### LINKED ISSUE #3783: Tapping off disabled message still enables quest
When enabling a 'disabled by default' quest there's a message that pops up with options to enable or cancel.

If instead of tapping on one of the buttons you tap outside of the popup then the popup goes away but the quest is enabled anyway.

This seems counter-intuitive and I'd expect it not to enable the quest.

kmpoppe:
Partly resolves #3783.

Using `.setCancelable(true)` will prevent accepting clicks outside the dialog and thus not cancel the dialog, which means you **have** to use either the Yes or Cancel buttons. Nothing else will cancel the dialog.

Well...

Almost nothing, except for rotating the device 🙄 

There is a way to prevent the [Dialog Dismissal on Device Rotation](https://stackoverflow.com/a/15729764), yet at 23:30 I'm unable to make that Java code work in Kotlin - which might also be partly due to the last change to the answer being 5 years and thus - in Android terms - an eternity ago. If someone feels like implementing that, feel free to commit those changes to my branch and this PR.

Kai

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/settings/questselection/QuestSelectionAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/settings/questselection/QuestSelectionAdapter.kt
@@ -309,8 +309,10 @@ class QuestSelectionAdapter(
                 visibleQuestTypeController.setVisible(item.questType, item.visible)
             }
             if (b && item.questType.defaultDisabledMessage > 0) {
-                AlertDialog.Builder(compoundButton.context)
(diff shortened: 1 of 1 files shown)


westnordost:
I'll have to reject this PR because it does not solve the issue completely, as you mention.

I know how to fix this properly, but explaining it here would about take as long as implementing it myself, so I should do that now because closing this PR without saying how it should be done would not be fair.

PR #3811, opened 2022-02-25
Show nearby address pins on housenumber quest

### LINKED ISSUE #3790: Show nearby addresses in housenumber quest
**Use case**

I'm adding housenumbers on a street with different numbers on each side (usually odds and evens), all of the houses are essentially identical (perhaps semi-detached or terraced) so it's not necessarily obvious exactly where I'm up to just based on rough GPS location and the map.

At the moment I can tap onto the next quest for a house (either street or building levels) and it tells me what number that is and then I can remember that when I enter the next one on that side, but it would be nicer if I could see nearby house numbers when I'm answering the quest and that would save me from making mistakes.

(Obviously for house numbers added enough time ago they show up on the map, but recent ones aren't rendered yet).

**Proposed Solution**

I've played around a bit with this locally and it does seem beneficial, however I'm opening an issue rather than just a PR as I have questions about implementation and not sure if it's wanted anyway.

* In MainFragment add "addr:housenumber" then "addr:housename" to the possibilities for the title of highlighted elements.
* In AddHousenumber add highlighted elements, possibly excluding any shop, amenity, etc.
* Possibly add an icon (smaller version of housenumber quest icon) to use in PinIcons. Here it would need to not return that for shops, amenities etc. so it wouldn't affect highlighted elements for shop quests where no icon is shown

arrival-spring:
Fixes #3790 

(example data for the purpose of the screenshot, not uploaded)

It feels like duplication to exclude amenity, shop etc twice but I think it is needed in both places:
- In housenumber quest highlighted elements: because otherwise it will pick up shops and their name will be shown as the title (name rightly comes before housenumber when looking for a title)
- In pin icons because otherwise the housenumber pin icon would be shown in other quests which use pins (e.g. any shop quest), since shops don't have their own icon


app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
```diff
@@ -59,5 +60,9 @@ import de.westnordost.streetcomplete.R
     when (map["tourism"]) {
         "information" -> return R.drawable.ic_pin_information
     }
+    if ((map.containsKey("addr:housenumber") or map.containsKey("addr:housename"))
+        and !map.containsAnyKey("amenity", "craft", "leisure", "office", "shop", "tourism")){
```
FloEdelmann, line 64: This should probably not be a [bitwise](https://kotlinlang.org/docs/basic-types.html#operations) `and`.


```suggestion
        && !map.containsAnyKey("amenity", "craft", "leisure", "office", "shop", "tourism")){
```

FloEdelmann, line 64: This should probably not use [bitwise logic](https://kotlinlang.org/docs/basic-types.html#operations) `and`/`or` infix functions.


```suggestion
    if ((map.containsKey("addr:housenumber") || map.containsKey("addr:housename"))
        && !map.containsAnyKey("amenity", "craft", "leisure", "office", "shop", "tourism")){
```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
```diff
@@ -129,6 +130,10 @@ class AddHousenumber : OsmElementQuestType<HousenumberAnswer> {
     override fun isApplicableTo(element: Element): Boolean? =
         if (!buildingsWithMissingAddressFilter.matches(element)) false else null
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways, relations with (addr:housenumber or addr:housename) " +
+            "and !amenity and !craft and !leisure and !office and !shop and !tourism")
```
FloEdelmann, line 135: Please use multiline strings to stay consistent with other element filters and to avoid string concatenation.


```suggestion
        getMapData().filter("""
            nodes, ways, relations with
              (addr:housenumber or addr:housename)
              and !amenity and !craft and !leisure and !office and !shop and !tourism
        """)
```

arrival-spring:
Thanks for the review!

westnordost:
Hmm... this implementation is indeed very hacky. I am not sure if the refactor necessary to make this not hacky is worth the enhancement this brings.

Independent of that, I think it would be better to show no icon, just the number. This could maybe be achieved by using a 1x1 transparent icon

arrival-spring:
Right, so should I do any more work on this, or close it as wontfix?

westnordost:
Hmm... I wish there were some objective guidelines I could use to decide how much complexity is too much complexity that would be transparent and accountable. Please understand my position, that bit by bit, more and more complexity creeps into this app and in the process, there is more and more things that need to be maintained, that can be enhanced, fine-tuned, that need fixes etc etc.. 
and at one point in time, you could employ one person full time just to triage and maintain the app like that without even any real own development. If there was money for that. (Recently, I have begun recording how much time I spend on SC every day just for these kind of tasks and it kind of shocked me. I will need to reduce the time I spend on this drastically soon - by a factor of 5 or so - for it to become sustainable in the long run and with the current influx of tickets, I don't know how this can be done.)

That being said, there is no good reason to reject this PR, if the getting the label text for the highlighted elements is done in a less hacky way: Move the code to an own function just like `getPinIcon` and have the logic that decides which tag is shown as a label there. You should search for usages of `getPinIcon`, it is used at one other place too.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
```diff
@@ -129,6 +130,12 @@ class AddHousenumber : OsmElementQuestType<HousenumberAnswer> {
     override fun isApplicableTo(element: Element): Boolean? =
         if (!buildingsWithMissingAddressFilter.matches(element)) false else null
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""nodes, ways, relations with
+            (addr:housenumber or addr:housename)
+            and !amenity and !craft and !leisure and !office and !shop and !tourism
```
westnordost, line 136: StreetComplete only asks housenumbers for buildings, so maybe instead do something like this?
```kotlin
.filter { it.tags.keys.all { key -> key.startsWith("addr:") || key == "building" } }
```
i.e. only show addresses that are "pure" addresses or that are buildings.

It may be more future proof

matkoniecz:
> Recently, I have begun recording how much time I spend on SC every day just for these kind of tasks and it kind of shocked me. I will need to reduce the time I spend on this drastically soon - by a factor of 5 or so - for it to become sustainable in the long run and with the current influx of tickets, I don't know how this can be done

Triage between high priority ("unbreak now" or "I personally want to implement it as an entertainment" or "actual bug") and everything else?

Options for everything else can include

- closing
- ask people like me to WONTFIX more aggressively
- marking with a special label ("waiting for PR")
- moving to a separate project (Github implemented issue transfer)
- moving to discussions
- restrict issue creation in main project, suggestions are created by default in a separate project and only high-priority issues are migrated into main project
- reduce emphasis about welcoming feature requests
- mix of above

In general if you notice the problem with long-term sustainability (and you mention it again! That is not a new thing!) then I would strongly advocate to prefer long-term involvement over overextending and crashing.

smichel17:
@westnordost Can you post https://github.com/streetcomplete/StreetComplete/pull/3811#issuecomment-1051300854 as a new discussion?

matkoniecz:
@smichel17 Well, I guess that he can do it he wants. Requesting to do stuff is not helping with being overloaded.

smichel17:
I can/will do it myself, but it only takes a few seconds and puts the right person's name on the post.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
arrival-spring: Hmm, then I start thinking about all of the other valid things that could be on an address node: `entrance`, `source:*` and maybe others I haven't thought of...

westnordost: True

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
```diff
@@ -59,5 +60,20 @@ import de.westnordost.streetcomplete.R
     when (map["tourism"]) {
         "information" -> return R.drawable.ic_pin_information
     }
+    if ((map.containsKey("addr:housenumber") || map.containsKey("addr:housename"))
```
westnordost, line 63: What about conscription number etc?

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
```diff
@@ -59,5 +60,20 @@ import de.westnordost.streetcomplete.R
...
+        return R.drawable.ic_none
+    }
     return null
 }
+
+fun getTitle(map: Map<String, String>): String? {
+    return if (map["name"] != null || map["brand"] != null) {
+        map["name"] ?: map["brand"]
+    }
+    else if (map["addr:conscriptionnumber"] != null && map["streetnumber"] != null) {
```
westnordost, line 74:
```suggestion
    } else if (map["addr:conscriptionnumber"] != null && map["streetnumber"] != null) {
```

Reads odd otherwise .oO("how can this still be executed if there is already a return?")

westnordost: Also, you mean `addr:streetnumber`, not `streetnumber`.

westnordost:
Which questions?

arrival-spring:
Oh, it seems my reviews don't show up! How do I make them show?

Copied below though:

In PinIcons:
`if ((map.containsKey("addr:housenumber") || map.containsKey("addr:housename"))
        && !map.containsAnyKey("amenity", "craft", "leisure", "office", "shop", "tourism")){`
Possibly this (hacky) check could be left out now that it uses a transparent icon. Not sure how offputting it would be to have something highlighted in e.g. opening hours quest where nothing is shown except housenumber

In AddHousenumber:
```override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
        getMapData().filter("""nodes, ways, relations with
            (addr:housenumber or addr:housename or addr:conscriptionnumber or addr:streetnumber)
            and !name and !brand
```
I _think_ this is better than excluding any amenity, shop etc., but risk is if getTitle function is changed to look for something else that it could then be shown by this

arrival-spring:
> Maybe there are in WIP state and not submitted?

Ah, didn't realise I had to, thanks

(Same thing had happened in #3735)

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddHousenumber.kt
```diff
@@ -129,6 +130,12 @@ class AddHousenumber : OsmElementQuestType<HousenumberAnswer> {
     override fun isApplicableTo(element: Element): Boolean? =
         if (!buildingsWithMissingAddressFilter.matches(element)) false else null
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""nodes, ways, relations with
+            (addr:housenumber or addr:housename or addr:conscriptionnumber or addr:streetnumber)
+            and !name and !brand
```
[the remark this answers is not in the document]
westnordost: This will highlight vacant shops for example though. But I guess this is OK, after all, their housenumber is already shown anyway.

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
```diff
@@ -59,5 +60,20 @@ import de.westnordost.streetcomplete.R
     when (map["tourism"]) {
         "information" -> return R.drawable.ic_pin_information
     }
+    if ((map.containsKey("addr:housenumber") || map.containsKey("addr:housename"))
+        && !map.containsAnyKey("amenity", "craft", "leisure", "office", "shop", "tourism")){
```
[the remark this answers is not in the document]
westnordost: Off-putting, for what?  I didn't really get anyway why housenumbers of shops should not be shown. So what if there are multiple housenumbers being displayed with the same number (because they are in the same building).

So anyway, I agree that this could be dropped.

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
```diff
@@ -59,5 +60,20 @@ import de.westnordost.streetcomplete.R
     when (map["tourism"]) {
         "information" -> return R.drawable.ic_pin_information
     }
+    if (map.containsKey("addr:housenumber") || map.containsKey("addr:housename") ||
+            map.containsKey("addr:conscriptionnumber") || map.containsKey("addr:streetnumber")
+        && !map.containsAnyKey("amenity", "craft", "healthcare", "leisure", "office", "shop", "tourism")){
```
westnordost, line 65:
```suggestion
    if (map.containsAnyKey("addr:housenumber", "addr:housename", "addr:conscriptionnumber", "addr:streetnumber")
        && !map.containsAnyKey("amenity", "craft", "healthcare", "leisure", "office", "shop", "tourism")) {
```

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
```diff
@@ -59,5 +60,20 @@ import de.westnordost.streetcomplete.R
...
+        return R.drawable.ic_none
+    }
     return null
 }
+
+fun getTitle(map: Map<String, String>): String? {
+    return if (map["name"] != null || map["brand"] != null) {
+        map["name"] ?: map["brand"]
+    } else if (map["addr:conscriptionnumber"] != null && map["streetnumber"] != null) {
+        map["addr:conscriptionnumber"] + "/" + map["streetnumber"]
```
westnordost, line 75: 1. it's `addr:streetnumber` , not `streetnumber`
2. `addr:streetnumber` is optional

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
arrival-spring: 1. argh! Sorry! I did know that really
2. Added an extra option to the if

app/src/main/java/de/westnordost/streetcomplete/map/PinIcons.kt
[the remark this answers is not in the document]
arrival-spring: Without this check, in e.g. opening hours quest shops with house numbers look different to those without.
Shops without would have a pink dot, shops with wouldn't.
So actually I am in favour of keeping it, and I could add a comment to the code to that effect?

westnordost:
I found a way to generalize this a bit more

PR #3814, opened 2022-02-27
Correct the pluralisation of tactile paving

peternewman:


FloEdelmann (approved):
Looks good to me. I think we can merge this immediately (even though it contains string changes) because translations won't be impacted by this.

PR #3821, opened 2022-02-28
Survey incomplete sidewalk tagging

arrival-spring:
Some places only have `sidewalk:left` or `sidewalk:right`. I think this will usually imply that the other side has none, but I think it still counts as incomplete data (and would block #3735 - sidewalk surface quest).

This adds a bit more complexity than I'd hoped to the sidewalk filter, I wanted to do
`and !sidewalk and !sidewalk:both and !(sidewalk:left and sidewalk:right)`
but the NOT operator doesn't seem to work with brackets (it complained I'd closed too many brackets).

I wasn't about to get into rewriting the filter code, and maybe this filter is more readable than the above anyway.

Tested and works as expected, and all tests pass.

(I don't _think_ this will have merge issues with #3735 but I'm happy to sort them if it does)


westnordost:
> `and !sidewalk and !sidewalk:both and !(sidewalk:left and sidewalk:right)`

Would be wrong anyway. That filter would always return true if e.g. only `sidewalk:both` but not any other `sidewalk:*` was set

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
```diff
@@ -32,7 +32,14 @@ class AddSidewalk : OsmElementQuestType<SidewalkSides> {
...
           and motorroad != yes
-          and !sidewalk and !sidewalk:left and !sidewalk:right and !sidewalk:both
+          and !sidewalk and !sidewalk:both
+          and (
+            (sidewalk:left and !sidewalk:right)
+            or
+            (sidewalk:right and !sidewalk:left)
+            or
+            (!sidewalk:left and !sidewalk:right)
+          )
```
westnordost, line 42: Hmm, what about other invalid combinations? Like `sidewalk` is set at the same time as `sidewalk:*`? Or `sidewalk:both` is set at the same time as `sidewalk:left`?

Rather than making the filter expression ever more complex, maybe you could use the `SidewalkParser.kt` to validate this. 

I.e., do the same as for the `AddCycleway` quest. Would also be more consistent.

Also, this approach has one flaw anyway: It treats wrong `sidewalk`-taggings like "not tagged", but that means that if there is a `highway=footway` nearby, this gets filtered out again. How it is done in `AddCycleway` takes this into account - i.e. roads with invalid, old or incomplete cycleway tagging are always asked again, independent of whether there is a cycleway nearby that might be the seperately-mapped cycleway.

arrival-spring:
Now branched from #3735 - sidewalk surface quest

Thanks for the tip, I've now done the filters here in a similar way to `AddCycleway`, so any incomplete or overloaded tagging is now surveyed every time.

Runs as expected, and all tests pass.

westnordost:
Something is wrong, the diff also includes changes from your other PR even though it is merged now

arrival-spring:
> Something is wrong, the diff also includes changes from your other PR even though it is merged now

It seems that's a [bug in Github](https://stackoverflow.com/a/26986320)

I've sorted it now

app/src/main/java/de/westnordost/streetcomplete/osm/sidewalk/SidewalkParser.kt
```diff
@@ -1,23 +1,32 @@
...
+    if (sidewalk != null && altSidewalk != null) return LeftAndRightSidewalk(INVALID, INVALID)
+
+    // has sidewalk tagging, but not known
+    if (tags.keys.containsAny(KNOWN_SIDEWALK_KEYS) && sidewalk == null && altSidewalk == null) {
+        return LeftAndRightSidewalk(INVALID, INVALID)
+    }
+
+    // has valid sidewalk tagging of only one tagging style
+    if (sidewalk != null) return sidewalk
+    if (altSidewalk != null)return altSidewalk
```
westnordost, line 29:
```suggestion
    if (altSidewalk != null) return altSidewalk
```

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
```diff
@@ -72,31 +37,43 @@ class AddSidewalk : OsmElementQuestType<SidewalkSides> {
...
+                }
             }
         }
+
+        // Also include any roads with invalid (overloaded) or incomplete sidewalk tagging
+        val hasInvalidSidewalkTagging = eligibleRoads.filter { way ->
+            way.hasInvalidOrIncompleteSidewalkTags()
+        }
+
+        return roadsWithMissingSidewalks + hasInvalidSidewalkTagging
```
westnordost, line 76:
```suggestion
        val roadsWithInvalidSidewalkTags = eligibleRoads.filter { way ->
            way.hasInvalidOrIncompleteSidewalkTags()
        }

        return roadsWithMissingSidewalks + roadsWithInvalidSidewalkTags 
```
naming

westnordost (changes requested):
Looks good to me, thank you! I find it very important though to have a test for `SidewalkParser`

app/src/test/java/de/westnordost/streetcomplete/quests/AddSidewalkTest.kt
```diff
@@ -38,6 +41,27 @@ class AddSidewalkTest {
...
+        val road = way(tags = mapOf(
+            "highway" to "residential",
+            "sidewalk" to "left",
+            "sidewalk:right" to "yes"
+        ))
+        val mapData = TestMapDataWithGeometry(listOf(road))
+        assertEquals(1, questType.getApplicableElements(mapData).toList().size)
+        assertTrue(questType.isApplicableTo(road)!!)
+    }
+
```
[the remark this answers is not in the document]
arrival-spring: Agreed, done and done

arrival-spring:
Tests added (and fixed a couple of edge cases whilst at it).

When using only one of `sidewalk:left` and `sidewalk:right` `SidewalkParser` currently returns `INVALID` for the other side (rather than `null`).

I'm not sure if this matters, I didn't change it because it would mean adding a check for any side being `null` in `AddSidewalk` (in addition to checking for any side being `INVALID`.

westnordost (approved):

westnordost:
Cool, you even found some bugs with that! The tests are really easy to read, well done!

> When using only one of sidewalk:left and sidewalk:right SidewalkParser currently returns INVALID for the other side (rather than null).

Hmm, I think this is fair. It would be OK either way. The important part is that the app asks the user to complete the information if it is incomplete / ambiguous this way. So, whatever you think is best.

Issue #3833, opened 2022-03-03
Disable width quests for placement=transition


DerDings:
[placement=transition](https://wiki.openstreetmap.org/wiki/Tag:placement%3Dtransition) is used on a short piece of way joining two different pieces of road/cycleway. It's width can not easily be measured. So the with quest should be disabled for those pieces.

![Screenshot_20220303-204132_1.png](https://user-images.githubusercontent.com/42044179/156640981-2e50509f-b10a-4066-b8e9-7a8f17326290.png)
a `placement=transition` piece of road

![Screenshot_20220303-203725_1.png](https://user-images.githubusercontent.com/42044179/156641208-09375e7a-58bc-426a-8002-66a8484654ae.png)
a `placement=transition` piece of cycleway

westnordost:
First time I hear of that tag. But if it is supported by JOSM, it seems to be fine.

PR #3835, opened 2022-03-04
do not ask for width on ways with placement=transition

mnalis:
as the width can not easily be measured there.

Closes https://github.com/streetcomplete/StreetComplete/issues/3833

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddCyclewayWidth.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddCyclewayWidth.kt
@@ -30,6 +30,7 @@ class AddCyclewayWidth(
         )
         and area != yes
         and access !~ private|no
+        and placement != transition
     """
     override val changesetComment = "Determine cycleways width"
     override val wikiLink = "Key:width"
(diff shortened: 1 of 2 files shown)


westnordost:
I wonder, should any other things also not be asked on such ways? Lanes? Sidewalks? etc?

mnalis:
I've checked that [placement=transition](https://wiki.openstreetmap.org/wiki/Tag:placement%3Dtransition) tag is implemented in JOSM so that should be OK. 
I've tested on https://www.openstreetmap.org/way/161782303 and it seems to work (regular SC asks, one with this PR doesn't)

mnalis:
> I wonder, should any other things also not be asked on such ways? Lanes? Sidewalks? etc?

I guess sidewalk on one side is still possible (and skipping asking for it would impact pedestrian routing if just that part of the connection would miss tagged sidewalk).


So, maybe `lanes`. While it seems that that still may be specified (the way I've linked above have both `lanes` and `sidewalk` tagged, for example), it might be more troublesome for SC users that what it's worth (but I'm not a user of `lanes` so I wouldn't really know). 
Let me know if you'd want me do PR for lanes too.

matkoniecz:
`lanes` quest seems fitting, maybe also skip `surface` quests for cycleways?

westnordost:
surface for cycleways? In what situation? Transition between separate and not-separate cycleway mapping? Then the same should apply for sidewalks (separate / not separate)

matkoniecz:
> Transition between separate and not-separate cycleway mapping?

Or for transition/connector between road and sidewalk that is allowing cycling. Like https://www.openstreetmap.org/way/308773316 (here surface was clearly asphalt, but sometime it is quite confusing)

I am not really sure is it worth skipping in this case

PR #3842, opened 2022-03-04
do not ask for lanes on ways with placement=transition

mnalis:
... as it can be hard to answer.

As mentioned in https://github.com/streetcomplete/StreetComplete/pull/3835#issuecomment-1058971788


mnalis:
(tested on https://www.openstreetmap.org/way/326136157 and it seems to work - display in original SC, does not display in this PR)

PR #3844, opened 2022-03-05
Incorrect tactile paving option for crossings

### LINKED ISSUE #3813: Allow tagging tactile_paving=incorrect for when only one side of a crossing has it
**Use case**
It affects also crossing ques, but is `no` really correct when only one side has paving?

_Originally posted by @matkoniecz in https://github.com/streetcomplete/StreetComplete/issues/3762#issuecomment-1039607154_

Probably `incorrect` is the right thing there. The wiki has "or if only one sloped kerb at a crossing has tactile paving" as an example of `incorrect`.

_Originally posted by @arrival-spring in https://github.com/streetcomplete/StreetComplete/issues/3762#issuecomment-1039612736_

OT, but I'd appreciate some clarity/a new option for this (I normally just tag no as it's not yes). Also on a crossing with an island, if its only on either side but not on the island which is right?

_Originally posted by @peternewman in https://github.com/streetcomplete/StreetComplete/issues/3762#issuecomment-1040544178_

There was a bit of a discussion here: https://github.com/streetcomplete/StreetComplete/issues/2738

I'd support having an other answers button of "Tactile paving only at the top or only at the bottom" and similarly for the crossing quest "Tactile paving only on one side" which would tag `incorrect` after a confirmation.

_Originally posted by @arrival-spring in https://github.com/streetcomplete/StreetComplete/issues/3762#issuecomment-1040650287_

See also #2738 .

I appreciate incorrect is low usage currently (0.11%) but I don't know about others but I'd certainly have been tagging no when it's only on one side (given "Does this crosswalk have tactile pavings on both sides?") although I'm not sure how many of those I've had. 

As mentioned I've also been tagging no if it's on the kerbs but not on the island as it's not on both sides of the two crossings, but I appreciate that probably needs a more general tagging discussion to find out what is the "correct" answer, or asking a blind person. @matkoniecz do you have contacts/feedback from your tactile maps?

**Proposed Solution**
Add an other answer for only on one side of the crossing

arrival-spring:
Fixes #3813 

As the same form is used for all tactile paving quests, showing the incorrect option only for nodes with `highway=crossing` or `highway=traffic_signals` (in the same way as the steps answer option in path quests) seemed like the best way of doing it.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingAnswer.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingAnswer.kt
@@ -1,0 +1,7 @@
+package de.westnordost.streetcomplete.quests.tactile_paving
+
+enum class TactilePavingAnswer(val osmValue: String) {
+    YES("yes"),
+    NO("no"),
+    INCORRECT("incorrect")
+}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingCrosswalkForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingCrosswalkForm.kt
@@ -3,11 +3,11 @@ package de.westnordost.streetcomplete.quests.tactile_paving
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.quests.AbstractQuestAnswerFragment
 import de.westnordost.streetcomplete.quests.AnswerItem
-import de.westnordost.streetcomplete.quests.tactile_paving.TactilePavingAnswer.YES
-import de.westnordost.streetcomplete.quests.tactile_paving.TactilePavingAnswer.NO
-import de.westnordost.streetcomplete.quests.tactile_paving.TactilePavingAnswer.INCORRECT
+import de.westnordost.streetcomplete.quests.tactile_paving.TactilePavingCrosswalkAnswer.YES
+import de.westnordost.streetcomplete.quests.tactile_paving.TactilePavingCrosswalkAnswer.NO
+import de.westnordost.streetcomplete.quests.tactile_paving.TactilePavingCrosswalkAnswer.INCORRECT
 
-class TactilePavingCrosswalkForm : AbstractQuestAnswerFragment<TactilePavingAnswer>() {
+class TactilePavingCrosswalkForm : AbstractQuestAnswerFragment<TactilePavingCrosswalkAnswer>() {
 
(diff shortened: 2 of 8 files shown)


matkoniecz (approved):
tested, works fine (is not appearing for kerb quests, as expected)

And big thanks for implementing it!

matkoniecz:
And big thanks for implementing it!

westnordost:
I was about to comment about that, because answering "on one side only" does not make sense when the app asks about just one kerb or just a bus stop. It is confusing to have such an answer option when it is not used.

Please instead just create a `TactilePavingCrosswalkForm` and leave the old `TactilePavingForm` as is. It may actually be less code this way.

arrival-spring:
Fair enough

Tested and working as expected

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingCrosswalkForm.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    override val contentLayoutResId = R.layout.quest_tactile_paving
+
+    override val otherAnswers get() = listOfNotNull(
+        createIncorrectAnswerItem()
+    )
+
+    private fun createIncorrectAnswerItem(): AnswerItem {
+        return AnswerItem(R.string.quest_tactilePaving_incorrect) { applyAnswer(INCORRECT) }
+    }
```
westnordost, line 20:
```suggestion
    override val otherAnswers get() = listOf(
        AnswerItem(R.string.quest_tactilePaving_incorrect) { applyAnswer(INCORRECT) }
    )
```

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingAnswer.kt
```diff
@@ -0,0 +1,7 @@
+package de.westnordost.streetcomplete.quests.tactile_paving
+
+enum class TactilePavingAnswer(val osmValue: String) {
```
westnordost, line 3: It would be consistent if this was named `TactilePavingCrosswalkAnswer`. Not a merge blocker though.

westnordost (approved):
It is actually less code than before :-)

Thank you for doing this change!

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/TactilePavingAnswer.kt
arrival-spring: Done

arrival-spring:
Thanks for doing the shortening, could have sworn I'd already done that!

Issue #3817, opened 2022-02-27
Bollard type quest should allow retag barrier type


HolgerJeromin:
I just had a "barrier bollard type quest" but had a boom barrier in front of me.
So imo it would be nice if there were a menu entry:
Other/is different barrier
To trigger the existing barrier quest.

arrival-spring:
That would seem to be consistent with other quests, to have such an option in 'other answers'.

kmpoppe:
By now, using @matkoniecz 's ent8r request, there are 44 notes where people couldn't answer this quest.

So maybe it's worth the effort implementing the "other option" now?

K

matkoniecz:
It seems a good idea to me.

mnalis:
Sooooo, does all that mean someone is volunteering to work on it? :smile: 

(If nobody else wants to do it, I can probably give it a try at weekend...)

matkoniecz:
@mnalis Feel free to implement it!

(I plan to do something today with one of cases where I volunteered: https://github.com/streetcomplete/StreetComplete/issues/assigned/matkoniecz )

mnalis:
Implemented. I'll also make another PR for similar option for `cycle barrier type` quest (with `not a cycle barrier` other answer)

PR #3845, opened 2022-03-05
Add other answer "not a bollard" to bollard type quest

mnalis:
When users chooses "other answers" and "not a bollard", the former `barrier=bollard` is instead retagged with `barrier=yes`, so regular `barrier type` quest can pop up. 

(Implemented in mostly the same way as `bench_backrest` quest with `it's a picnic table` answer.)

Tested and it seems to work correctly.

Closes https://github.com/streetcomplete/StreetComplete/issues/3817


westnordost:
I was about to ask in what situation another barrier would be mistagged as a bollard, but it looks like according to #3817, this happens to quite often. So, thanks for implementing!

app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/AddBollardType.kt
```diff
@@ -53,6 +54,11 @@ class AddBollardType : OsmElementQuestType<BollardType> {
     override fun createForm() = AddBollardTypeForm()
 
     override fun applyAnswerTo(answer: BollardType, tags: Tags, timestampEdited: Long) {
-        tags["bollard"] = answer.osmValue
+        when (answer) {
+            NOT_BOLLARD -> {
+                tags["barrier"] = "yes"
+            }
```
westnordost, line 60:
```suggestion
            NOT_BOLLARD -> tags["barrier"] = "yes"
```

westnordost (approved):

Issue #3563, opened 2021-12-04
Add answer option "It's another type of barrier" to "More Answers..."-section of Cycle-Barrier-Type and Bollard-Type quest


wielandb:
**Use case**
Since the Cycle-Barrier-Quest came out, I noticed some cycle-barriers in my area are not actually cycle-barriers but bollards or something else. When I came across some gates today that were tagged as bollards, I got the idea that the re-tagging could be done directly in StreetComplete as we already have a quest to determine barrier types.

**Proposed Solution**
Add the answer option "It's another type of barrier" to the "More Answers..." section of the quests "What kind of cycle barrier is this?" and "What kind of bollard is this?". When chosen, it converts `barrier=cycle_barrier` and `barrier=bollard` back to `barrier=yes` (similar in the way that a building gets "converted" back to `building=yes` when you answer that the building type is wrong in the housenumber quest), which would promt the Barrier-Type-Quest to show up, allowing quick correction of false bollards and cycle_barriers.

Of course, I would apprechiate feedback if falsely tagged `cycle_barrier`s and `bollard`s are a occurence for other users too to see if this addition is worth the effort.

ygra:
Haven't seen wrongly-tagged barriers here in the vicinity yet, but I could imagine it happens sometimes. And in those cases an answer option that spells out what's wrong instead of having to write a note can be useful.

matkoniecz:
Interesting idea, I will look at how often this quest results in notes.

For reference, it was #3487 that added it

https://ent8r.github.io/NotesReview/?view=map&query=Unable%20to%20answer%20%22What%20type%20of%20bicycle%20barrier%20is%20this%3F%22&status=null&before=null indicates that so far just 8 notes were opened for this quest.

So I will close this one without implementing.

PR #3846, opened 2022-03-05
Add other answer "not a cycle barrier" to "cycle barrier type" quest

mnalis:
When users chooses "other answers" and "not a cycle barrier", the former `barrier=cycle_barrier` is instead retagged with `barrier=yes`, so regular barrier type quest can pop up.

(Implemented in the same way as `bollard type quest` / `not a bollard` PR https://github.com/streetcomplete/StreetComplete/pull/3845)

Tested and it seems to work correctly.

Related to https://github.com/streetcomplete/StreetComplete/issues/3817, it was originally reported in https://github.com/streetcomplete/StreetComplete/issues/3563


app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierType.kt
```diff
@@ -22,6 +23,11 @@ class AddBicycleBarrierType : OsmFilterQuestType<BicycleBarrierType>() {
     override fun createForm() = AddBicycleBarrierTypeForm()
 
     override fun applyAnswerTo(answer: BicycleBarrierType, tags: Tags, timestampEdited: Long) {
-        tags["cycle_barrier"] = answer.osmValue
+        when (answer) {
+            NOT_CYCLE_BARRIER -> {
+                tags["barrier"] = "yes"
+            }
```
westnordost, line 29:
```suggestion
            NOT_CYCLE_BARRIER -> tags["barrier"] = "yes"
```

westnordost (approved):

PR #3847, opened 2022-03-05
show nearby steps in steps-related quests

mnalis:
Since steps can often be clustered together, it is useful to show nearby `highway=steps` ways when one is solving steps-related quests. This PR enables their highlighting. I've tested that it works.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/handrail/AddHandrail.kt
@@ -33,6 +33,9 @@ class AddHandrail : OsmFilterQuestType<Boolean>() {
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_handrail_title
 
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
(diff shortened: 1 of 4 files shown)


matkoniecz:
steps are shown already with map style, right?

mnalis:
They are; although I find them easy to miss, especially when they are small (e.g.micromapped in several small segments).

westnordost:
I think this would rather be confusing to users.

But well, do you have an example location where steps would not be displayed properly on the map and where this would help?

mnalis:
For example, comparison of vanilla vs my fork version at https://www.openstreetmap.org/way/562023209

small_Screenshot_20220306_235145_de westnordost streetcomplete small_Screenshot_20220306_234729_de westnordost streetcomplete mn debug

Do note that my fork also has other changes visible in the right picture (`hide` button, semi-transparent buildings, some different colors), but hopefully it shows main idea of purple lines for nearby steps highlighted well enough.

Anyway, I personally find it useful, and will keep it in my fork, so no hard feelings if you decide you don't want it in mainline.

westnordost:
Alright, then I'll close this mostly for consistency as other way-based quests also don't show nearby similar ways.

PR #3853, opened 2022-03-07
Use sealed interfaces instead of "hacky" enum items

FloEdelmann:
In #3845 and #3846, the "other barrier type" answers were implemented as new enum items in `BollardType` and `BicycleBarrierType`. That's a bit hacky though, as it requires specifying a dummy `osmValue`. Also, it's not consistent with other quest answer types.

In this PR, the implementation is changed to introduce new sealed interfaces `BollardTypeAnswer` and `BicycleBarrierTypeAnswer`, from which the old enum classes extend. Also, new single objects (without an invalid `osmValue`) now extend from the sealed interfaces.

After some reading[^1], I still don't quite get the difference between sealed classes and sealed interfaces. I think that in this case, both should work the same. But since sealed interfaces are a bit more flexible and save two characters (`()`) per inheritance definition, I decided to use `sealed interface`. Also, I converted all the `*Answer` sealed classes to sealed interfaces as well, to be consistent.

If that was the wrong decision, and `BollardTypeAnswer` and `BicycleBarrierTypeAnswer` should be sealed classes instead, please let me know, and I will change it.

CC @mnalis

[^1]: nice blog post here: https://quickbirdstudios.com/blog/sealed-interfaces-kotlin/

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/BollardType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/BollardType.kt
@@ -1,10 +1,13 @@
 package de.westnordost.streetcomplete.quests.bollard_type
 
-enum class BollardType(val osmValue: String) {
+sealed interface BollardTypeAnswer
+
+enum class BollardType(val osmValue: String) : BollardTypeAnswer {
     RISING("rising"),
     REMOVABLE("removable"),
     FOLDABLE("foldable"),
     FLEXIBLE("flexible"),
     FIXED("fixed"),
-    NOT_BOLLARD("not_bollard"),
 }
+
+object DifferentBarrierType : BollardTypeAnswer
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/BollardType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/BollardType.kt
@@ -10,4 +10,4 @@ enum class BollardType(val osmValue: String) : BollardTypeAnswer {
     FIXED("fixed"),
 }
 
-object DifferentBarrierType : BollardTypeAnswer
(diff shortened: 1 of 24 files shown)


westnordost:
> After some reading, I still don't quite get the difference between sealed classes and sealed interfaces. I think that in this case, both should work the same. But since sealed interfaces are a bit more flexible and save two characters (()) per inheritance definition, I decided to use sealed interface. Also, I converted all the *Answer sealed classes to sealed interfaces as well, to be consistent.

Sealed classes existed before. Sealed interfaces were supported later in Kotlin. The reason why most of these answers are sealed classes is because when they were created, sealed interfaces did not exist yet.
The way the sealed classes/interfaces are used in this app, interfaces seem to be more fitting. But as you say, there is not really a difference between the two.

app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/BollardType.kt
```diff
@@ -1,10 +1,13 @@
...
+enum class BollardType(val osmValue: String) : BollardTypeAnswer {
     RISING("rising"),
     REMOVABLE("removable"),
     FOLDABLE("foldable"),
     FLEXIBLE("flexible"),
     FIXED("fixed"),
-    NOT_BOLLARD("not_bollard"),
 }
+
+object DifferentBarrierType : BollardTypeAnswer
```
westnordost, line 13: IMO this should have something like `NotBollardBut...` or similar in its name. After all, this is a public singleton, not an enum constant. Enum constants are accessed like BollardType.FOLDABLE. Public singletons are directly accessed. Hence, we should avoid using the exact same names for two different singletons (i.e. there is also `object DifferentBarrierType : BicycleBarrierTypeAnswer`)

westnordost (changes requested):

westnordost (approved):

PR #3855, opened 2022-03-08
Add `amenity=food_court` to `isShopExpressionFragment`

FloEdelmann:
The following quests are asked for `amenity=food_court`:

* `AddAcceptsCash`
* `AddOpeningHours`
* `AddPlaceName`
* `AddSeating`
* `AddSmoking`
* `AddWheelchairAccessBusiness`

All these quests highlight all elements matching `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION`, but that didn't contain `amenity=food_court` yet.

I'm opening this as a PR because the comment above this function says that https://github.com/mnalis/StreetComplete-taginfo-categorize/blob/master/Makefile should be re-run afterwards. So @mnalis could you run the script, please? Should this PR be merged first or afterwards?

*(Noticed during review of [`f292530` (#3758)](https://github.com/streetcomplete/StreetComplete/pull/3758/commits/f292530131e719801aa4a0f9b1a9eee921ac842e))*

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/meta/OsmTaggings.kt
@@ -133,11 +133,11 @@ fun isShopExpressionFragment(prefix: String? = null): String {
         or """ +
         mapOf(
             "amenity" to arrayOf(
-                "restaurant", "cafe", "ice_cream", "fast_food", "bar", "pub", "biergarten", "nightclub",
+                "restaurant", "cafe", "ice_cream", "fast_food", "bar", "pub", "biergarten", "food_court",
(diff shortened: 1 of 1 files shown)


andrewharvey:
For `AddAcceptsCash` wouldn't that vary based on each vendor that serves the food court? The rest look good, often they will be noname but sometimes will have their own name.

FloEdelmann:
Note that this PR doesn't change the logic where the quest is asked, just what other elements are highlighted.

andrewharvey:
Sorry! I should have read the full text and not just skimmed the first two lines.

westnordost:
No, a food court is not in the typical area a shop is in. The "isShopTypeExpression" is for "Ladenflächen". A food court is not in a typical Ladenfläche, it's more like an (indoor) "Marktfläche".

PR #3860, opened 2022-03-09
Show place name for seating quest

### LINKED ISSUE #3857: Show place name for seating quest
Currently the new seating quest asks "What kind of seating does this place have?"

In many places showing place name would be very helpful, and it would be consistent with other quests.

Helium314:
fixes #3857 

The original string name without name is a bit misleading :`quest_seating_name_title`, if translations haven't yet started maybe it should be changed to `quest_seating_no_name_title`


matkoniecz:
It is in translations already: https://poeditor.com/projects/po_edit?order=&filter=&per_page=20&id=97843&id_language=127&search=quest_seating_name_title

but I would consider changing it anyway

I could rescue it for languages translated (if I have access to POEditor exports, in regular interface I would need to manually export each language)

Helium314:
Alright, so if that's ok I would change it to `quest_seating_no_name_title` and `quest_seating_name_title` (consistent with opening hours)

westnordost:
Due to the largely positive feedback on #3862, I am going to try this first which makes this PR and the accompanying issue obsolete.

Issue #3856, opened 2022-03-08
Don't show smoking quest without outdoor seating in some countries


Helium314:
In many countries smoking is forbidden inside restaurants/pubs/..., so asking in those countries is pointless if there is no outdoor seating, as the answer will always be no.
Well, except if the 'outside' answer means "you can just leave the pub and smoke on the street"... but the street does not belong to the pub, so smoking on the street shouldn't be tagged on the pub (not sure if this is the current interpretation by the smoking quest, because `outside` answer is still shown if `outside_seating=no`). 

The quest could even depend on answering the new seating quest first, i.e. require both `indoor_seating` and `outdoor_seating` (or `leisure=outdoor_seating` of course). So first the user answers the seating quest, and the smoking quest is only shown if there is outdoor seating (again: in countries with indoor smoking ban).

matkoniecz:
In which countries this quest will be asked (after enabling it) and should not be asked? Can you give some examples?

Helium314:
> In which countries this quest will be asked (after enabling it) and should not be asked? Can you give some examples?

Currently the quest will be asked in all countries, and I don't know any countries where it should be completely disabled.
But e.g. in Ireland smoking is not allowed in restaurants, so if a restaurant in Ireland has no outdoor seating, I would expect that smoking is not allowed, and there is no need to explicitly tag it.

Helium314:
Some quick and very incomplete check of the [Wikipedia list](https://en.wikipedia.org/wiki/List_of_smoking_bans):
No indoor smoking in Austria, Czech Republic, France, Hungary, Netherlands (also Greece, but not enforced).
In Sweden smoking isn't even allowed in outdoor seating areas.

andrewharvey:
I think most Australian states have or soon will ban smoking in outdoor dining areas, so no point asking it at all. New Zealand is really leading the way by phasing in a blanket ban on all smoking.

westnordost:
I was under the impression that this has already been discussed in the #539 ticket and the outcome was (that it's complicated and) the quest should (for that reason) be disabled by default? Same as e.g. Kosher quest which also doesn't make sense in some territories.

mnalis:
[...]

PR #3865, opened 2022-03-10
Smoking outside tuning

mnalis:
This PR tries to address superfluous `OUTSIDE` answer when it should be hidden, as reported in https://github.com/streetcomplete/StreetComplete/issues/3856

- do not show quest at all if there are neither indoor nor outdoor seating
- hide `OUTSIDE` answer if it is known that there is no outside seating

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
@@ -28,6 +28,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
              )
          )
          and takeaway != only
+         and !(indoor_seating = no and outdoor_seating = no)
          and (!smoking or smoking older today -8 years)
     """
     override val changesetComment = "Add smoking status"
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt
@@ -15,10 +15,11 @@ class SmokingAllowedAnswerForm : AListQuestAnswerFragment<SmokingAllowed>() {
         val isAlreadyOutdoor =
             tags["leisure"] == "outdoor_seating" || tags["amenity"] == "biergarten" ||
             (tags["outdoor_seating"] == "yes" && tags["indoor_seating"] == "no")
+        val noOutdoorSeating = tags["outdoor_seating"] == "no"
 
         return listOfNotNull(
             TextItem(NO, R.string.quest_smoking_no),
-            if (isAlreadyOutdoor) null else TextItem(OUTSIDE, R.string.quest_smoking_outside),
+            if (isAlreadyOutdoor || noOutdoorSeating) null else TextItem(OUTSIDE, R.string.quest_smoking_outside),
             TextItem(SEPARATED, R.string.quest_smoking_separated),
             TextItem(YES, R.string.quest_smoking_yes),
         )


peternewman:
>     * hide `OUTSIDE` answer if it is known that there is no outside seating

In the UK at least it's quite common, particular outside nightclubs, to have a small fenced area where people can smoke without having to leave the area the club controls (and pay for a new ticket etc). This fenced area is often on the pavement and may not have seating, which might break your current logic here.

See some average to rubbish photos of the like here:
https://www.vice.com/en/article/z3xwa4/this-is-what-englands-nightclub-smoking-areas-looked-like-last-night

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -28,6 +28,7 @@ class AddSmoking : OsmFilterQuestType<SmokingAllowed>() {
              )
          )
          and takeaway != only
+         and !(indoor_seating = no and outdoor_seating = no)
```
FloEdelmann, line 31: See [#3821](https://togithub.com/streetcomplete/StreetComplete/pull/3821):

> I wanted to do
> `and !sidewalk and !sidewalk:both and !(sidewalk:left and sidewalk:right)`
> but the NOT operator doesn't seem to work with brackets (it complained I'd closed too many brackets).

FloEdelmann: You should be able to rewrite it with [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) though:

```suggestion
         and (indoor_seating != no or outdoor_seating != no)
```

mnalis: Maybe, but I'm uncomfortable because I don't know how that parsing code handles null (undef) values, that's why I written it in what seems to me the "safe" way...

Also I find it much easier to read (given the fact that tags are not boolean `yes` / `no` values, but can have many values which differ from `no`, or no value at all).

mnalis: But interesting issue #3821, has it been **known** that `!` does not work with brackets, or was that just a _hypothesis_? Because it didn't complain in my case. I plan to do some checking how it actually behaves tonight to make sure.

mnalis:
> In the UK at least it's quite common, particular outside nightclubs, to have a small fenced area where people can smoke without having to leave the area the club controls

thanks @peternewman that is quite informative! I'm now undecided, should I just add the exception for `nightclub` (and likely `stripclub` as it might behave similarly?), or do pubs/cafes/restaurants also might have such arrangements and I should thus show `OUTSIDE` answer regardless of `outdoor_seating=no` ?

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
peternewman: When you work it out, it would be good to add some tests for that (if they don't exist already), to ensure it continues to behave in the same way in case this or something else is making use of that behaviour.

peternewman:
> thanks @peternewman that is quite informative! I'm now undecided, should I just add the exception for `nightclub` (and likely `stripclub` as it might behave similarly?), or do pubs/cafes/restaurants also might have such arrangements and I should thus show `OUTSIDE` answer regardless of `outdoor_seating=no` ?

Maybe pubs too as it might apply to some of the more clubby/youth ones, its more about having to queue for entry, possibly pay and/or have ID checks.

I don't think cafes or restaurants will have the queue/pay/ID issues, so probably won't bother. They're also more likely (possibly along with some pubs) to have outdoor seating at which point that bit kicks in too.

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
westnordost: `!` doesn't work with parenthesis, parsing for this is simply not implemented. It was known to me at least :-)
I once started to implement it because I thought it would be "cleaner" to also support parsing this despite it not being used that far but I stopped because it turned out to be more complex than I anticipated.

`indoor_seating = no` evaluates to true if `indoor_seating` is either not set or not `no`.

mnalis: thanks, this should be incorporated now... Now to testing if it actually works like it should

app/src/test/java/de/westnordost/streetcomplete/quests/smoking/AddSmokingTest.kt
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+
+    @Test fun `applicable to bakery with both seatings`() {
+        assertTrue(questType.isApplicableTo(node(tags = mapOf(
+            "shop" to "bakery",
+            "indoor_seating" to "bar_table",
+            "outdoor_seating" to "yes",
+        ))))
+    }
```
FloEdelmann, line 123: I'd combine similar test cases to make it more readable:

```suggestion
    /* we assume that seating is not present if not indicated for bakery and similar */
    @Test fun `not applicable to bakery without indicated seating`() {
        assertFalse(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
        ))))

        assertFalse(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "indoor_seating" to "no",
        ))))

        assertFalse(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "outdoor_seating" to "no",
        ))))
    }

    @Test fun `not applicable to bakery without any seating`() {
        assertFalse(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "indoor_seating" to "no",
            "outdoor_seating" to "no",
        ))))
    }

    @Test fun `applicable to bakery with any indicated seating`() {
        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "indoor_seating" to "bar_table",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "indoor_seating" to "yes",
            "outdoor_seating" to "no",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "outdoor_seating" to "terrace",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "indoor_seating" to "no",
            "outdoor_seating" to "yes",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "shop" to "bakery",
            "indoor_seating" to "bar_table",
            "outdoor_seating" to "yes",
        ))))
    }
```

app/src/test/java/de/westnordost/streetcomplete/quests/smoking/AddSmokingTest.kt
```diff
@@ -0,0 +1,220 @@
...
+
+    /* nighclubs etc. may have outdoor smoking areas even if no seating is present */
+    @Test fun `applicable to nightclub without any seating`() {
+        assertFalse(questType.isApplicableTo(node(tags = mapOf(
```
FloEdelmann, line 127: The comment above suggests that this should be `assertTrue`.

```suggestion
        assertTrue(questType.isApplicableTo(node(tags = mapOf(
```

app/src/test/java/de/westnordost/streetcomplete/quests/smoking/AddSmokingTest.kt
```diff
@@ -0,0 +1,220 @@
...
+    @Test fun `applicable to cafe with both seatings`() {
+        assertTrue(questType.isApplicableTo(node(tags = mapOf(
+            "amenity" to "cafe",
+            "indoor_seating" to "bar_table",
+            "outdoor_seating" to "patio",
+        ))))
+    }
+
+
+
```
FloEdelmann, line 211: Again combining similar tests:

```suggestion
    /* we assume that seating is present if not indicated for cafe and similar */
    @Test fun `applicable to cafe without indicated seating`() {
        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "indoor_seating" to "no",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "outdoor_seating" to "no",
        ))))
    }

    @Test fun `not applicable to cafe without any seating`() {
        assertFalse(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "indoor_seating" to "no",
            "outdoor_seating" to "no",
        ))))
    }

    @Test fun `applicable to cafe with any indicated seating`() {
        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "indoor_seating" to "bar_table",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "indoor_seating" to "yes",
            "outdoor_seating" to "no",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "outdoor_seating" to "terrace",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "indoor_seating" to "no",
            "outdoor_seating" to "yes",
        ))))

        assertTrue(questType.isApplicableTo(node(tags = mapOf(
            "amenity" to "cafe",
            "indoor_seating" to "bar_table",
            "outdoor_seating" to "patio",
        ))))
    }
```

FloEdelmann (changes requested):

mnalis:
Thanks for suggestions, @FloEdelmann! Do you know how/when those tests are actually run? 
I'd like to run them to verify it it works, before moving this out of draft.

FloEdelmann:
The tests are not run at all on GitHub Actions, they have to be run locally: Either via Android Studio, or in the commandline:

```
./gradlew :app:cleanTestDebugUnitTest :app:testDebugUnitTest --tests "de.westnordost.streetcomplete.*"
```

mnalis:
Thanks @FloEdelmann , I've added that command (in my fork GitHub workflow) to run tests, and it passes the checks now.

One more question: do you know if it is possible to check whether some option is shown in the list of answers or not?

Smoking quest sometimes offers `OUTSIDE` answer (and sometimes not) depending on various conditions (i.e. https://github.com/mnalis/StreetComplete/blob/smoking-outside-fix/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/SmokingAllowedAnswerForm.kt#L25), so I'd like to check that too if it is (easily) possible.

FloEdelmann:
> do you know if it is possible to check whether some option is shown in the list of answers or not?

Unfortunately, I don't know if that's possible here. Maybe only in the instrumented tests that have access to the Android runtime.

mnalis:
Thanks @FloEdelmann, I though that it might be more complicated... Then this is ready for review.

app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt
```diff
@@ -13,12 +13,23 @@ import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
...
+ - some places are eligible even if they don't have any seating (like nightclub)
+ - some places are eligible only if they are explicitely marked to have seating
+   as otherwise we assume they don't provide seating (like bakery, wine shop...)
+*/
     override val elementFilter = """
          nodes, ways, relations with
          (
-             amenity ~ bar|cafe|pub|biergarten|restaurant|food_court|nightclub|stripclub
+             amenity ~ bar|cafe|biergarten|restaurant|food_court
+             and (indoor_seating != no or outdoor_seating != no)
```
westnordost, line 30:
```suggestion
             amenity ~ bar|cafe|biergarten|restaurant|food_court and (indoor_seating != no or outdoor_seating != no)
```
mixing and and or while legal may be confusing for readers

westnordost (approved):
Looks good!

PR #3870, opened 2022-03-12
show name + location above quest title, simplify quest titles

westnordost:
See #3862

Screenshot_20220312-171402_StreetComplete Dev
Screenshot_20220312-171124_StreetComplete Dev
Screenshot_20220312-171350_StreetComplete Dev

Also, reorder all the remaining strings for better overview etc and add comments for certain strings (because IIRC they are displayed in POEditor and can be very useful for translators)


matkoniecz:
It still seems to be that it would be more natural to have this info below main question, not on top of it.

app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
```diff
@@ -0,0 +1,116 @@
...
+}
+
+private fun Resources.getHouseNumberLabel(tags: Map<String, String>): String? {
+    val houseName = tags["addr:housename"]
+    val conscriptionNumber = tags["addr:conscriptionnumber"]
+    val streetNumber = tags["addr:streetnumber"]
+    val houseNumber = tags["addr:housenumber"]
+
+    if (houseName != null) {
+        return getString(R.string.at_housename, houseName.inItalics())
```
matkoniecz, line 101: I would consider showing also numbers if present...

(I will add it if I run into such case)

app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
```diff
@@ -0,0 +1,116 @@
...
+       can not be seen on any sign. */
+    val floor = tags["addr:floor"] ?: tags["level:ref"]
+    if (floor != null) {
+        return getString(R.string.on_floor, floor)
+    }
+    val level = tags["level"]
+    if (level != null) {
+        return getString(R.string.on_level, level)
+    }
+    if (tags["tunnel"] == "yes" || tags["location"] == "underground") {
```
matkoniecz, line 88:
```suggestion
    if (tags["tunnel"] == "yes" || tags["tunnel"] == "culvert" || tags["location"] == "underground") {
```
Probably no quests affected right now but I would add it for completeness.

app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
westnordost: This was deliberate. Showing the housenumber (and level number) was introduced for better context. Before similar elements around it were highlihgted. Now, it is less useful. If the thing in question actually has a name (additionally to a housenumber), then it is much more helpful to show the name instead of the housenumber. Many shops have addresses tagged. But unlike for the name, there will not be (if at all) a big sign for the housenumber.

I removed it because
- with housenumber + name + feature name it gets really busy
- The more text there is, the more often it will not be read/ignored
- Also, oftentimes the "hint text" will take up two lines then

app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedHighwayConstruction.kt
```diff
@@ -19,15 +19,17 @@ class MarkCompletedHighwayConstruction : OsmFilterQuestType<CompletedConstructio
...
     override fun getTitle(tags: Map<String, String>): Int {
         val isRoad = ALL_ROADS.contains(tags["construction"])
         val isCycleway = tags["construction"] == "cycleway"
         val isFootway = tags["construction"] == "footway"
         val isSteps = tags["construction"] == "steps"
-
+        /* Alternative could be "Is this construction finished?" and just display the feature name
+        *  (e.g. "cycleway under construction") of the element above, but there are no iD presets
+        *  for "highway=construction + construction=*" so such road would just be named
+        *  "Road closed". Hence, keeping this (for now). */
```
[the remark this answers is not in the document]
matkoniecz: Maybe - there is https://github.com/openstreetmap/id-tagging-schema/issues/10 already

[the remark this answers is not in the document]
westnordost: From my end: No, I am not sure if this would be a meaningful preset.

app/src/androidTest/java/de/westnordost/streetcomplete/quests/existence/CheckExistenceLabelTest.kt
```diff
@@ -32,62 +30,60 @@ class CheckExistenceTitleTest {
...
     // https://github.com/streetcomplete/StreetComplete/issues/2840#issuecomment-831245075
-    @Test fun issue2840() {
-        val schoki = getQuestTitle(mapOf(
+    @Test fun schoki() {
+        assertEquals("Schoko Lädchen 3680 (Vending Machine)", getQuestLabel(mapOf(
             "amenity" to "vending_machine",
             "ref" to "3680",
             "operator" to "Schoko Lädchen",
-        ))
-        assertEquals(schoki, "Is Schoko Lädchen 3680 (Vending Machine) still here?")
```
FloEdelmann, line 86: I've updated the tests in here to check the label instead of the quest title (that is now always the same), and they still pass :+1:

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/AddBusStopName.kt
```diff
@@ -24,11 +24,7 @@ class AddBusStopName : OsmFilterQuestType<BusStopNameAnswer>() {
     override val icon = R.drawable.ic_quest_bus_stop_name
     override val questTypeAchievements = listOf(PEDESTRIAN)
 
-    override fun getTitle(tags: Map<String, String>) =
-        if (tags["tram"] == "yes")
-            R.string.quest_tramStopName_title
-        else
-            R.string.quest_busStopName_title
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_busStopName_title2
```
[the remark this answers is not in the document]
FloEdelmann: Ah, I see that *Bus Stop* is shown in the label, so this is fine then.

app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+    // if both name and feature name are available, put the feature name in parentheses
+    if (nameLabel != null) {
+        result += nameLabel
+        if (featureName != null) result += "($featureName)"
+    } else {
+        if (featureName != null) result += "$featureName"
+    }
+
+    return if (result.isEmpty()) null else result.joinToString(" ").parseAsHtml()
```
FloEdelmann, line 37: With the label below the title, this needs a bit more logic to show/hide the colon:

grafik

e.g. for https://www.openstreetmap.org/way/132149975

FloEdelmann (commented):
I really like the consistency of the label being shown in (nearly?) all quests and always showing relevant context information!

app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
```diff
@@ -0,0 +1,116 @@
...
+    val name = tags["name"]
+    val brand = tags["brand"]
+    val ref = tags["ref"]
+    val operator = tags["operator"]
+
+    return name
+        ?: brand
+        ?: if (ref != null && operator != null) "$operator $ref" else null
+        ?: operator
+        ?: ref
```
FloEdelmann, line 73: Do you think it's worth to check whether it's a reference number (e.g. `if (tags["emergency"] == fire_hydrant)`) and show "Reference number %s" in this case?

e.g. https://www.openstreetmap.org/node/6272420812:
grafik

But this would likely lead to questions about more and more elaborate logic for this.

westnordost: Hmm, well I suppose there is the number "1" engraved on that fire hydrant right? So if someone looks at the fire hydrant, he will see that the "1" is correct

westnordost: I acknowledge that it looks odd, but as you say, I wouldn't want to introduce special treatments for certain element types

app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
westnordost: ok

app/src/main/res/values/strings.xml
```diff
@@ -1,387 +1,1023 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- This is the source file for the app's strings in en-US.
```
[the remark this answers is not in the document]
FloEdelmann: For reference: With the generated quest list, I compared the old and new quest titles:

```diff
  AddAcceptsCash:
- Is cash payment accepted here?
+ Does […] accept cash payment?

  AddAirConditioning:
- Is this place air conditioned?
+ Does this place […] have air conditioning?

  AddBabyChangingTable:
- Does this place have a baby changing table?
+ Does this restroom have a baby changing table?

  AddBenchStatusOnBusStop:
- Is there a bench at this stop?
+ Is there a bench at this bus stop?

  AddBinStatusOnBusStop:
- Is there a waste basket at this stop?
+ Is there a waste basket at this bus stop?

  AddBusStopLit:
- Is this stop lit?
+ Is this bus stop lit?

  AddBusStopName:
- What’s the name of this stop?
+ What is the name of this bus stop?

  AddBusStopRef:
- What’s the reference number of this stop?
+ What is the reference number of this bus stop?

  AddBusStopShelter:
- Is there a shelter at this stop?
+ Is there a shelter at this bus stop?

  AddChargingStationCapacity:
- How many cars can be charged here at the same time?
+ How many cars can be charged at this charging station at the same time?

  AddDrinkingWater:
- Is the water potable here?
+ Is the water potable here? […]

  AddFuelSelfService:
- Can you pump gas yourself here?
+ Can you pump gas yourself at this fuel station?

  AddGeneralFee:
- Do you need to pay to enter here?
+ Do you need to pay to enter […]?

  AddHalal:
- Are halal products offered here?
+ Does […] offer Halal products?

  AddInformationToTourism:
- What type of tourist information is this?
+ What type of tourist information is […]?

  AddInternetAccess:
- What kind of Internet connection does this place offer?
+ What kind of Internet connection does […] offer?

  AddKosher:
- Are kosher products offered here?
+ Does […] offer kosher products?

  AddLevel:
- On which level number is this place located?
+ On which level number is […] located?

  AddOpeningHours:
- What are the opening hours here?
+ What are the opening hours of […]?

  AddPathSmoothness:
- What is the surface quality here?
+ What is the surface quality of this path here?

  AddPathSurface:
- What’s the surface here?
+ What surface does this path have here?

  AddPitchLit:
- Is it lit here?
+ Is this sport field lit?

  AddPitchSurface:
- What’s the surface here?
+ What surface does this sport field have?

  AddPlaceName:
- What’s the name of this place?
+ What is the name of this place? […]

  AddPowerPolesMaterial:
- What is this pole made of?
+ What is the material of this power or utility pole?

  AddRoadName:
- What is the name of this street?
+ What is the name of this road?

  AddRoadSurface:
- What surface does this piece of road have?
+ What surface does the road […] have here?

  AddSummitRegister:
- Is there a summit register here?
+ Is there a summit register at […]?

  AddTactilePavingBusStop:
- Does this stop have tactile paving?
+ Does this bus stop have tactile paving?

  AddToiletAvailability:
- Is there any restroom here?
+ Does […] have a restroom?

  AddVegan:
- Any vegan items on the menu here?
+ Does […] have vegan meals on the menu?

  AddVegetarian:
- Any vegetarian items on the menu here?
+ Does […] have vegetarian meals on the menu?

  AddWayLit:
- Is it lit here?
+ Is this stretch of way lit?

  AddWheelchairAccessBusiness:
- Is this place wheelchair accessible?
+ Is this location wheelchair accessible?

  AddWheelchairAccessOutside:
- Is this place wheelchair accessible?
+ Is […] wheelchair accessible?

  AddWheelchairAccessPublicTransport:
- Is this place wheelchair accessible?
+ Is this place wheelchair accessible?

  AddWheelchairAccessToilets:
- Is this place wheelchair accessible?
+ Is this restroom wheelchair accessible?

  AddWheelchairAccessToiletsPart:
- Is the restroom at this place wheelchair accessible?
+ Is the restroom at […] wheelchair accessible?

  CheckExistence:
- Is this still here?
+ Is this still here? […]

  CheckOpeningHoursSigned:
- Are the opening hours here signed?
+ Are the opening hours signed for […]?
```

FloEdelmann (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -31,12 +31,11 @@ class AddAddressStreet : OsmElementQuestType<AddressStreetAnswer> {
...
     // In Japan, housenumbers usually have block numbers, not streets
     override val enabledInCountries = AllCountriesExcept("JP")
-
     override val questTypeAchievements = listOf(POSTMAN)
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_address_street_title
 
-    override fun getTitleArgs(tags: Map<String, String>, featureName: Lazy<String?>): Array<String> =
+    override fun getTitleArgs(tags: Map<String, String>): Array<String> =
         arrayOfNotNull(tags["addr:streetnumber"] ?: tags["addr:housenumber"] ?: tags["addr:housename"])
```
FloEdelmann, line 39: Is it deliberate that this question (`What street does (house) […] belong to?`) is not simplified?

Is it because the tag priority here (`addr:streetnumber` > `addr:housenumber` > `addr:housename`) is different from the logic in `Resources.getHouseNumberLabel`?

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
FloEdelmann: @westnordost Is there a reason that the question sometimes starts with "What is" and sometimes with "What’s"?

If it should be consistent, now would be a good time to choose one, given that most strings are changed anyway.

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
westnordost: `AddAddressStreet` is asked not only for buildings. So it would be difficult to refer to the "thing" without saying housenumber. Or do you have an idea how to word it?

FloEdelmann: Why not leave it out completely?

> **What street does this belong to?**
> house number 27

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
westnordost: Will do, thanks!

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
westnordost: This what? Are you sure that it is possible even in English to word it like that?

FloEdelmann: I see no problems with that (but I might be wrong of course).

We have basically the same thing with "Is this still here?" and that seems to be a common expression: https://www.google.com/search?q=%22Is+this+still+here%3F%22

westnordost: @1ec5 suggested "What street is this on?", so I'll just use that

westnordost:
Overview of the changed strings, compiled by @FloEdelmann:

```diff
@@ AddAcceptsCash @@
- Does […] accept cash payment?
+ Is cash payment accepted here?

@@ AddAirConditioning @@
- Does this place […] have air conditioning?
+ Is this place air conditioned?

@@ AddBabyChangingTable @@
- Does this restroom have a baby changing table?
+ Does this place have a baby changing table?

@@ AddBenchStatusOnBusStop @@
- Is there a bench at this bus stop?
+ Is there a bench at this stop?

@@ AddBinStatusOnBusStop @@
- Is there a waste basket at this bus stop?
+ Is there a waste basket at this stop?

@@ AddBusStopLit @@
- Is this bus stop lit?
+ Is this stop lit?

@@ AddBusStopName @@
- What is the name of this bus stop?
+ What’s the name of this stop?

@@ AddBusStopRef @@
- What is the reference number of this bus stop?
+ What’s the reference number of this stop?

@@ AddBusStopShelter @@
- Is there a shelter at this bus stop?
+ Is there a shelter at this stop?

@@ AddChargingStationCapacity @@
- How many cars can be charged at this charging station at the same time?
+ How many cars can be charged here at the same time?

@@ AddDrinkingWater @@
- Is the water potable here? […]
+ Is the water potable here?

@@ AddFuelSelfService @@
- Can you pump gas yourself at this fuel station?
+ Can you pump gas yourself here?

@@ AddGeneralFee @@
- Do you need to pay to enter […]?
+ Do you need to pay to enter here?

@@ AddHalal @@
- Does […] offer Halal products?
+ Are halal products offered here?

@@ AddInformationToTourism @@
- What type of tourist information is […]?
+ What type of tourist information is this?

@@ AddInternetAccess @@
- What kind of Internet connection does […] offer?
+ What kind of Internet connection does this place offer?

@@ AddKosher @@
- Does […] offer kosher products?
+ Are kosher products offered here?

@@ AddLevel @@
- On which level number is […] located?
+ On which level number is this place located?

@@ AddOpeningHours @@
- What are the opening hours of […]?
+ What are the opening hours here?

@@ AddPathSmoothness @@
- What is the surface quality of this path here?
+ What’s the surface quality here?

@@ AddPathSurface @@
- What surface does this path have here?
+ What’s the surface here?

@@ AddPitchLit @@
- Is this sport field lit?
+ Is it lit here?

@@ AddPitchSurface @@
- What surface does this sport field have?
+ What’s the surface here?

@@ AddPlaceName @@
- What is the name of this place? […]
+ What’s the name of this place?

@@ AddPowerPolesMaterial @@
- What is the material of this power or utility pole?
+ What’s this pole made of?

@@ AddRoadName @@
- What is the name of this road?
+ What’s the name of this street?

@@ AddRoadSurface @@
- What surface does the road […] have here?
+ What surface does this piece of road have?

@@ AddSummitRegister @@
- Is there a summit register at […]?
+ Is there a summit register here?

@@ AddTactilePavingBusStop @@
- Does this bus stop have tactile paving?
+ Does this stop have tactile paving?

@@ AddToiletAvailability @@
- Does […] have a restroom?
+ Is there any restroom here?

@@ AddVegan @@
- Does […] have vegan meals on the menu?
+ Any vegan items on the menu here?

@@ AddVegetarian @@
- Does […] have vegetarian meals on the menu?
+ Any vegetarian items on the menu here?

@@ AddWayLit @@
- Is this stretch of way lit?
+ Is it lit here?

@@ AddWheelchairAccessBusiness @@
- Is this location wheelchair accessible?
+ Is this place wheelchair accessible?

@@ AddWheelchairAccessOutside @@
- Is […] wheelchair accessible?
+ Is this place wheelchair accessible?

@@ AddWheelchairAccessPublicTransport @@
- Is this place wheelchair accessible?
+ Is this place wheelchair accessible?

@@ AddWheelchairAccessToilets @@
- Is this restroom wheelchair accessible?
+ Is this place wheelchair accessible?

@@ AddWheelchairAccessToiletsPart @@
- Is the restroom at […] wheelchair accessible?
+ Is the restroom at this place wheelchair accessible?

@@ CheckExistence @@
- Is this still here? […]
+ Is this still here?

@@ CheckOpeningHoursSigned @@
- Are the opening hours signed for […]?
+ Are the opening hours here signed?

@@ AddAddressStreet @@
- What street does (house) […] belong to?
+ What street is this on?

@@ AddBoardType @@
- What is the topic of this information board?
+ What’s the topic of this information board?

@@ AddBridgeStructure @@
- What is the structure of this bridge?
+ What’s the structure of this bridge?

@@ AddHousenumber @@
- What is the house number of this building?
+ What’s the house number of this building?

@@ AddKerbHeight @@
- What is the height of this curb?
+ What’s the height of this curb?

@@ AddMaxHeight @@
- What is the height limit here?
+ What’s the height limit here?

@@ AddMaxPhysicalHeight @@
- What is the height limit here?
+ What’s the height limit here?

@@ AddMaxSpeed @@
- What is the speed limit for this street?
+ What’s the speed limit for this street?

@@ AddMaxWeight @@
- What is the weight limit here?
+ What’s the weight limit here?

@@ AddOrchardProduce @@
- What is being cultivated here?
+ What’s being cultivated here?

@@ AddPostboxRef @@
- What is the reference number of this postbox?
+ What’s the reference number of this postbox?

@@ AddPostboxRoyalCypher @@
- What is the royal cypher on this postbox?
+ What’s the royal cypher on this postbox?

@@ AddRoadSmoothness @@
- What is the surface quality of this road here?
+ What’s the surface quality of this road here?

@@ AddTracktype @@
- What is the surface firmness of this track?
+ What’s the surface firmness of this track?

```

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
peternewman: I'd never got round to opening an issue, but it always used to but me when I was asked `What street does (house) […] belong to?` for a railway station or an office, so this is a great improvement!

PR #3873, opened 2022-03-12
Refactor sidewalk surface quest

arrival-spring:
Refactoring sidewalk surface quest to use `AStreetSideSelectFragment` as suggested in [original PR](https://github.com/streetcomplete/StreetComplete/pull/3735/files#r813091412).

Things missing were:
- initStateFromTags (here to disable one side or the other if there's no sidewalk)
- a way to do a follow up question after tapping on a side (here to ask about generic surfaces)

Both of these are good things to have if cycleway and parking quests were to be refactored to use `AStreetSideSelectFragment`

I'm not 100% sure about the way in which I've added them both, but they do work.

The problem remaining though is the last answer button. `LastPickedValuesStore` assumes that it can convert the last answer to and from a string, however here the last answer value is something like `SurfaceAnswer(value=ASPHALT, note=null)`. This gets split on the comma and totally fails when it attempts to convert it back.

I'm not sure of the best way to solve this in general.


arrival-spring:
Also, downgrade is that the icons for each side are no longer shown in circles, and the title of the surface is displayed on each side, but you can't really see it behind the icon.

westnordost:
On the `LastPickedValuesStore`, you could require subclasses of `AStreetSideSelectFragment` to specify a a serializer and deserializer function.

> downgrade is that the icons for each side are no longer shown in circles

Why though? The side select puzzle supports setting a `de.westnordost.streetcomplete.view.Image` though, no? Couldn't the data model (be an interface) and have an `Image` instead of a `@DrawableRes Int`?

arrival-spring:
I've done the serializing and it all seems to work

> Why though? The side select puzzle supports setting a `de.westnordost.streetcomplete.view.Image` though, no? Couldn't the data model (be an interface) and have an `Image` instead of a `@DrawableRes Int`?

The problem I've run into with the above is that in `items` I want to have `RotatedCircleDrawable(resources.getDrawabale(...` but then I get `IllegalStateException: Fragment AddSidewalkSurfaceForm ... not attached to a context.` (from using `resources`)

westnordost:
can't items be 
`override val items: List<Item> get() = ...`
(i.e. a function instead of a value set in constructor?

(Sorry I didn't look into the code yet)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
```diff
@@ -76,167 +43,79 @@ class AddSidewalkSurfaceForm : AbstractQuestFormAnswerFragment<SidewalkSurfaceAn
...
-        val rightSide = rightSide
-        if (leftSide != null && rightSide != null) {
-            lastSelection =
-                if (isRoadDisplayedUpsideDown())
-                    LastSidewalkSurfaceSelection(rightSide, leftSide)
-                else
-                    LastSidewalkSurfaceSelection(leftSide, rightSide)
-        }
-    }
+        val leftStreetSideItem = StreetSideItem2(
```
arrival-spring, line 97: This feels like quite a bit of duplication here, but the only other way I could think of was to change `note` in `SurfaceAnswer` to `var`, which I wasn't sure was a good idea. (Then it could be done same as in sidewalk and shoulder with `items.find { it.value.toString() == arr[0] }!` and then just set the note value and return that.)

arrival-spring:
That was it, thanks.

Now ready for review.

The only change in functionality from before is that it now shows the surface on the road in the view after it has been selected (but somewhat hidden behind the icon).

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
arrival-spring: (not a very good quote there, sorry, I mean line 97 to 103, and same for right)

westnordost:
Sorry, forgot about this. Will review this soon

westnordost:
I'm working on this now. I needed to checkout the branch to read up on it, so I started to do all kinds of small adjustments (some unrelated to the PR but to the topic), so whatever I'll find maybe imperfect with this PR I'll fix it myself. Maybe I'll also get around to also let the `cycleway` quest to inherit from that or even refactor it to use that ViewController pattern I mentioned in the discussion forum some time ago.

westnordost:
Hmm, I'll do the ViewController stuff later after all. What I changed on your PR was to
- replace the serialize/deserialize with an abstract function where implementing classes must return the display item for any given value
- don't save to favs if a generic surface was selected (solves issue btw that would return corrupt data if the user included a "#" in his note)

arrival-spring:
Thanks for doing all that!

PR #3874, opened 2022-03-12
Add Vietnamese house number format

### LINKED ISSUE #3872: House number validator rejects slash-laden house numbers in Vietnam
The address quest requires a relatively rigid format for `addr:housenumber` that allows a single slash but requires a single digit or single letter to follow it:

https://github.com/streetcomplete/StreetComplete/blob/88d71c4a792376d82d5b1863553cf96f551088f7/app/src/main/java/de/westnordost/streetcomplete/quests/address/HousenumberAnswerValidator.kt#L5-L6

This doesn’t work for the vast majority of urban addresses in Vietnam. Vietnam’s urban address format includes house numbers that resemble a POSIX relative file path that starts from a lane off an arterial road and recurses through each alley until you reach the destination. For example, this is a [valid address](https://congan.com.vn/doi-song/song-o-noi-so-nha-dai-hon-so-chung-minh-thu_36586.html):

> Số 1806/127/2/6/15/48/2A Huỳnh Tấn Phát
> Khu phố 6, thị trấn Nhà Bè, huyện Nhà Bè
> Thành phố Hồ Chí Minh

translation:

> Number 1806/127/2/6/15/48/2A, Huỳnh Tấn Phát street
> Ward 6, Nhà Bè town, Nhà Bè district
> Ho Chi Minh City

or more verbosely (but no one would write this):

> Number 2A, alley 48, alley 15, alley 6, alley 2, alley 127, alley 1806, Huỳnh Tấn Phát street
> Ward 6, Nhà Bè town, Nhà Bè district
> Ho Chi Minh City

There’s even a word for the most deeply nested addresses, [_siêu xuyệt_](https://en.wiktionary.org/wiki/siêu_xuyệt#Vietnamese), but 2–4 slashes are not at all uncommon.

StreetComplete would be very helpful for surveying addresses in these dense neighborhoods, since it’s very difficult to collect them more systematically. [Sometimes these addresses even get deleted](https://www.openstreetmap.org/changeset/59631625) by Westerners who are unfamiliar with Vietnamese addressing and mistake them for vandalism or botched imports.

/ref https://github.com/openstreetmap/iD/pull/4235#discussion_r133900734

1ec5:
Added a regular expression that matches the most common Vietnamese house number formats. There’s a long tail of less common formats, but this regular expression balances the need to allow valid house numbers with the need to catch typos.

Here are some examples of house numbers that would be allowed:

* 1099A
* 1806/127/2/6/15/48/2A
* 73B/563B bis
* 40bis/1
* 35N/1C
* 42/7bis
* L1004A

This assumes the regular expressions are case-insensitive. I’m unsure whether it’s more correct to capitalize “Bis” and “Ter” or leave them lowercase, as in French. The house number signs are all uppercase, even for words that are normally written in mixed case.

References:

* Đỗ Phi Hùng (2012), “[Vẫn loay hoay trong ‘mê hồn trận’ số nhà](https://tuoitre.vn/van-loay-hoay-trong-me-hon-tran-so-nha-476867.htm)” [Still fussing over a “labyrinth” of house numbers], _Tuổi Trẻ_.
* Quốc Định (2012), “[Lạc trong ma trận số nhà](https://giadinh.net.vn/lac-trong-ma-tran-so-nha-172120716084712382.htm)” [Lost in a labyrinth of house numbers], _Gia đình_.

Fixes #3872.


westnordost:
Something seems to be wrong with the regex:

https://regexper.com/#%5Cp%7BL%7D%3F%28%3F%3A%5Cp%7BN%7D%7B1%2C4%7D%28%3F%3A%5Cs%3F%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%29%3F%5C%29*%5Cp%7BN%7D%7B1%2C4%7D%28%3F%3A%5Cs%3F%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%28%3F%3A%5Cs%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%29%3F%29%3F

kmpoppe:
I guess there was an additional `\` that shouldn't have been there. This lights green:

https://regexper.com/#%5Cp%7BL%7D%3F%28%3F%3A%5Cp%7BN%7D%7B1%2C4%7D%28%3F%3A%5Cs%3F%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%29%3F%29*%5Cp%7BN%7D%7B1%2C4%7D%28%3F%3A%5Cs%3F%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%28%3F%3A%5Cs%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%29%3F%29%3F

1ec5:
No, the `/` is intentional – it’s what allows slashes between the alley numbers. Some regex testing tools require the `/` to be escaped because of the `s/…/…/` literal syntax in JavaScript and `sed`, but as the `ES` line shows, this YAML file doesn’t require slashes to be escaped:

https://github.com/streetcomplete/StreetComplete/blob/9457ec3b2bb1d9d2c89902e43f0720e74b3ef912/res/country_metadata/additionalValidHousenumberRegex.yml#L3

1ec5:
For future reference, here’s a [compilation of house number laws in Vietnam by province](https://danluat.thuvienphapluat.vn/nhung-quy-dinh-can-biet-ve-bien-so-nha-146979.aspx). [The law for Ho Chi Minh City](https://thuvienphapluat.vn/van-ban/Bat-dong-san/Quyet-dinh-22-2012-QD-UBND-Quy-che-danh-so-va-gan-bien-so-nha-141155.aspx) provides for some syntaxes that aren’t covered here, like `H01.02` for apartment 2 on level 1.[^levels] I’ve excluded this syntax because `addr:floor` and `addr:unit` are well-established keys globally, though I don’t know if they’re used in Vietnam.

[^levels]: Southern Vietnam uses 0-based floor numbers, whereas northern Vietnam use 1-based floor numbers.

FloEdelmann:
It seems to work fine in https://regex101.com/r/DDnq9d/1, after I set the delimiters from `/` to `#` and added `^` to the start and `$` to the end to restrict each match to a single line.

westnordost:
Okay, for reference, this would be the correct visualization of the regex.



corrected in paint from

https://regexper.com/#%5Cp%7BL%7D%3F%28%3F%3A%5Cp%7BN%7D%7B1%2C4%7D%28%3F%3A%5Cs%3F%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%29%3F%5B%2F%5D%29*%5Cp%7BN%7D%7B1%2C4%7D%28%3F%3A%5Cs%3F%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%28%3F%3A%5Cs%28%3F%3Abis%7Cter%7Ck%C3%A9p%7C%5Cp%7BL%7D%29%29%3F%29%3F

westnordost:
This regex makes `123 bis A` and `123 bis bis` correct. Are you sure about that?

1ec5:
* [241/110 Bis A Nguyễn Văn Luông](https://infodoanhnghiep.com/thong-tin/Cty-TNHH-San-Xuat-Thuong-Mai-Hien-Tin-27022.html)
* [42 Bis B Trần Đình Xu](https://tracuudn.com/cong-ty/a869bb17-cong-ty-tnhh-truc-tin)
* [59 Bis F Hồ Hảo Hớn](https://datnen.vn/bat-dong-san/cho-thue-phong-tro-59-bis-f-ho-hao-hon-phuong-co-giang-quan-1-526734.view.html)
* [51 A Bis An Bình](https://tracuudn.com/cong-ty/545df92d-cong-ty-tnhh-mat-troi-hanh-phuc)

I guess the owner of 241/110 Bis subdivided the property into lettered lots without building an alley. But “bis bis” is probably pretty unlikely.

How thorough do we want to be, since there’s such a long tail of alternative formats? [This vendor is advertising](https://hoangkimplaza.com/lam-bien-nha-tai-thanh-pho-ho-chi-minh-2.html) house number signs in the following unsupported formats for Ho Chi Minh City specifically:

* 11E3
* 923W5
* LK27
* 18 BT8
* 55-TT2
* 128-Lô16D (_lô_ means “lot”)
* 16.D3

These formats are widely considered confusing because they aren’t as predictable. Lately, local authorities have begun phasing them out in favor of more slashes. Should we only support the more predictable formats, or should we make it more freeform like in https://github.com/streetcomplete/StreetComplete/pull/3874#issuecomment-1066045698?

westnordost:
Well, the app just warns the user (something like "This housenumber looks unusual, are you sure?") if the regex doesn't match, so it is OK to not have weird but maybe still valid combinations in the regex.

Then this can be merged as it is right?

1ec5:
If this is just for a warning that the user can easily bypass, then I think it’s mergeable. The 123/456A format is the 97.5% case anyhow.

PR #3878, opened 2022-03-14
Add smichel17 to codeowners

smichel17:
These are the sections of the app I'm most familiar with.

I'm busier nowadays and am having trouble keeping up with notifications when watching the whole repo, which means I might be missing pings or changes that I *should* be paying attention to. So, I'm going to switch my notifications back to "Participating and mentions". But I still want to get notifications when someone touches these files.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,3 @@
 res/country_metadata/orchardProduces.yml  @rugk
+app/src/main/java/de/westnordost/streetcomplete/quests/LastPickedValuesStore.kt @smichel17
+app/src/main/java/de/westnordost/streetcomplete/map/ @smichel17


FloEdelmann:
@mnalis I also noticed that today, so I asked @rugk already in https://github.com/streetcomplete/StreetComplete/pull/3764#issuecomment-1066685956 :wink:

rugk:
Ah it only works when you actually have write access, so well… okay, feel free to remove me there.

westnordost:
Uhm well, you **do** have read access. So, I consider this a bug / missing feature in github, so let's keep you in the file even if it doesn't work.

smichel17:
It makes some sense— it's the code**owners** file, not code*watchers*, and it's hard to argue that you own something if you're not allowed to modify it (though it seems many tech companies would like to deny that.. *glances at locked-bootloader phone :rage:*).

Actually, I guess the issue might be notification spam. If you don't have write access, you can't remove your name if you don't want the notifications.

westnordost:
Well, but you are certainly not the codeowner of the `streetcomplete.map` package, so I very much interpret this as _codewatchers_.

PR #3885, opened 2022-03-16
Viewcontroller

westnordost:
See #3830

What do you think?

The plan would be to eventually move all common view logic of different quest forms into those "ViewController"s. Of course it is not a 100% replacement for the A*SomethingSomething*Fragment but most of it.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddWidthForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddWidthForm.kt
@@ -37,7 +37,7 @@ class AddWidthForm : AbstractQuestFormAnswerFragment<WidthAnswer>() {
         binding.widthExplanationTextView.text = explanation
 
         lengthInput = binding.lengthInput.let {
-            LengthInputViewController(it.unitSelect, it.metersContainer, it.metersInput, it.feetInchesContainer, it.feetInput, it.inchesInput, )
+            LengthInputViewController(it.unitSelect, it.metersContainer, it.metersInput, it.feetInchesContainer, it.feetInput, it.inchesInput)
         }
         lengthInput.unitSelectItemResId = R.layout.spinner_item_centered_large
         lengthInput.isCompactMode = true


app/src/main/java/de/westnordost/streetcomplete/quests/max_height/AddMaxHeightForm.kt
```diff
@@ -97,30 +56,12 @@ class AddMaxHeightForm : AbstractQuestFormAnswerFragment<MaxHeightAnswer>() {
     }
 
     private fun userSelectedUnrealisticHeight(): Boolean {
-        val height = getHeightFromInput() ?: return false
-        val m = height.toMeters()
-        return m > 6 || m < 1.9
+        val m = lengthInput.length?.toMeters() ?: return false
+        return m > 6 || m < 1.8
```
FloEdelmann, line 60: Is the change from `1.9` to `1.8` deliberate?

app/src/main/java/de/westnordost/streetcomplete/quests/width/AddWidthForm.kt
```diff
@@ -34,10 +36,15 @@ class AddWidthForm : AbstractQuestFormAnswerFragment<WidthAnswer>() {
         binding.widthExplanationTextView.isGone = explanation == null
         binding.widthExplanationTextView.text = explanation
 
-        binding.lengthInput.maxFeetDigits = if (isRoad) 3 else 2
-        binding.lengthInput.maxMeterDigitsBeforeDecimalPoint = if (isRoad) 2 else 1
-        binding.lengthInput.selectableUnits = countryInfo.lengthUnits
-        binding.lengthInput.onInputChanged = {
+        lengthInput = binding.lengthInput.let {
+            LengthInputViewController(it.unitSelect, it.metersContainer, it.metersInput, it.feetInchesContainer, it.feetInput, it.inchesInput, )
```
FloEdelmann, line 40:
```suggestion
            LengthInputViewController(it.unitSelect, it.metersContainer, it.metersInput, it.feetInchesContainer, it.feetInput, it.inchesInput)
```

FloEdelmann (commented):
Looks good! (just from a code review on GitHub)

app/src/main/java/de/westnordost/streetcomplete/quests/max_height/AddMaxHeightForm.kt
westnordost: yes

westnordost:
Thanks for the review!
I meant also in general, the pattern and the naming thereof.

FloEdelmann:
I like the pattern, but I couldn't really think of (or find) a better name.

PR #3892, opened 2022-03-19
remove remains of getTitleArgs

matkoniecz:
as far as I can see this is not needed and not used but maybe I missed something (and it is actually needed for some obscure use or even more can be removed)


westnordost:
It is currently not needed, but does that mean there can be not instance where it is not needed in the future? I'd rather keep this, it's just ~11 lines of code after all

matkoniecz:
For me main issue was that I was confused is it used somehow in way that I am not noticing.

PR #3896, opened 2022-03-20
add special case for Norwegian Bokmål

### LINKED ISSUE #3890: Norwegian Bokmål has "no" code in iD and presets imported from iD, Android expects "nb" code and fails to use iD preset data
https://poeditor.com/projects/po_edit?id=97843&term=273463027&id_language=127#showComments273463027

> The search field only accept english (for instance "bakery" instead of the Norwegian word "bakeri"

after this is fixed

> so I "translated" this to "A(n)… (på engelsk)", which means "A(n)... (in english)". A bit longer than the source, but I think it should be OK...

should be changed back/reset

matkoniecz:
this is a bit ugly special case but istuation itself is also an ugly special case

fixes #3890

I investigated is Bokmål affected also elsewhere and found no evidence of that - I found only that it is listed twice in language selection list (once as Norwegian, once as Norwegian Bokmål) - and I am not really sure is it even a bug.

I also looked for a similar problematic cases and found no issues.

Still, likely that I missed something and something else is affected or it should be changed in a different way as it seems quite ugly hack.

### DIFF
--- a/app/src/main/assets/osmfeatures/default/nb.json
+++ b/app/src/main/assets/osmfeatures/default/nb.json
@@ -1,0 +1,5114 @@
+{
+    "no": {
+        "presets": {
+            "categories": {
+                "category-barrier": {
+                    "name": "Barriereobjekter"
+                },
+                "category-building": {
+                    "name": "Bygningsrelaterte egenskaper"
+                },
+                "category-golf": {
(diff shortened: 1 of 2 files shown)


matkoniecz:
Confirmed as a good idea by https://github.com/streetcomplete/StreetComplete/issues/3890#issuecomment-1073229990

Issue #2539, opened 2021-01-30
Clarification for vegetarian quest


Helium314:
According to the displayed text I am supposed to answer _yes_ for the vegetarian quest only if the place provides 'proper [vegetarian] eating'.
Does this mean that for places that do not offer 'proper eating' (e.g. a café offering only cake, or a bar selling peanuts) the answer is always no? According to the text it is the case, but answering 'no' for such places still feels completely wrong...

westnordost:
Do you have a suggestion how to improve the wording? Current wording:

"Only answer “yes” if the place provides proper eating options for this diet, ignore starters etc."

Helium314:
I thought about it for quite a while, but unfortunately nothing that wouldn't be horrendously long...

peternewman:
Would something around the word choice/selection/range help?

"Only answer “yes” if the place provides a proper choice of eating options for this diet, ignore starters etc."

A cafe only offering cake is fine I'd say, as long as the cake's aren't all topped with bacon (maybe maple cured) for example.

I guess the easier example is if you're answering the vegan quest for a cake cafe, then if there's one cake then yes, if they only sold a vegan cereal bar and no cakes, then no.

mnalis:
@peternewman I was under impression answer would **always** be "no" in cafes and other places which **do not serve full meals**? At least that is what "proper eating" implies to me - no snacks, but a full meal, one which you can eat for lunch and work all day on the calories taken. "ignore starters, etc ." seems to me that it reaffirms that stance. Or maybe I'm getting this wrong? 

It does look it could do with better explanation, if we seem to have understood it differently...

westnordost:
Really? Obviously, if the establishment does not serve "proper meals" at all, but just cakes for example, it is of course eligible for vegetarians.

westnordost:
The current text already mentions "proper options", not "proper meals". Cakes are not starters.

mnalis:
Well, It looks like it is _obvious_ to some, but not others (like original poster and myself). If cafe serves only cakes (made without animal fat?), I agree it does contain food eligible for vegetarians, but it does not seem what the text of the quest is asking for  - to me it **looks like it is explicitly excluding** all places where vegetarian can't get "A full day meal". So if restaurant had only had  pickles or vegetable salad as non-meat things on menu, it would be excluded by that quest definition, wouldn't it? 

[...]

PR #3902, opened 2022-03-21
Make the question better match quest_dietType_explanation

peternewman:
For example olives would be "any vegetarian/vegan items" but aren't a comparable choice. Any is "not none", which isn't really what the longer description aims at.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -847,7 +847,7 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
 
     <string name="quest_cycleway_width_title">"What’s the width of this cycleway here?"</string>
 
-    <string name="quest_dietType_halal_name_title2">Are halal products offered here?</string>
+    <string name="quest_dietType_halal_name_title2">Are Halal products offered here?</string>
     <string name="quest_dietType_explanation_halal">Halal products must be prepared according to special rules, with Halal certification agencies marking products in compliance.</string>
 
     <string name="quest_dietType_kosher_name_title2">Are kosher products offered here?</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -853,10 +853,10 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_dietType_kosher_name_title2">Are kosher products offered here?</string>
     <string name="quest_dietType_explanation_kosher">Kosher products must be prepared according to special rules, with kashrut certification agencies marking products in compliance.</string>
 
-    <string name="quest_dietType_vegan_title2">Any vegan items on the menu here?</string>
+    <string name="quest_dietType_vegan_title2">Are there vegan items on the menu here?</string>
     <string name="quest_dietType_explanation_vegan">Vegan meals contain no animal products (no meat, fish, poultry, milk products, eggs, honey …).</string>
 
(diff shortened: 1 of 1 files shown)


westnordost:
Didn't you notice the

> Only answer “yes” if the place provides a comparable choice of eating options for this diet

?

peternewman:
Yes, I did, and I'm aware of what the question is so go pretty much entirely on the icon, but as has been proved time and again, many people don't read the smaller text, so getting the main question to cover it as effectively as possible without making it too wordy seems very beneficial to me.

westnordost:
It's casual, yes, but that doesn't mean it is not well translatable. As it happens, casual language is usually better translatable to my experience.

westnordost:
Anyway, the current text came from a discussion on OSM-Slack US (#language channel). I learnt that e.g. "veg item on the menu" is synonymous to "veg option" and both mean proper meals, not "an item in a list". I am not convinced that your suggestion would be an improvement in clarity.

peternewman:
> It's casual, yes, but that doesn't mean it is not well translatable. As it happens, casual language is usually better translatable to my experience.

Casual language is generally more ambiguous so more chance of confusion.

> Anyway, the current text came from a discussion on OSM-Slack US (#language channel). I learnt that e.g. "veg item on the menu" is synonymous to "veg option" and both mean proper meals, not "an item in a list".

I'm not sure item has the same significance this side of the pond, compared to the vegetarian option.

> I am not convinced that your suggestion would be an improvement in clarity. **Edit**: IIRC the previous text was "Are there any...", so "Any" is just the casual form of that

The previous text, as per @FloEdelmann excellent diff ( https://github.com/streetcomplete/StreetComplete/pull/3870#discussion_r825452124 ), was:
```diff
@@ AddVegan @@
- Does […] have vegan meals on the menu?
+ Any vegan items on the menu here?

@@ AddVegetarian @@
- Does […] have vegetarian meals on the menu?
+ Any vegetarian items on the menu here?
```

Would you accept changing it to meals (Any veg meals on the menu here?), which seems like a much better fit all round (and matches the more detailed text)?

I suspect the underlying issue is it's trying to fulfil two purposes (and I suspect the reason someone suggest item), if you go to a doughnut shop, a single doughnut isn't a meal, but it is an item, but you're only buying one thing. But as part of a meal, appetisers/starters/sides are far more likely to be veg friendly, but you'll be unimpressed if that's all you get.

westnordost:
>> It's casual, yes, but that doesn't mean it is not well translatable. As it happens, casual language is usually better translatable to my experience.

> Casual language is generally more ambiguous so more chance of confusion.

Well it looks like we have a different opinion here. Note that I wrote "casual", not "colloquial" or "slang". In my experience, the casual language that is spoken on casual/normal interactions is the first thing you learn to use right.

> I'm not sure item has the same significance this side of the pond, compared to the vegetarian option.
> Would you accept changing it to meals (Any veg meals on the menu here?)

 I wouldn't accept changing it to meals because this is the very thing the people in the #language channel found sounded odd. "This is something a waiter would say". But well, if the understanding differs on the other side of the pond, that's good, so then you can use a better fitting wording for en-GB any everybody is happy.

mnalis:
see also https://github.com/streetcomplete/StreetComplete/issues/2539 - "comparable" was key word to me to indicate that beefsteak house which besides beefsteaks has vegetarian soups and salads and potatoes as side dishes does not count as "yes" in that quest.

PR #3903, opened 2022-03-21
Show local ref where present along with names

peternewman:
E.g. for bus stops

Fixes #3800

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/NameAndLocationLabel.kt
@@ -63,10 +63,12 @@ private fun getFeatureName(
 fun getNameLabel(tags: Map<String, String>): String? {
     val name = tags["name"]
     val brand = tags["brand"]
+    val localRef = tags["local_ref"]
     val ref = tags["ref"]
     val operator = tags["operator"]
 
-    return name
+    return if (name != null && localRef != null) "$name ($localRef)" else null
+        ?: name
         ?: brand
         ?: if (ref != null && operator != null) "$operator $ref" else null
         ?: operator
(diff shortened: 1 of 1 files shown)


peternewman:
Yes, but I forgot, and I think it doesn't count as it's a discussion not an issue!

kmpoppe:
GitHub will not automatically close the discussion, that's correct, but just keeping the reference helps :)

westnordost:
So `local_ref` should only be shown if there is a name? Could you please post some rationale for that?

I.e.
1. why with name and not with operator?
2. why not if neither name (nor operator) is available

peternewman:
> So `local_ref` should only be shown if there is a name? Could you please post some rationale for that?

I was just trying to change the bare minimum for the moment.

Given you didn't want house numbers shown if a name existed elsewhere, I was sticking to the same idea).

Personally I'm only aware of bus stops where it would add value for me.

> I.e.
> 
>     1. why with name and not with operator?

We could do it there too.

>     2. why not if neither name (nor operator) is available

I've added it, in theory in the bus stop case it could result in just "B" being shown, unless we want to try and translate strings for both ref and local ref and use them?

westnordost:
You put `local_ref` into parenthesis but `ref` isn't. Is this deliberate?

peternewman:
In my experience the local_ref is normally one or two characters, whereas ref is often longer (as it will be unique per country potentially) so I was just trying to differentiate the two types of ref and make it less likely people think there is a typo in the name (e.g. [Euston Station C](https://www.openstreetmap.org/node/469760112).

peternewman:
> Then maybe for consistency have parenthesis for `ref` too`

Done, I've taken the liberty of using square brackets to differentiate from localref. It shouldn't make much difference to average users, but would allow any power users to leave a note if something isn't using the expected tagging (e.g. if a post box or bus stop has used local_ref rather than ref).

> Looking at taginfo, most`ref` seem to be single or two letters too

The subtlety might be which tags they're used in combination with, and whether they're actually likely to appear on SC.

PR #3907, opened 2022-03-23
Housenumber+housename

westnordost:
### Changes
- better layout especially on smaller screens (ABC button does not so easily overlap with the input anymore). Already on master
- fixed that if one already filled the form and then selected "house name and house number" answer (=> form cleared), the OK button would stay
- fixed that when the housenumber hint is shown (= previous input), it would already be considered as a complete form in "house name and house number" form when the house name has been filled in
- the form for input of house name + house number reuses the forms for the house number input. Changed data structure. Which means:
  - \+ / - button and ABC button is available
  - "house name and house number" answer works correctly in Czech Republic, Slovakia and Japan

### Czech Republic
Screenshot_20220323-020154_StreetComplete Dev
### Slovakia
Screenshot_20220323-020251_StreetComplete Dev
### Japan
Screenshot_20220323-025127_StreetComplete Dev
### All others
Screenshot_20220323-023404_StreetComplete Dev


westnordost:
I also replaced the [French emanel sign](https://www.google.com/search?q=French+enamel+sign&tbm=isch) styling of housenumber signs to be a more generic/simple because
- it is a bit noisy
- it is a very specific style

westnordost:
@matkoniecz I added you as a reviewer because you did the original implementation. This is however more of a FYI, so if you have no time currently, you can have a look at it some other time.

PR #3931, opened 2022-03-30
Resurvey tactile_paving=incorrect in the TactilePavingCrosswalk quest

Cj-Malone:
Now TactilePavingCrosswalk allows answering with `incorrect` we should also be able to resurvey it.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt
@@ -22,6 +22,7 @@ class AddTactilePavingCrosswalk : OsmElementQuestType<TactilePavingCrosswalkAnsw
             !tactile_paving
             or tactile_paving = unknown
             or tactile_paving = no and tactile_paving older today -4 years
+            or tactile_paving = incorrect and tactile_paving older today -4 years
             or tactile_paving = yes and tactile_paving older today -8 years
           )
     """.toElementFilterExpression() }


matkoniecz:
For: this allows to update now correct info

Against: is it likely that tactile paving is incorrect in way not obvious for mappers?

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt
```diff
@@ -22,6 +22,7 @@ class AddTactilePavingCrosswalk : OsmElementQuestType<TactilePavingCrosswalkAnsw
             !tactile_paving
             or tactile_paving = unknown
             or tactile_paving = no and tactile_paving older today -4 years
+            or tactile_paving = incorrect and tactile_paving older today -4 years
```
westnordost, line 25:
```suggestion
            or tactile_paving ~ no|incorrect  and tactile_paving older today -4 years
```

Cj-Malone: I had left it separate in case there was a discussion on how often it should be resurveyed.

Cj-Malone:
@matkoniecz I don't have a strong opinion on this, but I think if we trust people to answer incorrect, we should trust people to do it again.

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingCrosswalk.kt
westnordost: I think it belongs into the "no" category.

Adding tactile paving to a crossing that has no or incomplete tactile paving does not require the whole inersection being rebuilt.
That tactile paving gets lost is only really possible after a whole intersection is either redesigned (I'd say) or degenerates over time.

matkoniecz:
The difference is that very small part all cases is really tricky. But much higher part of ones already tagged `=incorrect` will be in some unusual, tricky or confusing.

PR #3932, opened 2022-03-30
add step count on stile quest

### LINKED ISSUE #3141: ask about step_count at barrier=stile stile=stepover/ladder
### General
Affected tag(s) to be modified/added: [](https://wiki.openstreetmap.org/wiki/Key:step_count)
Question asked: **How many steps are at the side of this stile? If step count differs on each side, give higher number**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose - as discussed at talk-gb and OSM Wiki it is useful for accessibility. Especially for finding hiking routes good for older people who are still mobile, hiking but such steps may be a serious barrier them.

### Ideas for implementation

**Element selection:** 
`barrier=stile` `stile=stepover`

**Metadata needed:** No

**Proposed UI:** Reuse one from existing step count quest

matkoniecz:
fixes #3141

new text is being added

screen01


app/src/main/java/de/westnordost/streetcomplete/quests/step_count/AddStepCountStile.kt
```diff
@@ -0,0 +1,52 @@
...
+        ways with
+          access ~ private|no
+          and foot !~ permissive|yes|designated
+    """.toElementFilterExpression() }
+
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
+        val excludedWayNodeIds = mutableSetOf<Long>()
+        mapData.ways
+            .filter { excludedWaysFilter.matches(it) }
+            .flatMapTo(excludedWayNodeIds) { it.nodeIds }
```
FloEdelmann, line 31: Does this work as well?


```suggestion
        val excludedWayNodeIds = mapData.ways
            .filter { excludedWaysFilter.matches(it) }
            .flatMap { it.nodeIds }
```

res/graphics/quest/steps_brown.svg
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
FloEdelmann, line 1: Please keep the Android drawable filename and the SVG source filename in sync, so they can be matched in the quest list generation task (and by humans).

app/src/main/res/values/strings.xml
```diff
@@ -1455,5 +1455,6 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_wheelchairAccess_limited_description_public_transport">Partially means that there is no more than an ankle-high step to access it.</string>
     <string name="quest_wheelchairAccess_description_toilets">"A wheelchair sign like this might be visible if the restroom is accessible.
 Partially means that a wheelchair can enter and use the restroom, but no handrail or adapted hand basin is present."</string>
+    <string name="quest_step_count_stile_title">How many steps are at the side of this stile? If step count differs on each side, give higher number</string>
```
FloEdelmann, line 1458: I'd suggest simplifying the question:


```suggestion
    <string name="quest_step_count_stile_title">How many steps does this stile have? If the step count differs for each side, enter the bigger number.</string>
```

Also, I'd move the second sentence after the question to either a description text above the input field, or to a dialog that opens when clicking "Other Answers" > "Differs for each side".

FloEdelmann (changes requested):
Looks good all in all :)
Just a few minor suggestions.

app/src/main/res/values/strings.xml
matkoniecz: > How many steps does this stile have?

I would answer 6 if it would have 3 on each side.

res/graphics/quest/steps_brown.svg
matkoniecz: yes, I used wrong icon as a base

app/src/main/java/de/westnordost/streetcomplete/quests/step_count/AddStepCountStile.kt
matkoniecz: Yep, tested by having solely `        val excludedWayNodeIds = mutableSetOf<Long>()` and failing tests and then switching to this form.

This code existed in several places. Hmmm, maybe it should be extracted to some function?

westnordost: Uh, what is this?

Isn't `excludedWayNodeIds` now not just a list?

This'd wrong for two reasons:
1. it then can contain duplicate node ids. Yes, this is completely possible, e.g. two ways that cross share a node
2. more importantly, it is deliberate that this collection is a **Set** because `contains()` is executes in O(1) while on a List it executes in O(n)

app/src/main/res/values/strings.xml
westnordost: It doesn't make much sense to me to only count the number of steps on one side rather the total amount of steps. Imagine it was mapped as a very short `highway=steps`, then it would also be the total amount.
Where is this from, that the step count should be only on one side, but if it differs, the greater one?

matkoniecz: > It doesn't make much sense to me to only count the number of steps on one side rather the total amount of steps. 

This is a clearly established tagging for that situation and was already in use.

¯\_(ツ)_/¯

westnordost: There are 18376 `stile=stepover`, 1235 (7%) are tagged with `step_count`. 1235 usages are not that much but certainly enough for `step_count` to be considered "in use" for stepover stiles, but not that particular interpretation of how to count steps.

The [wiki page for stepover](https://wiki.openstreetmap.org/wiki/Tag%3Astile%3Dstepover) stile does not mention anything, neither does its discussion page. The [wiki page for step_count](https://wiki.openstreetmap.org/wiki/Key:step_count) only mention that the tag is used on stiles and nothing related on the discussion page.

Hence, given the missing documentation, no matter what some people on the mailing list write would be ideal in their opinion, the real world meaning of `step_count` on currently tagged stiles is simply "number of individual steps [for this stile]" and hence it would only be consistent to just ask for that: "How many steps does this stile have?".

matkoniecz: https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dstile has it (from what I remember I added it there based on outcome of discussion done as I considered adding this quest)

If you consider this schema as sufficiently silly to reject this quest then please close this PR (I am not planning running a proposal here as I have queue of even worse tagging and I do not consider it as so outrageously wrong to work on changing it)

matkoniecz: Maybe asking more people how this tag is used would make sense? I can generate list of all people who ever added `step_count` to stile, ordered by how many tags of currently present ones they added if anyone is interested in collecting such info.

westnordost: If that doesn't take too much of your time. The number of people who tagged this should not be too large. I can ask them via osm messages and document the result in the relevant wiki pages.

matkoniecz: Here it goes: user, example changeset, one of stiles from that changeset, total count of cases where they were first to add `step_count` and it is also present nowadays.

https://www.openstreetmap.org/user/SomeoneElse https://www.openstreetmap.org/changeset/118712057 https://www.openstreetmap.org/node/9592964605/history 537
https://www.openstreetmap.org/user/SK53 https://www.openstreetmap.org/changeset/103863609 https://www.openstreetmap.org/node/8680778970/history 277
https://www.openstreetmap.org/user/DaveD https://www.openstreetmap.org/changeset/116513251 https://www.openstreetmap.org/node/9442930384/history 141
https://www.openstreetmap.org/user/will_p https://www.openstreetmap.org/changeset/89479892 https://www.openstreetmap.org/node/7817798278/history 110
https://www.openstreetmap.org/user/John%20Stanworth https://www.openstreetmap.org/changeset/114488944 https://www.openstreetmap.org/node/9306669170/history 108
https://www.openstreetmap.org/user/Pink%20Duck https://www.openstreetmap.org/changeset/112503926 https://www.openstreetmap.org/node/9171867815/history 38
https://www.openstreetmap.org/user/craigloftus https://www.openstreetmap.org/changeset/111874426 https://www.openstreetmap.org/node/9132604165/history 36
https://www.openstreetmap.org/user/tomhukins https://www.openstreetmap.org/changeset/112133419 https://www.openstreetmap.org/node/9149556997/history 29
https://www.openstreetmap.org/user/mentor https://www.openstreetmap.org/changeset/22955272 https://www.openstreetmap.org/node/2732908666/history 20
https://www.openstreetmap.org/user/mueschel https://www.openstreetmap.org/changeset/106666440 https://www.openstreetmap.org/node/8849315034/history 17
https://www.openstreetmap.org/user/Borbus https://www.openstreetmap.org/changeset/70899778 https://www.openstreetmap.org/node/6519002388/history 15
https://www.openstreetmap.org/user/davespod https://www.openstreetmap.org/changeset/59213027 https://www.openstreetmap.org/node/5380326209/history 14
https://www.openstreetmap.org/user/r1608t https://www.openstreetmap.org/changeset/117621923 https://www.openstreetmap.org/node/9520639937/history 8
https://www.openstreetmap.org/user/davespod_bing https://www.openstreetmap.org/changeset/24090747 https://www.openstreetmap.org/node/2959254228/history 7
https://www.openstreetmap.org/user/spiregrain https://www.openstreetmap.org/changeset/106286955 https://www.openstreetmap.org/node/8829699657/history 6
https://www.openstreetmap.org/user/drnoble https://www.openstreetmap.org/changeset/110757277 https://www.openstreetmap.org/node/9064273917/history 4
https://www.openstreetmap.org/user/LuSirto https://www.openstreetmap.org/changeset/18563912 https://www.openstreetmap.org/node/2510185674/history 4
https://www.openstreetmap.org/user/10992 https://www.openstreetmap.org/changeset/115811859 https://www.openstreetmap.org/node/9395374276/history 4
https://www.openstreetmap.org/user/confusedbuffalo https://www.openstreetmap.org/changeset/112619350 https://www.openstreetmap.org/node/1184306845/history 2
https://www.openstreetmap.org/user/TheBlackMan https://www.openstreetmap.org/changeset/79468620 https://www.openstreetmap.org/node/7124941083/history 2
https://www.openstreetmap.org/user/Paul%20Berry https://www.openstreetmap.org/changeset/118544088 https://www.openstreetmap.org/node/8125930707/history 2
https://www.openstreetmap.org/user/Reinhard12 https://www.openstreetmap.org/changeset/110250481 https://www.openstreetmap.org/node/8923215121/history 2
https://www.openstreetmap.org/user/efemm https://www.openstreetmap.org/changeset/107364284 https://www.openstreetmap.org/node/8890697234/history 2
https://www.openstreetmap.org/user/uAva https://www.openstreetmap.org/changeset/113075975 https://www.openstreetmap.org/node/9207378760/history 2
https://www.openstreetmap.org/user/ChrissW-R1 https://www.openstreetmap.org/changeset/117857148 https://www.openstreetmap.org/node/369054711/history 1
https://www.openstreetmap.org/user/Local_Walker https://www.openstreetmap.org/changeset/105891652 https://www.openstreetmap.org/node/2548429034/history 1
https://www.openstreetmap.org/user/Jez%20Nicholson https://www.openstreetmap.org/changeset/103473991 https://www.openstreetmap.org/node/3379472794/history 1
https://www.openstreetmap.org/user/DanGregory https://www.openstreetmap.org/changeset/42921005 https://www.openstreetmap.org/node/4200073650/history 1
https://www.openstreetmap.org/user/TrekClimbing https://www.openstreetmap.org/changeset/110391248 https://www.openstreetmap.org/node/9044025683/history 1
https://www.openstreetmap.org/user/LivingWithDragons https://www.openstreetmap.org/changeset/111997001 https://www.openstreetmap.org/node/9139644382/history 1
https://www.openstreetmap.org/user/michael60634 https://www.openstreetmap.org/changeset/112630106 https://www.openstreetmap.org/node/9178965268/history 1
https://www.openstreetmap.org/user/animmo https://www.openstreetmap.org/changeset/116783764 https://www.openstreetmap.org/node/9460457870/history 1
https://www.openstreetmap.org/user/b-unicycling https://www.openstreetmap.org/changeset/119028502 https://www.openstreetmap.org/node/9613536909/history 1

peternewman: > I'd suggest simplifying the question:

Or something like:
```suggestion
    <string name="quest_step_count_stile_title">How many steps on one side side of this stile?</string>
```

> Also, I'd move the second sentence after the question to either a description text above the input field, or to a dialog that opens when clicking "Other Answers" > "Differs for each side".

Yes, this seems sensible to me too!

peternewman: > It doesn't make much sense to me to only count the number of steps on one side rather the total amount of steps. Imagine it was mapped as a very short `highway=steps`, then it would also be the total amount. Where is this from, that the step count should be only on one side, but if it differs, the greater one?

Remember it can't/shouldn't be mapped as **one** very short `highway=steps`; it would need to be two otherwise the incline would be wrong for half of it!

If you're asking them what they do, please ask how they'd map my real-world example from before (which was something like 5:1 or 4:0 or something. If you map the total, its impossible to differentiate from a fairly average 2:2 stile despite the fact it's significantly further to pass a dog up or fall down from or whatever:
https://github.com/streetcomplete/StreetComplete/issues/3141#issuecomment-940992747

Your top user:
> https://www.openstreetmap.org/user/SomeoneElse https://www.openstreetmap.org/changeset/118712057 https://www.openstreetmap.org/node/9592964605/history 537

Has already answered previously (so assuming they haven't changed their mind since last year, at least 38% of all mapped steps on stiles are based on the tallest side only):

> The larger of the two step counts either side si what I've always done [SomeoneElse](https://wiki.openstreetmap.org/wiki/User:SomeoneElse) ([talk](https://wiki.openstreetmap.org/wiki/User_talk:SomeoneElse)) 23:20, 17 April 2021 (UTC)

https://wiki.openstreetmap.org/wiki/Talk:Tag:barrier%3Dstile#Is_stile_type_used_for_anything_or_potentially_useful.3F

> I haven't mapped [step_count](https://wiki.openstreetmap.org/wiki/Key:step_count)=* to any stiles, but I think it would make most sense to be the maximum number of steps on either side. In most (?) cases, there will be the same number of steps on both sides.--[Confusedbuffalo](https://wiki.openstreetmap.org/w/index.php?title=User:Confusedbuffalo&action=edit&redlink=1) ([talk](https://wiki.openstreetmap.org/wiki/User_talk:Confusedbuffalo)) 11:36, 16 April 2021 (UTC)

> https://www.openstreetmap.org/user/confusedbuffalo https://www.openstreetmap.org/changeset/112619350 https://www.openstreetmap.org/node/1184306845/history 2

Likewise this user (if they haven't changed their mind since commenting).

And indirectly I think SK53 probably has, as I can't see how you can have a step up and down and only be one step (unless it was effectively on a kerb).

> For instance a squeezer with step_count=1 will generally be readily accessible [SK53](https://wiki.openstreetmap.org/wiki/User:SK53) ([talk](https://wiki.openstreetmap.org/wiki/User_talk:SK53)) 13:12, 15 April 2021 (UTC)

> https://www.openstreetmap.org/user/SK53 https://www.openstreetmap.org/changeset/103863609 https://www.openstreetmap.org/node/8680778970/history 277

If I'm not reaching too much with SK53 then that's 58% of all mapped step counts covered.

westnordost: I'll write them each a message

westnordost: I wrote them a message and asked with how many steps they'd tag 

without specifically mentioning whether they'd count both sides or only one side.

**Result**: All answered something between 3 and 5 (most 4). Only one wrote that it only occured to him now that this is not really the total step count and that (the habit to) counting the steps on one side perhaps comes from stiles like this
 where most will probably see 1 step only.

So, I will document this accordingly in the wiki.

@matkoniecz please remove this wording from the question and put it as a clarification / description into the answer-part of the form. I.e. "how many steps does this stile have?" and description "Just the number of steps to reach the top". The note about "if differs, chose the higher value" is superfluous because how people count steps is different anyway (as said, some would tag the above example 3, some 4, some 5) and thus such detailed instruction will not lead to higher precision

westnordost (changes requested):
(see last comment)

app/src/main/res/values/strings.xml
matkoniecz: I think that I would strongly prefer to keep current wording, as it makes more clear that one counts steps from one side.

BTW, thanks for OSM Wiki edits.

westnordost: Yes, the clarification may remain, but should move to the description area, not in the question itself.

matkoniecz:
screen11


BTW, I am considering adding an option to suppress display of object type. For example this quest will be asked on stiles only, stiles are mentioned in question and should be - it does not make sense to display "Stile" bonus text below.

I am not adding special support in "can't say" like "converted into gate" - I can do this if there is significant use of that quest.

FloEdelmann:
@matkoniecz I think you forgot to push.

matkoniecz:
@FloEdelmann Yes, thanks

app/src/main/java/de/westnordost/streetcomplete/quests/step_count/AddStepCountStile.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        return mapData.nodes
+            .filter { stileNodeFilter.matches(it) && it.id !in excludedWayNodeIds }
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? =
+        if (!stileNodeFilter.matches(element)) false else null
+
+    override val changesetComment = "Add step count to a stile"
```
westnordost, line 40:
```suggestion
    override val changesetComment = "Add step count to stiles"
```

app/src/main/res/values/strings.xml
```diff
@@ -1455,5 +1455,8 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_wheelchairAccess_limited_description_public_transport">Partially means that there is no more than an ankle-high step to access it.</string>
     <string name="quest_wheelchairAccess_description_toilets">"A wheelchair sign like this might be visible if the restroom is accessible.
 Partially means that a wheelchair can enter and use the restroom, but no handrail or adapted hand basin is present."</string>
+    <string name="quest_step_count_stile_title">How many steps are at the side of this stile?</string>
+    <string name="quest_step_count_stile_hint">Just the number of steps to reach the top. If step count differs on each side, give higher number.</string>
+
```
westnordost, line 1460: FYI the strings in the strings.xml are now sorted by topic and then alphabetically, but I'll fix this after merge because I'll just use the same string as for the step count quest.

res/graphics/quest/steps_count_brown.svg
```diff
@@ -0,0 +1,74 @@
...
+     showgrid="false"
+     inkscape:zoom="6.8203125"
+     inkscape:cx="34.822451"
+     inkscape:cy="64"
+     inkscape:window-width="1920"
+     inkscape:window-height="1140"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg20" />
```
westnordost, line 31: will fix this after merge

westnordost (approved):

app/src/main/res/values/strings.xml
matkoniecz: Oh, right. I keep forgetting as I add text without ever visiting this file.

westnordost:
> BTW, I am considering adding an option to suppress display of object type. For example this quest will be asked on stiles only, stiles are mentioned in question and should be - it does not make sense to display "Stile" bonus text below.

That would introduce inconsistencies in the UI plus add more customization (= more code) to quests. If the duplication is bothersome, then better not mention the type of element in the question if the wording is not going to weird otherwise. I did this in a cleanup commit.

10992-osm:
As [here](https://github.com/streetcomplete/StreetComplete/issues/3141#issuecomment-895570957), I still think that

> How many steps are at the side of this stile?

sounds a bit unnatural.  I think it would be better as

> How many steps does this stile have?

10992-osm:
> note [7b16206](https://github.com/streetcomplete/StreetComplete/commit/7b162061ecf09d57f3b5eb27dcf93c80d619dc0e) that changed it to
> 
> ```
>     <string name="quest_step_count_title">How many steps are here?</string>
>     <string name="quest_step_count_stile_hint">Just the number of steps to reach the top. If step count differs on each side, give higher number.</string>
> ```

Ah, OK, I didn't notice that.

Issue #3053, opened 2021-07-12
Does it have an air pump / compressor ?


mnalis:
### General
**Rationale**: especially for bicycle users, flat tires are a problem, and way to inflate tires is needed. Fuel stations especially often offer compressed air pump, but many times its availability is not mapped in OSM. This quests aims to improve that. When tagged on fuel station, this information is also useful for motor vehicles owners.

Affected tag(s) to be modified/added: [compressed_air](https://wiki.openstreetmap.org/wiki/Key:compressed_air) and [service:bicycle:pump](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:pump)
Question asked: **Does ____ have an air pump / compressor ?**

### Checklist
Checklist for quest suggestions (see [guidelines](https://github.com/streetcomplete/StreetComplete/wiki/Adding-new-Quests-to-StreetComplete):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary
   Located outside for majority of quests. Only sometimes in cases of `shop=bicycle` it might be necessary to enter the shop (if pump is not located outside).
- [x] 💤 Not an overwhelming percentage of quests have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of map data (Worth the effort)

### Ideas for implementation

**Element selection:** 
The question should be asked on following nodes/areas:
* [amenity=fuel](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dfuel)
* [shop=bicycle](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbicycle)
* [amenity=bicycle_repair_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station)

when they contain neither [compressed_air](https://wiki.openstreetmap.org/wiki/Key:compressed_air) nor [service:bicycle:pump](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:pump) tags.

(additionally, we may want to skip asking the quest if there is [amenity=compressed_air](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcompressed_air) or (another) [amenity=bicycle_repair_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station) very close by)

**Metadata needed:** 

[...]

PR #3934, opened 2022-03-31
Air pump quest(s) (manual pump / air compressor)

mnalis:
Closes: https://github.com/streetcomplete/StreetComplete/issues/3053

TODO:
- [ ] needs an vector icon (does anybody have idea where to find some; or is artistically inclined?)
 Currently it uses same icon for both manual bicycle pump and air compressor, but that can be changed if needed.
- [ ] should I add resurvey too? (and if so, suggestions on intervals?)
- [ ] better English texts? (esp. regarding https://github.com/streetcomplete/StreetComplete/issues/3053#issuecomment-1076823076)

Comments / suggestions welcome!


matkoniecz:
icon:

maybe some kind of wind/air movement symbol? But may be too cryptic.

screen04
screen05
screen06
screen07

Or just a wheel or part of it?

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
```diff
@@ -0,0 +1,36 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquests.Tags
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddBicyclePump : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
+        nodes, ways with
+         amenity = bicycle_repair_station or shop = bicycle
```
matkoniecz, line 17: For shop=bicycle [OSM Wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbicycle#Additional_keys) claims: 

> If a bicycle shop is not tagged with a key from the above list, it means that its status is unknown

so it should be fine. Maybe linking https://wiki.openstreetmap.org/w/index.php?title=Tag:shop%3Dbicycle&oldid=2305390#How_to_map in a comment would be a good idea?

Note that https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station takes no position on that, is `amenity=bicycle_repair_station service:bicycle:pump=yes` implying that it is sole service? I asked on https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dbicycle_repair_station#Is_presence_of_one_service_tag_implying_no_for_other.3F

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
```diff
@@ -0,0 +1,36 @@
...
+    """
+    override val changesetComment = "Add whether bicycle pump is available"
+    override val wikiLink = "Key:service:bicycle:pump"
+    override val icon = R.drawable.ic_quest_air_pump
+    override val questTypeAchievements = listOf(BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_air_pump_bicycle_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with compressed_air or amenity = compressed_air or service:bicycle:pump")
```
matkoniecz, line 29: I would highlight all `amenity=bicycle_repair_station` and all `amenity=fuel` in addition to that.

This would make easier to handle cases with multiple repair stands.

(are standalone `amenity = compressed_air` without name rendered properly?)

matkoniecz (changes requested):

matkoniecz:
Note also https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dfuel#compressed_air.3Dyes.2Fno - tagging availability of compressed air on `amenity=fuel` has some opposition

FloEdelmann:
A good website for simple CC-0 SVGs is https://www.svgrepo.com/. Search terms that may be interesting: "air pump", " air compressor", "air", " balloon".

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
mnalis: > (are standalone `amenity = compressed_air` without name rendered properly?)

If by "properly" you mean as an magenta dot, then yes, for example https://www.openstreetmap.org/node/2496359080

mnalis: good point about extra highlighting, added

westnordost:
Don't worry about icons, I'll create some. Just use some placeholders.

westnordost:
I'll do

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
mnalis: Thanks, I've documented in comment the meaning of undefined `service:bicycle:pump`.

Also I'll reply on wiki there shortly; in my experience, the whole point of introduction `X:Y=yes/no` instead of `X=Y1;Y2` was mostly just to fix that uncertainty, e.g.:
 
- using `service:bicycle:pump=yes`+`service:bicycle:rental=no` **only** defines that:
  -  yes, bicycle pump exists here
  -  no, bicycles are not rented here
  -  and says **nothing** about any and all other possible services (there are explicitly unknown)
- while traditional (example tagging) `service=bicycle_pump` was very ambiguous -  it did say that bicycle pump exists there; but it was not marked that:
  - it is known that bicycles are not being rented there
  - is the lack of other values (like "are bicycles washed here for a fee?") simply because there have not been surveyed, or because that they are known to not exist.

mnalis:
Apart from missing icons, this looks as good as I could make it. Ready for review.

app/src/main/res/values/strings.xml
```diff
@@ -654,6 +654,9 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_bikeParkingCapacity_title">"How many bikes can be parked here?"</string>
     <string name="quest_bikeParkingCapacity_hint">"Note that most normal stands can be used from both sides to park each one bike."</string>
 
+    <string name="quest_air_pump_bicycle_title">"Is there a working bicycle air pump available to general public?"</string>
```
matkoniecz, line 657: What about `amenity = bicycle_repair_station access=private` with an air pump?

Maybe at least exclude tagged as `access~private|no`? ( though it is [really rare](http://overpass-turbo.eu/?q=LyoKVGhpcyBoYcSGYmVlbiBnxI1lcmF0ZWQgYnkgdGhlIG92xJJwxIlzLXR1cmJvIHdpemFyZC7EgsSdxJ9yaWdpbmFsIHNlxLBjaMSsxIk6CsOiwoDCnGFtxI1pdMSaPcSYaWN5Y2xlX3JlxKNpcl9zdMSUaW_EjmFuxJdhY2Nlc3M9cMS3dsSUxJ5nbG9ixLzFiMKdCiovCltvdXQ6anPFpF1bdGnFjMaBxoMyNV07Ci8vxI_ElMSdciDFmnN1bHRzCigKICDGlCBxdcSSxJrEo3J0IGZvcjogxYjFisWMbsWOeT1ixZPFlcWXxZnFm2HFncWfxaHGisWkIMWmxajFqsWsxa7FsGnFssSVxbrGoyBub2RlWyLFi8WNxY8iPSLGvMWUxZbFmMWaxZzFnsWgxaLFpCLGiMeax4vFrceeIseOx5Blx6vGksakd2F5x5nHm8a4x53Hn8ehxr7HpMeBx4PHqMeGbserx7vHrnPHsMeyxbPHtceTxZpsx6lux7vGt8a5x7DIgMejx4DHpseEyJPIiMetxavHr8efyI3ElciPKcaSxqbHjm7Grsabxp3GnwrGjcSYx5Z5xpI-xpLIsXNrZcS9cXQ7&c=BPgZHWMPGC&R) )

(there would be still problem with private ones that have pump - they would get tagged `service:bicycle:pump=no` but it should be ridiculously rare)

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
```diff
@@ -0,0 +1,48 @@
...
+    override val questTypeAchievements = listOf(CAR, BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_air_pump_compressor_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
+            nodes, ways with
+            compressed_air
+            or service:bicycle:pump
+            or amenity ~ compressed_air|fuel|bicycle_repair_station
```
matkoniecz, line 40: https://www.openstreetmap.org/node/5603191863 was not shown, but maybe due to exceeding maximum distance

likely not actionable here

app/src/main/res/values/strings.xml
mnalis: The general idea of that _"...pump available to general public"_ phrase was primarily intended to exclude situations where e.g. bicycle repair shop has an pump for their own use, but their willingness to let you use it depends on how likeable a person asking is  - see https://github.com/streetcomplete/StreetComplete/issues/3053#issuecomment-1075853599. That's the best wording I've came with to indicate that `yes` answer should not depend on their goodwill. Improvements for wording are always welcome if someone has a better suggestion?

Anyway, I'm not sure that those `21` mapped cases worldwide are a problem requiring special handling; probably leaving a note explaining the situation so it can be mapped better in other editor is just fine in those cases. 
Note that for `access=customers` I'd leave the quest on - you should not be using that facility / be in that area if you're not a customer in any case (as warned in the intro and with signs on-site); and if you _are_ a customer who is using that facility, it is accessible to you to mark it, and is useful information for others  (e.g. _"sure I'll buy a drink/snack to be able to pump my tire here"_). 
But if people think it is a big enough problem, I could make another translation string to be asked for places marked `access=customer` which is less confusing in that case (e.g. _"Is there a working bicycle air pump available to all customers?"_)

matkoniecz: > probably leaving a note explaining the situation so it can be mapped better in other editor is just fine in those cases.

The problem is that mapper will answer correctly "no" and it will be tagged incorrectly. It is not clear that leaving note is necessary to avoid damaging data.

Maybe have separate questions for bicycle shop and repair station?

`access=customers` can be skipped I guess.

mnalis: Ah, good point about possible data damage if users don't leave note in that case. I've excluded `access~private|no` now. I don't see that separate questions for bicycle shop and bicycle repair stations would help here; after all it is quite possible that non-bicycle-only shop (e.g. Decathlon) has bicycle repair station in front (and may or may not enforce some rules on its use).

Perhaps a better text than _"Is there a working bicycle air pump available to general public?"_ would still help for customers case. I can think of _"Is there a working bicycle air pump available to everybody having access to this facility?"_, but that sounds just a bit too unwieldy to me. Anybody have nicer suggestion?

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
mnalis: Might be a distance, or the fact that the `amenity=fuel` area overlaps with that `amenity=compressed_air` node. 
When they are closer and not overlapping, they get shown OK (e.g. https://www.openstreetmap.org/way/342850258) so I think this quest logic is OK here; so there is likely nothing I could do here to fix it?

matkoniecz:
BTW, what you think about other quests about other properties for bicycle repair stations? Potentially one may ask also about other properties.

(and if having separate quests for `service:bicycle:chain_tool` and `service:bicycle:stand` and maybe other is unwanted it is worth noting this now)

app/src/main/res/values/strings.xml
```diff
@@ -654,6 +654,9 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_bikeParkingCapacity_title">"How many bikes can be parked here?"</string>
     <string name="quest_bikeParkingCapacity_hint">"Note that most normal stands can be used from both sides to park each one bike."</string>
 
+    <string name="quest_air_pump_bicycle_title">Is a working bicycle air pump available to everybody having access here?</string>
```
matkoniecz, line 657: Maybe "Can anyone use an air pump provided here?"

Or something similar?

(not a native speaker speaker)

matkoniecz (approved):
Icon needs to be added, ideally air pump question would be simplified SOMEHOW.

matkoniecz (commented):
I wanted to keep it in neither "changes requested" nor "approved" state but I see no way to achieve this.

mnalis:
> BTW, what you think about other quests about other properties for bicycle repair stations? Potentially one may ask also about other properties.
> 
> (and if having separate quests for `service:bicycle:chain_tool` and `service:bicycle:stand` and maybe other is unwanted it is worth noting this now)

It's interesting to some; but they do seem to have about an order of magnitude less uses on OSM than air pump - which coincides with my usage (most often I need just a pump). Perhaps seeing how well this quest gets used might give indication of interest to other related quests.

<details>

MTB folks in my experience seem to find `chain_tool` as one of most essential tools, while for example I've never had a chain snap in my life - but all I do is pedal on mostly flat terrain, and not downhill with vegetation jumping at me. Also stand is (as I understand it) important for many repairs if you have hydraulics for disc brakes, but my V-brakes equipped trekking bike can be easily turned upside down to do mostly any on-the-field repair, so while nice to have, it is no biggie for me.

So I'd personally usually need (in order of importance): air pump, patch kit, adjustible wrench, tire levers, hex keys, screwdrivers.  
</details>

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
```diff
@@ -0,0 +1,49 @@
...
+       nodes, ways with
+       amenity = fuel
+       and (
+           !compressed_air and !service:bicycle:pump
+           or compressed_air older today -6 years
+       )
+       and access !~ private|no
+    """
+    /* if service:bicycle:pump is undefined, nothing has been said about its existence;
+     * see https://wiki.openstreetmap.org/wiki/Tag:shop=bicycle#Additional_keys */
```
westnordost, line 27: I do not understand this comment. Why is this quest not asked if `service:bicycle:pump` is set to anything (no / yes)?

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_air_pump_compressor_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
+            nodes, ways with
+            compressed_air
+            or service:bicycle:pump
+            or amenity ~ compressed_air|fuel|bicycle_repair_station
+        """)
```
westnordost, line 42: Why highlight fuel stations regardless of if they have compressed air or not? This is not consistent. On the one side, you highlight **any** map feature with `compressed_air` (even set to `no`), on the other hand you highlight map features that might or might not have compressed air. Additionally, you highlight any manual bicycle pumps but the question is not about manual air pumps. This too is misleading.

```suggestion
        getMapData().filter("nodes, ways with compressed_air = yes or amenity = compressed_air")
```
... will highlight any feature with compressed air or separately mapped compressed air.

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
```diff
@@ -0,0 +1,45 @@
...
+    override val questTypeAchievements = listOf(BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_air_pump_bicycle_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
+            nodes, ways with
+            compressed_air
+            or service:bicycle:pump
+            or amenity ~ compressed_air|fuel|bicycle_repair_station
```
westnordost, line 37: same as for the other... though assuming that if not specified otherwise, a bicycle repair station has a working pump. (you may remove this if you think this cannot be assumed)
```suggestion
            nodes, ways with
            service:bicycle:pump = yes
            or amenity ~ bicycle_repair_station and service:bicycle:pump != no
```

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
matkoniecz: > though assuming that if not specified otherwise, a bicycle repair station has a working pump

this one was actually added after I proposed it: it was intended to help in cases of places with multiple repair stations by showing all of them for better orientation

westnordost: The highlighting must be consistent. Then I suggest to highlight exactly what the quest filter is asking for: any bicycle shops and any bicycle repair stations, regardless whether they have compressed air, a pump or not

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
mnalis: sorry, ended in wrong file; it was meant to explain that undefined `service:bicycle:pump` does **not** imply that there is none (cf. https://github.com/streetcomplete/StreetComplete/pull/3934#pullrequestreview-927120024). Fixed it in e747679d7331e5b3e463fe955d2f2091deadca0d now.

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
mnalis: I guess you're right on showing manual bicycle pumps, I probably see things too much from bicyclist point of view, so copy/pasting that code seemed logical. But now that you mention it, even if fuel station has manual bicycle pump, it would probably be on separately marked bicycle_repair_station which is covered by the other quest. So I've corrected that part in c89e209dc136330031ef2c4ef20705bd2e868cee.

app/src/main/res/values/strings.xml
mnalis: it turns out that `access=customers` is invalid usage here (e747679d7331e5b3e463fe955d2f2091deadca0d), so currently suggested string is:
`Is a working bicycle air pump available here for everyone?`

That looks passable to me, but I'd be happy for improvements, provided that  that keeps following information:

- it is about `bicycle air pump` (and not some other pump, like oil pump for hydraulic bicycle brakes, or any other pump not usable for bicycle tires) 
- it is in working condition (important, as in some countries like my own it cannot be assumed at all :cry: )
- it is available to any member of general public (and not only to customers or depending on goodwill of owner etc.)

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
mnalis: It makes sense to me to show it on _"any bicycle shops and any bicycle repair stations"_ as you say @westnordost, but I'd **_also_** highlight it on _nearby_ places which are **_explicitly_** marked as having compressed air or a bicycle pump, even if they are marked on some **_other_** nearby feature - otherwise the mapper could easily confuse one for another. Those happen relatively often: for example air pumps marked on [bicycle parkings](https://www.openstreetmap.org/node/8452855617), [vending machines](https://www.openstreetmap.org/node/1610293356), [bicycle rentals](https://www.openstreetmap.org/way/431729735), [picnic shelters](https://www.openstreetmap.org/node/8754095941), [cafes](https://www.openstreetmap.org/node/7200927563), [sport centers](https://www.openstreetmap.org/way/277433336) and [many more](https://overpass-turbo.eu/s/1hlj).

In that light, would something like 7ff7f4d2841e9969b18e89d0802713d18f281bf9 then be acceptable solution?

app/src/main/res/values/strings.xml
mnalis: it turns out that `access=customers` is invalid usage here (e747679d7331e5b3e463fe955d2f2091deadca0d), so currently suggested string is:
`Is a working bicycle air pump available here for everyone?`

That looks passable to me, but I'd be happy for improvements, provided that it keeps clear the following information:

- it is about `bicycle air pump` (and not some other pump, like oil pump for hydraulic bicycle brakes, or any other pump not usable for bicycle tires) 
- it is in working condition (important, as in some countries like my own it cannot be assumed at all :cry: )
- it is available to any member of general public (and not only to customers or depending on goodwill of the owner etc.)

arrival-spring: (native speaker here)
How about "Is there a working bicycle air pump available for everyone here?" (although bicycle air pump sounds a bit odd as I would just call it a bicycle pump)
and "Is there an air compressor available here?" reads better

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
westnordost: ok

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddBicyclePump.kt
westnordost: ok

app/src/main/res/values/strings.xml
mnalis: thanks @arrival-spring , I've changed them as you suggested then: bf5472f694346494cb81b7ee72a2f35731bd2d33! 

I did actually try similar string for air compressor before, but in the end, while it sounded OK to me, it looked highly suspicious to have both `there` _and_ `here` referring to same location in same sentence, so I decided to remove one of them. So it's great to have your confirmation that it is actually OK.

I take solace in the fact that any English-speaking person would have even harder time learning proper Croatian :laughing:

mnalis:
That should be it as far as I can tell; only icons remain be created by @westnordost . 

Related (especially for all of us non-native English speakers): if anyone has problems with strings, I've stumbled (of course, only _just_ after I didn't need it anymore for this quest :smile:) upon online form https://www.gingersoftware.com/products/sentence-rephraser which offers several different ways to express the same sentence, which might give some ideas.

westnordost:
added the graphics to master

westnordost:
I added the graphics to master, you need to integrate them

mnalis:
> I added the graphics to master, you need to integrate them

Oh, no problem, it seems I've [misunderstood earlier](https://github.com/streetcomplete/StreetComplete/pull/3934#issuecomment-1084864171), hence the delay. 

I've now updated references to icons, and it seems to be working fine when applied to current master:

small_Screenshot_20220410_160732_de westnordost streetcomplete mn debug

So should be ready to merge as far as I can tell.

westnordost (approved):

RubenKelevra:
Cool! But I feel like one info is missing: If you can use the air compressor for bicycles/motorcycles.

The more modern "automatic" fill stations are plastered with warning stickers that they are not safe to use for bicycles/motorbike tires.

The older manual ones on the hand are safe to use, as you can just tap them in very short bursts to avoid overfilling.

So the info is quite valuable to collect, if you want to find a gas station which can fill up your tire :)

mnalis:
@RubenKelevra Thanks! Yes, it is one of the things considered for possible followup quests, see for example https://github.com/streetcomplete/StreetComplete/issues/3053#issuecomment-1075848352. 

But as with https://github.com/streetcomplete/StreetComplete/pull/3934#issuecomment-1086431083, I think it would be good to wait a little to see how much this main airpump quest gets used (and what are main issues with it), before choosing adding quests to provide further details about it.

RubenKelevra:
Yeah that makes sense :)

Ah and some are free and some require to pay with a coin, like 50 Cent or 1 Euro which is kinda strange to me. But yeah.

westnordost:
I read up about these automatic air compressors.

It seems like the main issue with those automatic ones where you select a target pressure instead of having a +/- button is that they are configured to for car tyres. They work in a way that they put some air pressure in, then measure the pressure, then either put more pressure in or let out a bit, then measure again etc. Now, car tyres have a much bigger volume, so if the compressor overshoots its target, that's fine. Bicycle tyres however have a much smaller volume, so overshooting the target can lead to momentary overpressure that might make the tyre explode (apparently the felloe is a weak spot).

Most people seem to have not made any bad experiences with this and some people write that (bicycle) tyres can withstand large overpressure because just e.g. cycling over a kerb may already induce large overpressure. Others write that not the tyre but the felloe (especially aluminium ones that had been worn dorn by the brakes) is the weak spot in case of overpressure. In any case, the concern is real enough that many fuel stations put up a sign to forbid cyclists to use them.

Bottom line: I think it might be more valuable to **record the type of air compressor** (manual +/-, automatic with pressure targt etc...) than recording whether there is a 🚳 sticker.

RubenKelevra:
@westnordost yeah sure. The type of compressor is also interesting, but there are however automatic types of compressors which just measure much more frequently and thus can also handle bicycle tires.

I filled my MTB for years on an automatic compressor like 4 times a week and never had any overshoot.

So it's more about the brand/model than the general type (or the settings made by the manufacturer). I feel like if there's a sign, we should tag it as an access restriction – as it's simply not allowed.

If it's allowed, I don't think there's much of an issue doing this (just judging by my experience).

PR #3947, opened 2022-04-06
doc: fix 404 link in edit handle presentation

rugk:
It now links to https://lstu.fr/sc-data-flow (changed _ to - so the URL
stays as nice).
The URL behind that is now not the PDF anymore which does not exist
anymore, but
https://raw.githubusercontent.com/streetcomplete/StreetComplete/master/res/documentation/overview_data_flow.svg


rugk:
BTW as the diff of that thing is of course impossible, as it is binary, I also thought of saving it as a plain XML, that is `.fodp`  format, however, that is 3.8MB large as it is now not ZIP compressed anymore… :shrug: 

Feel free to ask for that, I can easily change it and commit it, if you want…

westnordost:
Thank you! I noticed the PDF was pretty broken, that's why I reexported as SVG.

Issue #2706, opened 2021-03-31
What kind of entrance is this?


1ec5:
### General
Affected tag(s) to be modified/added: [`entrance`](https://wiki.openstreetmap.org/wiki/Key:entrance)
Question asked: ~~**Who can use this building entrance?**~~ **What kind of entrance is this?**

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
    * More specific tags than `entrance=yes` would allow routers to route to the correct side of the building.
    * Some renderers like osm-carto distinguish `entrance=main` from other `entrance=*` values.
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary
- [ ] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
    * This depends on whether armchair mappers have been mapping only the main entrances or all the entrances they can see from aerial imagery.
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

**Element selection:** 

`entrance=yes` that is one of multiple `entrance=*` nodes along a `building=*` way. Maybe require the building to cover enough area or filter out things like `building=garage` to avoid spamming the user about insignificant buildings.

**Metadata needed:** 

**Proposed GUI:**

Select one:

* Main entrance – `entrance=main`
* Side or back entrance – `entrance=secondary`
* Entrance for staff only – `entrance=service` `access=private`
* Entrance for deliveries only – `entrance=service` `access=delivery`
* Exit only – `entrance=exit`
* Emergency exit – `entrance=emergency`

matkoniecz:
"Everyone (main entrance) – entrance=main" 

What about entrances to residential buildings where `entrance=staircase` may fit or houses where main one will be still private?

Or is it covered by " filter out things like building=garage to avoid spamming the user about insignificant buildings."?

andrewharvey:
How about just "What type of entrance is this?" instead of trying to conflate entrance= and access=?

1ec5:
Yes, I was assuming that the quest could generally filter out buildings that only have private entrances. Even so, a house’s front door could still validly be `entrance=main` even as `access=private` would apply. So maybe the only problem is the wording. Would it be weird to ask, “Which of the building’s entrances is this?” and answer simply that it’s the main entrance?

[...]

Issue #2464, opened 2021-01-07
Give priority to address (housenumber) quest


mnalis:
**Use case**
To be able to get "Address" (housenumber) quest, one currently first needs to solve "building type" quest. The reasoning is (as far as I get from other issues) is that some building types (garages, sheds...) do not have house numbers, so it makes no sense to ask for them there. This is highly unoptimal.

While valid observation that some building do not have address, there are few issues with this fact blocking the address quest:
- the "Address" quest is (or should be) a **high priority**. One of the main uses of map is to find something on it, and that something is in majority of cases defined by its address (the other common way is by searching for "closest amenity of some type"). 
- the "Building type" quest is (or should be) of a very **low priority**. Not only is there much lower usability of such data (most being covered by separate landuse polygons), but there is confusion/dispute between [building](https://wiki.openstreetmap.org/wiki/Key:building) and [building:use](https://wiki.openstreetmap.org/wiki/Key:building:use) (with most people mapping it the "wrong" way), also it is often hard to answer correctly (eg. there is often no way to know what was original purpose of the building when it was built without visiting cadastre), and even more often mapped incorrectly (eg. `building=house` when it should be `building=apartments` etc.), and even more problems because purpose is sometimes shared (shop+apartments or shop+house etc). Actually, if it was not being prerequisite for "Address" quest, it should probably be disabled by default as it is arguably likely to pollute the database with incorrect data when used by average mapper.
- the issue is additionally aggravated  by the fact that even after solving the "Building type" quest, and having auto-upload, the "Address" quest does not show immediately (like for example steps-related quests do), requiring another resurvey at the later time (#1550, #1510).

**Proposed Solution**

I propose to **make address quest a priority**. It is one of the most useful quests SC offers, and effectively disabling such high priority quest by low priority quest is counterproductive.

One way to do this would be to always ask for address quest first, and have an additional answer `This building does not have an address (it is garage or shed or similar)` (which can then let user choose building type, or simply mark it as having no address).

matkoniecz:
> While valid observation that some building do not have address, there are few issues with this fact blocking the address quest:

I see no other sane way of handling `building=yes`.

Ask housenumber for every building and leave trail of pointless notes? Merge building type and address quest into one?

matkoniecz:
> One way to do this would be to always ask for address quest first, and have an additional answer This building does not have an address (it is garage or shed or similar) (which can then let user choose building type, or simply mark it as having no address).

[...]

PR #3949, opened 2022-04-07
Entrance into building

### LINKED ISSUE #3941: What kind of entrance is here? (where no entrance is marked)
### General
Affected tag(s) to be modified/added: [entrance](https://wiki.openstreetmap.org/wiki/Key:entrance) / [noexit](https://wiki.openstreetmap.org/wiki/Key:noexit)
Question asked: **What kind of entrance is here?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose (also, it may enable #3064)
- [x] 🕓 Applies to a reasonable number of map data (Worth the effort) (hundreds in just Kraków, Poland - see https://overpass-turbo.eu/s/1hkN )

### Ideas for implementation

**Element selection:** find nodes shared between building and linear footway (maybe add also roads?), such as https://www.openstreetmap.org/node/9344552612

Skip cases where node is shared by building passage such as https://www.openstreetmap.org/node/5832373046

**Metadata needed:** None

**Proposed UI:**

Include option "there is no entrance here" that would apply `noexit=yes` There are cases of footway leading into wall which may be tagged as footway terminating on shared node with building (though I would not use such tagging)

matkoniecz:
fixes #3941

requires an icon (@westnordost  - are you interested in making one? Or should I try drawing a door?)

I am open to changing descriptions, but current one seems OK to me

screen09

Draft as an icon is missing.


app/src/main/res/values/strings.xml
```diff
@@ -1268,6 +1268,17 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_smoothness_obstacle_hint">Please provide a general answer for the smoothness. Single obstacles like kerbs, speed bumps or cattle grids can be ignored. Repeated or unexpected obstacles should count towards the answer.</string>
     <string name="quest_smoothness_hint">🚲, 🚗, 🚙, … indicate for which vehicles the surface should be okay to use. What matters is the big picture rather than single damages.</string>
 
+    <string name="quest_building_entrace_title">What kind of building entrance is it?</string>
```
FloEdelmann, line 1271:
```suggestion
    <string name="quest_building_entrace_title">What kind of building entrance is this?</string>
```

westnordost:
Uff, that's a lot of options...

How can it be determined from the outside if something is an emergency exit? The appropriate sign would rather be located on the inside, no? Same really with "exit only". If it is not a public building... and even if it is actually, there may be a door which is locked, but that doesn't mean that it is only an exit. It means that it is locked. Also, we don't want to encourage users to try to check if they can open doors (or peek inside). So, my question is, also in respect to the quest guidelines:

What is the useful purpose for recording the type of entrance and in such detail?

HolgerJeromin:
=main is quite interesting indeed (and rendered even on osm-carto). But the quest needs to support at least the most needed answers...

We could merge option 3,4,5 into "service". An entrance=exit is a bit an entrance=service at least from the outside (see Tobias). But this is drops a bit of information.

mnalis:
I could probably (in most situations) from outside differentiate between some (but probably not all) of them...

<details>
- dead end (obvious)
- garage (specific type of rollup / doors, although closed shop after hours might look similar for theft protection)
- maybe exit-only (closed door with no handles or whatever; unknown from outside if "regular" or "emergency only")
- maybe shop (reasonably well explained in menu entry itself if there is both entrance in the building and different entrance for the shop itself, if situated inside same building)
- main entrance (maybe: if this is a shop only building, is it main or shop? what if shop has multiple entrances with different qualities?)
- service (probably; I guess I'd mark main shop entry as "main" and entrance with "no unauthorized personnel" signs as "service"? Or should some of them be "shop" instead ?)
</details>

But it seems I'd need to dive in the wiki beforehand to get a firmer grasp about what it what exactly. Perhaps reducing the number of entries to most popular types would fix that?


As for the usefulness, I primarily see the usefulness in order to **_locate entrances to bigger buildings_** so next user don't have to work half a mile around a mall or big hospital building to find the regular entrance. So it would be quite useful in marking connection of footway and building with some key indicating whether:

- it is regular entrance (`entrance=main`? `entrance=yes`? I mean entrance where I, as a member of general public, may assume I'll be able to enter the building normally)
- something else (either mismapped connection so it needs to be `noexit=yes`, or more probably some of two exit-only variants, or `garage` or `service` where general public probably won't be able to enter normally) 

That yes/no quest should be quickly and easily answerable by the mapper, but seems to lack explicit tagging for the `no` case (and requiring user to distinguish between that plethora of all possible _"not an main entrance, but is not a `noexit=yes` either"_ variants as in current PR might be an overkill, as it likely cases much thinking and uncertainty for user - I hate such quests and tend not to enable them, like for example _"parking fee"_ quest  in many places).
  
Another problem is, in vast majority of cases I've seen in my area (so might be quite different elsewhere), there is no footway mapped connected to the building at all, so this quest wouldn't popup, and I have to see there is no entry mapped (in OsmAnd I run in parallel with max details, as SC does not show those I think) and and a note manually (as I do currently).

mnalis:
Also, (if it were to go in), might it be refined not to trigger when footway actually _**passes though**_ the building? Those are sometimes (but far from always, marked with [tunnel=building_passage](https://wiki.openstreetmap.org/wiki/Tag:tunnel%3Dbuilding_passage) and similar tags), and I don't think any of the currently offered answers match that use case.

 e.g. https://overpass-turbo.eu/s/1hl5 mentioned before shows a several such passages around https://www.openstreetmap.org/node/959409978#map=19/45.79891/15.89925&layers=N

westnordost:
@mnalis well that would be a mis-tagging then and it's good that the user needs to leave a note in that case.

matkoniecz:
> might it be refined not to trigger when footway actually passes though the building? Those are sometimes (but far from always, marked with [tunnel=building_passage](https://wiki.openstreetmap.org/wiki/Tag:tunnel%3Dbuilding_passage) or [covered=*](https://wiki.openstreetmap.org/wiki/Key:covered) or similar tags), and I don't think any of the currently offered answers match that use case (the closest is probably entrance=main but that is usually highly misleading too).

This should be handled by https://github.com/streetcomplete/StreetComplete/blob/c403fc9e0e031c83d09ad0e85ce16b025a8bab69/app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt#L28 if tagged well, otherwise user should leave note (hopefully it is clear?)

> How can it be determined from the outside if something is an emergency exit? The appropriate sign would rather be located on the inside, no? Same really with "exit only". If it is not a public building... and even if it is actually, there may be a door which is locked, but that doesn't mean that it is only an exit. It means that it is locked. Also, we don't want to encourage users to try to check if they can open doors (or peek inside).

Often it is clear also from outside (for example, emergency exit with marking about not blocking it or sign visible from outside). If it is not and it is not identified `=yes` would be suitable. Maybe its description can be tweaked to better cover this.

> What is the useful purpose for recording the type of entrance and in such detail?

For more complex buildings distinguishing between main entrance and service entrance is useful.

In could be for example used also to detect that main entrance is actually not yet mapped, after main entrances are mapped the quest asking about entrance code / flat range can be asked (for buildings with multiple main entrances, in at least some countries)

>  there is no footway mapped connected to the building at all, so this quest wouldn't popup

Yes, it relies on such detailed mapping being present already (which in some places IS present, enough to generate significant volume of quests. Berlin or Kraków has thousands of such quests).

HolgerJeromin:
We are discussing problems with exit entrances...

>> there is no footway mapped connected to the building at all, so this quest wouldn't popup

> Yes, it relies on such detailed mapping being present already (which in some places IS present, enough to generate significant volume of quests. Berlin or Kraków has thousands of such quests).

...but i do think that these thousands quests are >99% yes or main entrances and NOT the problematic "exit" ones.

westnordost:
>> What is the useful purpose for recording the type of entrance and in such detail?

> For more complex buildings distinguishing between main entrance and service entrance is useful.

I understand that the distinction between a main entrance and some other entrance is useful. But the list from the screenshot is much longer. Does anything speak against limiting this list to
- main entrance
- entrance to staircase
- other entrance
- no entrance (dead end)

matkoniecz:
Main negative here is that it is a lost opportunity to specify in more detail for cases where something more detailed than `entrance=yes` can be specified.

Especially as https://github.com/streetcomplete/StreetComplete/issues/2706 is impossible

> missing from screenshot even though it is 2nd most used

This got removed on attempt to trim the list and because it is a bit unspecific (there could be a main staircase entrance, secondary staircase entrance, service staircase, even emergency exit staircase)

westnordost:
> lost opportunity to specify in more detail

The way I see it, it is only a missed opportunity if the additional detail in data produced is actually useful. I don't see much value in differenciating between an exit only, or emergency exit only, or service entry only or "other" entry only - additionally to the verifiability issues.

> This got removed on attempt to trim the list and because it is a bit unspecific (there could be a main staircase entrance, secondary staircase entrance, service staircase, even emergency exit staircase)

TBH I thought the ability to put `entrance=staircase` was the very root motivation for this quest, hence the most valuable information to ascertain, because then and only on staircases, another quest may ask about the housenumber/flat number.
With the same argumentation as in the sentence cited, you could also say that `entrance=service` is unspecific, as it could be an servicestaircase or a normal service entrance.

mnalis:
>  I don't see much value in differenciating between an exit only, or emergency exit only, or service entry only or "other" entry only - additionally to the verifiability issues.

To me, `entrance=service` is the one with big sign `Authorized personnel ONLY` and thus is as good as a dead end for normal user (i.e. 99.99% of the users, who are not an employee there). However "other entry" (`entrace=secondary?` _is useful_ for normal user, for example if access to building is open there, but only on some occasions (like when there is a concert there or something). 

> TBH I thought the ability to put entrance=staircase was the very root motivation for this quest

Regarding that, I find [entrance=staircase wiki](https://wiki.openstreetmap.org/wiki/Tag:entrance=staircase) sorely lacking in content (one has to scavenge various talk pages and links in footnotes to get idea what it is about, and those seem to have conflicting information)

Idea: perhaps, there could be:
- one quest which has **_only 4 easily discernible values_** (as @westnordost suggests in https://github.com/streetcomplete/StreetComplete/pull/3949#issuecomment-1092765898) and triggers on `footway connects to building`, and
-  another followup quest (disabled by default?) for those wanting to map more details, which would trigger on all `entrance=yes` and ask what is detailed situation, offering all possible values (and thus requiring much more work and thinking from the mapper)?

matkoniecz:
> another followup quest (disabled by default?) for those wanting to map more details, which would trigger on all entrance=yes and ask what is detailed situation, offering all possible values (and thus requiring much more work and thinking from the mapper)?

Not viable, see https://github.com/streetcomplete/StreetComplete/issues/2706 (some `entrance=yes` have no better tagging...)

matkoniecz:
> Regarding that, I find [entrance=staircase wiki](https://wiki.openstreetmap.org/wiki/Tag:entrance=staircase) sorely lacking in content

I expanded it a bit
> With the same argumentation as in the sentence cited, you could also say that entrance=service is unspecific, as it could be an servicestaircase or a normal service entrance.

I would not consider "service staircase entrance" and "service entrance" without staircase as meaningful. Obviously, it is not the most important way to decide tagging, but note that https://wiki.openstreetmap.org/wiki/Proposed_features/entrance is without `staircase` value.


> TBH I thought the ability to put entrance=staircase was the very root motivation for this quest, hence the most valuable information to ascertain, because then and only on staircases, another quest may ask about the housenumber/flat number.

Yes, "another quest may ask about the housenumber/flat number." is primary motivation.

But note that `entrance=main` on `building=apartments` would continue to qualify Though, I guess, there is higher risk of people not using `main` for entrances like this one (even with the current description):



Which would be reduced if `staircase` answer is available.

mnalis:
> Not viable, see #2706 (some `entrance=yes` have no better tagging...)

But [maybe they do](https://github.com/streetcomplete/StreetComplete/issues/2706#issuecomment-817202174), though? I basically see `entrance=yes` as one of the **three** (more refined) [entrance=*](https://wiki.openstreetmap.org/wiki/Key:entrance) keys:
- `entrance=main` for primary entry+exit, available to **_everyone at all times_** (unless the place is closed, or additional access indicated explicitly, of course)
- `entrance=secondary` for non-primary entry+exit (i.e. available only in **_some_** cases)
- `entrance=exit` for exit-only (no entry possible)

All other values (or so it seems to me) basically derive from those, so:
- `entrance=service` is often basically `entrance=secondary`+`access=delivery` (or similar)
- `entrance=home` is basically `entrance=main` + `access=private` on `building=house|...`
- `entrance=staircase` is basically `entrance=main` + `access=private` + `wheelchair=*` on `building=apartments`
- `entrance=garage` is basically `entrance=secondary` + `motor_vehicle=yes` (on `amenty=parking` or `building=garage(s)`)   etc.

matkoniecz:
@mnalis Have you seen https://github.com/streetcomplete/StreetComplete/issues/2706#issuecomment-817202049 ?

mnalis:
> @mnalis Have you seen [#2706 (comment)](https://github.com/streetcomplete/StreetComplete/issues/2706#issuecomment-817202049) ?

Yes, and I've linked to https://github.com/streetcomplete/StreetComplete/issues/2706#issuecomment-817202174 which is the answer to that, with which I basically agree. However, probably only the users of that building (or very determined <strike>stalker</strike>mapper) can say **for sure** which of the entrances is `entrance=main` and which are `entrance=secondary` (also, one obscured by a tree might conceivably be `entrance=garage`, and some of them might be `entrance=exit` - but it's hard to say from that picture resolution) - which is why I have a problem with enabled-by-default-quest having such [very detailed / hard to answer list of answers](https://github.com/streetcomplete/StreetComplete/pull/3949#issuecomment-1093521029). 

<details>

Some of them might also be more precisely mapped as `entrance=home` maybe? (As it's not a building style I'm familiar with, I can't say with certainty if that is single/dual family residence, or a condo / apartment building, or a hotel, or maybe dormitory or terrace (which is also why I dislike problematic `building type quest` [blocking much more important quests](https://github.com/streetcomplete/StreetComplete/issues/2464)). 

</details>

Although (related to that) I do note that some people/[wiki](https://wiki.openstreetmap.org/wiki/Tag:entrance%3Dmain) equate `entrance=main` to mean `entrance=preferred_access_for_general_public` (i.e. `access=yes`) to be used exclusively on commercial/retail buildings, and never on `building=residential|house|...`; however that did not seem to me to be what always actually gets [mapped in practice](https://overpass-turbo.eu/s/1hwZ).

1ec5:
> An entrance=exit is a bit an entrance=service at least from the outside

`exit` is akin to a `oneway` indication for the doorway (but not the adjoining walkway, which can technically still handle two-way traffic). `service` vaguely implies some sort of access restriction (which would probably still need to be tagged).

For what it’s worth, typical outdoor signage for an exit would be an “Exit Only” or “Do Not Enter” sign. I don’t know if anyone has been mapping supermarket entrances to this level of detail, but this is typical of supermarkets in the U.S.:

[do not enter](https://www.walb.com/story/36669528/phoebe-purchases-vacant-former-grocery-store/)

westnordost:
>> Regarding that, I find [entrance=staircase wiki](https://wiki.openstreetmap.org/wiki/Tag:entrance=staircase) sorely lacking in content

> I expanded it a bit

Well this page now directly contradicts what can be read on `entrance=main`:

You write on `entrance=staircase`:
> Popular for staircases of apartment buildings, which can be also tagged as [entrance](https://wiki.openstreetmap.org/wiki/Key:entrance)=[main](https://wiki.openstreetmap.org/wiki/Tag:entrance%3Dmain).

On `entrance=main`, one can read:
> Not all buildings with a side [entrance](https://wiki.openstreetmap.org/wiki/Key:entrance) need to use entrance=[main](). This tag should only be used for the primary entrance to a publicly [access](https://wiki.openstreetmap.org/wiki/Key:access)ible building (like access=[yes](https://wiki.openstreetmap.org/wiki/Tag:access%3Dyes)).
> Use a different value for private buildings and residences, such as [yes](https://wiki.openstreetmap.org/wiki/Tag:entrance%3Dyes), [home](https://wiki.openstreetmap.org/wiki/Tag:entrance%3Dhome), [service](https://wiki.openstreetmap.org/wiki/Tag:entrance%3Dservice), [staircase](https://wiki.openstreetmap.org/wiki/Tag:entrance%3Dstaircase) etc.

It is the `entrance=main` wiki definition that I have been following for years - I only tag "main entrance" on larger publicly accessible buildings such as malls, schools, university buildings, hospitals etc. and that's what I generally see in the area I map. So I hear it is not in the original proposal from 10 years ago:

> Obviously, it is not the most important way to decide tagging, but note that https://wiki.openstreetmap.org/wiki/Proposed_features/entrance is without staircase value.

Exactly, the original proposal doesn't matter when the tag has seen such spread, the de-facto usage matters. You can't just retroactively deprecate the second most popular value like that.

---

@1ec5 For malls, supermarkets and the like the question is certainly easier to answer, but no chance in some nondescript building in a mixed landuse. I think 90% of all places where this quest will be asked are actually footways leading to cold-war style "apartment block on a lawn" type of buildings. For those, most relevant would be `staircase` and other (`yes`).

matkoniecz:
> Exactly, the original proposal doesn't matter when the tag has seen such spread, the de-facto usage matters. You can't just retroactively deprecate the second most popular value like that.

Oh I agree. I just thought that I am describing actual use. Thanks for pointing out that I am wrong, I will recheck situation and edit.

1ec5:
> I think 90% of all places where this quest will be asked are actually footways leading to cold-war style "apartment block on a lawn" type of buildings.

That would be an interesting statistic to get hard numbers on. Anecdotally, I’d imagine seeing plenty of cases involving [schools](https://www.openstreetmap.org/node/9460665112), [hotels](https://www.openstreetmap.org/node/6154792879), and even [single-family homes](https://www.openstreetmap.org/node/6217675114) where mappers didn’t originally bother to tag an `entrance`, but then again, the apartment blocks you’re describing are rare where I live. Maybe it’s possible to filter to the kind of building you want this quest answered for?

matkoniecz:
> Maybe it’s possible to filter to the kind of building you want this quest answered for?

Yes, in https://github.com/streetcomplete/StreetComplete/blob/21dc88c6b3ed7c947d12d186552ed4f5f07c7c6c/app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt#L19

It is entirely possible to limit it to some specific building types and skip say `building=house`

matkoniecz:
Staircase entrance added, secondary removed.

I still think that making possible to tag things like clear exit-only doors is a good idea.

matkoniecz:
exit only ones are signed reasonably often

for service entrance often situation is clear from the context

but I can remove this options if you really do not want them

see some clear unsigned service ones:

screen01
screen02


https://www.google.pl/maps/@50.0845536,19.9356153,3a,75y,165.05h,91.27t/data=!3m6!1e1!3m4!1si0iBaMGEozIQzWPY-sj6KA!2e0!7i13312!8i6656

https://www.google.pl/maps/@50.0843499,19.9353916,3a,15y,47.05h,87.67t/data=!3m6!1e1!3m4!1s2LcTZr9zIKcgVDoPhdjy1w!2e0!7i13312!8i6656

screen03
screen04


https://www.google.pl/maps/@50.0692271,19.9464206,3a,42.9y,184.14h,97.29t/data=!3m6!1e1!3m4!1sm3Z_I97J1uvq7kT_uvtjtg!2e0!7i13312!8i6656

https://www.google.pl/maps/@50.0659903,19.9461257,3a,90y,354.05h,105.96t/data=!3m6!1e1!3m4!1s9F2cpbIzb4MDoKKjzGKDMg!2e0!7i13312!8i6656

westnordost:
These are all on a public shopping center and alike. Are these the type of objects that are found with this quest?

app/src/main/res/values/strings.xml
```diff
@@ -1268,6 +1268,17 @@ If there are no signs along the whole street which apply for the highlighted sec
...
+    <string name="quest_building_entrace_title">What kind of building entrance is this?</string>
+    <string name="quest_building_entrance_main">Main (building may have multiple main entrances)</string>
+    <string name="quest_building_entrance_staircase">Staircase</string>
+    <string name="quest_building_entrance_service">Service</string>
+    <string name="quest_building_entrance_exit">Exit only</string>
+    <string name="quest_building_entrance_emergency_exit">Emergency exit only</string>
+    <string name="quest_building_entrance_garage">Garage</string>
+    <string name="quest_building_entrance_shop">Shop (storefront entrance), not a main building entrance</string>
+    <string name="quest_building_entrance_yes">Other entrance type</string>
+    <string name="quest_building_entrance_dead_end">Dead end, no entrance</string>
```
westnordost, line 1280: Regarding wording:

- doesn't mention that "main" should be used on public buildings etc. according to wiki
- hint that a building may have several main entrances seems superfluous because unlike in JOSM etc. the user doesn't see which other entrances are tagged as "main"
- I suggest "Staircase entrance (e.g. of apartment building)" because it is mostly/supposed to be used on apartment buildings
- I think for some values, "entrance" needs to be added because otherwise it is not well understood or doesn't make sense. E.g. "service" is not really translatable in German. Even German word for "service entrance" would be literally translated "servant entrance" which is itself a outdated term. I think the modern term would be "Nebeneingang" ("side entrance").
- "Garage": What does this mean, is this an entrance into a car garage? Or literally the garage door of a garage? (Why would anyone tag this? And why would a **footway** lead into a car garage rather than a `highway=service`?) 
- "other entrance type": Wording implies that the user read through all other options. Maybe instead just "unspecific entrance"

matkoniecz:
> These are all on a public shopping center and alike. Are these the type of objects that are found with this quest?

also with apartment buildings, for example ones with with built-in waste disposals

screen06


https://www.google.pl/maps/@50.0799472,19.8823393,3a,36.2y,175.34h,84.86t/data=!3m6!1e1!3m4!1sj_qAeBC27JFWe-gmB1w5cg!2e0!7i13312!8i6656

app/src/main/res/values/strings.xml
matkoniecz: > Wording implies that the user read through all other options

That is intentional. 

> Unspecific entrance

I got convinced by native speakers that "other" is better :) https://osmus.slack.com/archives/C01BU04GW7L/p1648903729432579

> Garage

removed

> I think for some values, "entrance" needs to be added because otherwise it is not well understood or doesn't make sense

done, for all (for consistency)

Other hints also followed

Current state:

```
    <string name="quest_building_entrace_title">What kind of building entrance is this?</string>
    <string name="quest_building_entrance_main">Main entrance on a public building</string>
    <string name="quest_building_entrance_staircase">Staircase entrance (e.g. of apartment building)</string>
    <string name="quest_building_entrance_service">Service entrance</string>
    <string name="quest_building_entrance_exit">Exit only</string>
    <string name="quest_building_entrance_emergency_exit">Emergency exit only</string>
    <string name="quest_building_entrance_shop">Shop entrance (storefront entrance), not a main building entrance</string>
    <string name="quest_building_entrance_yes">Other entrance type</string>
    <string name="quest_building_entrance_dead_end">Dead end, no entrance</string>
```

matkoniecz:
One thing that is still not done is cutting down answer list proposed in https://github.com/streetcomplete/StreetComplete/pull/3949#issuecomment-1092765898

I still think that even if most exit only/emergency exists/service entrances are not clearly distinguishable it is still nice to tags ones that are.

matkoniecz:
@westnordost Do you want to make an icon?

Do you think that service/exit/emergency/shop entrances are survivable or should it be removed?

screen02

app/src/main/res/values/strings.xml
```diff
@@ -697,6 +697,16 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
 
     <string name="quest_bridge_structure_title">What’s the structure of this bridge?</string>
 
+    <string name="quest_building_entrace_title">What kind of building entrance is this?</string>
+    <string name="quest_building_entrance_main">Main entrance of a public building</string>
+    <string name="quest_building_entrance_staircase">Staircase entrance (e.g. of apartment building)</string>
+    <string name="quest_building_entrance_service">Service entrance</string>
+    <string name="quest_building_entrance_exit">Exit only</string>
+    <string name="quest_building_entrance_emergency_exit">Emergency exit only</string>
+    <string name="quest_building_entrance_shop">Shop entrance (storefront entrance), not a main building entrance</string>
```
westnordost, line 706:
```suggestion
    <string name="quest_building_entrance_shop">Storefront entrance</string>
```

Too long otherwise IMO

app/src/main/res/values/strings.xml
```diff
@@ -697,6 +697,16 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
 
     <string name="quest_bridge_structure_title">What’s the structure of this bridge?</string>
 
+    <string name="quest_building_entrace_title">What kind of building entrance is this?</string>
```
westnordost, line 700: typo in title name ("entrace")

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt
```diff
@@ -0,0 +1,72 @@
...
+          highway ~ path|footway and area != yes and access !~ private|no
+    """.toElementFilterExpression() }
+
+    private val excludedWaysFilter by lazy { """
+        ways with (tunnel and tunnel != no) or covered=yes
+    """.toElementFilterExpression() }
+
+    override val changesetComment = "Add entrance info"
+    override val wikiLink = "Key:entrance"
+    override val icon = R.drawable.ic_quest_railway // TODO !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
westnordost, line 33:
```suggestion
    override val icon = R.drawable.ic_quest_door
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt
```diff
@@ -0,0 +1,72 @@
...
+    """.toElementFilterExpression() }
+
+    private val incomingWaysFilter by lazy { """
+        ways with
+          highway ~ path|footway and area != yes and access !~ private|no
+    """.toElementFilterExpression() }
+
+    private val excludedWaysFilter by lazy { """
+        ways with (tunnel and tunnel != no) or covered=yes
+    """.toElementFilterExpression() }
```
westnordost, line 29: You can combine both filters

```suggestion
    private val incomingWaysFilter by lazy { """
        ways with
          highway ~ path|footway
          and area != yes
          and access !~ private|no
          and (!tunnel or tunnel = no)
          and covered != yes
    """.toElementFilterExpression() }
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt
```diff
@@ -0,0 +1,72 @@
...
+        val excludedWayNodeIds = mutableSetOf<Long>()
+        mapData.ways
+            .filter { excludedWaysFilter.matches(it) }
+            .flatMapTo(excludedWayNodeIds) { it.nodeIds }
+
+        return mapData.nodes
+            .filter { withoutEntranceFilter.matches(it)
+                && it.id in buildingsWayNodeIds && it.id in incomingWayNodeIds
+                && it.id !in excludedWayNodeIds
+            }
```
westnordost, line 58: It's more performant to first check if the id is in the set and do the `withoutEntranceFilter.matches(it)` last in that expression

westnordost (commented):
Ok, the remaining answer options are fine.

app/src/main/res/values/strings.xml
matkoniecz: `Storefront entrance (not main entrance)` ?

westnordost: If it was a main entrance, the user would have chosen that already because it is displayed further up, no? Also, if there is a single shop in a building and that building has one entrance, "storefront entrance (not main entrance)" would be confusing

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt
matkoniecz: Actually no, it is not the same.

Note that node may touch OK way and banned way.

I added test for that.

matkoniecz: Is https://github.com/streetcomplete/StreetComplete/blob/ad891e038076a82651a65b96aa103b14fa75e157/app/src/test/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntranceTest.kt explaining difference?

Node may be both in a banned way and acceptable way.

In particular, https://www.openstreetmap.org/node/3091300654 should not trigger entrance quest

westnordost: right, I understand

westnordost (approved):

westnordost:
Ready to merge? It's still marked as draft.

matkoniecz:
yes, I just did final testing in emulator

HolgerJeromin:
Do we want to exclude carports?

matkoniecz:
roof is excluded already, I will also exclude carport - https://github.com/streetcomplete/StreetComplete/commit/b857f974d6fd1dde4c462f57fc8ac00e99d27bec

PR #3951, opened 2022-04-09
change AddSeating() quest position

mnalis:
show `Seating quest` earlier in list of quest, because:

- it is quicker and easier to determine than for example `AirConditioning` or `Smoking` quests (when the `Seating` quest is enabled seasonally)
- `Smoking` quest would benefit, as it will modify its behaviour if `Seating` data is available


matkoniecz:
Also, it is easier/similar to see a s ATM operator, easier to solve and is disabled by default so putting it higher makes it easier to find.

PR #3953, opened 2022-04-09
add AddAcceptsCards quest

### LINKED ISSUE #2704: What forms of payment are accepted here? (focusing on small businesses in the United States)
### General
Affected tag(s) to be modified/added: [`payment:cash`](https://wiki.openstreetmap.org/wiki/Key:payment:cash), [`payment:cheque`](https://wiki.openstreetmap.org/wiki/Key:payment:cheque), [`payment:credit_cards`](https://wiki.openstreetmap.org/wiki/Key:payment:credit_cards), [`payment:debit_cards`](https://wiki.openstreetmap.org/wiki/Key:payment:debit_cards)
Question asked: **What forms of payment are accepted here?**

The following answers are for the United States; I don’t think any answers could be generalized globally.

### Checklist
    * You don’t want to accidentally order at a sit-down restaurant, finish your meal, and only then realize you can’t pay.
    * You don’t want to accidentally get your hair cut and then realize you can’t pay.
    * Virtually every restaurant or shop in the U.S. posts either the logos of credit/debit cards they accept or a “cash only” sign.
    * Small businesses, especially in ethnic enclaves, are often cash-only establishments or only accept one credit card but not another.

[cash only](https://commons.wikimedia.org/wiki/File:*_CASH_ONLY_*_No_Debit_Cards;_No_Credit_Cards_%2833172689610%29.jpg)
[restaurant](https://commons.wikimedia.org/wiki/File:MetairieRoadMarkTwainTwain.JPG)
[hairdresser](https://commons.wikimedia.org/wiki/File:Fantastic_Sams_Ypsilanti.JPG)

### Ideas for implementation

Probably similar to the `accepts_cash` quest.

**Element selection:** 

`amenity=restaurant`, `amenity=fast_food`, `amenity=cafe`, `amenity=ice_cream`, `shop=*` without `payment:*=*`

Not `amenity=fuel`, because gas stations accept a wider variety of payment options that might be worth implementing in a separate quest.

Not `tourism=*`, `leisure=*`, or `office=*`, because these facilities generally require booking an appointment, seat, or room in advance, which means they’re less likely to display accepted payment methods at the door.

Some `shop=*` values like `doityourself`, `department_store`, and `supermarket` tend to be big-box stores, so the fact that they all accept every major credit card isn’t very interesting. I think this quest should try to focus on POI types that are more likely to be small businesses.

**Metadata needed:** 

I’m most familiar with the situation in the United States, so maybe limit the quest to the U.S. for now to avoid the issues that seemed to derail #1573?

**Proposed GUI:**

The `payment:*` tagging scheme is very complex, but here are the options that would be useful in 99.99% of the cases in the United States. It starts with a menu of options, each of which can be unset or set to Yes or No. Selecting “Credit card” or “Contactless” reveals a secondary menu of logos.

* Cash (`payment:cash`)
* Check (`payment:cheque`)
* Credit card (`payment:credit_cards`)
   * American Express (`payment:american_express`)
   * Discover (`payment:discover_card`)
   * Mastercard (`payment:mastercard`)
   * Visa (`payment:visa`)
* Debit card (`payment:debit_cards`)
* Contactless (`payment:contactless`)
   * Apple Pay (`payment:apple_pay`)
   * Google Pay (`payment:google_pay`)

The `payment:*` scheme isn’t exhaustive, allowing for a partial answer. For example, if a storefront window only displays credit card logos but doesn’t explicitly indicate whether cash is accepted, the user can simply leave the “Cash” option unset. Or they can set it to Yes knowing that cashless businesses are unheard of in that area. (There are some cashless stores these days, but they’re rare and concentrated in certain cities.)

matkoniecz:
fixes #2704

> Can you pay here with card?

> Is card payment accepted here?

which would be better?

We have already

> Is cash payment accepted here?

(active in Sweden for now) but it seems oddly formal to me, but of it is not worse it would be nice for consistency.

-----------

@westnordost Are you interested in making an icon? Or should I try making something? (probably cash quest symbol, but on rectangle, within circle of quest icon)

--------------

BTW, I am considering extending range of asked places to what similar quests asks: https://github.com/streetcomplete/StreetComplete/blob/18189f123416d9a0b332e50afab58b857452b416/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cash/AddAcceptsCash.kt#L17-L49 (which would likely imply extracting this listing to some shared place)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
@@ -14,10 +14,7 @@ class AddAcceptsCards : OsmFilterQuestType<CardAcceptance>() {
 
     override val elementFilter = """
         nodes with (
-          amenity=restaurant
-          or amenity=fast_food
-          or amenity=cafe
-          or amenity=ice_cream
+          amenity ~ restaurant|fast_food|cafe|ice_cream
           or (shop and shop !~ no|vacant|mall)
         )
         and !payment:credit_cards and !payment:debit_cards
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
@@ -27,7 +27,7 @@ class AddAcceptsCards : OsmFilterQuestType<CardAcceptance>() {
     override val changesetComment = "Add whether cards are accepted"
     override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
     override val wikiLink = "Key:payment"
-    override val icon = R.drawable.ic_quest_drinking_water // TODO FIX
+    override val icon = R.drawable.ic_quest_card
     override val isDeleteElementEnabled = true
     override val questTypeAchievements = listOf(CITIZEN)
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
@@ -1,0 +1,44 @@
+package de.westnordost.streetcomplete.quests.accepts_cards
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.osmquests.Tags
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
+
+class AddAcceptsCards : OsmFilterQuestType<CardAcceptance>() {
+
+    override val elementFilter = """
+        nodes with (
+          amenity=restaurant
+          or amenity=fast_food
+          or amenity=cafe
+          or amenity=ice_cream
+          or (shop and shop !~ no|vacant|mall)
+        )
+        and !payment:credit_cards and !payment:debit_cards
+        and !brand and !wikipedia:brand and !wikidata:brand
+        and (!seasonal or seasonal = no)
+    """
+    override val changesetComment = "Add whether cards are accepted"
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+    override val wikiLink = "Key:payment"
+    override val icon = R.drawable.ic_quest_drinking_water // TODO FIX
+    override val isDeleteElementEnabled = true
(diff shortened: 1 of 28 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -558,6 +558,12 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <!-- Quests -->
     <!-- sorted alphabetically by title -->
 
+    <string name="quest_accepts_cards">Can you pay here with card?</string>
```
1ec5, line 561: “With card” sounds awkward to me. The only word that refers to both is “plastic”, but that’s slang. These are slightly longer but read more naturally:

> Can you pay by credit or debit card here?
> Are credit or debit cards accepted here?

app/src/main/res/values/strings.xml
```diff
@@ -558,6 +558,12 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
...
+    <string name="quest_accepts_cards_debit_and_credit">Credit and debit cards allowed</string>
```
1ec5, line 562: At least in American English, you’d say credit and debit cards are “accepted”, not “allowed”.

app/src/main/res/values/strings.xml
matkoniecz: changed to Are credit or debit cards accepted here?

app/src/main/res/values/strings.xml
matkoniecz: thanks, English improved (I really, really appreciated feedback about language)

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
```diff
@@ -0,0 +1,45 @@
...
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddAcceptsCards : OsmFilterQuestType<CardAcceptance>() {
+
+    override val elementFilter = """
+        nodes with (
+          amenity=restaurant
+          or amenity=fast_food
+          or amenity=cafe
+          or amenity=ice_cream
```
westnordost, line 20:
```suggestion
          amenity ~ restaurant|fast_food|cafe|ice_cream
```
Can be shorter like this.

Also, @matkoniecz no space between key and value is allowed, but for consistency, please always add spaces as for other quests, it is formatted that way also.

app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
```diff
@@ -0,0 +1,42 @@
...
+        )
+        and !payment:credit_cards and !payment:debit_cards
+        and !brand and !wikipedia:brand and !wikidata:brand
+        and (!seasonal or seasonal = no)
+    """
+    override val changesetComment = "Add whether cards are accepted"
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+    override val wikiLink = "Key:payment"
+    override val icon = R.drawable.ic_quest_card
+    override val isDeleteElementEnabled = true
```
westnordost, line 28:
```suggestion
    override val isReplaceShopEnabled = true
```

Hmm, this is wrong. This refers to shops, so the shop should not be deleted but instead potentially replaced/marked as vacant. Agree, @matkoniecz ?

Cj-Malone:
Is there a reason why this was limited to nodes only?

matkoniecz:
> Is there a reason why this was limited to nodes only?

No, and I will fix it

Issue #3377, opened 2021-10-13
[v36.0] Can we get back the option for Compass Mode?


pkoby:
I tried the new version today with follow mode, and I can find no fault with the new functionality. I like the track of my path, and I think that pointing in the direction of movement is useful.

However, I don't think it's ideal for walking. For cycling, your focus is mostly on what's ahead, and you're not spending a lot of time looking around. When walking, though, I am constantly glancing around to see what's on either side, or weaving around a university campus aimlessly.

For walking, Compass Mode was excellent because I could face towards a quest and know exactly where to look. Now, I might see a quest to the side, but I don't know for sure that I'm facing the same way as the map (the direction takes awhile to catch up to a quick turn on foot).

Can I petition that we return to at least the option of using the compass? Perhaps it could be a triple toggle: tap once to find location, tap again to follow movement, tap a third time to follow compass. Then if you manually pan away, it turns black, tap again to return to red mode, then tap again to cycle. Maybe follow location only would also orient northwards?

westnordost:
@pkoby , I agree that the compass is useful in these situations. In fact, the compass is still used. It is now as before displayed as a err... triangle:

https://raw.githubusercontent.com/streetcomplete/StreetComplete/master/app/src/main/res/drawable-xhdpi/location_direction.png

mnalis:
@westnordost yes, triangle is shown (and is very useful!) but I can see how sometimes people (myself included) might prefer a whole screen (in `3D` mode) to be rotated in `compass` direction (as it was in previous SC versions below `36`). For example, on the left picture (simulated behavior in previous versions) it might be more to know what is exactly in front of you, and what on your sides, as opposed to the right picture (current behavior).

![Screenshot_20211013_233954_de westnordost streetcomplete debug2](https://user-images.githubusercontent.com/156656/137216881-bea01b4e-71a9-45ef-9c77-d339a11495d6.jpg)   ![Screenshot_20211013_233710_de westnordost streetcomplete debug2](https://user-images.githubusercontent.com/156656/137216882-d5756a76-26aa-47d8-8b2f-ca7685cca31c.jpg)

pkoby:
Thanks for all the quick input!

> consider both your movement direction and compass direction — probably a weighted average, depending on the speed of your movement (moving fast = ignore compass; moving slow = mostly compass). Would something like that work for you?

[...]

PR #3954, opened 2022-04-10
Orientation mode preference

mnalis:
Closes https://github.com/streetcomplete/StreetComplete/issues/3377

As accepted in https://github.com/streetcomplete/StreetComplete/issues/3377#issuecomment-944424594 this PR adds a new 
multiselect preference: `Navigation mode: orientation`, where the user can select between:
- `by movement direction` (default)
- `by compass direction` (old pre-v36.0 behavior)

Selecting preference part of the PR works. I seem to have [managed](https://github.com/mnalis/StreetComplete/actions/runs/2142784150) to get different behaviour depending on the setting chosen, although it does not do what I want it correctly yet. 

---
**TODO work in progress:**

My naive approach was to add a preference, and set different `bearing` depending on that preference - either current one depending on `movement direction` (i.e. based on GPX track) - which is easy one as it is how it currently work so no code changes needed,  or the old one depending solely on `compass direction` (which is being problematic).

- While it seems to work, it is **_incredibly chunky_**, doing map repositioning only every second or so. While before https://github.com/streetcomplete/StreetComplete/pull/3335 it **was very fluid**. Unfortunately the commit https://github.com/streetcomplete/StreetComplete/pull/3335/commits/99edceb43dc6af61aca52c1cc2ae0462b8959383 that seems to be responsible for the main changes seems to implement many new things and ways of doing them, and me not knowing Kotlin does not help in untangling that...

  I've tried playing with constants that seemed obvious (reducing numbers in `updateCameraPosition(600)`, `requestUpdates(2000, 1f)`) but while it seems to help a little, it is still incredibly chunky, so the current way is probably not a solution.

  Now I have been playing with reverting parts of code to do updates directly in `onCompassRotationChanged()` as the code was doing before, and that seems to help with fluidity, but then I get into sync issues when `centerCurrentPosition()` attempts to update map with something different and all hell breaks loose. So if someone wishes so throw some hints to me about how it should be done, it would be greatly appreciated; if not, I'll resume with try-and-fail method :smiling_face_with_tear: 

- Also, it is probably quite obvious from the code that I'm no Kotlin programmer. So I apologize if some things in code are horrible solutions, I basically tried cutting and pasting pieces of code that seemed to do similar things to what I wanted.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/Prefs.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/Prefs.kt
@@ -15,6 +15,7 @@ object Prefs {
     const val THEME_SELECT = "theme.select"
     const val LANGUAGE_SELECT = "language.select"
     const val THEME_BACKGROUND = "theme.background_type"
+    const val ORIENTATION_SELECT = "orientation.select"
 
     const val RESURVEY_INTERVALS = "quests.resurveyIntervals"
 
@@ -55,5 +56,9 @@ object Prefs {
 
     enum class ResurveyIntervals {
         LESS_OFTEN, DEFAULT, MORE_OFTEN
+    }
+
+    enum class NavigationOrientation {
+        MOVEMENT_DIRECTION, COMPASS_DIRECTION
     }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/StreetCompleteApplication.kt
@@ -37,6 +37,7 @@ import de.westnordost.streetcomplete.screens.main.mainModule
 import de.westnordost.streetcomplete.screens.main.map.mapModule
 import de.westnordost.streetcomplete.screens.measure.arModule
 import de.westnordost.streetcomplete.screens.settings.ResurveyIntervalsUpdater
+import de.westnordost.streetcomplete.screens.settings.NavigationOrientationUpdater
 import de.westnordost.streetcomplete.screens.settings.settingsModule
 import de.westnordost.streetcomplete.util.CrashReportExceptionHandler
 import de.westnordost.streetcomplete.util.getSelectedLocale
@@ -62,6 +63,7 @@ class StreetCompleteApplication : Application() {
     private val preloader: Preloader by inject()
     private val crashReportExceptionHandler: CrashReportExceptionHandler by inject()
     private val resurveyIntervalsUpdater: ResurveyIntervalsUpdater by inject()
+    private val navigationOrientationUpdater: NavigationOrientationUpdater by inject()
     private val downloadedTilesDao: DownloadedTilesDao by inject()
     private val prefs: SharedPreferences by inject()
     private val editHistoryController: EditHistoryController by inject()
@@ -125,6 +127,7 @@ class StreetCompleteApplication : Application() {
         setDefaultTheme()
 
         resurveyIntervalsUpdater.update()
+        navigationOrientationUpdater.update()
 
         val lastVersion = prefs.getString(Prefs.LAST_VERSION_DATA, null)
         if (BuildConfig.VERSION_NAME != lastVersion) {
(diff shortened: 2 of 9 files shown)


westnordost:
preliminary review:
- remove NavigationOrientationUpdater
- instead, let the LocationAwareMapFragment access the Prefs.ORIENTATION_SELECT in SharedPreferences directly
- if updating the orientation should happen at a higher interval and/or with an additional smoothing applied to the sensor data, please refer to the commit that removed/replaced the previous behavior because this is the code you (and I, as reviewer) want to have back in place.

westnordost:
Also note you have still work on #3934 pending for it to be merged.

mnalis:
> preliminary review:

thanks for taking a look!

> * remove NavigationOrientationUpdater
> * instead, let the LocationAwareMapFragment access the Prefs.ORIENTATION_SELECT in SharedPreferences directly

Hm, OK I'll try that way if you want, but I've done it this way (keeping it directly in `Boolean` variable) because of performance (and battery usage) fears - it is going to be be **_checked many times per second_** (for fluid motion, I estimate about 10-30 [FPS](https://en.wikipedia.org/wiki/Frames_per_second)?), and I have no idea how expensive those `SharedPreferences` accesses are (may be going via SQL to disk each time to query the DB, as far as I know)

westnordost:
`SharedPreferences` is held in memory as a `Map` as far as I know

westnordost:
But if you are concerned, it could be a local variable that is updated only once on `onStart`

YoshiRulz:
Bump and also offering to help. I was looking at the Issues for Kotlin/JVM => Kotlin MPP work for a first contrib, as that's what I have experience with, but would be happy to take over here if @mnalis is too busy to continue.

mnalis:
@YoshiRulz absolutely, please do jump in! :heart:  
Someone with kotlin experience would likely find it much easier to fix then myself!

westnordost:
I've not looked into this PR for a long time, let me know if you need any help. My last "preliminary review" should still be valid.

By the way, Kotlin/JVM => Kotlin MPP is highly appreciated. There is a lot to do, after an initial project setup, but/so I suggest to do this in as small PRs as possible. I.e. one PR for adding the mpp project setup (without actually moving any files over to the `common` module), and then, getting the stuff in `android` one by one ready to be moved to `common` by removing/replacing the dependencies on Java. Each in their own PR because if it gets too large of a "construction site", it will negatively affect other PRs, with merging and all, plus, will turn into one of those PRs with work on it happening sometimes and thus stretching on and on, sinking much time into review, rereview, discussion etc..

mnalis:
@YoshiRulz have you seen the OK https://github.com/streetcomplete/StreetComplete/pull/3954#issuecomment-1253028231? Are you still interested to work on it? (or maybe have you started working? please link to it here if you did)

YoshiRulz:
I've installed Android Studio and, after a quick rebase, it just worked. The debug app opens on my phone and I can see that there's a new setting which doesn't work, as you said. I guess I'll start on https://github.com/streetcomplete/StreetComplete/pull/3954#issuecomment-1094249373 now.

YoshiRulz:
Correction: it seems it *does* work, but the update function was written with velocity in mind and so only updates on movement. I've pushed my work [here](https://github.com/YoshiRulz/StreetComplete/tree/orientation-mode-pref).

westnordost:
You can just create a new PR when you are done or a new draft PR if you have questions / need help to replace this PR.

mnalis:
> Correction: it seems it _does_ work, but the update function was written with velocity in mind and so only updates on movement. I've pushed my work [here](https://github.com/YoshiRulz/StreetComplete/tree/orientation-mode-pref).

@YoshiRulz you may want to take a look at https://github.com/streetcomplete/StreetComplete/commit/99edceb43dc6af61aca52c1cc2ae0462b8959383 to see how it worked before (and to get idea how to restore that behaviour if `COMPASS_DIRECTION` preference is selected)

mnalis:
@YoshiRulz any luck with https://github.com/YoshiRulz/StreetComplete/tree/orientation-mode-pref ?
Are the any roadblocks there?

YoshiRulz:
I haven't yet looked at the old compass mode.

westnordost:
This has been stale for almost a year now. If this is implemented, it should rather be done in a new PR.

PR #3956, opened 2022-04-10
Upgrading assertj-core to 3.22.0

mccartney:
Upgrading assertj-core to 3.22.0 for not apparent reason other than staying up to date.
Currently used version of `assertj-core` is from 2017.

`./gradlew test` passes.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -130,12 +130,12 @@ dependencies {
     testImplementation("junit:junit:4.13.2")
     testImplementation("org.mockito:mockito-core:$mockitoVersion")
     testImplementation("org.mockito:mockito-inline:$mockitoVersion")
-    testImplementation("org.assertj:assertj-core:2.8.0")
+    testImplementation("org.assertj:assertj-core:3.22.0")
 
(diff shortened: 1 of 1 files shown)


matkoniecz:
2.8.0 -> 3.22.0

Have they described any breaking changes in release notes?

FloEdelmann (approved):
Since they are a bit scattered around different websites, here are the release notes for versions …

* `<=3.4.1`: https://joel-costigliola.github.io/assertj/assertj-core-older-releases.html
* `3.5.0 - 3.11.1`: https://joel-costigliola.github.io/assertj/assertj-core-news.html
* `>=3.12.0`: https://assertj.github.io/doc/#assertj-core-release-notes

The [v3.0.0 changelog](https://joel-costigliola.github.io/assertj/assertj-core-older-releases.html#assertj-core-3.0.0) only lists one breaking change:

> AssertJ Core 3.0 relies on Java 8.

I think that's fine then. Also, the tests are passing on my local machine (both unit tests and instrumented tests running on an emulator).

mccartney:
Thanks for answering before I could get to it. Also worth noting - that `assertj` seems to be used in like 2 tests in total.

PR #3964, opened 2022-04-14
Add more negative tests for prefix usage with TagOlderThan

peternewman:
To document and validate the current behaviour.

While doing some investigation of #3946

PR #3965, opened 2022-04-14
Fix the TagOlderThan tests and add some further tests

peternewman:
After #3964 !

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/filters/TagOlderThanTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/filters/TagOlderThanTest.kt
@@ -69,11 +69,23 @@ class TagOlderThanTest {
         ), newDate))
     }
 
-    @Test fun `does not match old element with tag only as a prefix`() {
-        assertFalse(c.matches(mapOf("opening_hours:signed" to "tag"), oldDate))
-    }
-
     @Test fun `does not match new element with tag only as a prefix`() {
         assertFalse(c.matches(mapOf("opening_hours:signed" to "tag"), newDate))
+    }
+
+    @Test fun `matches any element which is old`() {
+        assertTrue(c.matches(mapOf("foo" to "bar"), oldDate))
+    }
+
+    @Test fun `does not match any element which is new`() {
+        assertFalse(c.matches(mapOf("foo" to "bar"), newDate))
+    }
+    
+    @Test fun `matches element with no tags which is old`() {
+        assertTrue(c.matches(mapOf(), oldDate))
+    }
+
+    @Test fun `does not match element with no tags which is new`() {
+        assertFalse(c.matches(mapOf(), newDate))
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/filters/TagOlderThanTest.kt
```diff
@@ -69,11 +69,23 @@ class TagOlderThanTest {
...
+        assertFalse(c.matches(mapOf("foo" to "bar"), newDate))
+    }
+
+    @Test fun `matches element with no tags which is old`() {
+        assertTrue(c.matches(mapOf(), oldDate))
+    }
+
+    @Test fun `does not match element with no tags which is new`() {
+        assertFalse(c.matches(mapOf(), newDate))
+    }
```
FloEdelmann, line 90: Better test names and grouping:

```suggestion
    @Test fun `matches any old element without relevant tags`() {
        assertTrue(c.matches(mapOf("foo" to "bar"), oldDate))
        assertTrue(c.matches(mapOf(), oldDate))
    }

    @Test fun `does not match any new element without relevant tags`() {
        assertFalse(c.matches(mapOf("foo" to "bar"), newDate))
        assertFalse(c.matches(mapOf(), newDate))
    }
```

peternewman: Makes sense. I think in theory the empty maps should never exist...

app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/filters/TagOlderThanTest.kt
```diff
@@ -69,11 +69,17 @@ class TagOlderThanTest {
...
 
-    @Test fun `does not match old element with tag only as a prefix`() {
-        assertFalse(c.matches(mapOf("opening_hours:signed" to "tag"), oldDate))
-    }
-
     @Test fun `does not match new element with tag only as a prefix`() {
         assertFalse(c.matches(mapOf("opening_hours:signed" to "tag"), newDate))
     }
+
+    @Test fun `matches any old element without relevant tags`() {
```
peternewman, line 76: @westnordost , just to check that this is intended behaviour? It caught me by surprise a bit (especially given the function name), although doesn't impact on the current usages as they are also using a separate element filter.

westnordost: Hmm, I guess this behavior would be unexpected.

peternewman: I think if its fixed #3946 will need some alternative logic for when recycling/cycleway don't have a check date already.

I guess doing `TagOlderThan || ElementOlderThan` would make sense, or `AnyTagOlderThan("recycling:*")`

westnordost: Hmm... right. `check_date:cycleway` could be set but that doesn't mean that `cycleway` is set

westnordost: Besides, the parser does this:

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/ElementFiltersParser.kt#L243

so when written in the element filter expression, it **does** check for the tag being set. Hence, no documentation update in 

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/ElementFilterExpression.kt#L25-L61

is necessary.

PR #3970, opened 2022-04-18
Create .git-blame-ignore-revs

peternewman:
To automagically ignore some irrelevant revisions

From https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

See for example:
https://github.com/peternewman/StreetComplete/blame/patch-2/app/build.gradle.kts#L1-L3

Versus:
https://github.com/peternewman/StreetComplete/blame/master/app/build.gradle.kts#L1-L3

Note how the last line changes

We don't seem to be able to feed this into git log, but we could use it as the source of truth, or a base source (if we don't want to skip all these revisions in git blame) for updateContributorStatistics.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,3 @@
+# .git-blame-ignore-revs
+# mechanical linting
+9c6d3e25216d06a2c5afa71086949e1e195de926
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,23 @@
 # .git-blame-ignore-revs
+# some weird force-merge or something
+ae7a244dd60ccfc91cf2dc01bf9e60c8d6a81616
+# just a large re-import of orchard-produce images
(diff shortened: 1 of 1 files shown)


FloEdelmann:
I'm not sure if it makes sense. Linting (especially manual linting) could also introduce behavior changes, and not being able to find the relevant commit via `git blame` (at least in the GitHub UI) seems weird to me, given there isn't a way to quickly toggle this behavior.

Also, with this PR we would now have to duplicate lint commit hashes to two files instead of one. And even more, obvious lint commits that match the [exclude regex](https://github.com/streetcomplete/StreetComplete/blob/e79cca072b3958b8647cba04dc318dca70995ce9/build.gradle.kts#L79-L80) don't get written to this `.git-blame-ignore-revs` file automatically, so this means more work and more accidentally missed exclusions.

westnordost:
Your objection makes sense. I'll close this PR.

PR #3977, opened 2022-04-20
 Better indicate when there is no GPS signal

### LINKED ISSUE #3944: Better indicate when there is no GPS signal
**Use case**
When GPS signal is lost,  it can be hard for user to know, and thus wrong mapping can occur. It happened to me several times (that I noted!), especially when map layout in new location without GPS signal was similar to map layout where GPS was last available. (such GPS signal loss happens often especially if I turn off screen to save battery as there is nothing to map for some distance, or enter/exit buildings etc.) 

<Details>

<summary>Main visibility problems:</summary>

- if the map is in "3D follow mode" (i.e. GPS icon shows `arrowhead` instead of `dot`), it is not possible to say **at all** if the GPS signal is available or not (in "2D mode", i.e. when GPS icon is shown as a `dot`, the `dot` is blinking where there is no GPS signal, so it is at least somewhat visible, although it can be surprisingly easy to overlook)
- the GPS position on the map remain same shape and color, no matter if GPS is available or not
- direction the phone is pointed at (i.e. the blue section of the circle in front of GPS position dot on the map) is showing direction even if the location is totally wrong, which is highly misleading. 

</details>

See the video, especially about 00:00:09 - 00:00:14 for example of how misleading it can be.

https://user-images.githubusercontent.com/156656/161667162-e8c3edbf-226b-4c55-b193-335047a17c76.mp4

**Proposed Solution**

I propose to implement (ideally all of) following suggestions, in order to make it clear that the GPS position on the map is invalid and thus avoid user unknowingly mapping invalid data:

- if the GPS signal is lost, make the `arrowhead` GPS icon **blink also** (same as when `dot` GPS icon is blinking)
- if the GPS signal is lost, turn **blue** `dot` on the map indicating user position to **grey** `dot` (or `hexagon` or `smaller circle` or `hollow circle` or some other shape understandable to colorblind persons too).
- if the GPS signal is lost, **disable showing** direction the phone (i.e. the blue **section of the circle** in front of GPS position dot) 

(all of the above should return to normal when GPS signal is reacquired, of course)

westnordost:
fixes #3944

@mnalis Would you be available to test this? I have no idea how you were able to simulate a situation where the GPS signal was cut off.


westnordost:
I don't know what this gif is supposed to show, I don't see the GPS icon blinking. So anyway, @mnalis or @matkoniecz are you available to test this?

mnalis:
I'll be able to test it Monday or Tuesday, as I need access to location with big basement / lots of walls (or a dense forest) in order to reliably reproduce loss of GPS lock.

mnalis:
I've managed to [compile it](https://github.com/mnalis/StreetComplete/actions/runs/2197902514) and make a quick test; it seems that:
- blinking navigation arrow icon is working
- hiding the blue dot on the map (and blue transparent circle sector showing direction user is looking at) is **not** working (if I understand it correctly what PR was supposed to do)

Here is the example video just after signal loss:

https://user-images.githubusercontent.com/156656/165180866-2be10bad-d387-46b2-a634-63f0b0146981.mp4

(Steps: I had SC in foreground with working GPS signal just outside the building, then I turned screen off and entered the building thus losing GPS fix; after turning screen back on the SC which was still in foreground looked like video above).

westnordost:
Thank you for testing it! I will try to find the bug. Maybe I will also build a tinfoil box in which I could test it (if that is enough to block the signal)

westnordost:
Hm, it looks like there is not any callback that is called when location is lost. The only reason why the GPS button blinks in that case is that on initialization (when the app comes back into view), that button's state is first set to "searching" and only when the first location is reported back from the Android system, the state is set to "has location".

The same cane be done for the location dot of course. But note that this fix wouldn't help in the case when you lose location while the app is in foreground.

westnordost:
And it leads to some ugly blinking, i.e. the blue dot vanishes when the app comes back into view for ~1 second and pops back in after that. Every time you tab out and in. Given it would not really be a wholesome fix anyway, I rather not do that.

So, I'll merge it as is.

All works as expected when one disables location.

Issue #3978, opened 2022-04-20
Don't ask for opening hours for leisure=horse_riding


Hufkratzer:
According to [taginfo](https://taginfo.openstreetmap.org/tags/leisure=horse_riding#projects) and your [source code](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt) you ask for opening hours for leisure=horse_riding, but you don't ask it for leisure=sports_centre
> because these are often sports clubs which have no walk-in opening hours but training times

I think this is even more true for leisure=horse_riding. 

Only 218 of 15k (1,5%) leisure=horse_riding have some opening_hours tagged, which is a lower percentage than for leisure=sports_centre (7020 / 215k = 3,3%). This is probably not incomplete tagging but rather reflects reality. There may be an office somewhere on larger facilities which may have fixed opening hours, but this does not mean that the whole facility is closed when the office is closed. There are often privately owned horses on such facilities and their owners may need access to them independently from opening hours.

Hufkratzer:
The bug label was added automatically. I didn't want to say that it is a bug; I just didn't find a more appropriate way to submit this issue.

westnordost:
sports_centre was removed from this list in #1423

westnordost:
yes I know, I was just providing some context

Issue #2051, opened 2020-08-26
Give the users the possibility to mark private ground as such


cyclingcat:
Hi there,

when I deliberately walk/ride to a road surface or handrail quest I have to hide it surprisingly often, because I'm not able to answer it without entering private ground. (Sometimes it's possible to solve the quest with a peek over the fence, but often it's not.) This is probably due to the fact that I live in a well-mapped region where road surface quests are getting more and more rare, so the percentage of missing tags on private ground is quite high here, because noone has been able to answer it so far. Leaving a note in those cases is quite boring, but wouldn't it be possible to catch this case with StreetComplete in a reasonable way?

**Suggestion:**

Add a third menu item "Private ground" (or similar) to the "More answers" menu. If a user taps on this item, add an "access=private" tag, assuming that StreetComplete will never show quests about private objects which automatically hides this quest to other users in the future.

The cycling cat

westnordost:
In cases like these, you can and you are supposed to leave a note instead so that it can be tagged as private. I'd rather not add this to the other answers because I think a note is more valuable in this case: A way/stairs/road etc. that are private but not tagged as private are rarely alone. If you found out that a certain area is not public, a note will help to mark the whole area as private (all ways etc in the area).
Also, I'd like to pollute the other answers menu with as few items not directly related to the question itself as possible.

Issue #1423, opened 2019-06-10
Never ask for opening times for gyms/sports halls


cyclingcat:
During my StreetComplete actions in some villages nearby I often encounter sports halls for which opening hours quests are shown. However, none of them showed any opening hours visible from the outside, so I regularly have to hide the quest, which is quite annoying. I suppose that sports halls rarely have fixed opening times; they are open when some training or events take place and closed otherwise.

Proposal: Never show the opening hours quests when the object is tagged with "building=sports_hall".

This is a little bit related to #868 which led to excluding private objects from the opening hours quest - but here we have public (often municipal) buildings.

(StreetComplete 12.1 on Android 5.1)

The cycling cat

matkoniecz:
> Proposal: Never show the opening hours quests when the object is tagged with "building=sports_hall".

Or rather simply drop `leisure=sports_centre` (https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt)

westnordost:
A sports hall on school grounds that is not really public should not be tagged as `leisure=sports_centre` though. Or what kind of sports halls are we talking about?

cyclingcat:
The ones I'm talking about are those local handball/volleyball/table tennis clubs exercise and play their tournament matches. They are not always connected to school grounds, although they are probably used by schools, too.

The cycling cat

westnordost:
And these clubs do not have opening hours? Don't they have an office or something which would have opening hours?

westnordost:
Also, would there be a tagging available that distinguishes local sports clubs from those more commercial places like fitness studios and so on?

cyclingcat:
Hm... even if there are opening hours for those kinds of buildings: If they aren't accessible while standing in front of it and you thus have to do further research (e.g. on the websites of the village or the club) in order to answer the quest, this would this be out of the scope of SC, wouldn't it?

OSM wiki mentions "leisure=fitness_centre" (https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfitness_centre), but it seems to be a little more complicated, see https://wiki.openstreetmap.org/wiki/Gym_/_Fitness_centre.

The cycling cat

westnordost:
Yes, it wouldn't. Okay your argumentation convinced me, I'll remove `sports_centre` from the list.

PR #3982, opened 2022-04-21
Add more offices

dbdean:
Added a few more `office` options to the Opening Hours and Wheelchair quests. Also added some that seemed appropriate to the `isShopExpressionFilter` function too.

The following were added because they had more than [10K instances from taginfo](https://taginfo.openstreetmap.org/keys/?key=office#values), and had decent current usage of opening_hours:

- estate_agent (was already in wheelchair and shop expression)
- lawyer (was already in wheelchair)
- telecommunication
- educational_institution
- association
- ngo
- it
- accountant

I only included telecommunication, it, lawyer and accountant in the shop expression, as the others seemed less likely to be present in usual retail environments.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
@@ -96,6 +96,7 @@ fun isShopExpressionFragment(prefix: String? = null): String {
             ),
             "office" to arrayOf(
                 "insurance", "travel_agent", "tax_advisor", "estate_agent", "political_party",
+                "lawyer", "telecommunication", "it", "accountant"
             ),
             "craft" to arrayOf(
                 "shoemaker", "tailor", "photographer", "watchmaker", "optician",
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -79,6 +79,8 @@ class AddOpeningHours(
                 // common
                 "insurance", "government", "travel_agent", "tax_advisor", "religion",
                 "employment_agency", "diplomatic", "coworking",
+                "estate_agent", "lawyer", "telecommunication", "educational_institution",
+                "association", "ngo", "it", "accountant"
             ),
             "craft" to arrayOf(
                 // common
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
@@ -73,9 +73,11 @@ class AddWheelchairAccessBusiness : OsmFilterQuestType<WheelchairAccess>() {
                 // common
                 "insurance", "government", "travel_agent", "tax_advisor", "religion",
                 "employment_agency", "diplomatic", "coworking",
+                "estate_agent", "lawyer", "telecommunication", "educational_institution",
+                "association", "ngo", "it", "accountant"
 
                 // name & wheelchair
-                "lawyer", "estate_agent", "political_party", "therapist"
+                "political_party", "therapist"
             ),
             "craft" to arrayOf(
                 // common


FloEdelmann (changes requested):
I think many `office=it` elements are not customer-facing "shop"s, but rather, well, the offices of software companies. So asking for opening hours and wheelchair access may lead to many "can't answer" notes there.

I think `office=estate_agent`, `office=lawyer`, `office=accountant` are fine; I can't really tell for the others.

westnordost:
See @FloEdelmann , certain office=* are not not in the list because there haven't been tagged enough of them but because they are not "walk in". Anything that is not (highly likely) walk in should not be in these quests.

dbdean:
Absolutely, @FloEdelmann. Sorry if it wasn't clear that I had taken that into account.

The only ones I personally wanted was `estate_agent`, `lawyer` and `accountant`, but I wanted to make sure that I was a bit more consistent in guaging what walk-in places were from how things are actually mapped, rather than my perceptions.

I did this a bit haphazardly before, but I've taken the time to look up each of the existing and my proposed new quests using taginfo, to see if we can determine how people are tagging these type of offices in the wild already:

office= | existing osm quests | count | opening_hours | wheelchair
-- | -- | -- | -- | --
telecommunications | none | 29,480 | 38% | n/a
insurance | opening_hours, wheelchair | 51,480 | 27% | 7%
diplomatic | opening_hours, wheelchair | 14,896 | 21% | n/a
employment_agency | opening_hours, wheelchair | 10,396 | 20% | n/a
it | none | 13,602 | 20% | n/a
lawyer | wheelchair | 32,227 | 15% | 6%
estate_agent | wheelchair | 61,974 | 14% | 6%
ngo | none | 18,210 | 14% | n/a
association | none | 20,656 | 11% | 6%
government | opening_hours, wheelchair | 204,648 | 11% | 3%
accountant | none | 10,593 | 10% | n/a
educational_institution | none | 27,343 | 8% | n/a
religion | opening_hours, wheelchair | 10,113 | n/a | n/a
tax_advisor | opening_hours, wheelchair | 7,392 | n/a | n/a
political_party | wheelchair | 5,344 | n/a | n/a
travel_agent | opening_hours, wheelchair | 4,054 | n/a | n/a
therapist | wheelchair | 2,699 | n/a | n/a

I've sorted the table by existing proportion of items that already have `opening_hours`. n/a means that taginfo didn't have the number, which I believe means only a small number of examples (or none) exist.

I believe all of my proposed additions fit in well with the current usage out in the wild, at least as far as `opening_hours` go, but I am certainly open to compromising. I could see removing `educational_institution` for example. As I know that there are plenty of accountants around that have opening hours, I wouldn't want to remove anything that has an existing high level of usage than them, though.

Wheelchair usage is a bit harder to gauge, I'd certainly be happy to move some of these to `opening_hours` but not `wheelchair`, if it would help get this PR passed.

mnalis:
> I think many places with `office=it` are not customer-facing "shop"s, but rather, well, the offices of software companies. So asking for opening hours and wheelchair access may lead to many "can't answer" notes there.

True. It may be good idea for marking them with `access=private` if they are not "walk in" (assuming the quests would ignore offices marked private; as they should). Perhaps as `Other answers` option like `"Can't say, it is private"`?

@dbdean if changes to `isShopExpressionFragment()` get merged, please follow instructions in comment above it and open an issue/PR at https://github.com/mnalis/StreetComplete-taginfo-categorize so I can add them before next update run of it. (I'll try to spot / remember, but it's easy to miss for me in midst of all notifications)

dbdean:
@mnalis, on the 'Can't say - is private' option, see https://github.com/streetcomplete/StreetComplete/issues/2051.

westnordost:
I have not been replying further so far because 1. not sure if Flo will guide this or not and 2. more importantly, I reckon we need some kind of clear criteria when to include shop types that **maybe** are walk-in and maybe not. Because recently, `leisure=horse_riding` was removed from the list by @matkoniecz following #3978 because too few such amenities are walk-in like that (i.e. have publicly posted opening hours).
If there is no clear criteria, it will be a constant tug of war between people who want the question to be asked for more amenities and for people who want it to be asked for less amenities, with us maintainers in the middle.

Any ideas for that?

matkoniecz:
> `leisure=horse_riding` was removed from the list by @matkoniecz following https://github.com/streetcomplete/StreetComplete/issues/3978 because too few such amenities are walk-in like that (i.e. have publicly posted opening hours).

It was even in worse situation: it may have multiple separate opening hours (different for walk ins, different for signed up customers, different for arranging use, different for people with arranged meeting - and different objects will have different ones posted)

#1423 was a similar case

dbdean:
@FloEdelmann, when you are able, can I get an indication of what changes you would like to be made to get this PR in?

As I said, I am happy to compromise on which types of offices get included, but I would like to base the decision on [actual usage data from taginfo](https://github.com/streetcomplete/StreetComplete/pull/3982#issuecomment-1107697507), rather than our perceptions of what does and does not have opening hours.

FloEdelmann (approved):
Thanks @dbdean for looking at Taginfo and summarizing the results in a table! That certainly looks like tagging `opening_hours` to `office=it` is way more common than I thought.

There might be one missing information: How many elements with `office=it` that have real opening hours are already tagged with `opening_hours`? That percentage might be quite high already (making the quest for the rest of them quite spammy with "has no opening hours") or it might be a rather uniform distribution (making the quest useful).

I think it makes sense to assume a uniform distribution, as we can't really find out anyway. So let's go with your additions!

FloEdelmann:
> one solution might be to put some of these items on a 'does X still exist' list even if asking for opening hours isn't really needed. I think hotels, motels and BnBs would be good for that - they don't really have opening hours, but I would like to be asked if they are still operating every so often.

That wouldn't work with the current `CheckExistence` quest, as it will delete the node (and doesn't work for ways/areas). Maybe we could have another quest similar to `CheckShopType` that is asked regularly for all old shops/hotels/etc.

matkoniecz:
> That wouldn't work with the current CheckExistence quest, as it will delete the node (and doesn't work for ways/areas)

It could work for hotels and offices mapped as nodes.

PR #3992, opened 2022-04-27
Improve the ordering of some quests

peternewman:
To show ones which can be done from further away or are simpler first


FloEdelmann (approved):
LGTM

matkoniecz (commented):
seems to be an improvement based on looking on code - not tested

have you tried compiling and running it?

westnordost (approved):

Issue #3993, opened 2022-04-27
Fix #1977 for the new four option quests


peternewman:
**Use case**
Fix #1977 but for the new four option quests (seating and card types).

**Proposed Solution**
Either shrink the gap between answers, push the question up a bit again or at least make the default options match (as I've just spotted).

peternewman:
> at least make the default options match (as I've just spotted).

See #3994 for this part of it...

westnordost:
Ok I don't know what you think needs to be done here and wether it is already done / how much of it is done. Please open a new issue with screenshots to illustrate if there is still an issue somewhere.

PR #3994, opened 2022-04-27
Put the most likely/popular options first for seating

peternewman:
To match the app behaviour across the board...

Yes there's a small risk of muscle memory with changing the order (although while I feel like I've done quite a few of these I've only actually done 81...).

More importantly it makes it consistent with the card quest and e.g. the surface ones, where the most popular/likely options are first (far more places will have indoor seating or both, than takeaway only).

Some progress on, but doesn't completely finish #3993

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeatingForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeatingForm.kt
@@ -6,9 +6,9 @@ import de.westnordost.streetcomplete.quests.TextItem
 
 class AddSeatingForm : AListQuestAnswerFragment<Seating>() {
     override val items = listOf(
-        TextItem(Seating.NO, R.string.quest_seating_takeaway),
+        TextItem(Seating.INDOOR_AND_OUTDOOR, R.string.quest_seating_indoor_and_outdoor),
         TextItem(Seating.ONLY_INDOOR, R.string.quest_seating_indoor_only),
         TextItem(Seating.ONLY_OUTDOOR, R.string.quest_seating_outdoor_only),
-        TextItem(Seating.INDOOR_AND_OUTDOOR, R.string.quest_seating_indoor_and_outdoor),
+        TextItem(Seating.NO, R.string.quest_seating_takeaway),
     )
 }


FloEdelmann (approved):
Makes sense, I guess

westnordost:
What makes you think that "yes" is the most likely choice?

FloEdelmann:
Raw numbers from Taginfo:

* [`indoor_seating`](https://taginfo.openstreetmap.org/keys/indoor_seating#values): 3.2k `yes`, 0.4k `no` 
* [`outdoor_seating`](https://taginfo.openstreetmap.org/keys/outdoor_seating#values): 132k `yes`, 35k `no`
* [`takeaway`](https://taginfo.openstreetmap.org/keys/takeaway#values): 286k `yes`, 14.7k `only`, 9.2k `no`

westnordost:
5.99% of `amenity=restaurant` have [outdoor_seating](https://taginfo.openstreetmap.org/keys/outdoor_seating)=* tagged at all
4.69% of `amenity=restaurant` have [outdoor_seating](https://taginfo.openstreetmap.org/keys/outdoor_seating)=yes tagged

This does not imply that 78.30 % of restaurants have outdoor seating.

matkoniecz:
And many people do not tag absence of thing.

westnordost:
That's my point.

peternewman:
> And many people do not tag absence of thing.

> That's my point.

Surely that's in favour of yes being most popular, as people are more likely to tag the positive answers? Or do you mean tag in general as opposed to answer SC quests?

Looking at the more general case there are ~3x more restaurants ("A restaurant sells full sit-down meals") than fast food, and about the same cafes ("A generally informal place with sit-down facilities") as fast food ("A place concentrating on very fast counter-only service and take-away food"):
https://taginfo.openstreetmap.org/keys/amenity#values

Obviously many fast food places (e.g. McDonalds) also have seating, so I'd assume we're all in agreement that no/no is the **least** popular answer.

I suspect whether indoor **AND** outdoor or indoor only is more popular probably varies by country based on the temperature, but I figured having both at the top felt right (probably too long thinking about binary logic truth tables). Maybe technically indoor, both, outdoor, none is more accurate of reality, or probably technically indoor, both, none, outdoor, but neither of those feel very sensible layouts to me as stuff won't be where you'd expect it/you'll need to read more. Also because of the screen layout, you always have to move your finger a lot between the quest pin, the answer and the okay button, so you're not really optimising for that.

matkoniecz:
> Surely that's in favour of yes being most popular, as people are more likely to tag the positive answers? Or do you mean tag in general as opposed to answer SC quests?

It was about tagging in general.

People tend to mark places where

- something is present
- unusual situation is present

and tend to less often mark

- absence of a thing
- default/obvious/typical state

At least in Poland vast majority of eateries is indoor only, with indoor seating.

westnordost:
Order "indoor only", "both", "outdoor only", "no" would be ok. Switching around the last two ones is ok too.

mnalis:
In Croatia, most **abundant** ones are the cafes (to the point of tourists getting the impression that the whole country does nothing but sit outside and sip coffee for whole day, which is not _completely_ true) with **both indoor and outdoor** seating. Followed by **indoor only**, and then the remaining two options being equally unprobable. 

So over here originally suggested **both** => **indoor** => **outdoor** => **no** order seems ideal (and it also has an advantage of being logically ordered too; i.e. from most seating, via middle options, to least seating)

westnordost:
ok fine

peternewman:
Yeah I think that's probably the same in the UK @mnalis , despite a considerably less warm climate, everyone seems to try and get a few seats outside their shop so people can "enjoy" the weather.

PR #4002, opened 2022-05-02
Add empty link to mastodon profile for verification purposes

RubenKelevra:


westnordost:
?

Cj-Malone:
This would be used to verify "inofficial account | managed by RubenKelevra"... Seems to contradict itself.

@RubenKelevra [en.osm.town](https://en.osm.town/about) may be a better home than [mastodon.social](https://mastodon.social/about).

RubenKelevra:
It's not to verify the account, but the other way around. Its to verify the URL in the account.

While the account is not "official" the URL is.

And no, I won't move to a different server, especially not one which I don't know anything about (and which not even got 400 users total).

westnordost:
Sorry, I don't get it 😕

README.md
```diff
@@ -71,3 +71,5 @@ Many people are currently supporting this app through <a href="https://github.co
 
 <a href="https://osmfoundation.org/"><img src=".github/images/logo_osmf.png" height="58"/></a><br/>
 The <a href="https://osmfoundation.org/">OpenStreetMap foundation</a> was funding the development of doing <a href="https://wiki.openstreetmap.org/wiki/Microgrants/Microgrants_2020/Proposal/Map_Maintenance_with_StreetComplete">map maintenance with StreetComplete</a> (~ 3 weeks) in their <a href="https://blog.openstreetmap.org/2020/07/01/osmf-microgrants-program-congratulations-to-selected-projects/">first round</a> of the <a href="https://wiki.osmfoundation.org/wiki/Microgrants">microgrant program</a> in 2020.
+
+<a rel="me" href="https://mastodon.social/@streetcomplete"></a>
```
matkoniecz, line 75: It seems to identify this social media account as official, while it is not.

If it is doing anything else - can you link relevant documentation?

RubenKelevra: Here's the documentation about that:

https://docs.joinmastodon.org/user/profile/

The idea is to make the link in the profile verified. So it appears to the user to be "safe to click on" since it's the official link for the profile.

I do already operate the twitter profile, after talking to @westnordost about this via email to which terms he's comfortable with that.

I only want users to be sure that the link is safe if they click on it. :)

RubenKelevra: Maybe @rugk can chime in here as well, as he recommended to create a mastodon account - how he feels about adding a verified link. :)

rugk: Yeah. Its good and makes it look more official/supported by SC.

RubenKelevra:
@westnordost wrote:
> Sorry, I don't get it confused

Well, the link gets a checkmark to show that it's an official link. There's no such thing as "verifications of profiles" as official etc. on Mastodon. But I think it's better to have a link verification, to show that the profile is linking to the official project page.

README.md
matkoniecz: > However, Mastodon can cross-reference the links you put on your profile to prove that you are the real owner of those > links.
> (...)
> If you put a link in your profile metadata, Mastodon checks if the linked page links back to your Mastodon profile. If so, you get a verification checkmark next to that link, since you are confirmed as the owner.

So it seems that it would authenticate you as an official profile of SC on Mastodon. It seems that

> It's not to verify the account, but the other way around.

is not an accurate description.

matkoniecz: https://docs.joinmastodon.org/user/profile/

> However, Mastodon can cross-reference the links you put on your profile to prove that you are the real owner of those links.
> (...)
> If you put a link in your profile metadata, Mastodon checks if the linked page links back to your Mastodon profile. If so, you get a verification checkmark next to that link, since you are confirmed as the owner.

So it seems that it would authenticate you as an official profile of SC on Mastodon. It seems that

> It's not to verify the account, but the other way around.

is not an accurate description.

westnordost:
So, see @matkoniecz https://github.com/streetcomplete/StreetComplete/pull/4002#discussion_r863682676 , the mastodon account is as official as the twitter account. Hence, it would be misleading to have it show up as "verified". 

That being said, I'll add a mention of this account in https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/res/raw/credits_projects.yml

PR #4007, opened 2022-05-03
Add sign for `mutcd text spanish` no stopping style

FloEdelmann:
Related to streetcomplete/countrymetadata#1.

I've used https://commons.wikimedia.org/wiki/File:MUTCD-PR_R7-4.svg (Public Domain) and https://github.com/streetcomplete/StreetComplete/blob/master/res/graphics/street%20parking/no%20stopping%20sign/mutcd_text.svg (no attribution) as a base, so I think I don't have to give attribution?


westnordost:
LGTM, thanks!

matkoniecz:
> so I think I don't have to give attribution?

Still, it would be useful to document source even if attribution is not needed.

PR #4008, opened 2022-05-03
Add bicycle rental capacity quest

### LINKED ISSUE #2129: Bicycle Rental (docking station) capacity
### General
Affected tag(s) to be modified/added: [capacity](https://wiki.openstreetmap.org/wiki/Key:capacity)
Question asked: **What is the capacity of this bicycle rental dock?**

### Checklist
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort) - London has 292 candidates alone.

### Ideas for implementation
Could just be a rehash of #1181, as in directly part of the bike parking quest (possibly just with different text). Although personally I find the bike parking ones too ambiguous to answer whereas the docks have very clear demarked positions for wheels you can count.

**Element selection:** 
`amenity=bicycle_rental and capacity!=*`

**Proposed GUI:**
Same as the other two capacity quests.

FloEdelmann:
Fixes #2129.

Test/screenshot location: https://www.openstreetmap.org/node/9135891817


app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
```diff
@@ -0,0 +1,38 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+
+class AddBikeRentalCapacity : OsmFilterQuestType<Int>() {
+
+    override val elementFilter = """
+        nodes, ways with
+         amenity = bicycle_rental
+         and access !~ private|no
+         and bicycle_rental ~ docking_station|dropoff_point
```
FloEdelmann, line 18: I've enabled it for `bicycle_rental ~ docking_station|dropoff_point` (in #2129, only `docking_station` was suggested) as for both the capacity seems to be clearly visible.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
```diff
@@ -0,0 +1,38 @@
...
+         and (!capacity or capacity older today -6 years)
+    """
+
+    override val changesetComment = "Add bicycle rental capacities"
+    override val wikiLink = "Tag:amenity=bicycle_rental"
+    override val icon = R.drawable.ic_quest_bicycle_rental
```
FloEdelmann, line 24: As an icon, I used https://github.com/streetcomplete/StreetComplete/blob/v43.0/res/graphics/quest/bicycle_rental.svg for now. We might need to distinguish between bicycle rental capacity (#2129) and type (#3995) in the future, so maybe the icon still needs some work.

app/src/main/res/values/strings.xml
```diff
@@ -657,6 +657,8 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_bicycle_parking_type_building">Building</string>
     <string name="quest_bicycle_parking_type_handlebarholder">Handlebar holder</string>
 
+    <string name="quest_bicycle_rental_capacity_title">How many bikes can be rented or returned here?</string>
```
FloEdelmann, line 660: @ Native speakers (@peternewman @smichel17) wording suggestions? I'm not quite convinced with my suggestion yet.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
peternewman: Dockless may well have the same counting issues I experience with bike hoops hence why I raised the issue, see e.g.:
http://prod-upp-image-read.ft.com/6cccaaac-76d3-11e7-90c0-90a9d1bc9691

Or the example node you linked to is here:
https://goo.gl/maps/fbYQxYoayNf4ihCm9

Part of the issue is the tagging scheme doesn't differentiate between dockless and a bike rack with reserved slots for a rental scheme (where the bike locks itself), lumping them in together. It only differentiates the docking station (where the rack locks the bike).

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
peternewman: Maybe if the key was over the top of the bike (or cleverly formed into a wheel/the frame or something fancy), you could double up bikes like the other capacity quests?

app/src/main/res/values/strings.xml
peternewman: Yeah that wording would worry me that people would either try and count the bikes currently docked or the free slots.

How about (modifying the existing bike quest):
```suggestion
    <string name="quest_bicycle_rental_capacity_title">How many rental bikes can be parked here?</string>
```
```suggestion
    <string name="quest_bicycle_rental_capacity_title">How many rental bike spaces are there here?</string>
```
```suggestion
    <string name="quest_bicycle_rental_capacity_title">How many rental bike docks are there here?</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -316,12 +317,13 @@ fun questTypeRegistry(
...
     /* ↓ 2.solvable when right in front of it but takes longer to input --------------------- */
 
-    // bike parking: would be higher up if not for bike parking capacity which is usually not solvable when moving past
+    // bike parking/rental: would be higher up if not for bike parking/rental capacity which is usually not solvable when moving past
     AddBikeParkingCover(), // used by OsmAnd in the object description
     AddBikeParkingType(), // used by OsmAnd
     AddBikeParkingAccess(),
     AddBikeParkingFee(),
     AddBikeParkingCapacity(), // used by cycle map layer on osm.org, OsmAnd
+    AddBikeRentalCapacity(),
```
peternewman, line 326: Less ambiguous (for docked at least), so less thinking time than normal parking.
```suggestion
    AddBikeRentalCapacity(),
    AddBikeParkingCapacity(), // used by cycle map layer on osm.org, OsmAnd
```

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
FloEdelmann: You're right. I restricted the quest to `bicycle_rental=docking_station` in [`3773f50`](https://github.com/streetcomplete/StreetComplete/pull/4008/commits/3773f50cbf7865d76cb8fbcb32331c6f7243e469).

peternewman (changes requested):
Improve the quest order.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
peternewman: If the tagging was ever fixed, we could always tweak it to cover the non-dockless ones too.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
FloEdelmann:
```suggestion
    AddBikeRentalCapacity(), // less ambiguous than bike parking
    AddBikeParkingCapacity(), // used by cycle map layer on osm.org, OsmAnd
```

app/src/main/res/values/strings.xml
westnordost: > How many rental bike spaces are there here?

"there here"?

> How many rental bike docks are there here?

I am not sure if "dock" would be the most common word for rental stations. 

I find the original text or the 2nd suggestion (without "there here") best currently. Though maybe something with "at maximum capacity" would work too?

app/src/main/res/layout/quest_bike_rental_capacity.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
```
westnordost, line 1: Isn't this exactly the same layout as for the bicycle capacity quest? (If yes, why not use that layout. Actually, why not use the same form?)

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
westnordost: When/if the other quest type gets added, this icon could just be two keys on top of each other. The background color is already hint enough that it is about bicycles.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
```diff
@@ -0,0 +1,38 @@
...
+    override val isDeleteElementEnabled = true
+    override val questTypeAchievements = listOf(BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_bicycle_rental_capacity_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with amenity ~ bicycle_parking|bicycle_rental")
```
westnordost, line 31: What is the rationale for also showing bicycle parkings here? (In any case, the rationale should be documented as a comment)

westnordost (commented):
Looks good!

I wonder if releasing this quest should be delayed until a bicycle rental type quest has been created, so surveyors don't need to visit a place twice for solving the related quest types.

HolgerJeromin:
> I wonder if releasing this quest should be delayed until a bicycle rental type quest has been created, so surveyors don't need to visit a place twice for solving the related quest types.

We could disable the quest till the other quest is ready.

app/src/main/res/values/strings.xml
10992-osm: > (in reply to @peternewman)
> 
> > How many rental bike spaces are there here?
> 
> "there here"?

"there here" is acceptable to me, but it would probably also work without the *here* (as long as the name/type of the docking station is presented).

> 
> > How many rental bike docks are there here?
> 
> I am not sure if "dock" would be the most common word for rental stations.
> 
> I find the original text or the 2nd suggestion (without "there here") best currently. Though maybe something with "at maximum capacity" would work too?

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
FloEdelmann: Done in [`396dfbb`](https://github.com/streetcomplete/StreetComplete/pull/4008/commits/396dfbb7b6805f28de893eaf1e77960f85e177e2).

app/src/main/res/values/strings.xml
FloEdelmann: Changed in [`4515039`](https://github.com/streetcomplete/StreetComplete/pull/4008/commits/4515039b68a65d84cc6c7b8893aedc1bf70eab32).

> as long as the name/type of the docking station is presented

It is, just like in every other quest.

app/src/main/res/layout/quest_bike_rental_capacity.xml
FloEdelmann: The hint in the quest form doesn't make sense for this quest type:

> Note that most normal stands can be used from both sides to park each one bike.

But the quest form can of course be reused by just hiding that hint. Done in [`844328e`](https://github.com/streetcomplete/StreetComplete/pull/4008/commits/844328e02c45dc5b331202f47726e20a558327fe).

app/src/main/java/de/westnordost/streetcomplete/quests/bike_parking_capacity/AddBikeParkingCapacityForm.kt
```diff
@@ -17,6 +19,7 @@ class AddBikeParkingCapacityForm : AbstractQuestFormAnswerFragment<Int>() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        binding.clarificationText.isGone = !showClarificationText
```
westnordost, line 22: I fear this won't work like that :-/

Think about when the fragment is destroyed and recreated. The Android system can destroy the view and later recreate it when it is not in foreground.

I know, the proper way to pass parameters to a fragment is horribly verbose but I guess there is no other proper way to do this.

FloEdelmann: Hmm, bummer. But makes sense.

Fixed in [`04b3dac`](https://github.com/streetcomplete/StreetComplete/pull/4008/commits/04b3dac7e104cb786fccbdfb4dd6394855eb6501).

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
peternewman: I think it seems sensible, a drop-off point could look very similar, or even be identical to normal bike parking:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_rental#Types_of_bicycle_rental

Would it make sense to do the reverse in the normal bike quests too?

westnordost: The quest is not asked for dropoff points

peternewman: Ah yes, very true.

I think it's helpful in both directions though to see the other mapped bike parking, to help establish if something has been mis-mapped or removed if you're having to leave a note.

westnordost: The reason why I have been bringing this up is because I think what is shown when highlighting elements must be consistent, i.e. according to what the user would expect. That necessitates that there is a clear guideline what is shown here and what not.
This guideline has not been written down anywhere (yet), but I **think** all quests so far only show same elements that are nearby. E.g. the bench backrest quest just shows nearby benches, but not nearby picnic tables, bleachers or loungers.

FloEdelmann: You're right, with a few (very) minor exceptions:

* `AddBicyclePump` also highlights `compressed_air`, despite even excluding it in the quest
* `AddDrinkingWater` also highlights `amenity=drinking_water`, despite the quest not being asked for those
* `AddPowerPolesMaterial` also highlights power lines

But I guess you're right, let's keep it consistent then.

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_capacity/AddBikeRentalCapacity.kt
westnordost: Thank you for checking. For power poles it definitiely makes sense. The question for drinking water is ambiguous enough "water potable here?" that a user may confuse that the question is intended to be asked for a nearby dedicated drinking water fountain, so I think this is fine. For the bicycle pump, I wouldn't mind if this is removed as I guess there is probably no risk for confusion.

peternewman: > The reason why I have been bringing this up is because I think what is shown when highlighting elements must be consistent, i.e. according to what the user would expect. That necessitates that there is a clear guideline what is shown here and what not. This guideline has not been written down anywhere (yet), but I **think** all quests so far only show same elements that are nearby. E.g. the bench backrest quest just shows nearby benches, but not nearby picnic tables, bleachers or loungers.
> 
> Changing this guideline to "show same and elements that are somehow similar and related" I think would not be a good idea because there is no clear line to set what is still similar and what not, so it becomes inconsistent.

I'd argue the reverse, because although it might not be consistent, it should hopefully result in a better quality of data overall.

E.g. the bench quest lets you convert a bench to a picnic table:
https://github.com/streetcomplete/StreetComplete/blob/b0c5853bd0a9df24264b3e7fb2761ea02bbb1739/app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt#L40-L43

If there are bench and picnic table nodes nearby on the map, but only a picnic table remaining in real life, without seeing the other related nodes it would be quite easy to end up with two picnic tables.

Likewise OSM having a bike rental and a parking, but the parking having gone in reality, I might leave a confusing note saying this is actually a rental for a quest on the parking node.

Possibly a bit more controversial, bins and benches often go side by side. I've often found and left notes for the micro-mapping of them not being right (e.g. benches on the wrong side of a path, or bench and bin the wrong way round). Apart from the existence quest and a bit of Kim's Game memory, this is hard to do, but would be far simpler if they were shown on the alternate quest.

Essentially because the base SC map is so bare, we potentially miss other opportunities for the boots on the ground to improve the situation further.

mnalis: I agree with @peternewman here - better context leading to better data quality _should trump_ consistency. 

Which is actually why `AddBicyclePump` shows `compressed_air`, even if quest itself fills only `service:bicycle:pump=*` - bicycle shops and repair stations might have manual (i.e. `service:bicycle:pump`) or machine (i.e. `compressed_air`) air sources, and it might be tagged as either, according to individual mapper. Hiding that context from SC user would IMHO lead to lower data quality.

Afterall, we already disregard that consistency in other cases, like in the name of the ease of programming (e.g. `IS_SHOP_OR_DISUSED_SHOP_EXPRESSION` which contains things like `post_office` or `tanning_salon` is shown in quests like `AddHalal` (although `tanning_salon` is not likely to sell any food - Halal or not), instead of showing _only_ restaurants and other shops used in `AddHalal` quest element selection.

westnordost: > AddHalal

Because shop-type elements are shown when selecting a quest about a shop. This behavior is consistent. The highlighted elements are shown to help the user locate the object in question in the forest of similar elements. In this case, the fact that it is about a restaurant is not the important part, but that it is a shop.

westnordost: Anyway, I acknowledge the argument about the picnic table. Let's please discuss this in another ticket or discussion and collect there, which more elements should be shown if we were to soften the "show only same elements" guideline.

peternewman: > Let's please discuss this in another ticket or discussion and collect there, which more elements should be shown if we were to soften the "show only same elements" guideline.

See #4020!

app/src/main/res/values/strings.xml
```diff
@@ -657,6 +657,8 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_bicycle_parking_type_building">Building</string>
     <string name="quest_bicycle_parking_type_handlebarholder">Handlebar holder</string>
 
+    <string name="quest_bicycle_rental_capacity_title">How many rental bike spaces are there?</string>
```
westnordost, line 660:
```suggestion
    <string name="quest_bicycle_rental_capacity_title">How many rental bike spaces are here?</string>
```

westnordost (approved):

PR #4010, opened 2022-05-04
Add bicycle rental type quest

### LINKED ISSUE #3995: Quest suggestion: Bicycle Rental type

FloEdelmann:
Fixes #3995.


app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_type/AddBikeRentalType.kt
```diff
@@ -0,0 +1,37 @@
...
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_bicycle_rental_type_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with amenity ~ bicycle_parking|bicycle_rental")
+
+    override fun createForm() = AddBikeRentalTypeForm()
+
+    override fun applyAnswerTo(answer: BikeRentalType, tags: Tags, timestampEdited: Long) {
+        tags["bicycle_rental"] = answer.osmValue
+        if (answer == BikeRentalType.SHOP) tags["shop"] = "rental"
```
matkoniecz, line 35: `shop=bicycle` `service:bicycle:rental=yes` seems preferred and not requiring confirmation

matkoniecz (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_type/AddBikeRentalType.kt
FloEdelmann: From https://wiki.openstreetmap.org/wiki/Tag:amenity%253Dbicycle_rental:

> If the object to map is a bicycle shop which also happens to rent out bicycles, use `shop=bicycle` with `service:bicycle:rental=yes`.

> `bicycle_rental=shop` and `shop=rental`: A staffed, indoor shop whose **main business** is bike rental, e.g. for tourists in a touristical region.

That's why I added the confirmation prompt.

matkoniecz: maybe there should be options for both?

FloEdelmann: I thought about that first, but it's hard to find clear concise descriptions and meaningful but distinguishable images for both types.

matkoniecz: `Bicycle shop offering rental` vs `Human-operated bicycle rental`?

Maybe something with less prominent pile of bicycles as a photo? For example https://commons.wikimedia.org/wiki/File:Shkoder_088.JPG https://commons.wikimedia.org/wiki/File:Eizenga_Tweewielers,_Muntendam_(2017).jpg https://commons.wikimedia.org/wiki/File:Ferweda_Service_Center,_Groningen_(2020)_02.jpg https://commons.wikimedia.org/wiki/File:Rijwielspeciaalzaak_E.J._Noordhoff,_Delfzijl_(2019)_01.jpg or something else from https://commons.wikimedia.org/wiki/Category:Bicycle_shops_by_country

peternewman:
I like your pictures, they look better than the wiki ones!

More an issue with the tagging than anything else, but presumably you could get a load of bikes individually locked with bike locks in a dedicated spot, with the keys for those bike locks dispensed via machine...! I've never seen a key dispensing machine in the flesh for hire, so maybe that's actually how they work anyway?

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -316,9 +317,10 @@ fun questTypeRegistry(
 
     /* ↓ 2.solvable when right in front of it but takes longer to input --------------------- */
 
-    // bike parking: would be higher up if not for bike parking capacity which is usually not solvable when moving past
+    // bike parking/rental: would be higher up if not for bike parking/rental capacity which is usually not solvable when moving past
     AddBikeParkingCover(), // used by OsmAnd in the object description
     AddBikeParkingType(), // used by OsmAnd
+    AddBikeRentalType(),
```
peternewman, line 323: Minor re-ordering of the simpler quest:
```suggestion
    AddBikeRentalType(), // generally less overlap of possible types/fewer choices/simpler to answer
    AddBikeParkingType(), // used by OsmAnd
```

Only downside would be if there is rental next to parking it would flip-flop between the two.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_type/AddBikeRentalType.kt
peternewman, line 35: Isn't this more the usual SC style?
```suggestion
        if (answer == BikeRentalType.SHOP) {
          tags["shop"] = "rental"
        }
```

app/src/main/res/values/strings.xml
```diff
@@ -657,6 +657,14 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_bicycle_parking_type_building">Building</string>
     <string name="quest_bicycle_parking_type_handlebarholder">Handlebar holder</string>
 
+    <string name="quest_bicycle_rental_type_title">What type of bicycle rental is this?</string>
+    <string name="quest_bicycle_rental_type_docking_station">Docking station (locks bikes)</string>
+    <string name="quest_bicycle_rental_type_dropoff_point">Designated spot (bikes locked individually)</string>
```
peternewman, line 662: I don't really have a better suggestion right now, but it seems a bit odd that the comment in brackets for both could equally apply to the other one.

```suggestion
    <string name="quest_bicycle_rental_type_docking_station">Docking station (locks bikes in place)</string>
    <string name="quest_bicycle_rental_type_dropoff_point">Designated spot (bikes locked individually)</string>
```

peternewman (changes requested):
A few minor suggestions

FloEdelmann:
> I like your pictures, they look better than the wiki ones!

Thanks!

> I've never seen a key dispensing machine in the flesh for hire, so maybe that's actually how they work anyway?

There are currently only 10 places tagged with `bicycle_rental=key_dispensing_machine` worldwide: https://overpass-turbo.eu/s/1ihy
For these two, there is good Mapillary imagery:

* [node 9586518887](https://www.openstreetmap.org/node/9586518887): [Mapillary](https://www.mapillary.com/app/?pKey=155985169819163&lat=51.0647568&lng=3.1053712999991&z=17&focus=photo&x=0.5225364040039132&y=0.5371024550655779&zoom=1.401639344262295)
* [node 9722758036](https://www.openstreetmap.org/node/9722758036): [Mapillary](https://www.mapillary.com/app/?pKey=325511302960053&lat=51.0474674&lng=3.6549064&z=17&focus=photo&x=0.4337529047876969&y=0.6540920815734267&zoom=0)

So that's not much data of course, but the machines seems to be readily identifiable.

> More an issue with the tagging than anything else

Yep, I'd agree.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_type/AddBikeRentalType.kt
FloEdelmann: I like braces better, but only with correct indentation :wink: 

```suggestion
        if (answer == BikeRentalType.SHOP) {
            tags["shop"] = "rental"
        }
```

FloEdelmann: I'm not convinced by the images you suggested, and also couldn't find any better ones that were easily distinguishable from the bike rental shop image.

> with less prominent pile of bicycles

That also makes it less recognizable as a bicycle shop, as most of them prominently display what they sell (or rent, respectively).

I like your descriptions though. "Human-operated bicycle rental" is better than "Human-operated shop" even if we don't add that other answer, so I changed that already in [`663843c`](https://github.com/streetcomplete/StreetComplete/pull/4010/commits/663843cee71b68913aab83e46168157311078b7f).

app/src/main/res/values/strings.xml
FloEdelmann: You're right, that bugged me too. Maybe like this?

```suggestion
    <string name="quest_bicycle_rental_type_docking_station">Docking station which locks bikes</string>
    <string name="quest_bicycle_rental_type_dropoff_point">Designated spot (bikes locked individually)</string>
```

westnordost (commented):
The pictures kind of look really low quality, artifacts everywhere. Looking at some original pictures, they seem fine. Did you accidentally save them with a low JPEG quality? In GIMP, you can select to preview the output picture when using the quality slider. You should only slide it down as much as **no** artifacts are visible. IIRC something between 75-95 is fine, depends on the picture.

> "Is bike rental this shop’s main business? If not, please leave a note instead."

A dialog could be avoided by just putting this into the title. E.g. one option is "Bicycle rental shop", the other is "Bicycle shop that also rents bicycles".
On the other hand, I am not sure if this is is necessary at all. Since the question explicitly mentions that it is a bicycle rental, can't we expect that the user would leave a note then anyway (stating it is not a rental only)? At the very least, it is likely that this mistake is noticed the next time SC asks about opening hours because that quest would then display the object type as "bicycle rental shop" rather than "bicycle shop". (And TBH, there is a smooth transition between these two types of shops)

> Key dispensing machine

I strongly think this is so rare that it can be left out. The user can just leave a note in this case. 

If I understand this right, `bicycle_rental=key_dispensing_machine` is used **on the machine**, meaning it is a sort of vending machine right close to a an `amenity=bicycle_rental + bicycle_rental=dropoff_point` or docking station.
I cannot say whether such a key dispension machine is a very rare form of how bicycle unlocking is organized, but I'd say that it is not very likely that people who map such vending machines would not map this machine as  `amenity=bicycle_rental` without the appropriate subtag.

app/src/main/java/de/westnordost/streetcomplete/quests/bike_rental_type/AddBikeRentalType.kt
peternewman: Heh thanks, it was a bit of a quick write without checking!

app/src/main/res/values/strings.xml
peternewman: Yes, that feels better to me, I think I pondered something similar but then tried to match your brackets!

peternewman:
> If I understand this right, `bicycle_rental=key_dispensing_machine` is used **on the machine**, meaning it is a sort of vending machine right close to a an `amenity=bicycle_rental + bicycle_rental=dropoff_point` or docking station. I cannot say whether such a key dispension machine is a very rare form of how bicycle unlocking is organized, but I'd say that it is not very likely that people who map such vending machines would not map this machine as `amenity=bicycle_rental` without the appropriate subtag.

That makes sense, but makes the tagging even more crazy if you've then got two `bicycle_rental` tags adjacent to each other referring to the same service. Where would the cost or the name or whatever go?

westnordost:
> That makes sense, but makes the tagging even more crazy if you've then got two bicycle_rental tags adjacent to each other referring to the same service. Where would the cost or the name or whatever go?

So you think I understood this wrong, or is this note directed at the one who proposed this tagging scheme?

peternewman:
> So you think I understood this wrong, or is this note directed at the one who proposed this tagging scheme?

I'm honestly not sure which. The discussion page has this:

> Systems where one gets a bike at a fixed location and has to drop their bike at this same location again (e.g. Blue Bike in Belgium) - typically with a machine to retrieve the key... (Proposal: add [bicycle_rental:type](https://wiki.openstreetmap.org/w/index.php?title=Key:bicycle_rental:type&action=edit&redlink=1)=[key_dispensing_machine](https://wiki.openstreetmap.org/w/index.php?title=Tag:bicycle_rental:type%3Dkey_dispensing_machine&action=edit&redlink=1))
> ... and the variation where there is a (signposted) drop-off location, but the 'smart locks' are unlocked with an app (Proposal: bicycle_rental:type=dropoff_location). (One could also map the bike parking as area an additionally map the key_dispensing_machine, allthough this might get confusing)

In support of what you said, aside from it's mentioned about the smart locks rather than dispensing machines, and acknowledged as confusing.

How to map doesn't mention vending machines or anything:
https://wiki.openstreetmap.org/wiki/Tag:amenity=bicycle_rental#How_to_map

Presumably you meant:
```
amenity=bicycle_rental
bicycle_rental=dropoff_point
```

Then a node with:
```
amenity=bicycle_rental
bicycle_rental=key_dispensing_machine
```

My comment is if you do that, you'll never find any of:
```
amenity=bicycle_rental
bicycle_rental=key_dispensing_machine
```

But if you do as above, you'll have twice as many `amenity=bicycle_rental` and where do you tag the brand/operator etc...

The original proposal seems to suggest the simpler option with no node for the vending machine:
https://wiki.openstreetmap.org/wiki/Proposed_features/bicycle_rental:type#Tagging_the_infrastructure

FloEdelmann:
I've removed the "Key dispensing machine" option now, added a new "Bike store that also rents bikes" option and re-exported the graphics:

matkoniecz (approved):
note: I have not run code, only looked at it

westnordost (approved):

PR #4027, opened 2022-05-14
Always show available map tiles

### LINKED ISSUE #4015: Available old map tiles not used
When map tiles are older than 14 days, the files will remain in the cache, but the tiles are not displayed.
This can lead to situations where map data is available, but there is no background map shown.

**How to Reproduce**
1. Have a fresh install of StreetComplete
2. Download two areas (manual or automatic doesn't matter)
3. Wait a week
4. Open StreetComplete and download one of the areas again
5. Wait another week
6. Disable internet connection
7. Open StreetComplete and check both areas
8. See that map and quests in the area only downloaded 2 weeks ago has disappeared (not totally surprising)
9. See that map in the area downloaded 1 week ago has disappeared, but quests are still shown
10. Check `[internal storage]/Android/data/de.westnordost.streetcomplete/cache/tile_cache` and see a bunch of 2 week old files

So it seems the tile cache prefers not showing tiles over showing "too old" tiles.
At the same time the old unused tiles are not deleted, and apparently just sitting around waiting until the cache reaches maximum size.

Suggestions for solution, in order of my personal preference:
* Use available tiles if there is no internet connection (and download them again if there is connection)
* Always download map tiles when downloading an area to have the same download time for both
* Delete tiles after 14 days, as they aren't used for anything

**Versions affected**
StreetComplete 42.0 (F-Droid)

Helium314:
fixes #4015 by using `CacheControl` with a very high `maxStale` in `MapFragment` (used to download tiles when panning the map).

When panning, all map tiles available offline are used, no matter how old (well, up to 14000 days).
Since `CacheControl` in `MapTilesDownloader` is unchanged, old tiles will be updated as part of downloading / scanning for quests.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloadCacheConfig.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloadCacheConfig.kt
@@ -18,6 +18,13 @@ class MapTilesDownloadCacheConfig(context: Context) {
         .maxStale(DELETE_OLD_DATA_AFTER.toInt(), TimeUnit.MILLISECONDS)
         .build()
 
+    /* use separate cache control for tangram with large maxStale value to always show available
+    *  map tiles when panning the map, even without (or with bad) internet connection */
+    val tangramCacheControl = CacheControl.Builder()
+        .maxAge(12, TimeUnit.HOURS)
+        .maxStale(DELETE_OLD_DATA_AFTER.toInt(), TimeUnit.SECONDS)
+        .build()
+
     val cache: Cache?
 
     init {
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloadCacheConfig.kt
```diff
@@ -18,6 +18,13 @@ class MapTilesDownloadCacheConfig(context: Context) {
         .maxStale(DELETE_OLD_DATA_AFTER.toInt(), TimeUnit.MILLISECONDS)
         .build()
 
+    /* use separate cache control for tangram with large maxStale value to always show available
+    *  map tiles when panning the map, even without (or with bad) internet connection */
+    val tangramCacheControl = CacheControl.Builder()
+        .maxAge(12, TimeUnit.HOURS)
+        .maxStale(DELETE_OLD_DATA_AFTER.toInt(), TimeUnit.SECONDS)
```
westnordost, line 25: Wha is max stale if it is not set? Can it not simply not be set? If it must be set, maybe set to one year or something, reusing the DELETE_OLD_DATA_AFTER constant but interpreting the delay in seconds is confusing.

Helium314: Ok, I'll change it to some really long time explicitly.

> Can it not simply not be set?

Then stale cache will not be used, see https://github.com/square/okhttp/blob/master/okhttp/src/jvmMain/kotlin/okhttp3/CacheControl.kt#L172-L183

PR #4038, opened 2022-05-17
ask about bicycle shop details

### LINKED ISSUE #4012: Ask for services offered by a bicycle shop
### General
Affected tag(s) to be modified/added: [service:bicycle:retail](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbicycle)  [service:bicycle:repair](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbicycle) [service:bicycle:rental](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbicycle) ...
Question asked: **Mark services offered by this bicycle shop** / **Which services are offered by this bicycle shop?**

### Checklist
- [ ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary - may be necessary to disable it by default

### Ideas for implementation

**Element selection:** 
Skip objects where all (or any, or most) of this objects is tagged and all objects where any of this tags is present and was tagged recently
 

**Metadata needed:** 
None

**Proposed UI:**
Maybe resurvey with prefilled currently available things is a good idea?

matkoniecz:
fixes #4012

note: icons are missing

@westnordost Are you interested in making something for iicons here? Or should I try?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddBikeRepairAvailability.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddBikeRepairAvailability.kt
@@ -1,0 +1,41 @@
+package de.westnordost.streetcomplete.quests.bike_shop
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.osmquests.Tags
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+import de.westnordost.streetcomplete.quests.YesNoQuestAnswerFragment
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddBikeRepairAvailability : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
+        nodes, ways, relations with shop = bicycle
+        and (
+            !service:bicycle:repair
+            or service:bicycle:repair older today -6 years
+        )
+        and access !~ private|no
+    """
+
+    override val changesetComment = "Add whether bicycle shop offers repair"
+    override val wikiLink = "Key:service:bicycle:repair"
+    override val icon = R.drawable.ic_quest_car_air_compressor // TODO ADD AN ICON
+    override val questTypeAchievements = listOf(BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_bicycle_shop_repair_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter(IS_SHOP_OR_DISUSED_SHOP_EXPRESSION)
+
+    override fun createForm() = YesNoQuestAnswerFragment()
+
+    override fun applyAnswerTo(answer: Boolean, tags: Tags, timestampEdited: Long) {
+        tags.updateWithCheckDate("service:bicycle:repair", answer.toYesNo())
+    }
+}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddBikeRepairAvailability.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddBikeRepairAvailability.kt
@@ -24,6 +24,7 @@ class AddBikeRepairAvailability : OsmFilterQuestType<Boolean>() {
     """
 
     override val changesetComment = "Add whether bicycle shop offers repair"
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
     override val wikiLink = "Key:service:bicycle:repair"
     override val icon = R.drawable.ic_quest_car_air_compressor // TODO ADD AN ICON
     override val questTypeAchievements = listOf(BICYCLIST)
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -668,6 +668,14 @@ However, before uploading your changes, the app checks with a &lt;a href=\"https
     <string name="quest_bicycle_rental_type_human">Human-operated bike rental</string>
     <string name="quest_bicycle_rental_type_shop_with_rental">Bike store that also rents bikes</string>
 
+    <string name="quest_bicycle_shop_repair_title">Can you repair bicycles here?</string>
+
+    <string name="quest_bicycle_shop_second_hand_title">What type of bicycles can be bought here?</string>
+    <string name="quest_bicycle_shop_second_hand_only_new">Only new bicycles</string>
+    <string name="quest_bicycle_shop_second_hand_new_and_used">New bicycles and second-hand bicycles</string>
+    <string name="quest_bicycle_shop_second_hand_only_used">Only second-hand bicycles</string>
+    <string name="quest_bicycle_shop_second_hand_no_bicycles">No bicycles sold in this bicycle shop</string>
```
FloEdelmann, line 677:
```suggestion
    <string name="quest_bicycle_shop_second_hand_title">Are second-hand bicycles sold here?</string>
    <string name="quest_bicycle_shop_second_hand_only_new">No, only new ones</string>
    <string name="quest_bicycle_shop_second_hand_new_and_used">Yes, alongside new ones</string>
    <string name="quest_bicycle_shop_second_hand_only_used">Yes, exclusively</string>
    <string name="quest_bicycle_shop_second_hand_no_bicycles">No bicycles sold at all</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddSecondHandBicycleAvailability.kt
```diff
@@ -0,0 +1,60 @@
...
+            }
+            SecondHandBicycleAvailability.ONLY_SECOND_HAND -> {
+                tags.updateWithCheckDate("service:bicycle:retail", "yes")
+                tags["service:bicycle:second_hand"] = "only"
+            }
+            SecondHandBicycleAvailability.NO_BICYCLES_SOLD -> {
+                tags.updateWithCheckDate("service:bicycle:retail", "no")
+            }
+        }
+    }
```
FloEdelmann, line 59: Maybe extract the OSM tag values into variables of the `SecondHandBicycleAvailability` enum?

```kt
enum class SecondHandBicycleAvailability(osmValue: String?, retail: Boolean) {
    ONLY_NEW("no", true),
    NEW_AND_SECOND_HAND("yes", true),
    ONLY_SECOND_HAND("only", true),
    NO_BICYCLES_SOLD(null, false),
}
```

```suggestion
    override fun applyAnswerTo(answer: SecondHandBicycleAvailability, tags: Tags, timestampEdited: Long) {
        tags.updateWithCheckDate("service:bicycle:retail", answer.retail.toYesNo())
        answer.osmValue?.let { tags["service:bicycle:second_hand"] = it }
    }
```

app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddSecondHandBicycleAvailability.kt
```diff
@@ -0,0 +1,60 @@
...
+        )
+        and (
+            service:bicycle:retail != no
+            or service:bicycle:retail older today -6 years
+        )
+        and !second_hand
+        and access !~ private|no
+        """
+
+    override val changesetComment = "Add type of parking access"
```
FloEdelmann, line 28:
```suggestion
    override val changesetComment = "Add whether bicycle shop sells second-hand bicycles"
```

app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddSecondHandBicycleAvailability.kt
FloEdelmann: Or maybe with only one variable:

```kt
enum class SecondHandBicycleAvailability(osmValue: String?) {
    ONLY_NEW("no"),
    NEW_AND_SECOND_HAND("yes"),
    ONLY_SECOND_HAND("only"),
    NO_BICYCLES_SOLD(null),
}
```

```suggestion
    override fun applyAnswerTo(answer: SecondHandBicycleAvailability, tags: Tags, timestampEdited: Long) {
        if (answer.osmValue == null) {
            tags.updateWithCheckDate("service:bicycle:retail", "no")
        } else {
            tags.updateWithCheckDate("service:bicycle:retail", "yes")
            tags["service:bicycle:second_hand"] = answer.osmValue
        }
    }
```

app/src/main/res/values/strings.xml
matkoniecz: I had in one of drafts and I was wondering is it a bit too tricky. I guess I will go with it.

I just want to defend `No bicycles sold in this bicycle shop` vs `No bicycles sold at all` - if bicycle shop is replaced by say supermarket of greengrocer, them `No bicycles sold at all` is still applicable - but user should leave note instead.

FloEdelmann: Shouldn't the feature name "Bicycle shop" below the quest title suffice to raise concerns if there is e.g. a supermarket there now?

But I guess that being explicit doesn't hurt. It just sounds a bit repetitive.

matkoniecz: > Shouldn't the feature name "Bicycle shop" below the quest title suffice to raise concerns if there is e.g. a supermarket there now?

It should but I estimate that at least 10%, maybe up to 60% of people will miss it.

westnordost (commented):
Why is "second hand bikes" quest disabled by default but "bike repair services" not?

Yes, I can do icons

matkoniecz:
> Why is "second hand bikes" quest disabled by default but "bike repair services" not?

I expect one to require checking inside and one not, but if that is based too much on what I personally encountered I can apply the same "disabled by default" to both

westnordost:
> I expect one to require checking inside and one not, but if that is based too much on what I personally encountered I can apply the same "disabled by default" to both

I don't know, if I see a couple of obbviously used bikes in front of a bike shop, it could be **either** these are to be repaired / repaired bikes or it could be that they are selling second hand bikes.

westnordost:
Added icons

app/src/main/java/de/westnordost/streetcomplete/quests/bike_shop/AddBikeRepairAvailability.kt
```diff
@@ -0,0 +1,42 @@
...
+            !service:bicycle:repair
+            or service:bicycle:repair older today -6 years
+        )
+        and access !~ private|no
+    """
+
+    override val changesetComment = "Add whether bicycle shop offers repair"
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+    override val wikiLink = "Key:service:bicycle:repair"
+    override val icon = R.drawable.ic_quest_car_air_compressor // TODO ADD AN ICON
```
westnordost, line 29:
```suggestion
    override val icon = R.drawable.ic_quest_bicycle_repair
```

westnordost (approved):

PR #4041, opened 2022-05-17
Link to where the base language files are

peternewman:


peternewman:
I'd imagine that doing a .gitignore on the values-en version so it was never committed, and just copying it at compile time would also work and make some of this redundant?

westnordost:
Hm no, that's fine, thanks!

PR #4042, opened 2022-05-17
Make the both card types option consistent with indoor/outdoor seating

peternewman:
It also means you don't really need to pay quite as much attention to the question, as I'm sure some of us sometimes zone out.


westnordost:
Hmm, this makes sense

PR #4048, opened 2022-05-20
Remove "the" from OpenStreetMap and add acronym

arrival-spring:
I can't find an official style guide, but I've never come across OSM referred to with "the".

Also, the acronym should be introduced as it is used later.


FloEdelmann (approved):

peternewman (approved):
The about page says "OSM" but "the OSMF", so I think you're right, and apologies as I possibly introduced some of these:
https://www.openstreetmap.org/copyright/en

PR #4064, opened 2022-05-26
Ask for seating in bakeries

FloEdelmann:
See https://taginfo.openstreetmap.org/tags/shop=bakery#combinations:

grafik

I think it makes sense to ask there, as some bakeries might only be shops, and some might be more like a cafe, and determining this is valuable information.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt
@@ -14,7 +14,10 @@ class AddSeating : OsmFilterQuestType<Seating>() {
 
     override val elementFilter = """
         nodes, ways with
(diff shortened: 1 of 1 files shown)


matkoniecz:
in some regions it would be basically always "no", but this quest is disabled by default, so...

Helium314:
Should a separate node with `amenity=cafe` be added ins such cases? Wiki contains this somewhat weird sentence
> If a cafe is in the same building, add another point and the corresponding properties. (See [One feature, one OSM element](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element))

that doesn't indicate anything about connection between cafe and bakery (i.e. the same is true for **any** two places in the same building, and also in different buildings). But the specific mention seems to imply that a cafe/bakery should be mapped as two separate places.

The German Wiki on the other hand happily recommends combining `shop=bakery` with `cafe=yes` and seating tags.

PR #4066, opened 2022-05-27
Entrance reference/flat range

### LINKED ISSUE #3064: New quest: Add ref and/or addr:flats for entrance=staircase

matkoniecz:
Fixes #3064


app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    override fun applyAnswerTo(answer: EntranceAnswer, tags: Tags, timestampEdited: Long) {
+        when (answer) {
+            is FlatRange -> tags["addr:flats"] = answer.flatRange
+            is ReferenceCode -> tags["ref"] = answer.referenceCode
+            is ReferenceCodeAndFlatRange -> {
+                tags["addr:flats"] = answer.flatRange
+                tags["ref"] = answer.referenceCode
+            }
+            Unsigned -> tags["ref:signed"] = "no"
```
FloEdelmann, line 91: Assignments should be in their own block:

```suggestion
            is FlatRange -> {
                tags["addr:flats"] = answer.flatRange
            }
            is ReferenceCode -> {
                tags["ref"] = answer.referenceCode
            }
            is ReferenceCodeAndFlatRange -> {
                tags["addr:flats"] = answer.flatRange
                tags["ref"] = answer.referenceCode
            }
            Unsigned -> {
                tags["ref:signed"] = "no"
            }
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
```diff
@@ -0,0 +1,95 @@
...
+                    }
+                    if(entrance.tags.containsKey("ref")) {
+                        continue
+                    }
+                    returned.add(entrance)
+                }
+            }
+        }
+        return returned
+    }
```
FloEdelmann, line 79: Is it possible to refactor this without mutable lists to make it more concise and easier to read? Something like this:

```suggestion
    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
        // note: it does not support multipolygon buildings
        val buildingWays = mapData.ways.asSequence()
            .filter { buildingFilter.matches(it) }

        return buildingWays.flatMap { way ->
            val entrancesForApartments = way.nodeIds
                .mapNotNull { mapData.getNode(it) }
                .filter { it.tags["entrance"] == "main" || it.tags["entrance"] == "staircase" }

            if (entrancesForApartments.count() < 2) {
                return@flatMap emptyList()
            }

            return@flatMap entrancesForApartments.filter {
                it.tags["noref"] != "yes"
                && it.tags["ref:signed"] != "no"
                && !it.tags.containsKey("addr:flats")
                && !it.tags.containsKey("name")
                && !it.tags.containsKey("ref")
            }
        }
    }
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReferenceForm.kt
```diff
@@ -0,0 +1,134 @@
...
+        val flatRangeEnd = flatRangeEndInput?.nonBlankTextOrNull
+        val flatRange = if(flatRangeStart != null && flatRangeEnd != null) {
+            if(flatRangeStart == flatRangeEnd) {
+                flatRangeStart
+            } else {
+                "$flatRangeStart-$flatRangeEnd"
+            }
+        } else {
+            null
+        }
```
FloEdelmann, line 114:
```suggestion
        val flatRange = when {
            flatRangeStart == null || flatRangeEnd == null -> null
            flatRangeStart == flatRangeEnd -> flatRangeStart
            else -> "$flatRangeStart-$flatRangeEnd"
        }
```

FloEdelmann (commented):
Just some code style improvement suggestions:

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
matkoniecz: maybe but it still does not seem at all more readable to the current solution

or is there some other benefit?

FloEdelmann: I think it's much more readable, but that is subjective of course. It should not have a (noticeable) impact on performance or change the logic, so there are no other benefits.

matkoniecz: For me at least map-reduce-filter code is not at all more readable.

matkoniecz:
Good idea, I recoloured it to make it distinctive (in brown, to match colour of door in entrance type quest)

matkoniecz:
Final testing in emulator:

Peek 2022-06-03 15-33

westnordost:
FYI you can screen-record with Android Studio: 

image

And you can upload mp4 files to github. If the resulting mp4 is too large, you can re-encode and/or downsize them with HandBrake

westnordost:
From looking at the gif, do you have a tab order, i.e. that you can tap "next" on the keyboard to get to the next input field?

matkoniecz:
> FYI you can screen-record with Android Studio:

Ooo, one more nice feature that I missed! Thanks! (I was using Peek which allows to record anything on screen but this would be much better)

> From looking at the gif, do you have a tab order, i.e. that you can tap "next" on the keyboard to get to the next input field?

I have not tried this before - I guess that it would be blue checkmark button? Right now it is not implemented. Are there quest interfaces where it is implemented? Housenumber quest in house name mode is also without this (and I also implemented it, IIRC).

Multilanguage name input is without it, but it is less important there as you need to manually add languages, which places cursor in a new language

BTW, house number quest recently lost white border decoration - is it intentional? Should it be removed also here? Or restored there?

westnordost:
Nah, removing the border that was deliberate. 

The tab order / next button thing is some property of the edit text. You could probably look at building kevels quest or similar

On June 3, 2022 6:26:04 PM GMT+02:00, Mateusz Konieczny ***@***.***> wrote:
>> FYI you can screen-record with Android Studio:
>
>Ooo, one more nice feature that I missed! Thanks! (I was using Peek which allows to record anything on screen but this would be much better)
>
>> From looking at the gif, do you have a tab order, i.e. that you can tap "next" on the keyboard to get to the next input field?
>
>I have not tried this before - I guess that it would be blue checkmark button? Right now it is not implemented. Are there quest interfaces where it is implemented? Housenumber quest in house name mode is also without this (and I also implemented it, IIRC).
>
>Multilanguage name input is without it, but it is less important there as you need to manually add languages, which places cursor in a new language
>
>BTW, house number quest recently lost white border decoration - is it intentional? Should it be removed also here? Or restored there? 
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/4066#issuecomment-1146152542
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
```diff
@@ -0,0 +1,101 @@
...
```
westnordost, line 79:
```suggestion
    // TODO these vals should be at the top of the file actually
    private val entrancesFilter by lazy { 
        "nodes with entrance ~ main|staircase".toElementFilterExpression()
    }

    private val noEntranceRefFilter by lazy { """
        nodes with
          and noref != yes
          and ref:signed != no
          and !addr:flats
          and !name
          and !ref
    """.toElementFilterExpression() }

    override fun isApplicableTo(element: Element): Boolean? =
        if (!entrancesFilter.matches(element)) false else null

    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
        // note: it does not support multipolygon buildings
        val buildings = mapData.ways.asSequence()
            .filter { buildingFilter.matches(it) }

        val result = mutableListOf<Node>()
        for (building in buildings) {
            val buildingEntrances = building.nodeIds
                .mapNotNull { mapData.getNode(it) }
                .filter { entrancesFilter.matches(it) }

            if (buildingEntrances.count() < 2) continue

            result.addAll(buildingEntrances.filter { noEntranceRefFilter.matches(entrance) })
        }
        return result
    }
```

I understand that `flatMap` may not be easier to understand when one is not thinking in functional. I suggest this, because:
- the filter syntax is easily readable, better readable then code that does the same
- using the filter syntax for such things is consistent with the rest of the code base
- naming: usually we use `result`, not `returned`
- it's also shorter
- less deep nesting IMO improves readability

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
```diff
@@ -0,0 +1,101 @@
...
+import de.westnordost.streetcomplete.quests.building_entrance_reference.ReferenceCode
+import de.westnordost.streetcomplete.quests.building_entrance_reference.ReferenceCodeAndFlatRange
+import de.westnordost.streetcomplete.quests.building_entrance_reference.Unsigned
+
+class AddEntranceReference : OsmElementQuestType<EntranceAnswer> {
+
+    private val buildingFilter by lazy { """
+        ways with
+          building = apartments
+          and ruins != yes and disused != yes
```
westnordost, line 25: Does it make sense to require/exclude either/any of:
- `addr:housenumber` on the apartment building
- `addr:housenumber` on the entrance

?

westnordost (commented):
Better late than never, here's my review

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
matkoniecz: addr:housenumber on the apartment building - no, code on entrance is not replacement in such case for housenumber address

addr:housenumber on the entrance - I am not 100% sure, as while it is not excluding in principle then it would indicate that someone mapped this or relatively rare case in this countries where each entrance has own house number

app/src/main/res/values/strings.xml
```diff
@@ -1473,6 +1473,15 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_tracktype_grade4">Mostly soft</string>
     <string name="quest_tracktype_grade5">Soft</string>
 
+    <string name="quest_entrance_reference">What is signed at this entrance?</string>
+    <string name="quest_entrance_reference_select_flat_range_and_code">Code of entrance and range of flats</string>
+    <string name="quest_entrance_reference_select_flat_range_only">Range of flats</string>
+    <string name="quest_entrance_reference_select_code_only">Code of entrance</string>
+    <string name="quest_entrance_reference_nothing_signed">No sign with code or flat range</string>
+    <string name="quest_entrance_reference_reference_label">Code of entrance:</string>
```
westnordost, line 1481: Sure this and others should not be "entrance code"?

app/src/main/res/layout/quest_entrance_flat_range.xml
```diff
@@ -0,0 +1,90 @@
...
+                >
+
+                <EditText
+                    android:id="@+id/flatRangeStartInput"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:background="@drawable/background_rectangular_sign_blue"
+                    android:imeOptions="actionDone"
+                    android:textColorHint="#2fff"
+                    android:digits="0123456789.,- /IVX"
```
westnordost, line 47: and that does work? Or does there ned to be an "ABC" button to switch the keyboard from numeric to letters?

app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt
```diff
@@ -0,0 +1,126 @@
...
+        // not applicable (no matching address system)
+        // Kosovo (XK) -  https://t.me/osmkosovo/19219
+        // Serbia (RS) - https://t.me/osm_sr/6877
+        // Lithuania (LT) - https://lists.openstreetmap.org/pipermail/talk-lt/2022-June/thread.html#2723 https://github.com/streetcomplete/StreetComplete/issues/3064#issuecomment-1147833339
+        // Slovakia (SK) https://groups.google.com/g/osm_sk/c/IqZz0-iRK5I
+        // Czech Republic (CZ) https://t.me/openstreetmapcz/1242
+        // Croatia https://github.com/streetcomplete/StreetComplete/issues/3064#issuecomment-1157858833
+
+        // was unable to reach communities or get replies from Kazakhstan, Moldova, Bosnia and Herzegovina,
+        // Slovenia, Albania, Montenegro, Georgia (using what was on https://openstreetmap.community/ )
```
stefanb, line 52: Replied regarding Slovenia in forum thread: https://forum.openstreetmap.org/viewtopic.php?pid=865336#p865336

```suggestion
        // Croatia (HR) - https://github.com/streetcomplete/StreetComplete/issues/3064#issuecomment-1157858833
        // Slovenia (SI) - https://forum.openstreetmap.org/viewtopic.php?pid=865336#p865336

        // was unable to reach communities or get replies from Kazakhstan, Moldova, Bosnia and Herzegovina,
        // Albania, Montenegro, Georgia (using what was on https://openstreetmap.community/ )```

File: app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance_reference/AddEntranceReference.kt [hunk 9]
```diff
@@ -0,0 +1,127 @@
...
+        // Kosovo (XK) -  https://t.me/osmkosovo/19219
+        // Serbia (RS) - https://t.me/osm_sr/6877
+        // Lithuania (LT) - https://lists.openstreetmap.org/pipermail/talk-lt/2022-June/thread.html#2723 https://github.com/streetcomplete/StreetComplete/issues/3064#issuecomment-1147833339
+        // Slovakia (SK) https://groups.google.com/g/osm_sk/c/IqZz0-iRK5I
+        // Czech Republic (CZ) https://t.me/openstreetmapcz/1242
+        // Croatia (HR) - https://github.com/streetcomplete/StreetComplete/issues/3064#issuecomment-1157858833
+        // Slovenia (SI) - https://forum.openstreetmap.org/viewtopic.php?pid=865336#p865336
+
+        // was unable to reach communities or get replies from Kazakhstan, Moldova, Bosnia and Herzegovina,
+        // Albania, Montenegro, Georgia (using what was on https://openstreetmap.community/ )```
```
stefanb, line 53: Ooopsie, please fix the error from [my suggestion in comment in f33e672](https://github.com/streetcomplete/StreetComplete/pull/4066/commits/f33e6729ac5139872e9d4246063d8524a75b369b):
```suggestion
        // Albania, Montenegro, Georgia (using what was on https://openstreetmap.community/ )
```

app/src/main/res/values/strings.xml
matkoniecz: What would you prefer?

app/src/main/res/layout/quest_entrance_flat_range.xml
matkoniecz: Good point. It does not work.

screen01

app/src/main/res/values/strings.xml
westnordost: If this about a code that denominates an entrance, then "entrance code" I do prefer.

app/src/main/res/layout/quest_entrance_flat_range.xml
westnordost: So there should be an ABC button that works like the one in the housenumber quest.

matkoniecz:
@westnordost done

westnordost:
Code looks fine, I trust you tested it.

matkoniecz:
Yes, it is tested.

PR #4067, opened 2022-05-28
Don't ask about entrances on building=roof

dbdean:
It's often a bit difficult to answer the entrance question on building=roof, as there isn't often any clearly defined entrance, regardless of whether paths are entering the roofed area or not.


matkoniecz:
It may be worth considering to switch to closed list of valid values instead.

PR #4070, opened 2022-05-30
Improve performance when displaying pins

Helium314:
This PR addresses two related performance issues I encounter way too often (maybe because of my slow phone).

1. The app gets noticeably slower over time, after scrolling the map a lot, because `updatePins` may take several seconds when `quests` holds several thousand quests. And `updatePins` is called every time a quest is solved, plus often when scrolling.
The improvement here is removing all quests not currently visible. I found that above roughly 10000 quests, getting quests from DB again is negligible compared to the time `updatePins` takes. The `retrievedTiles` size limit is added to avoid permanent cleanup in areas with high quest density.
I am not sure how reasonable the limits are for faster phones that have more memory and performance than my S4 mini, but I guess in that case loading from DB is also faster, and the changes will not be noticed.
2. When scrolling the map quickly in areas with many quests, it sometimes takes very long (30s) for the quest pins to appear after scrolling is done. This is partially caused by the amount of pins (already addressed), but also because `updatePins` is called multiple times in parallel, and takes very long to execute.
Since pins for all `quests` are loaded at once anyway (i.e. not incremental), slow loading and fast scrolling may lead to situations where the same list of quest pins is loaded multiple times, massively slowing down the app. My solution for avoiding calling `updatePins` multiple times in parallel may not be the best, but still I found large improvements when testing.
It may be called a second time after finishing because quest list could have changed meanwhile.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
@@ -20,6 +20,7 @@ import de.westnordost.streetcomplete.data.visiblequests.QuestTypeOrderSource
 import de.westnordost.streetcomplete.screens.main.map.components.Pin
 import de.westnordost.streetcomplete.screens.main.map.components.PinsMapComponent
 import de.westnordost.streetcomplete.screens.main.map.tangram.KtMapController
+import de.westnordost.streetcomplete.util.math.contains
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
@@ -126,6 +127,19 @@ class QuestPinsManager(
         // area too big -> skip (performance)
         if (tilesRect.size > 4) {
             return
+        }
+        // clean up if too many quests are loaded
+        if (quests.size > 10000 && retrievedTiles.size > 10) { // should depend on device / heap size / cpu performance
+            synchronized(quests) {
+                // remove all quests not in current tilesRect
+                val bbox = tilesRect.asBoundingBox(TILES_ZOOM)
+                quests.values.removeAll { list -> list.none { bbox.contains(it.position) } }
+            }
+            synchronized(retrievedTiles) {
+                val tiles = tilesRect.asTilePosSequence().toSet()
+                // remove all retrievedTiles not in current tilesRect
+                retrievedTiles.removeAll { !tiles.contains(it) }
+            }
         }
         var tiles: List<TilePos>
         synchronized(retrievedTiles) {
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
@@ -53,6 +53,9 @@ class QuestPinsManager(
 
     private val viewLifecycleScope: CoroutineScope = CoroutineScope(SupervisorJob())
 
+    private var updatePinsIsRunning = false
+    private var updatePinsAgain = false
+
     /** Switch active-ness of quest pins layer */
     var isActive: Boolean = false
         set(value) {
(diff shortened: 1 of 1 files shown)


westnordost:
Hm well, in the layers branch I am working on, I completely changed this class (anyway). The class does not anymore cache all quests ever fetched from db but only those in view. While this might be slower when the view changes quickly between two tiles - forth and back - it solves the issue you are trying to solve with this PR.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
```diff
@@ -127,6 +131,19 @@ class QuestPinsManager(
...
+        if (quests.size > 10000 && retrievedTiles.size > 10) { // should depend on device / heap size / cpu performance
+            synchronized(quests) {
+                // remove all quests not in current tilesRect
+                val bbox = tilesRect.asBoundingBox(TILES_ZOOM)
+                quests.values.removeAll { list -> list.none { bbox.contains(it.position) } }
+            }
+            synchronized(retrievedTiles) {
+                val tiles = tilesRect.asTilePosSequence().toSet()
+                // remove all retrievedTiles not in current tilesRect
+                retrievedTiles.removeAll { !tiles.contains(it) }
```
westnordost, line 144:
```suggestion
                retrievedTiles.removeAll { it !in tiles }
```

westnordost:
So, these improvements will be dropped anyway soon and in the meantime cause merge issues. So not sure if it is worth to have this in for one or two SC versions

Helium314:
Well, it's up to you whether you consider it useful enough to warrant dealing with merge issues...
Personally I'll be using the PR anyway, at least until your changes are merged.

> The class does not anymore cache all quests ever fetched from db but only those in view. While this might be slower when the view changes quickly between two tiles - forth and back - it solves the issue you are trying to solve with this PR.

I just did a short test with reloading quests from DB every time `lastDisplayedRect` changes, and this is indeed not so nice when scroling forth and back over the edge of two zoom 14 tiles with high quest density (but my quick attempt of copying this behavior might be less efficient than what you have).

westnordost:
> zoom 14 tiles

changed it to zoom 16 though IIRC

westnordost:
The rationale of removing the cache (except for the tiles currently in view) was that 
- in normal usage of the app, you don't wildly scroll around across tiles but mostly remain close to where your location marker is
- **IF** caching is done, it should be done one level higher up: In the `OsmQuestController`.


Adding spatial caching (i.e. data is cached in a tile grid) for `OsmQuestController` and `MapDataController` would bring a huge performance boost, I am sure, because
1. reading operations on the database will be somewhat reduced
2. on download/update, `put` operations on map data can effectively return immediately (and the actual persisting into the database be done in a background thread) which means that the scanning for quests can start immediately and doesn't need to wait for the persisting to be done. This will shorten the download somewhat

I have been thinking about implementing that for a while, but never got around to do it as I am busy with other things.

Helium314:
Thanks!
I actually didn't expect to be added, as it's just for a rather short time.

I'm definitely interested in performance, so I'll look into some sort of caching as you described.

And probably also into the uplpoad thing, though for the upload I actually found `MapDataWithEditsSource.rebuildLocalChanges` to cause much more slowdowns than other parts.
This mostly applies when uploading many quests at once. For a few hundred quests to upload, `rebuildLocalChanges` takes some seconds after every single upload (obviously gets faster as the number of changes/uploads goes down).

westnordost:
True, but I think it is not the `rebuildLocalChanges` itself that takes long but the triggered call to the listeners that observe if the map data has been changed (and then thus check for all the quests for those updated elements). I do not think that `rebuildLocalChanges` can be avoided but what could maybe be avoided is to call the listeners at all in the case that original OSM data + local change == original OSM data after upload of the change.

RubenKelevra:
> Alright, I will merge it. Thank you!
> 
> If you are interested in more performance improvements, then a cache for `MapDataController` and `OsmQuestController` would be a grand addition. Another probably rather low hanging fruit would be to speed up the upload x2: Currently, for every edit uploaded, first the current version of the element it refers to is downloaded, then the edit applied to that, then uploaded. Meaning, for every edit, there are two API calls. But in 99% of cases, StreetComplete already got the newest version of that element in the local database, so what it could do is to apply the edit to the local version of the element first, try to upload that, and only if the API returns 409 Conflict, get the newest element from the API and do that again.

Probably also much better for the backend, as most of the time the context can be dropped faster again (not necessary to read the same data twice before updating).

RubenKelevra:
Actually that's a huge improvement even on faster devices. That's the first version of SC which completes drawing the quests first and then the background map.

Thanks! :)

PR #4075, opened 2022-06-03
Ask for summit register more often

Helium314:
Currently this quest is not asked because `route` relations are ignored.
With this PR, the quest is now asked for resurvey and if there is a summit cross (in my experience there is often a register if there is a cross).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
@@ -48,10 +48,10 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
         val hikingRoutes = mapData.relations
             .filter { it.tags["route"] == "hiking" }
             .mapNotNull { mapData.getRelationGeometry(it.id) as? ElementPolylinesGeometry }
-        if (hikingRoutes.isEmpty()) return emptyList()
 
         // yes, this is very inefficient, however, peaks are very rare
         return peaks.filter { peak ->
+            peak.tags["summit:cross"] == "yes" || peak.tags.containsKey("summit:register") ||
             hikingRoutes.any { hikingRoute ->
                 hikingRoute.polylines.any { ways ->
                     peak.position.distanceToArcs(ways) <= 10
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
@@ -52,6 +52,7 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
 
         // yes, this is very inefficient, however, peaks are very rare
         return peaks.filter { peak ->
+            peak.tags["summit:cross"] == "yes" || peak.tags.containsKey("summit:register") ||
             hikingPathsAndRoutes.any { hikingPath ->
                 hikingPath.polylines.any { ways ->
                     peak.position.distanceToArcs(ways) <= 10
@@ -67,7 +68,11 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
    """.toElementFilterExpression() }
 
     override fun isApplicableTo(element: Element): Boolean? =
-        if (!filter.matches(element)) false else null
+        when {
+            !filter.matches(element) -> false
+            element.tags["summit:cross"] == "yes" || element.tags.containsKey("summit:register") -> true
+            else -> null
+        }
 
     override fun createForm() = YesNoQuestAnswerFragment()
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
@@ -61,7 +61,7 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
     private val hikingPathsFilter by lazy { """
         ways with
           highway = path
-          sac_scale ~ mountain_hiking|demanding_mountain_hiking|alpine_hiking|demanding_alpine_hiking|difficult_alpine_hiking
+          and sac_scale ~ mountain_hiking|demanding_mountain_hiking|alpine_hiking|demanding_alpine_hiking|difficult_alpine_hiking
    """.toElementFilterExpression() }
 
     override fun isApplicableTo(element: Element): Boolean? =
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
@@ -45,12 +45,14 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
         val peaks = mapData.nodes.filter { filter.matches(it) }
         if (peaks.isEmpty()) return emptyList()
 
-        val hikingPaths = mapData.ways.filter { hikingPathsFilter.matches(it) }
-            .mapNotNull { mapData.getWayGeometry(it.id) as? ElementPolylinesGeometry }
+        val hikingPathsAndRoutes = mapData.ways.filter { hikingPathsFilter.matches(it) }
+            .mapNotNull { mapData.getWayGeometry(it.id) as? ElementPolylinesGeometry } +
+            mapData.relations.filter {  it.tags["route"] == "hiking" }
+                .mapNotNull { mapData.getRelationGeometry(it.id) as? ElementPolylinesGeometry }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
```diff
@@ -48,10 +48,10 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
         val hikingRoutes = mapData.relations
```
matkoniecz, line 48: this code is now dead and should be also removed

in general `getApplicableElements` can be simplified

main problem is now it will be now asked for peaks far away from any paths... Maybe it should be asked for path close to `highway=path` ways?

Helium314: I didn't want to remove it because it might start working again if at some point (hiking) relations are not ignored any more.

Far away from paths might not be that much of an issue if someone actually carried a cross up there?

matkoniecz: > if at some point (hiking) relations are not ignored any more

unlikely (it caused outrageous performance issues), and in such case old commits still exist

> Far away from paths might not be that much of an issue if someone actually carried a cross up there?

I am pretty sure that some crosses are on summits inaccessible by regular hikers

Helium314: I doubt it's worth the effort filtering out those rare inaccessible crosses, but ok.
Is it sufficient to replace `route=hiking` relations with `highway=path` ways?

matkoniecz: >  Is it sufficient to replace route=hiking relations with highway=path ways?

allowing also `highway=footway` and maybe `highway=track` may make sense

And it is worth testing is it murdering performance or not.

matkoniecz: > I doubt it's worth the effort filtering out those rare inaccessible crosses, but ok.

You can also inspect already mapped crosses and check how often they are inaccessible

Bigger benefit of asking near paths/tracks is that far more peaks that should be eligible will get this quest

Helium314: Oh, I didn't realize this was about all peaks in the `filter`, I was rather narrowly focused onto peaks with cross...
And also private paths/tracks shoulds be filtered out.

westnordost: Jup, this is why I do not think that approach is viable.

This quest is effectively dead, but I hesitate to remove it completely because it lying around doesn't hurt plus maybe one day, route relations can be included again. Then, we do not need to search deep in the git history of this project to make this quest work again.

Helium314: If `summit:cross` is not enough, how about requiring nearby paths to have `sac_scale` tagged? This should not happen in a city (well, maybe exclude `sac_scale=hiking`).

westnordost: Hmm yeah, I guess this would work. Of course it must be an inclusive list, not exclusive list

Helium314: All values in https://wiki.openstreetmap.org/wiki/Key:sac%20scale except `hiking`? Or maybe also exclude `difficult_alpine_hiking` as it includes climbing?

westnordost: I think everything except hiking is fine

app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
```diff
@@ -45,21 +45,25 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
...
                     peak.position.distanceToArcs(ways) <= 10
                 }
             }
         }
     }
 
+    private val hikingPathsFilter by lazy { """
+        ways with
+          highway = path
+          sac_scale ~ mountain_hiking|demanding_mountain_hiking|alpine_hiking|demanding_alpine_hiking|difficult_alpine_hiking
```
FloEdelmann, line 64:
```suggestion
        ways with
          highway = path
          and sac_scale ~ mountain_hiking|demanding_mountain_hiking|alpine_hiking|demanding_alpine_hiking|difficult_alpine_hiking
```

or maybe include `route=hiking` (for future use):

```suggestion
        ways with
          (
            highway = path
            and sac_scale ~ mountain_hiking|demanding_mountain_hiking|alpine_hiking|demanding_alpine_hiking|difficult_alpine_hiking
          ) or route = hiking
```

FloEdelmann (changes requested):
So nodes with `summit:cross = yes` or `summit:register` will still not be included if they are not close to an included way?

app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
matkoniecz: `or route = hiking` should be checked on relation - not on ways

Helium314: > So nodes with summit:cross = yes or summit:register will still not be included if they are not close to an included way?

I removed always asking on nodes with `summit:cross` because the only comment on this seemed to not be in favor of using it:

> I am pretty sure that some crosses are on summits inaccessible by regular hikers

Helium314: So the quest should be asked if either a hiking route or a path with appropriate `sac_scale` is nearby?

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
```diff
@@ -45,21 +45,27 @@ class AddSummitRegister : OsmElementQuestType<Boolean> {
...
+        val hikingPathsAndRoutes = mapData.ways.filter { hikingPathsFilter.matches(it) }
+            .mapNotNull { mapData.getWayGeometry(it.id) as? ElementPolylinesGeometry } +
+            mapData.relations.filter {  it.tags["route"] == "hiking" }
+                .mapNotNull { mapData.getRelationGeometry(it.id) as? ElementPolylinesGeometry }
 
         // yes, this is very inefficient, however, peaks are very rare
         return peaks.filter { peak ->
-            hikingRoutes.any { hikingRoute ->
-                hikingRoute.polylines.any { ways ->
+            hikingPathsAndRoutes.any { hikingPath ->
```
FloEdelmann, line 55: > I removed always asking on nodes with summit:cross because the only comment on this seemed to not be in favor of using it:
>
>> I am pretty sure that some crosses are on summits inaccessible by regular hikers

(by @Helium314 in https://github.com/streetcomplete/StreetComplete/pull/4075#discussion_r892125037 and @matkoniecz in https://github.com/streetcomplete/StreetComplete/pull/4075#discussion_r888950833)

Is it an issue though? If someone searches for quests in an inaccessible area, they either know how to walk or climb there or they will ignore the quest (or deliberately answer from a distance…).

I think that asking for all nodes with `summit:cross = yes` makes sense and will make the quest more useful again. Also, if `summit:register` was already tagged, resurveying should always be enabled IMHO, regardless whether a way/route is nearby.

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt
Helium314: I also think it makes sense, but @matkoniecz didn't like it.
I don't need to deal with people opening you-want-the-users-to-get-in-dangerous-situations issues, but I understand the desire to avoid them.

matkoniecz: I am not blocking it strongly - if you are sure that it is a good idea (especially if checking nearby sac_scale paths is not going to work or will exclude too many) then add it.

westnordost: reading vertically through this, Flo's point makes sense to me

FloEdelmann (approved):
Looks good to me :)

PR #4082, opened 2022-06-06
Fix trace uploads

goldbattle:
This fixes two issues:
1. On second notes, the GPX tracks are still being appended. I thought I tested this originally, but it seems to have slipped by, sorry. Now I explicitly send that the notes should attach the current tracks with the `hasGpxAttached` argument variable used to visualize the traces are going to be attached to the note. (you can see this issue [here](https://www.openstreetmap.org/traces/tag/streetcomplete))
2. Names with spaces are not handled properly. I encode the username so that the url is clickable from the note ([an example](https://www.openstreetmap.org/note/3211907)).


westnordost (approved):
Looks good, thank you!

Issue #4071, opened 2022-05-31
Map style not updated


Helium314:
There were a few updates to https://github.com/streetcomplete/streetcomplete-mapstyle over the past year, but the last commit that actually is used in StreetComplete is https://github.com/streetcomplete/streetcomplete-mapstyle/commit/6755a442956dffc3f3120a8058c83c1049f7dddb / https://github.com/streetcomplete/StreetComplete/commit/9778c814d7c23370a40bf254cbaf0733a6c02481 (more than a year old).

Is there a reason for not merging later updates?

matkoniecz:
Probably noone did it - I think that it would be fine to open PR doing this.

PR #4100, opened 2022-06-11
Add `updateMapStyle` Gradle task

FloEdelmann:
Automate the task that caused #4071.

CC @Helium314 @ENT8R

### DIFF
--- a/app/src/main/assets/map_theme/jawg/global.yaml
+++ b/app/src/main/assets/map_theme/jawg/global.yaml
@@ -1,6 +1,7 @@
 global:
     api_key: false # intended to be overwritten by StreetComplete
     language: false # intended to be overwritten by StreetComplete
+    language_script: Latn # intended to be overwritten by StreetComplete
     text_font_family: 'Roboto'
     text_stroke_color: 'white'
     text_water_color: '#fff'
--- a/app/src/main/assets/map_theme/jawg/global.yaml
+++ b/app/src/main/assets/map_theme/jawg/global.yaml
@@ -1,7 +1,6 @@
 global:
     api_key: false # intended to be overwritten by StreetComplete
     language: false # intended to be overwritten by StreetComplete
-    language_script: Latn # intended to be overwritten by StreetComplete
     text_font_family: 'Roboto'
     text_stroke_color: 'white'
     text_water_color: '#fff'
@@ -149,6 +148,5 @@ sources:
 textures:
     oneway-arrow:
         url: images/oneway_arrow@2x.png
-        # url: https://raw.githubusercontent.com/ENT8R/streetcomplete-mapstyle/master/images/oneway_arrow@2x.png
         filtering: mipmap
         density: 5
--- a/buildSrc/src/main/java/UpdateMapStyleTask.kt
+++ b/buildSrc/src/main/java/UpdateMapStyleTask.kt
@@ -67,14 +67,12 @@ private fun String.replaceOnlineWithLocalSections(): String {
     // uncomment local sections
     var shouldUncomment = false
     lines.forEachIndexed { index, line ->
-        if (shouldUncomment) {
-            if (line.trim().startsWith("#")) {
-                lines[index] = line.replaceFirst("#", "")
-            } else {
-                shouldUncomment = false
-            }
-        } else if (localStartLineIndices.contains(index)) {
+        if (localStartLineIndices.contains(index)) {
             shouldUncomment = true
+        } else if (shouldUncomment && line.trim().startsWith("#")) {
+            lines[index] = line.replaceFirst("#", "")
+        } else {
+            shouldUncomment = false
         }
     }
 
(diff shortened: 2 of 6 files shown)


app/src/main/assets/map_theme/jawg/global.yaml
```diff
@@ -1,7 +1,6 @@
 global:
     api_key: false # intended to be overwritten by StreetComplete
     language: false # intended to be overwritten by StreetComplete
-    language_script: Latn # intended to be overwritten by StreetComplete
```
FloEdelmann, line 4: Is this change intended?

`global.language_script` is still referenced in this file in line 80 and in

https://github.com/streetcomplete/StreetComplete/blob/18189f123416d9a0b332e50afab58b857452b416/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/SceneMapComponent.kt#L69

westnordost: > Is this change intended?

hm, no

buildSrc/src/main/java/UpdateMapStyleTask.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import java.net.URL
+
+open class UpdateMapStyleTask : DefaultTask() {
+
+    @get:Input var targetDir: String? = null
+
+    @TaskAction fun run() {
+        println("hello")
```
westnordost, line 14: hello

buildSrc/src/main/java/UpdateMapStyleTask.kt
```diff
@@ -0,0 +1,85 @@
...
+private fun String.normalizeLineEndings() = this.replace("\r\n", "\n")
+
+private fun String.replaceOnlineWithLocalSections(): String {
+    val lines = this.split("\n").toMutableList()
+
+    val onlineStartLineIndices = lines.indices.filter { lines[it].trim().startsWith("# for online testing") }
+    val localStartLineIndices = lines.indices.filter { lines[it].trim().startsWith("# for local testing") }
+
+    val onlineRanges = onlineStartLineIndices.zip(localStartLineIndices)
+    onlineRanges.forEach { (start, end) -> assert(start < end) }
```
westnordost, line 65: `assert` is kind of deprecated in Java. Not really (I think) but the behavior is not what one would expect. IIRC it can be configured as a JVM flag what should be the behavior of assert and by default it is a no-op in release builds. In any case, assert seems to be meant as a debugging tool rather than to enforce contracts. That's what I learnt ~10 years ago at least, feel free to disabuse me.

In any case, in Kotlin there are `check` (for state checks) and `require` (for ensuring contracts).

buildSrc/src/main/java/UpdateMapStyleTask.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    @get:Input var targetDir: String? = null
+
+    @TaskAction fun run() {
+        println("hello")
+
+        val targetDir = targetDir?.let { File(it) } ?: return
+        assert(targetDir.exists())
+
+        val githubDirectoryListingUrl = URL("https://api.github.com/repos/streetcomplete/streetcomplete-mapstyle/contents?ref=jawg")
```
westnordost, line 19: pass branch name as parameter instead?

westnordost (approved):
I think this can be merged, thank you for that work! 

In case you tasted blood and are interested in adding more automatism, there is one thing that would be most appreciated: Generating the taginfo JSON. You probably remember that it failed before.

app/src/main/assets/map_theme/jawg/global.yaml
FloEdelmann: So this line is always manually added in the StreetComplete repo after copying the map style from the mapstyle repo?

buildSrc/src/main/java/UpdateMapStyleTask.kt
FloEdelmann: oops

buildSrc/src/main/java/UpdateMapStyleTask.kt
FloEdelmann: Ah, I always forget that. I'll change it to `require` then.

app/src/main/assets/map_theme/jawg/global.yaml
westnordost: I think so

westnordost (approved):

app/src/main/assets/map_theme/jawg/global.yaml
FloEdelmann: With https://github.com/streetcomplete/streetcomplete-mapstyle/pull/129, this is now automated, too (commit [`0fdeda5`](https://github.com/streetcomplete/StreetComplete/pull/4100/commits/0fdeda5fa3aa3a2ae19a510e91ec15ff29591a67) was created by the task).

PR #4101, opened 2022-06-12
Use country metadata to control whether quests are enabled

FloEdelmann:
See https://github.com/streetcomplete/StreetComplete/discussions/4096.

Depends on https://github.com/streetcomplete/countrymetadata/pull/6.


app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt
```diff
@@ -151,7 +153,9 @@ class OsmQuestController internal constructor(
             scope.async {
                 val questsForType = ArrayList<OsmQuest>()
                 val questTypeName = questType::class.simpleName!!
-                if (!countryBoundaries.intersects(bbox, questType.enabledInCountries)) {
+                val countryCodes = countryBoundaries.getIntersectingIds(bbox)
+
+                if (!countryCodes.any { questType.isEnabled(countryInfos.get(listOf(it))) }) {
```
FloEdelmann, line 158: The corresponding test (["updates quests on map data listener replace for bbox"](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/test/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestControllerTest.kt#L299)) is now failing. I'm not sure whether the test needs to be adapted or I accidentally changed the logic here.

westnordost:
Hmm, I don't really like this.

Country metadata is so far composed of (afair/mostly) hard factual information, such as the start of the workday, the appearance of a cycle lane etc. 
Now, a quest being disabled in StreetComplete is not necessarily based on such facts but a matter of consideration:
- does the given property (cycle track, vibrating button, ...) appear in such numbers that it is worthwhile to ask surveyors about it (💤 no spam)?
- do the mapping practices or (local) tagging guidelines influence the quest in a way that it can't fulfill all the quest guidelines? (e.g. housenumber quest: housenumbers tagged not on buildings; all housenumbers imported for third party etc..)

Trying to shoehorn these reasons into the countrymetadata does not appear so wise because it is both less flexible and means more effort for contributor of a quest.

Additionally, in regards to adding such fuzzy stuff as `usuallyHasXYZ` to countrymetadata: From the point of view of maintenance of this repository, this sounds like a precedent for a nightmare to maintain. IIRC the readme for countrymetadata requests to always give the source (in a comment behind that line) for each contribution made, because that data should be based on hard facts. If there is data that is so subjective that it is a matter of consideration... well, that's to things that are difficult to marry.

FloEdelmann:
I've moved your comment to the discussion https://github.com/streetcomplete/StreetComplete/discussions/4096#discussioncomment-2933107. Let's keep the high-level conversation there.

FloEdelmann:
Closed because of https://github.com/streetcomplete/StreetComplete/discussions/4096#discussioncomment-2940702.

PR #2115, opened 2020-09-27
[WIP] What kind of memorial is this?

kmpoppe:
Closes #1572.

For the first throw, I've selected the entries, that have a wiki page, that are in the most used on taginfo AND that I could find pictures for that are either Public Domain or CC0.

Soruces are as follows:
`memorial=statue`:
https://commons.wikimedia.org/wiki/File:George_Washington_statue_in_the_Boston_Public_Garden_-_DSC08196.JPG
`memorial=bust`:
https://commons.wikimedia.org/wiki/File:Aristotelis_Valaoritis_bust_National_Garden_Athens,_Greece.jpg
`memorial=plaque`:
https://commons.wikimedia.org/wiki/File:Bleaklow_Plaque.jpg
`memorial=stolperstein`:
https://flic.kr/p/2jLQ3Xc
`memorial=stone`:
https://commons.wikimedia.org/wiki/File:Memorial_stone_of_battle_of_Liljendal.18080224.ojp.JPG
`memorial=war_memorial`:
https://commons.wikimedia.org/wiki/File:Women_of_World_War_II.jpg

Please comment on the pictures if anyone wants me to change something.

Reasons for leaving out others:
* `memorial=blue_plaque`- mainly in the UK, only around 370 uses. There might be a nice way of filtering whether the quest is being solved at a place in the UK and then display that option, but we'd still need to figure the licensing issue for a picture. There seems to be something special about them, haven't figured it out completely.
* `memorial=sculpture` and `memorial=cross` - no wiki pages, less than 1k uses
* `memorial=stele` and `memorial=obelisk` - quite frankly, I was unable to find pictures that made a distinction clear on first look. My (not so much-)favourite was a (not PD or CC0) picture in the German Wikipedia article about steles, that showed three stone blocks shaped like a small obelisk... not helpful.
* `memorial=ghost_bike` - 160 uses, heavily used only in Germany and the Netherlands, rest of the world: USA 21, South America: 3, EU/Turkey w/o GER and NLD: 25 - Yes, it warrants a wiki-page due to the importance of the subject, but I don't think it's that useful in SC.

Bonus feature that I'd love to do (@westnordost, can you advise if possible): If the user selects `war_memorial` ask them what conflict (`memorial:conflict`) it's for.

Comments welcome.

PR #4102, opened 2022-06-12
Complete memorial quest

matkoniecz:
/home/mateusz/Documents/install_moje/rescale-for-android
/home/mateusz/Documents/tmp
https://commons.wikimedia.org/wiki/Category:Khatchkars_in_Shirak_province

finish of @kmpoppe work, with help of others (@naposm @FloEdelmann at least), from #2115

There is asking about a bit irrelevant material of steles - it is present because it allows to show multiple images of them and there is no real chance to find one representing their variety.

It is possible that I was overly enthusiastic about khachkars, but they are not limited to a single country, are distinctive and tag is documented. Still, maybe I should remove them.

screen01


app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/AddMemorialType.kt
```diff
@@ -0,0 +1,45 @@
...
+    override val wikiLink = "Key:memorial"
+    override val icon = R.drawable.ic_quest_memorial
+    override val questTypeAchievements = listOf(QuestTypeAchievement.CITIZEN)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_memorialType_title
+
+    override fun createForm() = AddMemorialTypeForm()
+
+    override fun applyAnswerTo(answer: String, tags: Tags, timestampEdited: Long) {
+        when(answer) {
```
FloEdelmann, line 26: Off-topic and nitpick, but I noticed you often forget to add a space after `if` and `when`. Have you tried the Ktlint plugin for Android Studio (linked in the [contributing guidelines](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md?plain=1#L107))? That may help familiarizing with the style.

app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/AddMemorialType.kt
```diff
@@ -0,0 +1,45 @@
+package de.westnordost.streetcomplete.quests.memorial_type
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.osm.osmquests.Tags
+import de.westnordost.streetcomplete.data.user.achievements.QuestTypeAchievement
+
+class AddMemorialType : OsmFilterQuestType<String>() {
```
FloEdelmann, line 8: Please create an enum class and/or sealed interface for the answer. Maybe something like this:

```kt
sealed interface MemorialTypeAnswer

enum class MemorialType(val osmValue: String) : MemorialTypeAnswer {
    STATUE("statue"),
    BUST("bust"),
    PLAQUE("plaque"),
    WAR_MEMORIAL("war_memorial"),
    STONE("stone"),
    OBELISK("obelisk"),
}

enum class Stele(val osmMaterialValue: String) : MemorialTypeAnswer {
    WOODEN_STELE("wood"),
    STONE_STELE("stone"),
    KHACHKAR_STELE("stone"),
}
```

Then `applyAnswerTo` can be simplified to:

```kt
when (answer) {
    is MemorialType -> {
        tags["memorial"] = answer.osmValue
    }
    is Stele -> {
        tags["memorial"] = "stele"
        tags["material"] = answer.osmMaterialValue
    }
}

if (answer == KHACHKAR_STELE) {
    tags["stele"] = "khachkar"
}
```

app/src/main/res/authors.txt
```diff
@@ -84,6 +84,16 @@ kerb_height_flush_original.jpg CC0               Mateusz Konieczny (...)
 kerb_height_lowered_ramp.jpg   CC0               https://commons.wikimedia.org/wiki/File:Pram_Ramp.jpg (Michael3)
 kerb_height_no.jpg             Public Domain     https://commons.wikimedia.org/wiki/File:Ground_frost_damages.JPG (SeppVei)
 
+memorial_type_statue.jpg       CC-BY-SA 4.0      Rabax63 https://commons.wikimedia.org/wiki/File:Garibaldi_Monument.jpg
+memorial_type_bust.jpg         CC0               Jebulon https://commons.wikimedia.org/wiki/File:Aristotelis_Valaoritis_bust_National_Garden_Athens,_Greece.jpg
+memorial_type_plaque.jpg       Public Domain     Geotek https://commons.wikimedia.org/wiki/File:Bleaklow_Plaque.jpg
```
FloEdelmann, line 89: Maybe contrast can be tweaked a bit for `memorial_type_plaque`?

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/AddMemorialType.kt
matkoniecz: > Have you tried the Ktlint plugin for Android Studio (linked in the [contributing guidelines](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md?plain=1#L107))?

no, I installed it now

app/src/main/res/authors.txt
matkoniecz: https://github.com/streetcomplete/StreetComplete/pull/4102/commits/f5a9af05b214c11a5888d97f34b80cf5f64e1eb9

matkoniecz: screen05

FloEdelmann (approved):
Looks good to me :)

I'm not sure if the Khachkar stele should really be included, but I don't have a strong opinion on that.

matkoniecz:
@ZeLonewolf 

Should it be "slab as memorial" or "slab as a memorial"?

ZeLonewolf:
Memorial stone slab is probably the least awkward

matkoniecz:
So it would be "Stone stele (memorial slab)"?

Or maybe "Memorial stone slab (stele)"?

Or "Stone stele (memorial stone slab)"?

ZeLonewolf:
First option sounds the nicest to me.

kmpoppe:
Thank you all for your amazing work and for bringing this over the finish line 🥲

westnordost:
Khachkar seems to be oddly specific. Looks like an elaborate stone stele for me. Why is this not a subtag?

Could there not be hundreds of additional choices for different types of stone steles, different types of stones, different types of ... you get the idea ... if we allowed such differenciation on the top-level?

FloEdelmann:
It is a subtag: `memorial = stele`, `stele = khachkar`

app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/MemorialType.kt
```diff
@@ -0,0 +1,36 @@
...
+        }
+        is Stele -> {
+            tags["memorial"] = "stele"
+            tags["material"] = this.osmMaterialValue
+        }
+    }
+
+    if (this == Stele.KHACHKAR_STELE) {
+        tags["stele"] = "khachkar"
+    }
```
FloEdelmann, line 35: Maybe inlining the condition makes it more understandable:

```suggestion
        is Stele -> {
            tags["memorial"] = "stele"
            tags["material"] = this.osmMaterialValue

            if (this == Stele.KHACHKAR_STELE) {
                tags["stele"] = "khachkar"
            }
        }
    }
```

westnordost: Ah I see. Then I see no reason to include that in this quest. Unless there are stone steles, and khachkar are the only sub-type of stone steles that are defin(able).

matkoniecz: > khachkar are the only sub-type of stone steles that are defin(able)

right now - the only defined

I have no ideas for additional ones (except that gravestones are technically subtype of steles)

matkoniecz: > Could there not be hundreds of additional choices for different types of stone steles, different types of stones, different types of ... you get the idea ... if we allowed such differenciation on the top-level?

there could be - but right now khachkars are only known defined subtype

app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/MemorialType.kt
```diff
@@ -0,0 +1,36 @@
...
+}
```
peternewman, line 36: This would be better done like the stile quest (and others) which will make finishing/redoing my taginfo PR much easier:
https://github.com/streetcomplete/StreetComplete/blob/6aa3fb16b18f74955ce79a19b201d5bb60707616/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/StileTypeAnswer.kt#L11-L16

```suggestion
enum class MemorialType(val osmValue: String, val osmMaterialValue: String? = null, val osmSteleValue: String? = null) : MemorialTypeAnswer {
    STATUE("statue"),
    BUST("bust"),
    PLAQUE("plaque"),
    WAR_MEMORIAL("war_memorial"),
    STONE("stone"),
    OBELISK("obelisk"),
    WOODEN_STELE("stele", "wood"),
    STONE_STELE("stele", "stone"),
    KHACHKAR_STELE("stele", "stone", "khachkar"),
}

fun MemorialTypeAnswer.applyTo(tags: Tags) {
    tags["memorial"] = this.osmValue
    if (this.osmMaterialValue != null) {
        tags["material"] = this.osmMaterialValue
    }
    if (this.osmSteleValue != null) {
        tags["stele"] = this.osmSteleValue
    }
}
```

Unless we can already (or maybe we should add) a way to add a tag if the value isn't null?

peternewman (changes requested):
See if my review actually posts this time...

app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/MemorialType.kt
westnordost: I see. Well, I am trying to get to some sort of mutual decision and not "what I say goes".

From my point of view, the following reasons speak against it:
- ⚛️ Atomicy: A quest should ask the least possible (=taggable) detail
- 🪱 Can of worms: If such a specific sub-category of stone steles is included, what is the argument that can be used to not also include all kinds of further sub-categories (once theere is an established tagging for that). E.g. (mock/decorative) totem poles, cantabrian stelae, kurgan stelae, obelisks, stećak or those found in Copán and other places in central America (they are also very distinguishable from steles in Europe) etc.

The one reason that seems to speak for it is:
- the `stele=*` subtag is not exhaustive in the same way `service=*` for `highway=service` is not exhaustive. Hence, if it is not tagged now, it will probably never be taggable with StreetComplete

In my point of view, the reasons against it somewhat outweigh the reason for it. You don't agree?

matkoniecz: Yes, sadly it likely should be removed :(

westnordost:
One more comment. From the last PR. What do you think of this?: 

https://github.com/streetcomplete/StreetComplete/pull/2115#issuecomment-702124496

matkoniecz:
I guess that adding https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dsculpture makes sense.

How to name it to make clear that it is preferable to use other values? Maybe put it into "can't say" menu option?

westnordost:
Just put it at the end? (And/or) use the wording "(Other) Sculpture"?

matkoniecz:
> Just put it at the end?

and disable image reordering on selection?

westnordost:
Hmm I don't know, I don't see this much as a problem. Maybe because if I see a stele, I would not describe it as some form of sculpture. But even if, the other options are shown right next to it. Which user would select "sculpture" when he knows that "stele" etc can also be selected? He saw that option at least once.

app/src/main/java/de/westnordost/streetcomplete/quests/memorial_type/MemorialType.kt
peternewman: I see this was resolved with no changes or comments, was there a reason behind that?

Even if the Khachkar Stele is going to be dropped, the principle can be easily adapted for the others.

ZeLonewolf:
For the record, I've never heard the word "stele" before.  Though I accept that this is my own ignorance :)

matkoniecz:
> Which user would select "sculpture" when he knows that "stele" etc can also be selected?

ones confused by word "stele" and selecting familiar "sculpture" due to not being aware that "sculpture" is less specific.

matkoniecz:
screen01

mnalis:
Besides also personally never heard of "stele", what is the **_actual difference_** between stone [stele](https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dstele) and [obelisk](https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dobelisk)? Both seem to me to be vertical pillars/slabs made of stone used to mark something (some past event, human or place)?   

Even after looking at the (quite terse) OSM wiki I'm none the wiser. Wording of `stele` seems to indicate it is just like gravestone but which must not be placed at the grave? Wording of `obelisk` just mentions what it is physically, but wikipedia seems to indicate its top ends "pyramid-like shape". Is that the main distinguishing point? If so, it is unfortunate that `stele` picture in SC ends in "cone-like shape", which look very much like "weathered-down pyramid-like shape" 


**I'm worried that average SC user (not even familiar with the OSM wiki, much less with stonework arts) might be even more confused**. It would be good if it could somehow clarify it in SC text shown, and then explain the differences in more details on the wiki?

Example: If I see a stone slab related to fallen fighters in WW2 (which we happen to have tons of), I'd be torn between:
- war memorial
- stone stele
- obelisk

In that case I'd opt for war memorial as it seems most specific, even if it is also there to commemorate dead people (`stele`) and is a vertical slab of stone (`obelisk`). But in other cases (stone slab for some known figure who died not in war) - is it a `stele` or `obelisk` (as I marked it before, having never heard of stele, and [wikipedia agreeing](https://en.wikipedia.org/wiki/Obelisk#As_commemorative_monuments))?

matkoniecz:
> explain the differences in more details on the wiki

edited wiki a bit

my position in general is that main value here is from answering "is it just a plaque?"

matkoniecz:
>  Both seem to me to be vertical pillars/slabs made of stone used to mark something

steles are slabs (flat) while obelisks are pillars in a specific shape

westnordost (approved):

matkoniecz:
>  It would be good if it could somehow clarify it in SC text shown, and then explain the differences in more details on the wiki?

If anyone has specific improvements ideas - feel free to edit wiki or make PR or mention a better description for SC in a new issue (or here, until it is merged).

Right now I have no ideas for improvements and I think that it is not bad enough to block this quest.

> I'd be torn between:    war memorial (...)

See https://github.com/streetcomplete/StreetComplete/pull/2115#issuecomment-1152870691 https://wiki.openstreetmap.org/wiki/Talk:Key:memorial#war_memorial_statue%2Fstele%2Fstone - somehow there is no clear agreement that it is a problem so I decided to not make proposal for `war_memorial=yes` and deprecation of `memorial=war_memorial`

mnalis:
I'd suggest more different picture for `stele`, like https://en.wikipedia.org/wiki/Stele#/media/File:Stele_of_Arniadas.jpg or https://en.wikipedia.org/wiki/Stele#/media/File:Hilarri-4symb.jpg or  (or some other from https://en.wikipedia.org/wiki/Stele -- important is so it not looking like an obelisk - i.e. **clearly** showing that is wider, flat surface, and not ending in pointy tip)

(Update: Looking at the full-scale image of current SC picture at https://upload.wikimedia.org/wikipedia/commons/1/1b/Kerameikos%2C_Ancient_Graveyard%2C_Athens%2C_Greece_%284451597911%29.jpg it is also true, but when resized to part of mobile phone screen those distinctions are lost (e.g. even when paying extra attention at posted SC screenshot I did not see it is in fact flat, and the top looked pointy like in obelisk. One of the two suggested pictures would hopefully be clearer)

app/src/main/res/values/strings.xml
```diff
@@ -1016,6 +1016,17 @@ If there are no signs along the whole street which apply for the highlighted sec
...
 
+    <string name="quest_memorialType_title">What kind of memorial is this?</string>
+    <string name="quest_memorialType_statue">Statue</string>
+    <string name="quest_memorialType_bust">Bust</string>
+    <string name="quest_memorialType_plaque">Plaque</string>
+    <string name="quest_memorialType_war_memorial">War memorial</string>
+    <string name="quest_memorialType_stone">Stone</string>
+    <string name="quest_memorialType_obelisk">Obelisk</string>
+    <string name="quest_memorialType_stele_wooden">Wooden stele (memorial slab)</string>
+    <string name="quest_memorialType_stele_stone">Stone stele (memorial slab)</string>
```
mnalis, line 1027:
```suggestion
    <string name="quest_memorialType_obelisk">Obelisk (pointy pillar)</string>
    <string name="quest_memorialType_stele_wooden">Wooden stele (flat slab)</string>
    <string name="quest_memorialType_stele_stone">Stone stele (flat slab)</string>
```

to distinguish obelisk/pillar. "memorial" word can be removed, as the whole quest is about memorials.

matkoniecz: I \*think\* that for obelisks being four sided is also important, pointed round column is not an obelisk.

matkoniecz: maybe `memorial slab, flat` would be better to avoid tautology?

mnalis: `memorial slab, flat` seems fine to me, I was trying to shorten it down. 

As for obelisks being 4 sided, it definitely seems to be most common case by vast majority (`4-sided pointy pillar`?), but [wikipedia seems to list other shapes too as Obelisks](https://en.wikipedia.org/wiki/List_of_modern_obelisks) (e.g. [Killiney Hill](https://en.wikipedia.org/wiki/Killiney_Hill) Obelisk, [Heroes Monument](https://en.wikipedia.org/wiki/Heroes_Monument), [Monument to the abolition of slavery](https://en.wikipedia.org/wiki/Monumento_a_la_abolici%C3%B3n_de_la_esclavitud), [Bayonet-Obelisk of the War Memorial](https://upload.wikimedia.org/wikipedia/commons/thumb/b/b5/Brest_Brest_Fortress_9085_2150.jpg/100px-Brest_Brest_Fortress_9085_2150.jpg), [Colonna Mediterranea](https://en.wikipedia.org/wiki/Colonna_Mediterranea) etc.)

mnalis:
> Again, if you have something specific then I would be happy to use it

Here is quick mockup (sorry about quality, I'm not an artist) with https://en.wikipedia.org/wiki/Stele#/media/File:Rodney02.JPG (as additional picture in the middle between `Bust` and `War memorial`). 
Does not seem to look too bad to me, so perhaps something like that could be used?

sc_mockup

westnordost:
@mnalis not good, it is not clear how thick it is. Anyway, if you find something better, you can create a separate PR later.

mnalis:
No problem, I do not mean for those suggestions to block/delay this PR. I can alternatively open new one after this one is merged (I do not think I can open new one until this one is in, though?)
Let me know if one of those three replacement pictures in mockup looks better.

sc_stele_mockup_multi1

matkoniecz:
>  Let me know if one of those three replacement pictures in mockup looks better.

2 looks like some 4 cm high miniature

1 has this pole (flagpole? BTW, how you mark `historic=memorial` which is a flag?) in background, but really nicely shows as flat plate

3 is nice, though version without that red element would be nicer

If you want to replace image then 3 seems best (remember to update authors file)

> (I do not think I can open new one until this one is in, though?)

You can (could) - but you would need to open it in https://github.com/matkoniecz/Zazolc selecting complete-memorial-quest as a target branch

screen01

app/src/main/res/values/strings.xml
matkoniecz: For 4-sided: I opened https://en.wikipedia.org/wiki/Talk:Obelisk#Four_sided%3F as I am not 100% sure

matkoniecz:
@westnordost It may be too late for that but I would also list @mnalis in changeleog visible in app (though maybe it is too late for that and app got published?)

matkoniecz:
Done in https://github.com/streetcomplete/StreetComplete/commit/94dcc097867cf3ff8743b02de9d160d780353baa

PR #4105, opened 2022-06-13
handle surface vs tracktype conflicts

### LINKED ISSUE #662: Check and correct inconsistent surface attributes
The OSM tags "surface" and "tracktype" are often used together in order to describe a way tagged as "highway=track". However, sometimes the attributes are inconsistent with each other, e.g. "surface=asphalt, tracktype=grade2" or "surface=gravel, tracktype=grade1". Many maps (e.g. OsmAnd and OpenMTBMap/Velomap for Garmin devices) render paved ways differently which is very useful for road cyclists.

Ideas:

- If a surface quest is solved, the app could examine, if "tracktype" is set. If it's not, set it accordingly. If this attribute is already set, check if it's correct and in case it's not, correct it (maybe show a warning?)

- Discover these kinds of discrepancies and show them as normal "surface" quest as if the attributes aren't already set. If the quest has been solved by a user, proceed as stated above.

The cycling cat

matkoniecz:
ask for confirmation in case where surface mismatches tracktype
resurvey in cases where either surface or tracktype is definitely wrong
fixes #662 #4063

if accepted in general, I will further refactor it to avoid specifying surface/tracktype relation twice, in two separate places

see also https://josm.openstreetmap.de/ticket/22102 - especially https://josm.openstreetmap.de/ticket/22102#comment:7 that would mirror code here and may have feedback about strength of checks


app/src/main/java/de/westnordost/streetcomplete/osm/Surface.kt
```diff
@@ -1,12 +1,36 @@
...
-    "concrete", "concrete:lanes", "concrete:plates", "paving_stones",
+    "concrete", "concrete:plates", "paving_stones",
     "metal", "wood", "unhewn_cobblestone"
 )
+
+val ANYTHING_PAVED = ANYTHING_FULLY_PAVED + setOf(
+    "concrete:lanes"
+)
+
+fun isSurfaceAndTractypeMismatching(surface: String, tracktype: String): Boolean {
```
FloEdelmann, line 19: Typo: missing `k`
Also `track type` as camelCase should be `trackType` in parameter name and function name (and other places).

Partly fixed in my suggestion in next comment.

app/src/main/java/de/westnordost/streetcomplete/osm/Surface.kt
```diff
@@ -1,12 +1,36 @@
...
+            return true
+        }
+    }
+    if (tracktype == "grade3" || tracktype == "grade4" || tracktype == "grade5") {
+        if (ANYTHING_FULLY_PAVED.contains(surface)) {
+            return true
+        }
+    }
+    return false
+}
```
FloEdelmann, line 36:
```suggestion
fun isSurfaceAndTrackTypeMismatching(surface: String, trackType: String): Boolean =
    (trackType == "grade1" && ANYTHING_UNPAVED.contains(surface))
    || (trackType == "grade2" && SOFT_SURFACES.contains(surface))
    || ((trackType == "grade3" || trackType == "grade4" || trackType == "grade5")
        && ANYTHING_FULLY_PAVED.contains(surface))
}
```

app/src/main/java/de/westnordost/streetcomplete/osm/Surface.kt
matkoniecz: > camelCase should be trackType in parameter name and function name (and other places) – even though tracktype is the tag used in OSM.

not entirely sure, I treat "tracktype" as a word/code. Is "track type" even a thing in actual English? I thought that it is osmism not matching actual English terms.

matkoniecz:
@FloEdelmann I partially followed your suggestions in https://github.com/streetcomplete/StreetComplete/pull/4105/commits/5be361c8539c12564b404761b1eec67f3d95c29a

matkoniecz:
draft per

> if accepted in general, I will further refactor it to avoid specifying surface/tracktype relation twice, in two separate places

app/src/main/java/de/westnordost/streetcomplete/osm/Surface.kt
```diff
@@ -1,12 +1,30 @@
...
+fun isSurfaceAndTracktypeMismatching(surface: String, tracktype: String): Boolean {
+    if (tracktype == "grade1" && ANYTHING_UNPAVED.contains(surface)) {
+        return true
+    }
+    if (tracktype == "grade2" && SOFT_SURFACES.contains(surface)) {
+        return true
+    }
+    if ((tracktype == "grade3" || tracktype == "grade4" || tracktype == "grade5") && ANYTHING_FULLY_PAVED.contains(surface)) {
+        return true
+    }
```
FloEdelmann, line 28: It should be possible to use `in` instead of `.contains()` because they are `Set`s:

```suggestion
    if (tracktype == "grade1" && surface in ANYTHING_UNPAVED) {
        return true
    }
    if (tracktype == "grade2" && surface in SOFT_SURFACES) {
        return true
    }
    if ((tracktype == "grade3" || tracktype == "grade4" || tracktype == "grade5") && surface in ANYTHING_FULLY_PAVED) {
        return true
    }
```

app/src/main/java/de/westnordost/streetcomplete/osm/Surface.kt
```diff
@@ -1,12 +1,30 @@
...
+)
+
+fun isSurfaceAndTracktypeMismatching(surface: String, tracktype: String): Boolean {
+    if (tracktype == "grade1" && ANYTHING_UNPAVED.contains(surface)) {
+        return true
+    }
+    if (tracktype == "grade2" && SOFT_SURFACES.contains(surface)) {
+        return true
+    }
+    if ((tracktype == "grade3" || tracktype == "grade4" || tracktype == "grade5") && ANYTHING_FULLY_PAVED.contains(surface)) {
```
FloEdelmann, line 26: Not sure if this really improves readability, but just to get out the suggestion:

```kt
private fun String.tracktypeGrade(): Int = this.replaceFirst("grade", "").asIntOrNull() ?: -1

// then later:
if (tracktype.tracktypeGrade() in 3..5 && surface in ANYTHING_FULLY_PAVED) {
```

or maybe:

```kt
if (tracktype in setOf("grade3", "grade4", "grade5") && surface in ANYTHING_FULLY_PAVED) {
```

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceForm.kt
```diff
@@ -13,18 +14,38 @@ class AddRoadSurfaceForm : AImageListQuestAnswerFragment<Surface, SurfaceAnswer>
 
     override fun onClickOk(selectedItems: List<Surface>) {
```
FloEdelmann, line 15: I find the data flow here quite confusing. Passing `removeTracktype` to a function that primarily asks for surface description, or a function with `and` in its name that does two different things look like code smells.

Can it be refactored to something like this? Then each function has one single clear purpose:

```kt
override fun onClick(selectedItems: List<Surface>) {
    val surface = selectedItems.single()
    confirmTracktypeMismatch(surface) { shouldRemoveTracktype ->
        collectSurfaceDescription(surface) { description ->
            applyAnswer(value, description, shouldRemoveTracktype)
        }
    }
}

private fun confirmTracktypeMismatch(
    surface: Surface,
    onTracktypeConfirmed: (shouldRemoveTracktype: Boolean) -> Unit
) {
    val tracktype = osmElement!!.tags["tracktype"]
    if (tracktype == null) {
        onTracktypeConfirmed(false)
    } else if (isSurfaceAndTracktypeMismatching(surface.osmValue, tracktype)) {
        AlertDialog.Builder(requireContext())
            .setTitle(R.string.quest_generic_confirmation_title)
            .setMessage(R.string.quest_surface_tractypeMismatchInput_confirmation_description)
            .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ ->
                onTracktypeConfirmed(true)
            }
            .setNegativeButton(R.string.quest_generic_confirmation_no, null)
            .show()
    }
}

private fun collectSurfaceDescription(
    surface: Surface,
    onSurfaceDescribed: (description: String?) -> Unit
) {
    if (!surface.shouldBeDescribed) {
        onSurfaceDescribed(null)
    } else {
        AlertDialog.Builder(requireContext())
            .setMessage(R.string.quest_surface_detailed_answer_impossible_confirmation)
            .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ ->
                DescribeGenericSurfaceDialog(requireContext(), onSurfaceDescribed).show()
            }
            .setNegativeButton(android.R.string.cancel, null)
            .show()
    }
}
```

matkoniecz: I admit that any time I see lambda expression I need to stop for minute and think about it, but original code is hairy and smelly.

`confirmTracktypeMismatch` - what you think about `confirmPotentialTracktypeMismatch` or `investigateTracktypeMismatch`?

`confirmTracktypeMismatch` sounds a bit weird to me as confirmation will be needed only in rare cases

FloEdelmann: `confirmPotentialTracktypeMismatch` sounds good

app/src/main/java/de/westnordost/streetcomplete/osm/Surface.kt
westnordost: but this creates a `HashSet` every time it is called

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceForm.kt
matkoniecz: used with a minor improvements

matkoniecz:
I am now assuming that @westnordost is at least not disliking this idea.

westnordost:
Yes, idea sounds fine. It is still draft though, which is why I did not look at the code.

app/src/main/res/values/strings.xml
```diff
@@ -1463,6 +1463,7 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="quest_surface_detailed_answer_impossible_confirmation">Are you sure that it is impossible to specify the surface? Note the “Differs along the way” answer option that allows you to cut the way where the surface changes. Please use “Can’t say” if there is a single surface but it is not available as an answer.</string>
     <string name="quest_surface_detailed_answer_impossible_description">Please briefly specify the nature of the surface here, for example “sandy with patches of cobblestone”. The text length is limited to 255 characters.</string>
     <string name="quest_surface_detailed_answer_impossible_title">Describe surface</string>
+    <string name="quest_surface_tractypeMismatchInput_confirmation_description">This answer is not matching what is already mapped about this road. Are you sure that you are answering about the correct road? Are you sure that you personally surveyed along entire length, as always needed?</string>
```
westnordost, line 1466: A bit long and sounds a bit untrusting/condescending towards the user: "personally ... as always needed".

Thinking about this, I am not sure if a confirmation is necessary or useful at all. I mean, there is also no confirmation on a resurvey when the user selected a different surface than what was selected before. After all, the user is on-site, whatever he sees is the ground truth now, regardless of what was tagged before.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/SurfaceAnswer.kt
```diff
@@ -8,12 +8,16 @@ sealed interface SurfaceOrIsStepsAnswer
...
+data class SurfaceAnswer(val value: Surface, val note: String? = null, val replacesTracktype: Boolean = false) : SurfaceOrIsStepsAnswer
 
 fun SurfaceAnswer.applyTo(tags: Tags, key: String) {
     val osmValue = value.osmValue
     val previousOsmValue = tags[key]
 
+    if (replacesTracktype) {
+        tags.remove("tracktype")
+    }
+
```
westnordost, line 20: The previous values of the tags can be accessed here. Checking whether `tracktype` is in conflict with the selected surface answer and deleting it in that case rather belongs here than in the form, just like it is done for `smoothness`. There is no need to attach some boolean variable to the answer.

matkoniecz: I did this way to keep "should ask about tracktype conflict" and "should delete tracktype due to conflict" checks together and protect them from diverging.

app/src/main/res/values/strings.xml
matkoniecz: See #4063 - there are at least some cases where this resulted in wrong data.

And actually, if there is significant change in `surface` (`asphalt` <-> `dirt`) I would also consider asking. (do not ask on transitions like `dirt` <-> `grass` or `gravel` <-> `compacted`)

westnordost: Hm, I've read through #4063 and I do not find it so convincing.

First of all, it is not really clear why for the case which @mcliquid mentioned the wrong surface was selected by a user. We don't know if it is vandalism, fat fingers, carelessness etc.

If there was not any malintent however, then wouldn't this message need to be displayed **always** to reduce the number of wrong answers? Because whether or not something else was recorded before or nothing was recorded before, the situation is the same. The user is on-site now and can answer about the current situation.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/SurfaceAnswer.kt
westnordost: I understand. However, it is not the application logic's duty to know anything about what happens in the UI.

app/src/main/res/values/strings.xml
matkoniecz: There is difference between claiming that surface was `asphalt` and is still `asphalt` (very likely), claiming that surface changed from `dirt` to `asphalt` (rare) and claiming that surface changed from `asphalt` to `dirt` (really rare and unusual).

It may make sense to ask for confirmation in cases which are rare or very rare, while not asking for typical ones.

For cases where no surface data was tagged we have no indicator to suspect that it is case of a a rare case worth checking.

Also, in cases where no surface was tagged user is at worst adding false data which is less bad that destroying correctly mapped data.

westnordost: Hmm, okay

app/src/main/java/de/westnordost/streetcomplete/quests/surface/SurfaceAnswer.kt
matkoniecz: In this case I see it as GUI collecting info from user - including confirmation that existing tracktype data should be thrown away

FloEdelmann (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/surface/SurfaceAnswer.kt
```diff
@@ -14,10 +15,23 @@ fun SurfaceAnswer.applyTo(tags: Tags, key: String) {
     val osmValue = value.osmValue
     val previousOsmValue = tags[key]
 
+    var replacesTracktype = false
+    if (tags.containsKey("tracktype")) {
+        if (isSurfaceAndTracktypeMismatching(osmValue, tags["tracktype"]!!)) {
+            replacesTracktype = true
+        }
+    }
```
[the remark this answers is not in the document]
matkoniecz: I cannot decide whether I dislike more splitting command into multiple lines or overly elaborate nested ifs.

westnordost (approved):
Looks good.

It is not absolutely necessary, but I'd prefer if there was a test for `isSurfaceAndTracktypeMismatching`. Right now, there is neither a test for `SurfaceAnswer.applyTo` nor any regarding applying the surface for any of the `Add*Surface` classes.

app/src/test/java/de/westnordost/streetcomplete/osm/SurfaceKtTest.kt
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    fun `unknown tracktype does not crash or conflict`() {
+        Assert.assertEquals(false, isSurfaceAndTracktypeMismatching("paving_stones", "lorem ipsum"))
+    }
+
+    @Test
+    fun `unknown surface does not crash or conflict`() {
+        Assert.assertEquals(false, isSurfaceAndTracktypeMismatching("zażółć", "grade1"))
+    }
+}
```
westnordost, line 31:
```suggestion
package de.westnordost.streetcomplete.osm

import org.junit.Assert.assertFalse
import org.junit.Assert.assertTrue
import org.junit.Test

class SurfaceKtTest {
    @Test
    fun `poor tracktype conflicts with paved surface`() {
        assertTrue(isSurfaceAndTracktypeMismatching("asphalt", "grade5"))
    }

    @Test
    fun `high quality tracktype conflicts with poor surface`() {
        assertTrue(isSurfaceAndTracktypeMismatching("gravel", "grade1"))
    }

    @Test
    fun `high quality tracktype fits good surface`() {
        assertFalse(isSurfaceAndTracktypeMismatching("paving_stones", "grade1"))
    }

    @Test
    fun `unknown tracktype does not crash or conflict`() {
        assertFalse(isSurfaceAndTracktypeMismatching("paving_stones", "lorem ipsum"))
    }

    @Test
    fun `unknown surface does not crash or conflict`() {
        assertFalse(isSurfaceAndTracktypeMismatching("zażółć", "grade1"))
    }
}
```

westnordost (approved):
thank you for adding the tests!

matkoniecz:
https://github.com/streetcomplete/StreetComplete/pull/4105 seemed a good idea

later following appeared:

https://github.com/streetcomplete/StreetComplete/issues/4345
https://github.com/streetcomplete/StreetComplete/issues/5236
https://github.com/streetcomplete/StreetComplete/issues/5951 
https://github.com/streetcomplete/StreetComplete/pull/5962
https://github.com/streetcomplete/StreetComplete/pull/5980
https://github.com/streetcomplete/StreetComplete/issues/6524
https://github.com/streetcomplete/StreetComplete/issues/6258
https://github.com/streetcomplete/StreetComplete/issues/6272
https://github.com/streetcomplete/StreetComplete/issues/6524
https://github.com/streetcomplete/StreetComplete/issues/6611

maybe more

PR #4111, opened 2022-06-14
Add summit cross quest

### LINKED ISSUE #4095: Quest suggestion: Summit cross
*Note: this is a follow-up issue to #561.*

### General
Affected tag(s) to be modified/added: [`summit:cross`](https://wiki.openstreetmap.org/wiki/Key:summit:cross)
Question asked: **Is there a summit cross here?**

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose:
  [3.2k uses](https://taginfo.openstreetmap.org/keys/summit%3Across); [may be useful for orientation](https://github.com/streetcomplete/StreetComplete/issues/561#issuecomment-325673300).
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one): `yes` or `no`
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary:
  It's easy to answer if the peak is clearly defined. Crosses are not easy to see on aerial imagery.
- [x] 💤 Not an overwhelming percentage of quests have the same answer (No spam):
  [93% of current uses](https://taginfo.openstreetmap.org/keys/summit%3Across#values) are `yes`, but that may be because armchair mappers only add it when it is present. From my experience, not every summit has a cross.
- [x] 🕓 Applies to a reasonable number of map data (Worth the effort):
  Hard to say because the filter will be quite complex. But this [simple filter](https://overpass-turbo.eu/s/1jbS) already shows at least 700 quests in a small area. Also, it [may be one of few quests in mountainous areas](https://github.com/streetcomplete/StreetComplete/issues/561#issuecomment-325575505).

@matkoniecz [prefers](https://github.com/streetcomplete/StreetComplete/issues/561#issuecomment-325673300) adding a node with `man_made = cross` instead (which would not be possible with StreetComplete), two people on the [discussion page](https://wiki.openstreetmap.org/wiki/Talk:Tag:summit:cross%3Dyes) had the same suggestion. I'd argue that the tags don't rule each other out, i.e. it's the same situation as with `bench = yes` on a bus stop with an `amenity = bench` nearby. Maybe tagging `summit:cross = yes` might even help armchair mappers to find the place to add a `man_made = cross` node.

### Ideas for implementation

**Element selection:** Similar to [`AddSummitRegister`](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/summit_register/AddSummitRegister.kt) after #4075:

```
nodes with
  natural = peak
  and name
  and (!summit:cross or summit:cross older today -16 years)
```

* if they are close to a hiking path
  (`ways with highway = path and sac_scale ~ mountain_hiking|demanding_mountain_hiking|alpine_hiking|demanding_alpine_hiking|difficult_alpine_hiking`)
* or if they are close to a hiking route
  (`relations with route = hiking`)
  → currently non-functional, as [routes are not fetched](https://github.com/streetcomplete/StreetComplete/issues/561#issuecomment-1141411959)
* or if they have `summit:register = yes`
* or if they have `summit:cross` (always resurvey)

Resurvey interval 16 years (vs. 4 years in `AddSummitRegister`) because constructing a summit cross is much more difficult than installing a summit register.

**Metadata needed:** Ask in same countries as `AddSummitRegister` (list gathered in https://github.com/streetcomplete/StreetComplete/issues/561#issuecomment-325623974)

**Proposed UI:** Yes/No

FloEdelmann:
Fixes #4095.


app/src/main/java/de/westnordost/streetcomplete/quests/summit/SummitMarkingsUtil.kt
```diff
@@ -0,0 +1,48 @@
+package de.westnordost.streetcomplete.quests.summit
+
+import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
+import de.westnordost.streetcomplete.data.osm.geometry.ElementPolylinesGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.quest.NoCountriesExcept
+
+val COUNTRIES_WHERE_SUMMIT_MARKINGS_ARE_COMMON = NoCountriesExcept(
```
matkoniecz, line 8: Not sure is it a good idea to have both in the same category, there can be noncristian country with summit registers for example, or reverse (Poland may qualify for summit crosses for example - I am not entirely sure)

westnordost: I agree. I guess summit crosses may be present in any Christian country that has mountains and kind of shares the culture to put crosses on them (e.g. maybe Phillippines too? https://en.wikipedia.org/wiki/Mount_Samat#/media/File:Mount_Samat_remote_view.jpg).

Summit registers seem to be more a question of whether mountain hiking as a hobby is widespread.

But in any case, please do not feel pushed to do a worldwide research where one can find summit crosses. Instead, just add those countries where they are well known / where some are already tagged in OSM.

westnordost (commented):
Does the summit cross tag include "summit markings" that are not a cross?

E.g. https://de.wikipedia.org/wiki/Gipfelkreuz#/media/Datei:Gipfelkreuz_Kreuzberg.jpg does probably not count as a cross, but a summit marking. In general, would it not be better to ask and tag generally for any kind of summit marking?

mcliquid:
> Does the summit cross tag include "summit markings" that are not a cross?
> 
> E.g. https://de.wikipedia.org/wiki/Gipfelkreuz#/media/Datei:Gipfelkreuz_Kreuzberg.jpg does probably not count as a cross, but a summit marking. In general, would it not be better to ask and tag generally for any kind of summit marking?

In my humble opinion, this is not a "summit marking" but an old border stone of the German-Austrian border. These stones are still widely used on the old border, even though the stones are increasingly rotting. The "B" on the stone stands for Bavaria. 1961 stands for the time of the [iron curtain](https://en.wikipedia.org/wiki/Iron_Curtain).
Some more information about it is unfortunately only available in German here: https://www.sn.at/wiki/Grenzsteine

westnordost:
And the grey pole?

mcliquid:
> And the grey pole?

Sure! I'm terribly sorry about that. I was so focused on the stone!

app/src/main/java/de/westnordost/streetcomplete/quests/summit/SummitMarkingsUtil.kt
FloEdelmann: Ok. What I did in [`e69d228`](https://github.com/streetcomplete/StreetComplete/pull/4111/commits/e69d228ed3bbc0176f27874d56e8d1ebb46eb9a1):

* separated the quest filters
* enabled the quest in Poland
* disabled the quest in Slovenia, Greece and the US due to missing evidence that summit crosses are a thing there
* updated the sources for the other countries

FloEdelmann:
> Does the summit cross tag include "summit markings" that are not a cross?

I couldn't find any source that says that it does, or that there is another well-established tag that does.

> E.g. [de.wikipedia.org/wiki/Gipfelkreuz#/media/Datei:Gipfelkreuz_Kreuzberg.jpg](https://de.wikipedia.org/wiki/Gipfelkreuz#/media/Datei:Gipfelkreuz_Kreuzberg.jpg) does probably not count as a cross, but a summit marking.

Yeah, that's indeed a difficult case (also see the discrepancy between the file name and the file description there, introduced [a few minutes after file upload by the original author](https://commons.wikimedia.org/w/index.php?title=File:Gipfelkreuz_Kreuzberg.jpg&diff=next&oldid=109257120&diffmode=source)).

> In general, would it not be better to ask and tag generally for any kind of summit marking?

Probably yes, but while there is no currently agreed tagging for that, I think asking specifically for summit crosses is fine.

app/src/main/java/de/westnordost/streetcomplete/quests/summit/AddSummitCross.kt
```diff
@@ -25,8 +26,24 @@ class AddSummitCross : OsmElementQuestType<Boolean> {
...
+    override val enabledInCountries = NoCountriesExcept(
+        // Europe
+        "AT", // https://de.wikipedia.org/wiki/Gipfelkreuz
+        "CH", // https://de.wikipedia.org/wiki/Gipfelkreuz
+        "DE", // https://de.wikipedia.org/wiki/Gipfelbuch
+        "ES", // https://es.wikipedia.org/wiki/Cruz_de_la_cumbre
+        "FR", // https://fr.wikipedia.org/wiki/Croix_sommitale
+        "IT", // https://it.wikipedia.org/wiki/Croce_di_vetta
+        // not "NL": https://nl.wikipedia.org/wiki/Gipfelkreuz is about "foreign" summit crosses e.g. in the Alps
+        "PL", // https://it.wikipedia.org/wiki/Croce_di_vetta#Alcuni_esempi_di_croci_di_vetta
```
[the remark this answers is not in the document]
matkoniecz: http://www.magiagor.com.pl/osiagniecia-krzyz-na-gorskim-szlaku/ lists 67 known ones. I expect that vast majority do not have them, but we can simply remove it later

matkoniecz (approved):
I am still not sure is it a good quest in Poland but that can be easily changed

westnordost (approved):

PR #4140, opened 2022-06-19
Upgrade ktlint to v0.46.0

FloEdelmann:
Please see the individual commit descriptions.


app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
```diff
@@ -58,6 +58,7 @@ import de.westnordost.streetcomplete.quests.building_type.BuildingType.TRANSPORT
 import de.westnordost.streetcomplete.quests.building_type.BuildingType.UNIVERSITY
 import de.westnordost.streetcomplete.quests.building_type.BuildingType.WAREHOUSE
 
+@Suppress("ktlint:paren-spacing")
```
matkoniecz, line 61: I am against validator-specific comments, I would prefer entire rule to be disabled

matkoniecz (changes requested):

westnordost:
\*sigh\*, ktlint plugin, generating more work.

**I don't want this**
- Enable and fix annotation rule
- Enable and fix no-empty-first-line-in-method-block rule 
- Fix some spacing-between-declarations-with-annotations issues 
- Enable and fix spacing-between-declarations-with-comments issues 

**This is okay**
- Fix some no-multi-spaces issues

**This is debatable**
- Enable and fix multiline-if-else rule
I personally would not like this, but it'd be fine if I am being outvoted on this

app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingType.kt
FloEdelmann: Okay, I'll disable the rule globally and remove the `@Suppress` comments.

smichel17:
I maintain my dislike for if statements without parens (except when used like a ternary operator), but given my low level of activity here recently, I don't think my vote should be considered very important.

Issue #4122, opened 2022-06-15
StreetComplete refuses to operate without authorization to upload GPX tracks to OSM


gdt:
* summary

I was logged out, and on attempting to log in was presented with the OSM oauth page, which was fine.

* reproduce

1) Upgrade from an earlier version that for whatever reason lost credentials.
1) Solve a few quests (3).   (Parking lot type, do you have to pay, surface of parking lane, but surely that does not matter.)
1) See a screen "You have 3 not yet published changes.  You need to login to have them uploaded."
1) Click LOGIN
1) Enter username and password, click remember me, click login
1) On 'Authorize" screen, uncheck "upload GPS traces", because I don't want to permit StreetComplete to do that [1]
1) Click "Grant Access"
1) See error toast that says "Authorization failed.  All listed permissions must be granted".

It seems to me that not having track upload permission should just cause the app to work normally except that when trying to upload a track, fail with "permission is known not to be granted" rather than "tried and got an error from the server".  Or maybe just disable the setting that would upload - doesn't really matter exactly how this is handled, as long as it doesn't cause the app to completely fail to function.  

I don't see anything in settings about uploading tracks.  So presumably that means tracks are never uploaded?

I get it that "modify map" is not optional; that is the point.

[1] The act of mapping publishes information, basically a  location and a time.   The way I use SC, that's a place/time when I solve a quest.   I know SC  fairly recently captures track data and displays it to me, to be helpful in quest solving and understanding what's going on, and that's ok.  But I don't want to upload tracks, because that is more information than necessary to map, and it is too hard to figure out how dangerous that is and what is uploaded when.  (It seems obvious that track upload should be opt in and clicked for ok, but at the moment I don't understand that.)   So, my approach is to just not authorize uploading.

* versions

StreetComplete 43.2 from F-Droid, on Android 12L (via CalyxOS).  (This phone/OS has been using StreetComplete without issues for a while.)

Helium314:
Track upload *is* opt-in, it only happens if you choose to record a track (long press anywhere on the map) and "append" it to a note.
Creating / uploading is only possible in StreetComplete 44 or later.

riQQ:
It looks like 43.2 and higher was set up to necessarily require the *Upload tracks* permission.

See
https://github.com/streetcomplete/StreetComplete/releases/tag/v43.2
and
https://github.com/streetcomplete/StreetComplete/issues/4028#issuecomment-1128691587

gdt:
[...]

PR #4143, opened 2022-06-20
do not require READ_GPS_TRACES permission (fixes  #4141)

### LINKED ISSUE #4141: reduce required priviledges to upload GPS traces (do not require to "read your private GPS traces")
**Use case**

PR https://github.com/streetcomplete/StreetComplete/pull/3573 implemented ability to upload GPS traces.

As noted in https://github.com/streetcomplete/StreetComplete/issues/4122#issuecomment-1159446934 this seems to require that users grant not only a right to `upload GPS traces` (which is obviously needed) but also a right to `read your private GPS traces` which should **not** be needed. That is unnecessary requirement which allows the SC to access all history and tracks by the user marked private, and as such should be eliminated in order to preserve as much of user's privacy, following the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). 

**example**:
I've done some testing and it seems that GPS tracks get uploaded just fine (as [API 0.6 docs suggest](https://wiki.openstreetmap.org/wiki/API_v0.6#GPS_traces)) when  `read your private GPS traces` is **not** granted, however SC does not seem to detect successful upload, and forces re-login and retries uploading traces.

That results in traces being uploaded multiple times, while OSM Note get created only when both `upload GPS traces` and `read your private GPS traces`  are finally granted.

e.g. https://www.openstreetmap.org/note/3230296 was created when I finally allowed both of them, while several traces (see multiple instances of `_54432_10_19T12_53_20.000000Z.gpx` at https://www.openstreetmap.org/user/mnalis%20ALTernative/traces).

**Proposed Solution**

SC should **not** require `read your private GPS traces` permission.

Original author @goldbattle [offered to look into it](https://github.com/streetcomplete/StreetComplete/issues/4122#issuecomment-1159621000).

westnordost:
Note: Need to edit the oauth application to not ask for that permission anymore when ~everyone has updated to a version which includes this PR.

Requesting review by @goldbattle

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsUploader.kt
@@ -11,6 +11,7 @@ import de.westnordost.streetcomplete.data.osmtracks.Trackpoint
 import de.westnordost.streetcomplete.data.osmtracks.TracksApi
 import de.westnordost.streetcomplete.data.upload.ConflictException
 import de.westnordost.streetcomplete.data.upload.OnUploadedChangeListener
+import de.westnordost.streetcomplete.data.user.UserDataSource
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -23,6 +24,7 @@ import java.net.URLEncoder
 class NoteEditsUploader(
     private val noteEditsController: NoteEditsController,
     private val noteController: NoteController,
+    private val userDataSource: UserDataSource,
     private val notesApi: NotesApi,
     private val tracksApi: TracksApi,
     private val imageUploader: StreetCompleteImageUploader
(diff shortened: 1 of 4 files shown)


mnalis:
Unless I messed something when merging, it [fails to compile](https://github.com/mnalis/StreetComplete/runs/6975739189?check_suite_focus=true#step:5:139) for me:

```
> Task :app:compileDebugKotlin
e: /home/runner/work/StreetComplete/StreetComplete/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/edits/NoteEditsModule.kt: (8, 65): No value passed for parameter 'imageUploader'
```

westnordost:
alright, fixed now

goldbattle:
Looks good to me assuming `UserDataSource` stores the correct username that matches OSM. Will be able to test maybe this weekend, but I don't see any issue in the code directly.

You are correct with your comment about the oauth. Sorry, about not implementing correct the first time.

westnordost:
> Seems to work fine. When I try to upload without any GPS traces permission, I'm redirected to login again.

That is weird though...

mnalis:
> > Seems to work fine. When I try to upload without any GPS traces permission, I'm redirected to login again.

> That is weird though...

Huh? But that was **exactly what I was expecting** to happen, because you said in https://github.com/streetcomplete/StreetComplete/issues/4122#issuecomment-1156875826 :

> What I could do is to simply not require these permissions. This means that if you actually create a GPX track to "attach" to a note and try to upload it, you will automatically be logged out and prompted to login. Without explanation why you were logged out.

Or do you mean by that _"This is weird"_ that it would be nice if there were an message explaining to the user why they must re-login?

westnordost:
No,  i misread. I thought you wrote "without gps trace read permission "

PR #4145, opened 2022-06-20
Fix TimePicker hour/minute deprecations

TacoTheDank:
`currentHour` and `currentMinute` are deprecated in API 23. This should fix that deprecation for now.


westnordost:
I see no reason to produce more code just so that this particular deprecated function is not called as long as it is not removed in any new version of Android.

TacoTheDank:
Mm, alright.

PR #4152, opened 2022-06-23
use fake map to pass properties to Tangram-ES

westnordost:
Oftentimes there are thousands of quests to display. To display them, they are set as `Feature`s on a tangram-es `MapData` layer. Each `Feature` requires a `Map<String,String>` of properties to be passed in, which control how the particular feature will look on the map. Hence, when there are thousands of quests to display, thousands of `Map<String,String>` are created.

Additionally, when you solve or undo a quest, the whole tangram-es `MapData` layer must be re-set with all the pins that should be displayed. Hence, the above is done all the time (i.e. when solving a quest).

Finally, tangram-es does not actually do anything with the `Map<String,String>` other than iterating through it via `map.getEntries()`. Hence, we can just pass a `List` disguised as a `Map` and this is not slower, maybe a little faster. More importantly, it should reduce spikes in memory use (and thus GC activity) when solving a quest, which may incidentally also make it faster.

But now sure how much of a performance effect this would actually have. Asking @Helium314 for feedback.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/PinsMapComponent.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/PinsMapComponent.kt
@@ -5,6 +5,7 @@ import com.mapzen.tangram.geometry.Point
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.screens.main.map.tangram.KtMapController
 import de.westnordost.streetcomplete.screens.main.map.tangram.toLngLat
+import de.westnordost.streetcomplete.util.ListMap
 
 /** Takes care of displaying pins on the map, e.g. quest pins or pins for recent edits */
 class PinsMapComponent(ctrl: KtMapController) {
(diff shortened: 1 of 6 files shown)


Helium314:
Interestingly, performance appears to be slightly better when leaving the map as is is currently.
with this PR: 260-330 ms for 1600 quests (45-80 for creating list of pins)
without this PR: 250-300 ms for 1600 quests (35-60 for creating list of pins)
The difference is rather small, but I find it a bit too consistent to be coming from random fluctuations.

westnordost:
Thanks for testing. Weird though. Anyway, then let's trash it.

westnordost:
> with this PR: 260-330 ms for 1600 quests (45-80 for creating list of pins)
> without this PR: 250-300 ms for 1600 quests (35-60 for creating list of pins)

You meant creating the points, right? Because I didn't change anything on creating of the pins.

Quest ---creates---> Pin ---creates---> Point ----pass to-----> layer.setFeatures

Helium314:
> You meant creating the points, right? Because I didn't change anything on creating of the pins.

Right, it's `Points`s, sorry. This is the part I meant:

```kotlin
pins.map { pin ->
            // avoid creation of intermediate HashMaps.
            val tangramProperties = listOf(
                "type" to "point",
                "kind" to pin.iconName,
                "importance" to pin.importance.toString()
            )
            Point(pin.position.toLngLat(), ListMap(tangramProperties + pin.properties))
        }
```
Remaining ~200-250 ms is for `pinsLayer.setFeatures`

Issue #4148, opened 2022-06-22
Ask for shoulder on motorways and links even if there is not cycling/pedestrian infrastructure


mnalis:
**Use case**

I've recently noticed that I'm not being asked for shoulders where I think it would be reasonable (and shoulders exist), e.g. on this motorway: https://www.openstreetmap.org/note/3218529

It seems to because the code requires motorways to have cycling or pedestrian infrastructure in order for this quest to be asked:
https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/shoulder/AddShoulder.kt#L35-L36

Looking at the other part of that query for other types of road it seems not to have that limitation (it may even explicitly look for `sidewalk=no` in order to allow it, which is completely opposite!).

I do not get the reasoning (especially as in my part of the world, motorways are prohibited from having them by law).
Is that intended behaviour? (why?) 

It looks like an error to me, because e.g.
`highway=primary`+`motorroad=yes` (without foot/bicycle=yes) seems it would be allowed, but `highway=motorway` wouldn't.

**Proposed Solution**

Enable shoulder quest on `highway=motorway` too. (or add a comment explaining a logic / why it's excluded)

matkoniecz:
See https://github.com/streetcomplete/StreetComplete/commit/6908b2d9d42a0a0d5733269e43fbaeb196b87f9f - previously it was not asked at all, as nearly all motorways have shoulders

But in this case this info is especially valuable, so it started to be asked

> or add a comment explaining a logic / why it's excluded

feel free to make a PR

mnalis:
> [...] nearly all motorways have shoulders

Thanks, @matkoniecz; so the reasoning being `shoulder=yes` is implicit on motorway, so no need to spam users. 

I'm still confused then **why we do ask** about shoulders for example for `highway=trunk` + [motorroad=yes](https://wiki.openstreetmap.org/wiki/Key:motorroad)? According to wiki, it would imply _"[...] road on which the traffic rules are the same as on a motorway"_; yet we treat it differently then `highway=motorway`? Is that on purpose?

> I would suggest asking at least where `bridge=*` or `tunnel=*` as these are more likely to have a `shoulder=no` compared to a regular section, does that sound reasonable?

Makes sense to me, @andrewharvey; I can add that too in PR then (if nobody complains).

westnordost:
[...]

PR #4155, opened 2022-06-24
Ask shoulders on some specific parts of motorways (bridges/tunnels)

mnalis:
- Some parts of motorways might not have shoulders, even if the rest of the `highway=motorway` usually does, so ask for shoulders on that parts (bridges and tunnels).
- bridges have (often used) [more values](https://taginfo.openstreetmap.org/keys/bridge#values) than just generic `yes`, so account for them as bridges too
- documents the motorway part of the quest, as it can be confusing (see linked issue)

Closes: https://github.com/streetcomplete/StreetComplete/issues/4148

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shoulder/AddShoulder.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shoulder/AddShoulder.kt
@@ -11,6 +11,7 @@ class AddShoulder : OsmFilterQuestType<ShoulderSides>() {
 
     /* Trunks always, smaller roads only if they are either motorroads, bridges or tunnels or if
      * they are likely rural roads (high max speeds - implicit or explicit - or no sidewalk).
+     * Motorways are assumed to almost always have shoulders, except in some special cases (see #4148).
      *
      * The existence of a parking lane, cycle lane excludes the existence of a shoulder because they
      * are in themselves kind of a shoulder (with a special purpose)
@@ -24,7 +25,7 @@ class AddShoulder : OsmFilterQuestType<ShoulderSides>() {
               and (
                 motorroad = yes
                 or tunnel ~ yes|building_passage|avalanche_protector
-                or bridge = yes
+                or (bridge and bridge != no)
                 or sidewalk ~ no|none
                 or !maxspeed and highway = trunk
                 or maxspeed > 50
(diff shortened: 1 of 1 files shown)


mnalis:
Question: should I also make `tunnel` more generic (i.e. `tunnel and tunnel != no`) too? 
Currently is matches only `yes|building_passage|avalanche_protector`, but there are few other [values](https://taginfo.openstreetmap.org/keys/tunnel#values) used.

mnalis:
[tested](http://overpass-turbo.eu/?q=LyoKVGhpcyBoYcSGYmVlbiBnxI1lcmF0ZWQgYnkgdGhlIG92xJJwxIlzLXR1cmJvIHdpemFyZC7EgsSdxJ9yaWdpbmFsIHNlxLBjaMSsxIk6CsOiwoDCnMSEZ2h3YXk9bW90b3LFjsSaYW7El2LEt2TEkD0qIMWZxZvFncSQIT1ub8WIwp0KKi8KW291dDpqc29uXVt0aW1lxbLFtDI1XTsKLy_Ej8SUxJ1yIHJlc3VsdHMKKAogIMaHIHF1xJLEmsSjcnQgZsWVOiDFiMWKxLjFjcWPxZHFk8WVxZfFosWaxJjFpWXFoMaxxaRpxZ5lxafFqcWrxpfFhHlbIsWLxqt5Ij0ixZLFlMWWxY8ixboixZzGucSQx47HgseRxroixacixanHjih7e2LEqnh9fSnGhcelxobGiHDEt27GosaOxpDGksaUxoDEmG9kecaFPsaFx7Jza2XEvXF0Ow&c=BQLCmcVK6R) and it seems to work.

westnordost:
So, I'll have to close this PR as will not fix, to stay consistent with the decision to not ask anything about motorways in any quest. I am not sure how the part after the "or" managed to get into the code at all, maybe someone argued that as long as there is a motorway which explicitly allows foot and bicycle traffic, then it is safe enough for surveyors to ask about it because apparently it is accessible (somehow) to pedestrians.

matkoniecz:
> I am not sure how the part after the "or" managed to get into the code at all, maybe someone argued that as long as there is a motorway which explicitly allows foot and bicycle traffic, then it is safe enough for surveyors to ask about it because apparently it is accessible (somehow) to pedestrians.

I added it in https://github.com/streetcomplete/StreetComplete/commit/6908b2d9d42a0a0d5733269e43fbaeb196b87f9f based on such reasoning

PR #4158, opened 2022-06-26
Allow areas to also be marked as indoors

peternewman:
Unlike steps, there is no restriction for tagging indoor on areas, so there's no need to filter it here.

Fixes post-merge comments in #4151

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurfaceForm.kt
@@ -46,7 +46,7 @@ class AddPathSurfaceForm : AImageListQuestForm<Surface, SurfaceOrIsStepsAnswer>(
 
     private fun createMarkAsIndoorsAnswer(): AnswerItem? {
         val way = element as? Way ?: return null
-        if (way.isArea() || way.tags["indoor"] == "yes") return null
+        if (way.tags["indoor"] == "yes") return null
 
         return AnswerItem(R.string.quest_generic_answer_is_indoors) {
             applyAnswer(IsIndoorsAnswer)


matkoniecz (approved):
looks fine (though I have not tested code - have you did it @peternewman ? )

peternewman:
> looks fine (though I have not tested code - have you did it @peternewman ? )

No I haven't tested it either yet.

mnalis:
To add on discussion from https://github.com/streetcomplete/StreetComplete/pull/4151#discussion_r906437048:

 Selection query in [AddPathSurface.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPathSurface.kt#L15) is:
 `ways with highway ~ path|footway|cycleway|bridleway|steps`.

I was wondering how area could come into play at all, as I believed none of the highways mentioned allowed area, so I though any of them mapped as area can only be an mapping error. Now I see that [highway=footway](https://wiki.openstreetmap.org/wiki/Tag:highway=footway#Areas) wiki says that it might be an area (although it does note that there is no consensus on that), if it is additionally mapped with `area=yes`. (The others ways in that selection query may not be an area, according to their wiki.)

So it should probably be tested how SC handles that (`highway=footway` as a closed-way AKA area) - does it highlight whole area (as it highlights buildings when solving address quest for example), or does it highlight it only as a normal path (which would be wrong for our use). Here is [overpass-turbo query](http://overpass-turbo.eu/?q=LyoKVGhpcyBoYcSGYmVlbiBnxI1lcmF0ZWQgYnkgdGhlIG92xJJwxIlzLXR1cmJvIHdpemFyZC7EgsSdxJ9yaWdpbmFsIHNlxLBjaMSsxIk6CsOiwoDCnMSEZ2h3YXk9Zm9vdMWOxJphbsSXxLDFgD15ZcSGxZjEl3PEqGZhY8SexIUgbnVsbMWIwp0KKi8KW291dDpqc29uXVt0aW1lxbbFuDI1XTsKLy_Ej8SUxJ1yIHLFn8WsdHMKKAogIMaLIHF1xJLEmsSjcnQgxZJyOiDFiMWKxLjFjcWPxZHFk8WVxY8gxaHGs8aSYcWdxZ_Gs8WZxL7FpMWmZSE9KsWvxpnFhHlbIsWbYSI9IsWecyLFviLFi8ateceLIsWSxZTFlseQx4fFo3LFpcWnIiF-Ii4qx5sixLpkxZNyx6HHo8elx5Aoe3tixKp4fX0pxonHucaKxoxwxLduxqTGksWjbMaVCsaExJhvZHnGiT7GiciHc2tlxL1xdDs) to find potential locations for test.

peternewman:
> So it should probably be tested how SC handles that (`highway=footway` as a closed-way AKA area) - does it highlight whole area (as it highlights buildings when solving address quest for example), or does it highlight it only as a normal path (which would be wrong for our use). Here is [overpass-turbo query](http://overpass-turbo.eu/?q=LyoKVGhpcyBoYcSGYmVlbiBnxI1lcmF0ZWQgYnkgdGhlIG92xJJwxIlzLXR1cmJvIHdpemFyZC7EgsSdxJ9yaWdpbmFsIHNlxLBjaMSsxIk6CsOiwoDCnMSEZ2h3YXk9Zm9vdMWOxJphbsSXxLDFgD15ZcSGxZjEl3PEqGZhY8SexIUgbnVsbMWIwp0KKi8KW291dDpqc29uXVt0aW1lxbbFuDI1XTsKLy_Ej8SUxJ1yIHLFn8WsdHMKKAogIMaLIHF1xJLEmsSjcnQgxZJyOiDFiMWKxLjFjcWPxZHFk8WVxY8gxaHGs8aSYcWdxZ_Gs8WZxL7FpMWmZSE9KsWvxpnFhHlbIsWbYSI9IsWecyLFviLFi8ateceLIsWSxZTFlseQx4fFo3LFpcWnIiF-Ii4qx5sixLpkxZNyx6HHo8elx5Aoe3tixKp4fX0pxonHucaKxoxwxLduxqTGksWjbMaVCsaExJhvZHnGiT7GiciHc2tlxL1xdDs) to find potential locations for test.

It works absolutely fine and highlights the whole area (that happens to have an alternative title for areas), see for example:
https://www.openstreetmap.org/way/1008689617/history

I know this happens to be a road/answered via the road quest, it just happened to be the first one I found.

Maybe from that, highway=pedestrian should be moved into the path surface quest, or the indoor tag option should be added to both or something.

matkoniecz:
@peternewman So you tested it and it works fine, right?

westnordost:
ping

peternewman:
> @peternewman So you tested it and it works fine, right?

No sorry, I meant SC currently works with area mapped locations, answering @mnalis query about how it works with areas.

And even areas for this quest, such as https://www.openstreetmap.org/way/306452427 . There's also a test one I can check it on, so I just need to build an APK, install it and give it a go: https://www.openstreetmap.org/way/545679493

PR #4172, opened 2022-06-29
put address quests higher

matkoniecz:
this is based on UX testing - some experienced mappers were disappointed by SC and were unaware that it asks also about "important" data

At least one person specifically mentioned going into setting and looking at quest list.

Also, even if address data can be tricky to capture - it is still extremely useful and worth putting higher on the list.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -209,6 +209,15 @@ fun questTypeRegistry(
     element is based on wrong data while the note is not resolved */
     OsmNoteQuestType,
 
+    /* ↓ 0. special case  ------------------------------------------------------------------  */
+    // address: usually only visible when just in front + sometimes requires to take "other answer"
+    // on the other hand it is very useful, and listing it higher show new mappers that SC
+    // asks also about clearly useful data, so someone exploring SC by looking at quest list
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -209,6 +209,15 @@ fun questTypeRegistry(
     element is based on wrong data while the note is not resolved */
     OsmNoteQuestType,
 
+    /* ↓ 0. special case  ------------------------------------------------------------------  */
+    // address: usually only visible when just in front + sometimes requires to take "other answer"
+    // on the other hand it is very useful, and listing it higher show new mappers that SC
+    // asks also about clearly useful data, so someone exploring SC by looking at quest list
+    // will not see merely several quests about bus stops or the bench backrest quest
+    // It is also not very likely to crowd out other quests, as is limited to buildings
```
peternewman, line 217: I think that's likely to obscure things related to food and drink establishments (like seating, dietary etc).

Although I do appreciate its an important quest.

matkoniecz:
I agree that it would be an exception and change from current practice.

> opens a door for any number of similar requests to put "important data" to the front

I intended it as a sole such case, as it is enough to demonstrate clearly that SC asks also about important data by pumping this address quest to top.

But I see risk of getting more requests like this.

westnordost:
I'll close that then.

Issue #4164, opened 2022-06-27
Sidewalk surface quest: Surface image obscures text


Jean-BaptisteC:
**How to Reproduce**

- Find quest `What's the surface of the sidewalk here ?`
- Select surface
- Surface image is superimposed on illustration (Surface image is circular on illustration)

**Versions affected**
Android 12 - v45.0-beta1

westnordost:
What exactly is the issue here?

Jean-BaptisteC:
![Screenshot_20220627-181241~2](https://user-images.githubusercontent.com/87148630/176477991-2516be0f-723e-4065-b7c2-f3bca4dd9d50.png)

westnordost:
And what is the issue?

mnalis:
Ah, I see, so the **text** should be on **_uppermost_** layer, so it is always visible. Then below text would be the circle with material, and below that the background image of the road.

Issue #4035, opened 2022-05-16
Since yesterday cannot login anymore


tonsnoei:
Cannot login with the Android app. 

**How to Reproduce**
Install fresh android app
Press login button in the profile screen. 
A error toast is shown "authorization server not found" 

**Versions affected**
Latest

10992-osm:
Just to confirm, by latest you mean v43.2 (not v43.1)?

timothywashere:
Version 43.2 is now available on Google Play, yay

matkoniecz:
See https://github.com/streetcomplete/StreetComplete/issues/4028#issuecomment-1126990804

If you have 43.2 and it still fails: please comment

PR #4173, opened 2022-06-29
bug issue template: ask users what they expected to happen

mnalis:
- Sometimes it is not clear what users expected to happen when they report bug, so ask them to explicitly tell.
(e.g. latest example in https://github.com/streetcomplete/StreetComplete/issues/4164)
- Also clarify that we would like exact SC version number, instead of "latest"
(e.g. as in https://github.com/streetcomplete/StreetComplete/issues/4035)


matkoniecz (approved):

peternewman (approved):
LGTM.

We could use the issue forms stuff to be even more explicit, but perhaps that would be too restrictive?
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

PR #4125, opened 2022-06-16
Spatial cache for OsmQuestController, OsmNoteQuestController and MapDataController

Helium314:
A first version for the `OsmQuestController` cache discussed in #4079.
I left the comments to clarify my motivations.

Main benefits
* much faster `get` for quests in view (called on opening and solving quests)
* much faster `getAllVisibleInBBox` when tiles are cached, otherwise marginally slower
This translates to a ~30% improvement of `QuestPinsManager.onNewTilesRect`, as creating and updating pins is rather slow.

What is missing:
* unit tests
* decide cache size
* clear / trim cache on low memory
* maybe bugs?

PR #4181, opened 2022-07-01
Simpler bbox for ElementPointGeometry

Helium314:
When testing for #4125 I noticed that it's possible to save a surprising amount of memory by not storing node geometries.
I traced this to the (lazy) creation of bbox for the node, which creates two new `LatLon`s instead of re-using the node position.

I guess currently this change will not matter, but could be a slight improvement depending on how the cache will be done.


westnordost:
cool!

PR #4183, opened 2022-07-03
Don't ask path smoothness for highway=pedestrian

### LINKED ISSUE #4180: Path and Road smoothness both ask for highway=pedestrian
Currently both smoothness quests ask smoothness for highway=pedestrian.
This is not horrible, but confusing when hiding a smoothness quest. I think only one of the quests should apply.

I don't have any clear preference which quest shoud apply. RoadSmoothness makes a bit more sense, as then it's in line with surface quests.

Helium314:
fixes #4180

Smoothness for highway=pedestrian is already asked in the road smoothness quest.
According to the (now removed) comment, I added pedestrian to path smoothness with the idea of having the smoothness description text dependent on whether it's a path or a road. But anyway, looks like this wasn't implemented in the end.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddPathSmoothness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddPathSmoothness.kt
@@ -58,5 +58,4 @@ class AddPathSmoothness : OsmFilterQuestType<SmoothnessAnswer>() {
 }
 
 // smoothness is not asked for steps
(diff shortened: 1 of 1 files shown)


matkoniecz (approved):
looks good

have you tested it?

Helium314:
No, I did not.
Do you think it's necessary?

westnordost:
It's always better, one may never know what one overlooked.

matkoniecz:
I have run into compilation failures and severe bugs before that I have not spotted or expected.

Helium314:
Tested, and now there is no path smoothness quest asked for highway=pedestrian (only road smoothness)

PR #4185, opened 2022-07-04
Update tests for VisibleQuestsSource

### LINKED ISSUE #4182: de.westnordost.streetcomplete.data.quest.VisibleQuestsSourceTest tests are failing
getAllVisible does not return those that are invisible in team mode
getAllVisible returns only quests of types that are visible

https://gist.github.com/matkoniecz/e26c29fad1d7ecca8b2109f5bdf8927a

spotted when investigating something unrelated

Helium314:
fixes #4182 

`getAllVisible returns only quests of types that are visible` is removed. It assumes `OsmQuestSource.getAllVisibleInBBox` returns quest types that are not visible, which should never happen anyway.

`getAllVisible does not return those that are invisible in team mode` failed because now `VisibleQuestsSource.getAllVisible` uses the team mode filter only if team mode is enabled.

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSourceTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSourceTest.kt
@@ -93,23 +93,11 @@ class VisibleQuestsSourceTest {
         assertEquals(2, quests.filterIsInstance<OsmNoteQuest>().size)
     }
 
-    @Test fun `getAllVisible returns only quests of types that are visible`() {
-        val t1 = TestQuestTypeA()
-        val t2 = TestQuestTypeB()
-        val q1 = osmQuest(elementId = 1, questType = t1)
-        val q2 = osmQuest(elementId = 2, questType = t2)
-        on(osmQuestSource.getAllVisibleInBBox(bbox, questTypeNames)).thenReturn(listOf(q1, q2))
-        on(visibleQuestTypeSource.isVisible(t1)).thenReturn(false)
-        on(visibleQuestTypeSource.isVisible(t2)).thenReturn(true)
-
-        val quests = source.getAllVisible(bbox)
-        assertEquals(q2, quests.single())
-    }
-
     @Test fun `getAllVisible does not return those that are invisible in team mode`() {
         on(osmQuestSource.getAllVisibleInBBox(bbox, questTypeNames)).thenReturn(listOf(mock()))
         on(osmNoteQuestSource.getAllVisibleInBBox(bbox)).thenReturn(listOf(mock()))
         on(teamModeQuestFilter.isVisible(any())).thenReturn(false)
+        on(teamModeQuestFilter.isEnabled).thenReturn(true)
 
         val quests = source.getAllVisible(bbox)
         assertTrue(quests.isEmpty())


matkoniecz:
I am quite confused.

Why https://github.com/streetcomplete/StreetComplete/pull/4185/files#diff-98de34ab2347df101b52ab5c67a41ae5e9e9f61474eeea79e53c4ec05076a9e3L102 passes despite that [TestQuestTypeA is visible and TestQuestTypeDisabled is one being disabled?](https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/test/java/de/westnordost/streetcomplete/data/quest/TestQuestTypes.kt)

----------------------

Why https://github.com/streetcomplete/StreetComplete/pull/4185/files#diff-98de34ab2347df101b52ab5c67a41ae5e9e9f61474eeea79e53c4ec05076a9e3L102 passes and then https://github.com/streetcomplete/StreetComplete/pull/4185/files#diff-98de34ab2347df101b52ab5c67a41ae5e9e9f61474eeea79e53c4ec05076a9e3L106 fails anyway?

https://github.com/streetcomplete/StreetComplete/blob/92ffe9bd51215d82ff6c336f0a686d95f00b4adb/app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt#L86 is present

-------------------------------

in https://github.com/streetcomplete/StreetComplete/issues/4182#issuecomment-1173176137

> getAllVisible returns only quests of types that are visible assumes osmQuestSource.getAllVisibleInBBox may return quests that are not visible. If this is the case, getAllVisibleInBBox would be really bad name.

maybe renaming this function would be a good idea? `getAllRequestedInBBox`?

Helium314:
> I am quite confused.
> 
> Why https://github.com/streetcomplete/StreetComplete/pull/4185/files#diff-98de34ab2347df101b52ab5c67a41ae5e9e9f61474eeea79e53c4ec05076a9e3L102 is applied and then https://github.com/streetcomplete/StreetComplete/pull/4185/files#diff-98de34ab2347df101b52ab5c67a41ae5e9e9f61474eeea79e53c4ec05076a9e3L106 fails anyway?
> 
> https://github.com/streetcomplete/StreetComplete/blob/92ffe9bd51215d82ff6c336f0a686d95f00b4adb/app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt#L86
> is present

The `isVisible` in line 86 is only applied to the list of quest types for giving a list of quest types to `OsmQuestSource`, not to each individual quest returned by `OsmQuestSource.getAllVisibleInBBox`

> in [#4182 (comment)](https://github.com/streetcomplete/StreetComplete/issues/4182#issuecomment-1173176137)
> 
> > getAllVisible returns only quests of types that are visible assumes osmQuestSource.getAllVisibleInBBox may return quests that are not visible. If this is the case, getAllVisibleInBBox would be really bad name.
> 
> maybe renaming this function would be a good idea? `getAllRequestedInBBox`?

That's what the `getAllVisible` test already does.

Helium314:
> OK, I am now understanding what is going on - `getAllVisibleInBBox` is supposed to filter out invisible ones while this test checks what happens when `getAllVisibleInBBox` is buggy and returns also invisible ones.

Basically yes... and in that case it's probably better to fix `getAllVisibleInBBox`.

matkoniecz:
But `getAllVisibleInBBox` is not broken, filters out invisible and it was broken only in mockup in this quest, right?

https://github.com/streetcomplete/StreetComplete/blob/36f65b4aa918c418710ead05745e5d233c5092f1/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt#L89

https://github.com/streetcomplete/StreetComplete/blob/e67d1eb9107949162f5bc6a5b475631189a6d9ca/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt#L272

Helium314:
> But `getAllVisibleInBBox` is not broken, filters out invisible and it was broken only in mockup in this quest, right?

Yes, assuming the database returns only quest types on the provided list. So it's about whether we trust the database and our queries to do what they should...

> https://github.com/streetcomplete/StreetComplete/blob/36f65b4aa918c418710ead05745e5d233c5092f1/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt#L89
> 
> https://github.com/streetcomplete/StreetComplete/blob/e67d1eb9107949162f5bc6a5b475631189a6d9ca/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt#L272
> 
> Though maybe having a working test for that would be nice?

The first one may be useful (add a hidden id in `OsmNoteQuestControllerTest.getAll`).
The second one is checked in `OsmQuestControllerTest.getAllVisibleInBBox`.

PR #4188, opened 2022-07-05
Document that all quests should be answerable by users on foot

mnalis:
As mentioned in https://github.com/streetcomplete/StreetComplete/issues/4148#issuecomment-1173640775 and https://github.com/streetcomplete/StreetComplete/pull/4155#issuecomment-1173640311, all quests must be answerable by **users on foot** for consistency reasons, so document that in Quest guidelines.


FloEdelmann:
Maybe also add that to the issue template?

matkoniecz (approved):

mnalis:
updated as requested

QUEST_GUIDELINES.md
```diff
@@ -19,7 +19,7 @@ Consider the following:
...
 - 👨‍💻 **Users are no experts**: No knowledge about OpenStreetMap or any other background knowledge must be necessary
 - 🐿️ **Easy answer**: Users are out and about and impatient. A quick, straightforward and clear answer must be possible
-- 🕵️ **Publicly accessible**: The information the users are supposed to collect must be easily accessible from the outside, don't lead users into trouble. I.e. users will not be able to answer reliably if a building has a basement or not, if a restaurant has toilets or not etc. without entering it.
+- 🕵️ **Publicly accessible on foot**: The information the users are supposed to collect must be easily accessible from the outside by pedestrians, don't lead users into trouble. I.e. users will not be able to answer reliably if a building has a basement or not, if a restaurant has toilets or not etc. without entering it. Also information should not require special means of transport like a car or a bus (e.g. motorway surveying).
```
peternewman, line 22: Erm we specifically ask a number of those example quests which are deemed unsuitable (and other equivalents), although I appreciate they may be mostly disabled by default, shouldn't we call that out here?

PR #4190, opened 2022-07-05
Utilize more KTX extensions

TacoTheDank:
Context for sqlite transaction: https://androidx.tech/artifacts/core/core-ktx/1.8.0-source/androidx/core/database/sqlite/SQLiteDatabase.kt.html

Tell me if you would like any to be removed.


westnordost:
Sorry, I am a bit busy at the moment. Maybe you could review/redact your PR yourself a bit based on these general statement:

PRs that do not add a feature or change the behavior but just change how the code looks like are welcome if they make the code easier to read and understand or remove code duplication. Easier to read and understand code is not always shorter.

westnordost (approved):
Looked through it from top to bottom, looks good! Thank you!

Issue #3502, opened 2021-11-12
New generic other answer: "At slightly wrong position"


westnordost:
**Use case**
I am being asked a question like:
- bench backrest
- hydrant type
- hydrant diameter
- whether X is still here
- if a picnic table is covered
- anything about a bicycle parking
- etc. ... any small POI feature

and I can see that it is at a slightly wrong position by comparing my GPS trace and the location displayed on the map. I could leave a note, but it would be quicker if I could just move the node by a little within this app.

**Proposed Solution**
Technically similar as for split way, but much easier and much less code. Estimated code cost:
- 200 LOC for UI (`MoveNodeFragment.kt`)
- 10 LOC for the serializable data model (`MoveNodeTo.kt`)
- 20 LOC for the (`MoveNodeAction.kt`) + 40 LOC test code
- 40 LOC for adaptions in the quest controller and edits dao

So, about 300 LOC.

**Opinions**
Does this make sense? Is this worth the cost?

peternewman:
Yes please! I often open notes like this. It would also stop the note blocking the further quests.

You'd need to implement #2376 too.

You'd need to either not allow it for nodes like AEDs on the side of a wall, or have some more complicated system to move them, compared to a standalone bin or bench.

It's also probably more accurate to just do it yourself onsite, especially if the satellite stuff was enabled (even if there was an offset you could do them both relatively.

> Does this make sense? Is this worth the cost?

While it seems somewhat trivial if a bin and bench are the wrong way round, the whole point of a map is accuracy so improving that feels like it should be encouraged to me.

mnalis:
I personally would love it, as it would seem easier (and not requiring data network) than launching Vespucci at that coordinates (or leaving notes and post-processing later)

However, I am little worried that average SC user might move nodes wrongly (as they might not be aware of exact definition of GPS Dilution of Precision, what 68% confidence _**actually**_ means, and why approximations that SC shows in form of transparent circle are often wrong and by how much, and why that can't really be fixed). 
I.e. user might think their GPS precision is good and move nodes around according to that, while their position was more wrong then they realized. 

I would still love it, but perhaps hide it behind `Expert mode` checkbox in settings https://github.com/streetcomplete/StreetComplete/issues/471? Or show stern warning on each use (like for answering quests far away from current GPS location)?

FloEdelmann:
See also #2586.

westnordost:
[...]

PR #4202, opened 2022-07-12
Only fetch newest element from API if necessary

### LINKED ISSUE #4078: Improve performance on upload: Only fetch newest element from API if necessary
## Current Behavior

On upload of edits to OSM data,

1. first the app downloads the newest version of the element the edit refers to,
2. then applies the edit to it locally
3. and then uploads the edited element

This means that for every singular edit, two OSM API calls are made.

## The Issue

While this makes the implementation easier, the first call the OSM API is avoidable in 99% of cases: If the OSM element hasn't been edited in the meantime since the the app downloaded it, the app could just apply the edit to the locally persisted OSM element and upload that.

Only if the OSM API responds with a HTTP 409 Conflict error, we know that the upload needs to fetch the newest version of that element, apply the edit and try again once.

This improvement not only makes the upload faster, it also avoids unnecessary internet traffic.

## Challenges

- oftentimes, multiple edits are made on the same element in a row (e.g. add building type, add housenumber, add building levels). An implementation should keep this in mind - it should not run into HTTP 409 conflicts in cases where the edit "in the meantime" has just been done by the app itself
- the OSM API returns a HTTP 409 conflict error also in case one tries to upload a change in a changeset that as already been closed. So, this case needs to be handled separately

Helium314:
fixes #4078

I did *not* yet test whether the modified upload actually works, need to go for a walk first and find something to change.

Related tests are not building, and I failed at making them work, so help would be appreciated.


Helium314:
No, this is because of my changes

Helium314:
Anyway I got it now.
Needed to learn a bit more about how to use Mockito, and ran into trouble because `any()` apparently doesn't actually mean **any**... that's why there is `anyLong()`

westnordost:
Hmm, I think that this approach won't work with e.g. `SplitWayAction`.

Note the interface of `ElementEditAction`:

```kotlin
fun createUpdates(
    originalElement: Element,
    element: Element?,
    mapDataRepository: MapDataRepository,
    idProvider: ElementIdProvider
)
```

It expects a `MapDataRepository`, `MapDataApi` just implements this interface. But `MapDataWithEditsSource` does, too. (And `MapDataController` could, too)

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/split_way/SplitWayAction.kt#L45

`SplitWayAction` will fetch the elements it needs to perform the split from the `MapDataRepository` by itself.

My thinking is, wouldn't it be more straightforward (also in implementation) if on 1st try, the uploader would just pass a `MapDataController` to `edit.action.createUpdates`, and if that fails, pass a `MapDataApi` instead?

(Though, thinking about this more, this is where we would stumble upon an ugly performance workaround we made earlier - that StreetComplete discards all route relations right after download)

Helium314:
Could we just always use `MapDataApi` if `edit.action is SplitWayAction`?

> Right now I am asking myself why even the `element: Element?` is passed, after all the edit action could just fetch that by itself

If the edit action fetches the element, we may need to fetch it again in the uploader on a `ConflictException`, to know whether the changeset is closed or the element changed. Though this was not the case before this PR...

westnordost:
> Could we just always use MapDataApi if edit.action is SplitWayAction?

You mean, go with my suggestion but make an exception for `SplitWayAction` in order to not revert the throwing-away-of-the-route-relations? It seems to me that in its result, that solution would then differ not that much from the current solution. 

What I am thinking is whether that workaround to throw away all route relations on download (see `IGNORED_RELATION_TYPES`) could be revisited, especially when we can get the persisting of the relations to be done in the background thanks to your cache implementation.

Helium314:
See https://github.com/streetcomplete/StreetComplete/issues/2803#issuecomment-827639718 and below. I had identified the the db query for getting members of (huge) relations as the slow part.

Helium314:
I just tested how fast things are when all relations are in the database.
As expected `getMapDataWithGeometry` is still very fast when everything is cached, but loading an area with many huge relations into cache (same area as in #2803) loading from db may even take more than 70 s.

westnordost:
Okay, so either my approach should be taken but a hardcoded exception to this behavior (i.e. always use remote data) must be added for the `SplitWayAction` specifically (and any other future actions that may edit route etc. relations).

Or, we stay with your solution. However, as stated before, I find it not so clean because always a `mapDataApi` is passed to the action. So any other action that actually accesses the `mapDataRepository` to create the changes won't work with that solution (currently, excluding `SplitWayAction`, that's just `DeletePoiNodeAction`).

That there could be other action types in the future may seem abstract, but just look at the `ElementEditAction`s that already exist - it is really easy to add other types of actions. Even the super-complex splitting of ways is just 300 lines of code. For example, there could be a `MovePoiNodeAction` (#3502), a `AddEntrancesToBuildingOutlineAction` (or more generic) etc.. It would be great if this PR was future proof in the way that such possible extensions to the functionality of this app were covered.

Helium314:
I agree that passing `mapDataApi` and the local element is not good.

How about:
* don't pass any element
* remove `element` from `ElementEditAction.createUpdates` and use `mapDataRepository.get<Node/Way/Relation>`
* have a list of actions that are allowed to use local elements (or put it in a boolean for each action)

And do something like:
```kotlin
val localChanges by lazy { edit.action.createUpdates(edit.originalElement, mapDataController, idProvider) }
val remoteChanges by lazy { edit.action.createUpdates(edit.originalElement, mapDataApi, idProvider) }
val changes = if (edit.action.isAllowedToUseLocalChanges())
        localChanges
    else
        remoteChanges

return try {
    uploadChanges(edit, changes, false)
} catch (e: ConflictException) {
    // either changeset was closed, or element modified, or local element was cleaned from db
    if (changes === localChanges) {
        // anything of the 3 may be the problem, try again with remote changes
        try {
            uploadChanges(edit, remoteChanges, false)
        } catch (e: ConflictException) {
            // probably changeset closed
            uploadChanges(edit, remoteChanges, true)
        }
    } else {
        // probably changeset closed
        uploadChanges(edit, remoteChanges, true)
    }
}
```

westnordost:
> don't pass any element
> remove element from ElementEditAction.createUpdates and use mapDataRepository.get<Node/Way/Relation>

Yes, this seems reasonable. I wonder if there was a specific reason why I did not do it like this from the start. So, if there was no reason, then I am all for it. Makes it simpler.

> have a list of actions that are allowed to use local elements (or put it in a boolean for each action)

No, because whether or not an action is allowed to use local data is not a property of that action. The action shouldn't (need to) know that due to some performance consideration, StreetComplete as a whole drops relations on download before even persisting it.
Ideally, which action types should **not** be allowed to use local changes (I'd like to turn that around) would be defined at the same place where it is defined to throw away certain relation types - and be it just as a public constant.

Helium314:
> I wonder if there was a specific reason why I did not do it like this from the start.

I did't find any reason in other actions, or in the history... but actually this change isn't really necessary, I can also keep the element in there and fetch it in the uploader.

So I will put some list next to `IGNORED_RELATION_TYPES`.
My motivation for using a list of allowed actions was to be on the safe side in case this list is not considered when adding a new action. But I don't really care, so I'll use the list of not allowed actions.

Helium314:
> So, if there was no reason, then I am all for it. Makes it simpler.

I did not (yet) change this.
If you're ok with the way it is now, I will use/test it for a while.

Only after some testing I'll remove passing the element / fetching it in the action and test / use it for a while again.

Helium314:
> I will use/test it for a while

And I did, including some logging.
* Normal upload without conflicts: works as expected
* Conflict when uploading a SplitWayAction (way split in other editor): upload using remote changes fails, is tried again with new changeset, fails again and edit is dropped (no change from old behavior, as it's using changes on remote element)
* Conflict when uploading UpdateElementsTagsAction (added node to the way in other editor): upload using local changes fails, is tried again using remote changes, and works
* Conflict when uploading DeletePoiNodeAction (deleted node in other editor): upload using local changes fails, is tried again using remote changes, then tried again with new changeset, and then edit is dropped because "element deleted"
* Conflict when creating local changes (local element was deleted, db cleared): upload using remote changes works

app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -63,4 +65,8 @@ object ApplicationConstants {
         // no wiki entry, sounds like it could span large areas
         "power", "pipeline", "railway"
     )
+
+    val EDIT_ACTIONS_NOT_ALLOWED_TO_USE_LOCAL_CHANGES = setOf(
+        SplitWayAction::class
```
westnordost, line 70:
```suggestion
        /* because this action may edit route relations but route relations are not persisted 
           locally for performance reasons */
        SplitWayAction::class
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
```diff
@@ -1,39 +1,65 @@
...
+        val changesetId = if (newChangeset) changesetManager.createChangeset(edit.type, edit.source)
+            else changesetManager.getOrCreateChangeset(edit.type, edit.source)
+        return mapDataApi.uploadChanges(changesetId, mapDataChanges)
+    }
+
+    private fun ElementEdit.fetchElement(mapDataRepository: MapDataRepository) = when (elementType) {
+        ElementType.NODE     -> mapDataRepository.getNode(elementId)
+        ElementType.WAY      -> mapDataRepository.getWay(elementId)
+        ElementType.RELATION -> mapDataRepository.getRelation(elementId)
     }
```
westnordost, line 64: This is actually a weird choice for an extension function. Maybe I wrote this when I was new to extension functions. Now, I would write it like this:

```suggestion
    private fun MapDataRepository.get(elementType: ElementType, elementId: Long) = when (elementType) {
        ElementType.NODE     -> getNode(elementId)
        ElementType.WAY      -> getWay(elementId)
        ElementType.RELATION -> getRelation(elementId)
    }
```

Would you agree that this is better? (Or meinetwegen ElementEdit as parameter)

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
```diff
@@ -1,39 +1,65 @@
...
+                uploadChanges(edit, remoteChanges, true)
+            } else {
+                // anything of the 3 may be the problem, try again with remote changes
+                try {
+                    uploadChanges(edit, remoteChanges, false)
+                } catch (e: ConflictException) {
+                    // probably changeset closed
+                    uploadChanges(edit, remoteChanges, true)
+                }
+            }
```
westnordost, line 50: The control flow is actually (completely) different whether `useRemoveChanges` is true or false. I think it would make this easier to read if the `if(useRemoteChanges) ...`  was moved to the top level.

westnordost (approved):
Looks good! A few comments

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
Helium314: Right, this looks better.
The way I had is was (probably, can't remember) just because it's least changes from original `ElementEdit.fetchElement`.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
Helium314: done

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
Helium314: I renamed `get` to `fetch`, so conflicts with `MapDataController.get` are avoided

westnordost (approved):

westnordost:
Did the upload feel faster by the way? For the changelog it would be nice to know how much of an improvement this really is (roughly)

Helium314:
I didn't really notice the difference... though I did some logging, and time for uploading was reduced by somthing like 80-100 ms on average.

But I usually have auto-upload off, so time between consecutive uploads is mostly limited by `rebuildLocalChanges`, which is slower than upload if there are more than ~20 edits waiting (depends on phone performance).

PR #4203, opened 2022-07-12
Make statistics menu less sad - decorations and colour

### LINKED ISSUE #3414: Make statistics menu less sad - more interesting display of data (colours? patterns?)
related:  #3230 proposing to show how statistics changed as user enters the screen

Right now statistics menu has gray circles

It would be nice to experiment with

- colours, especially for highly ranked stats
- some extra decoration - maybe great blobs can sprout some leaves or
flowers?
- maybe graphs? Some people are absurdly motivated by graphs and like
graphs of everything (yes, I describe myself)

I am planning to do something with that, but not right now ( https://github.com/streetcomplete/StreetComplete/issues/assigned/matkoniecz ). So if someone is interested, feel free to steal ideas from this.

PS I though also about sounds but I have quite significant aversion to apps having sounds of any kind, so I am not fan of adding anything of this kind. Also, have no idea where to find suitable sounds on a suitable license or how to produce them.

matkoniecz:
fixes #3414

screen05

-------------------

example with extra placeholder to demonstrate styling

screen01
screen02
screen03
screen04


westnordost:
1. Could you put the laurel wreath into an own file in `/view`?
2. The drawing of the stalk is needlessly complicated (and looks a bit odd because there are so many corners). To draw an arc is actually really simple, see https://developer.android.com/reference/android/graphics/Canvas#drawArc(float,%20float,%20float,%20float,%20float,%20float,%20boolean,%20android.graphics.Paint). But actually, better than that would be if you just skipped the stalk and used a different drawable for the leaf, for example something like this: https://commons.wikimedia.org/wiki/File:Olivenkranz.png

matkoniecz:
For now I applied `drawArc`

screen07

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
```diff
@@ -0,0 +1,94 @@
...
+/*
+Drawable providing decoration, suitable for a circular background
+100 and more: fully grown wreath with all pretty elements
+99 to 10: may be losing elements as it gets smaller
+below: no decorative styling at all
+ */
+class LaurelWreath(val resources: Resources, private val percentageOfGrowth: Int) : Drawable() {
+    val laurelLeafOnStalk = resources.getBitmapDrawable(R.drawable.ic_laurel_leaf_rotated)
+    val laurelStalk = resources.getBitmapDrawable(R.drawable.ic_laurel_leaf_stalk)
+    val horizontalEndingLeaf = resources.getBitmapDrawable(R.drawable.ic_laurel_leaf_ending)
```
westnordost, line 24: should be private

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
```diff
@@ -0,0 +1,94 @@
...
+
+            var bitmap = laurelLeafOnStalk.bitmap
+            if (i == shownSegments) {
+                bitmap = horizontalEndingLeaf.bitmap
+            }
+            val stalkPaint = Paint()
+            stalkPaint.isAntiAlias = true
+            stalkPaint.setARGB(255, 0, 106, 0)
+            stalkPaint.style = Paint.Style.STROKE
+            stalkPaint.strokeWidth = 3f
```
westnordost, line 62: should be a private field. Doesn't Android Studio warn about new object allocations in the draw method?

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
```diff
@@ -0,0 +1,94 @@
...
+        isFilterBitmap = true
+    }
+
+    override fun draw(canvas: Canvas) {
+        // Get the drawable's bounds
+        val canvasWidth: Int = bounds.width()
+        val canvasHeight: Int = bounds.height()
+        val circleRadius: Float = Math.min(canvasWidth, canvasHeight).toFloat() / 2f
+
+        val niceSubtleGreen: Paint = Paint().apply { setARGB(255, 152, 184, 126) }
```
westnordost, line 37: should be a private field

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
```diff
@@ -0,0 +1,94 @@
...
+import de.westnordost.streetcomplete.ktx.flipHorizontally
+import de.westnordost.streetcomplete.util.ktx.getBitmapDrawable
+
+/*
+Drawable providing decoration, suitable for a circular background
+100 and more: fully grown wreath with all pretty elements
+99 to 10: may be losing elements as it gets smaller
+below: no decorative styling at all
+ */
+class LaurelWreath(val resources: Resources, private val percentageOfGrowth: Int) : Drawable() {
```
westnordost, line 21: drawables usually follow the naming convention `....Drawable`

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
```diff
@@ -0,0 +1,94 @@
...
+            stalkPaint.isAntiAlias = true
+            stalkPaint.setARGB(255, 0, 106, 0)
+            stalkPaint.style = Paint.Style.STROKE
+            stalkPaint.strokeWidth = 3f
+            val stalkCircleSize = circleRadius * locationBetweenCenterAndEdge * 1.1f
+            val frame = RectF(canvasWidth.toFloat() / 2 - stalkCircleSize, canvasHeight.toFloat() / 2 - stalkCircleSize, canvasWidth.toFloat() / 2 + stalkCircleSize, canvasHeight.toFloat() / 2 + stalkCircleSize)
+            val stalkLengthInDegrees = 180f * percentageOfGrowth / 100f - 10f
+            canvas.drawArc(frame, 90f, stalkLengthInDegrees, false, stalkPaint )
+            val upperLimitOfRightStalk = (90f - stalkLengthInDegrees + 360) % 360
+            canvas.drawArc(frame, upperLimitOfRightStalk, stalkLengthInDegrees, false, stalkPaint )
```
westnordost, line 68: drawing one arc for every segment seems unnecessarily complicated. You could just draw **one** arc.

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
```diff
@@ -0,0 +1,94 @@
...
+            // left side
+            canvas.withRotation(i * 180.0f / regularSegmentCount, canvasWidth / 2f, canvasHeight / 2f) {
+                // drawBitmap takes corner of the bitmap, we care about centering segments
+                canvas.drawBitmap(bitmap, circleCenterX - decorationSegmentImageWidth / 2f, canvasHeight * locationBetweenCenterAndEdge, antiAliasPaint)
+            }
+
+            // right side
+            val flippedBitmap = bitmap.flipHorizontally()
```
westnordost, line 76: FYI you can also apply a matrix transformation (to flip the bitmap) on drawing on the canvas, might be easier.

https://developer.android.com/reference/android/graphics/Canvas#drawBitmap(android.graphics.Bitmap,%20android.graphics.Matrix,%20android.graphics.Paint)

(But you are not exactly drawing thousands of wreaths each frame, so perfomance of creating dozens of intermediate bitmaps is no issue)

app/src/main/java/de/westnordost/streetcomplete/view/LaurelWreath.kt
matkoniecz: Are you sure that it should be moved out of that function?

westnordost: I am just parroting what the IDE warns about. Android Studio warns about creation of objects in the `draw` function for performance reasons. Since this paint is always the same, it also makes sense.

matkoniecz: > It's weird that your Android studio didn't display a warning there? Did you switch off warnings?

I switched off some that were consistently annoying and useless. I can enable them again, and triage them again - maybe they improved something.

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
```diff
@@ -164,15 +176,19 @@ class ProfileFragment : Fragment(R.layout.fragment_profile) {
             val shouldShow = statistics.rank != null && statistics.rank > 0 && statistics.count > 50
             val countryLocale = Locale("", statistics.countryCode)
             binding.localRankContainer.isGone = !shouldShow
-            binding.localRankText.text = "#${statistics.rank}"
-            binding.localRankLabel.text = getString(R.string.user_profile_local_rank, countryLocale.displayCountry)
+            if (shouldShow) {
+                binding.localRankText.text = "#${statistics.rank}"
+                binding.localRankLabel.text = getString(R.string.user_profile_local_rank, countryLocale.displayCountry)
+                binding.localRankText.background = LaurelWreathDrawable(resources, min(100 - statistics.rank!!, 100))
```
westnordost, line 182: uhh, can statistics.rank be null?

westnordost (commented):
Looks good, one small remark

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
matkoniecz: in such case https://github.com/streetcomplete/StreetComplete/pull/4203/files/91a1e78eda50992bad7a3ce4f4b2bbc236aa82a5#diff-a4897d401da7827ccc0867cd49c938294ee4c49d795d6a3173017c9ecbcf03b9L164 will set `shouldShow` to false

westnordost (approved):

PR #4204, opened 2022-07-13
Fix Attached GPS Tracks

goldbattle:
Enforce that GPS tracks should not be added if the note was not created after a track recording.
Should close issue #4199


westnordost:
Great, thank you!

So, the issue was that
1. the last recorded track is never cleared from memory and
2. the create-note-form would just take that last recorded track to attach to a note, **assuming** the last recorded track is cleared from memory after attaching

But that assumption is wrong. So you changed it so that the create-note-form does not assume that anymore but instead checks whether the user opted to attach a track or not.

PR #4208, opened 2022-07-14
Fix the GPX filename timestamps

peternewman:
Entirely untested, but strongly assumed based on https://github.com/streetcomplete/StreetComplete/commit/47447504ea8cf40fff8b2cce6b6b14148a43559e and the fact it's currently clearly wrong:
https://www.openstreetmap.org/traces/tag/streetcomplete/

And a bit of maths :nerd_face:


peternewman:
@goldbattle do you want to take a look?

westnordost:
What? Sorry, you need to start testing your own PRs.

westnordost:
Anyway, I looked through the code and looked where the Trackpoints are created, they are created from https://developer.android.com/reference/android/location/Location#getTime() which returns the time in milliseconds, so indeed the time of a Trackpoints is in milliseconds. 

So, thank you for noticing and providing a fix.

But please note that I now spent 20-30 minutes of my time for StreetComplete to reconstruct this and write this message, time that I could have spent on other things regarding StreetComplete and that would have been avoided if you tested it and provided some kind of proof that your assumption was correct. That's also more time than it would have taken if an issue was posted and I fixed it myself.

PR #4213, opened 2022-07-14
Camping site quests

mnalis:
This implements some (but not all) of requested quests for [tourism=camp_site](https://wiki.openstreetmap.org/wiki/Tag:tourism=camp_site), thus it mostly fixes https://github.com/streetcomplete/StreetComplete/issues/4198:

- [x] `drinking_water=*` `yes/no` quest
- [x] `power_supply=*` `yes/no` quest
- [x] `shower=*` `yes/no` quest
- [x]  3-answer quest `Who may camp here?` ( `Tents only` / `Caravan only` / `Both`) for setting [tents=*](https://wiki.openstreetmap.org/wiki/Key:tents) and [caravans=*](https://wiki.openstreetmap.org/wiki/Key:caravans) (with OtherAnswer available to set [backcountry=yes](https://wiki.openstreetmap.org/wiki/Key:backcountry))
- [ ] create non-placeholder icons

**Help needed with icons**: This PR reuses icon for `ic_quest_drinking_water` (which might be OK as it is, or it might benefit from being separate icon, e.g. if all icons have tent+some_image), but has only **_placeholder_** icons for `ic_quest_camp_power_supply`, `ic_quest_camp_shower` and `ic_quest_camp_type`. Artistically inclined people are invited to fix those!


matkoniecz:
It is in the draft state - do you expect review of what is prepared?

Or is it request for help with some part (which one)?

mnalis:
> It is in the draft state - do you expect review of what is prepared?
> Or is it request for help with some part (which one)?

Thanks @matkoniecz; I've just wanted to give interested people heads up and some extra time (especially as I have to wrap this one up _very soon_) to directly suggest better wording and perhaps make other suggestions while I work on rest of it. As soon as I implement all four parts (marked by checkboxes) and do basic testing, I'll mark it as ready for review.

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,10 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_camera_type_fixed">Fixed</string>
     <string name="quest_camera_type_panning">Panning</string>
 
+    <string name="quest_camp_drinking_water_title">"Does it have drinking water?"</string>
```
matkoniecz, line 845: what in case where there is shop selling bottled water?

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,10 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
+    <string name="quest_camp_shower_title">"Does it have showers?"</string>
+    <string name="quest_camp_power_supply_title">"Can you charge your phone/laptop here?"</string>
```
matkoniecz, line 847: what if only camp residents can charge? 

what if only some camp residents can charge? Say, only one with vans.

what if it is technically possible but extremely inaccessible (single connector in a bathroom, intended to be used for hair dryer)

what if you can charge phone (USB connector) but not laptop?

app/src/main/res/values/strings.xml
mnalis: If water was otherwise unavailable, I would leave a note and indicate a shop there in that case (so no, I don't think existence of a convenience shop qualifies as `drinking_water=yes`)
Perhaps `Does it have freely available drinking water` would be better question (even if longer)?

app/src/main/res/values/strings.xml
mnalis: I'm not quite sure what you mean by `what if only camp residents can charge`? Surely it is implied that only people having acquired legal access to camp may use its facilities (including power sockets)?

Perhaps `Can anyone charge their phone/laptop here?` would be better question? Or `Can any customer charge their phone/laptop here?` ?

As for the others, I have charged my phone in exactly such low-accessible places. Those are camps, not private hotel rooms - inconvenience with access to facilities - including power - is implied (e.g. in most smaller camps it is only one power socket / strip in common room, or socket in bathroom. That's why I always carry my own power strip, and multi-port USB charger to avoid fighting for scarce resources. Makes extra friends of the trip too :smile:). 
And if there is such strange situation (only DC USB port, no AC sockets), I'd expect user to leave a note (I've never encountered such situation myself though).

matkoniecz: I encountered case where there was one working socket, in bathroom stall (not very practical for charging :) )

+1 to `Can any customer charge their phone/laptop here?`

app/src/main/res/values/strings.xml
mnalis: As in other thread, I would assume all facilities on camp site are available only to those who have legally obtained access to that camp site (e.g. payed for their say if usual `fee=yes`). Anyway, would `Is drinking water available to all customers here?` sound better to you (akin to other thread question)? If not, feel free to suggest change for better wording.

app/src/main/res/values/strings.xml
mnalis: Applied `Can any customer charge their phone/laptop here?`

mnalis:
Tested and looks to work OK to me, so ready for review. 
It needs replacing placeholder icons however, which needs someone artistically inclined (as noted in updated PR summary).

mnalis:
(anyway, just a notice that I won't be available to give github attention for some time, so feel free to update/edit/merge this PR at will if you'd rather not wait for me)

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampDrinkingWater.kt
```diff
@@ -0,0 +1,37 @@
...
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddCampDrinkingWater : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
+        nodes, ways with (
+          tourism=camp_site
+        )
+        and (!drinking_water or drinking_water older today -4 years)
```
FloEdelmann, line 19:
```suggestion
        nodes, ways with
          tourism = camp_site
          and (!drinking_water or drinking_water older today -4 years)
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampPower.kt
```diff
@@ -0,0 +1,39 @@
...
+class AddCampPower : OsmFilterQuestType<Boolean>() {
+
+    /* We only resurvey power_supply=yes and power_supply=no, as it might have more detailed 
+     * values from other editors, and we don't want to damage them */
+    override val elementFilter = """
+        nodes, ways with (
+          tourism=camp_site
+        )
+        and (!power_supply or (power_supply older today -4 years and power_supply ~ yes|no))
+    """
```
FloEdelmann, line 22:
```suggestion
    /* We only resurvey power_supply = yes and power_supply = no, as it might have more detailed 
     * values from other editors, and we don't want to damage them */
    override val elementFilter = """
        nodes, ways with
          tourism = camp_site and (
            !power_supply
            or power_supply older today -4 years and power_supply ~ yes|no
          )
    """
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampShower.kt
```diff
@@ -0,0 +1,39 @@
...
+class AddCampShower : OsmFilterQuestType<Boolean>() {
+
+    /* We only resurvey shower=yes and shower=no, as it might have more detailed 
+     * values from other editors, and we don't want to damage them */
+    override val elementFilter = """
+        nodes, ways with (
+          tourism=camp_site
+        )
+        and (!shower or (shower older today -4 years and shower ~ yes|no))
+    """
```
FloEdelmann, line 22:
```suggestion
    /* We only resurvey shower = yes and shower = no, as it might have more detailed 
     * values from other editors, and we don't want to damage them */
    override val elementFilter = """
        nodes, ways with
          tourism = camp_site and (
            !shower
            or shower older today -4 years and shower ~ yes|no
          )
    """
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
```diff
@@ -0,0 +1,42 @@
...
+        nodes, ways with (
+          tourism=camp_site
+        )
+        and (!caravans or !tents)
+        and !backcountry
+    """
+    override val changesetComment = "Survey who may camp here"
+    override val wikiLink = "Key:caravans"
+    override val icon = R.drawable.ic_quest_camp_type
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
```
FloEdelmann, line 24: Maybe the `AddCampType` quest doesn't need to be disabled by default, since you can often see caravans and/or tents from the outside?

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampDrinkingWater.kt
```diff
@@ -0,0 +1,37 @@
...
+    override val changesetComment = "Specify whether there is drinking water in camp site"
+    override val wikiLink = "Key:drinking_water"
+    override val icon = R.drawable.ic_quest_drinking_water
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+    override val achievements = listOf(OUTDOORS)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_camp_drinking_water_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with tourism=camp_site")
```
FloEdelmann, line 30:
```suggestion
        getMapData().filter("nodes, ways with tourism = camp_site")
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampPower.kt
```diff
@@ -0,0 +1,39 @@
...
+    override val changesetComment = "Specify whether there is electricity available at camp site"
+    override val wikiLink = "Key:power_supply"
+    override val icon = R.drawable.ic_quest_camp_power_supply
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+    override val achievements = listOf(OUTDOORS)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_camp_power_supply_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with tourism=camp_site")
```
FloEdelmann, line 32:
```suggestion
        getMapData().filter("nodes, ways with tourism = camp_site")
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampShower.kt
```diff
@@ -0,0 +1,39 @@
...
+    override val changesetComment = "Specify whether there are showers available at camp site"
+    override val wikiLink = "Key:shower"
+    override val icon = R.drawable.ic_quest_camp_shower
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+    override val achievements = listOf(OUTDOORS)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_camp_shower_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with tourism=camp_site")
```
FloEdelmann, line 32:
```suggestion
        getMapData().filter("nodes, ways with tourism = camp_site")
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val achievements = listOf(OUTDOORS)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_camp_type_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with tourism=camp_site")
```
FloEdelmann, line 30:
```suggestion
        getMapData().filter("nodes, ways with tourism = camp_site")
```

FloEdelmann (commented):
Looks good :)
I have some quest filter formatting suggestions (to be consistent with existing quest filters) that I'll apply directly; please feel free to overwrite them again if they're not perfect.

I'm not approving, since real icons are still missing.

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
peternewman: Also shouldn't it have a slightly different message because you don't have to go "inside" somewhere to check it?

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_camera_type_fixed">Fixed</string>
     <string name="quest_camera_type_panning">Panning</string>
 
+    <string name="quest_camp_drinking_water_title">Is drinking water available to all customers here?</string>
+    <string name="quest_camp_shower_title">Does it have showers?</string>
+    <string name="quest_camp_power_supply_title">Can any customer charge their phone/laptop here?</string>
+    <string name="quest_camp_type_title">Who may camp here?</string>
+    <string name="quest_camp_type_tents_and_caravans">Both tents and caravans</string>
+    <string name="quest_camp_type_caravans_only">Only caravans</string>
```
peternewman, line 850: You might also need to try and differentiate from static caravans/mobile homes.

peternewman (changes requested):
Some improvements/suggestions and comments.

app/src/main/res/drawable/ic_quest_camp_power_supply.xml
```diff
@@ -0,0 +1,22 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
westnordost: mnalis wrote

>  It needs replacing placeholder icons however, which needs someone artistically inclined (as noted in updated PR summary).

app/src/main/java/de/westnordost/streetcomplete/quests/camping/CampType.kt
```diff
@@ -0,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.camping
+
+enum class CampType(val tents: Boolean, val caravans: Boolean) {
+    TENTS_AND_CARAVANS(true, true),
+    CARAVANS_ONLY(false, true),
+    TENTS_ONLY(true, false),
+    BACKCOUNTRY(false, false),
```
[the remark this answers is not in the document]
mnalis: yes backcountry supports tents, but proposed `(true,false)` already defines a `TENTS_ONLY` value, and this value needs to be **_unique_** or it won't work correctly. I could make the class take three booleans instead (`tents`, `caravans`, `backcountry`) and modify the rest of the code appropriately, but that is IMHO an overkill for such minor improvement in clarity?

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -399,6 +403,12 @@ fun questTypeRegistry(
 
     AddIsDefibrillatorIndoor(), // need to go inside in case it is inside (or gone)
 
+    // inside camping sites
+    AddCampType(),
+    AddCampDrinkingWater(),
+    AddCampShower(),
+    AddCampPower(),
```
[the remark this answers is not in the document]
mnalis: I'd prefer to leave them as-is. If we later want to reuse them, we can easily rename them then (we should also then move them to another directory, put in another class etc.)

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_camp_type_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("nodes, ways with tourism = camp_site")
+
+    override fun createForm() = AddCampTypeForm()
+
+    override fun applyAnswerTo(answer: CampType, tags: Tags, timestampEdited: Long) {
+        if (answer.tents || answer.caravans) {
```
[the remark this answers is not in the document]
mnalis: I closed that suggestion, so it doesn't matter anymore :smiley:

I wouldn't know how to do in Kotlin anyway; I though those class elements were defined by the boolean pair specified? (i.e. same pair of boolean on duplicate elements make those elements aliases)
But if you know how to do it more correctly so it works, feel free to suggest a code change.

app/src/main/res/values/strings.xml
mnalis: Uh, do you mean as in translation issue, and if so, do you have suggestions for better English (US) description?

https://wiki.openstreetmap.org/wiki/Key:caravans defines it as:
> "Caravan" is a British English term for specially made vehicles that have space to sleep while traveling, and often include other amenities such as electricity, basic cooking facilities and water supply. The caravan might be a trailer towed by another vehicle (ie travel trailer, camping trailer, popup camper), or a larger motorized vehicle (recreational vehicle (RV) or motorhome). 

As for more detailed mapping (e.g. [static_caravans=*](https://wiki.openstreetmap.org/wiki/Key:static_caravans), [permanent_camping=*](https://wiki.openstreetmap.org/wiki/Key:permanent_camping), [cabins=*](https://wiki.openstreetmap.org/wiki/Key:cabins) etc. I do not plan to implement them at this time; feel free to create additional PR after this one is accepted if you feel the additional tags are important enough.

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
mnalis: ok, I've made it enabled by default with comment in 270a6947c. We'll see if it is very disliked at final review.

@peternewman I think `default_disabled_msg_go_inside` is good enough IMHO; I don't think there is no need to introduce special new disabled categories with just a minor difference in meaning.

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
+    <string name="quest_camp_type_tents_only">Only tents</string>
+    <string name="quest_camp_type_backcountry">Backcountry camping (no facilities)</string>
```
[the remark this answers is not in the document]
mnalis: Putting it in the main answer section would IMHO create more confusion than it would help.

Also, it's not a big loss if it is not found - it mostly just changes implied defaults / skips the extra quests (see https://github.com/streetcomplete/StreetComplete/issues/4198#issuecomment-1179793163 and below) - initially I wasn't even planning to include it.

mnalis:
Hopefully I've addressed all comments/suggestions here. If so, all that remains is replacing placeholder icons. 

**Is anyone artistically inclined interested in creating real icons instead of placeholder ones?**

If nobody answers in a next two days, I can try creating icons to make this PR mergeable:
- camp type - https://commons.wikimedia.org/wiki/File:Noun_camp_2695642.svg
- shower - https://commons.wikimedia.org/wiki/File:Shower_icon_-_Noun_Project_4336.svg
- power - reuse existing icon
- water - already reusing existing icon

But I'd prefer not to waste time on that if they're going to be overwritten with something else soon after :smiling_face_with_tear:

app/src/main/res/drawable/ic_quest_camp_type.xml
```diff
@@ -1,16 +1,9 @@
-<vector
-    android:height="128dp"
-    android:viewportHeight="33.86667"
-    android:viewportWidth="33.86667"
-    android:width="128dp"
-    xmlns:android="http://schemas.android.com/apk/res/android">
+<vector xmlns:android="http://schemas.android.com/apk/res/android" xmlns:aapt="http://schemas.android.com/aapt"
```
matkoniecz, line 1: note that icon appears to be partially transparent - is it intentional and working well?

matkoniecz:
@mnalis Have you maybe tested this in your own SC version?

app/src/main/res/drawable/ic_quest_camp_type.xml
mnalis: Not particularly intentional, no. It's just the way icon came. Seems to be working just fine in menu and on map, though. Perhaps some other color than white would look nicer, if someone with inkscape skills could help (my attempt seem to create "borders" of some kind between the tent and background).

small_Screenshot_20220803_141451_de westnordost streetcomplete mn debug small_Screenshot_20220803_141445_de westnordost streetcomplete mn debug

mnalis:
> @mnalis Have you maybe tested this in your own SC version?

yes, this one: https://github.com/mnalis/StreetComplete/actions/runs/2789132496, see pics above.
I create android .xml from .svg with https://svg2vector.com/ (as I lack Android Studio environment).

Still needs replacing few other placeholder icons, and maybe making this one more colorful? 
If anyone wants to help, I'd be very glad for any help or suggestions with icons! In the meantime I try to help push this more near completion (although vector graphics is away from my comfort zone) to reduce to load on main dev...

matkoniecz:
I mostly though about actual use while mapping.

mnalis:
Yes, I've just done a camping bicycle route around lake Balaton with SC camping quest. I've had no problems with answering quests, but I'm of course biased in that regard as I wrote the quest...

> I tested in on campground familiar to me and I remain unsure whether I should answer yes or no for drinking water question in case where there is fast food pizzeria on camp grounds where you may buy water.

would this change help?
```diff
- Is drinking water available to customers here?
+ Is drinking water freely available to customers here?
```

Most any convenience shop/supermarket/cafe/restaurant/etc. will be able to sell you bottled water, I do not think that has anything to do with [drinking_water=yes](https://wiki.openstreetmap.org/wiki/Key:drinking_water) OSM tagging.

> And what to answer when you can charge phone but you need to hold it, as otherwise it will fall into sink (or have really, really long power extender to get it out of bathroom)

Well, the question being asked is: 
> **_Can_** customers charge their phone/laptop here?

and not _"**How convenient** is it for customers to charge their phone/laptop here?"_.

To me, it is a clear `yes` - **they can**. The convenience however ranges wildly (and AFAIK has no clear OSM mapping consensus for mapping such convenience, except in freeform `description=*` tag of course):
- there might be power outlets near many tent spaces (although they might be all used - quite common situation, so better have extension power cord with you! Also, even if socket happens to be very very close you can rarely if ever put a charger in your tent - unless you use previously mentioned own extension cord)
- there might be one (or few) separate space with one or two powers sockets (so better have extension cord with you, especially if going camping during peak season). You have to decide whether you're comfortable leaving them there unattended or if you're going to stand near it for hours (that risk/convenience problem generally applies to camping - what do you with wallet/phone when you go swimming etc.)
- there might be only a socket or two in bathrooms, and there may or may not be any shelves to conveniently put the phone on). I've done all of the following: had an extension cord which allowed leaving phone on the floor or sink, had a phone in my hand (USB-C fast chargers help a million!), had a flat charger and manually balanced a phone to lay on top of it, had a strong USB cable and having a phone dangle from charger on it like a fishing bait... Still, they all charged the phone.
- there might be no power at all. So you better carry battery backup in this case if it's important to you (alas, gone are the days when phones consumed little energy and battery banks had ability to charge from solar power). Only this case is `no` answer IMO, all others are `yes`.

My point is: **_camping has many inconveniences_** associated with it, and any person camping more than first time will **_surely be aware_** of most of them. And this quest is intended only for campers i.e. people which have payed to enter the camp (and is thus disabled by default). 

I would not think any camper expects to have the convenience of their own private power socket (or water faucet or toilet for that matter!) installed in their tent (like they would have if for example they rented a hotel room).

Perhaps putting Other Answer with _"Yes, but it is inconvenient"_ which open alert window with _"Well, DOH, rent a hotel room next time instead"_ :smile: or something less sarcastic like _"If it is possible to charge the phone, you should answer `yes`, no matter of convenience"_ might help? But I'm dubious.

matkoniecz:
`Is drinking water freely available to customers here?` - that should work great!

> The convenience however ranges wildly (and AFAIK has no clear OSM mapping consensus for mapping such convenience

If OSM tagging is not distinguishing between this cases and all should be tagged the same then it should be fine, I guess.

app/src/main/res/drawable/ic_quest_camp_type.xml
matkoniecz: > Seems to be working just fine in menu and on map, though

Yes, as it shown on top of nontransparent white pin, so it works out fine.

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val elementFilter = """
+        nodes, ways with
+          tourism = camp_site
+          and (!caravans or !tents)
+          and !backcountry
+    """
+    override val changesetComment = "Survey who may camp here"
+    override val wikiLink = "Key:caravans"
+    override val icon = R.drawable.ic_quest_camp_type
+    // you can often see caravans and/or tents from the outside, so usually there is no need for this quest to be disabled by default
```
matkoniecz, line 23: but lack of caravans right now is not answering question whether caravans are allowed or not

res/graphics/authors.txt
```diff
@@ -138,6 +138,9 @@ building levels/
     building_levels_icon.svg
     building_levels_illustration.svg
 
+camping/
+    camp.svg                    https://commons.wikimedia.org/wiki/File:Noun_camp_2695642.svg, Bluerasberry CC-BY 4.0
```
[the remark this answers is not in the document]
matkoniecz: note that Bluerasberry is uploader to Commons, while artwork authorship is credited to [sbts2018](https://thenounproject.com/sbts2018/)

screen26

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
peternewman: A bit like the seating quests (or pedestrian crossing audible indication), answering about presence is far easier than absence!

app/src/main/res/values/strings.xml
peternewman: > I'm not quite sure what you mean by `what if only camp residents can charge`? Surely it is implied that only people having acquired legal access to camp (i.e. _"camp residents"_?) may use its facilities (including power sockets)?

In the UK at least it's quite common for there to be hookups for caravans (on blue Ceeforms), you may need to pay for those or they are only for caravan spots. It used to be quite common for there to only be those or a shaver socket in the bathroom, with no sensible ability to charge a phone/laptop via a domestic socket. I think that's becoming less common now because most mobiles need charging daily.

The detail on the wiki is pretty vague:
https://wiki.openstreetmap.org/wiki/Key:power_supply
https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dcamp_site#Tagging

But I'm sure both caravan users and cyclists without an adaptor lead will be frustrated if they turn up to a site and find they can't plug in as someone has tagged it in the opposite way to what they were expecting despite it answering the SC quest question.

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
peternewman: > ok, I've made it enabled by default with comment in [270a694](https://github.com/streetcomplete/StreetComplete/commit/270a6947c5f9c7f33fce73ab42546ebc42fa3914). We'll see if it is very disliked at final review.

I (and @matkoniecz ) aren't entirely convinced by that.

> @peternewman I think `default_disabled_msg_go_inside` is good enough IMHO; I don't think there is no need to introduce special new disabled categories with just a minor difference in meaning.

Yeah as mentioned the message is more general than the string name.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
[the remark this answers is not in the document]
peternewman: Okay, it seems a shame to not ask e.g. the shower (probably less likely as in your vehicle) and power quests (probably more likely as you need to plug in) at caravan sites from the get go.

app/src/main/java/de/westnordost/streetcomplete/quests/camping/CampType.kt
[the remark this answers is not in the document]
peternewman: Personally I would suggest that means your code needs altering; I don't think we should be recording wrong information (`tents=no`) just to simplify the code. As I say, I think backcountry is orthogonal to tents/caravans, so it should either only set backcountry, or set tents too (if that can be guaranteed at all backcountry sites).

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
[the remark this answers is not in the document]
peternewman: It looks like the enum won't care if they're duplicate values:
https://discuss.kotlinlang.org/t/ensuring-uniqueness-of-enum-constants/16661

And you can match on the enum itself rather than the value if you need to.

app/src/main/res/values/strings.xml
peternewman: I'm concerned about someone seeing a static caravan on-site and the word caravan and picking that answer, despite it potentially not being possible to turn up with a caravan/RV. I'm the wrong side of the pond to confirm whether clarity is needed for the US people, but I'd suggest the string file wants a comment at the minimum so it isn't miss-translated by others.

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_camera_type_fixed">Fixed</string>
     <string name="quest_camera_type_panning">Panning</string>
 
+    <string name="quest_camp_drinking_water_title">Is drinking water freely available to customers here?</string>
+    <string name="quest_camp_shower_title">Are there showers here?</string>
+    <string name="quest_camp_power_supply_title">Can customers charge their phone/laptop here?</string>
```
matkoniecz, line 847: Maybe 

> Is there any way for customers to charge their phone or laptop here?

to make more clear that even poorly accessible charging counts? (not a native speaker, not entirely sure)

matkoniecz: I think that it is almost ready for final review by `@westnordost` - though maybe you may want to reject/use/process this idea first.

mnalis: thanks, that sounds reasonable translation, I've updated it.

mnalis:
> (Just ping me if this is ready for final review. I haven't been following this PR. As usual, icons are not required, I'll make them if they are missing.)

Thanks @westnordost, I've tried playing with icons but got stuck at transparency/fill issues, so I'd appreciate if you can take care of icons when you do final review!

Other than that, I think it's as finished as I could make it. 
Two more things need decision, though:

- Is "caravan" good enough US translation in order not to be confused with static caravan? Or should it be changed or some string translation hints be added (and where?)  - see https://github.com/streetcomplete/StreetComplete/pull/4213#pullrequestreview-1061503670
- should the camp_type quest be enabled or disabled by default, there have been arguments for both, so it's a matter of commenting or uncommenting the line - see https://github.com/streetcomplete/StreetComplete/pull/4213#pullrequestreview-1061503670

mnalis:
> recently I was talking with friends and it turns out that in some regions even "is this camping site paid" maybe would be a viable quest

In my limited experience of near-to-Croatia parts of Europe it does seem spammy (see https://github.com/streetcomplete/StreetComplete/issues/4198#issuecomment-1179793163), but I guess it might differ worldwide. If `camp fee` quest would be useful in some countries (and `backcountry` answer as currently implemented is insufficient), I think another issue/PR should be opened, as this one if overstuffed already (probably would need some research where it would not be spammy).

peternewman:
> (and `backcountry` answer as currently implemented is insufficient), I think another issue/PR should be opened, as this one if overstuffed already

I strongly feel this PR shouldn't be merged with the backcountry answer in its current state; leaving a note when required occasionally would be better than the current other answer which can result in incorrect tagging. Looking at taginfo, more than a quarter of `backcountry=yes` are also tagged with `tents=yes`, so setting the opposite for one when we tag the other seems crazy to me:
https://taginfo.openstreetmap.org/tags/backcountry=yes#combinations

mnalis:
> I strongly feel this PR shouldn't be merged with the backcountry answer in its current state; leaving a note when required occasionally would be better than the current other answer which can result in incorrect tagging. Looking at taginfo, more than a quarter of `backcountry=yes` are also tagged with `tents=yes`, so setting the opposite for one when we tag the other seems crazy to me: https://taginfo.openstreetmap.org/tags/backcountry=yes#combinations

@peternewman I'm not sure of which incorrect tagging are you talking about, can you provide an example of what would be marked incorrectly in which case?
If user chooses `Backcountry camping (no facilities)` answer, it will be marked  **only** with `backcountry=yes` tag (and no other quests about facilities will be asked). 

I understand that it might be argued that some additional tags might be asked for backcountry camps too; but surely there is **no** _**incorrect** tagging_ if [tourism](https://wiki.openstreetmap.org/wiki/Key:tourism)=[camp_site](https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dcamp_site) gets tagged _only_ with additional `backcountry=yes`?

Or did I misunderstand your point?

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
```diff
@@ -0,0 +1,42 @@
+package de.westnordost.streetcomplete.quests.camping
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.OUTDOORS
+import de.westnordost.streetcomplete.osm.Tags
```
peternewman, line 9: Add a required import:
```suggestion
import de.westnordost.streetcomplete.osm.Tags
import de.westnordost.streetcomplete.quests.camping.CampType.BACKCOUNTRY
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/CampType.kt
```diff
@@ -0,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.camping
+
+enum class CampType(val tents: Boolean, val caravans: Boolean) {
+    TENTS_AND_CARAVANS(true, true),
+    TENTS_ONLY(true, false),
+    CARAVANS_ONLY(false, true),
+    BACKCOUNTRY(false, false),
```
peternewman, line 7: At the minimum I'd suggest this should have a comment added
```suggestion
    // Only the enum value is used here, the tent and caravan values are ignored.
    BACKCOUNTRY(false, false),
```

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    override fun createForm() = AddCampTypeForm()
+
+    override fun applyAnswerTo(answer: CampType, tags: Tags, timestampEdited: Long) {
+        if (answer.tents || answer.caravans) {
+            tags["tents"] = answer.tents.toYesNo()
+            tags["caravans"] = answer.caravans.toYesNo()
+        } else {
+            tags["backcountry"] = "yes"
+        }
```
peternewman, line 40: From:
https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrest.kt

To me this is less confusing and with less unintended consequences if the rest was tweaked in future.
```suggestion
        when (answer) {
            BACKCOUNTRY -> tags["backcountry"] = "yes"
            else -> {
                tags["tents"] = answer.tents.toYesNo()
                tags["caravans"] = answer.caravans.toYesNo()
            }
        }
```

You could potentially go as far as this:
https://stackoverflow.com/questions/32830904/use-of-boolean-in-if-expression

Then the backcountry one could be nulled entirely if it's not going to set anything and the others could just tag if not null.

peternewman (changes requested):
Some suggestions to clarify the backcountry behaviour

peternewman:
> @peternewman I'm not sure of which incorrect tagging are you talking about, can you provide an example of what would be marked incorrectly in which case? If user chooses `Backcountry camping (no facilities)` answer, it will be marked **only** with `backcountry=yes` tag

Apologies, I'd got caught out by your `BACKCOUNTRY(false, false),`, not checked the other code and assumed it was going to tag `tents=no` too. I've added some code suggestions for how to improve the clarity of that a lot (IMHO), including the all-important comment, and a fairly simple extension for how to potentially make it even better and remove the gotcha false values entirely.

> (and no other quests about facilities will be asked).

This bit is currently untrue looking at the quest filters, you'll still be asked about water, showers and power even if its tagged as backcountry (where they are less likely and its possibly spammy).

mnalis:
> Apologies, I'd got caught out by your `BACKCOUNTRY(false, false),`, not checked the other code and assumed it was going to tag `tents=no` too. I've added some code suggestions for how to improve the clarity of that a lot (IMHO), including the all-important comment, and a fairly simple extension for how to potentially make it even better and remove the gotcha false values entirely.

Thanks, I've added you code-readability improvements!

> This bit is currently untrue looking at the quest filters, you'll still be asked about water, showers and power even if its tagged as backcountry (where they are less likely and its possibly spammy).

Ah yes.  I've had a mind about disabling quests for them, but after more careful reading of https://wiki.openstreetmap.org/wiki/Key:backcountry I've decided to still ask for them, as it seems at least some of them are still possible (just less likely, i.e. assumed `no` by default). And as `backcountry=yes` is only a small subset of camps (according to [taginfo](https://taginfo.openstreetmap.org/tags/tourism=camp_site#combinations)) and camps are relatively rare themselves, I do not think continuing to ask for them would be too spammy. But if there is consensus to disable some of them for `backcountry=yes`, I'm not against.

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_camera_type_fixed">Fixed</string>
     <string name="quest_camera_type_panning">Panning</string>
 
+    <string name="quest_camp_drinking_water_title">Is drinking water freely available to customers here?</string>
```
westnordost, line 845:
```suggestion
    <string name="quest_camp_drinking_water_title">Is drinking water freely available to campers here?</string>
```

Customers sounds odd in the context of accommodation. I've checked with en-US and en-GB dictionaries - camper is fine to refer to a person that camps (it can also mean the caravan/motorhome, but context is clear here)

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
+    <string name="quest_camp_shower_title">Are there showers here?</string>
+    <string name="quest_camp_power_supply_title">Is there any way for customers to charge their phone or laptop here?</string>
```
westnordost, line 847:
```suggestion
    <string name="quest_camp_power_supply_title">Is there any way for customers to charge their phone, laptop etc. here?</string>
```

otherwise sounds a bit too specific (some laptops are now chargeable via USB)

app/src/main/res/values/strings.xml
```diff
@@ -842,6 +842,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
+    <string name="quest_camp_type_title">Who may camp here?</string>
+    <string name="quest_camp_type_tents_and_caravans">Both tents and caravans</string>
+    <string name="quest_camp_type_caravans_only">Only caravans</string>
```
westnordost, line 850:
```suggestion
    <string name="quest_camp_type_tents_and_caravans">Both tents and RVs</string>
    <string name="quest_camp_type_caravans_only">Only RVs</string>
```

That's British English. I've looked through merriam-webster defintions and this: https://familytravelfever.com/difference-rv-motorhome-trailer/
RV is a catch-all for both motorhomes (=camper vans) and travel trailers

app/src/main/java/de/westnordost/streetcomplete/quests/camping/AddCampType.kt
westnordost: Still, I think this is fine. Camp sites are open areas usually, so you don't really need to go "inside" the way you would do for shops etc.

westnordost (approved):
Nice, clean implementation. I have nothing else to add other than the tweaks on the wordings. I will do the icons after merging.

matkoniecz:
From not-a-native-speaker: I would have no idea what "RV" is, even with context.

westnordost:
Okay

riQQ:
Also as a non-native speaker: RV is clear for me.

westnordost:
:shrug: I changed it to "motorhomes, travel trailers, ..." or something

PR #4215, opened 2022-07-15
buildings type selection in one list

### LINKED ISSUE #3387: Building type quest - allow to access full list by scrolling, not by clicking "show more"
**Use case**
On mobile it is more natural to scroll list than to click on button.

**Proposed Solution**

Current: list with top answers and blue "show more" button. Pressing button replaces top values with a full list.

Proposed: list with top values and  full list below - accessible by scrolling, not by pressing button.

Seems clearly better to me (multiple people in testing tried to scroll and completely failed to notice big blue button). But I want to check before implementing anything.

matkoniecz:
fixes #3387 

Current: list with top answers and blue "show more" button. Pressing button replaces top values with a full list.

This: list with top values and full list below - accessible by scrolling, not by pressing button.

Seems clearly better to me (multiple people in testing tried to scroll and completely failed to notice big blue button). I also tested it by myself and it is much, much better.

Main negative is that I am not entirely happy about how I solved problem of selecting answer in the list where marked answer should be what is in category, not one of top answers before categories.

Maybe there is a smarter way to implement this? Maybe I should write own equivalent of RecyclerView?

Maybe it is actually fine as is?

screen01


matkoniecz:
Switching `indexOf` to `lastIndexOf` and adding `.reversed()` in getter, so that internal RecyclerView will work as expected (select item in the category, not the one in top entries) seems a bit ugly to me

> If you open a category, will the favourite answers be collapsed 

no, as it is not in category

I tried putting it into category, but category header wastes space of one entry without benefit, so I rewrote it

> If not, wouldn't that make sense

I think no - at least I sometimes get back to top answers, it would add more jumping/disappearing.

And putting it into category waste screen space

westnordost:
> Switching indexOf to lastIndexOf and adding .reversed() in getter, so that internal RecyclerView will work as expected (select item in the category, not the one in top entries) seems a bit ugly to me

And why do you have to do that?

matkoniecz:
without reversing listing following happens:

- user skips entry in top selections
- opens group
- select one of entries there, entry which was appearing also in top selection

Now the first entry (in top items) is selected instead, group which was opened closes

This is confusing and not desirable.

westnordost:
I see. Well, refactoring to make the `selectedIndex` the variable and `selectedItem` the computed property should do the trick, I think. Please try that, because currently it is indeed very hacky.

matkoniecz:
@westnordost 

Do you know or why https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/view/image_select/GroupedImageSelectAdapter.kt#L66 exists? It appears to be copying (saving?) variable and shadowing it. What is the point of that?

westnordost:
Try removing that line and you'll know.

matkoniecz:
Removing it resulted in need to add an extra `!!` in one place where Kotlin was not smart enough. And no bad effects that I noticed so I was curious whether I am overlooking something.

Either way I tried raw index numbers as primary way of identifying items and it was not working well as retracting/expanding groups was invalidating them.

I ended simply remembering group of item - now it works well in all cases without any weird behaviour or hacks in code.

I still would appreciate review but now I am happy enough to merge it.

westnordost:
> Removing it resulted in need to add an extra !! in one place where Kotlin was not smart enough. And no bad effects that I noticed so I was curious whether I am overlooking something.

Ehm, but the IDE explains why the `!!` is needed. And that should be reason enough to not use it there. If we just sticked `!!` everywhere, we might as well not have null safety.

app/src/main/java/de/westnordost/streetcomplete/view/image_select/GroupedImageSelectAdapter.kt
```diff
@@ -51,37 +75,42 @@ class GroupedImageSelectAdapter<T>(val gridLayoutManager: GridLayoutManager) :
 
     override fun onBindViewHolder(holder: ItemViewHolder, position: Int) {
         holder.bind(_items[position])
-        holder.isSelected = selectedItem?.let { _items.indexOf(it) == position } == true
-        holder.isGroupExpanded = getGroup(selectedIndex) == position
+        holder.isSelected = selectedItem?.let { _items.lastIndexOf(it) == position } == true
```
westnordost, line 78: There is still `lastIndexOf` here. Shouldn't it be

```suggestion
        holder.isSelected = selectedIndex() ==position
```

now?

app/src/main/java/de/westnordost/streetcomplete/view/image_select/GroupedImageSelectAdapter.kt
```diff
@@ -51,37 +75,42 @@ class GroupedImageSelectAdapter<T>(val gridLayoutManager: GridLayoutManager) :
...
         }
         if (selectedItem != null) {
-            val selectedIndex = _items.indexOf(selectedItem)
-            notifyItemChanged(selectedIndex)
+            notifyItemChanged(selectedIndex())
 
-            if (selectedItem.isGroup) {
-                if (prevSelectedItem == null || getGroup(_items.indexOf(prevSelectedItem)) != selectedIndex) {
-                    expandGroup(selectedIndex)
+            if (selectedItem!!.isGroup) {
```
westnordost, line 111: Please get rid of the !! here.

westnordost (changes requested):
Looks good, except for the `!!`s. You tested this?

matkoniecz:
> You tested this?

Yes, definitely (after my initial idea had unexpected weird behaviour, hack had also weird behaviour just in rare case and index based solution was crashing in many cases I definitely tested also this, and to some surprise it actually worked once I fixed some mistakes, without bugs in the design itself :) )

app/src/main/java/de/westnordost/streetcomplete/view/image_select/GroupedImageSelectAdapter.kt
```diff
@@ -51,36 +68,42 @@ class GroupedImageSelectAdapter<T>(val gridLayoutManager: GridLayoutManager) :
...
+            notifyItemChanged(previousItemIndex)
+            if (previousSelectedItemGroup != null) {
+                if (selectedItem == null || previousSelectedItemGroup != selectedItemGroup) {
                     retractGroup(previousGroupIndex)
                 }
             }
         }
-        if (selectedItem != null) {
-            val selectedIndex = _items.indexOf(selectedItem)
+        val copyOfSelectedItem = selectedItem
```
westnordost, line 101: "Copy" may be misleading, because it is not a copy. It points to the same object in memory.

matkoniecz:
@westnordost 

I marked it as a draft, as I discovered a bug that I have not spotted before, then discovered that it is present in the original code but is not triggered.

It appears that something goes wrong when list is long and one entry is retracted and another expanded. Sometimes newly expanded group is not actually shown.

For example, on `master` with addition of few more groups to make list longer ( https://github.com/matkoniecz/Zazolc/commit/bcb426eabc0394f08081975c92f8a74d580f8f85 ) following can be observed:

[device-2022-08-04-091017.webm](https://user-images.githubusercontent.com/899988/182785893-f9488e7c-f418-4e80-b397-6dc1b74fb08e.webm)

For now I remain confused what is going on. Initially I tried to understand what I broke, but as it is present also in original code (but not triggered!) so it is going a bit trickier to fix and I will need to learn more about internals of what this code is using.

westnordost:
No idea what may cause this from the top of my head. Maybe even the internal logic of the adapter is just fine and it is an issue with the RecyclerView that needs a workaround - after all, scrolling is purely UI feature.

westnordost:
The code for scrolling to ...  is here:

https://github.com/streetcomplete/StreetComplete/blob/941c75cf7cb27c9a7855adcc99d1fdd66ea89f3d/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestForm.kt#L89

Issue #4120, opened 2022-06-15
Inconsistent name requirements for some amenity quests


peternewman:
https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/main/java/de/westnordost/streetcomplete/quests/seating/AddSeating.kt#L16-L22
https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt#L16-L22
https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/main/java/de/westnordost/streetcomplete/quests/air_conditioning/AddAirConditioning.kt#L17-L22
https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/main/java/de/westnordost/streetcomplete/quests/smoking/AddSmoking.kt#L25-L39
Compared to:
https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddVegetarian.kt#L17-L25
Or also:
https://github.com/streetcomplete/StreetComplete/blob/18f7db06775732bfee022299e53898b883c4a98b/app/src/main/java/de/westnordost/streetcomplete/quests/level/AddLevel.kt#L35-L37

I'd have just done a PR, but I'm not sure which is the right way to go. I've sometimes found kiosks with no obvious name just what they sell (and mindful of the suspicious name check in iD left it blank) but then the subsequent quests don't appear. Likewise I can complete the other quests more quickly, but not if I have to key in the name. But then if the name changes/was wrong it all gets reset.

Name or brand certainly seems like a good compromise if we have to have some identifier (but in a lot of places they are isolated enough you can tell what's going on still).

**How to Reproduce**
Try and tag vegetarian on a node with no name, compared to seating.

**Versions affected**
44

FloEdelmann:
Out of curiosity: How did you spot these @peternewman?

peternewman:
Skipping the name on a kiosk and café and noticing that it didn't let me answer the dietary quests but went straight to card and air-con. I then filled some dummy data (which I subsequently undid) and they worked.

PR #4241, opened 2022-07-26
Ask wheelchair access for nameless businesses

Helium314:
According to https://github.com/streetcomplete/StreetComplete/issues/4120#issuecomment-1156437799

> most "name" requirements are relicts from when the feature name of each element was not always displayed.

So it looks like the `(name or brand)` requirement for the wheelchair quest is a leftover from earlier times and can be removed.


peternewman (approved):

matkoniecz:
Looks good to me.
Have you tested is this new code working?

Helium314:
Yes, tested and shows wheelchair quest for nameless places.

PR #4243, opened 2022-07-27
Fix crash in aerial view

Helium314:
Aerial view is currently not used, so this fix will not have any noticeable effect.
But in case aerial view is resurrected, it shouldn't crash when toggling `show3DBuildings` (happens because in aerial view there is no buildings layer where `extrude` could be toggled).


westnordost:
Cool, thank you!

PR #4246, opened 2022-07-27
this quests are useful also for cyclists

matkoniecz:
Is there some design reason why it was not done before? Is it safe to just add achievement to quest in such way?

Especially `tracktype` and `surface` is more important to cyclists than car drivers.


BalooUriza:
Is it?  You need a pretty decent track to get most cars down them but even
an omafiets is usable on all but the worst tracks.

riQQ:
In Germany, a lot of tracks are paved or compacted roads for agriculture that are also (easily) used by bikes.

westnordost:
but not conveniently

PR #4271, opened 2022-08-06
Load quest pins/dots up to zoom 14

Helium314:
When changing `TILES_ZOOM` from 14 to 16, the zoom level necessary for loading new quest pins / dots was also changed from 14 to 16.
Since this change, dots (visible from between levels 16 and 14, see [here](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/assets/map_theme/jawg/streetcomplete.yaml#L113)) are displayed, but never updated / loaded when panning the map without zooming in.

With this change, dots are now updated at zoom levels 14-15.999 too (if no more than 16 z16 tiles are visible, as per restriction a few lines below the change).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
@@ -111,6 +111,7 @@ class QuestPinsManager(
     fun onNewScreenPosition() {
         if (!isActive) return
         val zoom = ctrl.cameraPosition.zoom
+        // require zoom >= 14, which is the lowest zoom level where quests are shown
         if (zoom < 14) return
         val displayedArea = ctrl.screenAreaToBoundingBox(RectF()) ?: return
         val tilesRect = displayedArea.enclosingTilesRect(TILES_ZOOM)
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
@@ -111,7 +111,7 @@ class QuestPinsManager(
     fun onNewScreenPosition() {
         if (!isActive) return
         val zoom = ctrl.cameraPosition.zoom
-        if (zoom < TILES_ZOOM) return
+        if (zoom < 14) return
         val displayedArea = ctrl.screenAreaToBoundingBox(RectF()) ?: return
         val tilesRect = displayedArea.enclosingTilesRect(TILES_ZOOM)
         // area too big -> skip (performance)


Helium314:
actually the same could be done for `StyleableOverlayManager`...

westnordost:
One does not really have anything to do with the other. The cutoff at tile 14 is simply a performance consideration (not too many tiles should be pulled from DB at once).
So that they are visible but are not fetched from DB while at zoom 14 is actually deliberate.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
```diff
@@ -111,7 +111,7 @@ class QuestPinsManager(
     fun onNewScreenPosition() {
         if (!isActive) return
         val zoom = ctrl.cameraPosition.zoom
-        if (zoom < TILES_ZOOM) return
+        if (zoom < 14) return
```
peternewman, line 114: Wouldn't another constant rather than a magic number be a better idea? Then the same value could potentially be used in the other file too?

westnordost: 🤷 or a comment

Helium314:
But performance is already considered independen of zoom level in
https://github.com/streetcomplete/StreetComplete/blob/658110fcaf11160ed64a7da056c50629a4bf4999/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt#L117-L118

westnordost:
Yes, true. But I do still remember that I added this for performance reasons. I.e. that not even `screenAreaToBoundingBox` and `enclosingTilesRect` are executed when zoom < 14. (`onNewScreenPosition` is executed each frame when the map is moving)

Helium314:
I've been using it for a while now, and loading the dots definitely has a noticeable delay on low zoom. But it's not any worse than before v45.
If you mainly don't want this for performance reasons, maybe we can add it after cache is done?

westnordost:
I didn't say I don't want it. I was just telling you why I added it exactly in this way in the first place. I think you now spent more time around that code than me, so I think it is fair if you decide what's best.

Helium314:
I think this is not much of a performance issue, as I find it slow but acceptable on S4 mini (and it will get better with the cache).
Personally I sometimes like to have some overview, and I find it somewhat confusing/annoying that quests are displayed, but don't change when panning the map on low zoom.
With the proposed change this would improve, and I don't think it would aversely affect most users.

Though there is one potential drawback: if a user zooms out to z15, moves the map a bit and zooms in again, there might be be 2 slow jobs started in `onNewTilesRect` running at the same time (one for zoom out, one for move), with the first one actually being useless after the second ist started.
Maybe the old job could be canceled when starting a new one, this should also be beneficial for performance independent of this PR.

westnordost:
> slow jobs started in onNewTilesRect running at the same time

Are they running at the same time? Are they not serialized?

Helium314:
Access to `questsInView` is synchronized, and `pinsMapComponent.set(pins)` seems to be synchronized at some point in tangram if I remember earlier tests correctly.
But what may occur:
1. user zooms out / moves the map
2. `onNewTilesRect` job is started (fetch quests, fill `questsInView`, call `pinsMapComponent.set(pins)`)
3. user moves the map
4. another `onNewTilesRect` job is started, but if there are many quests in previous view, the job from 2. may still be fetching quests or setting pins. Now actually the result from 2. is not needed any more, but je job will still complete before this jobs sets any pins
5. user moves the map
6. another job is started...

If the jobs take long, there will be a considerable delay for the last one, which is the one that actually is relevant, as it's for the most recent tilesRect.

I did test this a bit (without this PR), and in my normal usage had 2 jobs running in parallel quite often, and sometimes 3 or even 4. Note that running in parallel likely means the first job is actually doing somthing, while the others are waiting.

PR #4276, opened 2022-08-07
Don't set check_date:opening_hours unless surveyed

### LINKED ISSUE #4273: Opening hours check date is updated when marked as not signed
In the 'Are these opening hours still correct?' quest, if the user selects 'No opening hours sign', then two changes are made:

* `opening_hours:signed` is set to `no`, as expected, and
* `check_date:opening_hours` is set to the current date.

To me, updating `check_date:opening_hours` implies that the opening hours are correct at the current time. But the since the opening hours haven't actually been checked, they could be wrong/out of date.

**How to Reproduce**
Um, find an 'Are these opening hours still correct?' quest, and select 'No opening hours sign'.

**Expected Behavior**
I'm not sure exactly what the right thing to do here is. Maybe the current behaviour is actually the desired behaviour?

My suggestion would be to use `check_date:opening_hours:signed` instead, although I don't think this key exists yet and I don't know what the guidance is for when it's alright to introduce new keys. 

If I've understood the code, then the opening hours quest isn't shown when `opening_hours:signed` is `no`.
And setting `check_date:opening_hours:signed` to the current date would stop the opening hours _signed_ quest from being shown.

**Versions affected**

v45.1

spagy:
Fixes #4273 

Previously, when resurveying opening hours, if a user said there was no
opening hours sign then we'd update check_date:opening_hours to the
survey date. This implies that the opening hours are correct but the
user hasn't actually checked that.

Instead of doing that, we now update check_date:opening_hours:signed.
This means that the quest will be hidden again for a while but we avoid
marking the opening hours as recently checked if they haven't actually
been confirmed.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -157,7 +157,7 @@ class AddOpeningHours(
     override fun applyAnswerTo(answer: OpeningHoursAnswer, tags: Tags, timestampEdited: Long) {
         if (answer is NoOpeningHoursSign) {
             tags["opening_hours:signed"] = "no"
-            tags.updateCheckDateForKey("opening_hours")
+            tags.updateCheckDateForKey("opening_hours:signed")
             // don't delete current opening hours: these may be the correct hours, they are just not visible anywhere on the door
         } else {
             val openingHoursString = when (answer) {
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -169,6 +169,7 @@ class AddOpeningHours(
             tags.updateWithCheckDate("opening_hours", openingHoursString)
             if (tags["opening_hours:signed"] == "no") {
                 tags.remove("opening_hours:signed")
+                tags.remove("check_date:opening_hours:signed")
             }
         }
     }


matkoniecz:
Have you tested it?

If you are not really planning to work on it let me know and I will test it is it actually working (you can help by finding some old shop with opening_hours check date).

But if you are interested - then https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#development has some docs (and yes, initial setup will take some time)

There are few things that could go wrong (not actually suppressing immediate asking "are opening hours signed", crashing or something)

Helium314:
It might also be good to remove the `opening_hours:signed` check date when removing `opening_hours:signed`.

spagy:
> Have you tested it?

i'm not sure exactly what testing is needed. here's how i tested:

* ran the app, with the change, on my phone (without logging in to my osm account)
* found a shop with an opening hours resurvey quest
* selected 'No opening hours sign'
* clicked undo, so i could see the changes made
* verified that `check_date:opening_hours` wasnt updated, but `check_date:opening_hours:signed` was.

i don't know whether the shop in the quest that i tested with has an opening hours sign, so i didn't want to try uploading it the change. i'm _assuming_ the upload process should work fine, but if it's worth testing that too i'm happy to go out and find an opening hours resurvey quest for shop that _doesn't_ have an opening hours sign.

please let me know if there's anything else that'd be good to test.

> It might also be good to remove the `opening_hours:signed` check date when removing `opening_hours:signed`.

yeah, that makes sense to me. i'll do that!

matkoniecz (approved):

matkoniecz:
Then it is fine! I was asking as some people submitted PRs without even compiling code.

I will not merge it immediately as maybe westnordost is noticing something that I have not noticed.

> but check_date:opening_hours:signed was.

as in added? (as it is unlikely to be present before)

> please let me know if there's anything else that'd be good to test.

The quest for presence of opening hours sign has not appeared immediately, right?

spagy:
the `check_date:opening_hours:signed` tag is now also removed when `opening_hours:signed` is removed. the code builds and works fine, as far as i can tell.

> as in added? (as it is unlikely to be present before)

yes `check_date:opening_hours:signed` _is_ added.

and the quest for the presence of the opening hours sign isn't shown again immediately.

westnordost:
>> Have you tested it?

> i'm not sure exactly what testing is needed. here's how i tested:

There are tests in /app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt that need to be updated / new tests need to be added.

If you are not prepared to do that (it would involve you download the Android Studio IDE and run the tests), you can close the PR and the reported issue will stay open until someone (maybe myself) will implement it fully.

I am also not sure if the implementation is complete or will work that way correctly. After all, does the quest that looks for whether the opening hours are signed look for the check date on opening hours? Or the check date on opening hours signed? Doesn't it have to look for both now?

peternewman (changes requested):
I'm pretty certain this will need tweaking too, or we'll never re-survey non-signed opening hours which SC added:
https://github.com/streetcomplete/StreetComplete/blob/7c97885cb6c24f9f7defed22e4497a495ab7e4ab/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt#L41-L44

Or we'll continually resurvey them.

This lot likely needs tweaking too:
https://github.com/streetcomplete/StreetComplete/blob/7c97885cb6c24f9f7defed22e4497a495ab7e4ab/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt#L69-L86

Or in other words there are lots more corner cases around resurveying and transitions to/from signed which ought to be tested too. It will also need to now handle the both styles of existing opening hours check date tagging for a resurvey...

https://github.com/streetcomplete/StreetComplete/blob/d57c74c36a09cf877c431fe7bed6af86aa95a5d4/app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSignedTest.kt#L32-L87

Have you run the tests, as they unfortunately aren't run automatically here for true CI, I'm pretty certain these will fail now:
https://github.com/streetcomplete/StreetComplete/blob/8eaf0606fceb814ab120bc0a996655d5bde595b5/app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt#L50-L65

The tests could definitely do with some unsigned to signed tests too.

peternewman:
Great minds @westnordost ...

> If you are not prepared to do that (it would involve you download the Android Studio IDE and run the tests), you can close the PR and the reported issue will stay open until someone (maybe myself) will implement it fully.

This isn't entirely true, you should be able to run it as a GitHub action here:
https://github.com/spagy/StreetComplete/actions/workflows/test.yml

westnordost:
Alright I'll close this then. A proper change would require redoing this anyway.

PR #4283, opened 2022-08-12
Improve performance when updating pins

Helium314:
This addresses issues mentioned in https://github.com/streetcomplete/StreetComplete/pull/4271#issuecomment-1210658453, and another on I didn't mention there.
Main purpose of this is to avoid calling `pinsMapComponent.set`, as this takes some 100-500 ms on my phone depending on number of quests.

The jobs for updating setting pins are made cancelable, and canceled when another job is started. Reason is that we're interested in the most recent result only.
I added some `synchronized` sections, as otherwise waiting would happen somewhere deeper in `visibleQuestsSource.getAllVisible` / `pinsMapComponent.set`, where the job can't be (easily) canceled.

In actual usage (with added logging) I found that the `pinsUpdateJob` is canceled reasonably often without.
When quickly panning the map on "low" zoom (z16), pins load (very) noticeably faster. Though no one would do this in actual use, it still helps e.g. when moving camera to user location from far away.

The second improvement is calling `pinsMapComponent.set` from `updateQuestPins` only if quest in view were added/deleted. This is a noticeable improvement when uploading a bunch of quests that are not currently visible.


westnordost (commented):
Hm, there are at least two performance improvements in this one. I am not sure about the coroutine job thing. Maybe you could first create a PR that checks if any of the added/removed quests are actually in view because that change seems to be straightforward useful. (Though I try to keep the *Manager.kt classes in sync because they are very similar, i.e. the StyleableMapDataManager could make use of the same improvement, no?

> coroutineContext.isActive

I think you are looking for [yield](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html)

Helium314:
> I think you are looking for [yield](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html)

`yield` will throw an exception if the job is canceled, though I don't know how expensive "exceptions are expensive" is, and whether this is actually relevant. Though compared to the gain of not executing some slow part it's probably negligible...

Anyway: `yield` inside `synchronized` does not compile, with some "yield inside critical section" message. Though it could probably be solved using some (volatile) booleans like `currentlyFetchingQuests` and `currentlySettingPins` instead of `synchronized`...

> StyleableMapDataManager could make use of the same improvement

I didn't do any performance tests with overlays, but at very least for fetching map data it will be useful.

> Maybe you could first create a PR that checks if any of the added/removed quests are actually in view

Will do (for both managers), though I won't have the time for a week or so.

westnordost:
First time I hear about exceptions being expensive. Is it because a stack trace is generated? Can't be too expensive, that. Do you have a source?

Anyway. You are right, yield inside a synchronized is a bad idea.

Helium314:
> Do you have a source?

No, I only remember reading that somewhere, probably comment on Stackoverflow... but can't find it again.

riQQ:
Yes, exceptions are expensive and the expensive part is creating the stack trace by unwinding the stack.
A source: https://stackoverflow.com/questions/567579/how-expensive-are-exceptions

westnordost:
One think that came to my mind: You know, your great cache PR will get merged sooner or later anyway. And after it is merged, the "fetching of the quests" will usually not be a I/O operation anymore, so to me it looks like there is no opportunity to cancel the coroutine before the first call to the map update reaches to point in the code where the map pins are set anyway (because the fetching from the cache is done in negligible amount of time). So, wouldn't this patch be non-functional anyway?

Helium314:
Well, it will still help in some case(s) when the areas are not cached.
I had thought about this before starting this PR, and also did some tests with and without cache (though mostly without), checking if / where the job is canceled when using the app normally.

Without cache, cancelling mostly happens before fetching from db.
With cache it happens much less often, and if, then mostly before setting the pins.
But there are some situations where cancelling before fetching still happend several times in quick succession: When having the map somewhere away from my location, and then moving the map to my location by tapping the pointer pin, the map scrolls very quickly. During the scroll there are several calls to `onNewTilesRect` each loading some area along the way the camera travels. So during some ~100 ms camera travel we may have accumulated a lot of `tileRects` to load. If we have data for those tiles in DB, it can take some 10 seconds until quests near my location are displayed (without this PR).
I had noticed this slow loading after move on some occasions, but so far wasn't able to understand what was going on...
Anyway, this is probably not happening very often, especially to users who only care about nearby things.

So while cache clearly diminishes usefulness of this PR, I think it can still be useful with cache, because
* it still helps by sometimes cancelling before setting pins (because setting pins may take several 100 ms)
* it may help when centering the camera on the user from far away (only if SC has quests along the way of camera travel)

westnordost (commented):
Ok, let's include this.

Though, I do wonder about the `synchronized`. I do not understand the first one (with getAllVisible).

What is it for?

If you want to wait before the new job starts until either the previous job ran to completion or was cancelled successfully, maybe it would be easier (and more readable) to use `job.join()` before launching the new job instead? (see also `CoroutineIntentService`)

---

Also, ideally, the same improvement should be done in the `StyleableOverlayManager`, but it is not required for this PR to be merged.

Helium314:
The purpose of the `synchronized` is to make sure that `if (!coroutineContext.isActive)` is checked at latest possible point before calling `getAllVisible`.
I'm not really happy with how it looks currently, but it was the best I found.

So it could be something like
```kotlin
        pinsUpdateJob?.cancel()
        pinsUpdateJob = viewLifecycleScope.launch {
            val quests = withContext(Dispatchers.IO) {
                pinsUpdateJob?.join()
                visibleQuestsSource.getAllVisible(bbox)
            }
            setQuestPins(quests)
        }
```
to have the same functionality? (didn't try it yet)


> Also, ideally, the same improvement should be done in the StyleableOverlayManager, but it is not required for this PR to be merged.

I'm definitely going to add it in this PR.

westnordost:
> The purpose of the synchronized is to make sure that if (!coroutineContext.isActive) is checked at latest possible point before calling getAllVisible.

So if you leave out the synchronized, what later point would there be then?

Helium314:
When leaving out `synchronized`, the `if (!coroutineContext.isActive)` check will happen right after the job was started, so the coroutine will always be active.
With `synchronized`, the check will happen after a potential previous call to `visibleQuestsSource.getAllVisible(bbox)` returned. And not call `visibleQuestsSource.getAllVisible(bbox)` if the job was canceled meanwhile.

westnordost:
> When leaving out synchronized, the if (!coroutineContext.isActive) check will happen right after the job was started

No, first `visibleQuestsSource.getAllVisible(bbox) ` is called, and access to the quest data is already serial (by the database, and when you are finished with the cache thing, by the controller).

Helium314:
But even with the serialized access, the waiting will happen somewhere inside `visibleQuestsSource.getAllVisible(bbox)` (probably right before accessing cache or DB). And when waiting at this point, `job.cancel()` does not cancel anything. The `synchronized` moves the waiting to a place where the job can still be canceled after the waiting is done.

westnordost:
Hmm, I see. But then, what about the `join` I suggested? This seems to be easier to achieve that(?)

```
launch {
  job?.cancelAndJoin()
  job = launch { doTheThing(); }
}
```
This makes sure the next job only starts when the previous finished

Helium314:
As mentioned, I didn't try it yet. But I agree, this should do the same thing without requiring the ugly `synchronized` section.

Helium314:
When panning the map wildly:

```kotlin
        viewLifecycleScope.launch {
            styledElementsUpdateJob?.cancelAndJoin()
            styledElementsUpdateJob = launch {
                val mapData = withContext(Dispatchers.IO) {
                    mapDataSource.getMapDataWithGeometry(bbox)
                }
                setStyledElements(mapData)
            }
        }
```
appears to have no benefit at all (still loads all seen areas sequentially).

-----
```kotlin
        val oldStyledElementsUpdateJob = styledElementsUpdateJob
        styledElementsUpdateJob = viewLifecycleScope.launch {
            val mapData = withContext(Dispatchers.IO) {
                oldStyledElementsUpdateJob?.cancelAndJoin()
                mapDataSource.getMapDataWithGeometry(bbox)
            }
            setStyledElements(mapData)
        }
```
is definitely better and (visually) only loads the last visited area, but apparently still loads too much stuff (may take some ~30-40 s from last call to `onNewTilesRect` until `setStyledElements` is done)

-----
```kotlin
        styledElementsUpdateJob?.cancel()
        styledElementsUpdateJob = viewLifecycleScope.launch {
            val mapData = withContext(Dispatchers.IO) {
                synchronized(mapDataSource) {
                    if (!coroutineContext.isActive) null
                    else mapDataSource.getMapDataWithGeometry(bbox)
                }
            } ?: return@launch
            setStyledElements(mapData)
        }
```
is clearly the fastest here (~10-15s from last call to `onNewTilesRect` until `setStyledElements` is done, which is not much longer than the 8-10 s my phone needs to load without panning)

westnordost:
You didn't include `if (!coroutineContext.isActive) null` in the first two code snippets.

Helium314:
I just tried that, like
```kotlin
        val oldStyledElementsUpdateJob = styledElementsUpdateJob
        styledElementsUpdateJob = viewLifecycleScope.launch {
            val mapData = withContext(Dispatchers.IO) {
                oldStyledElementsUpdateJob?.cancelAndJoin()
                if (!coroutineContext.isActive) null
                else mapDataSource.getMapDataWithGeometry(bbox)
            } ?: return@launch
            setStyledElements(mapData)
        }
```
but there is no improvement. Still the `synchronized` version is clearly the fastest.

Helium314:
`oldStyledElementsUpdateJob?.cancelAndJoin()` never takes more than a few ms, so apparently it doesn't wait until the old job is completed.
Thus `mapDataSource.getMapDataWithGeometry(bbox)` is (nearly) always called.

westnordost:
> oldStyledElementsUpdateJob?.cancelAndJoin() never takes more than a few ms, so apparently it doesn't wait until the old job is completed.

That is really weird, it seems to contradict the documentation: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/cancel-and-join.html

There must be something else happening there, I don't think something this basic is not working as documented.

Also, your snippet seems to **always** call `setStyledElements`, isn't the point of it all to not do it? e.g. like that:

```kotlin
val oldJob = job
job = viewLifecycleScope.launch {
    oldJob?.cancelAndJoin() // suspends, i.e. this job can be cancelled while it waits for oldJob to be completed
    val mapData = withContext(Dispatchers.IO) { mapDataSource.getMapDataWithGeometry(bbox) }
    if (!coroutineContext.isActive) return@launch // may have been cancelled while querying the map data
    setStyledElements(mapData)
}
```

Helium314:
Tested with kotlin playground:
```kotlin
import kotlinx.coroutines.*

val viewLifecycleScope: CoroutineScope = CoroutineScope(SupervisorJob())
var styledElementsUpdateJob: Job? = null
var t = 0L

fun onNewTilesRect() {
    t = System.currentTimeMillis()
    val oldStyledElementsUpdateJob = styledElementsUpdateJob
    styledElementsUpdateJob = viewLifecycleScope.launch {
        val t1 = System.currentTimeMillis()
        oldStyledElementsUpdateJob?.cancelAndJoin() // suspends, i.e. this job can be cancelled while it waits for oldJob to be completed
        println("waiting for join took ${System.currentTimeMillis()-t1}ms")
        withContext(Dispatchers.IO) { Thread.sleep(1000) }
        if (!coroutineContext.isActive) return@launch // may have been cancelled while querying the map data
        println("done, ${System.currentTimeMillis() - t}ms since most recent call to onNewTilesRect")
    }
}

fun main() {
    onNewTilesRect()
    runBlocking { delay(300) }
    onNewTilesRect()
    runBlocking { delay(300) }
//    onNewTilesRect()
    runBlocking { delay(4000) }
    println("Hello, world!!!")
}
```
prints
```
waiting for join took 0ms
waiting for join took 748ms
done, 1751ms since most recent call to onNewTilesRect
Hello, world!!
```

If the last `onNewTilesRect()` is uncommented, it prints
```
waiting for join took 0ms
waiting for join took 44ms
done, 1045ms since most recent call to onNewTilesRect
Hello, world!!!
```
Not sure what, but something appears to be wrong here.

westnordost:
```
import kotlinx.coroutines.*

val scope: CoroutineScope = CoroutineScope(SupervisorJob())
var job: Job? = null
var i = 0
val t = System.currentTimeMillis()

fun onNewTilesRect() {
    val id = ++i
    
    val oldJob = job
    job = scope.launch {
        println("+${System.currentTimeMillis()-t} start ($id)")
        oldJob?.cancelAndJoin()
        println("+${System.currentTimeMillis()-t} joined ($id)")
        withContext(Dispatchers.IO) { Thread.sleep(1000) }
        println("+${System.currentTimeMillis()-t} wake up ($id)")
        if (!coroutineContext.isActive) return@launch
        println("+${System.currentTimeMillis()-t} finish ($id)")
    }
}

fun main() {
    onNewTilesRect()
    runBlocking { delay(300) }
    onNewTilesRect()
    runBlocking { delay(300) }
    onNewTilesRect()
    runBlocking { delay(300) }
    onNewTilesRect()
    runBlocking { delay(4000) }
    println("finish")
}
```

https://pl.kotl.in/UXw5Ze5tO

Helium314:
+153 start (1)
+157 joined (1)
+463 start (2) -> Nothing else happens for 2, this is good.
+763 start (3)
+816 joined (3) -> We don't want this, because job 1 is not yet finished. But it still happens... because job 2 is canceled and thus finished while waiting for job 1.

But I still don't know how to properly do it without `synchronized`.

westnordost:
> +153 start (1)
> +157 joined (1)
> +463 start (2) -> Nothing else happens for 2, this is good.
> +763 start (3)
> +816 joined (3) -> We don't want this, because job 1 is not yet finished. But it still happens... because job 2 is canceled and thus finished while waiting for job 1.

Sorry for posting earlier without comment. I should have added "`withContext(Dispatchers.IO) { Thread.sleep(1000) }` seems to allow the coroutine to be cancelled".
Note that the `finish` log is only printed for (4).

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
```diff
@@ -63,8 +63,10 @@ class QuestPinsManager(
 
     private val visibleQuestsListener = object : VisibleQuestsSource.Listener {
         override fun onUpdatedVisibleQuests(added: Collection<Quest>, removed: Collection<QuestKey>) {
-            pinsUpdateJob?.cancel()
-            pinsUpdateJob = viewLifecycleScope.launch { updateQuestPins(added, removed) }
+            pinsUpdateJob = viewLifecycleScope.launch {
+                pinsUpdateJob?.join() // don't cancel, as updateQuestPins only updates existing data
```
westnordost, line 67: I think this does not make sense. At the point where what is in pinsUpdateJob is executed, pinsUpdateJob is already `this` job. What happens if a job joins on itself?

Helium314: Oh, right... it should be using `oldPinsUpdateJob`.

> What happens if a job joins on itself?

Don't know, but I had unintentionally tried `cancelAndJoin` in on the job itself, and it was just canceled and nothing else. So I guess in "best" case nothing happens.

Helium314:
> Sorry for posting earlier without comment. I should have added "`withContext(Dispatchers.IO) { Thread.sleep(1000) }` seems to allow the coroutine to be cancelled".

Looks like it's canceled when `withContext` finishes. If you move the log line inside the `withContext` block you always get the wakeup message.
So `if (!coroutineContext.isActive) return@launch` actually is useless here.

westnordost:
Mh, that's good to know. So whenever a coroutine context within a coroutine finishes, it checks whether the coroutine should exit now. Nice.

Helium314:
Hmm, but we still have no (simple) replacement for the `synchronized`...
Is it really that bad?

changing
```kotlin
            val quests = withContext(Dispatchers.IO) {
                synchronized(visibleQuestsSource) {
                    if (!coroutineContext.isActive)
                        // can't return@launch here, so return null and then @launch
                        return@synchronized null
                    visibleQuestsSource.getAllVisible(bbox)
                }
            } ?: return@launch
```
to
```kotlin
            val mapData = withContext(Dispatchers.IO) {
                synchronized(mapDataSource) {
                    if (!coroutineContext.isActive) null
                    else mapDataSource.getMapDataWithGeometry(bbox)
                }
            } ?: return@launch
```
at least makes it look a little bit more readable...

westnordost:
Huh? I was under the impression that the kotlin play example above showed that using cancel and join jobs would work just fine. E.g. the "finish" (=calling update map pins) was only called once, how we want it to be. https://github.com/streetcomplete/StreetComplete/pull/4283#issuecomment-1232692254

Helium314:
Well, for that it works.
But purpose of the `synchronized` was to call `Thread.sleep` as little as possible (i.e. cancel right before calling). This is the placeholder for `mapDataSource.getMapDataWithGeometry`. Actually a bad placeholder, because `Thread.sleep` is not serialized, but `mapDataSource.getMapDataWithGeometry` is.

`cancelAndJoin` is an improvement over the current state, but test in https://github.com/streetcomplete/StreetComplete/pull/4283#issuecomment-1232022492 showed that from last call to `onNewTilesRect` until the map data actually being updated it may still take much longer than when using `synchronized`.
This is because all the previous calls to `mapDataSource.getMapDataWithGeometry` get processed before the one (last) call we actually need to display the data in view. And with `synchronized` we simply don't call if the previous call isn't done. And then we can cancel right before calling.
In worst case with `synchronized` there is one call we need to wait for before calling `mapDataSource.getMapDataWithGeometry`.
With `cancelAndJoin` there isn't really a limit, and we may need to wait for 5 previous calls.

westnordost:
Hmm, ok 😵

Helium314:
Oh... I just found that `mutex.withLock` could be used to replace `synchronized`. Would that be better?

westnordost:
It wouldn't really make it shorter, would it? Personally, I only use mutexes if I need to do more complicated stuff that `synchronized` offers.

Helium314:
No, it would only add another line and some imports.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
```diff
@@ -126,22 +135,31 @@ class QuestPinsManager(
...
-        viewLifecycleScope.launch {
-            val quests = withContext(Dispatchers.IO) { visibleQuestsSource.getAllVisible(bbox) }
+        pinsUpdateJob?.cancel()
+        pinsUpdateJob = viewLifecycleScope.launch {
+            val quests = withContext(Dispatchers.IO) {
+                synchronized(visibleQuestsSource) {
+                    if (!coroutineContext.isActive) null
+                    else visibleQuestsSource.getAllVisible(bbox)
+                }
+            } ?: return@launch
```
westnordost, line 145: Not sure if I took note of that before, but as far as I know, the Dispatchers.IO has a very limited threadpool and this construct blocks one of them.
Is it possible to put the `withContext` just around the `getAllVisible`?

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
```diff
@@ -58,7 +63,11 @@ class QuestPinsManager(
 
     private val visibleQuestsListener = object : VisibleQuestsSource.Listener {
         override fun onUpdatedVisibleQuests(added: Collection<Quest>, removed: Collection<QuestKey>) {
-            viewLifecycleScope.launch { updateQuestPins(added, removed) }
+            val oldPinsUpdateJob = pinsUpdateJob
+            pinsUpdateJob = viewLifecycleScope.launch {
+                oldPinsUpdateJob?.join() // don't cancel, as updateQuestPins only updates existing data
+                updateQuestPins(added, removed)
```
westnordost, line 69: I don't get it, why not?

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/StyleableOverlayManager.kt
```diff
@@ -40,6 +43,8 @@ class StyleableOverlayManager(
 
     private val viewLifecycleScope: CoroutineScope = CoroutineScope(SupervisorJob())
 
+    private var styledElementsUpdateJob: Job? = null
```
westnordost, line 46: how about just naming it updateJob? After all, we are already in this class, it's clear that it is for the styled elements. (Same with the other class.)

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
Helium314: When panning the map while uploading, canceling on `onUpdated` will cancel the job for loading data currently in view, and then only update some elements / quests. Data in view may _not_ be loaded

I noticed this while testing, and not seeing any quests in an area you know you have downloaded is really bad.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
Helium314: My decision for blocking Dispatchers.IO actually was deliberate, but for the same reason.

I didn't find any clear information, but I assumed `viewLifecycleScope.launch` uses [Dispatchers.Default](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-default.html), which only has one thread for each CPU core, while [Dispatchers.IO](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-i-o.html) has up to 64 threads.

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/StyleableOverlayManager.kt
Helium314: done

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
westnordost: I see, ok

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/StyleableOverlayManager.kt
```diff
@@ -55,7 +60,11 @@ class StyleableOverlayManager(
 
     private val mapDataListener = object : MapDataWithEditsSource.Listener {
         override fun onUpdated(updated: MapDataWithGeometry, deleted: Collection<ElementKey>) {
-            viewLifecycleScope.launch { updateStyledElements(updated, deleted) }
+            val oldSUpdateJob = updateJob
+            updateJob = viewLifecycleScope.launch {
+                oldSUpdateJob?.join() // don't cancel, as updateStyledElements only updates existing data
```
westnordost, line 65:
```suggestion
            val oldUpdateJob = updateJob
            updateJob = viewLifecycleScope.launch {
                oldUpdateJob?.join() // don't cancel, as updateStyledElements only updates existing data
```
typo

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
westnordost: But when it is cancelled, it means that it is cancelled because another tile rect comes into view, no? And that means that either the updates are not in view anymore or they are part of the data just fetched by bbox from the db anyway.

Helium314: > But when it is cancelled, it means that it is cancelled because another tile rect comes into view, no?

Only cancelling as part of `onNewTilesRect`.
But here (in `onUpdatedVisibleQuests`) there is no new `TilesRect` coming in view, only some quests should be added/removed.

westnordost: I see, okay

westnordost:
Just for the record: This performance improvement is only in because setting the pins with tangram-es is so relatively slow. If it was faster and/or tangram-es would not be used to display pins, I think the complexity this introduces and thus the potential for bugs would not be worth it.

westnordost:
Hm I noticed now in the alpha which includes this change that sometimes while scrolling nothing seems to load at all. 

I didn't systematically try to reproduce the problem, so I am not sure if nothing appeared at all until I at least moved to a different tile or if it just took a lot more time. But in any case, there seems to be some kind of hiccup now.

However, this does not have to be related to the changes in this PR, maybe the issue was present before and I just did not notice it.

Helium314:
Hmm... I haven't seen this since the change to not cancel the updateJob in `onUpdated`, and I had used a build including this PR for a while now. After all, every time the job is canceled a new one should be started.

And I failed to reproduce the problem so far...

PR #4285, opened 2022-08-13
Allow manual runs of the lint workflow too

peternewman:


westnordost:
Does this work? That there is a trailing colon looks suspect, not sure if this is even valid YML.

riQQ:
Yes, this works. The same syntax is used here in the official docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events.

peternewman:
Yeah that's where I got it from, and their validator passes it, and I can also now see the button in my master having merged into there:
https://github.com/peternewman/StreetComplete/actions/workflows/lint.yml

But I'm guessing that might not show for others...

westnordost (approved):

PR #4297, opened 2022-08-18
Don't update pins and elements if changes are outside viewed area

Helium314:
This is the "straightforward useful" part of #4283, plus (more or less) the same thing for `StyleableOverlayManager`, as per https://github.com/streetcomplete/StreetComplete/pull/4283#pullrequestreview-1071963887.

I did *not yet* test this, other than whether it compiles.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
@@ -18,6 +18,7 @@ import de.westnordost.streetcomplete.data.visiblequests.QuestTypeOrderSource
 import de.westnordost.streetcomplete.screens.main.map.components.Pin
 import de.westnordost.streetcomplete.screens.main.map.components.PinsMapComponent
 import de.westnordost.streetcomplete.screens.main.map.tangram.KtMapController
+import de.westnordost.streetcomplete.util.math.contains
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/screens/main/map/QuestPinsManager.kt
```diff
@@ -140,11 +141,16 @@ class QuestPinsManager(
...
         val pins = synchronized(questsInView) {
-            added.forEach { questsInView[it.key] = createQuestPins(it) }
-            removed.forEach { questsInView.remove(it) }
+            addedInView.forEach { questsInView[it.key] = createQuestPins(it) }
+            removed.forEach { if (questsInView.remove(it) != null) deletedAny = true }
             questsInView.values.flatten()
         }
+        if (!deletedAny && addedInView.isEmpty())
+            return
         pinsMapComponent.set(pins)
```
westnordost, line 154:
```suggestion
        if (deletedAny || addedInView.isNotEmpty())
            pinsMapComponent.set(pins)
```

Just to make it consistent with the code in StyleableOverlayManager

westnordost (approved):
Code looks good! Of course, as always, it must be tested before it is merged.

Helium314:
Not sure if this is enough: I just used SC with this PR, and didn't encounter any issues when uploading (triggers the call to `updateQuestPins` / `updateStyledElements`). Done with and without overlay enabled.

westnordost:
Hmm, i think this is enough,  than you!

PR #4312, opened 2022-08-25
update element ids referenced in note on upload (fixes #4309)

### LINKED ISSUE #4309: Broken link in created note (internal ID?)
StreetComplete created a broken link in a [note](https://www.openstreetmap.org/note/3328360):

 > Unable to answer "What surface does this piece of road have?" for https://osm.org/way/-27 via StreetComplete 45.2:

I guess that the road has been split in the app, but not uploaded yet, so an internal ID has been added to the link.

**How to Reproduce**
I'm not sure. 

**Expected Behavior**
The app should only add working links with online IDs to notes.

**Versions affected**
StreetComplete 45.2 (according to note)

westnordost:
Done, but must be tested before merge.


matkoniecz:
Is it going to work fine when unsynced notes somehow referred to way `-1` and `-19`? And `-1` replacement is run first? Or is it somehow protected against it?

westnordost:
That is an excellent remark! I think that would break it

HolgerJeromin:
Perhaps adding a placeholder string on creation and change that on upload is a cleaner approach.

Issue #2126, opened 2020-10-01
Posted stop ID numbers


BalooUriza:
### General
Affected tag(s) to be modified/added: [](https://wiki.openstreetmap.org/wiki/Key:)
Question asked: What is the ID number for this stop?

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

**Element selection:** 
Elements with `highway=bus_stop` but no `ref=*` tag.  Almost certainly also applicable to just about any other public transit stop, not just bus stops.

**Metadata needed:** 
I believe this applies worldwide given the widespread adoption of GTFS.

**Proposed GUI:**
Default to numeric keypad as most stop IDs are exclusively numerical, but allow other characters to be entered as occasionally stop IDs get weird.

BalooUriza:
Rationale being that many transit systems put stop ID numbers at their stops so that riders can contact the transit agency to discover what routes will be arriving when.  This is also helpful for creating GTFS feeds (official or otherwise).  The problem is the stop IDs published in some agencies GTFS do not actually reflect the real, on the ground, stop ID.

Stop IDs are mappable in OSM by adding the `ref=*` tag to a transit stop.

westnordost:
I think this is not eligible, for the following reasons:

- **👨‍💻 Users are no experts** / **🐿️ Easy answer**: Users don't know what to look for or where. Is the ref on the pole? Somewhere on the shelter? Somewhere on the the bus schedule? In reality, there are usually a lot of numbers everywhere, users won't know if any has a certain meaning.
- **🤔 No unanswerable quests**: Maybe in Oklahoma, all bus stops have refs, in other places, there may not even be one which would mean that in other places, this quest is not actually answerable. Additionally, we cannot assume that the ref is numerical.

Furthermore, not necessarily a reason to not implement it, but simply to say it is not such a big loss if this information can not be recorded with this app:
[...]

Issue #3059, opened 2021-07-13
Ask ref for fire_hydrant in France


millecentdix:
Hi,
For attribut emergency=fire_hydrant, we are asked about the fire_hydrant:type, it will be interesting to ask the ref number too.
Thanks for reading :)

matkoniecz:
1) Note that quest suggestion has a helpful template: https://github.com/streetcomplete/StreetComplete/issues/new?assignees=&labels=&template=quest-suggestion.md

It would be the best to either open a new issue while filling it (or even better, copy it to that issue)

---------------

2)

Note that at least in some regions fire hydrants have no numbers obtainable by field survey (for example in Poland).

Where fire hydrants almost always have numbers?

----------------

See #922 - especially

> How is the ref helpful data?
> In which countries do hydrants have refs?
> How to distinguish the ref from other numbers (if any) that may possibly be found on the hydrants? Per-Country-guids necessary?

millecentdix:
Damed ! I haven't found the guideline suggestion template.
In France, most fire hydrants have a number, but not all.
I don't no if there is a regional ref.

matkoniecz:
Looking at https://commons.wikimedia.org/wiki/File:Fire-fighting-facility_node-5537552160.jpg and https://commons.wikimedia.org/wiki/Category:Fire_hydrants_in_France# in general it seems that many (most? almost all?) fire hydrants in France have a clear `ref` signed.

Still, do you know how this refs are used? For orientation purposes, maybe by firefighters?

millecentdix:
I have find informations in multiple technical guide edited by French firefighters. Hydrants are all referenced, but the reference making can various.
This ref is used to identify the hydrant, completed with address or GNSS location.
https://sdis41.fr/uploads/DECI/RECUEIL%20DES%20FICHES%20TECHNIQUES_V2.pdf page 12
https://www.pyrenees-atlantiques.gouv.fr/content/download/17382/114697/file/RDDECI64%20d%c3%a9finitif.pdf page 45

westnordost:
![image](https://user-images.githubusercontent.com/4661658/125657130-9c41636d-0e65-46df-b806-fb87e4204f89.png)

So there are two numbers? How will the user be able to differenciate?

tiptoptom:
In Germany and some other countries there is a sing at each underground fire hydrant which shows the diameter, the position in relation to the sign and optional the reference number.
Here two examples:
[...]

PR #4315, opened 2022-08-26
add railStopRef quest

candux:
Hello,

this PR adds a quest to fill out missing ref attributes on rail platforms. It's similar to https://github.com/streetcomplete/StreetComplete/issues/2126 for bus stops.
We limited the region of this quest to Germany, Switzerland and France for the moment because in other countries there could exist a reference, but it's not shown on a sign. As far as we know, if a reference exists, it is always shown on a platform sign in these three countries

**Checklist**

* [x]  :construction: To be added tag is established and has a useful purpose: Yes, the attribute is used by many map applications
* [x]  :thinking: Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one): Yes, if there is no ref, the user can select "no number or Code visible"
* [x]  :chipmunk: Easily answerable by everyone from the outside but a survey is necessary: if there is a platform ref, it's there on a big sign
* [x]  :zzz: Not an overwhelming percentage of elements have the same answer (No spam)
* [x]  :clock4: Applies to a reasonable number of elements (Worth the effort)

**Background**
We are developers from German (DB), Swiss (CFF SBB) and French (SNCF) train companies working on train stations indoor mapping. We organized a OpenStreetMap hackathon in Strasbourg (France) in August 2022 to make together contributions to this topic. Within the frame of the hackathon, this quest was developed.

We just launched the Open Rail Foundation to promote collaborative open-source projects between European railway actors such as UIC, DB, CFF-SBB and SNCF because we have the same problems in all the countries. So, we think it’s worth sharing ideas, resources, code, data, and work on OpenStreetMap data and open source tools. We hope this contribution will be useful for routing inside stations.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -41,6 +41,7 @@ import de.westnordost.streetcomplete.quests.bus_stop_lit.AddBusStopLit
 import de.westnordost.streetcomplete.quests.bus_stop_name.AddBusStopName
 import de.westnordost.streetcomplete.quests.bus_stop_ref.AddBusStopRef
 import de.westnordost.streetcomplete.quests.bus_stop_shelter.AddBusStopShelter
+import de.westnordost.streetcomplete.quests.rail_stop_ref.AddRailStopRef
 import de.westnordost.streetcomplete.quests.camera_type.AddCameraType
 import de.westnordost.streetcomplete.quests.camping.AddCampDrinkingWater
 import de.westnordost.streetcomplete.quests.camping.AddCampPower
@@ -223,6 +224,8 @@ fun questTypeRegistry(
     AddBusStopName(), // requires text input
     AddBusStopRef(), // requires text input
     AddBusStopLit(), // at least during day requires to stand in it to see if there is a light in the shelter
+
+    AddRailStopRef(), // requires text input
 
     AddRailwayCrossingBarrier(), // useful for routing
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/rail_stop_ref/AddRailStopRef.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/rail_stop_ref/AddRailStopRef.kt
@@ -1,0 +1,34 @@
+package de.westnordost.streetcomplete.quests.rail_stop_ref
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.NoCountriesExcept
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.PEDESTRIAN
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddRailStopRef : OsmFilterQuestType<RailStopRefAnswer>() {
+
+    override val elementFilter = """
+        ways with
+        (
+          (railway = platform and public_transport = platform)
+        )
+        and !ref and noref != yes and ref:signed != no and !~"ref:.*"
+    """
+    override val enabledInCountries = NoCountriesExcept("DE", "FR", "CH")
+    override val changesetComment = "Determine Rail Platform refs"
+    override val wikiLink = "Tag:public_transport=platform"
+    override val icon = R.drawable.ic_quest_railway
+    override val achievements = listOf(PEDESTRIAN)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_railStopRef_title2
+
+    override fun createForm() = AddRailStopRefForm()
+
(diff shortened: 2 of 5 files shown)


westnordost:
I have never seen a "ref" number for any railway or subway stop in Germany. Or I do not know what to look for. If it is not easily surveyable by laymen on the ground, it cannot be a quest in StreetComplete. In particular, I do not see 

> 👨‍💻 Users are no experts: No knowledge about OpenStreetMap or any other background knowledge must be necessary
> 🕵️ Publicly accessible on foot: The information the users are supposed to collect must be easily accessible from the outside by pedestrians

matkoniecz:
> We organized a OpenStreetMap hackathon in Strasbourg (France)

Maybe in France it is more visible? Like "Ask ref for fire_hydrant in France" #3059

Helium314:
> I have never seen a "ref" number for any railway or subway stop in Germany. Or I do not know what to look for

Isn't it just the track/platform number?
Looking at the [wiki](https://wiki.openstreetmap.org/wiki/Tag:public_transport=platform) it might also be `local_ref`, though judging by the _comment_ column both `ref` and `local_ref` appear to be more or less the same in this context.

westnordost:
Hm well, platforms in Germany usually do not have signed numbers, only the tracks do. E.g. one platform would lead to track 1 and track 2.

FloEdelmann:
@candux Friendly ping :slightly_smiling_face:

candux:
The goal of this quest is to add the ref attribute to all platforms. This attribute contains the track number which are accessible on this platform. If there are multiple possible tracks, they are separated by a semicolon. See this example: https://www.openstreetmap.org/relation/5708837

The title of this quest is in fact misleading. `platform_ref` would probably be better. I will also add a better description inside the quest.

> I have never seen a "ref" number for any railway or subway stop in Germany. Or I do not know what to look for. If it is not easily surveyable by laymen on the ground, it cannot be a quest in StreetComplete. In particular, I do not see

@westnordost Does this quest look reasonable to you that way?

westnordost:
OK, I understand you want to tag the (1-2) track numbers a platform leads to. Let's see:

1. Can you point us to a wiki page that explains how this is tagged? Is it `ref`? Is it `platform_ref` Has it not really been defined yet?
2. The quest can only be asked in places where track numbers are clearly signed. So not in e.g. subway systems, light rail etc., only regular trains. (I know, some light rail systems do have track numbers signed, but not all, and then usually only at larger stations)
3. The quest must be crystal clear what it asks about, lest wrong data is added by surveyors. Requiring the user to input some value separated by semicolon is "too much", there must be some custom UI that allows for 1 - 2 track numbers to be input. I am not sure what would be the best UI for that, actually, Suggestions welcome.

Helium314:
>     1. Can you point us to a wiki page that explains how this is tagged? Is it `ref`? Is it `platform_ref` Has it not really been defined yet?

According to https://wiki.openstreetmap.org/wiki/Tag:railway%3Dplatform `ref` should be used for "_Platform reference number_" (nothing about the track).

According to https://wiki.openstreetmap.org/wiki/Tag:public_transport=platform `local_ref` should contain "_Track number/platform number_".
So this actually requires that only one of those numbers exists (e.g. in Czechia there are both track and platform numbers, but quest is disabled there anyway).
But also on this page it's stated that actually `ref` is used for the track, which can also be seen when checking tagging at random railway stations in countries where the quest should be enabled.
This is quite inconsistent with bus stop ref... but well, it's definitely widely used that way.

>     2. The quest can only be asked in places where track numbers are clearly signed. So not in e.g. subway systems, light rail etc., only regular trains. (I know, some light rail systems do have track numbers signed, but not all, and then usually only at larger stations)

Judging from 
> We are developers from German (DB), Swiss (CFF SBB) and French (SNCF) train companies working on train stations indoor mapping. 

I assume this quest is intended only for "regular" trains, so light_rail/tram/subway/monorail should definitely be excluded.

westnordost:
Anyway, I'll close this PR because as stated, it would have to be implemented with a different UI and possible different tags.

mnalis:
(for parties interested who missed it, that quest had since been implemented in [SCEE](https://wiki.openstreetmap.org/wiki/SCEE) (StreetComplete _"Expert edition"_ fork)

mnalis:
> was it by any chance done fulfilling
>>  it would have to be implemented with a different UI and possible different tags

Well, it is different implementation, only inspired by this PR. (still not sure that it would qualify for SC, though, it that was the intention of the question) 

In quest settings (SCEE feature which SC don't have) it allows user to set whether the `ref` or `local_ref` should be tagged, and the quest selection is significantly more complex than what was suggested in this PR in order to remove false positives (e.g. not only extra tag filters, but also needing 2+ `railway=rail` railways nearby that `railway=platform` etc; see [SCEE/AddRailwayPlatformRef.kt](https://github.com/Helium314/SCEE/blob/modified/app/src/main/java/de/westnordost/streetcomplete/quests/railway_platform_ref/AddRailwayPlatformRef.kt)  for details)

PR #4326, opened 2022-08-30
Refactor: separate items from forms

westnordost:
**Reasons (in order of soundness):**
1. consistency: for some this separation is necessary; let's do it for all then. It's easier for new contributors if they just see one pattern in the code, not several
2. separation of concerns: the definition how the selectable items look does not have anything directly to do with the UI logic of the form
3. leads to slightly shorter files (because there are more files now)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierTypeForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierTypeForm.kt
@@ -3,26 +3,12 @@ package de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.quests.AImageListQuestForm
 import de.westnordost.streetcomplete.quests.AnswerItem
-import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type.BicycleBarrierType.DIAGONAL
-import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type.BicycleBarrierType.DOUBLE
-import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type.BicycleBarrierType.SINGLE
-import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type.BicycleBarrierType.TILTED
-import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type.BicycleBarrierType.TRIPLE
-import de.westnordost.streetcomplete.view.image_select.Item
 
 class AddBicycleBarrierTypeForm :
     AImageListQuestForm<BicycleBarrierType, BicycleBarrierTypeAnswer>() {
 
-    override val items = listOf(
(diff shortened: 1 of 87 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/AddBicycleBarrierTypeForm.kt
```diff
@@ -3,26 +3,12 @@ package de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type
...
 class AddBicycleBarrierTypeForm :
     AImageListQuestForm<BicycleBarrierType, BicycleBarrierTypeAnswer>() {
 
-    override val items = listOf(
-        Item(SINGLE, R.drawable.barrier_bicycle_barrier_single, R.string.quest_barrier_bicycle_type_single),
-        Item(DOUBLE, R.drawable.barrier_bicycle_barrier_double, R.string.quest_barrier_bicycle_type_double),
-        Item(TRIPLE, R.drawable.barrier_bicycle_barrier_triple, R.string.quest_barrier_bicycle_type_multiple),
-        Item(DIAGONAL, R.drawable.barrier_bicycle_barrier_diagonal, R.string.quest_barrier_bicycle_type_diagonal),
-        Item(TILTED, R.drawable.barrier_bicycle_barrier_tilted, R.string.quest_barrier_bicycle_type_tilted),
-    )
```
peternewman, line 22: Personally to me this is more readable and understandable with all the info in one array.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/BicycleBarrierTypeItem.kt
```diff
@@ -0,0 +1,23 @@
...
+    TILTED ->   R.string.quest_barrier_bicycle_type_tilted
+}
+
+private val BicycleBarrierType.iconResId: Int get() = when (this) {
+    SINGLE ->   R.drawable.barrier_bicycle_barrier_single
+    DOUBLE ->   R.drawable.barrier_bicycle_barrier_double
+    TRIPLE ->   R.drawable.barrier_bicycle_barrier_triple
+    DIAGONAL -> R.drawable.barrier_bicycle_barrier_diagonal
+    TILTED ->   R.drawable.barrier_bicycle_barrier_tilted
+}
```
peternewman, line 23: Than having to update both of these two, plus there's no risk of something having a title for SINGLE but forgetting to add the icon for example (but maybe the compile would catch it anyway).

peternewman (commented):
Just my personal opinion but I prefer the older style. Can't we just pull them into separate files?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_type/BicycleBarrierTypeItem.kt
westnordost: > (but maybe the compile would catch it anyway).

yes

matkoniecz:
it is not strong or really valid objection. But extra step with `asItem`, with images and labels assigned separately seemed always more alien to me, and less readable to me. Than generating them in a list, with each entry having label and image assigned together.

PR #4329, opened 2022-09-01
Street parking overlay

### LINKED ISSUE #4177: overlay : add parking_lane editing
**Use case**
There is a great editor for parking_lane tag [here](https://zlant.github.io/parking-lanes/#18/44.13594/4.80888).
image

**Proposed Solution**
The same as the overlay sidewalk for pedestrians an overlay would enhance a specific editing for cars.

westnordost:
image

Screenshot_20220901-162136_StreetComplete Dev

Screenshot_20220901-162144_StreetComplete Dev

Screenshot_20220901-162158_StreetComplete Dev

Fixes #4177

- `#D00055`: invalid or not set yet
- `#444444`: No parking, standing or stopping - dashed if just "no" (not signed explicitly)
- `transparent`: Separately mapped parking or not relevant (e.g. on service road)
-  `#007EEC`: Parking off the roadway - dashed if in "parking bays"
- `#03C3B8`: Parking half off roadway, i.e. half on kerb
- `#F7C204`: Parking on roadway - dashed if only in marked spaces

Chokers are shown as icons on the map, to help decide in fuzzy cases whether it should be still parking on roadway with some chokers on the road or whether it should already count as parking off the roadway.

Similarily, the width of the road if it is mapped is shown to help with the same.

Any parking areas are highlighted in blue.


tordans:
Love to see this come to life!

_A thought about the map:_ Would it be possible to "gray out" the map when I clicked a road segment in order to highlight the separately mapped parking spaces? I think I see those in the first image, but they could be overlooked quite easily.

_A thought on the labels:_ I found myself talking about "45°" and "90°" instead of Diagonal and Perpendicular because I can never remember those terms. Maybe those would be a good addition to the label, eg like "Diagonal ~45°", "Perpendicular 90°".

westnordost:
>  "gray out" 

Yeah, the graying out is "on my list" of things I'd like to add. Maybe there is time for that in the frame of implementing the cycleway overlay.

> Diagonal and Perpendicular because I can never remember those terms.

Well, a picture says more than a thousand words

tordans:
I suggest to collect feedback at https://github.com/streetcomplete/StreetComplete/discussions/4362 (or separate discussion posts).

PR #4333, opened 2022-09-02
implement ATM cash-in quest

mnalis:
Closes: https://github.com/streetcomplete/StreetComplete/issues/4292
([tested](https://github.com/mnalis/StreetComplete/actions/runs/2976059075) and seems to work ok)

Note: I re-used _"ATM Operator"_ quest `ic_quest_money` icon ("wad of notes" image)
Maybe it is fine as-is (for similar quests to have same icon), or a modified icon should be created.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/atm_cashin/AddAtmCashin.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/atm_cashin/AddAtmCashin.kt
@@ -1,0 +1,36 @@
+package de.westnordost.streetcomplete.quests.atm_cashin
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddAtmCashin : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -605,6 +605,7 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
 
     <string name="quest_airConditioning_title">"Is this place air-conditioned?"</string>
 
+    <string name="quest_atm_cashin_title">"Does this ATM allows depositing cash?"</string>
```
andrewharvey, line 608:
```suggestion
    <string name="quest_atm_cashin_title">"Does this ATM allow depositing cash?"</string>
```

andrewharvey: Or you could say.

> Can you deposit cash at this ATM?

mnalis: thanks @andrewharvey , added that as it sounds even better!

matkoniecz:
>([tested](https://github.com/mnalis/StreetComplete/actions/runs/2976059075) and seems to work ok)

just that it builds or also in use?

mnalis:
In use too, e.g. https://www.openstreetmap.org/node/5181271721/history
It adds the `cash_in=yes` (or `no`) tag, and then the quest is not shown again (as it shouldn't be).

westnordost (changes requested):
Could you use camel case "CashIn"?

app/src/main/java/de/westnordost/streetcomplete/quests/atm_cashin/AddAtmCashin.kt
```diff
@@ -0,0 +1,36 @@
...
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddAtmCashin : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
+        nodes with
+         amenity = atm
+         and !cash_in
+    """
```
westnordost, line 19:
```suggestion
    override val elementFilter = "nodes with amenity = atm and !cash_in"
```
I think this fits quite well into one line 🤷

mnalis:
> Could you use camel case "CashIn"?

CamelCase added.

westnordost (approved):

PR #4339, opened 2022-09-02
Respect ignored relations when uploading

### LINKED ISSUE #4319: Ignored relation type still downloaded
When investigating some unusually slow upload, I found that I have some huge relations of `type=route` in the database, despite the type being ignored according to
https://github.com/streetcomplete/StreetComplete/blob/0c20e22aa2679a6ac972c881916f031cb6ee9f1f/app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt#L54-L65

I am not sure how this happened, but the relations must have been downloaded when uploading edits (updating tags and splitting ways).
Anyway, I think relation types should be ignored consistently and not sneak into the db when uploading...

Helium314:
This should fix #4319 using the solution suggested in https://github.com/streetcomplete/StreetComplete/issues/4319#issuecomment-1235212882, but is not yet tested.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
@@ -1,5 +1,6 @@
 package de.westnordost.streetcomplete.data.osm.edits.upload
 
+import de.westnordost.streetcomplete.ApplicationConstants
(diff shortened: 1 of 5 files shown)


westnordost (approved):
Code-wise it looks good (thanks for the unit test!), needs one live test of course.

Helium314:
A little fix was necessary, but now it's tested and works.

westnordost:
Great, thanks!

PR #3914, opened 2022-03-24
Patch SnakeYAML dependency

FloEdelmann:
Fixes #3889.

Running `../gradlew dependencies` in the `app` directory verifies that all occurrences of `org.snakeyaml:snakeyaml-engine` are indeed replaced with the patched `org.bitbucket.snakeyaml:snakeyaml-engine:8209bb9484` version from https://jitpack.io/#org.bitbucket.snakeyaml.snakeyaml-engine.src.master/

This compiles and works fine in the emulator with Android 11. @westnordost can you try to reproduce the issue in Android 5?

Issue #3889, opened 2022-03-19
Crash in org.snakeyaml.engine.v2.api.LoadSettingsBuilder


westnordost:
There is a reproducible crash for org.snakeyaml.engine.v2.api.LoadSettingsBuilder (dependency of [kaml](https://github.com/charleskorn/kaml)).

```
java.lang.ExceptionInInitializerError
at org.snakeyaml.engine.v2.api.LoadSettingsBuilder.<init>(LoadSettingsBuilder.java:66)
at org.snakeyaml.engine.v2.api.LoadSettings.builder(LoadSettings.java:79)
at com.charleskorn.kaml.YamlParser.<init>(YamlParser.kt:33)
at com.charleskorn.kaml.Yaml.decodeFromReader(Yaml.kt:51)
at com.charleskorn.kaml.Yaml.decodeFromStream(Yaml.kt:47)
at de.westnordost.streetcomplete.ktx.ResourcesKt.getYamlStringMap(Resources.kt:16)
at de.westnordost.streetcomplete.about.ChangelogFragmentKt$readChangelog$2.invokeSuspend(ChangelogFragment.kt:93)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:39)
at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
Caused by: java.util.regex.PatternSyntaxException: Syntax error in regexp pattern near index 13:
^\$\{\s*((?<name>\w+)((?<separator>:?(-|\?))(?<value>\w+)?)?)\s*\}$
^
at java.util.regex.Pattern.compileImpl(Native Method)
at java.util.regex.Pattern.compile(Pattern.java:411)
[...]

PR #4358, opened 2022-09-06
Upgrade kaml to v0.48.0

FloEdelmann:
[`kaml` v0.48.0](https://github.com/charleskorn/kaml/releases/tag/0.48.0) was just released, which updates `snakeyaml-engine` to v2.4.
[`snakeyaml-engine` v2.4](https://bitbucket.org/snakeyaml/snakeyaml-engine/wiki/Changes) includes the fix to support Android 5 (see #3889).

Thus, we can drop our workaround/patch introduced in #3914.

@westnordost Can you verify that these changes work correctly on Android 5? I tested on an Android 11 emulator and it works fine.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -103,13 +103,6 @@ configurations {
     all {
         // it's already included in Android
         exclude(group = "net.sf.kxml", module = "kxml2")
-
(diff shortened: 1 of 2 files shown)


westnordost:
Nice, one workaround/TODO less

PR #4369, opened 2022-09-11
Suggest last answers in fire hydrant diameter quest

FloEdelmann:
Similar to the building levels quest. Most cities have only a handful of different diameters, and in one street there's usually only one or two different ones. Clicking a button is much faster than entering the value manually each time.


app/src/main/java/de/westnordost/streetcomplete/util/LastPickedValuesStore.kt
```diff
@@ -28,11 +28,11 @@ class LastPickedValuesStore<T : Any>(
     private fun getKey() = Prefs.LAST_PICKED_PREFIX + key
 }
 
-/* Returns the `target` most-common non-null items in the first `historyCount`
+/** Returns the [target] most-common non-null items in the first [historyCount]
```
smichel17, line 31:
```suggestion
/**
 * Returns the [target] most-common non-null items in the first [historyCount]
```
Or indent the other lines below to match

westnordost:
Hmm, I have some reservations against this:

1. Worth the extra space (and UI + implementation complexity)?
  - Building levels had **two** separate input fields instead of just one for the fire hydrant diameter quest. Less time saved here
  - The map was completely **drowned** in building levels quests, so it was much more important there was some kind of shortcut (before the buttons were added, many other ideas how to make it faster were floated around). Not so with the hydrant diameter quest
  - Surveying the building levels information was usually very quick, so quick (and repetitive) that you'd have acquired the information faster than you could input it. With the fire hydrant quest, the gathering of the information usually takes longer than inputting it, because you have to look out for the sign (which you also cannot read while too far away)

2. Choice of UI
Why this UI to auto-complete the one value you got to enter? The line of buttons makes the UI considerably larger and the UI was only chosen for the building levels UI because there are two input fields to auto-fill. When there is just one field, I am not sure that UI makes sense.

3. Repetition of RecyclerView in layouts for different countries
Maybe if there is going to be a common element for all the different layouts, there should be just one main layout and then simply a `inflateIntoThatContainerViewTheFollowingLayout(when(countryCode) { "DE" -> ... })`. (I'm looking to get away from that MCC magic anyway mid/long-term because it is quite intransparent and platform specific)

peternewman:
I don't feel too strongly either way about this, and I've only answered this quest a few times so far, but some general thoughts.

>     1. Worth the extra space (and UI + implementation complexity)?
Although it is a bigger number so a few extra digits to type too.

>     2. Choice of UI
>        Why this UI to auto-complete the one value you got to enter? The line of buttons makes the UI considerably larger and the UI was only chosen for the building levels UI because there are two input fields to auto-fill. When there is just one field, I am not sure that UI makes sense.

It would be much slower to use a drop-down or similar, and an extra click, for the same effect.

>        (You have a smartphone with an extreme aspect ratio - 1:2? - The larger the form, the more from the map is hidden. It may look ok on your device but on other devices, there may be just a slit left for the map.

While I'd certainly appreciate more map space, would putting the presets at the top of the form solve it (assuming you can reach the buttons comfortably), then the mostly irrelevant bottom of the form could remain hidden on smaller screens.

FloEdelmann:
> Building levels had two separate input fields; map was drowned in building levels quests; time to input is not the bottleneck here

As @peternewman said, the numbers are longer for fire hydrant diameters. There aren't as many fire hydrant quests as building levels quests of course, but in my town there's usually one every 40 meters or so. So it's still one of the more frequent quests. Especially when answering the quest with the phone mounted to a bike handlebar (as I often do), entering a diameter via the keyboard requires stopping the bike. I know that's not a common use case though.

> extra space

We could also overlay the buttons over the hydrant sign, that way it won't take up any extra space. Since the button(s) will only be shown anyway after having solved the quest at least once (and the bottom of the signs is faded), nothing important will be hidden.

> Choice of UI

I don't think the UI is difficult to understand, do you? Do you have a better suggestion?

> implementation complexity

Most of the heavy lifting happens in `LastPickedValuesStore` anyway, so I think the effort is worth it.

> Repetition of RecyclerView in layouts for different countries

Yeah, that duplication can likely be avoided.

westnordost:
> I don't think the UI is difficult to understand, do you? Do you have a better suggestion?

The street name quest has a ▼ button next to the input which lets you one of the nearby road names. This would be an idea, it takes up practically no extra space and (thus) does not populate the UI with more buttons (the busier UI, i.e. the more knobs and wheels to turn, the less clear it gets). The only downside is that one extra tap to open the dropdown is required.

riQQ:
The state of the dropdown could be persisted. Then it's one additional tap only once.

app/src/main/res/layout/quest_fire_hydrant_diameter_last_picked_button.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    style="@style/Widget.AppCompat.Button.Small"
```
westnordost, line 4: Independently of the rest, I wonder if there is a reason why we do not use material design theme here? In particular, I guess the material-style to choose here would be the "outlined button"

westnordost (commented):
Otherwise, the code looks spotless

matkoniecz:
@FloEdelmann 

> Clicking a button is much faster than entering the value manually each time.

Are you editing in area where it really helps? In my area I get this question maybe once per editing session (and in Poland so far 100% of hydrants were missing such sign, but that is a different story)

FloEdelmann:
> Are you editing in area where it really helps?

Yes, as I wrote [above](https://github.com/streetcomplete/StreetComplete/pull/4369#issuecomment-1245892626), in my town there's usually a fire hydrant every 40 meters or so. Most are mapped but were lacking the diameter (although that changed because I answered the quest for most of them already). Tapping a button would have been much quicker than having to enter one of the same three common diameters every time.

app/src/main/res/layout/quest_fire_hydrant_diameter_last_picked_button.xml
FloEdelmann: I guess there is no reason, I just used the default style that the android layout designer suggested.

westnordost:
So, did you consider a ▼ button?

FloEdelmann:
I didn't (and still don't) have much time to try to change this. But I'll get to it in the next weeks.

westnordost:
Well, unless you think your current solution is better. But IMO when compared, one tap more is better than having 1-5 buttons more in the UI.

FloEdelmann:
I changed it to a dropdown now. The :small_red_triangle_down: button is shown only when there are suggestions and the input field is empty. I actually like this more than I had expected.

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_diameter/AddFireHydrantDiameterForm.kt
```diff
@@ -79,4 +117,31 @@ class AddFireHydrantDiameterForm : AbstractOsmQuestForm<FireHydrantDiameterAnswe
...
+        popup.show()
+    }
+}
+
+private fun getHydrantDiameterSignLayoutResId(countryCode: String): Int = when (countryCode) {
+    "DE" -> R.layout.quest_fire_hydrant_diameter_sign_de
+    "FI" -> R.layout.quest_fire_hydrant_diameter_sign_fi
+    "NL" -> R.layout.quest_fire_hydrant_diameter_sign_nl
+    "PL" -> R.layout.quest_fire_hydrant_diameter_sign_pl
+    "UK" -> R.layout.quest_fire_hydrant_diameter_sign_uk
```
westnordost, line 145:
```suggestion
    "GB", "IE" -> R.layout.quest_fire_hydrant_diameter_sign_uk
```

Also, what about Belgium (BE)? They used to be defined before. It seems to have the same style as one of the others in this list, but I do not remember which.

westnordost (changes requested):
I like that you think it's better :-)

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_diameter/AddFireHydrantDiameterForm.kt
FloEdelmann: Ah, I thought those two files were duplicated because a country could consist of multiple MCCs. Good that we get rid of that now.

Belgium used the German sign before.

westnordost (approved):

PR #4382, opened 2022-09-15
Improve icons in the Link Collection view

endim8:
improve icons for organicmaps & figuregrounder, use osm icon for openorienteeringmap (better than nothing)

new icons:
exsdfjsbkj


app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -411,7 +411,7 @@ private val links = listOf(
         "https://oomap.co.uk/global/",
         "OpenOrienteeringMap",
         LinkCategory.GOODIES,
-        null,
+        R.drawable.ic_link_openstreetmap,
```
westnordost, line 414:
```suggestion
        R.drawable.ic_link_openorienteeringmap,
```

westnordost:
Hmm to be honest, I like the current figuregrounder-icon better. But I wouldn't mind if that current icon was round instead of a square.

I also noticed that openorienteeringmap has a favicon - I created an SVG of it and added it.

Regarding organic maps, see @FloEdelmann - I think it is better to show 1:1 the app icon

matkoniecz:
> Hmm to be honest, I like the current figuregrounder-icon better.

The same, new one looks like a blob. Have they changed icon?

westnordost:
I'll close this then

PR #4385, opened 2022-09-16
Incline quest where mtb:scale:uphill=* is used

matkoniecz:
useful as otherwise you need high-quality elevation model
to do routing, and in some cases it may be unavailable at all in sufficient detail

I am writing this as I wanted to tag mtb scale info in some places to improve routing
for bicycle trekking where people like me are NOT interested at all in MTB routes
but mtb:scale=1 (or just mtb:scale=0) is still fine for them.

So mtb:scale=1 mtb:scale:uphill=3 is fine downhill but not uphill for such people,
but it is hard to check in which way given path goes uphill

Recommended by wiki: https://wiki.openstreetmap.org/w/index.php?title=Key:mtb:scale&uselang=en#mtb:scale:uphill=0-5

screen13

BTW, if someone cares/is aware about mtb tagging: comments on https://wiki.openstreetmap.org/wiki/Talk:Key:mtb:scale#Is_it_possible_to_have_mtb:scale:uphill=1_on_asphalt_road? and https://wiki.openstreetmap.org/wiki/Talk:Key:mtb:scale#Can_mtb:scale:uphill_be_lower_than_mtb:scale? are welcome as I am confused a bit by this tagging (and it seems a possible candidate for overlays)


app/src/main/res/values/strings.xml
```diff
@@ -653,6 +653,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_barrier_bicycle_type_multiple">Chicane - more than two barriers</string>
     <string name="quest_barrier_bicycle_type_not_cycle_barrier">Not a cycle barrier, but some other barrier</string>
 
+    <string name="quest_bicycle_incline_title">Which direction leads upwards for this way?</string>
```
westnordost, line 656: Or just 

```suggestion
    <string name="quest_bicycle_incline_title">Which direction leads upwards here?</string>
```
? After all, the type of thing is already mentioned in the quest hint label. (Could maybe also delete the special string for the steps quests then)

app/src/main/java/de/westnordost/streetcomplete/quests/incline_direction/AddBicycleIncline.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_bicycle_incline_title
+
+    override fun createForm() = AddInclineForm()
+
+    override fun applyAnswerTo(answer: InclineDirection, tags: Tags, timestampEdited: Long) {
+        tags["incline"] = when (answer) {
+            UP -> "up"
+            UP_REVERSED -> "down"
+        }
```
westnordost, line 60: Could go into an 

`fun InclineDirection.applyTo` extension function, search for  ".applyTo(" in the code to see examples.

app/src/main/java/de/westnordost/streetcomplete/quests/incline_direction/AddBicycleIncline.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
+        // sadly for very long ways shape may be complex and it may be confusing which answer should be given
+        // once multiple quest markers are appearing it becomes completely unclear
+        // see for example https://www.openstreetmap.org/way/437205914
+        return mapData
+            .filter { tagFilter.matches(it) }
+            .filter {
+                val geometry = mapData.getGeometry(it.type, it.id) as? ElementPolylinesGeometry
+                val overlyLong = geometry?.polylines?.filter { it.measuredLength() > OsmQuest.minLengthForMultiMarkers(false) }
```
westnordost, line 41: This is arbitrary. Please just specify a constant here instead of referring to some unrelated function. Also, 800+ meters sounds a bit long for choosing to not display this quest for long ways.

Also, why can the user not simply split up the way in such a case?

westnordost (changes requested):
Looks good to me. A few remarks:

app/src/main/java/de/westnordost/streetcomplete/quests/incline_direction/AddBicycleIncline.kt
matkoniecz: That is actually a related function: if there are multiple markers on a windy paths then it becomes utterly unclear which direction should be actually applied.

screen14

> Also, 800+ meters sounds a bit long for choosing to not display this quest for long ways.

Though I guess that some small distance can be also used, with assumption that it will always have a single marker

>Also, why can the user not simply split up the way in such a case?

It is not entirely obvious (at least I have not thought about it) and in many cases would be done solely as workaround for SC interface so probably should not be encouraged

app/src/main/res/values/strings.xml
matkoniecz: For steps it can be actually useful to strongly focus on steps, but here generic way is definitely pointless

app/src/main/java/de/westnordost/streetcomplete/quests/incline_direction/AddBicycleIncline.kt
westnordost: Why would it be done as a workaround to SC interface?

matkoniecz: In case where incline changes it should be split, and if parts of it are small enough this stops being relevant.

In case where way has continuous this case primary motivation for splitting way would be to get its direction displayed clearly in SC (so: workaround to SC interface)

westnordost: > In case where incline changes it should be split, and if parts of it are small enough this stops being relevant.

Well you have the same for any length of way.

app/src/main/java/de/westnordost/streetcomplete/quests/incline_direction/AddBicycleIncline.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
+        // sadly for very long ways shape may be complex and it may be confusing which answer should be given
+        // once multiple quest markers are appearing it becomes completely unclear
+        // see for example https://www.openstreetmap.org/way/437205914
+        return mapData
+            .filter { element -> tagFilter.matches(element) }
+            .filter { element ->
+                val geometry = mapData.getGeometry(element.type, element.id) as? ElementPolylinesGeometry
+                val overlyLong = geometry?.polylines?.filter { it.measuredLength() > 200 }
+                overlyLong?.size == 0
```
westnordost, line 40:
```suggestion
                geometry?.polylines?.all { it.measuredLength() < 200 } == true
```

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/incline_direction/AddBicycleIncline.kt
westnordost: or `geometry?.polylines?.none{ it.measuredLength() > 200 } == true`

westnordost:
So my standpoint in the [unresolved comment thread](https://github.com/streetcomplete/StreetComplete/pull/4385#discussion_r972970354) is that whether it is kind of impossible to declare what is the incline of a MTB track is pretty much independent of the length of the track in OSM. The track above may just as well be split into many segments already, this doesn't mean that anything on the ground is different.

But the code I dispute does just lead to less quests being displayed, i.e. **it is not a merge-blocker**. 

However, in cases cases were it is not really answerable (because it goes up and down and up and down...), you need to think of how to handle this case because (if that actually occurs), it would not fulfill this point from the guideline:

> 🤔 No unanswerable quests: All generated quests need to be actually answerable (no false-positives). This means that any answer given by the user must result in something being tagged. 

I.e. for cases where the answer is something like "it does not really have a upwards or downwards incline, it's more of a bumpy ride of some kind", there should be an equivalent tagging. Maybe e.g. remove/replace the `mtb:scale:uphill` value with something else here?

matkoniecz:
> Maybe e.g. remove/replace the mtb:scale:uphill value with something else here?

In such case it may not be the best idea, as there is difference between up and down hoops on compacted surface and up and down hoops on surface of loose gravel and loose big stones.

I posted some suggestions on tagging mailing list for feedback.

matkoniecz:
Done. So far I used fairly popular tag and I confirmed that it is actually used in this way by contacting mappers who used it already.

There are some limited voices on tagging mailing list that it should be deprecated and replaced by some other tag: if that happens SC can be easily changed.

PR #4386, opened 2022-09-16
New quest ref for emergency access points

### LINKED ISSUE #4384: Add `ref` for `highway=emergency_access_point` and `emergency=access_point`
### General
Affected tag(s) to be modified/added:
[highway=emergency_access_point](https://wiki.openstreetmap.org/wiki/Tag:highway%3Demergency_access_point)
[highway=access_point](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Daccess_point)
Question asked: **Is there a reference number on this (emergency) access point?**

### Checklist
"Answer: no" would add [noref=yes](https://wiki.openstreetmap.org/wiki/Key:noref)
Regarding wiki, all access points have a number on the sign.
There are "only" around 1400 missing ref's in Germany.
Probably not really worth, but there a lot of similar quests which could be duplicated for this case.
Regarding taginfo there are only 4-5 % of all access points without a ref-number which are worldwide 2.342 issues for this quest.

**Disclaimer: I'm perfectly fine with closing the issue if it doesn't make sense.**

### Ideas for implementation

**Element selection:** 
(highway=emergency_access_point or emergency=access_point) and ref!=*
Overpass Turbo for Germany: https://overpass-turbo.eu/s/1lVd

**UI:**
UI and behaviour could be identical with the bus ref quest or the post_box ref quest.

mcliquid:
* New Quest: "add ref" for `(highway = emergency_access_point or emergency = access_point) and !name and !ref and noref != yes and ref:signed != no and !~"ref:.*"`
* Added quest icon
* Add Option to Delete
* Add other Answer "Is Assembly Point"
  * Remove highway-Tag if set

Fix #4384


app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRef.kt
```diff
@@ -0,0 +1,42 @@
...
+        (
+          (highway = emergency_access_point)
+          or
+          (emergency = access_point)
+        )
+        and !name and !ref and noref != yes and ref:signed != no and !~"ref:.*"
+    """
+
+    override val changesetComment = "Determine emergency access point refs"
+    override val wikiLink = "Tag:emergency=access_point"
```
matkoniecz, line 21:
```suggestion
    override val wikiLink = "Key:ref"
```

Though https://github.com/streetcomplete/StreetComplete/blob/e67d1eb9107949162f5bc6a5b475631189a6d9ca/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/EditType.kt#L18 mentions that t can be also linking edited feature

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
```diff
@@ -0,0 +1,53 @@
...
+            .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ -> applyAnswer(NoAccessPointRef) }
+            .setNegativeButton(R.string.quest_generic_confirmation_no, null)
+            .show()
+    }
+
+    override fun isFormComplete() = ref.isNotEmpty()
+
+    private fun createMarkAsAssemblyPointAnswer(): AnswerItem? {
+        val node = element as? Node ?: return null
+        if (node.tags["emergency"] == "assembly_point") return null
```
matkoniecz, line 47: Note that it would mean that for

`highway = emergency_access_point emergency=assembly_point` this quest would reappear. It is necessary to exclude objects already tagged with `emergency=assembly_point` then

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRef.kt
```diff
@@ -0,0 +1,42 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddAccessPointRef : OsmFilterQuestType<AccessPointRefAnswer>() {
+
+    override val elementFilter = """
+        nodes with
+        (
+          (highway = emergency_access_point)
```
matkoniecz, line 13:
```suggestion
          (highway = emergency_access_point and emergency != assembly_point)
```

to prevent infinite loop if user selects that it is an assembly point (alternative: remove `highway = emergency_access_point` in such case?)

note: untested code

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
matkoniecz: Wait, is this function even being used? See `IsAssemblyPointAnswer` in `applyAnswerTo`

matkoniecz (changes requested):
Please, add authorship info to https://github.com/streetcomplete/StreetComplete/blob/master/res/graphics/authors.txt#L255 (unless icon is placeholder and @westnordost will make a new one)

Also, for icon I suspect that inner "Emergency" can be removed to simplify it and make it more multilanguage-friendly (and making separate icon for each language is not a good idea :) )

matkoniecz:
In general: congratulations and thanks for submission! This quest is a bit borderline with relatively low usage but it fixes something validly reported as a problem by QA tool and requiring in-place survey and adding quest to more remote areas which are a bit quest-starved right now.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -466,4 +467,7 @@ fun questTypeRegistry(
 
     /* at the very last because it can be difficult to ascertain during day. used by OsmAnd if "Street lighting" is enabled. (Configure map, Map rendering, Details) */
     AddWayLit(),
+
+    /* added it at the end because it could be not easy to understand this quest */
+    AddAccessPointRef(),
```
westnordost, line 472: This is not a reason to put it at the end. It should be put according to how far you have to be away from it to be able to answer this.
So, it should be somewhere near AddPostboxRef etc

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRef.kt
westnordost: yes, this is fine

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRef.kt
```diff
@@ -0,0 +1,42 @@
...
+          or
+          (emergency = access_point)
+        )
```
westnordost, line 16:
```suggestion
        (
          highway = emergency_access_point
          or emergency = access_point
        )
```
no need for the additional brackets

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRef.kt
```diff
@@ -0,0 +1,42 @@
...
+        and !name and !ref and noref != yes and ref:signed != no and !~"ref:.*"
+    """
+
+    override val changesetComment = "Determine emergency access point refs"
+    override val wikiLink = "Tag:emergency=access_point"
+    override val icon = R.drawable.ic_quest_access_point
+    override val achievements = listOf(EditTypeAchievement.LIFESAVER)
```
westnordost, line 23:
```suggestion
    override val achievements = listOf(EditTypeAchievement.LIFESAVER, EditTypeAchievement.RARE)
```

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    override val contentLayoutResId = R.layout.quest_ref
+    private val binding by contentViewBinding(QuestRefBinding::bind)
+
+    override val otherAnswers get() = listOfNotNull(
+        AnswerItem(R.string.quest_ref_answer_noRef) { confirmNoRef() },
+        createMarkAsAssemblyPointAnswer()
+    )
+
+    private val ref get() = binding.refInput.text?.toString().orEmpty().trim()
```
westnordost, line 23: Rather use
```suggestion
    private val ref get() = binding.refInput.nonBlankTextOrNull
```
and correct other places accordingly

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
westnordost, line 47: Why not always offer this option? Why do you check if it is a node?

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    override fun isFormComplete() = ref.isNotEmpty()
+
+    private fun createMarkAsAssemblyPointAnswer(): AnswerItem? {
+        val node = element as? Node ?: return null
+        if (node.tags["emergency"] == "assembly_point") return null
+
+        return AnswerItem(R.string.quest_accessPointRef_answer_assembly_point) {
+            applyAnswer(IsAssemblyPointAnswer)
```
westnordost, line 50: This should also show a confirmation dialog, one that shows a picture like can be seen here https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dassembly_point

app/src/main/res/values/strings.xml
```diff
@@ -1551,4 +1551,7 @@ Partially means that a wheelchair can enter and use the restroom, but no handrai
     <string name="overlay_sidewalk">Sidewalks</string>
 
     <string name="overlay_street_parking">Street parking</string>
+
+    <string name="quest_accessPointRef_title">What’s the identification number of this?</string>
```
westnordost, line 1555: I find "here" better than "of this". @peternewman what do you think?

westnordost (changes requested):
The "Emergency" should be deleted from the icon. It is too small

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
matkoniecz: How it should be handled in `applyAnswer(AccessPointRef(ref))`? 

`applyAnswer(AccessPointRef(ref!!))`

?

`!!` should be avoided, but admittedly it is not worse than current situation where bug could in theory let empty string end there,

westnordost: yes, !!

matkoniecz: BTW, I applied it to the current code - including one which was a base for this quest

in https://github.com/streetcomplete/StreetComplete/commit/0aaffb13d359714036271a3c212d6c65772321c4

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
westnordost: IMO, just always offer this answer option, so the implementation is the same as for that other answer (no ref at all)

app/src/main/res/values/strings.xml
westnordost: "What’s the identification number **here**?", not "of here"

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
westnordost: AddMaxspeedForm which shows the living street icon

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
westnordost: See https://github.com/streetcomplete/StreetComplete/pull/4386#discussion_r973759355

app/src/main/res/values/strings.xml
peternewman: Postbox has:
`What’s the identification number of this collection box?`

Which does feel better, as it's not really a place as such, I'm assuming it's lots of different things (well two currently), which would be shown in the description row already.

Plus the old number, when it could be letters too, but we had that for postboxes too.

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRef.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_accessPointRef_title
+
+    override fun createForm() = AddAccessPointRefForm()
+
+    override fun applyAnswerTo(answer: AccessPointRefAnswer, tags: Tags, timestampEdited: Long) {
+        when (answer) {
+            is NoAccessPointRef -> tags["noref"] = "yes"
+            is AccessPointRef ->   tags["ref"] = answer.ref
+            is IsAssemblyPointAnswer -> {
```
peternewman, line 33: Some of these can have ref's too, e.g. letters, like everyone from one building or team go to A, another B etc:
https://www.manutan.co.uk/img/S/GRP/ST/AIG3603434.jpg

Although not all, and it's probably less common/only bigger sites.

res/graphics/authors.txt
```diff
@@ -253,6 +253,7 @@ italian police type/
     polizia.svg                     modified from https://commons.wikimedia.org/wiki/File:Italian_traffic_signs_-_icona_polizia.svg, Gigillo83, License: CC-BY-SA 3.0
 
 quest/
+    access_point.svg
```
peternewman, line 256: Shouldn't this be called something more generic like emergency, or descriptive like plus/cross, given the same thing would also be appropriate for first aid or other emergency quests.

More generally what about adding a map pin to the centre to make it a bit more specific?

westnordost: No, icon is fine

peternewman: Fair enough, is it worth renaming it though to be more descriptive like most of the others are?

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
westnordost: Sorry, I've looked at the implementation in `AddMaxSpeed` and it is different than I remembered after all. What you need to do is to display a custom layout in the `AlertDialog` that asks the "are you sure..." question.

This custom layout (defined in to be created `dialog_assembly_point.xml` in the `res/layout` folder would only contain an `ImageView` that shows that assembly point feature and a `TextView` that displays the text like "bla bla... looks something like this?".

The method you are looking for is `setView(layoutResId: Int)` for `AlertDialog.Builder`. E.g.

```kotlin
    private fun confirmAssemblyPoint() {
        AlertDialog.Builder(requireContext())
            .setTitle(R.string.quest_generic_confirmation_title)
            .setView(R.layout.dialog_assembly_point)
            .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ -> applyAnswer(IsAssemblyPointAnswer) }
            .setNegativeButton(R.string.quest_generic_confirmation_no, null)
            .show()
    }
```

westnordost:
But then this PR wouldn't be yours anymore. I added a hint how to do it in the comment, maybe this will help you.

app/src/main/res/values/strings.xml
```diff
@@ -1551,4 +1551,8 @@ Partially means that a wheelchair can enter and use the restroom, but no handrai
     <string name="overlay_sidewalk">Sidewalks</string>
 
     <string name="overlay_street_parking">Street parking</string>
+
+    <string name="quest_accessPointRef_title">What’s the identification number here?</string>
+    <string name="quest_accessPointRef_answer_assembly_point">It’s an Assembly Point</string>
+    <string name="quest_accessPointRef_detailed_answer_impossible_confirmation">An Emergency Assembly Point is where people assemble in case of an emergency. Mostly it’s signed like this:</string>
```
westnordost, line 1557:
```suggestion
    <string name="quest_accessPointRef_detailed_answer_impossible_confirmation">An Emergency Assembly Point is where people assemble in case of an emergency. It usually looks similar to this:</string>
```

app/src/main/res/layout/dialog_confirm_assembly_point.xml
```diff
@@ -0,0 +1,23 @@
...
+        android:text="@string/quest_accessPointRef_detailed_answer_impossible_confirmation"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textAppearance="@android:style/TextAppearance.Theme.Dialog" />
+
+    <ImageView
+        android:id="@+id/assemblyPointImage"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        tools:src="@drawable/ic_assembly_point"
```
westnordost, line 20:
```suggestion
        android:src="@drawable/ic_assembly_point"
```
`tools` namespace is only shown in the IDE, not in the app

app/src/main/res/drawable/ic_assembly_point.xml
```diff
@@ -0,0 +1,49 @@
...
+      android:fillType="evenOdd"/>
+  <path
+      android:pathData="M14.76,169.1 L8.9,163.5 35.28,137H16.71L24.53,129.05H48.95V152.9L41.14,160.85V142.59L14.76,169.1"
+      android:fillColor="#ffffff"
+      android:fillType="evenOdd"/>
+  <path
+      android:pathData="M163.24,169.1 L169.1,163.51 142.72,137H161.28L153.47,129.05H129.05V152.91L136.86,160.86V142.6L163.24,169.1"
+      android:fillColor="#ffffff"
+      android:fillType="evenOdd"/>
+</vector>
```
westnordost, line 49: Some older versions of Android do not support `fillType=evenOdd`

```suggestion
<vector xmlns:android="http://schemas.android.com/apk/res/android"
    android:width="400dp"
    android:height="400dp"
    android:viewportWidth="178"
    android:viewportHeight="178">
    <path
        android:pathData="M0,0h178v178h-178z"
        android:fillColor="#fff"/>
    <path
        android:pathData="M5,5h168v168h-168z"
        android:fillColor="#008855"/>
    <path
        android:pathData="M44.9,117.94 L44.9,88.84C44.9,86.32 47.1,84.64 49.4,84.64H76.7C76.6,88.08 79,90.5 80.9,93.04H74C69,93.04 66.5,96.15 66.5,101.74V117.94H56.9V96.04H52.7V117.94zM70.7,127.84 L70.7,101.44C70.7,98.41 72.7,96.94 75.5,96.94H102.5C105.8,96.94 107.3,99.73 107.3,101.14V127.84H101.6V106.54H97.4V127.84H80.9V106.54H76.4V127.84zM133.1,117.94 L133.1,90.04C133.1,86.99 131.5,84.64 128.6,84.64H101.3C101.3,87.95 99,90.54 97.1,93.04H103.7C108.4,93.04 111.5,97.12 111.5,101.74V117.94H121.1V96.04H125.3V117.94z"
        android:fillColor="#fff"/>
    <path
        android:pathData="M97.86,58.49A8.85,8.85 0,0 1,80.15 58.49,8.85 8.85,0 1,1 97.86,58.49z"
        android:fillColor="#fff"/>
    <path
        android:pathData="M122.54,71.07A8.86,8.86 0,0 1,104.83 71.07,8.86 8.86,0 1,1 122.54,71.07z"
        android:fillColor="#fff"/>
    <path
        android:pathData="M72.28,70.69A8.86,8.86 0,0 1,54.57 70.69,8.86 8.86,0 1,1 72.28,70.69z"
        android:fillColor="#fff"/>
    <path
        android:pathData="M97.7,83.41A8.86,8.86 0,0 1,79.99 83.41,8.86 8.86,0 1,1 97.7,83.41z"
        android:fillColor="#fff"/>
    <path
        android:pathData="M14.76,8.9 L8.9,14.5 35.28,41H16.71L24.53,48.95H48.95V25.1L41.14,17.15V35.4L14.76,8.9"
        android:fillColor="#fff"/>
    <path
        android:pathData="M163.24,8.9 L169.1,14.49 142.72,41H161.28L153.47,48.95H129.05V25.09L136.86,17.14V35.4L163.24,8.9"
        android:fillColor="#fff"/>
    <path
        android:pathData="M14.76,169.1 L8.9,163.5 35.28,137H16.71L24.53,129.05H48.95V152.9L41.14,160.85V142.59L14.76,169.1"
        android:fillColor="#fff"/>
    <path
        android:pathData="M163.24,169.1 L169.1,163.51 142.72,137H161.28L153.47,129.05H129.05V152.91L136.86,160.86V142.6L163.24,169.1"
        android:fillColor="#fff"/>
</vector>
```

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
```diff
@@ -0,0 +1,55 @@
...
+    private fun createMarkAsAssemblyPointAnswer(): AnswerItem? {
+        return AnswerItem(R.string.quest_accessPointRef_answer_assembly_point) {
+            AlertDialog.Builder(requireContext())
+                .setTitle(R.string.quest_generic_confirmation_title)
+                .setView(R.layout.dialog_confirm_assembly_point)
+                .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ -> applyAnswer(IsAssemblyPointAnswer) }
+                .setNegativeButton(R.string.quest_generic_confirmation_no, null)
+                .show()
+        }
+    }
```
westnordost, line 54: Can you please have the code in the same way as for the "confirmNoRef" answer? (No extra function to create the answer, but an extra function for confirming it). For consistency.

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    private fun confirmNoRef() {
+        AlertDialog.Builder(requireContext())
+            .setTitle(R.string.quest_generic_confirmation_title)
+            .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ -> applyAnswer(NoAccessPointRef) }
+            .setNegativeButton(R.string.quest_generic_confirmation_no, null)
+            .show()
+    }
+
+    private fun AssemblyPointAnswer() {
```
westnordost, line 43: please rename to `confirmIsAssemblyPoint` or similar

westnordost (changes requested):
Looks good now, just one thing

westnordost:
>https://user-images.githubusercontent.com/3351668/192159392-9a8e0b09-6a84-461e-8c7d-b78abb34fcf5.png

"An emergency [...] is where people assemble in case of an emergency" sounds doppelt-gemoppelt. How about leaving out the first "emergency"? Would be more consistent with the other string, too.

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/access_point_ref/AddAccessPointRefForm.kt
westnordost, line 43:
```suggestion
        AnswerItem(R.string.quest_accessPointRef_answer_assembly_point) { confirmAssemblyPoint() }
    )

    private val ref get() = binding.refInput.nonBlankTextOrNull

    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
        super.onViewCreated(view, savedInstanceState)

        binding.refInput.doAfterTextChanged { checkIsFormComplete() }
    }

    override fun onClickOk() {
        applyAnswer(AccessPointRef(ref!!))
    }

    private fun confirmNoRef() {
        AlertDialog.Builder(requireContext())
            .setTitle(R.string.quest_generic_confirmation_title)
            .setPositiveButton(R.string.quest_generic_confirmation_yes) { _, _ -> applyAnswer(NoAccessPointRef) }
            .setNegativeButton(R.string.quest_generic_confirmation_no, null)
            .show()
    }

    private fun confirmAssemblyPoint() {
```

HolgerJeromin:
The screenshot shows "highway milestone".
Is that a bug?

PR #4404, opened 2022-09-22
Updated the steps handrail quest icon the make the background a real circle

wjtje:
Currently the steps handrail quest icon has a different background shape as other quests.


westnordost:
Oh, weird. The android drawable seems correct though, no idea how this happened

PR #4406, opened 2022-09-23
Cycle barrier installation quest

mnalis:
Closes https://github.com/streetcomplete/StreetComplete/issues/4293

The code is implemented and seems to compile and work fine (debug apk [available here](https://github.com/mnalis/StreetComplete/actions/runs/3109721535)); however:

- [x] probably should change to 2 pictures per row, instead of 3 pictures
- [x] needs better pictures (current ones are from [wiki](https://wiki.openstreetmap.org/wiki/Key:cycle_barrier:installation) and as noted are [not really good match](https://github.com/streetcomplete/StreetComplete/issues/4293#issuecomment-1237507377) (also, we miss image for `foldable` cycle barrier at all, I took zoomed-in part of one from bollard as an example), with `authors.txt` update.
- [x] needs shorter and better texts for answers (as they are overlaying significant parts of pictures)
- [ ] it needs [an icon](https://github.com/streetcomplete/StreetComplete/issues/4293#issuecomment-1230491478) (currently uses `ic_quest_no_bicycles`

So if someone could find/take/produce square images of the features, that would be greatly appreciated, as well as any comments on the other issues.

Currently it looks like this (so you can see above issues, e.g. text covering important parts of the images, etc.):
small_Screenshot_20220923_033307_de westnordost streetcomplete mn debug


westnordost:
there are like... 2 `foldable` values, not documented. What is the use case for even recording the difference between foldable, openable and removable? How is a surveyor supposed distinguish between those? (Why is it not simply removable/openable or not?)

mnalis:
Sure, it seems rare enough (at the moment), so we can drop it if you like, and have user leave a note instead. 

`foldable` is documented at https://wiki.openstreetmap.org/wiki/Key:cycle_barrier:installation though (albeit without a picture) as:
> Barrier can be folded down with a hinge at its base.  

I cannot say what is the significant **_difference in use_** as I was not involved in tag creation (if it was up to me, both for `bollard` and for `cycle_barrier` it would be simple tag like `permanent=yes/no` and that is it - I too do not see much usability between mapping if something is foldable vertically or swingable horizontally or removable or slideable or... but it seems people involved though it was better to differentiate them).

But we seem to ask that difference at `bollard type` quest too:
https://github.com/streetcomplete/StreetComplete/blob/a4d68fecc300b1062b9aaa0b961ebe847af9f598/app/src/main/java/de/westnordost/streetcomplete/quests/bollard_type/BollardType.kt#L5-L11

Let me know how you'd prefer to handle this one (it's all the same to me, this is just a first draft).

westnordost:
bollard is an established tagging scheme already, `cycle_barrier:installation` is not. It is not too late to question the sense of some values. 
So far, usage seems to have been pushed mostly by @SupaplexOSM - maybe he can answer that.

SupaplexOSM:
When I proposed `cycle_barrier:installation`, I oriented the variance of it's values on the [bollard](https://wiki.openstreetmap.org/wiki/Key:bollard) values - with the exception that there is a value "openable" for cycle barriers, which would make no sense for bollards. Semantically, one could perhaps argue that "removable" is sufficient besides "fixed" as the only second category, but since "removable" is already distinguished from "foldable" and other forms of "removing" for bollards, I thought it is consistent to distinguish such forms for cycle barriers as well (as similar/the same mechanisms are used here).

Currently, the values "removable" and "openable" [are used fairly in the same quantity](https://taginfo.openstreetmap.org/keys/cycle_barrier:installation#values) - but I don't know if this is also due to a "wrong" understanding of the term "removable". However, I think these values should appear in the quest/should be distinguished from each other to keep consistency with the bollard key, but others ("foldable") are not needed because of their rarity.

legofahrrad:
> - [ ] needs shorter and better texts

To be consistent with the bollard type quest, it should be "What type of cycle barrier is this?". But we have exactly the same question for the cycle barrier type (single/double/triple/tilted/...). Could both quests have the same question or would this be confusing and misleading to users?

westnordost:
It would be confusing.

mnalis:
Ok, I've removed `foldable` and updated other pictures... How does this look? Suggestions (about pictures, or texts, or ...) ?

small_Screenshot_20221007_232017_de westnordost streetcomplete mn debug

westnordost:
Hm, the pictures are practically worthless and helping the surveyor see what he should look for. It may be better to not show the whole bicycle barrier but just the part that matters, e.g. the hinge etc.

For (fixed) bicycle barriers, isn't there already a picture (from the other bicycle barrier quest)?

mnalis:
> Hm, the pictures are practically worthless and helping the surveyor see what he should look for

Hm, I was aiming to showing concrete example how it might usually look like so it would be instantly recognizable, and providing info about _"what they should look for"_ via text overlay. But perhaps that is the wrong approach.

> It may be better to not show the whole bicycle barrier but just the part that matters, e.g. the hinge etc.

Something like this mockup would be more OK with you then @westnordost? If yes, I can do that. While I certainly see some advantages there, I'd like other opinions which version seems more understandable to novice user.
 
mockup_detail2


> For (fixed) bicycle barriers, isn't there already a picture (from the other bicycle barrier quest)?

Yes, there is. I could use for example this one, if you think that would be better (and save few kB) ?
barrier_bicycle_barrier_triple

>  Also, the space is very inefficiently used

It used to be 3 rows, but I have reduced it to two rows since we had 4 answers which both fit 2x2 answers, and provide somewhat larger images to hopefully make detail little easier to see. Now that we're back to 3 answers, I can certainly revert to 3 rows in one line, if you find that bigger pictures are more of a distraction then help.

> Basically, you have absolute freedom in how the items are being displayed

As it is, that absolute freedom (in my case) is quite limited by my technical abilities (i.e. I can copy and paste from other quest if it is not too complicated, but I can not really invent what I would like - e.g. even seemingly easy changes like having text on upper part of 3x1 image selection [instead of on lower part] are likely beyond my current abilities).

legofahrrad:
This one is also a good photo for `openable` (hinged upwards), when you crop it, IMO: https://commons.wikimedia.org/wiki/File:Cycle_barrier_single_removable_detail.jpg

mnalis:
Thanks! Might have helped finding it if it was **_called_** `openable` instead of `removable` in file name :smile: 

Perhaps showing both might help like in this mockup? Or is it more confusing than helping?
mockup_detail3s

westnordost:
https://github.com/streetcomplete/StreetComplete/pull/4406#issuecomment-1272387969 is better. For consistency, should also show the fixed bicycle barrier from close up.

Note about with or without key is necessary? Because, what if you need a key for openable too? I do not like your last mockup with swing gate + hinged upwards in one picture. Also, is it really necessary to mention how it can be opened?

mnalis:
> [#4406 (comment)](https://github.com/streetcomplete/StreetComplete/pull/4406#issuecomment-1272387969) is better. For consistency, should also show the fixed bicycle barrier from close up.

ok, will try that.

> Note about with or without key is necessary? 

I don't know. It does not seem _that important_ to me. It was copied from `What type of bollard is this?` quest:
small_Screenshot_20221010_000738_de westnordost streetcomplete mn debug
I think that was related to possible confusion of users who might not call something removable if it is locked (or something like that). I'm more than happy to reduce the text here; but if it is unwanted here, we _might_ also consider if there is benefit for those texts in `bollard type` quest either.

> Because, what if you need a key for openable too? 

yeah, that one would definitely get too long with both explanations.

> I do not like your last mockup with swing gate + hinged upwards in one picture.

It's ok, I'm not a big fan of it myself (although I'm sure someone with more graphic sense could make a much nicer version). I was experimenting because @legofahrrad seems to thinks both pictures should be shown to users or else they might not realize that picture of horizontal-swing `openable` barrier might **also** represent vertical-lift `openable` barrier (and I'd like to _avoid having two different answers with same tagging_, as that caused issues in the past: https://github.com/streetcomplete/StreetComplete/issues/4293#issuecomment-1256406384)

> Also, is it really necessary to mention how it can be opened?

See above (especially if the picture shows **_only one_** `openable` barrier) the text should give user the clue that it can be openable **both** in `horizontal` and `vertical` directions. If the picture showed both version, we could probably safely remove that explanation from the text.

Extra texts are also there to address @SupaplexOSM [concern](https://github.com/streetcomplete/StreetComplete/pull/4406#issuecomment-1256450885) that `openable` / `removable` distinction might not be entirely clear to users.

But I myself am content also with terse `fixed` / `openable` / `removable` texts, if that would be preferable ?

mnalis:
OK, here is [test APK](https://github.com/mnalis/StreetComplete/actions/runs/3229730590), which compiles and works fine for me. Does it now looks better @westnordost ? I've done:

- 3 images per row to minimize screen estate usage
- show only details of all 3 options
- short descriptions

small_Screenshot_20221011_230122_de westnordost streetcomplete mn debug

westnordost:
Yes, it looks better! Is this ready for review?

mnalis:
Ready for review now. Note that it still misses a different icon (currently it reuses `ic_quest_no_bicycles`)

westnordost:
Any ideas for an icon?

mnalis:
Nothing particularly good...  Only that it might be based on `ic_quest_no_bicycles` and expand on it? 

(However there is no precedent - for `What type of bollard is this?` we had **_only one quest_** so _"forbidden car"_ was a nice choice. However, for bicycle we have **_two such quests_** `What type of bicycle barrier is this?` (which already uses  _"forbidden bicycle"_ icon) and now this one `Can this cycle barrier be opened?`)

If we were to expand on `ic_quest_no_bicycles`, problem is that it is already quite detailed, so not much additional graphics would fit... maybe just add `2` somewhere on it to indicate it is additional quest to that?

Dunno, I'm not really good with that artistic stuff :smiling_face_with_tear: , perhaps somebody else have an idea?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_installation/AddBicycleBarrierInstallation.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class AddBicycleBarrierInstallation : OsmFilterQuestType<BicycleBarrierInstallationAnswer>() {
+
+    override val elementFilter = "nodes, ways with barrier = cycle_barrier and cycle_barrier and cycle_barrier != tilted and !cycle_barrier:installation"
+    override val changesetComment = "Specify cycle barrier installation"
+    override val wikiLink = "Key:cycle_barrier:installation"
+    override val icon = R.drawable.ic_quest_no_bicycles
+    override val isDeleteElementEnabled = true
+
+    override val achievements = listOf(BLIND, WHEELCHAIR, BICYCLIST, LIFESAVER)
```
westnordost, line 22: Please remove BLIND and WHEELCHAIR

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_installation/AddBicycleBarrierInstallation.kt
```diff
@@ -0,0 +1,37 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BLIND
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.WHEELCHAIR
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.LIFESAVER
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddBicycleBarrierInstallation : OsmFilterQuestType<BicycleBarrierInstallationAnswer>() {
+
+    override val elementFilter = "nodes, ways with barrier = cycle_barrier and cycle_barrier and cycle_barrier != tilted and !cycle_barrier:installation"
```
westnordost, line 16: Please format this into a multiline string for better readability

westnordost (changes requested):
Code looks good, two nitpicks

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_installation/AddBicycleBarrierInstallation.kt
mnalis: OK, if you wish. Note: those were copied from somewhat similar `AddBarrierType.kt`, and made sense to me - e.g. `cycling barrier` will stop wheelchair user as efficiently (or more so) as `bollard` will...

mnalis: removed in 1f97b2b2dd024aa2d82315a3324364998494a19a

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_bicycle_barrier_installation/AddBicycleBarrierInstallation.kt
mnalis: done

legofahrrad:
> grafik

How should the average SC user know that the “hinge to the top” is also tagged as `openable`? The photo shows some kind of swing gate, which is also correct for `openable` but I guess there will come many notes, if users see “hinge to the top” cycle barriers but no matching picture in the app.
However, it looks really good all in all :)

mnalis:
> How should the average SC user know that the “hinge to the top” is also tagged as `openable`?

I'm guessing from the word itself. `Openable` => `Something that can be opened`. The _exact_ method of opening (is it horizontal swing, vertical pivot, sliding barrier, vertical pull, retractable parts etc.) is just extra detail. 
If in other languages the equivalent of `openable` is more problematic, that can be solved in [POeditor translations](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app).

And if Notes start popping up asking for clarification (it can be checked a month of few after this gets released, for example on https://ent8r.github.io/NotesReview/), one can open a new issue so it can be improved. Feel free to do so then @legofahrrad! (I'd be happy to help with guidance on how to do it)

Because, if we insist on _100% absolute perfection on the first try_, the PR will simply never get released :smiley_cat:

westnordost (approved):

paulklie:
Was thinking about creating a unique icon for this quest, to allow better differentiation to the other bicycle barrier quest.

What do you think about this:
image

westnordost:
More than two elements and on top of this basically three layers that cast a shadow onto another make it look a bit busy. Try looking at it in a small size

paulklie:
Hmm, yea I agree.
How about this one:
image
I changed the background colour, because, while it is about cycling barriers, it is not really relevant to cycling (but mostly about emergency vehicles).

westnordost:
> mostly about emergency vehicles

Hm, you are on to something there

paulklie:
How about only changing the background color to yellow?

riQQ:
@paulklie please start new issues for stuff like this. In the new issue, you can link to prior discussion and related issues.

westnordost:
Well, how about my suggestion? (`res/graphics/building/fire_truck.svg`)

PR #4408, opened 2022-09-23
fix conversion of tiles from/to bboxes (#4125)

westnordost:
@Helium314 this should be correct now. Does it check out for #4125?


Helium314:
Looks good, makes more sense than my workaround attempt, and all tests pass.

PR #4413, opened 2022-09-26
Skip AcceptCards for POIs with access=private/no/customers

HolgerJeromin:
This skips the quest for a few hundreds cantines (=`no`/`private`) and embedded restaurants (=`customers` like in waterparks).


matkoniecz:
Have you tested the code is it working as expected (hides some case that should be hidden, otherwise works without change)?

HolgerJeromin:
> Why? Waterparks are usually large enough that when you are in the area, you will already be able to answer this because you are a customer.

In a waterpark or other customer based stuff you will perhaps paying by a different payment system (own rfid). 

But yes, removed customer again.

HolgerJeromin:
> Have you tested the code is it working as expected (hides some case that should be hidden, otherwise works without change)?

Yes, tested.
Thanks for merging.

PR #4419, opened 2022-09-27
link OSM Wiki tag pages in undo view

matkoniecz:
I researched situation to ensure that all tags added/modified by SC will have the target links being present, at least in the form of automatically generated compound pages (some removed tags may not have wiki pages - I think it is fine but I can research also this).

screen01
screen

implements #3442


app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
```diff
@@ -151,18 +152,30 @@ class UndoDialog(
...
+fun StringMapEntryChange.tagString(): String {
+    val keyText = key
+    val escapedKey = Html.escapeHtml(keyText)
+    val valueText = when (this) {
+        is StringMapEntryAdd -> value
+        is StringMapEntryModify -> value
+        is StringMapEntryDelete -> valueBefore
+    }
+    val escapedValue = Html.escapeHtml(valueText)
+    val shownKey = "<a href=\"https://wiki.openstreetmap.org/wiki/Key:$escapedKey\">$escapedKey</a>"
```
westnordost, line 172: What if that key is not documented? Surely some keys that StreetComplete uses are not documented? Maybe?

app/src/main/java/de/westnordost/streetcomplete/osm/Freeform.kt
```diff
@@ -0,0 +1,60 @@
...
+
+/** Returns true if value is from some well defined set of values
+ *  Return false otherwise
+ *  Note that it returns false where StreetComplete is using a limited subset of values
+ *  even if tag is effectively freeform for OSM mappers in general
+ *  For example building=* is not considered as freeform.
+ *  Note that freeform tags where SC is using dedicated subset of values but tag is completely freeform
+ *  for example source:maxheight where sole value set by StreetComplete is "ARCore"
+ * */
+fun isTagValueFromWellDefinedSet(key: String, value: String): Boolean {
```
westnordost, line 11: Wrong thinking? Shouldn't this function be called isTagValueNOTFromWellDefinedSet?

(But then, better call is `isFreeformTagValue`)

app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
matkoniecz: > What if that key is not documented? Surely some keys that StreetComplete uses are not documented? Maybe?

Every single key used by StreetComplete is documented. I verified that as part of doing that taginfo listing (and created some tag pages and or improved their documentation).

That took some time, but was worth doing anyway (for example OSM Wiki pages about surfaces are now better matching both reality, OSM mapping consensus and StreetComplete).

Some **removed** tags may not have wiki pages - I think it is fine but I can research also this.

app/src/main/java/de/westnordost/streetcomplete/osm/Freeform.kt
matkoniecz: I initially named it this way: but for example `building` is effectively freeform in OSM but for SC purposes it is a bit different and individual values can be linked as SC is using specific set of values.

While `opening_hours` is not freeform but linking its values is doomed to fail.

Maybe it should be rather `isTagValueEditedByStreetCompleteLinkableToWiki` or something similarly specific to how it is being used?

app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
matkoniecz: And if new tags would become used by SC without having own wiki page then I plan to fix this. Likely by creating stub pages, maybe by changing SC where applicable like https://github.com/streetcomplete/countrymetadata/commit/f9ec20a28a1dab6fb3f877e8891a7ef322ddb68c (I found this bug exactly because I was checking whether all possible links are actually existing)

app/src/main/java/de/westnordost/streetcomplete/osm/Freeform.kt
westnordost: `isTagValueLinkableToWiki` would be fine. But again, the function currently returns the exact opposite of what the function sounds like then.

westnordost:
I am really not really convinced that it makes sense to link to the values, because:
- this is a list that only excludes the tags that should **not** be tagged. This is begging to break for cases that haven't been thought of yet and for future quests where the **tag value** does not have an own wiki page
- in any case, this will produce some maintenance work and whenever maintenance work is due, it is because something broke, not something is not supported (because of the above)
- it doesn't really make sense to have a wiki page for each tag value, even if it is not free form, e.g. `bench=yes`. The documentation will be in `Key:bench`. To create wiki pages as stubs just so they can be linked ... sounds like busywork with no benefit

I'd suggest to remove linking to values altogether and only replace it with something auto-generated from the OSM wiki. (E.g. maybe there is an API for the mediawiki or it can be scraped for which tags exist wiki pages. taginfo must do it somehow, iD must do it somehow).

matkoniecz:
> I'd suggest to remove linking to values altogether and only replace it with something auto-generated from the OSM wiki. (E.g. maybe there is an API for the mediawiki or it can be scraped for which tags exist wiki pages. taginfo must do it somehow, iD must do it somehow).

It definitely exists. I even have it in https://github.com/matkoniecz/Zazolc/blob/taginfo/buildSrc/src/main/java/UpdateTaginfoListingTask.kt#L270

The problem here is that without being aware which values are linkable it is necessary to fetch every single wiki page and it takes quite a lot of time what is painful

It is also possible to piggy back on https://github.com/matkoniecz/openstreetmap-website/blob/master/config/wiki_pages.yml - here main negative is that it is rarely updated (but that is not a big problem, keys will be still linked)

> This is begging to break for cases that haven't been thought of yet and for future quests where the tag value does not have an own wiki page

Though in such case failure is quite a soft failure anyway.

> it doesn't really make sense to have a wiki page for each tag value, even if it is not free form, e.g. bench=yes

`=yes` `=no` values are excluded from being linked as pointless

westnordost:
> It is also possible to piggy back on https://github.com/matkoniecz/openstreetmap-website/blob/master/config/wiki_pages.yml - here main negative is that it is rarely updated (but that is not a big problem, keys will be still linked)

This sounds like a good source. Even if it is not updated often, so what. The file is huge, bug it could be reduced to only contain the keys (and not values) within `en:`. Is this the outdated-ness the only reason why you didn't use this as a source?

> =yes =no values are excluded from being linked as pointless

And =limited, =separate etc etc, because they are all documented on the page for `Key`.

Note that StreetComplete tags all kinds of tags out of its control because it uses the iD presets and NSI presets. So some automatically created solution sounds more sound to me.

matkoniecz:
> Is this the outdated-ness the only reason why you didn't use this as a source?

Primarily. And also need to rely on more third part stuff that can be randomly broken at any time.

So I guess that I should switch to it.

westnordost:
Alternatively, you can just remove the "link to tag values" and go for linking to keys-only. Linking to values could optionally be added at a later time. (As I wrote, I don't think it is so important).

matkoniecz:
And typical well written wiki page will list values. Sometimes they may be a bit tricky to locate or page may be poorly written, so maybe this can be added later. Though given requirement of extra third-party dependency...

app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
```diff
@@ -151,18 +151,25 @@ class UndoDialog(
...
+    val valueText = when (this) {
+        is StringMapEntryAdd -> value
+        is StringMapEntryModify -> value
+        is StringMapEntryDelete -> valueBefore
+    }
+    val escapedValue = Html.escapeHtml(valueText)
+    val keyText = key
+    val escapedKey = Html.escapeHtml(keyText)
+    val shownKey = "<a href=\"https://wiki.openstreetmap.org/wiki/Key:$escapedKey\">$escapedKey</a>"
+    return "$shownKey = $escapedValue"
```
FloEdelmann, line 172:
```suggestion
    val escapedKey = Html.escapeHtml(key)
    val escapedValue = Html.escapeHtml(valueText)
    val keyLink = "<a href=\"https://wiki.openstreetmap.org/wiki/Key:$escapedKey\">$escapedKey</a>"
    return "$keyLink = $escapedValue"
```

westnordost (approved):

PR #4423, opened 2022-09-27
How is drinking water provided here?

matkoniecz:
implements drinking water side of #4390 (but `amenity=fountain` part remains and may be or may not be a good idea to implement it)

I see value in mapping whether given amenity=drinking_water is a proper tap or just providing miserable jet of water making impossible to fill water bottle.

Also, I noticed that many bare amenity=drinking_water are in various ways quite likely to be problematic (misplaced, used for water taps without drinkable water or inaccessible to public and so on)

see https://github.com/streetcomplete/StreetComplete/pull/4423#issuecomment-1260759795 for the next interface version with grammar fixed and better well support

screen02


mnalis:
I would additionally **_most definitely distinguish_** between those two:

manual_pump  200px-Radbrunnen_in_St _Anna

- one on the left I'd call _"manual pump"_ or _"hand pump"_ (and tag it with `pump=manual`+`man_made=water_well`) (and **not** as a "water well" as currently this PR calls it)
- one the right I'd actually call a _"water well"_ (and tag it with `pump=no`+`man_made=water_well`)

as the actual  difference (both visual and when using) between the two is much greater than say `bubbler` and `water_tap`, so i think it is definitely deserving to differentiate.

Related discussion: https://github.com/gravitystorm/openstreetmap-carto/issues/1224#issuecomment-355811835

matkoniecz:
@mnalis 

> pump=no

Is this really big mechanism visible on the right really not qualifying as a pump? I need to look at it again, I thought that it is one more problematic tagging scheme. But maybe bucket pulled from the well is simply not considered to be a pump?

BTW, if this tagging scheme works well then we can have an extra question asking to distinguish between these two.

Helium314:
> Is this really big mechanism visible on the right really not qualifying as a pump?

Personally I wouldn't call lifting a bucket _pumping_; I usually connect pumping with pressure.
But apparently there _are_ pumps that are just lifting buckets: https://en.wikipedia.org/wiki/Chain_pump

mnalis:
> Is this really big mechanism visible on the right really not qualifying as a pump? I need to look at it again, I thought that it is one more problematic tagging scheme. But maybe bucket pulled from the well is simply not considered to be a pump?

pulling a bucket is not considered pumping... It is actually well defined in the OSM wiki for [man_made=water_well](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dwater_well#Examples) and shown in examples section there.

app/src/main/res/values/strings.xml
```diff
@@ -927,6 +927,15 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_drinking_water_not_potable_signed">Not safe to drink (signed)</string>
     <string name="quest_drinking_water_not_potable_unsigned">Not safe to drink (unsigned)</string>
 
+    <string name="quest_drinking_water_type_title">How drinking water is provided here?</string>
```
riQQ, line 930:
```suggestion
    <string name="quest_drinking_water_type_title">How is drinking water provided here?</string>
```

matkoniecz:
screen04

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/AddDrinkingWaterType.kt
```diff
@@ -0,0 +1,33 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.OUTDOORS
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddDrinkingWaterType : OsmFilterQuestType<DrinkingWaterType>() {
+
+    override val elementFilter = """nodes with
+        (
```
westnordost, line 12:
```suggestion
    override val elementFilter = """
        nodes with
        (
```

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/AddDrinkingWaterType.kt
```diff
@@ -0,0 +1,34 @@
...
+    override val elementFilter = """
+        nodes with
+        (
+            amenity = drinking_water
+            or
+            (disused:amenity = drinking_water and !amenity and older today -1 years)
+        )
+        and (!seasonal or seasonal = no)
+        and !man_made and !natural and !fountain and !pump"""
+
```
westnordost, line 20:
```suggestion
        and !man_made and !natural and !fountain and !pump
    """

```

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt
```diff
@@ -0,0 +1,31 @@
...
+        if (tags["disused:amenity"] == "drinking_water") {
+            tags.remove("disused:amenity")
+            tags["amenity"] = "drinking_water"
+        }
+    } else {
+        if (tags["amenity"] == "drinking_water") {
+            tags.remove("amenity")
+        }
+    }
+}
```
westnordost, line 31: Since the tagging logic is more complicated than key=value (more complicated than e.g. for `building=*`), it is more clear to put all of the logic into the `applyTo` function and let `DrinkingWaterType` be an enum with no further properties. I.e.

```kotlin
fun DrinkingWaterType.applyTo(tags: Tags) {
    when(this) {
        WATER_FOUNTAIN_GENERIC -> tags["fountain"] = "drinking"
        WATER_WELL_WITHOUT_PUMP -> {
            tags["man_made"] = "water_well"
            tags["pump"] = "no"
        }
        //...etc
    }
    if (this != DISUSED_DRINKING_WATER) {
        if (tags["disused:amenity"] == "drinking_water") {
            tags.remove("disused:amenity")
            tags["amenity"] = "drinking_water"
        }
    }
    //...etc
}

**Review by westnordost — COMMENTED** [review]
Looks very good! You injected a reasonable quest into an unreasonable tagging scheme (by only tagging a few values)

File: app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt (hunk 4 above)
Comment (matkoniecz) [reply app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt]: done, testing it right now

this enum's parameters definitely has outgrown its usefulness

File: app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt [hunk 5]
```diff
@@ -0,0 +1,59 @@
...
+        WATER_TAP -> tags["man_made"] == "water_tap"
+        HAND_PUMP -> {
+            tags["man_made"] = "water_well"
+            tags["pump"] = "manual"
+        }
+        WATER_WELL_WITHOUT_PUMP -> {
+            tags["man_made"] = "water_well"
+            tags["pump"] = "no"
+        }
+        SPRING -> tags["natural"] == "spring"
```
FloEdelmann, line 33: Probably a typo:

```suggestion
        WATER_FOUNTAIN_GENERIC -> tags["fountain"] = "drinking"
        WATER_FOUNTAIN_JET -> tags["fountain"] = "bubbler"
        WATER_FOUNTAIN_BOTTLE_REFILL_ONLY -> tags["fountain"] = "bottle_refill"
        WATER_TAP -> tags["man_made"] = "water_tap"
        HAND_PUMP -> {
            tags["man_made"] = "water_well"
            tags["pump"] = "manual"
        }
        WATER_WELL_WITHOUT_PUMP -> {
            tags["man_made"] = "water_well"
            tags["pump"] = "no"
        }
        SPRING -> tags["natural"] = "spring"
```

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt
```diff
@@ -0,0 +1,59 @@
...
+                tags.remove("amenity")
+                changed = true
+            }
+            if (tags["amenity"] != "drinking_water") {
+                tags["disused:amenity"] = "drinking_water"
+                changed = true
+            }
+            if (!changed || tags.hasCheckDate()) {
+                tags.updateCheckDate()
+            }
```
FloEdelmann, line 46: If `tags["amenity"]` is equal to `drinking_water`, set `changed` to true. If it isn't, also set `changed` to true. So the `changed` variable will always be true afterwards and is not needed. BTW, you can use `else` here instead, which makes this more obvious.

FloEdelmann (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt
```diff
@@ -0,0 +1,59 @@
...
+                tags.remove("amenity")
+                changed = true
+            }
+            if (tags["disused:amenity"] != "drinking_water") {
+                tags["disused:amenity"] = "drinking_water"
+                changed = true
+            }
+            if (!changed || tags.hasCheckDate()) {
+                tags.updateCheckDate()
+            }
```
FloEdelmann, line 46: Maybe it can be simplified to this?

```suggestion
            tags.remove("amenity")
            tags["disused:amenity"] = "drinking_water"
            if (!tags.hasChanges() || tags.hasCheckDate()) {
                tags.updateCheckDate()
            }
```

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // handle change from disused:amenity=drinking_water to amenity=drinking_water
+    if (this != DISUSED_DRINKING_WATER) {
+        if (tags["disused:amenity"] == "drinking_water") {
+            tags.remove("disused:amenity")
+        }
+        if (tags["amenity"] != "drinking_water") {
+            tags["amenity"] = "drinking_water"
+        }
```
FloEdelmann, line 57: Then this would probably also work:

```suggestion
        tags.remove("disused:amenity")
        tags["amenity"] = "drinking_water"
```

matkoniecz: yes, with additional checks in the query filter it should work

and if something is tagged with `disused:amenity=drinking_water amenity=drinking_water` then it deserves manual checks... Argh, there are 21 such places (yes, I just opened notes and notified mappers)

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water_type/DrinkingWaterType.kt
```diff
@@ -0,0 +1,48 @@
...
+    SPRING,
+    DISUSED_DRINKING_WATER,
+}
+
+fun DrinkingWaterType.applyTo(tags: Tags) {
+    when (this) {
+        WATER_FOUNTAIN_GENERIC -> tags["fountain"] = "drinking"
+        WATER_FOUNTAIN_JET -> tags["fountain"] = "bubbler"
+        WATER_FOUNTAIN_BOTTLE_REFILL_ONLY -> tags["fountain"] = "bottle_refill"
+        WATER_TAP -> tags["man_made"] = "water_tap"
```
FloEdelmann, line 24: Please reorder to match order above (this got lost in some earlier comment apparently):

```suggestion
        WATER_TAP -> tags["man_made"] = "water_tap"
        WATER_FOUNTAIN_JET -> tags["fountain"] = "bubbler"
        WATER_FOUNTAIN_BOTTLE_REFILL_ONLY -> tags["fountain"] = "bottle_refill"
        WATER_FOUNTAIN_GENERIC -> tags["fountain"] = "drinking"
```

FloEdelmann (approved):
Code and screenshot look fine now, thanks for the changes! I haven't tested this in real life though.

matkoniecz:
> I haven't tested this in real life though.

I did it. (that is why tagging mailing list got some questions about water taps and is likely to get more of that)

I also tested it right now in emulator and I would say that it is ready for merge.

BTW, now two water quests will share icon. But I have no idea about how to differentiate them. Maybe put some label on top of water drop in one asking about drinkability status?

matkoniecz:
screen06

Not really sure about using that for drinkability status... Two identical icons seem fine in this case.

FloEdelmann:
This quest's icon could be similar to https://www.svgrepo.com/svg/150968/fountain

westnordost (approved):
Unit tests for the check date logic (disused / not disused drinking water) would be ideal, but not required for it to be merged.

westnordost:
Thank you for adding the tests!

matkoniecz:
@dbdean How it would be even tagged?

PR #4431, opened 2022-09-30
try to reduce confusion of building vs function

matkoniecz:
current form with 'Church'/'Chapel'/'Synagogue' is causing confusion in cases where religious amenity is in an office/residential building

Not entirely sure should it be using new strings to force translations, not 101% sure is it making sense also in English.

It is intended to help especially with cases like this one: sign about church inside this residential building (one of protestant churches which has very limited amount of people, so they just use a flat not `building=church`)

Some people are really confused in such cases.

screen10
https://www.google.com/maps/place/Ko%C5%9Bci%C3%B3%C5%82+Adwentyst%C3%B3w+Dnia+Si%C3%B3dmego+-+zb%C3%B3r+w+Nowej+Hucie/@50.0796708,20.0315918,3a,90y,121.24h,107.24t/data=!3m6!1e1!3m4!1s-5zoo2QEkuPLTD_tuRm45g!2e0!7i13312!8i6656!4m9!1m2!2m1!1zWmLDs3I!3m5!1s0x4716453fd49e47e5:0x653a97260f2f7778!8m2!3d50.0796171!4d20.0317451!15sCgVaYsOzcpIBHHNldmVudGhfZGF5X2FkdmVudGlzdF9jaHVyY2jgAQA


westnordost (approved):

westnordost:
Yeah, I think this makes sense

PR #4433, opened 2022-09-30
stop asking about surface on private (foot-)paths

legofahrrad:


app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddFootwayPartSurface.kt
```diff
@@ -25,6 +25,7 @@ class AddFootwayPartSurface : OsmFilterQuestType<SurfaceAnswer>() {
             and !note:footway:surface
           )
         )
+        and access !~ private|no
```
matkoniecz, line 28: what about `access=no foot=yes`  taggings? Or `foot=designated`? (some RoW in England have such tagging from wha)

legofahrrad: `and access !~ private|no and (!foot or foot !~ private|no)` would be a better filter then, I guess?

westnordost: Watch your brackets, mister! ;-)

FloEdelmann: I think it should be

```suggestion
        and (access !~ private|no or foot !~ private|no)
```

westnordost:
Code looks good (also for the other). Did you test it (you can create a debug APK from the actions tab)?

legofahrrad:
Sorry, didn't tested it yet (same for #4434)

legofahrrad:
I built a debug apk but it didn't worked as expected. Maybe I did something wrong?
I tried this at this way: https://www.openstreetmap.org/way/300720572
The quests shouldn't show up, because of `access=no` but there were the quests about the surfaces.

mnalis:
@legofahrrad did you run GitHub Action on correct branch?  When you click `Run workflow` it allows you to choose on which branch do you want to run it. It seems you've maybe run it against the default `master` branch, instead against your `patch-2` branch?

legofahrrad:
tried again with `patch-2`, worked now for footpaths but cyclepath surface quest still shows up. The problem: My cyclepath surface PR is on `patch-1`
Tried once again for `patch-1` and worked as expected :)

westnordost:
So, in short you tested it and it works?

legofahrrad:
Yup, works as expected 👍🏻

PR #4434, opened 2022-09-30
stop asking about surface on private (cycle-)paths

legofahrrad:


app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -25,6 +25,7 @@ class AddCyclewayPartSurface : OsmFilterQuestType<SurfaceAnswer>() {
             and !note:cycleway:surface
           )
         )
+        and (access !~ private|no or (foot and foot !~ private|no))
```
matkoniecz, line 28: what about `access=no bicycle=yes`  taggings? (some RoW in England have such tagging from what I remember)

Though maybe excluding bicycle-only ways is fine

legofahrrad:
```suggestion
        and (access !~ private|no or (foot and foot !~ private|no))
        and !bicycle or bicycle !~ private|no
```

matkoniecz: I think you need `or` at proper level, this would block asking on `access=yes foot=yes bicycle=private` and will not enable asking and `access=no bicycle=yes` ways

FloEdelmann: I think it should be

```suggestion
        and (access !~ private|no or foot !~ private|no or bicycle !~ private|no)
```

PR #4435, opened 2022-09-30
Fix overlays not showing properly

Helium314:
onNewScreenPosition currently checks `if (overlay == null)` before setting overlay field to the new value


westnordost:
Cool thank you!

PR #4440, opened 2022-09-30
Add new quest for fire hydrant ref

### LINKED ISSUE #3059: Ask ref for fire_hydrant in France

mcliquid:
Fix #3059

I have added France, Austria and Switzerland currently to the list of allowed countrys as I know that there are ref's for fire hydrants definitly popular. At least in South Germany I've never get used to see one and never learned it during my time as a firefighter in Germany. But I could imagine that it is at least different in each state, since the formation of volunteer fire departments is a state matter.

The Icon contains text. It could be too small and not translateable, but I think "REF" is international?


riQQ:
If the text is deemed to be problematic, another option would be to use the "tag" icon used in the name quest instead.

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_ref/AddFireHydrantRef.kt
```diff
@@ -0,0 +1,38 @@
...
+    override val changesetComment = "Determine fire hydrant refs"
+    override val wikiLink = "Key:ref"
+    override val icon = R.drawable.ic_quest_fire_hydrant_ref
+    override val achievements = listOf(EditTypeAchievement.LIFESAVER)
+    override val isDeleteElementEnabled = true
+    override val enabledInCountries = NoCountriesExcept(
+        "CH", "FR", "AT"
+    )
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_accessPointRef_title
```
FloEdelmann, line 28: Remember to change the quest title

res/graphics/authors.txt
```diff
@@ -322,6 +322,7 @@ quest/
     fire_hydrant.svg
     fire_hydrant_diameter.svg
     fire_hydrant_grass.svg
+    fire_hydrant_ref.svg             modified from fire_hydrant.svg by mcliquid
```
westnordost, line 325: The "REF" letters are too small for the icon. I suggest you either overlay the hydrant picture with the "label" icon (at position of current REF but a little large) or with the "ID" from the mailbox ref quest icon (centered and rather on top)

westnordost:
So it will be displayed in this size. 



Yeah, I think it is okay. Maybe I'd make the label a big bigger still. Also, the blue ribbon is too small to notice (and ends in nothing), I'd remove it. Also, the shadow for the label is wrong, it should go down just like for the hydrant (4px down IIRC).

westnordost:
> @westnordost How can I assign myself to issues and PRs? Or is this an option that is reserved for admins?

No need to assign, in the issue overview this PR is already linked. Noone will get the idea to create another PR for this issue. And regarding being mentioned in the changelog credits, I always go over the commit history manually anyway.

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_ref/AddFireHydrantRef.kt
```diff
@@ -0,0 +1,38 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddFireHydrantRef : OsmFilterQuestType<FireHydrantRefAnswer>() {
+
+    override val elementFilter = """
+        nodes with
+        (
+          emergency=fire_hydrant
+        )
```
westnordost, line 15:
```suggestion
        emergency=fire_hydrant
```

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_ref/AddFireHydrantRefForm.kt
```diff
@@ -0,0 +1,43 @@
...
+import android.view.View
+import androidx.appcompat.app.AlertDialog
+import androidx.core.widget.doAfterTextChanged
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.databinding.QuestRefBinding
+import de.westnordost.streetcomplete.quests.AbstractOsmQuestForm
+import de.westnordost.streetcomplete.quests.AnswerItem
+import de.westnordost.streetcomplete.util.ktx.nonBlankTextOrNull
+
+class AddFireHydrantRefForm : AbstractOsmQuestForm<FireHydrantRefAnswer>() {
```
westnordost, line 13: Hm, as this is 100% exact the same as for postbox refs and emergency access point refs (and more?), I wonder if it is worth to have a generic `AddRefForm`?

app/src/main/res/values/strings.xml
```diff
@@ -968,6 +968,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_fireHydrant_type_wall">Wall</string>
     <string name="quest_fireHydrant_type_pond">Pond</string>
 
+    <string name="quest_fireHydrantRef_title">What’s the identification number here?</string>
```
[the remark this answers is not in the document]
westnordost: I am not sure why he suggested that. IMO, as the question is the same, it's better. Maybe rename the actual string to `quest_genericRef_title` so translators don't get the idea to mention "fire hydrant" and users in certain languages become confused why it's asking about fire hydrant ref for some emergency access points or postboxes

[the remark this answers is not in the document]
FloEdelmann: I didn't recall that the `quest_accessPointRef_title` is actually generic and does not mention fire hydrants at all. But to make this clear (also for translators), renaming this to only one `quest_genericRef_title` string sounds good to me.

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_ref/AddFireHydrantRefForm.kt
FloEdelmann: I think reducing code complexity/duplication is always worth it. Currently, I can't think of any necessary specialization. If we find some in the future, we can still undo this generalization.

Helium314: I agree. It's a simplification, and it's should be simple to revert if it turns out bad later.

westnordost: Yes, that's okay

app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_ref/AddFireHydrantRef.kt
```diff
@@ -0,0 +1,36 @@
...
+        and !name and !ref and noref != yes and ref:signed != no and !~"ref:.*"
+    """
+
+    override val changesetComment = "Determine fire hydrant refs"
+    override val wikiLink = "Key:ref"
+    override val icon = R.drawable.ic_quest_fire_hydrant_ref
+    override val achievements = listOf(EditTypeAchievement.LIFESAVER)
+    override val isDeleteElementEnabled = true
+    override val enabledInCountries = NoCountriesExcept(
+        "CH", "FR"
```
FloEdelmann, line 23: As [`b8586d8`](https://github.com/streetcomplete/StreetComplete/pull/4440/commits/b8586d84d9fc5798c2f81e7271ea8057ba73615c) was "just for testing", should Austria be added back?

```suggestion
        "CH", "FR", "AT"
```

FloEdelmann (approved):

Issue #4079, opened 2022-06-03
Improve I/O performance (especially on download)


westnordost:
## Current Behavior

Currently, there is no cache for the downloaded map data and the generated quests. All map data, its geometry and all quests are directly persisted to the on-disk database and when needed, retrieved from there.

Hence, the implementation is relatively straightforward, but frequent retrievals of data from the database are a bottleneck in the app's performance.
Quests and map data are put into and retrieved from the local database (e.g.)...

- on download, all the downloaded data and the geometry generated from it is persisted into the database. Only **after** this is done, quests are generated, i.e. the scanning for the quests needs to wait until after all that data has been persisted
- for some quests, surrounding related map data is displayed when the form is opened
- on solving quests, the updated element is scanned for eligibility for new quests - some quests require to look at the surrounding map data
- on scrolling the map, quests (and map data) are continuously retrieved in the displayed bounding box

The database does not have a (real) spatial index, so the more data is in the database, the slower a query by bounding box becomes. Additionally, creating and then discarding a lot of objects from database this way is not something the garbage collector is happy about, resulting in additional slowdowns due to garbage collection cycles (especially on a tight memory budget, e.g. Samsung S4 Mini).

## Solution

If `MapDataController` (and `OsmQuestController`) would additionally have an in-memory spatial cache (e.g. the elements are sorted into a raster of tiles), it would avoid querying the database in most cases. Also, the object allocations can be avoided this way. The garbage collector will be happy about this.

## Challenges

- caches need to be properly synchronized aunless one wants to deal with weird Heisenbugs
- one must be careful to always keep the cache and the database in sync
- it should be some sort of cyclic cache, e.g. the tiles longest not accessed are cleaned up first

## Remarks

If you intend to look into this, post here first because @Helium314 vaguely registered interest to busy himself with it before.

Helium314:
I had some thoughts about this:

## Cache
There is [`LruCache`](https://developer.android.com/reference/androidx/collection/LruCache) which could be pretty useful.
* it automatically drops the least recently accessed entry once the size limit is reached.
* by overriding `create` the `LruCache` can load data that is not currently in the cache
* overriding `entryRemoved` could also be useful, e.g. to avoid dropping cache close to the user
* for searching through the entire cache (e.g. looking for `ElementKey` or `QuestKey`) the `snapshot` can be used

## OsmQuestController
[...]

Issue #4077, opened 2022-06-03
Improve performance on upload: Notice if nothing changed


westnordost:
## Current behavior

Whenever the map data changes (download, upload) or the user creates or deletes a local edit, `MapDataWithEditsSource` notifies its listeners about the changed map data. 

Most notably amongst the listeners is the `OsmQuestController` which will re-scan the changed element(s) for quests and in turn triggers the map to refresh displaying the quests. This is done for example when the user answered a quest and it takes relatively long (a few hundred milliseconds up to a second or so).

After an edit has been uploaded, the copy of the original OSM element is updated and the edit is deleted. This triggers `MapDataWithEditsSource` to rebuild the cache (=the local changes applied to the original OSM map data) and in turn triggers notifying the listeners about changed.

## The Issue

If the element the edit has been applied to and uploaded has not been changed in the meantime by someone else, which consistitutes about 99% of the times,

- the previous original OSM element + applied edit
- and the updated OSM element after upload

are identical. In the case that the element didn't change at all, notifying the listeners that the map data changed is superfluous and could theoretically be avoided. Currently it is not avoided, which means not only are the local changes rebuilt on upload of each singular edit (I think this is not avoidable) but that updated element is scanned for quests plus the map being updated as a consequence of scanning for quests. Again, for each singular edit.

At least what could be done is to reduce the number of elements for which the listener is called to those elements that actually changed. This requires to compare the edited elements before `rebuildLocalChanges` has been called with the edited elements after it has been called.

## Challenges

- since edits and OSM elements are managed independently (`MapDataController`, `ElementEditsController`) and thus updating one and deleting the other during upload theoretically constitutes of two separate operations, it is possible that `MapDataWithEditsSource` is actually notified of a change in data twice for the same uploaded element. Needs to be confirmed is this actually the case

## Remarks

If you intend to look into this, post here first because @Helium314 vaguely registered interest to busy himself with it before.

westnordost:
@Helium314, I think you started some implementation in the frame of another ticket already?

I noticed that the solution for this ticket will not only improve performance on upload, but also on the normal usage for most users, as auto-upload after each change is active. So currently after each quest solution/edit for users that have auto-upload enabled, the quest icons and map data (in case of overlays displayed) are updated twice in fast succession, each with usually the same data. Thus, the solution to this ticket will speed up this normal usage of the app by a factor of 2.

Helium314:
[...]

PR #4452, opened 2022-10-02
Cache: split up fetch tilesRect if it makes sense

Helium314:
This is basically what is mentioned in https://github.com/streetcomplete/StreetComplete/pull/4125#issuecomment-1249154809, dealing with diagonal scrolling as mentioned in https://github.com/streetcomplete/StreetComplete/issues/4079#issuecomment-1152616022.

This was _not_ tested in the current state, only the version I had previously (https://github.com/streetcomplete/StreetComplete/compare/master...Helium314:StreetComplete:cache_divide_fetch_rect).
And it still needs some unit tests...

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/download/tiles/TilesRect.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/download/tiles/TilesRect.kt
@@ -3,6 +3,7 @@ package de.westnordost.streetcomplete.data.download.tiles
 import de.westnordost.streetcomplete.data.osm.mapdata.BoundingBox
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.data.osm.mapdata.splitAt180thMeridian
+import de.westnordost.streetcomplete.util.ktx.firstAndLast
 import kotlinx.serialization.Serializable
 import kotlin.math.PI
 import kotlin.math.asinh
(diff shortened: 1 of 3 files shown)


westnordost:
Weird, the first link does not work (doesn't show the comment). Maybe you can just cite what you wrote there?

Helium314:
Well, it's not much...
I quote a part of (then) further plans/ideas
> * Split bbox to fetch from db into up to two smaller bboxes (to fix https://github.com/streetcomplete/StreetComplete/issues/4079#issuecomment-1152616022). I do have some code for this, but I'm inclined to only add it in a separate PR.

and write
> See https://github.com/Helium314/StreetComplete/commit/085fcf8e2a6d1adbf786835ce5f54a6ca535ae57
Seems to work quite well.

westnordost:
But anyway, I get the idea (1= in cache, X+Y not in cache), right?

<table>
<tr><td>1</td><td>1</td><td>1</td><td>X</td></tr>
<tr><td>1</td><td>1</td><td>1</td><td>X</td></tr>
<tr><td>1</td><td>1</td><td>1</td><td>X</td></tr>
<tr><td>Y</td><td>Y</td><td>Y</td><td>X</td></tr>
</table>

Now, is this really relevant? It does not make a difference for displaying the data on the map (because all the data is always replaced on every change). And getting the data from the cache is fast anyway that it shouldn't really matter much.

What's more, splitting it up in two smaller bboxes should not help in querying less tiles, because both segments (X + Y) still span the whole width/height of the tiles-in-view.

Helium314:
> Now, is this really relevant? It does not make a difference for displaying the data on the map (because all the data is always replaced on every change). And getting the data from the cache is fast anyway that it shouldn't really matter much.

I don't really understand. In https://github.com/streetcomplete/StreetComplete/issues/4079#issuecomment-1152616022 you said "I think it would be an improvement in many cases.", and now you question whether it is relevant?
In the few tests I did (before I wrote https://github.com/streetcomplete/StreetComplete/pull/4125#issuecomment-1249154809), I found that fetching from two small bboxs is faster than a large one. The only situation where fetching the full `minTileRect` was faster than `upToTwoMinTileRects` was if the whole area didn't contain any elements.

> And getting the data from the cache is fast anyway that it shouldn't really matter much.

But this PR changes _nothing_ about getting data from the cache. It's only about the cache fetching data from DB.

> What's more, splitting it up in two smaller bboxes should not help in querying less tiles, because both segments (X + Y) still span the whole width/height of the tiles-in-view.

It's definitely querying less. Without this PR, all 16 tiles are fetched from DB (in a single query), while with this PR there are two queries for the _X_ and _Y_ tiles each.
So the only question is whether two small queries are still actually faster than a large one. And in my tests: yes, almost always. Even when the number of queried tiles is reduced from 4 to 3 there is a (marginal) improvement.

westnordost:
> I don't really understand. In https://github.com/streetcomplete/StreetComplete/issues/4079#issuecomment-1152616022 you said "I think it would be an improvement in many cases.", and now you question whether it is relevant?

What I meant in that comment was the general idea to get the minimum bounding box. Which you already implemented. I mentioned the last table in that post as a reservation that this method does not bring an improvement in all situations, e.g. not when scrolling diagonally. This is why I wrote _"But nevertheless, I think it would be an improvement in many cases."_, i.e. in cases where one does not scroll diagonally.

> In the few tests I did (before I wrote https://github.com/streetcomplete/StreetComplete/pull/4125#issuecomment-1249154809), I found that fetching from two small bboxs is faster than a large one. The only situation where fetching the full minTileRect was faster than upToTwoMinTileRects was if the whole area didn't contain any elements.

> So the only question is whether two small queries are still actually faster than a large one. And in my tests: yes, almost always. Even when the number of queried tiles is reduced from 4 to 3 there is a (marginal) improvement.

I see, so that's good then. I'll review the code within the next few days.

Do you have some numbers on by how much this improves performance on fetching the data initially when it is not in cache yet (e.g. scrolling diagonally an area that is not in cache yet), for the changelog?

Helium314:
Some results:
* ca 20% of fetches are divided into 2 bboxes
* for these, 2 small bbox fetches take between 50% and 70% of time that a full bbox fetch takes, which is roughly half a second gain for each split fetch. I usually used highest zoom level on that trip, and I would expect more gain for lower zoom.
* for spatialCache, there was no gain from splitting, maybe because `fetch` is really a single database query, while mapDataCache `fetch` is more complicated.

westnordost:
So 10-14% performance gain but only for the initial scrolling (before it is in cache)

Helium314:
Yes, it's not a huge difference...

westnordost:
Also note that the speed of bbox-db-queries in absence of a spatial index will be very dependent on how much data is in there. Basically, every query iterates through all the data, regardless of how big the bbox query is. So the two time factors are bbox query size because the construction of the Java objects from the cursor (and its geometry etc.) takes time, and the doing the checking on every row on the database table. While the latter stays about constant for constant-sized bboxes, the former will increase with the size of the database.

westnordost:
Yeah, I don't know... while there are other improvements whose impact is measured in a performance increase of "factor X"s (e.g. also #4077 - factor 2, at least for those with auto-sync on), I don't think it is particularly worth it to invest time here (and increase complexity).
Maybe not outright delete this branch, but put it on the back burner?

Helium314:
> at least for those with auto-sync on

For those who have it off, it should actually result in much faster mass-upload now that `rebuildLocalChanges` is faster thanks to caching.

> Maybe not outright delete this branch, but put it on the back burner?

alright

PR #4461, opened 2022-10-04
Remove waterway=fuel from compressed air quest

### LINKED ISSUE #4460: The compressed air quest should not ask at fuel stations for boats
When you are near a gas station for boats which is tagged as amenity=fuel and waterway=fuel it is not relevant to ask for compressed air. Boats do not need to inflate tyres. If they are big enough they already have an air compressor on board.

**How to Reproduce**
Navigate to a boat fuel station which is tagged with amenity=fuel and waterway=fuel like https://www.openstreetmap.org/node/1692354523 and the quest shows up.

**Expected Behavior** 
The quest should not show up for boat gas stations tagged with  amenity=fuel and waterway=fuel.
For the expected behavior the selector should be adjusted somewhere here https://github.com/streetcomplete/StreetComplete/blob/5b8a765adf65829cabe174d7804022d6e4e16df7/app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt#L21

**Versions affected**
v47.0

mcliquid:
Fix #4460

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
@@ -22,6 +22,7 @@ class AddAirCompressor : OsmFilterQuestType<Boolean>() {
            or compressed_air older today -6 years
        )
        and access !~ private|no
+       and waterway != fuel
     """
 
     override val changesetComment = "Survey availability of air compressors"


mnalis:
No, this PR should be rejected. 

See the comments in the issue, it seems to be mistagged node, and SC code is correct to ask the quest there.
There exist many fuel stations that service both ground vehicles **and** boats.

PR #4469, opened 2022-10-06
improve performance when scanning for quests

Helium314:
`OsmFilterQuestType` quests always uses element tags, so there is no need to check whether an element without tags is applicable.
This change results in ~15-30% faster creation of quests in a bbox.


westnordost (changes requested):
see previous comment. Also, better add a comment that explains what and why it is being done. I'd suggest to also rename `elementsWithTags` to `onlyElementsWithTags`

westnordost (approved):
👍 Nice find and implementation by the way! Such a performance gain for just 10 lines of code is pretty neat!

PR #4470, opened 2022-10-06
Add tests for MapDataCache

Helium314:
Adds missing tests in https://github.com/streetcomplete/StreetComplete/pull/4125#issuecomment-1262126349 (except __get by bbox trims data__).

> "is in spatial cache" logic in line 121ff for ways and line 157ff for relations, each. Should probably be covered in several tests due to the complexity

is imo covered in https://github.com/streetcomplete/StreetComplete/pull/4125/commits/b79adee201b85eb0b787318944466fa4feaa0525, and maybe also in some other tests.

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCacheTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCacheTest.kt
@@ -10,6 +10,7 @@ import de.westnordost.streetcomplete.util.math.enclosingBoundingBox
 import org.junit.Assert.*
 import org.junit.Test
 import org.mockito.Mockito.verify
+import org.mockito.Mockito.verifyNoMoreInteractions
 
 internal class MapDataCacheTest {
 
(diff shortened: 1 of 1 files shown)


app/src/test/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCacheTest.kt
```diff
@@ -735,4 +735,85 @@ internal class MapDataCacheTest {
...
+
+        val expectedMapData = MutableMapDataWithGeometry().apply {
+            listOf(node1, node2, node3).forEach { put(it, ElementPointGeometry(it.position)) }
+            listOf(way1, way2, way3, rel1, rel2).forEach { put(it, null) }
+            boundingBox = nodesBBox
+        }
+        assertEquals(expectedMapData, cache.getMapDataWithGeometry(nodesBBox))
+        verify(elementDB).getAll(nodesRect.asBoundingBox(16))
+        // second time it's cached
+        assertEquals(expectedMapData, cache.getMapDataWithGeometry(nodesBBox))
```
westnordost, line 764: Wouldn't you have to test for no further interactions with elementDB here? (Or is this implied in mockito - not sure)

Helium314: Not sure how I can test that...
What I did is line 753:
`on(elementDB.getAll(nodesRect.asBoundingBox(16))).thenReturn(elementsInsideBBox).thenThrow(IllegalStateException())`
so on second request there is an exception. You can check this be adding `cache.clear()` before the last `assertEquals`.

westnordost: Ah I see, that should work, I didn't notice that.

Maybe it would be better readable to use `verifyNoMoreInteractions``?

Helium314: I didn't know about that...

PR #4471, opened 2022-10-07
Migrate to kotlinx-datetime

### LINKED ISSUE #4240: Migrate from java.time to kotlinx-datetime
A further step to make the core part of StreetComplete platform independent (#1892), usages of `java.time` classes should be replaced with kotlinx-datetime.
I have had a cursory look at v0.4.0 so far and it looks stable enough now, though some convenience functions are missing.

**Steps to do**
- add [kotlinx-datetime](https://github.com/Kotlin/kotlinx-datetime) as dependency
- Ctrl+F for usages of `java.time` in the project
- replace usages with equivalent classes from kotlinx-datetime
- (you'll notice, a lot of convenience functions are missing, so add those as extension functions to `util/ktx/*.kt` instead of littering the code with unwieldy constructs)

**Note**
There is no 100% guarantee that every use case java.time is used in StreetComplete can be covered with kotlinx-datetime 0.4.0, after all it is still marked as "experimental" and I have had only a cursory look. But it looks quite complete by now.

YoshiRulz:
resolves #4240

On the whole this was rather easy as the kotlinx-datetime API seems to be based on Java's.


app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
```diff
@@ -1,12 +1,31 @@
...
+
+typealias LocalDateTime = kotlinx.datetime.LocalDateTime
+
+typealias LocalTime = kotlinx.datetime.LocalTime
+
+typealias Month = kotlinx.datetime.Month
+
+typealias ZoneId = kotlinx.datetime.TimeZone
+
+typealias ZoneOffset = kotlinx.datetime.FixedOffsetTimeZone
```
westnordost, line 25: What are these typealiases for? Why not simply import them?

If some of this is to keep the same names for classes in other files - this is not necessary. Transitioning to kotlinx-datetime is intended to be a one-way-road, we do not intend to go back to java.time. Hence, the API does not need to be the same, just not more complicated (an less complicated is fine too!)

app/src/test/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsParserTest.kt
```diff
@@ -24,7 +25,7 @@ class StatisticsParserTest {
             ),
             2345,
             78,
-            OffsetDateTime.parse("2007-12-03T10:15:30+01:00").toInstant().toEpochMilli(),
+            Instant.parse("2007-12-03T10:15:30+01:00").toEpochMilli(), //TODO doesn't support +HH:MM suffix? https://github.com/Kotlin/kotlinx-datetime#converting-instant-and-local-datetime-to-and-from-string
```
westnordost, line 28: The `lastUpdate` field in the to-be-parsed JSON is created here

https://github.com/streetcomplete/sc-statistics-service/blob/45d2d73a81e0941817ae343fe093f9a94802cdb9/get_statistics.php#L60

i.e. `date('c', $last_update)`.
In the [documentation](https://www.php.net/manual/en/datetime.format.php), this is described as:

<html><body>


c | ISO 8601 date | 2004-02-12T15:19:21+00:00
-- | -- | --



</body>
</html>

Your link mentions that only ISO-8601 is supported. So, I don't know... is one of the documentations incorrect?

app/src/main/java/de/westnordost/streetcomplete/osm/ResurveyUtils.kt
```diff
@@ -20,8 +20,8 @@ val LAST_CHECK_DATE_KEYS = listOf(
     "survey_date"
 )
 
-fun LocalDate.toCheckDateString(): String =
-    DateTimeFormatter.ISO_LOCAL_DATE.format(this)
+@Suppress("NOTHING_TO_INLINE")
+inline fun LocalDate.toCheckDateString(): String = this.toString()
```
westnordost, line 24: Please add a unit test that confirms that `toString` on the kotlinx.datetime.LocalDate actually outputs the date in the desired format

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -12,20 +12,24 @@ import de.westnordost.osmapi.traces.GpsTrackpoint
...
         // https://stackoverflow.com/a/49862573/7718197
-        val name = DateTimeFormatter
-            .ofPattern("yyyy_MM_dd'T'HH_mm_ss.SSSSSS'Z'")
-            .withZone(ZoneOffset.UTC)
-            .format(Instant.ofEpochMilli(trackpoints[0].time)) + ".gpx"
+        val name = Instant.ofEpochMilli(trackpoints[0].time).atZone(ZoneOffset.UTC).let {
+            // doesn't seem to have a formatter function; could have done `.toJavaInstant()` and passed that to DateTimeFormatter but it would have been the 1 remaining instance of `java.time` --yoshi
+            ("${it.year.toPaddedString(4)}_${it.monthNumber.toPaddedString(2)}_${it.dayOfMonth.toPaddedString(2)}"
+                + "T${it.hour.toPaddedString(2)}_${it.minute.toPaddedString(2)}_${it.second.toPaddedString(2)}.${it.nanosecond.toPaddedString(6)}Z.gpx")
+        }
```
westnordost, line 32: Maybe create a private extension function to convert an Instant to a string in that format? I find this a bit hard to read.

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
```diff
@@ -17,3 +36,59 @@ fun isApril1st(): Boolean {
...
+
+/** https://github.com/Kotlin/kotlinx-datetime#date--time-arithmetic */
+fun LocalDateTime.minusHours(hours: Long): LocalDateTime = ZoneId.systemDefault().let { tz ->
+    this.toInstant(tz).minus(hours, DateTimeUnit.HOUR, tz).toLocalDateTime(tz)
+}
+
+/** https://github.com/Kotlin/kotlinx-datetime#date--time-arithmetic */
+fun LocalDateTime.plusHours(hours: Long): LocalDateTime = ZoneId.systemDefault().let { tz ->
+    this.toInstant(tz).plus(hours, DateTimeUnit.HOUR, tz).toLocalDateTime(tz)
+}
```
westnordost, line 76: if `.plus(Long, DateTimeUnit)` is the kotlinx-datetime api to add/subtract hours/days/minutes/... from a datetime, it would probably be more consistent to have extension functions with the same signature on `LocalDateTime` too instead of trying to mirror the java.time API. It is not really more verbose than the java.time counterpart.

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
```diff
@@ -17,3 +36,59 @@ fun isApril1st(): Boolean {
...
+    instant.atZone(zone)
+
+inline fun LocalTime_Companion.ofSecondOfDay(secondOfDay: Long): LocalTime =
+    LocalTime.fromSecondOfDay(secondOfDay.toInt())
+
+inline fun TimeZone_Companion.systemDefault(): ZoneId = ZoneId.currentSystemDefault()
+
+inline fun Instant.atZone(timeZone: ZoneId): LocalDateTime = this.toLocalDateTime(timeZone)
+
+inline fun Instant.toEpochMilli(): Long = this.toEpochMilliseconds()
```
westnordost, line 64: As I wrote in other comment, here and for other functions, it is OK IMO to use the kotlinx-datetime API and not try to mirror the java.time API as long as it is not more complicated. Some additional letters don't matter :-P

westnordost (commented):
Great job so far! I was puzzled first why only the imports changed, but saw that you solved this entirely through extension functions, mirroring the java.time API.

1. 
`Instant.now().toEpochMilli()` should be the same as `System.currentTimeMillis()`, right? The latter does not even need an explicit import because it is in java.lang but at the same time it is not cross-platform for Kotlin. So any `currentTimeMillis()` need replacement too.

Maybe it is worth just defining a public function called `currentEpochMillis()` or something?

2. Did you notice any (other) shortcuts that could be taken? E.g. any usages of the whole date time stuff that is always the same and could just be shortened to use one extension or static public function instead?

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/create/CreateNodeAction.kt
```diff
@@ -27,7 +29,7 @@ data class CreateNodeAction(
         mapDataRepository: MapDataRepository,
         idProvider: ElementIdProvider
     ): MapDataChanges {
-        val node = Node(idProvider.nextNodeId(), position, tags, 1, Instant.now().toEpochMilli())
+        val node = Node(idProvider.nextNodeId(), position, tags, 1, Clock.System.now().toEpochMilli())
```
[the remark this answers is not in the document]
YoshiRulz: `Instant.now()` was deprecated with the latter in `ReplaceWith`

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
YoshiRulz: I only wanted to keep the diff small. I thought about inlining them in a third commit but decided to wait in case a lot of changes were requested.

app/src/test/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsParserTest.kt
YoshiRulz: False alarm, it uses the same format (at least on JVM) https://github.com/streetcomplete/StreetComplete/pull/4471/commits/bea813ad99b66c6062db4d9cf103ef0e0f87e5a3

app/src/main/java/de/westnordost/streetcomplete/osm/ResurveyUtils.kt
YoshiRulz: https://github.com/streetcomplete/StreetComplete/pull/4471/commits/c31617ce3ae573285c9f06ea5c2c9758aa17a5c6

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
YoshiRulz: https://github.com/streetcomplete/StreetComplete/pull/4471/commits/077ce5add873e07888735c9ccf692e941967db79

YoshiRulz:
Anything else before I start on your cleaner API suggestions?

app/src/main/java/de/westnordost/streetcomplete/osm/ResurveyUtils.kt
YoshiRulz: Turns out there was already a test for this:
https://github.com/streetcomplete/StreetComplete/blob/0e01b1a92e4390806a100c2d28e24cd9a5dcc12e/app/src/test/java/de/westnordost/streetcomplete/data/meta/ResurveyUtilsTest.kt#L22-L24

westnordost:
Maybe have the extension function for the track name private in the `TracksApiImpl` but otherwise, go ahead!

westnordost:
reading vertically, through the changes, I think it is not a good idea to use `toString` in code because it is not explicit what format exactly is expected and hence practically undocumented except for some single unit test somewhere. When there is an explicit `toCheckDateString`, it can have a documentation comment that describes exactly how the output will look.

Generelly, I never trust `toString` to have a certain output or that that certain output may not change at any time. As far as I know, `toString` is meant to be a string representation for e.g. logs and debugging in Java. (That's why there is `MyEnumConstant.name` next to `MyEnumConstant.toString`, I figure)

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
YoshiRulz: https://github.com/streetcomplete/StreetComplete/pull/4471/commits/3a21417f2144b30f3943bde58def2ae5faee72c6

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
YoshiRulz: https://github.com/streetcomplete/StreetComplete/pull/4471/commits/05906c64861f6b10b777604588ebc6bf99c1d478

YoshiRulz:
Rebased and force pushed ([backup](https://github.com/YoshiRulz/StreetComplete/tree/kotlify-datetime-bak)). @westnordost I'm done with changes

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
```diff
@@ -60,3 +59,9 @@ private inline fun <T> wrapExceptions(block: () -> T): T =
         // request timeout is a temporary connection error
         throw if (e.errorCode == 408) ConnectionException(e.message, e) else e
     }
+
+private fun LocalDateTime.toTrackFilename(): String {
+    fun Int.f(len: Int): String = this.toString().padStart(len, '0')
+    return ("${this.year.f(4)}_${this.monthNumber.f(2)}_${this.dayOfMonth.f(2)}"
+        + "T${this.hour.f(2)}_${this.minute.f(2)}_${this.second.f(2)}.${this.nanosecond.f(6)}Z.gpx")
```
westnordost, line 66: No need for `this`
```suggestion
    fun Int.f(len: Int): String = toString().padStart(len, '0')
    return ("${year.f(4)}_${monthNumber.f(2)}_${dayOfMonth.f(2)}"
        + "T${hour.f(2)}_${minute.f(2)}_${second.f(2)}.${nanosecond.f(6)}Z.gpx")
```

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
```diff
@@ -1,19 +1,48 @@
...
+fun LocalDateTime.minusInSysTZ(value: Long, unit: DateTimeUnit): LocalDateTime =
+    TimeZone.currentSystemDefault().let { tz ->
+        this.toInstant(tz).minus(value, unit, tz).toLocalDateTime(tz)
+    }
+
+/** https://github.com/Kotlin/kotlinx-datetime#date--time-arithmetic */
+fun LocalDateTime.plusInSysTZ(value: Long, unit: DateTimeUnit): LocalDateTime =
+    TimeZone.currentSystemDefault().let { tz ->
+        this.toInstant(tz).plus(value, unit, tz).toLocalDateTime(tz)
+    }
```
westnordost, line 48:
```suggestion
fun LocalDateTime.minusInSysTZ(value: Long, unit: DateTimeUnit): LocalDateTime {
    val tz = TimeZone.currentSystemDefault()
    return toInstant(tz).minus(value, unit, tz).toLocalDateTime(tz)
}

/** https://github.com/Kotlin/kotlinx-datetime#date--time-arithmetic */
fun LocalDateTime.plusInSysTZ(value: Long, unit: DateTimeUnit): LocalDateTime {
    val tz = TimeZone.currentSystemDefault()
    return toInstant(tz).plus(value, unit, tz).toLocalDateTime(tz)
}
```

I find this easier to read. Also, I'd suggest to either drop the `InSysTZ` in the function name (the doc comment should be enough of a warning) or write it out to `InSystemTimeZone`.

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
```diff
@@ -1,19 +1,48 @@
...
-fun LocalDate.toEpochMilli(): Long = this.toInstant().toEpochMilli()
+fun LocalDate.toEpochMilli(): Long = this.toInstant().toEpochMilliseconds()
 
 fun Instant.toLocalDate(): LocalDate =
-    this.atZone(ZoneId.systemDefault()).toLocalDate()
+    this.toLocalDateTime(TimeZone.currentSystemDefault()).date
 
 fun isApril1st(): Boolean {
-    val now = LocalDate.now()
+    val now = systemTimeNow().toLocalDate()
```
westnordost, line 27: Worth making this into an extension function like you did for `LocalDateTime.Companion.now()` some lines below? I have seen a few such usages IIRC during review.

westnordost (commented):
Review complete! The shortcuts look really good!

Two general notes:

---

I somewhat dislike Kotlin's `measureTimeMillis` (and `measureTimedValue`) (in production code) because it makes it look like the measuring of the time is the main event.

To me, it looks a bit like having a method like `getSomeData` actually not only return some data, but also do other things, i.e. have side-effects. (Such functions are a commonly known anti-pattern.)

Instead, the measuring of the time is the side effect, not the work done in-between the measuring. This is why I have abstained from using `measureTimeMillis` so far, except maybe in some test code where the main thing actually **is** measuring the time of operations not the operations themselves.

Further (minor) reasons why I dislike `measureTimeMillis`:
- it adds an (unnecessary) level of indentation to the code. I do find code with less indentation easier to read.
- it adds a scope. I.e. variables declared within that scope are not accessible to the outside. IMO there is no good reason for that
- such Kotlin inline/block functions are always a bit opaque regarding what is `this` now (and whether there is an `it`. One would need to look into the documentation for each of these. This is a problem for all these functions generally of course, but this is why I try to reduce usage of such functions or delegate to declared functions

---

Regarding replacement of `max(0, value)` with `value.coerceAtLeast(0L)` (etc.):

Why?

I always found the `coerce...` functions less expressive but maybe just because I am used to `min` and `max` from other programming languages. The name is also slightly confusing, as there exists a concept of "type coercion" in other languages which however is completely unrelated to these functions.

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
YoshiRulz: https://github.com/streetcomplete/StreetComplete/pull/4471/commits/f77366a26d0e4c2e7a55efb94fe6848201e08455

app/src/main/java/de/westnordost/streetcomplete/data/osmtracks/TracksApiImpl.kt
YoshiRulz: https://github.com/streetcomplete/StreetComplete/pull/4471/commits/0466eace8700043627df4c27c1c5a0f76bc4b84e

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
YoshiRulz: `systemTimeNow()` is called 8 times, 6 of those are `systemTimeNow().toLocalDate()`.

westnordost:
> I forgot to mention that I also replaced a few System.currentTimeMillis calls in androidTest when rebasing. I presume that needs to run on Android, how do I do that?

Do you have Android Studio installed? You can run it in an emulator. Add/start/manage emulators via this button

image

Otherwise, you can also connect your Android phone and run the tests there. But this requires a few additional steps (turn on developer mode on the phone, turn on USB debugging on the phone, maybe installing some drivers that allow your computer to see the phone, accepting your computer as trusted on your phone)

app/src/main/java/de/westnordost/streetcomplete/util/ktx/LocalDate.kt
westnordost: Ok, was just a suggestion.

westnordost (approved):
Looks really nice!

YoshiRulz:
My phone is set up to launch and debug, but I'm running the tests with `:app:testDebugUnitTest` and that doesn't include a deploy at all.

Am I good to do a final rebase now? It looks like there's a new conflict to resolve as well.

westnordost:
> My phone is set up to launch and debug, but I'm running the tests with :app:testDebugUnitTest and that doesn't include a deploy at all.

Hmm, not sure. I usually always run the test in the UI, i.e.

image

But I'd have thought that it does exactly the same. Don't worry about it, I will test it after the merge.

I resolved the conflict.

FloEdelmann:

PR #4472, opened 2022-10-07
Don't call MapDataWithEditsSource listeners if nothing changed

### LINKED ISSUE #4077: Improve performance on upload: Notice if nothing changed

Helium314:
fixes #4077 

This is a longer version of what was linked in https://github.com/streetcomplete/StreetComplete/issues/4077#issuecomment-1265693266, but it's not doing more work and I think it's easier to read that way.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -22,6 +22,7 @@ import de.westnordost.streetcomplete.data.osm.mapdata.Node
 import de.westnordost.streetcomplete.data.osm.mapdata.Relation
 import de.westnordost.streetcomplete.data.osm.mapdata.Way
 import de.westnordost.streetcomplete.data.upload.ConflictException
+import de.westnordost.streetcomplete.util.ktx.copy
 import de.westnordost.streetcomplete.util.math.intersect
 import java.util.concurrent.CopyOnWriteArrayList
 
@@ -62,7 +63,32 @@ class MapDataWithEditsSource internal constructor(
             val modifiedElements = ArrayList<Pair<Element, ElementGeometry?>>()
             val modifiedDeleted = ArrayList<ElementKey>()
             synchronized(this) {
+                /* We don't want to callOnUpdated if none of the changes affects map data provided
+                 * by MapDataWithEditsSource.
+                 * This is the case if
+                 *  * All keys in deleted are already in deletedElements.
+                 *  * The modified versions of all elements in updated are the same before and after
+                 *    rebuildLocalChanges, except for the timestamp (expected to have few ms
+                 *    difference) and version (never updated locally).
+                 */
+                val deletedUnchanged = deletedElements.containsAll(deleted)
+                val elementsThatMightHaveChanged = updated.mapNotNull { element ->
+                    val key = ElementKey(element.type, element.id)
+                    if (element.isEqualExceptVersionAndTimestamp(updatedElements[key]))
+                        null // we already have the updated version, so this element is unchanged
+                    else
+                        key to get(element.type, element.id) // elementKey and element as provided by MapDataWithEditsSource
+                }
+
                 rebuildLocalChanges()
+
+                val nothingChanged = deletedUnchanged && elementsThatMightHaveChanged.all {
+                        val updatedElement = get(it.first.type, it.first.id)
+                        // old and new elements are equal except version and timestamp, or both are null
+                        it.second?.isEqualExceptVersionAndTimestamp(updatedElement) ?: (updatedElement == null)
+                    }
+                if (nothingChanged)
+                    return
 
                 for (element in updated) {
                     val key = ElementKey(element.type, element.id)
@@ -438,3 +464,6 @@ class MapDataWithEditsSource internal constructor(
         listeners.forEach { it.onCleared() }
     }
 }
+
+private fun Element.isEqualExceptVersionAndTimestamp(element: Element?): Boolean =
+    element?.copy(timestampEdited = timestampEdited, version = version) == this


app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -438,3 +464,6 @@ class MapDataWithEditsSource internal constructor(
         listeners.forEach { it.onCleared() }
     }
 }
+
+private fun Element.isEqualExceptVersionAndTimestamp(element: Element?): Boolean =
+    element?.copy(timestampEdited = timestampEdited, version = version) == this
```
westnordost, line 469: it's unexpected (and certainly slower) to copy an element for a comparison. I'd very much prefer something like
```kotlin
/** returns whether this element equals given [element], ignoring any versioning info */
private fun Element.equalsInData(element: Element): Boolean = 
  id == element.id && tags == element.tags && when {
  is Node -> position == element.position // (or .equalsInOsm??)
  is Way -> ...etc
  is Relation ->...etc
}

```

 even if it is a few more lines of code

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -62,8 +63,33 @@ class MapDataWithEditsSource internal constructor(
...
             synchronized(this) {
+                /* We don't want to callOnUpdated if none of the changes affects map data provided
+                 * by MapDataWithEditsSource.
+                 * This is the case if
+                 *  * All keys in deleted are already in deletedElements.
+                 *  * The modified versions of all elements in updated are the same before and after
+                 *    rebuildLocalChanges, except for the timestamp (expected to have few ms
+                 *    difference) and version (never updated locally).
+                 */
+                val deletedUnchanged = deletedElements.containsAll(deleted)
```
westnordost, line 74: rename to deleted(Elements)**Is**Unchanged? (And `elementsThatMightHaveChanged` to something mentioning `byKey`?

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -62,8 +63,33 @@ class MapDataWithEditsSource internal constructor(
...
+                val elementsThatMightHaveChanged = updated.mapNotNull { element ->
+                    val key = ElementKey(element.type, element.id)
+                    if (element.isEqualExceptVersionAndTimestamp(updatedElements[key]))
+                        null // we already have the updated version, so this element is unchanged
+                    else
+                        key to get(element.type, element.id) // elementKey and element as provided by MapDataWithEditsSource
+                }
+
```
westnordost, line 82: What if `deletedUnchanged` is true and `elementsThatMightHaveChanged` is empty at this point? Why does `rebuildLocalChanges` need to be called then?

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -62,8 +63,33 @@ class MapDataWithEditsSource internal constructor(
...
                 rebuildLocalChanges()
 
+                val nothingChanged = deletedUnchanged && elementsThatMightHaveChanged.all {
+                        val updatedElement = get(it.first.type, it.first.id)
+                        // old and new elements are equal except version and timestamp, or both are null
+                        it.second?.isEqualExceptVersionAndTimestamp(updatedElement) ?: (updatedElement == null)
+                    }
+                if (nothingChanged)
```
westnordost, line 90: In what situation could `elementsThatMightHaveChanged` **not** be empty but `nothingChanged` be true

westnordost (changes requested):
Some questions... also, should add tests as this is a critical section of code

Also:

> but it's not doing more work and I think it's easier to read that way.

Thank you, yes!

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
Helium314: Does it?
I don't know, and I didn't bother because with cache it's a lot faster than it used to be.
But if it's not necessary maybe we can even return before.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
Helium314: When there are two edits on the same element, and `onUpdated` is called after the first edit is uploaded.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -62,8 +62,33 @@ class MapDataWithEditsSource internal constructor(
...
+                    val key = ElementKey(element.type, element.id)
+                    if (element.isEqualExceptVersionAndTimestamp(updatedElements[key]))
+                        null // we already have the updated version, so this element is unchanged
+                    else
+                        key to get(element.type, element.id) // elementKey and element as provided by MapDataWithEditsSource
+                }
+
                 rebuildLocalChanges()
 
+                val nothingChanged = deletedIsUnchanged && elementsThatMightHaveChangedByKey.all {
```
westnordost, line 84:
```suggestion
                /* nothingChanged can be false at this point when e.g. there are two edits on the 
                   same element, and onUpdated is called after the first edit is uploaded. */
                val nothingChanged = deletedIsUnchanged && elementsThatMightHaveChangedByKey.all {
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
westnordost: okay then let's leave it as it is. Not updating `version` and `timestamp` of the elements may have some unwanted side-effects on upload/conflict solving and quest creation, respectively. Updating the select elements on the other hand is more implementation and testing effort.

westnordost:
What should be tested:
- does not call onUpdated when all deleted elements are already deleted
- does not call onUpdated when all updated elements stayed the same
- does not call onUpdated when all updated elements stayed the same except for version and timestamp
- does call onUpdated when not all deleted elements are already deleted
- does call onUpdated when not all updated elements stayed the same

westnordost:
In terms of numbers, I guess it is reasonable to claim 2x faster quest solving (with auto-sync on) and 2x faster upload times?

Helium314:
I added the tests, but the last one fails.
I think this is because using `mapDataChangesAre` the changes are not actually edits applied on top of some initial elements... how do I do that?

westnordost:
Uhm, not sure I understand the problem. But `mapDataChangesAre` is just a helper function. You can create a different helper function as you like you fix it if it does not do what one would expect.

westnordost:
Maybe I ran into the same problem as you when adding tests to the now dismissed #4481 PR. Shouldn't you be able to use `editsControllerNotifiesMapDataChangesAdded`?

Helium314:
Thanks! `editsControllerNotifiesMapDataChangesAdded` and `originalElementsAre` made it work as intended.

PR #4481, opened 2022-10-10
Add support for edits working on several elements

westnordost:
Currently, `ElementEdit`s refer to exactly one element.  Each `ElementEdit` hence contains the original element the edit was made on and its element type and id.

This is necessary:

1. **most importantly:** to update the id of the element the edit refers to after upload of created elements (new elements get temporary negative ids assigned and get their proper id assigned after upload by the OSM API).
2. downloading the current version of an element for which the upload of an edit resulted in a conflict
3. clearing all edits made on an object one just created with another edit when that first edit is deleted via undo-function
4. show the correct quest title looking at the tags of the element edited by an edit in the undo dialog

This reaches a limitation when an edit touches several elements or when an edit creates an element (i.e. has no original element to refer to). Creating (free-floating) nodes already only worked through a hack (using a "dummy" element). Creating a vertex of a way (=create node and modify way) would not work for ways that were created in this app because with the above, it wouldn't be possible to update the id of the way that edit refers to.

### The goal

We want an action that can (without hacks) do exactly that:

Create a new node with tags at a given position and make it a node of the given way(s). (ways in plural because it is practically no extra complexity to support adding it to several ways even if this is not used (yet)).

### This solution

1. The `ElementEdit` does not refer to one element anymore. Instead: The `ElementEditAction` (each `ElementEdit` has one action) contains all the information it needs to perform (e.g. the original element the edit was made on)
2. The element edits are no longer indexed by element in the database. In fact, the elements of an element edit action are never updated. Instead: `CreatedElementsController` keeps track of the information which temporary/local ids refer to which ids assigned by the OSM API and is able to return that assigned id given a temporary/local id. When an `ElementEditAction` gets passed a `MapDataRepository`, it is wrapped in a `MapDataRepositoryWithUpdatedIds` which automatically resolves old ids to new ids using `CreatedElementsSource`.

Point 2,3,4 are not solved with this (so far).

### TODO

- upgrade database
- 2 remaining failing tests (should only continue after rebasing on top of #4472)
- solve 2,3,4

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/created_elements/CreatedElementsDaoTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/created_elements/CreatedElementsDaoTest.kt
@@ -15,13 +15,13 @@ class CreatedElementsDaoTest : ApplicationDbTestCase() {
         dao = CreatedElementsDao(database)
     }
 
-    @Test fun addGetDelete() {
+    @Test fun putGetDelete() {
         assertTrue(dao.getAll().isEmpty())
 
         val elements = listOf(
-            ElementKey(ElementType.NODE, 1),
-            ElementKey(ElementType.WAY, 1),
-            ElementKey(ElementType.NODE, 3),
+            CreatedElementKey(ElementType.NODE, 1, null),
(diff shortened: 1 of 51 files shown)


westnordost:
scratch that, it will not work to solve 2,3,4 with this approach.

PR #4484, opened 2022-10-11
Change order of footway quests

Henry00572:
Since `AddPathSurface` relies on the key `segregated,` the corresponding quest `AddCyclewaySegregation` should be asked before `AddPathSurface` by default.


westnordost:
`AddCyclewaySegregation` is only asked for paved surfaces. Hence, `AddPathSurface` is asked before anyway.

Henry00572:
I see, I didn't consider this. The thing is, we currently get ways with `surface,` `cycleway:surface` and `footway:surface` (https://www.openstreetmap.org/way/364908727/history), but I guess this can't be avoided in any other way.

PR #4486, opened 2022-10-11
remove "surface:grade" when updating "smoothness"

mnalis:
[surface:grade=*](https://wiki.openstreetmap.org/wiki/Key:surface:grade) is/was an less popular (and poorly defined) alternative to [smoothness=*](https://wiki.openstreetmap.org/wiki/Key:smoothness). It is still somewhat popular, with about `25k` uses.

So, when updating `smoothness` in any way, we should remove `surface:grade` (which is sometimes used in parallel) as otherwise we'd be leaving conflicting information.

I've also added a [test](https://github.com/mnalis/StreetComplete/actions/runs/3230052001) to check for correct removal.

App [compiles and seems to work fine](https://github.com/mnalis/StreetComplete/actions/runs/3229730590),  e.g.:
small_Screenshot_20221011_222352_de westnordost streetcomplete mn debug

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddPathSmoothness.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/smoothness/AddPathSmoothness.kt
@@ -39,16 +39,19 @@ class AddPathSmoothness : OsmFilterQuestType<SmoothnessAnswer>() {
         when (answer) {
             is SmoothnessValueAnswer -> {
                 tags.updateWithCheckDate("smoothness", answer.value.osmValue)
+                tags.remove("surface:grade")
                 tags.remove("smoothness:date")
             }
             is WrongSurfaceAnswer -> {
                 tags.remove("surface")
+                tags.remove("surface:grade")
                 tags.remove("smoothness")
                 tags.remove("smoothness:date")
                 tags.removeCheckDatesForKey("smoothness")
             }
             is IsActuallyStepsAnswer -> {
                 tags["highway"] = "steps"
+                tags.remove("surface:grade")
                 tags.remove("smoothness")
                 tags.remove("smoothness:date")
                 tags.removeCheckDatesForKey("smoothness")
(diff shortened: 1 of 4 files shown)


westnordost:
Did you run the tests?

mnalis:
> Did you run the tests?

Yes, see the link in:

> I've also added a [test](https://github.com/mnalis/StreetComplete/actions/runs/3230052001) to check for correct removal.

And that `Test with Gradle` workflow action [didn't fail](https://github.com/mnalis/StreetComplete/actions/runs/3230052001) (and it _does_ fail if I [break assumptions](https://github.com/mnalis/StreetComplete/actions/runs/3230020085/jobs/5287982396#step:5:368) in my fork in unrelated quests), so I assume the tests are passing.

westnordost:
Hm, why does `on wrong surface, deletes everything smoothness-related` not fail?

Anyway, please add the identical tests to `AddPathSmoothnessTest` and tests for `SurfaceAnswer`.

Oh.. there is no test for `SurfaceAnswer`... but the tests are scattered in `AddRoadSurfaceTest` etc... hmpf.

mnalis:
> Hm, why does `on wrong surface, deletes everything smoothness-related` not fail?

Because it didn't add `smoothness:grade`, so it wasn't surprised when it wasn't removed.
I've extended that test now to both add `smoothness:grade` and check it is removed. (030338d98e51090c04d26254e9f9739987f9ddbf)

> Anyway, please add the identical tests to `AddPathSmoothnessTest` and tests for `SurfaceAnswer`.

Ummmmm... gladly, but where **is** `AddPathSmoothnessTest`?

> Oh.. there is no test for `SurfaceAnswer`... but the tests are scattered in `AddRoadSurfaceTest` etc... hmpf.

Huh, I've found this one: https://github.com/streetcomplete/StreetComplete/blob/fc9e478f0818eb4e340097ff2267ef3048ae607d/app/src/test/java/de/westnordost/streetcomplete/quests/surface/AddRoadSurfaceTest.kt#L11

but it does not seem to have anything `smoothness` related, so I guess it should not have anything `surface:grade` related either?

westnordost:
Okay, thank you. I will see if I can clean this up further after the merge

Issue #4483, opened 2022-10-10
Handle addr:substreet and addr:parentstreet as used in the UK in house street name quest


peternewman:
As detailed here, the UK now sometimes uses some alternative tags:
https://wiki.openstreetmap.org/wiki/Addresses_in_the_United_Kingdom#addr:substreet,_addr:street_&_addr:parentstreet

I should note I'm just a user here who only happened to notice due to a discrepancy around street names and not being able to select this street ( https://www.openstreetmap.org/way/37107064 ) as the name, which it turns out was because it didn't have a name set and that wasn't asked because it's only a service road.

**How to Reproduce**
Go to https://www.openstreetmap.org/way/230252827
Note how the street name quest is being asked for the building, despite substreet and parentstreet being set.

**Expected Behavior**
Don't ask if either addr:substreet or addr:parentstreet is already set.

**Versions affected**
SC v47.1

peternewman:
The address overlay will need tweaking too...

westnordost:
Uhm, i don't understand these tags

westnordost:
So what to do for the address overlay then?

peternewman:
I'm not sure, partly because I've not seen it yet because the release cadence combined with what feels like a bug in Google Play means its not offering me the higher numbered alpha release to update to.

What does it do about house names (rather than, or as well as numbers), or are they shown on the three dot menu?

I guess for editing existing ones, it should show two street fields if they're set (N.B. iD doesn't either currently), so as not to damage or misrepresent existing data, or it could just show the substreet, and change the label.

For adding new addresses, I guess either solution could work too, either a way to switch on the three dots (like house name hopefully has), or something else. If I'm honest, if I hadn't gone down this particular rabbit hole because the service road didn't have a name set, I wouldn't have known, so I'm not convinced your average punter will necessarily know if it's a sub street/whether you can tell onsite. So adding it as a street and then someone with more knowledge correcting that later doesn't seem terrible to me.

I think the main thing we want to avoid is us ending up with:
```
addr:housenumber: 29
addr:substreet: Acacia Avenue
addr:parentstreet: High Road
```

Becoming
```
addr:housenumber: 29
addr:substreet: Acacia Avenue
addr:street: Acacia Avenue
addr:parentstreet: High Road
```

Because people can't see that a substreet is already set.

To me an entirely missing address being added as:
```
addr:housenumber: 42
addr:street: Acacia Avenue
```

[...]

PR #4490, opened 2022-10-12
Do not ask AddAddressStreet quest if addr:substreet or addr:parentstreet are present

mnalis:
This PR (as agreed in https://github.com/streetcomplete/StreetComplete/issues/4483#issuecomment-1276016771) skips the `AddAddressStreet` quest if either `addr:substreet` or `addr:parentstreet` are already defined, as answering this quest there would likely only lead to more confusing data, instead of actually helping OSM.
It also adds tests for checking if those are being ignored.

It [compiles](https://github.com/mnalis/StreetComplete/actions/runs/3237694006), [passes](https://github.com/mnalis/StreetComplete/actions/runs/3237694861) the [tests](https://github.com/mnalis/StreetComplete/blob/ignore-uk-parentstreet-substreet/app/src/test/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetTest.kt#L34-L52) and seems to work fine; e.g.  it skips for example this building: https://www.openstreetmap.org/way/230252831 (v6) which  lacks `addr:street`, but has `addr:substreet` (and on which the `AddAddressStreet` quest was previously being asked):

small_Screenshot_20221012_213319_de westnordost streetcomplete mn debug small_Screenshot_20221012_224128_de westnordost streetcomplete mn debug

Closes: https://github.com/streetcomplete/StreetComplete/issues/4483

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
@@ -17,8 +17,8 @@ class AddAddressStreet : OsmElementQuestType<StreetOrPlaceName> {
 
     private val filter by lazy { """
         nodes, ways, relations with
-          (addr:housenumber or addr:housename) and !addr:street and !addr:place and !addr:block_number
-          or addr:streetnumber and !addr:street
+          (addr:housenumber or addr:housename) and !addr:street and !addr:place and !addr:block_number and !addr:substreet and !addr:parentstreet
+          or addr:streetnumber and !addr:street and !addr:substreet and !addr:parentstreet
     """.toElementFilterExpression() }
 
     // #2112 - exclude indirect addr:street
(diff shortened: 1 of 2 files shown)


matkoniecz (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/address/AddAddressStreet.kt
```diff
@@ -17,8 +17,8 @@ class AddAddressStreet : OsmElementQuestType<StreetOrPlaceName> {
 
     private val filter by lazy { """
         nodes, ways, relations with
-          (addr:housenumber or addr:housename) and !addr:street and !addr:place and !addr:block_number
-          or addr:streetnumber and !addr:street
+          (addr:housenumber or addr:housename) and !addr:street and !addr:place and !addr:block_number and !addr:substreet and !addr:parentstreet
+          or addr:streetnumber and !addr:street and !addr:substreet and !addr:parentstreet
```
FloEdelmann, line 21: Could be refactored to

```suggestion
          !addr:street and (
            (addr:housenumber or addr:housename) and !addr:place and !addr:block_number
            or addr:streetnumber
          ) and !addr:substreet and !addr:parentstreet
```

But I don't know if that's easier to understand.

westnordost: I think it is fine this way

westnordost:
Looks fine, thank you!

Issue #4494, opened 2022-10-13
Separately mapped sidewalk is not detected


mnalis:
Sidewalk quest get asked even if there are nearby separate footways marked separately with `highway=footway`+`footway=sidewalk`. The sidewalk quest should be skipped in such cases.

**How to Reproduce**
- enable `sidewalk` quest and navigate for example to this way: https://www.openstreetmap.org/way/34926444
- notice that `Does this street have a sidewalk` quest gets asked
- notice that there are parallel sidewalks separately mapped already on both sides (on that particular segment of the street, there is not even an green buffer zone between street and sidewalk, like it is a little to the east)

![small_Screenshot_20221013_143327_de westnordost streetcomplete](https://user-images.githubusercontent.com/156656/195609048-7a99eed7-5eb7-4ad0-904b-47e0c2d4f644.jpg) ![josm_cut2](https://user-images.githubusercontent.com/156656/195611050-3bcc06d8-b68c-4762-915a-4e15f1b69342.png)

**Expected Behavior**
- SC should detect separately mapped footway and skip the quest, according to my understanding.

Code itself is unfortunately too complex for me to follow, but comments in the code seem to indicate it should work that way:

 https://github.com/streetcomplete/StreetComplete/blob/d215e40b2112fd2f803387a63e82f97a3f5a82ca/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt#L42-L45
  
**Versions affected**
upstream StreetComplete `v47.1`, Android 10

Same or similar problem seems to also be present in earlier versions too, e.g. https://www.openstreetmap.org/note/3186207, https://www.openstreetmap.org/note/3186207 etc

mnalis:
(and there are many similar examples relatively nearby; e.g. https://www.openstreetmap.org/way/804793377, https://www.openstreetmap.org/way/629025256, https://www.openstreetmap.org/way/27314747 etc.)

Helium314:
Sidewalk is only detected as sidewalk if it's close enough, for minimum distance at which a way is recognizes as a possible sidewalk see https://github.com/streetcomplete/StreetComplete/blob/d215e40b2112fd2f803387a63e82f97a3f5a82ca/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt#L78

mnalis:
[...]

PR #4499, opened 2022-10-14
include sidewalk width in calculation (and document it)

mnalis:
This PR:
- adds minimum sidewalk width to the calculation
- adds comments to better document what and why we do

Closes: https://github.com/streetcomplete/StreetComplete/issues/4494

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
@@ -113,6 +113,7 @@ private fun estimateCyclewaysWidth(tags: Map<String, String>, isLane: Boolean): 
     return leftWidth + rightWidth
 }
 
-private const val BROAD_LANE = 4f
-private const val LANE = 3f
+// those values are pessimistic by design, they should not be realistic
+private const val BROAD_LANE = 3f
+private const val LANE = 2f
 private const val SHOULDER = 2f
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
@@ -113,7 +113,7 @@ private fun estimateCyclewaysWidth(tags: Map<String, String>, isLane: Boolean): 
     return leftWidth + rightWidth
 }
 
-// those values are pessimistic by design, they should not be realistic
-private const val BROAD_LANE = 3f
-private const val LANE = 2f
+// when estimating, use lowest probable width to no over-estimate the road width
+private const val BROAD_LANE = 3.5f
+private const val LANE = 2.5f
 private const val SHOULDER = 2f
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
@@ -113,7 +113,7 @@ private fun estimateCyclewaysWidth(tags: Map<String, String>, isLane: Boolean): 
     return leftWidth + rightWidth
 }
 
-// when estimating, use lowest probable width to no over-estimate the road width
-private const val BROAD_LANE = 3.5f
-private const val LANE = 2.5f
+// rather under-estimate then over-estimate the road width
+private const val BROAD_LANE = 3.75f // standard interstate/motorway width
+private const val LANE = 2.75f // widest trucks are about 2.6m
 private const val SHOULDER = 2f
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
@@ -82,7 +82,7 @@ class AddSidewalk : OsmElementQuestType<LeftAndRightSidewalk> {
             (estimateRoadwayWidth(tags) ?: guessRoadwayWidth(tags)) +
             (estimateParkingOffRoadWidth(tags) ?: 0f) +
             (estimateCycleTrackWidth(tags) ?: 0f) +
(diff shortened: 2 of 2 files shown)


westnordost:
Did you read https://github.com/streetcomplete/StreetComplete/issues/4494#issuecomment-1278117729 ?

mnalis:
I did (well, first revision anyway; it seems github do not send emails with updated comments), but I'm not completely sure what you'd like me to do exactly... I've tried to guess, and did this:

- reduce sidewalk width to `1m` (way below cited **absolute minimum** of `1.5m` in US or minimum legal `1.8m` in my country)
- reduce LANE from `3m` to `2m` (i.e. highly unrealistic as **two SUVs wouldn't be able to pass each other without crashing**) and BROAD_LANE from `4m` to `3m`. That seems strange to me, but does work with your formula that residental with 2 lanes is estimated at 4m.

So, does this look more OK to you?
**_OR_** did you rather want those functions to estimate as they did, but then multiply their result with `0.75f` or something?

app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
```diff
@@ -113,6 +113,7 @@ private fun estimateCyclewaysWidth(tags: Map<String, String>, isLane: Boolean):
     return leftWidth + rightWidth
 }
 
-private const val BROAD_LANE = 4f
-private const val LANE = 3f
+// those values are pessimistic by design, they should not be realistic
+private const val BROAD_LANE = 3f
+private const val LANE = 2f
```
westnordost, line 118: okay let's....
```suggestion
// when estimating, use lowest probable width to no over-estimate the road width
private const val BROAD_LANE = 3.5f
private const val LANE = 2.5f
```

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
```diff
@@ -75,11 +75,14 @@ class AddSidewalk : OsmElementQuestType<LeftAndRightSidewalk> {
...
 
+    /* Calculate when footway is too far away from the road to be considered its sidewalk.
+       It is an estimate, and we deliberately err on the side of showing the quest too often. */
     private fun getMinDistanceToWays(tags: Map<String, String>): Float =
         (
             (estimateRoadwayWidth(tags) ?: guessRoadwayWidth(tags)) +
             (estimateParkingOffRoadWidth(tags) ?: 0f) +
-            (estimateCycleTrackWidth(tags) ?: 0f)
+            (estimateCycleTrackWidth(tags) ?: 0f) +
+            1f    // absolute minimum legal sidewalk width is usually above 1.5m, but be even more pessimistic
```
westnordost, line 85:
```suggestion
            1.5f    // assumed sidewalk width
```
let's assume 1.5m then...

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
```diff
@@ -75,11 +75,14 @@ class AddSidewalk : OsmElementQuestType<LeftAndRightSidewalk> {
...
         ) / 2f +
         4f // + generous buffer for possible grass verge
```
westnordost, line 87: and remove that, or reduce to ~2m or something.

westnordost (commented):
What about AddCycleway::getMinDistanceToWays?

app/src/main/java/de/westnordost/streetcomplete/osm/RoadWidth.kt
```diff
@@ -113,6 +113,7 @@ private fun estimateCyclewaysWidth(tags: Map<String, String>, isLane: Boolean):
     return leftWidth + rightWidth
 }
 
-private const val BROAD_LANE = 4f
-private const val LANE = 3f
+// when estimating, use lowest probable width to no over-estimate the road width
+private const val BROAD_LANE = 3.5f
+private const val LANE = 2.5f
```
westnordost, line 118: I've read the wikipedia article, I think this is fair.

```suggestion
// rather under-estimate then over-estimate the road width
private const val BROAD_LANE = 3.75f // standard interstate/motorway width
private const val LANE = 2.75f // widest trucks are about 2.6m
```

app/src/main/java/de/westnordost/streetcomplete/quests/sidewalk/AddSidewalk.kt
mnalis: In the example in the picture of https://github.com/streetcomplete/StreetComplete/issues/4494#issuecomment-1277812505, the grass verge was `3.85m`. That one is `highway=secondary` though, for `highway=residential` it may probably be narrower elsewhere, although in my city it often seems to be about `3-4m` size too (as often in has trees or shrub planted there). 

E.g. this (red arrow) josm measurement othogonal to `highway=residential` shows `3.49m` verge (and `7.60m` between middle of the road to the middle of the sidewalk):

josm2

(but I can change it whatever way you want, or you can of course change that directly if you find it more efficient)

mnalis: > What about AddCycleway::getMinDistanceToWays?

If I'm looking correctly, it seems to use same `estimateRoadwayWidth()`  functions, so would automatically use updated `const val LANE` etc.

https://github.com/streetcomplete/StreetComplete/blob/00f00f60b192bda332dae5249b5fdcd099148f29/app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt#L142-L147

mnalis: so current `+4f` seems to be OK for current `kerb+(gutter?)+grass verge` I guess?

westnordost: okay, fine

PR #4506, opened 2022-10-15
Use english overlay title for notes

Helium314:
Currently overlays use user locale for the overlay title, which leads to notes like _In context of "Straßenbeleuchtung" overlay_.
This PR copies the use of English resources from `AbstractOsmQuestForm`.


westnordost:
Oh, makes sense, thank you

Issue #4495, opened 2022-10-13
Allow "sidewalk is marked separately" other answer on  "What's the surface of the sidewalk here?" quest


mnalis:
**Use case**
I'm often offered quest `What's the surface of the sidewalk here?`, which wants to add `sidewalk:left:surface` and `sidewalk:right:surface` to the road. However, the sidewalks are already mapped separately as `highway=footway`.

It seems to happen when roads have old-style `sidewalk=both` / `sidewalk=left` tags. 
E.g. https://www.openstreetmap.org/note/3162147, https://www.openstreetmap.org/note/3112565, https://www.openstreetmap.org/note/3112572, etc.

In that case, user should have the "other answer" to say that the sidewalks are marked separately instead.

**Proposed Solution**
Show "other Answer" which would say _"the sidewalks are already mapped separately"_, and when chosen, fix the sidewalk tags as follows:

- `sidewalk=both` => `sidewalk:both=separate`
- `sidewalk=left` => `sidewalk:left=separate`
- `sidewalk=right` => `sidewalk:right=separate`

I can try to do the PR if that seems reasonable?

legofahrrad:
Why don't you just use the Sidewalk Overlay to change it?

legofahrrad:
Similar issue:  https://github.com/streetcomplete/StreetComplete/issues/3920, especially https://github.com/streetcomplete/StreetComplete/issues/3920#issuecomment-1165988346

mnalis:
> Why don't you just use the Sidewalk Overlay to change it?

For basically the same reason I don't long click and choose _"open in Vespucci"_. 
Overlay seems fine for longer-time editing of one feature, but I find it **_very cumbersome for one-off edits_** so far.

<Details>
E.g. instead of:

- click other other
- click _"the sidewalks are already mapped separately"_

With _Sidewalk Overlay_ I instead have to:
- click hamburger
- click overlay
- find and click "Sidewalks"
- click "OK" (and not outside of window which I often do, which cancels the action)
- click on the road
- click on the left side
- verify situation on the screen (zooming and panning as needed) and choose "separate" or "no" to click on it
- click on the right side
- verify situation on the screen (zooming and panning as needed)  and choose "separate" or "no" to click on it
- click checkbox to confirm
- click hamburger
- click overlays
- find and click previous overlay I was using (usually shops or maybe addresses)
- click "OK" (and not outside of window which I often do, which cancels the action)

</details>

[...]

Issue #3920, opened 2022-03-26
Sidewalk surface quest: Enable to change the sidewalk tag


cyclingcat:
Hi there,

the sidewalk surface quest currently cannot be solved correctly if the `sidewalk` tag itself is wrong. For example, the quest asks for the sidewalk surfaces on both sides separately although a sidewalk exists on only one side (or vice versa), due to an earlier tagging error or a change at this place. Currently there is no other possibility than to leave a note, but this situation occured surprisingly often, at least during my first attempts with this new quest.

**Ideas:**

1. Give the user the possibility to answer "The sidewalk situation is other than shown here" (or similar, comparing to the "Other surface" menu item in the smoothness quest), which deletes the `sidewalk` tag - the sidewalk quest would immediately reappear and after this one is solved, the sidewalk surface quest reappears, which can be solved correctly afterwards.

2. Add a button "No sidewalk" to the selection of possible surfaces, which changes the `sidewalk` tag appropriately. This approach alone doesn't cover the situation when there are sidewalks on both sides of the street when it's currently mistagged as one-sided.

BTW: Contrary to the road surface quest, the list of the possible surfaces doesn't get reordered by recent choices. Is this a deliberate decision? For example, the fourth item "Concret lanes" (= "Fahrspuren aus Beton") always stays at its quite high position despite the fact that it probably won't get selected very often.

The cycling cat

arrival-spring:
(I implemented this quest)

I did think about this when making the quest (see https://github.com/streetcomplete/StreetComplete/issues/1593#issuecomment-1029403540 and https://github.com/streetcomplete/StreetComplete/pull/3735#issuecomment-1038111396)

And using https://ent8r.github.io/NotesReview/ I can find no open notes yet from this quest.

I think option 2 would not be consistent with other quests so would favour option 1.

Regarding reordering it's to do with the interface being used and it would take a fair bit of work to make it order by recently used.

mnalis:
I also think option 1 would be ideal.

I actually opened a note several days ago to raise this exact issue (https://www.openstreetmap.org/note/3095650)  after it happened to me several times, but didn't get to reporting it yet. You don't see it in notes list as I've opened note manually, not as answer to a quest (as that seems to block quests on that element, so I tend to avoid doing that)

cyclingcat:
> And using https://ent8r.github.io/NotesReview/ I can find no open notes yet from this quest.

[...]

PR #4508, opened 2022-10-15
Add "The sidewalk situation is different" OtherAnswer to sidewalk surface quest

mnalis:
When  _"The sidewalk situation is different"_, it removes `sidewalk=*` and `sidewalk:*=*` tags, allowing the user to subsequently give correct on-the-ground answer. It also adds new test to check new option works correctly.

(When reviewing, you may want to use ["?w=1"](https://github.com/streetcomplete/StreetComplete/pull/4508/files?w=1) to ignore whitespace changes in GitHub as I needed to restructure some code making the diff seem bigger than it is).

Closes: https://github.com/streetcomplete/StreetComplete/issues/4495
(also closes https://github.com/streetcomplete/StreetComplete/issues/3920)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
@@ -6,6 +6,7 @@ import androidx.appcompat.app.AlertDialog
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.osm.sidewalk.Sidewalk
 import de.westnordost.streetcomplete.osm.sidewalk.createSidewalkSides
+import de.westnordost.streetcomplete.quests.AnswerItem
 import de.westnordost.streetcomplete.quests.AStreetSideSelectForm
 import de.westnordost.streetcomplete.view.controller.StreetSideDisplayItem
 import de.westnordost.streetcomplete.view.controller.StreetSideSelectWithLastAnswerButtonViewController.Sides.BOTH
@@ -17,6 +18,10 @@ class AddSidewalkSurfaceForm : AStreetSideSelectForm<Surface, SidewalkSurfaceAns
 
     private var leftNote: String? = null
     private var rightNote: String? = null
+
+    override val otherAnswers = listOf(
+        AnswerItem(R.string.quest_sidewalk_answer_different) { applyAnswer(SidewalkIsDifferent) }
+    )
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
@@ -86,7 +86,7 @@ class AddSidewalkSurfaceForm : AStreetSideSelectForm<Surface, SidewalkSurfaceAns
         if (left?.shouldBeDescribed != true && right?.shouldBeDescribed != true) {
             streetSideSelect.saveLastSelection()
         }
-        applyAnswer(SidewalkSurfaceAnswer(
+        applyAnswer(SidewalkSurface(
             left?.let { SurfaceAnswer(it, leftNote) },
             right?.let { SurfaceAnswer(it, rightNote) }
         ))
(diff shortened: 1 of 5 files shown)


mnalis:
I need help on this, if some of Kotlin programmer have some spare time. It should be complete and test added, but if fails to compile.
 
I've tried to copy what other quests do, but even after dozen tries I still cannot seem to get it right.

The [compile fails](https://github.com/mnalis/StreetComplete/actions/runs/3256130817/jobs/5346188924) with errors like `Unresolved reference: it` etc.:

```
e: /home/runner/work/StreetComplete/StreetComplete/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt: (55, 42): Unresolved reference: left
> Task :app:compileDebugKotlin FAILED
34 actionable tasks: 34 executed
e: /home/runner/work/StreetComplete/StreetComplete/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt: (55, 73): Unresolved reference: it
e: /home/runner/work/StreetComplete/StreetComplete/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt: (56, 43): Unresolved reference: right
e: /home/runner/work/StreetComplete/StreetComplete/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurface.kt: (56, 75): Unresolved reference: it
```

I'm pretty sure it's some simple thing I'm missing, but as I do not actually know anything about kotlin programming, it is eluding me.

Help?

Helium314:
I did not try, but my first guess is that the compiler does not "know" that `else` (line 54) implies that `answer` is a `SidewalkSurface`. You could try replacing `else` with `is SidewalkSurface`.

mnalis:
This now [successfully compiles](https://github.com/mnalis/StreetComplete/actions/runs/3277233982) (thanks @Helium314 for hint!), seems to work fine when manually testing (i.e. removes sidewalk related tags when chosen) and also [pass gradle tests](https://github.com/mnalis/StreetComplete/actions/runs/3277154883).

So, ready for review.

app/src/main/res/values/strings.xml
```diff
@@ -1311,6 +1311,7 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_sidewalk_value_yes">sidewalk</string>
     <string name="quest_sidewalk_value_no">no sidewalk</string>
     <string name="quest_sidewalk_value_separate">sidewalk, but displayed separately on map</string>
+    <string name="quest_sidewalk_answer_different">The sidewalk situation is different</string>
```
westnordost, line 1314:
```suggestion
    <string name="quest_sidewalk_answer_different">Sidewalk situation is different</string>
```

Shorter text because space in popup is limited

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
```diff
@@ -18,6 +19,10 @@ class AddSidewalkSurfaceForm : AStreetSideSelectForm<Surface, SidewalkSurfaceAns
     private var leftNote: String? = null
     private var rightNote: String? = null
 
+    override val otherAnswers = listOf(
+        AnswerItem(R.string.quest_sidewalk_answer_different) { applyAnswer(SidewalkIsDifferent) }
```
westnordost, line 23: For other answers that tag something, we always ask the user if he is sure **if I remember correctly**, so it would be a consistent UX to do that. See e.g. the AddSmoothnessForm, where you also got the option to say that it is a different surface.

westnordost (changes requested):
One remark, otherwise looks good

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddSidewalkSurfaceForm.kt
mnalis: > For other answers that tag something, we always ask the user if he is sure **if I remember correctly**

It seems to be dependent on the quest, some change it without extra confirming, for example:
https://github.com/streetcomplete/StreetComplete/blob/d59747374351de6e4328a0c0f5f0a334e678b6e3/app/src/main/java/de/westnordost/streetcomplete/quests/bench_backrest/AddBenchBackrestForm.kt#L17-L19

or `AddBusStopShelterForm`, `TactilePavingCrosswalkForm`, `WayLitForm`, `AddReligionForm`, ...

As it is not consistent and we already have both situations in different quest; in this specific case it seemed (to me) better choice to not to bother the user with (in my opinion needless) extra confirmation, especially as the following quest will immediately ask for new sidewalk situation. 

But if you think it would be better to ask for confirmation anyway in this case, let me know and I'll implement it.

westnordost: I see, if it is not consistent, then I shall not force you to chase this.

westnordost (approved):

PR #4509, opened 2022-10-15
Fix jagged puzzle images on high resolution screens

tapetis:
This PR removes the gaps between the street images and fixes the jagged display of images in the `StreetSideSelectPuzzle` on high resolution screens. The main cause of these issues is the interactions of Android's density-independent pixels and the implemented rounding operations. For certain display densities, the assumed width of the `StreetParkingDrawable` was off by one compared to the drawables drawn to its canvas. 

Before:
Jagged

After:
Not Jagged

The changes in `StreetParkingDrawable.kt` fix the gaps between the two street drawables that make up one street icon. The change in `StreetSideSelectPuzzle.kt` fixes the gap between the repeated drawables. Finally, the increased margins in `view_side_select_puzzle.xml` remove the jagged edges between the top and bottom halves of the street.

I tested my changes with the affected quests and could not find any problems.


app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingDrawable.kt
```diff
@@ -56,7 +57,7 @@ class StreetParkingDrawable(
         val backgroundResId = getStreetDrawableResId(parkingOrientation, parkingPosition)
         val nyanResId = if (isApril1st()) R.drawable.car_nyan else null
 
-        for (y in 0..(ceil(1.0 * bounds.height() / height).toInt())) {
+        for (y in 0 until ceil(bounds.height() / height).toInt()) {
```
westnordost, line 60: Why replace .. with until?

PR #4510, opened 2022-10-15
Disable Jetifier

tapetis:
This PR disables the Jetifier tool tool, which was used to migrate support-library-dependent libraries to rely on the equivalent AndroidX packages. As can be seen from the screenshot of Android Studio's build analyzer below, there are no longer any dependencies that require the use of the Jetifier tool. Turning it off didn't make a noticeable difference in build times for me, though.

Jetifier


westnordost:
Cool! Did you test it though?

Issue #3567, opened 2021-12-06
Resurvey quest dismissed by splitting way


lukashass:
After answering a resurvey quest (.i.e a quest for a value that is already present but old) with "Differs along the way..." and splitting the way no new quests appear for the newly created split ways. I had this problem with a "What is the surface of this path?" quest, but this probably applies to others, as well.

I suspect that no quests appear for the split ways since they have been created/updated by splitting and thus are not eligible for resurvey.

**How to Reproduce**

1. Answer a resurvey quest with "Differs along the way..." and split the way.
2. :boom: No new quests appear for the split ways.

**Versions affected**

StreetComplete v38.0

HolgerJeromin:
Resurvey checks for the "last touched for x years" state of a osm way.
By splitting you touch the way, so the resurvey is not triggered afterwards.
I got bitten by that a few times. 
All fixes for that were hacks so IMO no one was implemented right now.

westnordost:
> I suspect that no quests appear for the split ways since they have been created/updated by splitting and thus are not eligible for resurvey.

Yes, correct. I am sorry but this cannot be solved because OSM has no notion of keeping change-dates for single tags, only for the whole element. If a `check_date:<key>` tag is present, it will be used correctly but if there is none, you are out of luck.

westnordost:
(Also anyway another problem is that there is no notion of splitting ways in OSM. Splitting ways in OSM is a hack: Actually new ways are created, thus no history carries over to the new way(s), it is treated like a new object.)

matkoniecz:
as workaround you may do edit in Vespucci, create note or wait few years for quest to reappear

riQQ:
> (Also anyway another problem is that there is no notion of splitting ways in OSM. Splitting ways in OSM is a hack: Actually new ways are created, thus no history carries over to the new way(s), it is treated like a new object.)

Just to add more information. The original way id with its history is preserved in the chunk with the most nodes. But the other chunks are new ways without any history.

https://github.com/streetcomplete/StreetComplete/blob/e96252ff6beed429393313f031dbb1377a1049eb/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/split_way/SplitWayAction.kt#L114-L129

FloEdelmann:
Could a workaround be setting `check_date` to the last modification date when splitting a way?

[...]

PR #4523, opened 2022-10-17
Don't change timestamp when splitting a way

Helium314:
Fixes resurvey quests disappearing when splitting a way (only with auto-sync off), as discussed in #3567.
Tested and works.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/split_way/SplitWayAction.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/split_way/SplitWayAction.kt
@@ -131,9 +131,10 @@ private fun getSplitWayAtIndices(
 
     return nodesChunks.mapIndexed { index, nodes ->
         if (index == indexOfChunkToKeep) {
-            Way(originalWay.id, nodes, tags, originalWay.version, currentTimeMillis())
+            // keep the original timestampEdited, so resurvey quests are still shown after splitting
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/split_way/SplitWayAction.kt
```diff
@@ -131,9 +131,10 @@ private fun getSplitWayAtIndices(
 
     return nodesChunks.mapIndexed { index, nodes ->
         if (index == indexOfChunkToKeep) {
-            Way(originalWay.id, nodes, tags, originalWay.version, currentTimeMillis())
+            // keep the original timestampEdited, so resurvey quests are still shown after splitting
```
matkoniecz, line 134: as it refers to both Way creations, maybe keep it one level higher - before `if (index == indexOfChunkToKeep)`?

matkoniecz:
Also, is it working only with auto-sync disabled? In such case maybe also mention it in the comment.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/split_way/SplitWayAction.kt
westnordost: The comment? makes sense

Helium314:
Makes sense, both done

PR #4524, opened 2022-10-18
Store last side selection for only one side separately

### LINKED ISSUE #4503: Last answer shortcut button for only one streetside
**Use case**
In the quest for sidewalk surfaces (for example) the previous answer is displayed as a shortcut in the bottom left of the quest popup. However, when the quest only asks for one side (e.g. quest for `sidewalk:right:smoothness` becaue `sidewalk=right` is already set), this shortcut doesn't show up.

Other examples include the `cycleway` and `shoulder` quests on `highway`s with `oneway=yes`.

**Proposed Solution**
It might just be a matter of changing one if-statement, but I'm not sure where the relevant code is. (Is it https://github.com/streetcomplete/StreetComplete/blob/00f00f60b192bda332dae5249b5fdcd099148f29/app/src/main/java/de/westnordost/streetcomplete/view/controller/StreetSideSelectWithLastAnswerButtonViewController.kt?)

Helium314:
This fixes #4503. It's closed, but I decided to give it a try because I regularly get annoyed by not being able to repeat answers quickly.

If there is only one side, left and right are stored separately. This might lead to unexpected last picked values being shown sometimes, but storing separately might be necessary to ensure correct `(de)serializeLastSelection`.


westnordost:
Surprisingly little code. 

Why did you choose to have different persisted selections for each the left side and the right side only? (Which one is applied depends on how you rotated the map)

Helium314:
> Why did you choose to have different persisted selections for each the left side and the right side only? (Which one is applied depends on how you rotated the map)

I wasn't sure whether calling `serializeLastSelection(it, true)` to store and `deserializeLastSelection(it, false)` is guaranteed to work.
Currently `serializeLastSelection` never depends on the side (i.e. ignores true/false), and thus storing only one value works as well.

westnordost:
Good point. As long as that functionality is there, it should not be broken by introducing a conflicting change. If you would say it would make sense from a UI point of view to only save one (that is applied to either left or right side), you could investigate why the interface is like this right now and whether this is obsolete and will be obsolete.
There must have been a reason why this was introduced in the first place.

Helium314:
Looks like it was added in https://github.com/streetcomplete/StreetComplete/commit/c89f12ac5fbe443be223c66b66d34a568651419b#diff-ccb08f0f1d63cea747239a61ae7ceaa85b4268a8c4ae97604d1fb951b0b964df, but `isRight` seems to be unused there.

westnordost:
Yeah well, that's the refactor I did to unify of the code in the different classes.

peternewman:
> Why did you choose to have different persisted selections for each the left side and the right side only? (Which one is applied depends on how you rotated the map)

It seems sane to me as I'd commented elsewhere, if you consider a road:
`=_=-=_=`

Where the pavement is consistent on each side of the road (when present), but different between the sides, a choice like that simplifies the tagging, whereas if there was only one historic value for not "both" sides, it would always be wrong.

Helium314:
> Yeah well, that's the refactor I did to unify of the code in the different classes.

Is (de)serialize really dependent on left/right here? It looks like serialize was just
`serializeLastSelection(selection: LastSelection<I>): String = "${selection.left}#${selection.right}"`,
with both `left` and `right` being `<I?>`

> Where the pavement is consistent on each side of the road (when present), but different between the sides, a choice like that simplifies the tagging, whereas if there was only one historic value for not "both" sides, it would always be wrong.

Good point. Though there values might be switched depending on how the map is rotated...

peternewman:
> Good point. Though there values might be switched depending on how the map is rotated...

But that's also true of the two sided last selection too IIRC. From a practical point of view, if you're walking along that road it seems unlikely you'd randomly spin the map 180 degrees for no good reason!

app/src/main/java/de/westnordost/streetcomplete/view/controller/StreetSideSelectWithLastAnswerButtonViewController.kt
```diff
@@ -171,38 +175,59 @@ class StreetSideSelectWithLastAnswerButtonViewController<I>(
...
-        val l = if (isUpsideDown) lastSelectionRight else lastSelectionLeft
-        val r = if (isUpsideDown) lastSelectionLeft else lastSelectionRight
+        val l: StreetSideDisplayItem<I>?
+        val r: StreetSideDisplayItem<I>?
+        if (showSides == Sides.BOTH) {
+            l = if (isUpsideDown) lastSelectionRight else lastSelectionLeft
+            r = if (isUpsideDown) lastSelectionLeft else lastSelectionRight
+        } else {
+            l = if (isUpsideDown) lastSelectionOnlyRight else lastSelectionOnlyLeft
+            r = if (isUpsideDown) lastSelectionOnlyLeft else lastSelectionOnlyRight
```
westnordost, line 229: But, does that work? Let's say you have 
- lastSelectionOnlyLeft  = cycle track
- lastSelectionOnlyRight = cycle lane


Now, there is a oneway road for which only the right side is displayed. Hence, the last answer button is displayed as cycle lane. But then, you rotate the road upside down. Now, according to this logic, tapping on the last answer button which **displays** a cycle lane is going to add a cycle track, no?

Helium314: Meanwhile I also found some cases where this didn't work properly... I'll have a look at it later (converting to draft while this is not solved)

westnordost: I think we did establish in the main discussion that the left/right serialization is a relict from the past that is actually not used anymore anywhere. So maybe the better solution would be to have just one `lastSelectionOneSideOnly` instead.

Helium314: I removed the serialization, and now only one value is stored. But I didn't switch to `lastSelectionOneSideOnly`, because in deserialization the side is relevant.

westnordost: Oh.

So that's why there is this `isRight: Boolean` in both the serializer and deserializer

westnordost: How do you changes solve the issue that when you turn the map (at most) 180°, a different not previously shown item is suddenly depicted on the button?

Helium314: The shown last solution for one side is the same (from same settings value), only deserialized for the correct side.

westnordost: I don't understand.

Helium314:
```kotlin
        lastSelectionOnlyLeft = prefs.getString("$lastSelectionPreferencePrefix.oneSide", null)?.let { deserializeLastSelection(it, false) }
        lastSelectionOnlyRight = prefs.getString("$lastSelectionPreferencePrefix.oneSide", null)?.let { deserializeLastSelection(it, true) }
```
Both `lastSelectionOnlyLeft` and `lastSelectionOnlyRight` use the value stored in `"$lastSelectionPreferencePrefix.oneSide"`. The only difference is whether `isRight` is set to true in `deserializeLastSelection`

westnordost:
Code looks fine. Though I am not sure about the 180° thing... gonna test it out.

westnordost:
Works well, thank you!

PR #4543, opened 2022-10-23
Update deprecated github actions

mnalis:
This PR fixes GitHub compilation deprecation warnings by using newer version of actions `v3` instead of `v2`, before GitHub removes support completely. 
Tested and it makes warnings go away, while action results work fine (as it did before depreciations).

---

GitHub was complaining  that:

> Node.js 12 actions are deprecated. For more information see:
  https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
  Please update the following actions to use Node.js 16

and 
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more
information see:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build-debug-apk.yml
+++ b/.github/workflows/build-debug-apk.yml
@@ -12,9 +12,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: set up JDK 11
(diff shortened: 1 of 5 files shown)


FloEdelmann:
Thanks!

PR #4544, opened 2022-10-23
Fix crash when onDestroy is not called in StyleableOverlayManager

### LINKED ISSUE #4540: Crash with overlay active and downloading after reentering the app
I frequently get reported this crash report now.

I am not completely sure how to reproduce it, but it is relatively recent and it seems to only happen when an overlay is active but the user left the app for a while, then returned to it, then downloaded something.

My first suspicion would be that this `StyleableOverlayManager` is still lingering from before the app was sent to background and while its parent `MainFragment` had been destroyed and recreated already, along with the tangram `MapController`, it is still around because it still listens to the `MapDataWithEditsSource`.
I.e. for some reason, it does not unsubscribe on updates from `MapDataWithEditsSource` when its parent `MainFragment`s view is destroyed.

Since it is relatively recent that this occurs, (can't say exactly when - at least 47.1), could this have something to do with `updateJob`, @Helium314 ? Though, I don't see anything wrong with the code... maybe I am missing something...

If anybody can find an exact reproduction, it would be appreciated too. Since this issue seems to be happening so often, I would really like to get this fixed before releasing v48.0 (which I *planned* to release today 🤷 ).

```
java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mapzen.tangram.NativeMap.getCameraPosition(com.mapzen.tangram.CameraPosition)' on a null object reference
	at com.mapzen.tangram.MapController.getCameraPosition(MapController.java:496)
	at de.westnordost.streetcomplete.screens.main.map.tangram.CameraManager.pullCameraPositionFromController(CameraManager.kt:169)
	at de.westnordost.streetcomplete.screens.main.map.tangram.CameraManager.getCamera(CameraManager.kt:63)
	at de.westnordost.streetcomplete.screens.main.map.tangram.KtMapController.getCameraPosition(KtMapController.kt:179)
	at de.westnordost.streetcomplete.screens.main.map.StyleableOverlayManager.onNewScreenPosition(StyleableOverlayManager.kt:110)
	at de.westnordost.streetcomplete.screens.main.map.StyleableOverlayManager$mapDataListener$1.onReplacedForBBox(StyleableOverlayManager.kt:72)
	at de.westnordost.streetcomplete.data.osm.edits.MapDataWithEditsSource.callOnReplacedForBBox(MapDataWithEditsSource.kt:462)
	at de.westnordost.streetcomplete.data.osm.edits.MapDataWithEditsSource.access$callOnReplacedForBBox(MapDataWithEditsSource.kt:32)
	at de.westnordost.streetcomplete.data.osm.edits.MapDataWithEditsSource$mapDataListener$1.onReplacedForBBox(MapDataWithEditsSource.kt:137)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController.onReplacedForBBox(MapDataController.kt:271)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController.putAllForBBox(MapDataController.kt:91)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataDownloader$download$2.invokeSuspend(MapDataDownloader.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
```

tapetis:
This PR fixes #4540. I was not able to reproduce the crash with these changes.

While I haven't been able to find any problems with the overlays during testing, I don't know if there is some logic in the app that was based on the previous leaky behavior.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/EditHistoryPinsManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/EditHistoryPinsManager.kt
@@ -34,15 +34,17 @@ class EditHistoryPinsManager(
     private val resources: Resources
 ) : DefaultLifecycleObserver {
 
-    /** Switch active-ness of edit history pins layer */
-    var isActive: Boolean = false
+    private val viewLifecycleScope: CoroutineScope = CoroutineScope(SupervisorJob())
+
+    /** Switch visibility of edit history pins layer */
+    var isVisible: Boolean = false
         set(value) {
             if (field == value) return
             field = value
-            if (value) start() else stop()
+            if (value) show() else hide()
         }
 
-    private val viewLifecycleScope: CoroutineScope = CoroutineScope(SupervisorJob())
+    private var isStarted: Boolean = false
 
     private val editHistoryListener = object : EditHistorySource.Listener {
         override fun onAdded(edit: Edit) { updatePins() }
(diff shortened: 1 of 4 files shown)


westnordost:
Thank you, I will apply it in modified form. If this indeed solves the crash (we will see), then as a follow-up, we should go through all `onDestroy` methods and see what we can move to `onStop` etc.

PR #4546, opened 2022-10-23
update to KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED 

mnalis:
Bigger list of keys to remove, following major update of `isShopExpressionFragment()` introduced in https://github.com/streetcomplete/StreetComplete/commit/4f462a1d15c1eb9660e879901c403a4af4ff7be8

Closes: https://github.com/mnalis/StreetComplete-taginfo-categorize/issues/11


westnordost:
Nice, thank you @mnalis!

westnordost:
Would you say that this approach still makes sense though, looking at the size of it? E.g. maybe move back to an keys-that-should-not-be-removed list?

mnalis:
Yes, I still think it is good idea not to remove keys we don't understand.

[keys-that-should-not-be-removed](https://github.com/mnalis/StreetComplete-taginfo-categorize/blob/master/sc_to_keep.txt) list is not much shorter either...

matkoniecz:
> E.g. maybe move back to an keys-that-should-not-be-removed list?

That is also obnoxiously long list, with missing entries having worse effects.

@mnalis Thanks for maintaining that!

Issue #4538, opened 2022-10-21
Parking access quest asked for HGV-only parking lots


tapetis:
In Germany (and probably many other countries) it is possible to specify which vehicles may be parked in a parking lot or on the side of the road. An additional sign (e.g. Zusatzzeichen 1010-51 for HGVs) can be attached to the parking sign (Zeichen 314) for this purpose (see [this](https://www.stvo2go.de/verkehrszeichen-parken-zusatzzeichen/) description in German for more details).

In OSM, many HGV parking spaces are currently tagged as just `amenity=parking` and `hgv=designated`. It can be argued that one should also tag `access=no` or `motor_vehicle=no` to make it clearer that only trucks may park there (see [here](https://wiki.openstreetmap.org/wiki/DE_talk:Tag:amenity%3Dparking#LKW_Parkplatz) and [here](https://forum.openstreetmap.org/viewtopic.php?id=68542)). However, this is often not the case. The same problem exists with bus parking, for example.

When there is no `access` tag, StreetComplete ask the user "Who is allowed to park here?" and in most cases adds the `access=yes` tags to these parking areas. Afterwards, the area has wrong tags and is displayed as if cars are also allowed to park there as specified by the [access hierarchy](https://wiki.openstreetmap.org/wiki/Key:access#Land-based_transportation).

Here are some `amenity=parking` that were previously only tagged as `hgv=designated`, which now also have `access=yes` tagged by StreetComplete:

- https://www.openstreetmap.org/way/353163903/history ([Mapillary](https://www.mapillary.com/app/?pKey=589521965364323)),
- https://www.openstreetmap.org/way/493334668/history ([Mapillary](https://www.mapillary.com/app/?pKey=564130354549717)), and
- ~~https://www.openstreetmap.org/way/294331631/history ([Mapillary](https://www.mapillary.com/app/?pKey=299076235049840))~~.

As can be seen from the Mapillary links, the first two linked parking lots are more or less clearly designated as truck parking. More ways like this can be found easily using [overpass](https://overpass-turbo.eu/s/1mWn).

I think the app should handle these truck parking lots that are tagged only as `hgv=designated` better. A straightforward solution would be to simply not display the `AddParkingAccess` quest for such parking lots. Another solution would be for the quest to allow the user to specify that only certain types of vehicles are allowed to park, rather than only adding the general `access` tag.

westnordost:
[...]

PR #4547, opened 2022-10-24
If more specific access (like hgv=*) is already specified, skip the quest

mnalis:
We skip them, in order to not introduce more problems.

PR [compiles sucessfully](https://github.com/mnalis/StreetComplete/actions/runs/3313012182) and seems to work fine. (e.g. no longer asks for parking access on [w79658967](https://www.openstreetmap.org/way/79658967), while normally asking elsewhere).
 
Closes: https://github.com/streetcomplete/StreetComplete/issues/4538


westnordost:
Looks good

Issue #4526, opened 2022-10-18
Don't ignore contradictory `*:surface=*` and `surface=*` tags


Discostu36:
**Use case**
When SC asks for `footway:surface=*` and `cycleway:surface=*` on a segregated `highway=path`, it does not change `surface=*`, even when it is contradictory. 

See for example [this path](https://www.openstreetmap.org/way/134592975).
<table>
<tr><td>biclycle</td><td>designated</td></tr>
<tr><td>cycleway:surface</td><td>paving_stones</td></tr>
<tr><td>foot</td><td>designated</td></tr>
<tr><td>footway:surface</td><td>paving_stones</td></tr>
<tr><td>highway</td><td>path</td></tr>
<tr><td>segregated</td><td>yes</td></tr>
<tr><td>surface</td><td>asphalt</td></tr>
</table>

**Proposed Solution**
Three possibilities (in ascending effort):
1. StreetComplete should delete a mismatching `surface=*` tag 
2. StreetComplete should change it according to `footway:surface` and `cycleway:surface`, but only if both are the same
3. StreetComplete should also change `surface=*` tag if `footway:surface` and `cycleway:surface` are not the same, but both are paved or unpaved (for example `footway:surface=paving_stones` and `cycleway:surface=asphalt` → `surface=paved`). If one way is paved and the other one unpaved, it should add them via semicolon (e.g. `surface=asphalt;dirt`).

mnalis:
Make sense to me.

If someone wants to give it a try, leave a comment stating so. 
(if noone steps up, I can give it a try in few days, for option `2`  with fallbacking to `1` or maybe `3`)

Discostu36:
One thing I forgot about is that options 2 and 3 of course could also be used to add missing `surface`-Tag, not only change it.

RubenKelevra:
> One thing I forgot about is that options 2 and 3 of course could also be used to add missing `surface` tag, not only change it.

Nah, if there are two explicit surface information, I would keep them both. Explicit is always better than implicit IMHO. For example that could be the only part of the way which has the same surface and in the next part they differ again.

With just one tag I would second guess if the information is complete for this part.

mnalis:
> Nah, if there are two explicit surface information, I would keep them both. Explicit is always better than implicit IMHO

Agreed, but if I understood @Discostu36 correctly (hopefully I did, as i did the PRs :smile:), it would keep both of them, **and then add** `surface` too. E.g. final result of such tagging would be:

[...]

PR #4548, opened 2022-10-24
Update surface when cycleway:surface and footway:surface match

mnalis:
(or remove `surface=*` if they _don't_ match)

[Compiles OK](https://github.com/mnalis/StreetComplete/actions/runs/3315165654) and seems to work as expected, e.g.:

- when quests answer with `cycleway:surface` same as `footway:surface` (or other way around), it also sets `surface=*` to the same value
- if  `cycleway:surface` and  `footway:surface` are different, it removes `surface=*` tag (if existing)

Closes: https://github.com/streetcomplete/StreetComplete/issues/4526


westnordost:
Does this work? Better add unit tests!

mnalis:
(pinging @matkoniecz as original author of the quest for his opinion too; but everyone is invited and welcomed to give their opinion)

Helium314:
> is that about null handling of foot (i.e. if foot is not defined, foot != no evaluates to true?)

Yes, `foot != no` is true when there is no `foot`.

Helium314:
This will remove `surface=paved` e.g. if cycleway surface is asphalt and footway surface is paving stones. Not sure if this is desired...

Helium314:
I think that adding `paved` is not relevant, but there is not much complexity involved.
You simply need to
```kotlin
        if (tags["cycleway:surface"] != null && tags["footway:surface"] != null) {
            if (tags["footway:surface"] != tags["cycleway:surface"] && tags["footway:surface"] in ANYTHING_FULLY_PAVED && tags["cycleway:surface"] in ANYTHING_FULLY_PAVED) {
                tags["surface"] = "paved"
            }
        }
```

mnalis:
Thanks @Helium314 ; @matkoniecz also contacted me with information that most of that code/constants has already been implemented in https://github.com/matkoniecz/Zazolc/blob/overlays/app/src/main/java/de/westnordost/streetcomplete/osm/surface/Surface.kt#L121 which should hopefully be submitted soon, so I think I'll avoid duplicating that code for now.

So the general idea here is to fix the two failing tests (or rather, improve on the quest logic?) in this PR so it can get in, and then after @matkoniecz submits and merges his code, those constants can be reused for those quests too in another PR (unless someone objects to the idea)

mnalis:
Ok, tests are ready and [pass now](https://github.com/mnalis/StreetComplete/actions/runs/3357973659), so ready for next review.

Note that extensive automated tests I've added discovered few [inconsistencies with _**previous**_ code](https://github.com/streetcomplete/StreetComplete/pull/4548#issuecomment-1290703523) (mostly in handling access permissions), so I've fixed those too.

westnordost:
> Note that extensive automated tests I've added discovered few https://github.com/streetcomplete/StreetComplete/pull/4548#issuecomment-1290703523 (mostly in handling access permissions), so I've fixed those too.

Could you please put this into an own PR?

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -25,7 +25,11 @@ class AddCyclewayPartSurface : OsmFilterQuestType<SurfaceAnswer>() {
             and !note:cycleway:surface
           )
         )
-        and (access !~ private|no or (foot and foot !~ private|no) or (bicycle and bicycle !~ private|no))
+        and (
+          !access and bicycle !~ private|no
+          or access and access !~ private|no and bicycle !~ private|no
+          or access and access  ~ private|no and bicycle and bicycle !~ private|no
+        )
```
westnordost, line 32: Why this?

westnordost:
Or never mind, that will be too much work.

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
mnalis: This addresses second bullet point of this comment: https://github.com/streetcomplete/StreetComplete/pull/4548#issuecomment-1290703523

Basically, the access logic didn't work correctly in original code (i.e. pre-this PR).
For example, the `cycleway surface` was still being asked for `bicycle=private` etc. (which, as I understand it, shouldn't happen).

So I've written in way that is:
- actually working correctly (passes the tests that check for all combinations)
- more readable (IMHO), as it checks each general `access=*` state separately on each line
- (taking a little more space)

mnalis: > quests in general are asked for ways that are either accessible by default, on foot or with the vehicle type you ought to use there

I agree (and that is what I've tried to do), but the English sentence above have the similary problem the code had; and it is interaction between those **cannot** be defined with simple **"or"** (e.g. `foot=no` overrides `access=yes`).

> How much of a theoretical case is this, though?

That I don't know (and I worry it is unlikely that ad-hoc overpass queries trying to find out would survive such attempts on word-wide queries)


> I am still struggling what exactly the cited section of code filters.... would it not evaluate to false for a way that has access=no but which you can access on foot?

It would evaluate to `false`, and correctly IMHO, because `foot` does not matter at all in `cycleway surface` quest. (You qouted `AddCyclewayPartSurface.kt` here!). There is very similar (but with key names changed) code in `AddFootwayPartSurface.kt` where your example of `access=no+foot=yes` would evaluate to `true`:

```
          !access and foot !~ private|no
          or access and access !~ private|no and foot !~ private|no
          or access and access  ~ private|no and foot and foot !~ private|no
```

The new logic seems quite readable to me:
- if general `access` is not defined, check if `foot` specifically is disallowed
- if general `access` is allowed, check if `foot` overrides this
- if general `access` is disallowed, check if `foot` overrides this

There might be **_shorter_** ways to write this, but much **_less readable_** (in my opinion). I found original code not only give incorrect results, but was also treacherous/misleading and confusing (partly due to how SC handles `null` values. YMMV)

> Also, for consistency, such change necessitates also changing dozens of other quest filters to be consistent (hence my initial knee-jerk reaction that this should be an own PR, as it has the potential to grow out of control).

Well I don't disagree that it should be solved globally, but would rather not assign such a task to _myself_ at this time. I've added the automated tests because you've asked me to, and then tried to handle situation when they failed (by fixing incorrect quest code). 

---

Anyway, let me know how you would prefer this to be resolved. IMHO, the current solution is the best one (I'm obviously biased), but if you want me to revert to old access-handling code and remove/comment-out the tests finding inconsistencies (or some other solution), let me know and I'll try to oblige.

westnordost: I prefer to not have that in this PR then. If this is done, it should be done consistently across all quests where such a filter or a similar filter exists. But I am not at all convinced about either that it is supposedly easier to read, that it covers a relevant case or even that the logic would be correct.  As written in my last post.

mnalis: OK, done as requested (reverted to old access checking code, and commented out the tests that would be failing)

westnordost:
Please do not comment out tests. We have version control. If you want to keep those tests around in case a PR like this is accepted, please put them in some branch.

mnalis:
OK, removed failing tests completely. Ready for review.

westnordost (approved):
Looks good to me.

I prefer if maps with more than 1-2 values are defined like this

```diff
- assertIsNotApplicable("highway" to "path", "bicycle" to "designated", "segregated" to "yes", "cycleway:surface" to "unpaved", "note:cycleway:surface" to "it's complicated")
+ assertIsNotApplicable(
+    "highway" to "path",
+    "bicycle" to "designated",
+    "segregated" to "yes",
+    "cycleway:surface" to "unpaved",
+    "note:cycleway:surface" to "it's complicated"
+)
```
for readability but I won't require you to change this to merge this PR.

The test names are all easy to read and descriptive about what they test.

Especially the part about applicability I find a bit unnecessarily extensive, but now that you've already written it, might as well keep it.

Issue #4501, opened 2022-10-14
Parking lane quest deletes previous data


Mannivu:
I've noticed that the parking lane quest (activated via the dedicated layer) deletes the old tags on the way. See for example [this way](https://www.openstreetmap.org/way/28045002/history) in which SC deleted `parking:condition:both:time_interval`, `parking:condition:both:default` and `parking:condition:both` keys.

westnordost:
This is deliberate. Whenever the parking has been changed, any subtag may have become wrong, so it is removed. I know, it is unfortunate that there is still no quest in StreetComplete with which the `condition` can be (re)surveyed, but in any case it is better to than to delete potentially wrong data then keep it.

E.g. `smoothness` is also removed when the `surface` is changed.

Mannivu:
I understand, but in a case like the one I presented, the  only data missing from the way was the `parking:lane:both:parallel` key, everything else was right. Maybe the app could ask the user to check if the existing data are still up to date (unfortunately, I am not aware if this is possible, so if this can be done, I understand that the deleting of the existing info might be better than leaving potentially right data).

westnordost:
Well yeah, **you** know that but the app doesn't. The situation could just as well be that the street was redesigned.

If the user changes nothing, the app also doesn't delete anything (duh!). Asking if the existing data is still up to date would necessitate that StreetComplete understood these condition-tags and it does not understand them (yet) because there is also no quest for it. Also, the OSM street-parking guy (@SupaplexOSM) started to work on a proposal to straighten up the whole parking schema, so (that) parsing of conditions is in flux, too (thankfully, the current scheme is quite bad):

https://wiki.openstreetmap.org/wiki/Proposed_features/street_parking_revision

westnordost:
Anyway, I am sorry and I acknowledge that this is problematic, but of two evils (deleting too much vs. keeping wrong data) it is IMO the lesser one. As far as I know, @matkoniecz has some half-finished implementation of parking conditions lying around, so maybe there will be something soon to fill that hole.

Mannivu:
Yeah, I thought the app couldn't (yet) handle this kind of check. Don't
worry, I didn't know this was intended, but I understand the thinking
behind this behaviour.

Manuel, ***@***.***

Il ven 14 ott 2022, 17:57 Tobias Zwick ***@***.***> ha
scritto:

Discostu36:
[...]

PR #4549, opened 2022-10-24
Improve cleanup of old parking tags

### LINKED ISSUE #4534: Street parking overlay creates duplicated tags
The street parking overlay seems to tag ways with duplicated information for `:left` or `:right` and `:both`. From what I have seen so far, either the `:right` or `:left` key appears to duplicate the value of the `:both` key. I did not find any inconsistent information between these keys.

I do not know how to reproduce this issue, but I found many examples of this type of tagging on osmose, which reports this issue as "[Too many parking:lane:[side]](https://osmose.openstreetmap.fr/en/issues/open?item=3161&class=31614)". Listed below are some ways that now have duplicated `parking:condition` and/or `parking:lane` tags after they were edited using StreetComplete: 

- https://www.openstreetmap.org/way/1104454759/history
- https://www.openstreetmap.org/way/22943015/history
- https://www.openstreetmap.org/way/1104455061/history
- https://www.openstreetmap.org/way/1105429683/history
- https://www.openstreetmap.org/way/1104422600/history

Most of the changesets that introduced this type of tagging were created using v47.1 and v47.2 of StreetComplete.

tapetis:
Closes #4534.

I have not seen any problems while testing and the new test cases all work as well. However, I am not really sure if the new code covers all possible parking situations and user selections.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParking.kt
@@ -106,8 +106,8 @@ fun LeftAndRightStreetParking.applyTo(tags: Tags) {
      * function. But on the other hand, when the physical layout of the parking changes (=redesign
      * of the street layout and furniture), the condition may very well change too, so better delete
      * it to be on the safe side. (It is better to have no data than to have wrong data.) */
-    val rightSideChanged = right != null && !right.isSupplementingOrEqual(currentParking?.right)
-    val leftSideChanged = left != null && !left.isSupplementingOrEqual(currentParking?.left)
+    val rightSideChanged = !right.isSupplementingOrEqual(currentParking?.right)
+    val leftSideChanged = !left.isSupplementingOrEqual(currentParking?.left)
 
     val keysToRemove = mutableListOf<Regex>()
     if (right == left) {
@@ -115,7 +115,8 @@ fun LeftAndRightStreetParking.applyTo(tags: Tags) {
         if (conditionLeft != null && conditionRight == conditionLeft) {
             keysToRemove.add(Regex("^parking:condition:.*"))
         }
-    } else if (rightSideChanged && leftSideChanged) {
+    }
+    if (rightSideChanged && leftSideChanged) {
         keysToRemove.add(Regex("^parking:(lane|condition):.*"))
     } else if (rightSideChanged) {
         keysToRemove.add(Regex("^parking:(lane|condition):(both|right).*"))
@@ -163,7 +164,7 @@ fun LeftAndRightStreetParking.applyTo(tags: Tags) {
     }
 }
 
-private fun StreetParking.isSupplementingOrEqual(other: StreetParking?): Boolean =
+private fun StreetParking?.isSupplementingOrEqual(other: StreetParking?): Boolean =
     this == other || (
     this is StreetParkingPositionAndOrientation &&
         other is StreetParkingPositionAndOrientation &&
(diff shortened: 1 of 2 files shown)


westnordost:
Very nice, you even added all those new test cases! Will review asap

app/src/test/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingKtTest.kt
```diff
@@ -160,7 +179,125 @@ class StreetParkingTest {
...
+            LeftAndRightStreetParking(
+                StreetParkingPositionAndOrientation(ParkingOrientation.PARALLEL, ParkingPosition.ON_STREET),
+                StreetParkingPositionAndOrientation(ParkingOrientation.PARALLEL, ParkingPosition.ON_STREET)
+            ),
+            arrayOf(
+                StringMapEntryModify("parking:lane:both", "parallel", "parallel"),
+                StringMapEntryAdd("parking:lane:both:parallel", "on_street"),
+                StringMapEntryDelete("parking:lane:left:parallel", "on_street"),
+                StringMapEntryDelete("parking:lane:right:parallel", "on_kerb"),
+                StringMapEntryDelete("parking:condition:right", "customers")
```
westnordost, line 295: So, here I would expect the `parking:condition:right` to be deleted because the right side changed from 
parallel on-kerb parking to parallel on-street parking. (I added that line to the test).

Or am I missing something?

app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParking.kt
```diff
@@ -81,19 +81,32 @@ private val ParkingPosition.estimatedWidthOnRoadFactor: Float get() = when (this
...
+    // parking:condition:<left/right/both>
+    val conditionRight = right?.toOsmConditionValue()
+    val conditionLeft = left?.toOsmConditionValue()
+
+    /* Clean up previous parking tags when the associated side is changed. This can include removing
+     * parking:condition:*, which is a bit peculiar because most values are not even set in this
+     * function. But on the other hand, when the physical layout of the parking changes (=redesign
+     * of the street layout and furniture), the condition may very well change too, so better delete
+     * it to be on the safe side. (It is better to have no data than to have wrong data.) */
+    val keysToRemove = buildList {
```
westnordost, line 93: In general I like to avoid such Kotlin constructs that implicitly change what `this` is because it adds a layer of confusion and a source of error (you need the IDE to help you display what `this` is or know the documentation by heart)

app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParking.kt
```diff
@@ -81,19 +81,32 @@ private val ParkingPosition.estimatedWidthOnRoadFactor: Float get() = when (this
...
+        if (right == left) {
+            add(Regex("^parking:lane:.*"))
+            if (conditionLeft != null && conditionRight == conditionLeft) {
+                add(Regex("^parking:condition:.*"))
             }
```
westnordost, line 98: Huh? Everything is removed when left and right are the same? And something with conditions?

I don't quite understand this, shouldn't this if-else-block be about checking whether each side **changed** only?

westnordost:
I added more granular parsing to enable implementation of the suggestions made here https://github.com/streetcomplete/StreetComplete/issues/4501#issuecomment-1290420809

They write that StreetComplete should also not delete parking tags for the side that is changed **if** it is merely supplemented to be complete tagging rather than changing anything.

westnordost:
I'm currently looking into the code for that and maybe changing something.

westnordost:
> Is it even possible to delete an already existing side using the app?

No

westnordost:
Hum, I spotted some other issue while adding some tests but I forgot what it was. It was probably some edge case to the "do clear the tags if something changed". But anyway, I guess I should revert/redo some of all this to account for https://github.com/streetcomplete/StreetComplete/pull/4549#issuecomment-1290757629 ... or at least think a bit more about this first.

westnordost:
Hmm, I thought about it some more and I think it's better to not clear the whole `parking:*` space but only replace/remove the tags set by that quest. I'll work on this now.

westnordost:
But this PR now contains so much stuff for the conditional removal of subtags that I better start fresh and only selectively copy over some things from this PR

PR #4552, opened 2022-10-25
Fix parking tags2

westnordost:
Closes https://github.com/streetcomplete/StreetComplete/issues/4534.
Supercedes  #4549..

@tapetis , do you want to have a look? I ended up with having none of the original PR in there, because all of it was not relevant anymore.

I also removed parking condition stuff from StreetParking (parsing) and the UI altogether, which made the whole thing somewhat simpler. Which sign exactly forbids parking, standing, stopping is something that could be recorded in another overlay.


app/src/main/res/raw/changelog.yml
```diff
@@ -28,6 +28,7 @@ v48.0: |
   <li>Enable quest that asks if cash is accepted in United Kingdom (#4517)</li>
   <li>Clarify some wordings for buildings (#4431), by @matkoniecz</li>
   <li>Do not ask for parking access if some access is already tagged (#4538, #4547), by @mnalis</li>
+  <li>Parking overlay: Simplify selecting that that there is no parking (#4534)</li>
```
FloEdelmann, line 31:
```suggestion
  <li>Parking overlay: Simplify selecting that there is no parking (#4534)</li>
```

app/src/main/res/raw/changelog.yml
```diff
@@ -36,6 +37,7 @@ v48.0: |
   <li>Add links to documentation for tags changed as shown in undo dialog (#3442, #4419), by @matkoniecz</li>
   <li>The obstrusive downloading-view is gone, downloading and uploading is now shown in the upper left corner</li>
   <li>Resurvey quests no longer vanish after splitting the way - only works if auto-sync is off (#3567, #4523), by @Helium314</li>
+  <li>Fix the parking overlay would unnecessarily remove some tags and not remove others it should remove when changing the answer (#4501, #4552), thanks @tapetis</li>
```
FloEdelmann, line 40:
```suggestion
  <li>Fix which tags the parking overlay removes when changing the answer (#4501, #4552), thanks @tapetis</li>
```

westnordost:
> Does the app clean up the serialized parking selection when it is updated via Google Play?

Hm no it doesn't. Do you have an idea how to solve this with little code?

westnordost:
Then I'll just wrap this in a try catch

PR #4579, opened 2022-10-31
Add move node action

Helium314:
a start for fixing #3502

Moving a node works, and geometries of elements containing the node are updated and quest geometries are displayed correctly.
I added a "move node" other answer, which is enabled only for nodes not part of a way/relation (to avoid dealing with https://github.com/streetcomplete/StreetComplete/issues/3502#issuecomment-967127770).

There is still some work to be done:
* decide where to enable the move node answer (really all quests with free floating nodes?)
* add some text explaining the users why they can't move all element, as being able to move only some might be confusing
* proper UI instead of abusing the create note layout (see comments in `MoveNodeFragment.kt`)
* unit tests
* try uploading a move node action, and reverting it

If anyone wants to help (especially with the UI), feel free to do so

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -1,5 +1,6 @@
 package de.westnordost.streetcomplete.data.osm.edits
 
+import de.westnordost.streetcomplete.data.osm.edits.move.MoveNodeAction
 import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
 import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometryCreator
 import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometryEntry
@@ -427,7 +428,21 @@ class MapDataWithEditsSource internal constructor(
             updatedElements[key] = element
             updatedGeometries[key] = createGeometry(element)
         }
-        return MapDataUpdates(updated = updates, deleted = deletedKeys)
+
+        // get affected ways and relations and update geometries if a node was moved
+        val elementsWithChangedGeometry = mutableListOf<Element>()
+        if (edit.action is MoveNodeAction) {
+            val waysContainingNode = getWaysForNode(edit.elementId)
+            val affectedRelations = getRelationsForNode(edit.elementId) + waysContainingNode.flatMap { getRelationsForWay(it.id) }
+            for (element in waysContainingNode + affectedRelations) {
+                val key = ElementKey(element.type, element.id)
+                deletedElements.remove(key)
+                updatedElements[key] = element
+                updatedGeometries[key] = createGeometry(element)
+                elementsWithChangedGeometry.add(element) // questController needs to know that element was affected (though it was not updated in OSM sense)
+            }
+        }
+        return MapDataUpdates(updated = updates + elementsWithChangedGeometry, deleted = deletedKeys)
     }
 
     private fun createGeometry(element: Element): ElementGeometry? {
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
@@ -1,0 +1,146 @@
+package de.westnordost.streetcomplete.screens.main.bottom_sheet
+
+import android.content.res.Configuration
+import android.graphics.Point
+import android.os.Bundle
+import android.view.View
+import androidx.annotation.UiThread
+import androidx.appcompat.app.AlertDialog
+import androidx.core.os.bundleOf
+import androidx.fragment.app.Fragment
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.edits.ElementEditType
+import de.westnordost.streetcomplete.data.osm.edits.ElementEditsController
+import de.westnordost.streetcomplete.data.osm.edits.move.MoveNodeAction
+import de.westnordost.streetcomplete.data.osm.geometry.ElementPointGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.ElementKey
+import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
+import de.westnordost.streetcomplete.data.osm.mapdata.Node
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
+import de.westnordost.streetcomplete.data.overlays.OverlayRegistry
+import de.westnordost.streetcomplete.data.quest.QuestTypeRegistry
+import de.westnordost.streetcomplete.databinding.FragmentCreateNoteBinding
+import de.westnordost.streetcomplete.overlays.IsShowingElement
+import de.westnordost.streetcomplete.util.ktx.getLocationInWindow
+import de.westnordost.streetcomplete.util.ktx.popIn
+import de.westnordost.streetcomplete.util.ktx.viewLifecycleScope
+import de.westnordost.streetcomplete.util.math.distanceTo
+import de.westnordost.streetcomplete.util.viewBinding
+import kotlinx.coroutines.launch
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import org.koin.android.ext.android.inject
+
+/** Fragment that lets the user move an OSM node */
+class MoveNodeFragment :
+    Fragment(R.layout.fragment_create_note), IsCloseableBottomSheet, IsShowingElement {
+
+    private val binding by viewBinding(FragmentCreateNoteBinding::bind)
+    private val bottomSheetBinding get() = binding.questAnswerLayout
+    private val okButton get() = bottomSheetBinding.okButton
+    private val okButtonContainer get() = bottomSheetBinding.okButtonContainer
+
+    private val elementEditsController: ElementEditsController by inject()
+    private val questTypeRegistry: QuestTypeRegistry by inject()
+    private val overlayRegistry: OverlayRegistry by inject()
+
+    override val elementKey: ElementKey by lazy { ElementKey(node.type, node.id) }
+
+    private lateinit var node: Node
+    private lateinit var editType: ElementEditType
+
+    private val hasChanges get() = getPosition() != node.position
+    private val isFormComplete get() = hasChanges //&& getPosition()?.distanceTo(node.position)?.let { it > 2.0 } ?: false // this is ugly
+
+    interface Listener {
+        fun getMapPositionAt(screenPos: Point): LatLon?
+
+        fun onMovedNode(editType: ElementEditType, position: LatLon)
+    }
+    private val listener: Listener? get() = parentFragment as? Listener ?: activity as? Listener
+
+    private fun getPosition(): LatLon? {
+        val createNoteMarker = binding.markerCreateLayout.createNoteMarker
+        val screenPos = createNoteMarker.getLocationInWindow()
+        screenPos.offset(createNoteMarker.width / 2, createNoteMarker.height / 2)
+        return listener?.getMapPositionAt(screenPos)
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val args = requireArguments()
+        node = Json.decodeFromString(args.getString(ARG_NODE)!!)
+        editType = questTypeRegistry.getByName(args.getString(ARG_QUESTTYPE)!!) as? OsmElementQuestType<*>
+            ?: overlayRegistry.getByName(args.getString(ARG_QUESTTYPE)!!)!!
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        okButton.setOnClickListener { onClickOk() }
+        okButtonContainer.popIn()
+        okButton.popIn()
+    }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {
+        super.onConfigurationChanged(newConfig)
+        binding.markerCreateLayout.centeredMarkerLayout.setPadding(
+            resources.getDimensionPixelSize(R.dimen.quest_form_leftOffset),
+            resources.getDimensionPixelSize(R.dimen.quest_form_topOffset),
+            resources.getDimensionPixelSize(R.dimen.quest_form_rightOffset),
+            resources.getDimensionPixelSize(R.dimen.quest_form_bottomOffset)
+        )
+    }
+
+    private fun onClickOk() {
+        val pos = getPosition() ?: return
+        viewLifecycleScope.launch {
+            val action = MoveNodeAction(pos)
+            elementEditsController.add(editType, node, ElementPointGeometry(node.position), "survey", action)
+            listener?.onMovedNode(editType, pos)
+        }
+    }
+
+    override fun onClickMapAt(position: LatLon, clickAreaSizeInMeters: Double): Boolean {
+        // todo: maybe move map to this position? what would a "normal" person expect?
+        return false
+    }
+
+    // todo:
+    //  nicer view with that crosshair and distance?
+    //   don't use the note layout!
+    //  hide ok button if not moved, or moved very far -> see MainFragment.onMapIsChanging
+    //  confirm large move
+    //  confirm very small move?
+    //  switch to satellite view??
+
+    @UiThread override fun onClickClose(onConfirmed: () -> Unit) {
+        if (!hasChanges) {
+            onConfirmed()
+        } else {
+            activity?.let {
+                AlertDialog.Builder(it)
+                    .setMessage(R.string.confirmation_discard_title)
+                    .setPositiveButton(R.string.confirmation_discard_positive) { _, _ ->
+                        onConfirmed()
+                    }
+                    .setNegativeButton(R.string.short_no_answer_on_button, null)
+                    .show()
+            }
+        }
+    }
+
+    companion object {
+        private const val ARG_NODE = "node"
+        private const val ARG_QUESTTYPE = "quest_type"
+
+        fun create(elementEditType: ElementEditType, node: Node): MoveNodeFragment {
+            val f = MoveNodeFragment()
+            f.arguments = bundleOf(
+                ARG_NODE to Json.encodeToString(node),
+                ARG_QUESTTYPE to elementEditType.name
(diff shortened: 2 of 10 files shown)


westnordost:
Yeahh, nice!!

> decide where to enable the move node answer (really all quests with free floating nodes?)

Nothing comes to mind why not. I.e. should be decided/treated similarily as for the split way.

> add some text explaining the users why they can't move all element, as being able to move only some might be confusing

Ideally, the option should just not be shown in that case.

> proper UI instead of abusing the create note layout (see comments in MoveNodeFragment.kt)

I'd really really like to see a UI like this: https://github.com/streetcomplete/StreetComplete/issues/3502#issuecomment-967151011 I 

think you cannot draw arrow-shapes in tangram-es, but that's not necessary. See the pointer pin, it is also a normal view. Or actually, it can be just a custom Drawable, just like e.g. the LaurelWreathDrawable, the StreetParkingDrawable, the DatesActiveDrawable and the lot. You have an ImageView with layout width and height to `match_parent` and assign it that TBD `ArrowDrawable`.
On every screen move, you need to pass it the screen positions of the start and end geo position of the node-moving and the drawable will rerender itself in the image view. The `ArrowDrawable` will really be just a view lines of code.

The display at the bottom for the moved distance will be equally easy. You can/should use the same layout as for the length-display in the measuring activity. Just use `fun LatLon.distanceTo` to get the distance in meters. Best round it to 10cm or so. (Also, in the US etc. , it should probably be Foot+Inch. Just have a look at the Measure Activity.)

Helium314:
> > decide where to enable the move node answer (really all quests with free floating nodes?)
> 
> Nothing comes to mind why not. I.e. should be decided/treated similarily as for the split way.

Maybe nodes that are member of a non-multipolygon relation could also be allowed? After all, nodes that are member of a relation with type in `IGNORED_RELATION_TYPES` will be eligible for move too.

Helium314:
After some improvements now I think it's ready for review.

What is still left:
* arrow icon (currently it's using the quest pin)
* the speech bubbles in there really don't look suitable, but I have no idea what to use here
* "explanatory text" at the bottom is basically a placeholder
* inform user if the node was moved too far? I tried that, but spamming toasts while the camera is moving is really bad for performance...

westnordost:
Should this be reviewed or are you still working on it? What about the "ArrowDrawable" as depicted here? https://github.com/streetcomplete/StreetComplete/issues/3502#issuecomment-967151011

Helium314:
I think it's fine now.
If you want I can replace the pin with an arrow, though I think the pin is ok as well, and people are used to it from positioning notes.

westnordost:
I'd very much like the UI as I proposed, so that the user knows from where to where he is moving the node and how many meters he is moving the node.

Helium314:
The user already knows from where to where the place is moved, and how mayn meters. My question was only about arrow or pin:
Screenshot_20221111-175514_Street�Complete_Dev

westnordost:
This UI looks as if there is **another** water source (at the previous position), as this kind of highlighting is used in quests only to highlight exactly that: The location of other benches, other shops, other water sources.

I do think that it makes sense to show other objects of the same kind on the map for obvious reasons. But this also means that this UI is misunderstandable.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -184,6 +186,15 @@ class MapDataWithEditsSource internal constructor(
...
                     }
+
+                    if (edit.action is MoveNodeAction || edit.action is RevertMoveNodeAction) {
+                        val waysContainingNode = getWaysForNode(edit.elementId)
+                        val affectedRelations = getRelationsForNode(edit.elementId) + waysContainingNode.flatMap { getRelationsForWay(it.id) }
+                        for (elem in waysContainingNode + affectedRelations) {
+                            mapData.put(elem, getGeometry(elem.type, elem.id))
+                        }
+                    }
+
```
westnordost, line 197: What is this for? I thought MoveNodeAction would only work for free-floating nodes?

Also, in any case, unit tests would be very useful here.

westnordost (commented):
A preliminary review, i.e. everything except the UI

app/src/main/res/drawable/ic_undo_move_node.xml
```diff
@@ -0,0 +1,21 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
Helium314: It's basically the arrow from `ic_slim_arrow_up_16dp.xml`

app/src/main/java/de/westnordost/streetcomplete/overlays/AbstractOverlayForm.kt
```diff
@@ -347,6 +352,13 @@ abstract class AbstractOverlayForm :
             if (element.isSplittable()) {
                 answers.add(AnswerItem(R.string.split_way) { splitWay(element) })
             }
+
+            if (element is Node // add moveNodeAnswer only if it's a free floating node
+                && mapDataWithEditsSource.getWaysForNode(element.id).isEmpty()
+                && mapDataWithEditsSource.getRelationsForNode(element.id).isEmpty()) { // todo: allow relations except multipolygon?
+                answers.add(AnswerItem(R.string.move_node) { moveNode() })
```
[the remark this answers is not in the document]
Helium314: I don't think this is worth it.
For overlays it's irrelevant as the visible area is cached.

I just checked for some differences when opening quests on a node for first and second time after starting SC.
I didn't notice any difference, independent on whether the quest is showing highlighted elements, and I claim the vast majority of SC users are using a faster phone than a S4 Mini and thus would even notice less.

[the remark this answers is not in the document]
Helium314: Though the check could already be started async inside `onCreate`...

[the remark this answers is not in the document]
Helium314: I was thinking about that, but decided to use the same as you wanted to avoid using the word _node_, and the only alternative that came in my mind was _point_, which imo doesn't sound right.

And maybe "At slightly wrong _position_" should be changed to "At slightly wrong _location_", as this is also used when clicking _can't say_.

app/src/main/res/drawable/ic_undo_move_node.xml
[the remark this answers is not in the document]
Helium314: So should I create a svg from the image and claim it's the source?

app/src/main/res/values/strings.xml
```diff
@@ -147,6 +147,13 @@ The info you enter is directly added to OpenStreetMap in your name, without the
     <string name="quest_split_way_many_splits_confirmation_description">These are quite a few splits. You can always further split up the way later.</string>
     <string name="split_way">Split way</string>
 
+    <string name="move_node">At slightly wrong position</string>
+    <string name="quest_move_node_message">If it is at a wrong position, you can move this feature.\nMove it now?</string>
+    <string name="node_moved_not_far_enough">Please move it further</string>
+    <string name="node_moved_too_far">Moved too far. Consider leaving a note</string>
+    <string name="node_moved">Moved %s</string>
+    <string name="move_node_description">Drag the map to place the feature at the correct position. Use the zoom function and try to be as precise as possible. Please be aware that your displayed location and the map may not be exact.</string>
```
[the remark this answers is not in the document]
Helium314: I was thinking of @mnalis [comment](https://github.com/streetcomplete/StreetComplete/issues/3502#issuecomment-966807247), my own experiences of frequently being far outside the accuracy circle, and of the lack of orientation of some users (observed from notes with attached images).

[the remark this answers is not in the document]
Helium314: But I agree that the part about the not exact map is not helping.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
Helium314: It's used only for free floating nodes, but the action works also for nodes that are part of a way or relation.
And I think allowing some (non-multipolygon) relations might actually make sense, e.g. for moving a bus stop.

app/src/main/res/drawable/ic_undo_move_node.xml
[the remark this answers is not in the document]
westnordost: Yeah, and maybe rotate it a bit. An "arrow up" carries an additional meaning that may be confusing. An arrow just pointing wherever doesn't have this meaning.

app/src/main/java/de/westnordost/streetcomplete/overlays/AbstractOverlayForm.kt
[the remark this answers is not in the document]
westnordost: Ok

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
westnordost: Oh okay, that is quite foresighted of you. But then, please also add tests that confirm this behavior.

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+                MeasureDisplayUnitMeter(2)
+        binding.createNoteIconView.setImageResource(editType.icon)
+        highlightSimilarElements()
+    }
+
+    private fun highlightSimilarElements() {
+        val feature = featureDictionaryFuture.get().byTags(node.tags).isSuggestion(false).find().firstOrNull()
+        val tagsToFind = feature?.tags ?: node.tags
+        val mapData = mapDataWithEditsSource
+            .getMapDataWithGeometry(node.position.enclosingBoundingBox(MAX_MOVE_DISTANCE + 5.0))
```
westnordost, line 113: use `viewLifecycleScope.launch { ... }` with `withContext(Dispatchers.IO)`.

See also `MainFragment::showHighlightedElements`, try to mimic this behavior.

Actually, better yet, ... while this solution with using the feature dictionary is clever, it is inconsistent to the highlighting done while the quest is displayed. E.g. the "bench backrest" quest shows other benches **and** picnic tables. Now if a user taps "at slightly different position", only other benches are shown. Its even much worse for any shop-related quests, because the `MoveNodeFragment` would only show exactly the same shop type as the user is moving. This **will** 100% lead to confusion.

I believe it would be possible with only little refactoring to just let `MainFragment` show the highlighted elements of in whatever quest's context this fragment was started with. After all, the quest type is passed to this fragment. All `MainFragment::showHighlightedElements` needs is a position and a `getHighlightedElements` of a quest type currently.

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+                ARG_NODE to Json.encodeToString(node),
+                ARG_QUEST_TYPE to elementEditType.name
+            )
+            return f
+        }
+    }
+}
+
+private const val MIN_MOVE_DISTANCE = 2.0
+private const val MAX_MOVE_DISTANCE = 30.0
```
westnordost, line 200: Can you add a rationale (in the comment) why a minimum move distance makes sense and why it is at 2 meters? (The same for max move distance if applicable)

westnordost (changes requested):
Hmm, the icon looks like the icon for "resize" or "measure". But anyway, I may later change it, no blocker. Also, I would really like that arrow. But no worries, I will do it myself after merge then.

The code looks from start to end really well. The only issue right now I have with this PR is the inconsistency in highlighting the "related elements" with the quests.

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
Helium314: Well, then the simplest solution would be to use `showInBottomSheet(MoveNodeFragment.create(editType, node), clearPreviousHighlighting = false)`, and hide overlay only if `editType !is Overlay`.
Only (potential) issue here is that the quest pin remains at the original position,  though indicating starting position is not necessarily bad.

Helium314:
>  But anyway, I may later change it, no blocker. Also, I would really like that arrow. But no worries, I will do it myself after merge then.

Thanks, I would prefer if you did those changes, because I'm really slow when it comes to vector graphics / Inkscape /...

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
Helium314: I'm not sure about the actual values, that's rather some guess how much makes sense. I added the reasons for limiting the distance.

westnordost:
> Thanks, I would prefer if you did those changes, because I'm really slow when it comes to vector graphics / Inkscape /...

Hm? Perhaps a misunderstanding. https://github.com/streetcomplete/StreetComplete/issues/3502#issuecomment-967151011 would not involve creating vector graphics at all. It would involve creating a custom `Drawable` class that draws itself on a canvas.

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+}
+
+// Require a minimum distance for two reasons:
+// 1. The map is not perfectly precise, especially displayed road widths may be off by a few meters,
+//     so it may be hard to tell whether something really is misplaced (e.g. bench along a path)
+//     without good aerial imagery.
+// 2. The value added by moving nodes by such small distance, even if correct, is rather low.
+private const val MIN_MOVE_DISTANCE = 2.0
```
westnordost, line 204: A third good reason would be:

What is the size of a (e.g. shop) icon shown at maximum zoom in meters? Basically, the larger the icon, the less sense it makes to move it by some centimeters. So if a hydrant icon is displayed at e.g. 2x2 meters (I don't know), the position imprecision is already about 1m because it is not really possible for the user to ascertain exactly what is the center of the icon. The icons are not shown on a pin after all, like they are in iD
grafik

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
```diff
@@ -0,0 +1,207 @@
...
+// Move node functionality is meant for fixing slightly misplaced elements. If something moved far
+// away, it is reasonable to assume there are more substantial changes required, also to nearby elements.
+private const val MAX_MOVE_DISTANCE = 30.0
```
westnordost, line 207: Another reason: the related elements are only shown in a certain radius. The move distance may not be larger than that radius.

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
westnordost: Hm, I didn't think of that. This might be the easiest solution. I can't think of any downside.

And original quest pin could just be hidden manually in the function that shows the MoveNodeFragment, no?

Helium314:
> Hm? Perhaps a misunderstanding. [#3502 (comment)](https://github.com/streetcomplete/StreetComplete/issues/3502#issuecomment-967151011) would not involve creating vector graphics at all. It would involve creating a custom `Drawable` class that draws itself on a canvas.

But a drawable _is_ a vector graphic? (I don't think raster graphics would be acceptable here)

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
Helium314: I was thinking about that, but this probably depends on the device. For tablet the size might be much different from a phone, and even between phones this could be different.
On my phone I would estimate shop icons are a bit larger than 2x2 m, but not by much.

westnordost:
No, a `Drawable` is an abstract class in Android. It is implemented by `VectorDrawable`, but you can create your own drawables as you wish. See for example `LaurelWreathDrawable` or `DatesActiveDrawable`

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
westnordost: Why would this depend on the device? The icons on the map are scaled, too. I.e. on devices with a low DPI, the icons should appear at the same size as on devices with a high DPI.

Helium314:
Ah ok, I was confusing this with the stuff that's in drawable folder.

westnordost:
It's actually quite fun to create custom drawables, because it is so little code for so much effect. Also custom drawables can be animated really easily too (`LaurelWreathDrawable` is animated)

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
Helium314: I wasn't aware the map scales together with the icons

Helium314: So the size is the same everywhere

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/MoveNodeFragment.kt
Helium314: Adjusted, so this is now happening.

westnordost (approved):

westnordost:
Mh, I noticed only now that one item on your todo list hasn't been crossed out. Did you test that in the meantime?

Helium314:
I did some uploads, but never tried reverting an uploaded move node action

Helium314:
Now also reverted a move node action, works as it should.

peternewman:
I was going to suggest bus stops should be excluded from the relation, as they're often part of a route relation, but I see they've also got a `public_transport=stop_position` which also wants moving, so I guess they'll sadly have to be left excluded. They probably also often move more than 30m too.

PR #4585, opened 2022-10-31
Attempt to add tutorial for overlays

matkoniecz:
Note: graphics are not done at all

Note: I am not entirely sure whether technical display/hiding of tutorial screens is done properly or a horrific hack


westnordost:
> Note: graphics are not done at all

Should be similar to the main tutorial screen. If you are not sure what could be shown, try making some mockups  (in paint or whatever) and this can be discussed.

> Note: I am not entirely sure whether technical display/hiding of tutorial screens is done properly or a horrific hac

Why plural? Do you plan to have a tutorial for each overlay? Otherwise, it should just work the same as the initial tutorial screen.

westnordost:
Otherwise, just post here if you need feedback or help. But remember, I am not the only one who can give (technical) feedback.

matkoniecz:
> Should be similar to the main tutorial screen. If you are not sure what could be shown, try making some mockups (in paint or whatever) and this can be discussed.

Yes, I have some plans but wanted to win with technical side first.

> Why plural? Do you plan to have a tutorial for each overlay?

No, but this single tutorial has more than one screen

> Otherwise, it should just work the same as the initial tutorial screen.

I started with using the same technical solution, but that was showing below overlay listing

westnordost:
>> Otherwise, it should just work the same as the initial tutorial screen.

> I started with using the same technical solution, but that was showing below overlay listing

Uhm well, just show the tutorial instead of the overlay listing (or close it) on first click on the overlays button? 

Also,  the overlay selection needs to be redone anyway with the number of overlays growing some time.

matkoniecz:
> Uhm well, just show the tutorial instead of the overlay listing (or close it) on first click on the overlays button?

That is how it is implemented right now.

westnordost:
So, this is marked as draft. Do you need help or do you want a code review? If yes, about what / which part?

matkoniecz:
> do you want a code review

Yes, code review - about code itself except `step1Transition` `step2Transition` as graphics are not changed from intro tutorial.

Including proposed content to be shown as text in `strings.xml` (ideally it would get stable and reviewed before working on graphics illustrating it)

matkoniecz:
@FloEdelmann If you would have interest and time - I would appreciate review/nitpicking/complaints about this code

(except `step1Transition` `step2Transition` functions as graphics are not changed from intro tutorial)

FloEdelmann (commented):
Only wording, the code looks fine! Note that I didn't run the code.

westnordost:
For the graphics / animations, an idea would be:

1. show a part of a map as a vector graphic (i.e. recreate some section of made-up map in tangram-es colors as a vector graphic using Inkscape or something)

2. Animate in coloring of the same vector graphic as it would look like in a made-up overlay on top of the previous graphic. For animation, "fade-in" would be easiest but not fancy. Fancy and not really much more difficult would be to have a graphic of a paintbrush or paint roller move over the area from top-left to bottom-right and at the same time move the mask (`clip-path`) of the overlay-graphic in the same direction.  
  I.e. the latter graphic would be an [animated vector drawable](https://developer.android.com/develop/ui/views/animations/drawable-animation) just like [some other graphics in this project](https://github.com/streetcomplete/StreetComplete/search?q=animated-vector), in particular, the mail.

3. For the 3rd slide, my only idea is to show the tools being mentioned float up from the bottom or being held by a hand like a deck of cards.

app/src/main/res/values/strings.xml
```diff
@@ -1610,7 +1610,7 @@ Partially means that a wheelchair can enter and use the restroom, but no handrai
     <!-- Overlays tutorial screen -->
     <string name="tutorial_overlays_welcome">Congratulations!</string>
     <string name="tutorial_overlays_welcome_subtitle">you found a new way to edit</string>
-    <string name="tutorial_overlays_intro">You can now view data, spot wrong entries and fix them without waiting for a quest. Overlays are showing a full overview, not just a few objects where survey is obviously needed.</string>
```
Victino1ef207e: `merge`

westnordost:
With the next version moving the overlays button to a somewhat more discoverable location, a tutorial would be quite useful. Did you abandon this?

westnordost:
This is what I'd imagine it. 

It is shown on first tap on the overlays button.

Three steps:

### 1.

[[a blank osm map is shown]]

Neat!

You found a new way to edit data: Overlays.
 
### 2.

[[a paint-roller moves over the map and leaves all roads colored in different colors]]

Each overlay displays one kind of data, for example shops or street surfaces.

The data is highlighted in different colors to give you an overview over what has been recorded so far.

Missing data is always shown as red.

### 3.

[[one highlighted way is visually selected, flashes and its color is changed]]

Tap on any highlighted data to edit it. You can also split ways or report notes, as with quests.

Additionally, some overlays allow you to also add new data using the crosshair and the big plus button, such as the address overlay.

PR #4600, opened 2022-11-02
Fix pivot point in StreetParkingDrawable

### LINKED ISSUE #4590: Regression: thumbnails not shown in street parking overlay
**How to Reproduce**
Attempt to add parking in the parking overlay, and find that the thumbnails are not showing for many of the options as they have in earlier versions.

image
image

**Expected Behavior**
Thumbnails should be shown so type of parking can be selected easily.

**Versions affected**
V90.0-beta 1 on Android 11

tapetis:
Closes #4590 (sorry).

Before my changes in 99be3a9, the car illustrations were shown correctly, but some cars were clipped when editing in a country with left-handed traffic:

After my changes in 99be3a9, the pivot point for the `Canvas` rotation / scaling was incorrect and the illustrations were not shown correctly:

With this PR both issues are fixed:

I tested my changes both with left-handed and right-handed traffic and did not notice any new issues.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingDrawable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingDrawable.kt
@@ -44,10 +44,13 @@ class StreetParkingDrawable(
     override fun draw(canvas: Canvas) {
         if (!isVisible) return
 
+        if (isUpsideDown) {
+            val pivotY = bounds.height() / ceil(height / width / 2f)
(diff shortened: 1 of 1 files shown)


westnordost:
Noice!

So I understand there was a bug in the first place but it didn't show because the drawable always painted itself 2x as many times as necessary and you solved the root issue, thank you!

westnordost:
Since the drawable is also shown in the street side select puzzle: Did you test if it shows correctly both in left-hand and right-hand driving countries there as well?

PR #4602, opened 2022-11-02
serialize street side values, not display items. Decide about its app…

### LINKED ISSUE #4591: No Bike lane shows DO NOT ENTER icon instead
I picked no bike lane on a way. Then on another way I picked preset in the lower left corner and instead it showed me the icon of "do not enter" sign. Ultimately the tags that got applied to that way were correct, so this appears to be just a visual bug local to the client.

Screenshot_20221030-214519_StreetComplete
Screenshot_20221030-214524_StreetComplete

**How to Reproduce**
Apply "no bike lane" on a way, use the same preset on another way.

**Expected Behavior**
Correct icon shows

**Versions affected**
v49-beta1

westnordost:
…earance only at the moment where it should be displayed/applied (fixes #4591)


Helium314:
This fixes the issue, but I think the cause might still be a bug (possibly introduced in #4524):
The value stored in the settings file is `<string name="imageListLastPicked.AddCyclewayForm.oneSide">NONE</string>`, which doesn't say anything about oneway. But the UI ends up showing the one way sign, so I think there might just be a mixup between left and right somewhere...

Anyway, this is removing the left/right in deserialize, which makes sense to me.

westnordost:
Hum... you are right... what could be done about that....

westnordost:
The core issue is that the left/right side of what is saved for the "last answer" button is applied according to the street orientation on the screen. So, what's left on the button could be applied to the right side of the road and the other way round, effectively switching sides of what the user answered last.
This itself is not an issue, but mixed with values like `NONE_NO_ONEWAY` which depend on the context they have been answered in (it has to be a oneway, it has to be the side of the oneway that is in contraflow), it becomes one.

However, pondering this for a while, I think this is a very minor caveat and can be mitigated by displaying  `NONE_NO_ONEWAY` the as `NONE` if it is not in contraflow.

PR #4615, opened 2022-11-06
Update surface to generic paved if footway and cycleway parts are different but paved

mnalis:
This improves https://github.com/streetcomplete/StreetComplete/issues/4526 (originally closed by PR https://github.com/streetcomplete/StreetComplete/pull/4548):

- in situation when cycleway and footway are _differently paved_ it will also update `surface = paved` (previously, it would just always delete `surface` tag in such case).
- actually removes `smoothness` when `surface` changes (we did test for it, but never actually did it)
- some more fixes for compilation error for tests that were present in vanilla `v49.0` (introduced when different branches were merged)

(This is recommended way to tag different but `paved` surfaces, see second block in `Ref`= `S5` on https://wiki.openstreetmap.org/wiki/Bicycle#Miscellaneous)

---

[Compiles OK](https://github.com/mnalis/StreetComplete/actions/runs/3406512293) and manually tested that it works fine. It also includes automated test that [pass too](https://github.com/mnalis/StreetComplete/actions/runs/3406460771).


mnalis:
Oh sorry, you're the only one the github shows by default, so it seemed inviting button to press :smiley: 
I now see I can press the cog icon to request others too. Everyone is welcome to review of course, but if clicking that button is not required for processing of PR to proceed, than I won't be clicking it in the future. :heart: 

(That request-review functionality is then probably less useful in opensource projects, and more useful in corporate environment with strict hierarchy when someone in power is selecting who should do what, I guess?)

app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
```diff
@@ -40,9 +41,18 @@ class AddCyclewayPartSurface : OsmFilterQuestType<SurfaceAnswer>() {
...
-                tags["surface"] = tags["cycleway:surface"]!!
+                if (tags["surface"] != tags["cycleway:surface"]!!) {
+                    tags["surface"] = tags["cycleway:surface"]!!
+                    tags.remove("smoothness")
+                }
+            } else if (tags["footway:surface"] in ANYTHING_FULLY_PAVED && tags["cycleway:surface"] in ANYTHING_FULLY_PAVED) {
+                if (tags["surface"] != "paved") {
+                    tags["surface"] = "paved"
+                    tags.remove("smoothness")
+                }
```
westnordost, line 52: Well, I don't know about that...

1. As far as I remember, StreetComplete asks for the surface of all roads with `surface=paved` if not also a note is supplied. We do not really want that here, though

3. You seem to forget that there are categorizations below `paved` and `unpaved` possible. E.g. for `footway:surface=unhewn_cobblestone` , `cycleway:surface=sett` , `surface=cobblestone`, it would lead to `surface` being retagged to `paved`.

mnalis: > Well, I don't know about that...

I've decided to go that additional step, because there is [significant number of situations](https://overpass-turbo.eu/s/1nFA) (500 ways just for one city) where a combination of `asphalt` + `paving_stones` is popular for such segregated foot+bicycle ways, so removing `surface` _completely_ seems too harsh.   

> As far as I remember, StreetComplete asks for the surface of all roads with `surface=paved` if not also a note is supplied

- I can change `AddPathSurface.kt` to skip quest on `surface=paved` if both `footway:surface` and `cycleway:surface` are present (`AddRoadSurface.kt` and `AddPitchSurface.kt` which also check that are not relevant here)

- Alternatively,  I can add a `surface:note=different footway and cycleway surfaces` when I set `surface=paved` in this PR

Which one would you prefer?

> You seem to forget that there are categorizations below paved and unpaved possible

They are, but _**very few**_. For example, for `cobblestone`, there is [10 cases wordwide](https://overpass-turbo.eu/s/1nFt), so additional complexity did not seem worth it (also, all 10 cases currently existing are `footway:surface=sett` + `cycleway:surface=sett` + `surface=cobblestone`, which would correctly be converted by code to more precise `surface=sett`). But even if they were more popular, less precise `surface=paved` would be preferred to removal of `surface=*` (which is what currently happens), so this PR would be step in right direction.

westnordost: > so removing surface completely seems too harsh.

Well, it was a mistake before, just the mistake is removed. Also, StreetComplete will then just ask about the surface, doesn't it?

> I can change AddPathSurface.kt to skip quest on surface=paved if both footway:surface and cycleway:surface are present (AddRoadSurface.kt and AddPitchSurface.kt which also check that are not relevant here)
> Alternatively, I can add a surface:note=different footway and cycleway surfaces when I set surface=paved in this PR
> 
> Which one would you prefer?

Neither.

mnalis: > Well, it was a mistake before, just the mistake is removed.

I'm not sure if you talk about mistake in `v48` and older (i.e. invalid `surface=*` being left behind) or mistake in `v49.0` (i.e. `surface=*` being removed which is not wrong, but might have been handled better in some cases). If the latter, yes, this PR tries to **_improve on_** what the previous one (PR https://github.com/streetcomplete/StreetComplete/pull/4548) accomplished.

>  Also, StreetComplete will then just ask about the surface right after, doesn't it?

After some extra looking... No, it will not.

Which also means there is no need for workarounds I incorrectly suggested in last post. 
So, allow me to correct my answer on your initial remark:

>  As far as I remember, StreetComplete asks for the surface of all roads with surface=paved if not also a note is supplied

with "no, it will not ask for the `surface=*` on ways with `segregated=yes`, on which quests in this PR operate. "

westnordost: No, I talk about the mistake that would be introduced if the tagging was like this:

`footway:surface=paving_stones`
`surface=paving_stones`

and `cycleway:surface=asphalt` was added. 

And this is what you already implemented.

> No, not as far as I can tell, because path surface=* will be asked only on:

Ah okay, that's fine then.

I am a bit confused right now, I will revisit this PR later based on the information in this thread.

mnalis: OK. Previously, your example would remove `surface=*` tag. This PR improves that so in such example `surface=paved` would be added instead.

mnalis: This is also recommended way to tag it, see second block in `S5` on https://wiki.openstreetmap.org/wiki/Bicycle

westnordost (commented):
Hm, I notice that I refactored the `SurfaceAnswer.applyTo` function to also remove `smoothness` if the surface was changed. Hence, part of this PR is obsolete/wrong.

However, I kept you waiting long enough and to save on further back and forth, I will make the changes myself before merging.

smichel17:
@mnalis On review requests: I've found them useful in $dayjob not because of hierarchy but more like a shorthand for "I know you worked on this feature in the past, or have some stake in maintaining it, can you check if this is an ok change?" I think same thing applies here; someone might want to request my review when touching the recent answers code, for example.

Issue #3030, opened 2021-07-08
StreetComplete is excluded from adb backup


gdt:
**Description**

In preparing for a phone migration I did adb backup and unpacked it.  The streetcomplete app was missing.
The FAQ does not say that it being missing from backups is a feature.   (Yes, I realize that the local state is minor.)

**How to Reproduce**

Do adb backup and then use something like Android Backup Extractor https://github.com/nelenkov/android-backup-extractor to unpack the backup.   Look for StreetComplete without success.

**Versions affected**

StreetComplete 33.0

westnordost:
Why should it be included?

gdt:
So that when you back up your phone and restore it, it's there.   That's the point of backup, as I see it.

I see the point that I think you're trying to make that there isn't a lot of state, but there is still the count of completed quests, and more importantly the records of quests the user said they didn't know, and configuration of quest types not to do, etc.

Why do you think the app and the settings should disappear across backup/restore?

gdt:
(This is prompted by a real situation of backup/restore for phone upgrade.)

westnordost:
I don't think the app should disappear. But f.e. the downloaded quest cache, the downloaded tiles etc. Even the quest count and stats is simply resynced after login

gdt:
Sure, the cache stuff should be omitted, as well as things that can be synced.  And it's fine to omit the oauth cookies as a security measure.

I did "adb backup" and then I unpacked it and I see most apps, but things like Signal are missing that explicitly exclude themselves.

I wonder if you just do "adb backup" and unpack if you see StreetComplete there.
I am using an old copy of an extractor that I no longer remember where I got i from, but this might be the normal thing now:
https://github.com/nelenkov/android-backup-extractor

gdt:
Backups are disabled in line 29 of app/src/main/AndroidManifest.xml

smichel17:
It was added in https://github.com/streetcomplete/StreetComplete/commit/611005bd3e4ebd0cffc9e5f117c3920ff13eecdd. The commit message is,

> do not allow backup
> 
> backup has some security implications. Plus, the local data of StreetComplete will almost all the time be above the 25MB limit, so backup does not work most of the time anyway. To backup downloaded quests makes no sense. The only thing that makes sense would be not-uploaded-yet answers, but they should not be stored for a lengthy period anyway plus we wouldn't want these answers to be spread over all of the user's devices (cause then there would be upload conflicts)

[...]

PR #4639, opened 2022-11-19
Remove allowbackup=false

tiziodcaio:
So you allow backups and we can use with seedvault

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -32,7 +32,6 @@
 
     <application
         android:name="de.westnordost.streetcomplete.StreetCompleteApplication"
-        android:allowBackup="false"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"


matkoniecz:
have you seen reasoning in https://github.com/streetcomplete/StreetComplete/commit/611005bd3e4ebd0cffc9e5f117c3920ff13eecdd ?

And #3030 ?

Cj-Malone:
Having just set up a new phone, I'd have liked my enabled quests transferred. Theoretically if I had both phones at the same time I could have done the new preset share, but that's not ideal.

westnordost:
Is it possible to select programmatically what gets backed up and what does not? Because as written before, there is a certain maximum cloud backup size (local backup is kind of rather for the technical affine users who could just as well copy from the filesystem) which is easily exceeded by any one quest download.

mnalis:
> Theoretically if I had both phones at the same time I could have done the new preset share, but that's not ideal.

Workaround: Even if you don't have both phones at the same time, you can click that "copy" button, and then paste the links in some text file (or send yourself an email or whatever), and then import quest by opening that link on a new phone later. (but yes, I agree it's not terribly efficient, especially if you have multiple presets. Then again, one probably doesn't changes phone too often). 

I don't know about android internals on how to make it automatic, but in docs there seems to be section called  [Include and exclude files](https://developer.android.com/guide/topics/data/autobackup#IncludingFiles) which might help this PR to include only user settings and presets, and exclude other stuff (quests should probably be uploaded before transitioning to new phone, or that might make backup more complicated)

westnordost:
Hm well, currently, all is saved in the one shared pref file + the database (which is also a file). There is no distinction made between data that should be kept when moving to a new phone and data that should not be kept. This would require having two database files and deciding for each table in which it should go.
The same with the shared preferences.

I am not really opposed to this, as long as it can be implemented without a lot of code. However, this would be the prerequisite to enable backup, IMO.

westnordost:
Anyway, I will close this. Feel free to create a new PR if the above can be implemented with a reasonable amount of code.

PR #4694, opened 2022-12-17
fix #4688

### LINKED ISSUE #4688: NSI not adding names
Today I added a shop and i searched for the brand name, i clicked on the correct brand and SC added the POI. However, I noticed that the name of the shop was not added ([this is the POI](https://www.openstreetmap.org/node/10267676636)) even if the entry in the NSI should automatically enter the name (see [the entry](https://nsi.guide/index.html?t=brands&k=shop&v=perfumery&tt=marionnaud#marionnaud-d22512)).

**How to Reproduce**
Add a shop and look for "Marionnaud": the app adds a POI without adding the name.

**Expected Behavior**
The POI should be added with the name

**Versions affected**
Android 12 SP1A.218012.016 (MIUI Global 13.0.1 Stable on XIAOMI 2201116TG); StreetComplete v 50.0

matkoniecz:
making it PR as I have not figured out what was being fixed in https://github.com/streetcomplete/StreetComplete/commit/1cc3c96aa24e6a8fd7ae4d82f4797c127966f430 as far as user visible changes go

also, sadly I have not figured out how to add a nice test in this case that would test for this regression

fix #4688

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/shops/ShopsOverlayForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/shops/ShopsOverlayForm.kt
@@ -294,7 +294,11 @@ private suspend fun createEditAction(
         }
     }
 
-    newNames.applyTo(tagChanges)
+    if (!isFeatureWithName) {
+        // in this case name input was not even shown so newNames will be empty
+        // newNames should not be applied as it will erase names provided by NSI
+        newNames.applyTo(tagChanges)
+    }
     if (newNames.isEmpty() && isNoName) {
         tagChanges["name:signed"] = "no"
     }


matkoniecz:
Also, maybe filling `newNames` (and still not showing them to user) would be a better fix.

westnordost:
Thank you for the fix!

matkoniecz:
Do you remember what https://github.com/streetcomplete/StreetComplete/commit/1cc3c96aa24e6a8fd7ae4d82f4797c127966f430 was fixing?

Was it supposed to have some user-visible effects or was it changing internal system to start behaving as advertised?

westnordost:
Yes,
- `noname` and `name:signed` was not removed when a name was specified
- other previously specified names were not removed when they were removed in the UI

PR #4706, opened 2022-12-27
fix cycleway quest is visible for roads in zone 30 (and maxspeed=walk)

westnordost:
see https://github.com/streetcomplete/StreetComplete/commit/3daf11aff37687ebe4021c083ed18c6d26ce8d2b#r94339607


westnordost:
Thank you for testing this!

matkoniecz:
Is it safe to delete this branch now?

Issue #4664, opened 2022-12-03
Support new street parking tagging scheme


westnordost:
The proposal for a [new street parking schema](https://wiki.openstreetmap.org/wiki/Proposed_features/street_parking_revision) will very likely be approved.

Part of the proposal is to deprecate the old schema. Hence, in order to not use deprecated tags, the app must use the new schema too.

**Affected files**
- [StreetParkingParser.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingParser.kt) + [tests](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/test/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingParserKtTest.kt)
- [StreetParkingCreator.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingCreator.kt) + [tests](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/test/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingCreatorKtTest.kt)
- maybe [StreetParking.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParking.kt) + [StreetParkingItem.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingItem.kt) + [tests](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/test/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingKtTest.kt) and [StreetParkingOverlay.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/overlays/street_parking/StreetParkingOverlay.kt) + [StreetParkingOverlayForm.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/overlays/street_parking/StreetParkingOverlayForm.kt) in case new selections should be added or changed

**Implementation**
The parser that parses the old scheme should first be renamed to `OldStreetParkingParser` and be used as a fallback for a while as long as roads that are tagged with the old schema are still around.

<strike>
<b>Coordination required</b>

[...]

PR #4717, opened 2023-01-02
Update parking overlay to new schema

tordans:
This draft PR is just a collection of hints for https://github.com/streetcomplete/StreetComplete/issues/4664.
I don't know enough about Kotlin/Android/Mobile development to take this further. Maybe someone finds helpful notes here…

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking copy/StreetParking.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/street_parking copy/StreetParking.kt
@@ -1,0 +1,71 @@
+package de.westnordost.streetcomplete.osm.street_parking
+
+import de.westnordost.streetcomplete.osm.street_parking.ParkingOrientation.DIAGONAL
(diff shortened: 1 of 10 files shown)


westnordost:
Hmm,
- the data structure (StreetParking.kt), its representation (StreetParkingDrawable.kt, StreetParkingItem.k) and the creator of the OSM tags (StreetParkingCreator.kt) do not really need to be touched
- As written in the original issue, what mainly needs to be done is to 1. rename the current StreetParkingParser.kt and associated tests to OldStreetParkingParser.kt, 2. create a NewStreetParkingParser.kt (copying the old one and adapting it) that parses the new scheme plus associated tests. Worrying about the odd ends and bits (such as updating comments, renaming constants or adding/removing a value from the data structure) can be done after this.
- you don't need to know anything about mobile development as parsing does not include any android parts. Is there anything in particular you do not understand in the `StreetParkingParser.kt`. You know, it just takes a Map<String, String> as input and returns the data structure as output, there is no magic to it other than maybe some kotlin syntax sugar

westnordost:
obsolete

Issue #4289, opened 2022-08-16
non-free dependency: Google AR Core


IzzySoft:
I've just scanned the APK built by F-Droid, and see:

```
Offending libs:
---------------
* Google ARCode (/com/google/ar/core): NonFreeDep

1 offenders.
```

This would mean all affected builds need to be disabled, as that's a no-go for F-Droid. So the questions are:

* I've looked at the corresponding smali code (generated via `apktool d de.westnordost.streetcomplete_4504.apk`), this does not look like a "stub". Am I mistaken?
* If it's no stub (which is to be assumed, as it's [explicitly implemented](https://github.com/streetcomplete/StreetComplete/blob/master/app/build.gradle.kts#L202)), is this essential – or can it be excluded e.g. by using a build flavor, or replaced by something else altogether?

Being a StreetComplete user myself, and just having made a big [announcement at Mastodon](https://mastodon.technology/@IzzyOnDroid/108820687843825531) how useful the app is, I'd hate seeing it disabled at F-Droid – so I really, really hope you can find a way to mitigate this.

matkoniecz:
F-Droid can maintain a patch disabling this and remove functionality in few quest relying on it which measure width of some objects.

There is already fallback to measuring width with a tape and entering data manually.

matkoniecz:
If SC can be changed without making it worse in way that makes maintaining patch easier, I guess that they can submit such pull request.

Or if there is a drop-in replacement for this library (afaik no) that also would be nice to know.

licaon-kter:
I've seen and asked about this in https://github.com/streetcomplete/StreetComplete/pull/3709 at that time and was at ease since it said "but it's foss"

But now looking at https://github.com/google-ar/arcore-android-sdk and comparing with the AAR served by maven https://mvnrepository.com/artifact/com.google.ar/core/1.32.0 which has .so's and classes and whatnot, I fail to see WHERE is the source code exactly. :(

IzzySoft:
Thanks for the fast answer! Indeed I'm not aware of any "smooth replacement" either (but I'm no dev, so I might simply have not heard of it). Thanks for the detail that it's at least not "crucially essential"! @licaon-kter maybe you could take a look at how to achieve that, so builds can be re-enabled?

HolgerJeromin:
Adding some references for open AR replacements:
https://github.com/streetcomplete/StreetComplete/issues/2093#issuecomment-1090606995
[...]

PR #4740, opened 2023-01-08
Outsource arcore into own app

westnordost:
...and related fixes


matkoniecz:
Is `measure_info_title` string still needed?

westnordost:
https://github.com/streetcomplete/StreetMeasure/blob/master/app/build.gradle.kts#L69-L92

matkoniecz:
this fixed #4289

PR #4750, opened 2023-01-13
Ask fire hydrant diameter again after 6 years if unsigned

FloEdelmann:
I noticed in my neighborhood that during construction work that moved a street lantern, a fire hydrant sign was installed for an (underground) fire hydrant that has had no sign for a long time (and was otherwise unchanged by the construction works). Since that hydrant was tagged with `fire_hydrant:diameter:signed = no`, StreetComplete would have not asked about that hydrant anymore.

I suggest to ask the diameter quest again after 6 years for hydrants tagged with `fire_hydrant:diameter:signed = no`. In that time, a sign might have been added. Or if the sign was actually installed the first time but quite far away or hidden, resurveying it (maybe by another mapper) might result in the diameter eventually being added.


matkoniecz:
> If in Poland, no fire hydrants have this info signed, the quest should be disabled for Poland.

That happens but is really rare in my experience.

> and be backed up by (OSM) data collected so far

there are firefighters adding `diameter` based on fire hydrant model (and were doing before SC was a thing) and some imports so this will require parsing SC edits (or maybe my city and where I visited are unusual places but that seems unlikely)

westnordost (approved):

Issue #4718, opened 2023-01-03
Ameliorate address quest (needed for India)


letypequividelespoubelles:
**Use case**
In India, a lot of addresses (at least where I live) are tagged with addr:housenumber=Letter-XXX & addr:place=* but no addr:street=*

It's really a pain to add addresses there (via the address overlay) because :
1) we can't do "+1" at each incrementation of housenumber because the housenumber doesn't contain only number (like C-245 for exemple)
2) as there is no street tag, we need to add via addr:place=* tag, so need to "select other", and then write the name of the place.

**Proposed Solution**
1) it should be possible to do +1 / -1 for housenumber like C-246, it'll be faster.
2) a) streetcomplete should remember that we tag addr:place= instead of addr:street=
   b) streetcomplete should remember the name of the place, so that we don't write it everytime
  c) SC should propose name of nearby place when starting to write the name of the place to avoid errors

PS : for 2) it should be the same for the "What is the streetname of this housenumber" quest.

Thanks !

westnordost:
All three should be possible. I'll mark it as an enhancement, so anybody interested can implement this.

letypequividelespoubelles:
@Helium314 Thanks

>is `<letter>-<number>` the only format you have issues with, or are there other (common) ones?

I don't know any other common ones in India so it should be enough (but India is huge).

>It will put the form in place-mode if the previous answer was a place. Switching back from place to street needs to be done using the spinner.

Perfect. That way, will it be possible to add street and place tag ? (because some roads have name, so address is mainly place, but adding street in addition is better)

>b) is might be too much work for me

SC already remember street name, can't we implement the same for place ? 

>c) as well, plus I don't see how it would work realistically, as I found areas where almost 1000 m search radius would be necessary.

If 1), 2)a)b) are already implemented, c) is only the cherry on the cake so no need to spend too much time on this

westnordost:
>@westnordost is something like https://github.com/Helium314/StreetComplete/commit/965c06edec1b167db51482f735b7d404b422e8b9 ok for 2. a)?

Looks like an improvement.

Something that will have a much larger impact on the overall productivity with the housenumbers layer would be the following feature which is indicated a bit in b) / c):
- get all elements in a certain radius around the current position with a housenumber
- if the street field is not filled in by the user yet: after each change to the housenumber field (either by typing or pushing the +/-), check the `addr:streetname` of any neighbouring housenumber and fill it in

[...]

PR #4753, opened 2023-01-15
Parse housenumbers starting with letter

Helium314:
Improves #4718

This allows housenumber parsing for numbers like C-245, and some similar housenumbers I found when randomly checking housenumbers in India: up to two letters, and `-`, `/` or nothing as separator.

Though maybe the added parsing doesn't integrate nicely with the other code, maybe this should be improved?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/address/HouseNumbers.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/address/HouseNumbers.kt
@@ -89,6 +89,14 @@ data class HouseNumberWithLetter(
 ) : StructuredHouseNumber() {
     override fun toString() = "$number$separator$letter"
 }
+/** e.g. C-12 */
+data class HouseNumberWithLetterAtStart(
+    override val number: Int,
+    val separator: String,
+    val letter: String
+) : StructuredHouseNumber() {
+    override fun toString() = "$letter$separator$number"
+}
 /** e.g. 12/3 */
 data class HouseNumberWithNumber(
     override val number: Int,
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/address/HouseNumbersParser.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/address/HouseNumbersParser.kt
@@ -10,8 +10,12 @@ fun parseHouseNumbers(string: String): HouseNumbers? {
         when (range.size) {
             1 -> SingleHouseNumbersPart(parseHouseNumberParts(range[0]) ?: return null)
             2 -> {
-                val start = parseHouseNumberParts(range[0]) ?: return null
-                val end = parseHouseNumberParts(range[1]) ?: return null
+                val start = parseHouseNumberParts(range[0])
+                val end = parseHouseNumberParts(range[1])
+                if (start == null && end != null) {
+                    return@map SingleHouseNumbersPart(parseHouseNumberParts(part) ?: return null)
+                }
+                if (start == null || end == null) return null
                 if (start < end) {
                     HouseNumbersPartsRange(start, end)
                 }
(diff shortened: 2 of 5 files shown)


app/src/main/java/de/westnordost/streetcomplete/osm/address/HouseNumbersParser.kt
```diff
@@ -27,7 +31,20 @@ fun parseHouseNumbers(string: String): HouseNumbers? {
...
+        return if (separatorWithNumber != null && c.isAtEnd()) {
+            HouseNumberWithLetterAtStart(
+                separatorWithNumber.groupValues[2].toInt(),
+                separatorWithNumber.groupValues[1],
+                letterAtStart
+            )
+        } else {
+            null
+        }
+    }
```
westnordost, line 47: This seems to be unnecessarily complicated. Why not

letterAtStart = ...
if (letterAtStart != null) separator = ... ?: ""
houseNumber = ....

if (letterAtStart != null ) return HouseNumberWithLetterAtStart(...)

app/src/main/java/de/westnordost/streetcomplete/osm/address/HouseNumbers.kt
```diff
@@ -89,6 +89,14 @@ data class HouseNumberWithLetter(
 ) : StructuredHouseNumber() {
     override fun toString() = "$number$separator$letter"
 }
+/** e.g. C-12 */
+data class HouseNumberWithLetterAtStart(
```
westnordost, line 93: 1. C-12a, C-12/3 not possible with this data structure. But I don't see a reason why not. In the end, the letter at the beginning is more or less something like a `addr:block_number` baked into the `addr:housenumber`.

2. I think the `compareTo` method in `StructuredHouseNumber` should be adapted too?

3. Given 1, maybe it would make sense first to inquire why `addr:block_number` is not used in places where `C-13` is used as housenumber

Helium314: > 3. Given 1, maybe it would make sense first to inquire why `addr:block_number` is not used in places where `C-13` is used as housenumber

What is clear is that it is not used (`addr:block_number` is used very rarely, and not with a letter).
@letypequividelespoubelles do you have some more information about this? Is the letter really actually a block number?

letypequividelespoubelles:
@Helium314 Yes, the letter refers to the block number. But the written address is C-245 for exemple, and no one talks about house 245 in C block, but of house C-245. 
But maybe it's better to wait, I'll ask again Indian community in telegram to be perfectly clear.

westnordost:
Which city by the way? There are very few planned cities in India. Is the block system used elsewhere (too)?

letypequividelespoubelles:
New Delhi. Idk for other cities in India.

westnordost:
Hm, where exactly?

In another part of Delhi, blocks seem to be numbered. I wonder if the block number is used in the address, too?
https://www.openstreetmap.org/way/259086505#map=17/28.64830/77.19268

Helium314:
There are also a few other cities with similar addresses... I used this overpass query to find house numbers that are not only numbers and "usual" separators:
```
[out:json][timeout:25];
(
  node["addr:housenumber"~".*?[^0-9/-].*"]({{bbox}});
  way["addr:housenumber"~".*?[^0-9/-].*"]({{bbox}});
  relation["addr:housenumber"~".*?[^0-9/-].*"]({{bbox}});
);
out body;
>;
out skel qt;
```

letypequividelespoubelles:
>where exactly? 

You can check different location like Defence colony, Nizamuddin, Greater Kailash, Lajpat Nagar, Lodhi Colony, Anan Niketan, etc. All these colonies are subdivised in "letter" block, and all housenumber inside are of the form letter-number. I didn't find any block mapped though, so can't provide you OSM link, sorry. 

But here is a picture of a rare plan showing block I made this morning.

photo_2023-01-16_12-51-41


Concerning addresses in India, I haven't found anything more than [this wiki talk](https://wiki.openstreetmap.org/wiki/Talk:India), which doesn't provide any guide rules. 

The key addr:block_number seems to be used[ just in Japan](https://taginfo.openstreetmap.org/keys/?key=addr%3Ablock_number#map), and addr:block used [mainly in India](https://taginfo.openstreetmap.org/keys/?key=addr%3Ablock#map). So maybe the best is to map with addr:housenumber=number ; addr:block=letter ; addr:place= place.

I tried to ask indian community, but ... not very efficient. (I'm not Indian, I'm from Europe and spend 1-2 years in India).

Helium314:
So probably better use `addr:place`, `addr:block` and `addr:housenumber`, Which would remove the need for this PR, by just allowing to switch to something like the Japan-style UI.

From the wiki entry I get that there is / was the attempt to make addresses so that nominatim shows the "right" things. Here it would probably be better to have nominatim extract the "right" information instead...

westnordost:
So, hm, close this then? After all, even in Delhi, baking block labels into the housenumber does not really work because there are also numbered blocks https://www.openstreetmap.org/way/259086505#map=17/28.64830/77.19268 and `addr:housenumber=12-31` would be interpreted differently.

PR #4754, opened 2023-01-15
Add seasonal grit bin quest

### LINKED ISSUE #4749: Quest suggestion: Is this grit bin here the whole year?
### General
Affected tag(s) to be modified/added: [seasonal](https://wiki.openstreetmap.org/wiki/Key:seasonal)
Question asked: **Is this here the whole year?**

### Checklist
  → `seasonal` tag is established, but not widely used on `amenity = grit_bin` yet
  → can be used to re-enable #2726 for non-seasonal grit bins
  → survey is necessary with at least half a year in between. That's why the quest should be disabled by default, like the indoor/outdoor seating quest
- [ ] 💤 Not an overwhelming percentage of quests have the same answer (No spam)
  → not sure, might be depending on the area, see https://overpass-turbo.eu/s/1q7A
  → yes, see https://overpass-turbo.eu/s/1q7z

### Ideas for implementation

**Element selection:** `amenity = grit_bin and !seasonal`
**Proposed UI:** Yes/No quest form
**Quest icon:** Maybe a calendar, similar to https://www.svgrepo.com/svg/362042/calendar-clock

### Related issues

* #2726
* https://github.com/streetcomplete/StreetComplete/issues/2744#issuecomment-885065544
* https://github.com/streetcomplete/StreetComplete/issues/3943

FloEdelmann:
Fixes #4749.

Example element: https://www.osm.org/node/10542534426

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/seasonal/AddSeasonal.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/seasonal/AddSeasonal.kt
@@ -27,6 +27,6 @@ class AddSeasonal : OsmFilterQuestType<Boolean>() {
     override fun createForm() = YesNoQuestForm()
 
     override fun applyAnswerTo(answer: Boolean, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
-        tags.updateWithCheckDate("seasonal", answer.toYesNo())
+        tags.updateWithCheckDate("seasonal", (!answer).toYesNo())
     }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/seasonal/AddSeasonal.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/seasonal/AddSeasonal.kt
@@ -1,0 +1,32 @@
+package de.westnordost.streetcomplete.quests.seasonal
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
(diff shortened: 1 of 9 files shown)


westnordost:
Better ask if this is here also in the summer? It seems to be easier to answer of a question and has the same effect.

app/src/main/java/de/westnordost/streetcomplete/quests/seasonal/AddSeasonal.kt
```diff
@@ -0,0 +1,32 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddSeasonal : OsmFilterQuestType<Boolean>() {
```
westnordost, line 12: As this is about grit bins only, why not name it that way?

Helium314: Probably because it could be extended... could be useful e.g. for mobile toilets that are stationary, but removed in winter (I know some).

westnordost: That's fair. However, for mobile toilets for example, the question is somewhat unanswerable. Because you would have to check... I don't know, once every month to see if it is still there, and when you made 12 ✔️ , then you can answer "Yes, it's always there". For grit bins (winter) and outdoor seating (summer), you at least know which time of the year you have to look out for.

FloEdelmann: Changed in [`d57dcdd`](https://github.com/streetcomplete/StreetComplete/pull/4754/commits/d57dcdd7311e85f7916cc43ba00f849f3a21ef07).

westnordost (approved):

PR #4756, opened 2023-01-16
Don't ask whether vacant shop is still the same shop

### LINKED ISSUE #4752: SC asks me if a vacant shop is the same shop
In the "Shops" layer, if I click "It's vacant", StreetComplete asks me if the shop is still the same shop or a different one, as when you rename a shop. In this case tho I find it quite confusing, also because both choices seems to have identical effects.

I think it shouldn't ask that question 
or
it could ask "Is the name still visible?" (as suggested in the [disused:* wiki page](https://wiki.openstreetmap.org/wiki/Key:disused:*): _"It no longer has an operator as such - no beer is sold - but the name is still useful for navigation."_) and keep/delete the name=* based on the reply.

StreetComplete v50.2

Helium314:
fixes #4752

Tag cleanup should definitely happen when a shop is tagged as vacant (maybe with exception of name, but that would require a separate "other answer")


westnordost (approved):

westnordost:
Thank you for fixing this!

PR #4764, opened 2023-01-26
Add alt_name to Vkusno i tochka

linux2good:
Most POI's of this brand have it, so it would be nice to automatically add it with shop overlay.


matkoniecz:
See https://github.com/streetcomplete/StreetComplete/blob/749434f8cd53c4744e141da6c8be4ba0062efd32/app/src/main/assets/authors.txt

We get them from https://github.com/osmlab/name-suggestion-index - so changes should be applied there. Maybe this should be more prominently mentioned?

Also note that

> Most POI's of this brand have it

is likely not a good reason to apply it to every single shop of this brand

Issue #3401, opened 2021-10-17
Improve opening/emptying times picker


Emilius123:
The current time-picker dialog for opening/emptying hours seems to not be intuitive. I have personally seen many people fail at correctly inputting the times.

While the Android Timepicker in itself is a thing that should be fixed sometime, the problem seems to lie in the way, StreetComplete asks for the times:

1. What days of the week is this open? -- Many will select all days the place is open, even if the times itself are different
2.  When does this open and close? -- Even if the days were given correctly, most will just give the times of total opening and total closure, not including any breaks that might have been.

Adding the times totally correctly is just too complicated, most users might not even notice the "Add opening hours" button.

While this problem might be solved in many ways, my suggestion would be to use a grid-like system, just as the [JOSM OpeningHoursEditor](https://wiki.openstreetmap.org/wiki/JOSM/Plugins/OpeningHoursEditor) does:
![image](https://user-images.githubusercontent.com/47276235/137646923-803afd14-76db-45c4-abe9-c9846fa17cbe.png)

Even if not by default, such a system would probably be very desireable for some.

matkoniecz:
> The current time-picker dialog for opening/emptying hours seems to not be intuitive. I have personally seen many people fail at correctly inputting the times.

Interesting. I am in middle of UX testing and I identified some problems (see say #3364  #3394 #3341) but - to my surprise - people had no problem with opening hours interface.

I wonder what is the cause of difference, I was not expecting significant cultural impact here.

> What days of the week is this open? -- Many will select all days the place is open, even if the times itself are different

Some run into it initially but - again to my surprise - they figured out how to fix it.

> When does this open and close? -- Even if the days were given correctly, most will just give the times of total opening and total closure, not including any breaks that might have been.

In place where I was testing it such places are really rare so I had no opportunity to see it in action.

> While this problem might be solved in many ways, my suggestion would be to use a grid-like system, just as the JOSM OpeningHoursEditor does:

That is not feasible due to screen size.

PR #4766, opened 2023-01-26
Add 'Uncheck all' button to Weekday Picker

jakeisnt:
I've been using StreetComplete frequently, and often it's faster to clear the selection of all of the weekdays when specifying opening hours than to uncheck all of the ones you don't want. This PR attempts to introduce a neutral button, 'Uncheck all', to the dialog that unchecks all of the checkboxes while leaving the window open.

TODO:
- Fix overriding the button's onClick to allow it to refresh the array on click. It's necessary to override the method to prevent the dialog from closing when pressing the neutral button. I'm not sure why, but the list isn't being refreshed.
- Add internationalization to 'clear all'

Any help would be appreciated!

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/WeekdaysPickerDialog.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/WeekdaysPickerDialog.kt
@@ -1,6 +1,9 @@
 package de.westnordost.streetcomplete.quests.opening_hours
 
 import android.content.Context
+import android.view.View
+import android.widget.ArrayAdapter
+import android.widget.ListView
 import androidx.appcompat.app.AlertDialog
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.osm.opening_hours.model.Weekdays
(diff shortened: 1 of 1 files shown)


westnordost:
>  and often it's faster to clear the selection of all of the weekdays when specifying opening hours than to uncheck all of the ones you don't want

Why? What are the usual opening days of shops in your area?

matkoniecz:
> The status quo is that when adding the second set of hours, I'll have to uncheck each of Monday, Tuesday, Wednesday, Thursday, and Friday, then check off Saturday and/or Sunday. 

Can you attach a screen capture how you add them? Or step by step explanation how you do this?

This should not happen, once some days are added then Mon-Fri should be no loger preselected

westnordost:
Then this is a bug. Once you defined Monday to Friday, if you add another set of weekdays, Monday-Friday should not be selected (but instead Saturday or nothing).

matkoniecz:
I uploaded it here (I hope that it is fine):
https://user-images.githubusercontent.com/899988/215093965-36aaa88b-4381-48ec-a1e8-b1294557d294.mp4

westnordost:
I find the suggested name a bit long, and I am not sure if it is more clear.

I suggest to close this as misunderstanding/question answered for now and monitor if the misunderstanding remains to be an isolated case or if more people tend to misunderstand it.

jakeisnt:
The semantic discrepancy here has to do with what's being added. Both options add  hours, so 'adding days' doesn't make as much sense. They add additional hours to different aspects of the form, so that distinction is also necessary. 

The length of these strings is not significantly different from the current second string, so I think it's fine.

That said, I'll close this as the code change is no longer relevant. Keep the text change in mind!

mnalis:
I do agree that the current wording is less then perfect... (while I personally have caught on early, I do remember it was requiring some thinking / trying out the first time - e.g. https://github.com/streetcomplete/StreetComplete/issues/3401#issuecomment-946265276)... 

For me additionally it forces brain to jump out of "instinct" mode, because all three options start with "Add ...", so are hard to **visually** distinguish / reflexively click the correct one (i.e. without engaging the higher brain functions, i.e. reading+translating+understanding mode).

Perhaps something along the lines of `Extend Hours` or `Extra hours" (instead of current "Add hours") would be less misleading and easier to distinguish, yet short enough? Or do people have better ideas?

PR #4768, opened 2023-01-27
Support new parking schema (and add staggered parking)

### LINKED ISSUE #4664: Support new street parking tagging scheme

westnordost:
Fixes #4664

Support new parking tagging schema:
- understand the tags from the new `parking:<side>` schema
- still understand old tags from `parking:lane` schema. If both old and new are tagged, new tags take precedence
- when (re)tagging, always tag according to the new schema and clear any tags of the old schema
- all tested of course

<br clear="left">

### More details

- `parking:lane:<side>:<type>=painted_area_only` becomes `parking:<side>=lane` + `parking:<side>:markings=yes` + `parking:<side>:staggered=yes` in the new schema. It is also still available as a selection because the description text for staggered parking would become too long ("On street on alternating sides **or** only in dedicated spaces"), IMO
- added two selectable options (per orientation) for ["staggered" parking](https://wiki.openstreetmap.org/wiki/Street_parking#Staggered_parking_on_narrow_roads). Did not use the word "staggered" for the source (en-US) wording after some discussion in the #language channel on [slack US](https://osmus.slack.com/archives/C01BU04GW7L/p1674827093328859). Other suggested wordings included _"room for parking on one side at a time", "haphazard parking"_. Note in particular this comment: _“Staggered left- and right-side parking” isn’t terrible, but it implies a degree of planning that just isn’t there with the unmarked case you brought up."_
- the short descriptions of the selections are now shown in the preview WYSIWYG view because especially the "staggered" stuff is not that self explanatory
- `shoulder` as a value is (currently) not supported. If it will be supported, then only as a synonym to `on_kerb` with the original tagging being preserved if it is already tagged with `shoulder`. Reasons summarized here: https://wiki.openstreetmap.org/wiki/Talk:Street_parking#Suggestion_to_remove_parking:side=shoulder


app/src/main/java/de/westnordost/streetcomplete/osm/street_parking/StreetParkingParser.kt
```diff
@@ -1,29 +1,68 @@
...
-    val left = createParkingForSide(expandedTags, "left")
-    val right = createParkingForSide(expandedTags, "right")
+    // first try to parse new schema
+    var left = createParkingForSide(expandedTags, "left")
+    var right = createParkingForSide(expandedTags, "right")
+    if (left == null && right == null) {
+        // then get the values for left and right
+        left = createParkingForSideOldSchema(expandedTags, "left")
+        right = createParkingForSideOldSchema(expandedTags, "right")
+    }
```
FloEdelmann, line 15: Moving the comment now sounds misleading. Maybe this is better?

```suggestion
    // then get the values for left and right, first try to parse new schema
    var left = createParkingForSide(expandedTags, "left")
    var right = createParkingForSide(expandedTags, "right")
    if (left == null && right == null) {
        // try parsing the old schema instead
        left = createParkingForSideOldSchema(expandedTags, "left")
        right = createParkingForSideOldSchema(expandedTags, "right")
    }
```

westnordost: thanks, I'll change it

scaidermern:
> * `shoulder` as a value is (currently) not supported. If it will be supported, then only as a synonym to `on_kerb` with the original tagging being preserved if it is already tagged with `shoulder`. Reasons summarized here: https://wiki.openstreetmap.org/wiki/Talk:Street_parking#Suggestion_to_remove_parking:side=shoulder

I strongly suggest to stick to the official Wiki documentation. There has been a proposal, the OSM community agreed on it, therefore the tagging should not be changed.

If you are unhappy with the result then start a new proposal. Until then please stick to the official tagging.

Apart from that, thanks for implementing this tagging scheme!

westnordost:
Well, no, it's not planned for the reasons stated. An approved wiki proposal does not make issues in its real-world application of the schema go away, it can only do its best to avoid such issues in the first place by having many eyes look on it. I looked on it too, but I didn't notice the issue, then.
I also didn't notice that `painted_area_only` from the old schema was deleted. The new schema does not really provide for this situation (which occurs in the real world) which is why I chose the tags as described in the starter post.

For the record, I have no personal investment in not supporting the `shoulder` value as selectable option (i.e. not allowing to distinguish between "on the curb" parking and "beside the road but not on a curb" parking) , but then to actually be able to cover all the situations in the real world, it'd be at least necessary to:
- add and document a value like `half_on_shoulder`
- change the documentation of `on_kerb` (etc.) to make it clear that this value is only to be used on sidewalks with curbs (or when there is no sidewalk but a curb, I guess... maybe just remove the mention of sidewalks altogether)
- change the documentation of `shoulder` to make clear that it has nothing to do with the `shoulder=*` key and should instead only be used on soft shoulders **and** verges but not on hard shoulders (on which `lane` should be used instead)

Discostu36:
_Note: There was overlap between me writing this comment and the answer by @westnordost, but I decided to submit it anyway because I think most of it is still relevant._

I agree with @scaidermern that you should follow the accepted tagging scheme. Especially, you should stick to the definition of `on_kerb`, that cars are parking on the sidewalk. Your current solution would lead to inconsistent and ambivalent tagging, like:
* `sidewalk:right=no` with `parking:right=on_kerb`: This doesn't make sense by wiki definition, but would be tagged with your preset in cases where there is no sidewalk, but cars are parking next to the street. 
*  `sidewalk:right=yes`, `parking:right=on_kerb` and `sidewalk:right:width=3`: With your preset, data consumers couldn't be sure anymore if the width of the sidewalk was really shared between cars and pedestrians. Because maybe there is an area "adjacent to the street", between street and pavement, but it doesn't have "dedicated spaces" and that's why the user chose this option.

So, please don't dilute the `on_kerb` tag by cramming all forms of off-street parking into it. How you write it, it even sounds like that a dilution of the tag was a desired effect because then you would have a reason to propose renaming `on_kerb` into `off_street` in a few months.

Don't misunderstand my comment: I do agree that the current tagging might be further improved. Maybe, `on_sidewalk` would have been a better value than `on_kerb` (because there might be no kerb) and maybe `off_street` would be a better tag value than `shoulder` (to also include verges), but these are things that should be discussed at the appropriate places and decided by the community and not by the maintainer of one of the most successful editors.

westnordost:
You assume in both your examples that all places that have curbs have sidewalks and that parking up on the curb always takes away width from the sidewalk. These assumptions are not correct.

[This sidewalk](https://www.google.de/maps/@53.5881326,9.959352,3a,63y,1.82h,68.45t/data=!3m6!1e1!3m4!1sz8hyY0TQSttUL7gmMnr4aw!2e0!7i13312!8i6656) is about 1.5m wide, not 3.5m, even though the cars are parking `on_kerb`.
If cars were allowed to park up on the curb [here](https://www.google.com/maps/@40.3776026,-80.0567853,3a,75y,191.16h,68.72t/data=!3m6!1e1!3m4!1sZdwUOLKCptmKGH2ARlvmTg!2e0!7i13312!8i6656), still `on_kerb` would be tagged even in absence of a sidewalk. (If this still classifies as a curb, but on the other hand, [the curb of a sidewalk one street down](https://www.google.com/maps/@40.3786451,-80.0543722,3a,73.6y,64.21h,83.11t/data=!3m6!1e1!3m4!1snX3ETz0yTBJ7MxkprnBCVw!2e0!7i16384!8i8192) is equally low)

> it even sounds like [...]

Please don't do this.

westnordost:
Anyway, I am in discussion with the main author of the proposal and happy to implement any solution that may arise. `half_on_shoulder` with the other prerequisites I listed is one, the suggestion I made on the wiki discussion page is another.

Discostu36:
> You assume in both your examples that all places that have curbs have sidewalks and that parking up on the curb always takes away width from the sidewalk. These assumptions are not correct.

No, I assume in my examples that the tag `on kerb` is badly named and actually doesn't have much to do with kerbs at all, as it is defined as "Parking on the footpath area" on wiki. The cars in your first example are not parking on the footpath area and should not be tagged `on_kerb`. But I guess they do show the need for a tag meaning "not on street, not in a parking bay, and not on the sidewalk" that is better than `shoudler`.

> > it even sounds like [...]
> 
> Please don't do this.

Sorry, that was unnecessary.

westnordost:
> The cars in your first example are not parking on the footpath area and should not be tagged on_kerb

Uh, but what then? That's really not a `shoulder` by any definition of "shoulder". @SupaplexOSM, please help me here, surely [this](https://www.google.de/maps/@53.5881326,9.959352,3a,63y,1.82h,68.45t/data=!3m6!1e1!3m4!1sz8hyY0TQSttUL7gmMnr4aw!2e0!7i13312!8i6656) is `on_kerb`, right?

Helium314:
Looks like `on_kerb` is an utterly misleading choice that should be named sth like `on_sidewalk`, or there _will_ be more people assuming `on_kerb` actually means __on kerb__.

SupaplexOSM:
For me, `on_kerb` parking is neither about a kerb nor a sidewalk :D

The crucial factor for _on_kerb_ parking imho is that the space where the vehicles are parked is _either_ part of a dedicated pedestrian area _or_ would be part of a dedicated pedestrian area if no cars were parked there. In situations [like this one](https://www.mapillary.com/app/?lat=52.46860439997201&lng=13.465393499972038&z=19.577418906036474&panos=true&pKey=343937230846946&focus=photo&x=0.20327987788381136&y=0.5158772089241602&zoom=0.36542055968751996), there is no sidewalk/no space to walk beside the vehicles - so there can be on kerb parking without a sidewalk. (But if you remove the street signs that allow parking there (and punish still parking cars there), than it becomes a sidewalk again.)

A dedicated pedestrian area often is demarcated from the carriageway by a kerb. Then the case is clear, I think. Another possibility is, that the pedestrian area is simply created by a different surface (e.g. paving stones, different colors etc.), without the presence of a kerb (e.g. common in living streets). In this case, however, `on_street` or `street_side` parking is usually more appropriate, depending on the structure of the street.

In constructionally "untended" (?) areas (I mean streets with low management efforts), there is sometimes a section of "undefined" space beside the street or between the street and a footpath of some kind, which is sometimes used for parking (often informally, depending on the legal situation). Thats an typical example what `shoulder` should be used for in my understanding. In [this example](https://www.google.de/maps/@53.5881326,9.959352,3a,63y,1.82h,68.45t/data=!3m6!1e1!3m4!1sz8hyY0TQSttUL7gmMnr4aw!2e0!7i13312!8i6656), there is kind of an unmanaged section between the carriageway and the sidewalk paving used for parking, but since there is a kerb that demarcates the "beginning" of the dedicated pedestrian area (although you can't use the space in full width) it's `on_kerb` parking.

I'm surprised that this topic is such a big issue because from my mapping experiences and all example images I saw in the last time, it's always possible to choose a distinct (or lets say, satisfying) category out of the five. But obviously everybody here have an onther understanding of the categories so maybe we should start to just collect sample pictures for the wiki (from usable sources)?

SupaplexOSM:
P.S. Yes, the terms for the values might be not perfect, but like everywhere in OSM they are just strings connected with a definition and a discourse – and, to be honest, I am not convinced by any of the other suggestions. E.g. it's not about "sidewalks", as I elaborated above, and "off street" parking in context of street parking is a term that means parking _not_ on or beside the street but on parking lots, underground or multi storey parking facilities far from the street.

Plus the values are in use since long time and in absolutely most cases (!) they fit very well, also by their literal meaning.

Discostu36:
> For me, on_kerb parking is neither about a kerb nor a sidewalk :D

That is a bit surprising, considering that you proposed to keep that tag value named `on_kerb` and wrote the definition "Parking on the footpath area" with the word "footpath" linking to the Wikipedia article about sidewalks. 

> In [this example](https://www.google.de/maps/@53.5881326,9.959352,3a,63y,1.82h,68.45t/data=!3m6!1e1!3m4!1sz8hyY0TQSttUL7gmMnr4aw!2e0!7i13312!8i6656), there is kind of an unmanaged section between the carriageway and the sidewalk paving used for parking, but since there is a kerb that demarcates the "beginning" of the dedicated pedestrian area (although you can't use the space in full width) it's `on_kerb` parking.

Is there a universal law that a kerb always marks the border of a dedicated pedestrian area, even if this area is obviously not intended for pedestrians?

But apart from the edge cases: Do you agree with me that the preset above is lacking the information that `on_kerb` is not about all unmarked parking adjacent to the street, but only about the cases where cars park on the area next to the road that is designated to pedestrians?

SupaplexOSM:
> Is there a universal law that a kerb always marks the border of a dedicated pedestrian area, even if this area is obviously not intended for pedestrians?

No, kerbs are just demarcating different traffic areas (but often the motorized traffic area from the pedestrian traffic area). To discuss, which value is the best fitting for edge cases with kerbs but obviously no pedestrian area (even if there would be no parked cars), I think we need pictures (might be on street, street side or even shoulder parking).

> But apart from the edge cases: Do you agree with me that the preset above is lacking the information that `on_kerb` is not about all unmarked parking adjacent to the street, but only about the cases where cars park on the area next to the road that is designated to pedestrians?

I can understand why @westnordost does not offer the value in the context of StreetComplete, because StreetComplete is about simple decisions for casual mappers without background knowledge -- but street parking can be very complex and diverse ;) So I think it's a good compromise if the app can read the Value but not write it (and also not delete it). For detailed/professional street parking mapping, you can still use all values you want.

It is similar with the "staggered" issue: no one thought of this during the proposal process, because it is clear that other mapping techniques are available for such special situations if you want it to be precise - i.e. separate mapping. StreetComplete can't do that. So StreetComplete either has to remain very inaccurate or make use of a "hack", as in this case with the staggered value for just one side. I think that's OK, because it doesn't break anything. We just have to document it accordingly (and evaluate it over time, in case street complete users are systematically miss/missuse/missunderstand the intended categories).

I just want to avoid transferring the rather "simple" principle of selecting categories in StreetComplete to OSM in general.

Discostu36:
> I can understand why @westnordost does not offer the value in the context of StreetComplete, because StreetComplete is about simple decisions for casual mappers without background knowledge -- but street parking can be very complex and diverse ;) So I think it's a good compromise if the app can read the Value but not write it (and also not delete it). For detailed/professional street parking mapping, you can still use all values you want.

Sorry, but I don't understand your answer to my question, especially not what value you think StreetComplete should only read but not write.

SupaplexOSM:
> Sorry, but I don't understand your answer to my question, especially not what value you think StreetComplete should only read but not write.

I mean: I'm fine with the decision to not show `shoulder` as a selection in StreetComplete but to leave `shoulder` values on street segments untouched if users confirm them by selecting "on_kerb" (thats how I understand how it will be working now?), because `shoulder` is a very rare value in most places and you can't explain the distinction in a single image and some words (as this thread shows ;-).

westnordost:
> For me, on_kerb parking is neither about a kerb nor a sidewalk :D

Hm, that's not a good start to define a tag value named `on_kerb` with a description exclusively mentioning sidewalks. 😅 Anyway, not wanting to pick on the weakest point in your post:

> The crucial factor for on_kerb parking imho is that the space where the vehicles are parked is either part of a dedicated pedestrian area or would be part of a dedicated pedestrian area if no cars were parked there.

If this is to be the intended meaning of this tag from your point of view, the issue is that it is not documented that way in the wiki. It is the typical OSM "for me, tag X means Y". As `on_kerb` has been inherited from the old `parking:lane` schema, the expectation is that the meaning of that value did not change.
But more importantly, **that definition itself is problematic** because it is very fuzzy and leaves a lot of room for interpretation. After all, what exactly **would be** a dedicated pedestrian area **if** no cars were parked there? I can see people arguing and interpreting this either way:
- one may say, as legislation generally requires pedestrians to use a sidepath of a road if one is available, any space ([including dirt, compacted](https://www.mapillary.com/app/?pKey=1352720785204433&lat=52.4602547&lng=13.4783764&z=17&focus=photo)) would count as a "dedicated pedestrian area" if it was usable by pedestrians. So, any such "shoulder", too.
- in the other end of the interpretation spectrum, one may say that a dedicated pedestrian area adjacent to the street must be a [paved sidewalk with no barriers in-between](https://www.mapillary.com/app/?lat=52.468604399972&lng=13.465393499972&z=17&panos=true&pKey=343937230846946&focus=photo&x=0.20327987788381136&y=0.5158772089241602&zoom=0.36542055968751996) (like trees, lamp posts)

Compare that with "kerb". It is a simple physical property that is either there or not. Of course, there are flat (or no) kerbs too, but for that there is [`kerb=flush`](https://wiki.openstreetmap.org/wiki/Key:kerb) etc.
As I wrote in the talk page to the new street parking scheme, choosing values by "constructional categories" is always going to be problematic.

> simple decisions for casual mappers without background knowledge -- but street parking can be very complex and diverse ;) 

I have to strongly refuse the notion that street parking, actually, any tagging of on the ground info, is so complex that it can only be mapped properly by experts. One does not need background knowledge to accurately record the physical properties of a parking: It is marked, is there a kerb, in which orientation park the cars, is it on or adjacent to the street, etc.. 
Only if the definition of a tag values becomes complex, mapping it becomes complex too. This is what I have been saying - `shoulder` (and now apparently also `on_kerb`) are problematic because their definition is fuzzy and complex.

In any case, this discussion should really rather happen in the wiki, we will not find a solution in this issue tracker.

To summarize, I read that Michael, Alex and Alex find it important to map the distinction between parking adjacent to the road with or without curb. If this is the case, then please take note of the issues I listed, in particular that the schema currently does not allow to record "half on the street but no curb" parking.

SupaplexOSM:
It's an interesting finding, that obviously two opposite categorisation approaches collide here with each other ("physical" vs. "constructional") and in OSM it seems to depend on the feature/it's functions which approach is seen to offer more advantages.

The problem I have with the physical one is, that it don't allow distinctions that are regularly made in real world: E.g. with just looking on the existence of kerbs, markings etc, you hardly can distinguish a street-side parking bay (DE: Parkbucht) although everyone recognises such a parking bay in most cases when they just see it. You might need a longer "decision tree" regarding different physical properties/layouts to do this on a pure physical base. (Long time of my younger life I was a hobby meteorologist and to distinguish/name the shapes of clouds, you don't get far with catchy definitions, but instead turn to picture tables - maybe thats why I prefer this approach ;)

So what I can offer to solve this issue is to help sharpen criterias (you can tell it definition, if you want) over the next time and to collect sample images to better document the values. If there is a community mood to just delete "shoulder" from the list, I'm fine with that, of cause, but I won't be the one to push this ;) And I'm not sure whether the wiki talk page is a good place for that since there are not much people watching the page yet.

Discostu36:
> And I'm not sure whether the wiki talk page is a good place for that since there are not much people watching the page yet.

I'd suggest moving the discussion to the community forum instead of the Wiki talk page, as I think that will reach a bigger part of the community.

PR #4772, opened 2023-01-29
Display name in undo dialog

### LINKED ISSUE #4767: Display name of amenity in Undo menu
**Use case**

Sometimes I want to undo a quest (from multi-level Undo menu), but if there have been several quests, I am unsure which quest refers to which object (e.g. several closely-spaced shops, cafes etc. and `opening_hours` quest), because Undo menu (as opposed to Quest itself) doesn't show amenity name/type. E.g.

small_Screenshot_20230126_180201_de westnordost streetcomplete small_Screenshot_20230126_180220_de westnordost streetcomplete

**Proposed Solution**

It would be great if the Undo entity when pressed would also contain name/type of the object on which the Quest was answered. E.g. in example above, second picture would also say _"Super Hrvoje (cafe)"_ below _"Are this opening hours still correct?"_. That would allow me to check if I'm undoing a right quest (instead of a closeby quest of same type!)

Helium314:
fix #4767 

This shows name and location label below the question in undo dialog, just like in the quest form.
I'm not sure how useful the location part is here, but there were no negative reactions on https://github.com/streetcomplete/StreetComplete/issues/4767#issuecomment-1407126969

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
@@ -11,6 +11,7 @@ import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import android.widget.TextView
 import androidx.appcompat.app.AlertDialog
+import de.westnordost.osmfeatures.FeatureDictionary
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.edithistory.Edit
 import de.westnordost.streetcomplete.data.edithistory.EditHistoryController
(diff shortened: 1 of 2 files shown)


westnordost:
Code looks perfect. Did you test it?

westnordost (approved):

westnordost:
I'll test it myself in a moment

Helium314:
I did test it, and it it works, but I just found that for very long questions there is not enough space.

Helium314:
Not sure how to handle this, maybe put icon and text in a sub-layout? I think neither text nor icon can be used for top/bottom constraints, as text may be much smaller or much larger than icon.

westnordost:
What do you mean, it doesn't fit? Doesn't the dialog just get larger then?

Helium314:
The `divider` is constrained to bottom of `icon`, and the icon top is constrained to bottom of `createdTimeText`. The `titleText` does not matter at all here.

You can try it in Android Studio by entering a very long `titleText` or `titleHintText`. Actually this could be an issue even without the added `titleHintText` in some languages with long texts.

Helium314:
Looks like an [`Barrier`](https://stackoverflow.com/questions/47120035/constrainlayout-how-to-constrain-a-views-top-to-the-bottom-of-the-view-of-larg) does what I want, I'll try to get it working.

Helium314:
The barrier below is straightforward, but I didn't get the barrier below `createdTimeText` to work...

westnordost:
It's okay, I'll do it

PR #4779, opened 2023-01-30
add missing …

matkoniecz:
as I understand we should have … there as it appears in other answers menu and will bring new window (not sure about value of that extra … but would be nice to do it for consistency)

as PR as I am not sure should it be synchronised somehow with releases


HolgerJeromin:
ref https://uxdesign.cc/dot-dot-dot-7ce6170bfc7f#c9df

PR #4781, opened 2023-01-30
Set default theme to system on Android 11+

### LINKED ISSUE #4710: Use "System Default" theme by default
Please use the "System Default" theme by default.

Helium314:
fixes #4710 

This sets the default theme to "system" on Android 11+, and keeps it on "auto" for older versions.
Tested on Android 9 using the version in this PR, and a test version that set theme to "system" on Android 9+ instead of 11+.


westnordost (approved):
Cool, thank you!

Issue #4784, opened 2023-01-31
Street names are automatically title-cased in locales where they shouldn't be


balsoft:
Because of [this line](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/view/controller/AutoCorrectAbbreviationsViewController.kt#L51), whenever one is typing in a name for a street or a place, it always capitalizes the first letter of each word in the street name. While this is great for many languages where uppercase letters are used at the start of each word in toponyms, this is a problem in other languages where that's not the case. For example, in Georgian language, while capital letters are included in the Unicode standard, they are not used for title casing. This makes it really annoying to input street names, as one has to edit the words afterwards to remove the capitalization.

I'm not sure what the best solution for this would be. Perhaps add a list of "exception" languages which are not to be titlecased, or even a simple toggle to disable the feature.

**How to Reproduce**

Start solving a "Street Name" quest somewhere in Georgia, type in street name in Georgian. The first letters are automatically capitalized, even though they shouldn't be.

Here's what it looks like:

https://user-images.githubusercontent.com/18467667/215851298-f3f8194d-74bb-4378-a82d-e9f1d5544de4.mp4

**Expected Behavior**

No tilte-casing occurs.

**Versions affected**

v50.2 (latest from F-Droid)

westnordost:
So this first needs some research in which languages (or countries?) a title case is not used. Would you care to do this research?

westnordost:
By the way, you can tap on the street it belongs to instead of writing it each time manually.

balsoft:
> By the way, you can tap on the street it belongs to instead of writing it each time manually.

Yep, I'm aware of that; However, this also happens for place names (which are quite common here, in the form of microdistricts, and here I can't use that feature with place names).

balsoft:
> So this first needs some research in which languages (or countries?) a title case is not used. Would you care to do this research?

I think a good first approximation would be to look at endonyms of countries, to see which ones are and aren't using title-casing: https://en.wikipedia.org/wiki/List_of_countries_and_dependencies_and_their_capitals_in_native_languages

matkoniecz:
I found https://bugs.ruby-lang.org/issues/14839 which is quite related

[...]

PR #4785, opened 2023-01-31
Autocorrect: use title-case instead of upper-case

balsoft:
In some languages, title-case for letters is different from upper-case. This difference is captured by the [titlecase] function in Kotlin. Use it instead of [uppercase].

Fixes https://github.com/streetcomplete/StreetComplete/issues/4784

[titlecase]: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/titlecase.html
[uppercase]: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/uppercase.html

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/view/controller/AutoCorrectAbbreviationsViewController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/view/controller/AutoCorrectAbbreviationsViewController.kt
@@ -47,7 +47,7 @@ class AutoCorrectAbbreviationsViewController(private val editText: EditText) {
             fixedReplace(s, wordStart, wordStart + lastWordBeforeCursor.length, replacement)
         } else if (lastWordBeforeCursor.length > 3) {
             val locale = abbrs.locale
-            val capital = lastWordBeforeCursor.substring(0, 1).uppercase(locale)
+            val capital = lastWordBeforeCursor.get(0).titlecase(locale)
             s.replace(wordStart, wordStart + 1, capital)
         }
     }


matkoniecz:
You may be able to trigger build at https://github.com/balsoft/StreetComplete/actions

balsoft:
Thanks! Seems like I can run the workflow, but it doesn't appear to be attached to the commit in this repo.

balsoft:
Looks like [the build succeeds](https://github.com/balsoft/StreetComplete/actions/runs/4058140607) and the resulting APK exhibits the correct behavior:

https://user-images.githubusercontent.com/18467667/215881358-10a148a6-eacd-4ad9-b790-e37b571e1875.mp4

app/src/main/java/de/westnordost/streetcomplete/view/controller/AutoCorrectAbbreviationsViewController.kt
```diff
@@ -47,7 +47,7 @@ class AutoCorrectAbbreviationsViewController(private val editText: EditText) {
             fixedReplace(s, wordStart, wordStart + lastWordBeforeCursor.length, replacement)
         } else if (lastWordBeforeCursor.length > 3) {
             val locale = abbrs.locale
-            val capital = lastWordBeforeCursor.substring(0, 1).uppercase(locale)
+            val capital = lastWordBeforeCursor.get(0).titlecase(locale)
             s.replace(wordStart, wordStart + 1, capital)
```
westnordost, line 51: Uh wait how can this even compile? `String.get(0)` returns a `Char`, which is fine because `titlecase` is a method of `Char`. But `s` is a `Editable` and the `replace` method expects a `CharSequence` (implemented by `String`), not a `Char`:

https://developer.android.com/reference/android/text/Editable#replace(int,%20int,%20java.lang.CharSequence)

westnordost: Oh, nevermind. `titlecase` returns a `String`. That's fine then. Thank you!

balsoft: `titlecase` returns a `String`, since in some languages title-casing a single letter may result in multiple letters.

PR #4793, opened 2023-02-05
New quest: asking if place has backup electricity generator

### LINKED ISSUE #4789: Ask if places have a backup generator
### General
Affected tag(s) to be modified/added: [backup_generator](https://wiki.openstreetmap.org/wiki/Key:backup_generator) (disclaimer: I created this page to document an already in use tag)
Question asked: **Does this place have a backup generator?**

### Checklist
It is very useful to know if a place will still operate during load shedding [1] and some brands even advertise which locations have a generator on their website: [example](https://romanspizza.co.za/find-a-store).
Disabled by default as it requires entering a place at a particular time, similar to air conditioning quest.

### Ideas for implementation

**Element selection:** 
All shop-like elements.

**Metadata needed:**
Countries where backup generators are common, at least South Africa and by the looks of where the tag is used already, Philippines and Indonesia. 

**Proposed UI:**
Yes/No question, much like the air conditioning quest.

[1] South Africa has recently and continuously been having load shedding where power is off for hours at a time on a [schedule in different areas](https://www.eskom.co.za/distribution/customer-service/outages/municipal-loadshedding-schedules/), recent article: ["South Africa’s load shedding is practically permanent now"](https://www.citizen.co.za/news/south-africa/load-shedding/load-shedding-schedule-06-february-2023/)

michalgwo:
Fixes #4789

I found it very useful for countries like South Africa or Lebanon. I'm not sure about restricting it to specific countries, on one hand, backup generators don't exist in some countries, but on the other, a situation can change quickly, and countries that didn't have power outages in the past, can fall into this at some point in time, so maybe they can just be disabled by default, just like the question about air conditioning, what do you think?

Screenshot_1675639368 Screenshot_1675639372 Screenshot_1675639393 Screenshot_1675639214

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
@@ -13,16 +13,15 @@ import de.westnordost.streetcomplete.quests.YesNoQuestForm
 import de.westnordost.streetcomplete.util.ktx.toYesNo
 
 class AddBackupGenerator : OsmFilterQuestType<Boolean>() {
-    val amenities = listOf(
+    private val amenities = listOf(
         "bar", "cafe", "fast_food", "food_court", "pub", "restaurant", "cinema",
-        "nightclub", "planetarium", "theatre", "internet_cafe", "car_wash", "telephone",
-        "vending_machine"
+        "nightclub", "planetarium", "theatre", "internet_cafe", "car_wash", "vending_machine"
     )
-    val tourisms = listOf(
-        "zoo", "aquarium", "theme_park", "hotel", "hostel", "motel", "guest_house",
-        "apartment", "camp_site", "museum", "gallery"
+    private val tourisms = listOf(
+        "theme_park", "hotel", "hostel", "motel", "guest_house", "apartment", "camp_site",
+        "museum", "gallery"
     )
-    val leisures = listOf(
+    private val leisures = listOf(
         "adult_gaming_centre", "amusement_arcade", "bowling_alley", "escape_game",
         "miniature_golf", "sauna", "trampoline_park", "tanning_salon"
     )
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
@@ -1,0 +1,58 @@
+package de.westnordost.streetcomplete.quests.backup_generator
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddBackupGenerator : OsmFilterQuestType<Boolean>() {
+    val amenities = listOf(
+        "bar", "cafe", "fast_food", "food_court", "pub", "restaurant", "cinema",
+        "nightclub", "planetarium", "theatre", "internet_cafe", "car_wash", "telephone",
+        "vending_machine"
+    )
+    val tourisms = listOf(
+        "zoo", "aquarium", "theme_park", "hotel", "hostel", "motel", "guest_house",
+        "apartment", "camp_site", "museum", "gallery"
+    )
+    val leisures = listOf(
+        "adult_gaming_centre", "amusement_arcade", "bowling_alley", "escape_game",
+        "miniature_golf", "sauna", "trampoline_park", "tanning_salon"
+    )
+
+    override val elementFilter = """
+        nodes, ways with
+        (
+              (shop and shop !~ no|vacant)
+              or amenity ~ ${amenities.joinToString("|")}
+              or tourism ~ ${tourisms.joinToString("|")}
+              or leisure ~ ${leisures.joinToString("|")}
+        )
+        and access !~ private|no
+        and !backup_generator
+    """
+    override val changesetComment = "Survey availability of backup generator"
+    override val wikiLink = "Key:backup_generator"
+    override val icon = R.drawable.ic_quest_backup_generator
+    override val isReplaceShopEnabled = true
+    override val achievements = listOf(CITIZEN)
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside_regional_warning
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_backupGenerator_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter(IS_SHOP_OR_DISUSED_SHOP_EXPRESSION)
+
+    override fun createForm() = YesNoQuestForm()
+
+    override fun applyAnswerTo(answer: Boolean, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        tags["backup_generator"] = answer.toYesNo()
+    }
+}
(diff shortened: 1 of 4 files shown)


matkoniecz:
I think that making it generally available is fine? In case of users being nuisance and pointlessly answering it in places where this adds no useful info it can become restricted.

app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
```diff
@@ -0,0 +1,58 @@
...
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddBackupGenerator : OsmFilterQuestType<Boolean>() {
+    val amenities = listOf(
+        "bar", "cafe", "fast_food", "food_court", "pub", "restaurant", "cinema",
+        "nightclub", "planetarium", "theatre", "internet_cafe", "car_wash", "telephone",
+        "vending_machine"
+    )
+    val tourisms = listOf(
+        "zoo", "aquarium", "theme_park", "hotel", "hostel", "motel", "guest_house",
```
matkoniecz, line 22: what should be answered for say zoo that has generators to keep vulnerable animals alive but would be closed due power outage?

app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
```diff
@@ -0,0 +1,58 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddBackupGenerator : OsmFilterQuestType<Boolean>() {
+    val amenities = listOf(
```
matkoniecz, line 16: telephone seems surprising to me here

app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
michalgwo: You are right, it is the same with the aquarium

app/src/main/java/de/westnordost/streetcomplete/quests/backup_generator/AddBackupGenerator.kt
michalgwo: I copied filters from another quest and must have missed that one while deleting not related ones :sweat_smile: It's removed

westnordost:
I see several general issues with this quest:
That list of places to ask for will likely have to be extended over the time and/or may not be even be kind of complete now. Have a look at e.g. `AddOpeningHours`. There are **a lot** of different places that **could** be added. But ultimately it comes down to weighing if a business can operate without power or not(?). Or well, many places do not need power to operate. So how would that even be surveyable in such cases? It's not simply looking which stores stay open but... you have to ask them? See if their AC is on (if they have one, that is)? Anyway, what I am saying is that

1. it is really hard to survey. Entering a store/place may not be enough. It can also change any time, depending if the shop owner thinks it is worth powering up the generator this time around; or if the generator works properly; or **what** is powered by the generator and what is not. (I could tell you an anecdote about a certain hostel in ဘားအံ, but instead I'll just link [a blog article of mine from 2014](https://westnordost.de/blog/thailand/122/solar-powered)). To the point where I do not think it is worth it, at least for SC mappers.

2. it is really hard to decide which places to include and which not. Theoretically, any place could make use of electricity (for light), but then again, in places where the electrical power grid is not reliable, everyone has some kind of setup of car batteries + solar or something around to at least have light. That does not mean that a laundromat would power their washing machines with that setup. So, this means that (if the quest is actually used by people) there will be a constant forth and back on which places to include and which not, which just adds up to the maintenance effort

3. the other "you have to go inside to check / regionally may not make sense" quests were for example about cuisine/diet. Halal **may** make sense outside of Arab countries, e.g. in Arab ghettos in other places in the world. I don't see that with electricity generators. So, to restrain that quest only to a few confirmed countries is the better choice here. (Actually, it will not be a few, given extensive research, I think many countries in Africa would fall in that category. However, that research also adds up to the effort).

4. the (on-site) mapper communities in such countries are really really small or non-existent, which is why I doubt it is worth investing time there.

All in all, I would not even recommend this quest for Myanmar where I have first-hand experience for the reasons stated above.

michalgwo:
@westnordost you have good points, but I still think it's worth to do it.

> 1. it is really hard to survey. Entering a store/place may not be enough. It can also change any time, depending if the shop owner thinks it is worth powering up the generator this time around; or if the generator works properly; or **what** is powered by the generator and what is not. (I could tell you an anecdote about a certain hostel in ဘားအံ, but instead I'll just link [a blog article of mine from 2014](https://westnordost.de/blog/thailand/122/solar-powered)). To the point where I do not think it is worth it, at least for SC mappers.

This could be the biggest challenge here, but generally just knowing that place has a backup generator and could use it when needed is very useful information.


> 2. it is really hard to decide which places to include and which not. Theoretically, any place could make use of electricity (for light), but then again, in places where the electrical power grid is not reliable, everyone has some kind of setup of car batteries + solar or something around to at least have light. That does not mean that a laundromat would power their washing machines with that setup. So, this means that (if the quest is actually used by people) there will be a constant forth and back on which places to include and which not, which just adds up to the maintenance effort

We can use it for places that definitely need electricity to operate, and there is a high probability that they will operate during a power cut. Maybe @arrival-spring or anyone living in a place with frequent power shortages can help with that?


> 3. the other "you have to go inside to check / regionally may not make sense" quests were for example about cuisine/diet. Halal **may** make sense outside of Arab countries, e.g. in Arab ghettos in other places in the world. I don't see that with electricity generators. So, to restrain that quest only to a few confirmed countries is the better choice here. (Actually, it will not be a few, given extensive research, I think many countries in Africa would fall in that category. However, that research also adds up to the effort).

I think we can just exclude high developed countries.


> 4. the (on-site) mapper communities in such countries are really really small or non-existent, which is why I doubt it is worth investing time there.

Just one mapper per city is enough to make a huge difference.

westnordost:
> We can use it for places that definitely need electricity to operate

Which would these be?

michalgwo:
> Which would these be?

Cinemas, nightclubs, theatres (they need light), car washes, vending machines, theme/amusement parks, gaming centers, amusement arcades, bowling alleys (technically you can every time you throw a ball, go and pick it up, but I doubt a lot of bowling alleys are adapted to that), escape games (the need light), saunas, banks - all of these definitely need electricity to operate.

But maybe I exaggerated a bit with "definitely need electricity", I think we should also include places that highly depend on power. Like supermarkets or restaurants, theoretically, you can buy something without electricity, but they need it to keep food fresh in the fridges/freezers, cash registers to print receipts, to take card payments, restaurants additionally need electricity for cooking - if power outages are frequent, supermarkets and restaurants can't operate without a generator. 

Hotels also are worth including, because it will surely have a huge impact on the decision whether to book this hotel or not.

westnordost:
- Car washes only need electricity if they are automated
- nightclubs, theatres,... can be open air
- Some ticket machine (types) do not need external power supply (e.g. parking ticket machines in my area do not)
- I've seen amusement parks in India that were completely hand-operated ... I even have a short video from that still, let's see.... didn't find the one I was looking for, but there is another. Also, amusement parks are generally too large that it makes sense to look for a (single) generator there

https://user-images.githubusercontent.com/4661658/217904542-8bd3d6ec-50fe-4990-bcc7-9cea1dcd38df.mp4

- only these American-style automated bowling alleys need electricity
- the heat for saunas is can be generated without electricity
- escape games do not necessarily need light if it is bright enough otherwise
- banks also do not necessarily need electricity, otherwise there wouldn't have been any banks before its invention
- (also, I notice that if there is no power, a generator in an internet café wouldn't do much good because there is no internet)

So, the list of places that **definitely** need electricity is very small. Places that "most probably" depend on power... well... in places where there is no reliable electricity network or otherwise 3rd world countries, shops and places will simply not be that dependent on electricity: Mechanical cash registers (still) exist, so do just some drawer with money and coins in it. Even for card payment, I remember that during the 56k modem times and even much later still, the machine would not even query your secret pin but instead you were prompted by the cashier to provide a signature on the receipt - AFAIK this was done every time an connection with the bank could not be established. And yeah, electricity for cooking... no, they are more likely to use wood, coal, gas or something.

westnordost:
In any case, I noticed that `backup_generator` is not (even) included in the iD presets. Maybe this is a better place (to start) because iD doesn't have such stringent "quest guideline" rules as StreetComplete (necessarily) has.

valerio-bozzolan:
(Let me say that this discussion is amazing.)

Maybe the real question should be "Does it work under a power outage?" yes / no but indeed it's probably something to be discussed on OSM first.

westnordost:
So anyway. I'll close this as will not fix. Sorry that you put all the effort into it already.

I really do get the intention of this, and on the surface it may make sense to want to know which businesses will still operate during rolling blackouts, but whether or not they are open is somewhat independent of whether they have a generator standing around somewhere. Not only could they draw from other sources (batteries, solar...) or not actually be dependent on electricity (flashlights, candles, petrolium lamps, torches ...). 

The other reasons I mentioned in https://github.com/streetcomplete/StreetComplete/pull/4793#issuecomment-1419174097 . In particular, this is almost impossible to survey since one can't expect a generator to be in plain sight like in this jewelry shop in Mandalay and also only really surveyable **during** a blackout.

 

To be honest, in regions where the electricity grid is that unstable, people have generally more pressing things to do than documenting which shops operate continuously despite this.

PR #4797, opened 2023-02-07
Allow viewing photos after attaching to note

Helium314:
See discussion in https://github.com/Helium314/SCEE/issues/392. The photo is shown in a dialog with transparent background.
I hope the delete button is now big enough.

screen

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -187,6 +187,9 @@ dependencies {
 
     // opening hours parser
     implementation("ch.poole:OpeningHoursParser:0.27.0")
+
+    // image view that allows zoom and pan
+    implementation("com.github.chrisbanes:PhotoView:2.3.0")
 }
 
 /** Localizations that should be pulled from POEditor */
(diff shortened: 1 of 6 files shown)


westnordost:
Nice! Hmm, the delete button is indeed a bit big now (also, scaling it will make it blurry, the actual graphic either have to be made larger or at least the view displaying it). I propose 32dp size for the delete button and 72dp or so for the image thumbnail.

How come the third picture is cut off that way? I wonder if there is a way it can be made clear that this scrolls vertically. Maybe always show the scroll bar? Or show some kind of gradient from background-color to transparent at its edges?

Helium314:
> scaling it will make it blurry

This is weird, I wonder why Android Studio can scale it without blurring (in the preview)

> How come the third picture is cut off that way?

That's normal, but previously I think 4 or 5 pictures were necessary.
Is there really anything to do? When the users see that the 3rd photo they attach is cut off, they might try to scroll.
But a scrollbar might still help. Can the scrollbar be always shown, but only if scrolling is actually possible?

Helium314:
Ok, the scrollbar shows up for a few seconds once it becomes necessary. That's hopefully enough.

westnordost:
> This is weird, I wonder why Android Studio can scale it without blurring (in the preview)

Android has some optimizations built-in for vector graphics. It usually caches a bitmap variant of the vector graphic after rendering it once. The size of that bitmap depends on the defined size of the vector graphic.

westnordost:
>  But a scrollbar might still help. Can the scrollbar be always shown, but only if scrolling is actually possible?

I think yes. Should be some XML property. I'd need to have a look at the docs too

westnordost:
I made some improvements I didn't want to bother you with

PR #4798, opened 2023-02-07
Fix #4790

### LINKED ISSUE #4790: Cycleway quest applies wrong tags
Likely this is related to #4774, but this happens in in 51.0 beta1 (from GitHub)

* Open the cycleway quest for https://www.openstreetmap.org/way/43943837 (`cycleway=lane`)
* see current situation tagged as "???" (which is expected)
* either: set the available side to "none"
  * see added tags `oneway:bicycle=no`, `cycleway:left=lane`, `cycleway:right=no` (the first two are incorrect)
* or: set the available side to "advisory"
  * see added tags `oneway:bicycle=no`, `cycleway:both=lane`, `cycleway:right:lane=advisory` (the first two are incorrect)
* get the quest again, this time showing both sides
* set new side to "none"
* see tags are now correct

westnordost:
@Helium314 are you interested in reviewing this?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayCreator.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayCreator.kt
@@ -71,7 +71,6 @@ private fun expandBareTags(tags: Tags, isLeftHandTraffic: Boolean) {
     }
     tags.remove("cycleway")
     tags.expandBareTagIntoSide("cycleway", "lane", side)
-    tags.expandBareTagIntoSide("cycleway", "lane", side)
     tags.expandBareTagIntoSide("cycleway", "oneway", side)
     tags.expandBareTagIntoSide("cycleway", "segregated", side)
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayCreator.kt
```diff
@@ -49,6 +52,37 @@ fun LeftAndRightCycleway.applyTo(tags: Tags, isLeftHandTraffic: Boolean) {
...
+        "both"
+    }
+    if (!tags.containsKey("cycleway:$side")) {
+        tags["cycleway:$side"] = cycleway
+            .removePrefix("opposite_") // opposite_track -> track etc.
+            .replaceFirst("opposite", "no") // opposite -> no
+    }
+    tags.remove("cycleway")
+    tags.expandBareTagIntoSide("cycleway", "lane", side)
+    tags.expandBareTagIntoSide("cycleway", "lane", side)
```
Helium314, line 74: This line is duplicate, and the second line will do nothing. Did you mean `tags.expandBareTagIntoSide("cycleway", "", side)`?

Helium314:
Tested a bit on oneways, and couldn't find any weird results.
Codelooks fine, except for the place I indicated. Though I only roughly looked over the tests.

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayCreator.kt
westnordost: Whoops, thank you!

westnordost:
Thank you, @Helium314 !

PR #4800, opened 2023-02-09
Vacant office fixes

dbdean:
Changes to make SC treat `office`=`vacant` similar to `shop`=`vacant`. While there aren't that many worldwide ([813](https://taginfo.openstreetmap.org/tags/office=vacant)), this stops vacant offices coming up in the name quest or any of the other shop-related quests.

This also ensures that vacant offices are resurveyed on occasion to check if they are still vacant.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
@@ -125,7 +125,7 @@ fun isShopExpressionFragment(prefix: String? = null): String {
     val p = if (prefix != null) "$prefix:" else ""
     return ("""(
         ${p}shop and ${p}shop !~ no|vacant|mall
-        or ${p}office and ${p}office != no|vacant
+        or ${p}office and ${p}office !~ no|vacant
         or ${p}healthcare and healthcare != hospital
         or ${p}craft
         or ${p}tourism = information and ${p}information = office
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
@@ -267,4 +267,5 @@ val IS_DISUSED_SHOP_EXPRESSION = """
     nodes, ways, relations with
       ${isShopExpressionFragment("disused")}
       or shop = vacant
+      or office = vacant
 """.toElementFilterExpression()
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
@@ -125,7 +125,7 @@ fun isShopExpressionFragment(prefix: String? = null): String {
     val p = if (prefix != null) "$prefix:" else ""
     return ("""(
         ${p}shop and ${p}shop !~ no|vacant|mall
-        or ${p}office and ${p}office != vacant
+        or ${p}office and ${p}office != no|vacant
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/osm/Shop.kt
```diff
@@ -125,7 +125,7 @@ fun isShopExpressionFragment(prefix: String? = null): String {
     val p = if (prefix != null) "$prefix:" else ""
     return ("""(
         ${p}shop and ${p}shop !~ no|vacant|mall
-        or ${p}office and ${p}office != vacant
+        or ${p}office and ${p}office !~ no|vacant
```
dbdean, line 128: office = no is almost non-existent, but seemed best to match the shop usage

app/src/main/java/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -22,8 +22,8 @@ class AddPlaceName(
         nodes, ways with
         (
           shop and shop !~ no|vacant
+          or office and office !~ no|vacant
```
dbdean, line 25: office = no is almost non-existent, but seemed best to match the shop usage

westnordost:
I think this makes sense. What do @Helium314 and @matkoniecz think?

matkoniecz:
Ideally `office=no` and `officer=vacant` would be not used but workaround for bad tagging is fine to have to avoid confusing SC mappers.

Helium314:
In general I agree, though I'm not sure about asking `CheckShopType` for `office=vacant`.
Looking at actual uses, sometimes it's used like [`shop=vacant`](https://www.openstreetmap.org/node/4254833206), sometimes it's used in combination with [`building=office` (history)](https://www.openstreetmap.org/way/62043817/history), which likely will result in people leaving notes. But even when used like `shop=vacant`, it may have been a company office [somewhere in the building](https://www.openstreetmap.org/node/3199946665).

JLZIMMERMANN:
This way of thinking is really welcome :
shop=vacant
office=vacant
both need to be considerated.

westnordost:
> Looking at actual uses, sometimes it's used like [shop=vacant](https://www.openstreetmap.org/node/4254833206), sometimes it's used in combination with [building=office (history)](https://www.openstreetmap.org/way/62043817/history), which likely will result in people leaving notes.

I wonder if `office=vacant` somewhere in the building is an acceptable tagging though? It is really hard to verify whether the office space on the left side of the 5th floor of an office building is still vacant or what is even in there right now except if that particular office building has a very clear sign at the entrance.

But if it is acceptable and under the assumption that it is clearly signed at the entrance or in the lobby, then it would also be okay to ask about it, wouldn't it?

dbdean:
Is there an access tag that can be used to say employees only? If that tag was on the building or the office tag, then SC could ignore it to avoid such spammyness.

dbdean:
Access = private would work I think. We could add an alternate answer 'It is not accessible to the public' -> tag `access`=`private`.

matkoniecz:
> Is there an access tag that can be used to say employees only?

`access=private`? In general for many `office=` it is hard to answer with OSM data is it internal company office only for employees and invited guests or is it something accepting walk-ins.

westnordost:
Hm well, I think given the low volume of office=vacant, there'll not be so many places where `office=vacant` is used in scenarios as @Helium314 described. Also, the the volume is so low that it is also probably not worth that much time to ponder this. So let's merge it.

PR #4809, opened 2023-02-13
healthcare ways should be treated as areas

dbdean:
This PR ensures that ways tagged with `healthcare`=`*` are considered to be areas, not ways. Before this PR, an healthcare=audiologist would show up as the boundary of the shop:

image

With this PR, it correctly shows up as an area:

image


westnordost:
Yeah, true, makes sense!

PR #4820, opened 2023-02-17
Implemented duplicating presets functionality

### LINKED ISSUE #4609: Allow cloning quest presets
**Use cases**
1. I want to have one (or multiple) custom presets that have few quests disabled that I never want to answer, and use that preset as a base for creating new presets.
2. I have multiple quest presets that in a way all focus on one type of quest, but are only different in the level of detail of enabled quests. It would be nice to be able to copy one present so that I only need to enable/disable a few quests instead of going through all quests all over again.

**Proposed Solution**
1. In the "Manage presets" screen, add an action to the per-preset three dot menu to clone a preset
2. or: when creating a new preset, always copy the settings of the currently selected preset

Additionally, I think it would make sense to automatically select a newly created preset, assuming that the whole point of creating a new preset is to use or modify it.

michalgwo:
Fixes #4609

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
@@ -21,6 +21,10 @@ class QuestTypeOrderController(
             }
             override fun onAddedQuestPreset(preset: QuestPreset) {}
             override fun onRenamedQuestPreset(preset: QuestPreset) {}
+            override fun onDuplicatedQuestPreset(preset: QuestPreset, duplicatedPresetId: Long) {
+                val orders = getOrders(duplicatedPresetId)
+                setOrders(orders, preset.id)
+            }
             override fun onDeletedQuestPreset(presetId: Long) {
                 clear(presetId)
             }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
@@ -21,14 +21,15 @@ class QuestTypeOrderController(
             }
             override fun onAddedQuestPreset(preset: QuestPreset) {}
             override fun onRenamedQuestPreset(preset: QuestPreset) {}
-            override fun onDuplicatedQuestPreset(preset: QuestPreset, duplicatedPresetId: Long) {
-                val orders = getOrders(duplicatedPresetId)
-                setOrders(orders, preset.id)
-            }
             override fun onDeletedQuestPreset(presetId: Long) {
                 clear(presetId)
             }
         })
+    }
+
+    fun copyOrders(presetId: Long, newPresetId: Long) {
+        val orders = getOrders(presetId)
+        setOrders(orders, newPresetId)
     }
 
     fun setOrders(orderItems: List<Pair<QuestType, QuestType>>, presetId: Long? = null) {
--- a/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
@@ -28,8 +28,8 @@ class QuestTypeOrderController(
     }
 
     fun copyOrders(presetId: Long, newPresetId: Long) {
-        val orders = getOrders(presetId)
-        setOrders(orders, newPresetId)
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestPresetsAdapter.kt
```diff
@@ -115,6 +120,10 @@ class QuestPresetsAdapter(
                 renameItem.setOnMenuItemClickListener { onClickRenamePreset(preset); true }
             }
 
+            val duplicateItem = popup.menu.add(R.string.quest_presets_duplicate)
+            duplicateItem.setIcon(R.drawable.ic_add_24dp)
```
FloEdelmann, line 124: I'd rather use the `copy` icon: https://pictogrammers.com/library/mdi/icon/content-copy/

FloEdelmann (commented):
Looks really good!

app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestPresetsAdapter.kt
michalgwo: Fixed ;)

westnordost (commented):
The `QuestPresetsController` only takes care of naming the presets. It knows nothing of the quest type orders or enabled quests. Hence, I think it is not the right position to introduce a `duplicate` method in that class. From `QuestPresetsController`s point of view, it is just another "add".

Instead, the logic should lie either in the adapter or in a very tiny `*Controller` class that I guess is only used for duplicating. Since it is really just a few lines of code, I propose to do it in the adapter, like this, but a new controller class would be fine too:

In `QuestPresetsAdapter::duplicateQuestPreset`, have this code within the IO scope (or with better names if you :

```kotlin
val newPresetId = questPresetsController.add(name)
questTypeOrderController.copyOrders(presetId, newPresetId)
visibleQuestTypeController.copyVisibilities(presetId, newPresetId)
```
(or use `questTypeOrderController.setOrders(questTypeOrderController.getOrders(presetId), newPresetId)` etc. I have no strong opinion about this)

Should you implement it via a new controller class, I think it would be cleaner if the adapter class would **only** use that controller class and the controller class would in turn manage access to the other controller classes. I.e. 
`class QuestPresetsManagementController(...) { fun add(name:String) { questPresetsController.add(name) } ... }` etc.

michalgwo:
Thank you for the review, I implemented it in the adapter.

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/QuestTypeOrderController.kt
```diff
@@ -27,6 +27,11 @@ class QuestTypeOrderController(
         })
     }
 
+    fun copyOrders(presetId: Long, newPresetId: Long) {
+        val orders = getOrders(presetId)
+        setOrders(orders, newPresetId)
```
westnordost, line 32:
```suggestion
        questTypeOrderDao.setAll(newPresetId, questTypeOrderDao.getAll(presetId))
        if (newPresetId == selectedPresetId) onQuestTypeOrderChanged()
```

I think I prefer this because it does not take the detour of getting the quest type classes.

app/src/main/java/de/westnordost/streetcomplete/data/visiblequests/VisibleQuestTypeController.kt
```diff
@@ -42,6 +42,12 @@ class VisibleQuestTypeController(
 
     private val selectedPresetId: Long get() = questPresetsSource.selectedId
 
+    fun copyVisibilities(presetId: Long, newPresetId: Long) {
+        val duplicatedQuestTypes = getVisible(presetId)
+        val questTypes = questTypeRegistry.associateWith { it in duplicatedQuestTypes }
+        setVisibilities(questTypes, newPresetId)
```
westnordost, line 48:
```suggestion
        visibleQuestTypeDao.putAll(newPresetId, visibleQuestTypeDao.getAll(presetId))
        if (newPresetId == selectedPresetId) onQuestTypeVisibilitiesChanged()
```

I think I'd prefer this. It is more precise (`getVisible` does not return exactly what is set through `visibleQuestTypeDao.getAll` returns) and does not take the detour of getting the quest type classes.

westnordost (approved):
No complaints now, thank you! 

I'll mark this as approved but not merge it yet because it will be included in the next feature release (and not in the next bugfix update).

Also, on merging, I'll want to add two tests for the added functions in the controller classes.

PR #4843, opened 2023-02-24
Check shop existence

matkoniecz:
fixes https://github.com/streetcomplete/StreetComplete/discussions/4232

Seems OK and plan on merging if noone will have feedback about this, but I want to give chance to review. No new text is being used here.

Why as a separate quest? Check existence quest is before opening hours, wheelchair quest and so on.

This one should be after opening hours etc, otherwise they would not be resurveyed in some cases. For example if check existence AND opening hours quests are enabled AND shop has `opening_hours` tagged AND shop node was not edited for 13 years, then opening hours resurvey quest should be asked, not check existence quest.

Otherwise user would confirm existence, what would update object and opening hours resurvey would not be asked.

And why it is useful at all if opening hours already acts as a de facto existence checker? OH is quite time-consuming and some people could disable it without enabling wheelchair quest or similar.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
@@ -27,7 +27,6 @@ class CheckShopExistence : OsmFilterQuestType<Unit>() {
     override val elementFilter = """
         nodes with (
           (${isShopExpressionFragment()})
-          or (${isShopExpressionFragment("disused")}
              and !man_made
              and !historic
              and !military
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
@@ -36,7 +36,6 @@ class CheckShopExistence : OsmFilterQuestType<Unit>() {
              and !railway
              and amenity !~ clinic|library|music_school|post_office
              and healthcare !~ clinic
-          )
         ) and (
           older today -2 years
           or ${LAST_CHECK_DATE_KEYS.joinToString(" or ") { "$it < today -2 years" }}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
@@ -30,8 +30,6 @@ class CheckShopExistence : OsmFilterQuestType<Unit>() {
              and !attraction
              and !aeroway
              and !railway
-             and amenity !~ clinic|library|music_school|post_office
-             and healthcare !~ clinic
         ) and (
           older today -2 years
           or ${LAST_CHECK_DATE_KEYS.joinToString(" or ") { "$it < today -2 years" }}
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
@@ -1,21 +1,17 @@
 package de.westnordost.streetcomplete.quests.shop_type
 
 import de.westnordost.streetcomplete.R
-import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
 import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
 import de.westnordost.streetcomplete.data.osm.mapdata.Element
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
 import de.westnordost.streetcomplete.data.osm.mapdata.filter
-import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
 import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
-import de.westnordost.streetcomplete.osm.IS_SHOP_EXPRESSION
 import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistenceForm.kt
```diff
@@ -0,0 +1,23 @@
...
+                },
+                AnswerItem(R.string.quest_generic_hasFeature_yes) {
+                    applyAnswer(Unit)
+                }
+            )
+        } else {
+            emptyList()
+        }
+
+    private var isAskingAboutExistence: Boolean = true
```
FloEdelmann, line 22: What is this variable used for? I think it's never overwritten, so it could be removed, right?

matkoniecz: oh right, that is leftover from the initial design

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistenceForm.kt
```diff
@@ -0,0 +1,17 @@
...
+class CheckShopExistenceForm : AbstractOsmQuestForm<Unit>() {
+    override val buttonPanelAnswers get() =
+        listOf(
+            AnswerItem(R.string.quest_generic_hasFeature_no) {
+                replaceShop()
+            },
+            AnswerItem(R.string.quest_generic_hasFeature_yes) {
+                applyAnswer(Unit)
+            }
+        )
```
FloEdelmann, line 16: This can be simplified even more: No need for a getter, and reformatting to single lines to improve readability (this is subjective of course).

```suggestion
    override val buttonPanelAnswers = listOf(
        AnswerItem(R.string.quest_generic_hasFeature_no) { replaceShop() },
        AnswerItem(R.string.quest_generic_hasFeature_yes) { applyAnswer(Unit) },
    )
```

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
```diff
@@ -0,0 +1,60 @@
...
+class CheckShopExistence : OsmFilterQuestType<Unit>() {
+
+    // opening hours quest acts as a de facto checker of shop existence, but some people disabled it.
+    // separate from CheckExistence as very old shop with opening hours should show
+    // opening hours resurvey quest rather than this one (which would cause edit date to be changed
+    // and silence all resurvey quests)
+
+    override val elementFilter = """
+        nodes with (
+          (${isShopExpressionFragment()})
```
westnordost, line 29: Are you sure that the extra `(...)` are necessary here? In the code it looks like that function already adds `(...)`

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
```diff
@@ -0,0 +1,60 @@
...
+             and !man_made
+             and !historic
+             and !military
+             and !power
+             and !attraction
+             and !aeroway
+             and !railway
+             and amenity !~ clinic|library|music_school|post_office
+             and healthcare !~ clinic
```
westnordost, line 38: Why not?

westnordost (commented):
Makes sense! Useful for people who disabled the opening hours quest and adds very little code to maintain.

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
matkoniecz: rechecking these every two years seems a bit too much, but that may be assuming too stronly that my area is too typical and this kind of places can be far more unstable elsewhere

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
matkoniecz: added for safety but actually seems not needed - removed

matkoniecz:
> adds very little code to maintain

Yeah, added for chance to implement something not growing into megaproject :)

app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistence.kt
westnordost: But it is just a simple yes/no question. IMO it's okay and not worth the extra brainpower to try to maintain a balance here. (After all, if you keep the exclusion here, the next thing I or someone else would request is that the existence of these would also be asked but less frequently.)

matkoniecz: I was rather thinking about risk of someone making mistake and answering incorrectly.

I thought about asking them with longer timeout and I can add this.

westnordost: How would asking with longer timeouts solve the issue you note? Such larger facilities seem more difficult to miss, too. So I do not really see this issue.

matkoniecz: Some `amenity=clinic` are not very prominently signed (I encountered them during testing and were initial ones added, that is how I ended asking you about this hospitals and doing https://github.com/streetcomplete/StreetComplete/commit/fd0a1c3c0853fc1caf3e96e35d71f971f9ba3c69 )

westnordost: Well, but still, asking less often or even not asking at all in this quest is not going to solve this. After all, the opening hours of clinics are also asked - every year. And most users will have the opening hours quest activated.

matkoniecz: OK, I removed special handling of major objects.

westnordost:
I'll approve this but not merge it yet because it is a new feature (albeit it does not require new strings).

westnordost (approved):

PR #4846, opened 2023-02-25
Check location time difference using elapsedRealtime

### LINKED ISSUE #4652: GPS position error
Hello,
I am new on StreetComplete, found it easy and fun to use.
But I found a difference between my real pisition and the GPS position dusplayed on the app.
This is the only app on my smartphone with such an issue...and I use quite a large number of map apps, depending on my needs.
Thanks for your help....this issue make it difficult to use the app.
Of course I have selected the "exact position" autorisation.

Thanks,

HerveB29

Helium314:
should fix #4652

This is the most (I think) minimal fix for GPS locations having a wrong `location.time`.
If everything is working correctly, it shouldn't matter whether `elapsedRealtimeNanos` or `time` is used, but `elapsedRealtimeNanos` shouldn't have issues with system and GPS clock being slightly different.

A short test / comparison on my S4 mini plus:
`gpsLocation.time` is about 1.6 s ahead (!) of `nowAsEpochMilliseconds()` when the phone was running a few days, and about 1.45 s ahead after a restart of the phone.
`gpsLocation.elapsedRealtimeNanos` is about 5-20 ms behind `SystemClock.elapsedRealtimeNanos()`, and this time difference doesn't change with restart.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/util/location/FineLocationManager.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/util/location/FineLocationManager.kt
@@ -86,7 +86,7 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
 
 // Based on https://web.archive.org/web/20180424190538/https://developer.android.com/guide/topics/location/strategies.html#BestEstimate
 
-private const val TWO_MINUTES = 1000L * 60 * 2
+private const val TWO_MINUTES = 1000L * 1000 * 1000 * 60 * 2
 
 /** Determines whether this Location reading is better than the previous Location fix */
 private fun Location.isBetterThan(previous: Location?): Boolean {
@@ -98,7 +98,7 @@ private fun Location.isBetterThan(previous: Location?): Boolean {
     if (previous == null) return true
 
     // Check whether the new location fix is newer or older
-    val timeDelta = this.time - previous.time
+    val timeDelta = this.elapsedRealtimeNanos - previous.elapsedRealtimeNanos
     val isMuchNewer = timeDelta > TWO_MINUTES
     val isMuchOlder = timeDelta < -TWO_MINUTES
     val isNewer = timeDelta > 0L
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/util/location/FineLocationManager.kt
```diff
@@ -86,7 +86,7 @@ class FineLocationManager(context: Context, locationUpdateCallback: (Location) -
 
 // Based on https://web.archive.org/web/20180424190538/https://developer.android.com/guide/topics/location/strategies.html#BestEstimate
 
-private const val TWO_MINUTES = 1000L * 60 * 2
+private const val TWO_MINUTES = 1000L * 1000 * 1000 * 60 * 2
```
westnordost, line 89: uh what? This ain't two minutes :-D

westnordost (commented):
It would be good if you added compact comments that explained why it is done.

app/src/main/java/de/westnordost/streetcomplete/util/location/FineLocationManager.kt
Helium314: It looks like a lot, but actually it is...

Before Android 13 location can only return `elapsedRealtimeNanos`, but not `elapsedRealtime` (milliseconds), so I needed to convert the 2 minutes from milliseconds to nanoseconds.
I'll rename it to something like `TWO_MINUTES_IN_NANOSECONDS`.

westnordost:
perfect, thank you!

app/src/main/java/de/westnordost/streetcomplete/util/location/FineLocationManager.kt
```diff
@@ -98,9 +98,13 @@ private fun Location.isBetterThan(previous: Location?): Boolean {
     if (previous == null) return true
 
     // Check whether the new location fix is newer or older
-    val timeDelta = this.time - previous.time
-    val isMuchNewer = timeDelta > TWO_MINUTES
-    val isMuchOlder = timeDelta < -TWO_MINUTES
+    // we use elapsedRealtimeNanos instead of epoch time because some devices have issues
+    // that may lead to incorrect GPS location.time (e.g. GPS week rollover, but also others)
+    // the use of nanoseconds is necessary because it iss the only way to get
```
FloEdelmann, line 103: Typo: iss → is

matkoniecz: https://github.com/streetcomplete/StreetComplete/commit/ac56d0866a9c968e41f4b406d28f13be5e8cf155

PR #4850, opened 2023-02-28
Add optional addr:block input in some countries

### LINKED ISSUE #4718: Ameliorate address quest (needed for India)

Helium314:
fixes #4718

This is a first version of adding optional `addr:block` input.
I added a separate form because the Japanese one mentions block number, but here it might also be a letter or longer.

For now, it's only for the address overlay. I'll add it to the housnumber quest once a few things are solved / clarified:
* The list of countries where `addr:block` is common should be in one place, but I have no idea where to best put it
(according to taginfo, `addr:block` is used in India, Egypt and Uruguay, so the input can only be enabled there)
* Switching to `addr:block` input happens via some "toggle block input" other answer. This is fine for overlay, but not for a question (housenumber quest), where I would need to change the text of the other answer after it's created. How should this be done properly?
* Is the way of switching (`toggleBlockInput`) ok? This creates a new `AddressNumberAndNameInputViewController`.
* Block names in Uruguay are sometimes [quite long](https://www.openstreetmap.org/node/7099034942), so the input field would need to be longer or variable width (edit: also in Egypt)
* Currently `addr:block` is using `HouseAndBlockNumber`, and the decision whether to put `blockNumber` in `addr:block` or `addr:block_number` depends on the countrry code. I guess a less hacky way is preferred? Something like adding a `HouseNumberAndBlock` class maybe?


app/src/main/res/values/strings.xml
```diff
@@ -621,6 +621,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="label_housenumber">house number</string>
     <!-- In Japan, addresses are chiefly house number + block number (because most streets don't have a name) -->
     <string name="label_blocknumber">block number</string>
+    <!-- In some countries, addresses may have a house number and a block (number or letters -->
```
FloEdelmann, line 624:
```suggestion
    <!-- In some countries, addresses may have a house number and a block (number or letters) -->
```

app/src/main/java/de/westnordost/streetcomplete/overlays/address/AddressOverlayForm.kt
```diff
@@ -110,9 +112,13 @@ class AddressOverlayForm : AbstractOverlayForm() {
             streetOrPlaceBinding.streetOrPlaceSelect.isGone = true
         }
 
+        showNumberOrNameInput(getAddressNumberLayoutResId(countryInfo.countryCode, if (addressNumber != null) addressNumber is HouseAndBlockNumber else lastBlockNumber != null))
```
FloEdelmann, line 115:
```suggestion
        val useOptionalBlockLayout = if (addressNumber != null) addressNumber is HouseAndBlockNumber else lastBlockNumber != null)
        showNumberOrNameInput(getAddressNumberLayoutResId(countryInfo.countryCode, useOptionalBlockLayout)
```

app/src/main/java/de/westnordost/streetcomplete/overlays/address/AddressOverlayForm.kt
```diff
@@ -50,11 +50,13 @@ class AddressOverlayForm : AbstractOverlayForm() {
     private var streetOrPlaceName: StreetOrPlaceName? = null
 
     private var isShowingHouseName: Boolean = false
-    private var isShowingPlaceName = false
+    private var isShowingPlaceName: Boolean = false
+    private var isShowingBlockNumber: Boolean = false
```
FloEdelmann, line 54: Probably needs to be saved in `onSaveInstanceState` like the other two. Also, this is never set to `true`, right?

app/src/main/java/de/westnordost/streetcomplete/overlays/address/AddressOverlayForm.kt
```diff
@@ -169,10 +186,11 @@ class AddressOverlayForm : AbstractOverlayForm() {
 
         number?.streetHouseNumber?.let { lastHouseNumber = it }
         if (number is HouseAndBlockNumber) { number.blockNumber.let { lastBlockNumber = it } }
+            else lastBlockNumber = null
```
FloEdelmann, line 189:
```suggestion
        lastBlockNumber = if (number is HouseAndBlockNumber) number.blockNumber else null
```

matkoniecz:
> The list of countries where addr:block is common should be in one place, but I have no idea where to best put it

https://github.com/streetcomplete/countrymetadata

matkoniecz:
> where I would need to change the text of the other answer after it's created. 

which exactly text needs to be changed?

app/src/main/java/de/westnordost/streetcomplete/overlays/address/AddressOverlayForm.kt
Helium314: Is the `let` here really useless? It looks like, but I thought it's probably here for a reason...

app/src/main/java/de/westnordost/streetcomplete/overlays/address/AddressOverlayForm.kt
Helium314: > Probably needs to be saved in onSaveInstanceState

Probably. Though maybe it could also be read from `numberView.findViewById<EditText?>(R.id.blockNumberInput) != null`?

> Also, this is never set to true, right?

Set to true if there is a block number input in `isShowingBlockNumber = numberView.findViewById<EditText?>(R.id.blockNumberInput) != null`

matkoniecz:
> iirc other answers for quests should be more like "it's a block number...", but then this answer would need to be changed to something like "it has no block number".

note that house name seems to not have such functionality, so it can be probably also skipped here (user can exit and reenter form I guess)

app/src/main/java/de/westnordost/streetcomplete/overlays/address/AddressOverlayForm.kt
FloEdelmann: Introduced in a438117f7a85846ae0ab782d0695efb75cfd6360. It likely was a refactoring from a nullable `blockNumber`, where `?.let { … }` makes sense.

westnordost:
>  I added a separate form because the Japanese one mentions block number, but here it might also be a letter or longer.

It would be okay to allow (via ABC button) the user to enter letters for the block number input too. Let's not exclude the possibility that there are blocks like "5A" etc. in Japan.

Open for discussion: Not sure if it is necessary to limit showing this option to countries that are known to use blocks for addresses. After all, it does not busy the main UI but is behind an "other answer". For example Mannheim also has blocks.

> Is the way of switching (toggleBlockInput) ok? This creates a new AddressNumberAndNameInputViewController.

Looks okay to me. But is it actually valid in Kotlin to have a lateinit var change its value?

> Currently addr:block is using HouseAndBlockNumber, and the decision whether to put blockNumber in addr:block or addr:block_number depends on the countrry code. I guess a less hacky way is preferred? Something like adding a HouseNumberAndBlock class maybe?

Yes. Question is, who's job is it to determine which format to use. The user of the data structure (the part of the code that tags the OSM tags) or the creator of the data structure (the form). And IMO, it is the latter.

Helium314:
> But is it actually valid in Kotlin to have a lateinit var change its value?

As far as I understand it's mostly a normal var, and the lateinit is basically a way to avoid initializing wiht null and thus requiring a nullable type. Also: no issues so far when testing with the current state of the PR.

Helium314:
> It would be okay to allow (via ABC button) the user to enter letters for the block number input too

Ok, so use to the Japanese layout, with some modifications?
* The block (number) field should be allowed to grow when the text is long
* The default keyboard should be text, except in Japan
* The text below the block(number) should be "block", but stay "block number" in Japan

westnordost:
> The block (number) field should be allowed to grow when the text is long
> The default keyboard should be text, except in Japan
> The text below the block(number) should be "block", but stay "block number" in Japan

Hm, in this case a separate layout seems like the better idea. Especially if it should be allowed to grow, it should be laid out vertically.

I do wonder whether block + housenumber should be considered as (sufficiently) complete address. I notice that 75% of all objects tagged with `addr:block` do also have a `addr:street` tagged. However, if the street name is not mandatory for the address to be considered unique in one city, rather the block should be treated on the same level as the street name. I.e. the input above would be a street name or place name or block (name) input.

Helium314:
Changed:
* add `HouseNumberAndBlock`, use it in `blockInput` in `AddressNumberInputViewController`
* allow switching to block input everywhere except in Japan
* use (slightly modified) Japanese form all the time

Missing:
* actual testing (only tried to compile)
* show "block" in form instead of "block number" (except in Japan)
* default to letter input in form (except in Japan)

Helium314:
> Hm, in this case a separate layout seems like the better idea

Ok, I was too quick here ^^

>  I notice that 75% of all objects tagged with addr:block do also have a addr:street tagged

Do most of the others have `addr:place`? I remember that was usually the case when I did some random checks (not always though).

westnordost:
What do you mean? That if they do not have `addr:street`, they will have `addr:place` for sure?

In that case 

> the input above would be a street name or place name or block (name) input.

would not be the best choice.

However, with street name, block name and house number each on top of each other, I worry that the form may grow too big. How growable are we talking about, here? If not so much text is too be expected, maybe they could be horizontal to each other after all, but the text shrinks instead if the maximum size is reached.

Helium314:
> If not so much text is too be expected, maybe they could be horizontal to each other after all, but the text shrinks instead if the maximum size is reached.

The longest I found are similar to `Viviendas Jc 24`, so one line and shrinking text is probably fine, as long as the field is allowed to grow somewhat.

> That if they do not have addr:street, they will have addr:place for sure?

Almost. There are some addresses in Uruguay and very few in India that have neither place nor street. I think we can still have place or street required.

westnordost:
> so one line and shrinking text is probably fine, as long as the field is allowed to grow somewhat.

Not sure if the combination of both is possible Surely hasn't been done yet in the project.

Helium314:
> I'll try and use a separate layout if I can't make it work.

Well, now at least I remember why I despise those things.
Behavior on phone is quite different from Android Studio, and the field grows only if the number input is active. So long numbers are fine, but long text isn't.

Only way I got it working reasonably is with input `textMultiLine`, `maxlines` 2 and a `maxwith`.
The 2 lines look quite ok imo, and if maxwidth is small enough there are no issues with overlapping fields / buttons.

Helium314:
I re-added the separate `block` form, as it allows for better separation between `block` and `block_number` related things.
The form works as described above (usable for long `block`s), and doesn't show the "ABC" button as `addr:block` rarely contains numbers without other text.

I also added the block input to the `AddHousenumber` quest, and it's showing the last used input. Switching happens via other answer without a way of switching back, as suggested in https://github.com/streetcomplete/StreetComplete/pull/4850#issuecomment-1448741884.

Helium314:
Should `getShortHouseNumber` and `getHouseNumberHtml` in [NameAndLocationLabel](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/util/NameAndLocationLabel.kt) also show block / block number?

westnordost:
I'd say so.  Doesn't it for block number+ housenumber in japan right now?

Helium314:
No, a search for "block" in [NameAndLocationLabel](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/util/NameAndLocationLabel.kt) return no results.

Helium314:
> Should getShortHouseNumber and getHouseNumberHtml in [NameAndLocationLabel](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/util/NameAndLocationLabel.kt) also show block / block number?

Actually `getShortHouseNumber` should rather not contain block (number), as it might become too long (it's only shown on the map)

westnordost:
Is this finished / ready for review? Is something missing?

Helium314:
Yes, it's ready.

I didn't add the block / block_number to getHouseNumberHtml yet, because that doesn't really belong in this PR.

westnordost (changes requested):
Clean code overall, nothing to complain about. 

However, I tested it, there are some issues left with it:
- I changed some small things myself (see my commit)
- layout issues with the growing block number input: Layout goes out of view to the left, overlaps with the abc button and the current housenumber (as hint) is shown as "...9" instead of full.

Screenshot_20230306_151648_StreetComplete Dev

Helium314:
The issue seems to be that my phone has set a smaller icon/text size than the Android default.
Changing the `maxwitdh` to 110 dp helps, but when the text is larger than normal then it breaks the form again...

The previous housenumber hint is unchanged, also in the normal form I get "...66" as hint when the previous number was 66666.

Helium314:
Looks like maxwidth should be form width minus 250 dp

westnordost (approved):
I tested again, works flawlessly

Issue #4374, opened 2022-09-13
new building type: security checkpoint 


Identitaet:
**Use case**

Often next to barriers there are little buildings where a security guard sits and does security stuff 
**Proposed Solution**

I'm a bit unsure how one would tag.

But the solution is to add it

wielandb:
There is [building=guard_post](https://wiki.openstreetmap.org/wiki/Tag:building%3Dguard_post), but it only has 24 uses :(

westnordost:
True about the low usage. People have been struggling to find a proper tag for that. There is `building=gatehouse` (German translation: Pförtnerhäuschen), however the wiki defines this as a building that spans over a road. 

@SK53 added to the wiki in November 2021 that this tag (should) also describe it as 

> small modern buildings at the entrances to large commercial, industrial, military or academic (university) sites) housing security personnel who control access to the site

but @matkoniecz removed this again in June 2022, pointing instead to the tag mentioned, `building=guard_post`. Given that the tag hasn't been documented very long, I think `building=guard_post` is fine, the low usage is not an issue because there is not any more popular tag to describe the thing.

I am worried though about that the usage number of `building=gatehouse` went up rapidly only after @SK53 added the cited paragraph to the wiki. So maybe the de-facto meaning of `building=gatehouse` has already become what he described it as?

westnordost:
By the way, so far I have usually tagged such guard posts as simply "commercial", i.e. a nondescript/multipurpose building on a commercial or industrial compound. IMO that's also fine. There are uncountable special purpose buildings around, we cannot accomodate all anyway.

matkoniecz:
>  I have usually tagged such guard posts as simply "commercial", i.e. a nondescript/multipurpose building on a commercial or industrial compound

I prefer to tag them specially as they are (at least to me) quite distinct and not really commercial. Especially ones on military bases.

Ones in gated residential areas and in industrial sites can sort of fir commercial label, but...

matkoniecz:
There is also https://en.wikipedia.org/wiki/Guardhouse https://taginfo.openstreetmap.org/tags/building=guardhouse#overview that seems to apply both to tiny booths and large buildings. Somehow I ignored/missed it while making writeup last time?

Would it cover https://commons.wikimedia.org/wiki/Category:Crossing_keeper_cabins ? Note say https://www.openstreetmap.org/way/589832071

SK53:
SK53 here. 

[...]

PR #4852, opened 2023-03-01
make clear that building guardhouse is not only military

matkoniecz:
implements idea from https://github.com/streetcomplete/StreetComplete/issues/4374#issuecomment-1448842173

(PR as contain a new string)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_type/BuildingTypeItem.kt
@@ -101,6 +101,7 @@ private val BuildingType.descriptionResId: Int? get() = when (this) {
     RESIDENTIAL ->     R.string.quest_buildingType_residential_description
     COMMERCIAL ->      R.string.quest_buildingType_commercial_generic_description
     CIVIC ->           R.string.quest_buildingType_civic_description
+    GUARDHOUSE ->      R.string.quest_buildingType_guardhouse_description
     else ->            null
 }
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -870,6 +870,7 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_buildingType_ruins_description">derelict, ruined and unidentifiable building</string>
     <string name="quest_buildingType_fire_station">Fire station building</string>
     <string name="quest_buildingType_guardhouse">Guardhouse</string>
+    <string name="quest_buildingType_guardhouse_description">building constructed to house security staff, common at entrances to controlled areas</string>
 
     <!-- "stop" as in "bus stop". But could be a tram stop, railway stop etc. too -->
     <string name="quest_busStopBench_title2">"Is there a bench at this stop?"</string>


michaelblyons:
I don't have a good replacement word in mind, but I'm just a little concerned about "to house" in case that makes someone think it's like a dormitory or a barracks. Is there a better way to communicate that it's the place that they perform their duties?

matkoniecz:
building constructed **for** security staff, common at entrances to controlled areas

riQQ:
Or *accommodate*?

mnalis:
I'd also add `small` before `building`

westnordost:
Why is a description necessary? Why is it not clear from the name plus the picture?

mnalis:
> Why is a description necessary? Why is it not clear from the name plus the picture?

According to https://github.com/streetcomplete/StreetComplete/issues/4374#issuecomment-1448842173, it is not clear.

Also in some languages it seems to imply military facilities, and the picture does help either in that case (as it looks exactly like an military base entrance too).

westnordost:
Can you post the reasons here?

mnalis:
> Can you post the reasons here?

If you mean copy/paste that relevant comment by @mpawelski , sure, here it is:

>I just learned about this building=guardhouse because my old [note](https://www.openstreetmap.org/note/3027301) was resolved and I wanted to suggest adding this type to StreetComplete. Nice that this was added!

> But I would suggest adding some small text description to it, because IMO just a "Guardhouse" ("Wartownia" in polish) doesn't explain that well what is it. Something like the the description in [wiki](https://wiki.openstreetmap.org/wiki/Tag:building%3Dguardhouse) would be great. Maybe just shorter, like for example:

> "small building with guards inside, common at supervised parking lots and other areas that are surveilled by guards"

But if you mean if I can explain why exactly how it confuses @mpawelski, only they can give authoritative answer on that (I tried to summarize my understanding in my comment above in https://github.com/streetcomplete/StreetComplete/pull/4852#issuecomment-1452393970).

westnordost:
Yes,  i man how this is confusing. 

And if it is a translation issue / there is no general word for that (but several), those could be put into the translation of the name.

mpawelski:
> Yes, i mean how this is confusing.

It is very confusing and not clear enough to me and I though that I'm probably not the only one. That's why I thought it would be an improvement to add some small text description like there are for other building types.

As other mentioned, currently the text and icon is more suggesting that it's only for military-related areas. You could argue that ones can deduce that it also applies to places like parking lots, but I really doubt it's that obvious to most people. I want to make it clear from the first sight.

That's why I would actually like to see some examples in the text. Maybe something like this?:

"building constructed for security staff, common at entrances to controlled areas, such as parking lots, military bases, apartment buildings"




> And if it is a translation issue / there is no general word for that (but several), those could be put into the translation of the name.

It is _somehow_ related to translations issues but finding a good words for all use case can be really hard/impossible and not that shorter than small text with a description.

For example, no one calls "guardhouse" ("Wartownia" in polish) a building in a parking lot. I think there is no good word for it. I would call it "[kanciapa](https://sjp.pwn.pl/szukaj/kanciapa.html)" but it's actually more general wold that mean "a small building".

matkoniecz:
> For example, no one calls "guardhouse" ("Wartownia" in polish) a building in a parking lot.

"Budka strażnicza"?

mpawelski:
> > For example, no one calls "guardhouse" ("Wartownia" in polish) a building in a parking lot.
> 
> "Budka strażnicza"?

Sounds a bit better.
But I still think it has to big connotation with military. Having text with examples like proposed in my previous comment would make it clear that parking is allowed. 

[Parking booth](https://www.google.com/search?q=parking+booth&client=firefox-b-d&sxsrf=AJOqlzWZR0lWiRe_HzvmxqGA-X7n-NDPew:1677845912945&source=lnms&tbm=isch&sa=X&ved=2ahUKEwiq2ION37_9AhXNmYsKHSgqAmAQ_AUoAXoECAIQAw&biw=1920&bih=927&dpr=1) also looks like another good name that makes it clear that it's not only military related. But how we would display it? "Guardhouse. Parking booth".?

westnordost:
_"Wartownia / Budka strażnicza / ..."_

matkoniecz:
That still sounds quite military, more than just "Budka strażnicza". 

"Wartownia" fits basically only armed forces and at least suggests that people inside are armed.

"Budka strażnicza / portiernia / wartownia"?

portiernia seems strongly on civilian side, but it more often refers to place in a building, next to an entrance - not a separate building. But is also used for this booths.

mpawelski:
"Guardhouse / Parking booth"  in English and "Wartownia / Budka parkingowa" in Polish?

I'm proposing "Parking booth" for English because I'm not sure that just a "Guardhouse" is clear enough. But I'm not native speaker so maybe it is clear.

In Polish I would actually do "Wartownia / Budka parkingowa / Portiernia". Because "Portiernia" is also a good name that suggest there is a "portier" inside, someone that allows you to enter some building (and I wouldn't call him "security")

mnalis:
> Because "Portiernia" is also a good name that suggest there is a "portier" inside, someone that allows you to enter some building (and I wouldn't call him "security")

Although, that might get confused with [amenity=reception_desk](https://wiki.openstreetmap.org/wiki/Tag:amenity=reception_desk) / [camp_site=reception](https://wiki.openstreetmap.org/wiki/Tag:camp_site=reception) / [office=reception](https://wiki.openstreetmap.org/wiki/Tag:office=reception) which are **also** commonly situated in the entrance part of the **big** (main) building.

PR #4868, opened 2023-03-11
Ask wheelchair toilets for restaurants etc.

### LINKED ISSUE #4866: Ask if restaurants have a wheelchair accessible toilet where `toilets` is not already mapped
**Use case**

Wheelchair accessibility of toilets is important information(`toilets:wheelchair`). The quest `AddWheelchairAccessToiletsPart` collects this for anywhere with
```
toilets = yes
and name
and access !~ no|private
```
(and resurvey)

In some countries, essentially all restaurants have toilets and many are not marked with `toilets=yes` so this quest would not show.
(See #1272 and #4860 for reasons for not asking for toilets for all restaurants).

**Proposed Solution**
Extend the element selection to include `amenity~restaurant|pub|bar` or `amenity~cafe|fast_food and indoor_seating=yes` without `toilets=no`. Also, add an other answer "There is no toilet here" to add `toilets=no`
Options:
1) Only apply that extra bit only in countries known to require toilets in restaurants (research to be done).
2) Apply that everywhere, since `AddWheelchairAccessToiletsPart` is disabled by default anyway.

arrival-spring:
Fixes #4866 

Tested and works as expected and tests all pass

I wonder if the requirement for a place to have a name when only toilets=yes is tagged for the quest to show is left over from before feature types were shown. However, that's unrelated to this change so I left it.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
@@ -30,10 +30,7 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAns
 
     override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
         when (answer) {
-            is WheelchairAccessToilets -> {
-                tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
-                tags["toilets"] = "yes"
-            }
+            is WheelchairAccessToilets -> answer.applyTo(tags)
             is NoToilet -> tags.updateWithCheckDate("toilets", "no")
         }
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
@@ -7,7 +7,7 @@ import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.
 import de.westnordost.streetcomplete.osm.Tags
 import de.westnordost.streetcomplete.osm.updateWithCheckDate
 
-class AddWheelchairAccessToilets : OsmFilterQuestType<WheelChairAccessAnswer>() {
+class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAnswer>() {
 
     override val elementFilter = """
         nodes, ways with amenity = toilets
@@ -28,10 +28,13 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelChairAccessAnswer>() 
 
     override fun createForm() = AddWheelchairAccessToiletsForm()
 
-    override fun applyAnswerTo(answer: WheelChairAccessAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+    override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
         when (answer) {
-            is WheelchairAccess -> tags.updateWithCheckDate("wheelchair", answer.osmValue)
-            else -> throw IllegalStateException("Other wheelchair answer")
+            is WheelchairAccessToilets -> {
+                tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
+                tags["toilets"] = "yes"
+            }
+            is NoToilet -> tags.updateWithCheckDate("toilets", "no")
         }
     }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
@@ -7,7 +7,7 @@ import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.
 import de.westnordost.streetcomplete.osm.Tags
 import de.westnordost.streetcomplete.osm.updateWithCheckDate
 
-class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccess>() {
+class AddWheelchairAccessToilets : OsmFilterQuestType<WheelChairAccessAnswer>() {
 
     override val elementFilter = """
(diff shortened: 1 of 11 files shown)


westnordost (changes requested):
Please use the following sealed interface:

```kotlin
sealed interface WheelchairAccessToiletsAnswer
data class WheelchairAccessToilets(val access: WheelchairAccess): WheelchairAccessToiletsAnswer
object NoToilet : WheelchairAccessToiletsAnswer
```

for the `AddWheelchairAccessToilets` quest and make `AddWheelchairAccessToiletsForm` not inherit from `WheelchairAccessForm` so that this data structure is not used for all the other wheelchair quests.

westnordost:
> I wonder if the requirement for a place to have a name when only toilets=yes is tagged for the quest to show is left over from before feature types were shown. 

I think yes. So, feel free to remove it.

arrival-spring:
> Please use the following sealed interface:

That makes more sense, thanks, and done

> I think yes. So, feel free to remove it.

Done

westnordost (commented):
Since WheelchairAccessToilets data structure is used in several places, you could get rid of the duplication by adding a function like 

`fun WheelchairAccessToilets.applyTo(tags: Tags)`

not required for it to be merged, though.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToiletsPart.kt
```diff
@@ -8,12 +8,21 @@ import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.
...
+             toilets = yes
+             or
+             (
+                (
+                    amenity~restaurant|pub|bar
+                    or amenity~cafe|fast_food and indoor_seating=yes
+                )
+                and !toilets
+             )
+         )
```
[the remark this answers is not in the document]
arrival-spring: I did think the filter could be a bit neater, thanks

westnordost:
Oh well, I meant WheelchairAccessToiletsAnswer.applyTo, sorry. Anyway, thank you!

westnordost:
Hmm, something occurred to me.

Setting `wheelchair=yes` on `amenity=restaurant` (etc) is already supposed to cover that also the toilets are wheelchair accessible. So, this is quite duplicate here.

arrival-spring:
Ah yes good point.

Although the wiki page for wheelchair doesn't seem to say hat wheelchair=yes and toilets=yes implies toilets:wheelchair=yes so maybe it is good to map it explicitly.

And I guess this also has benefit for wheelchair=limited

Helium314:
Does it make sense to exclude elements with `wheelchair=no`? I don't have much experience with wheelchair mapping, but I wouldn't expect an accessible toilet if the whole place is tagged as not accessible.

matkoniecz:
In theory it could be possible for rare cases where say toilet is at level 0 and restaurant in general upstairs/downstairs but it is extremely rare.

Limiting this quest to `wheelchair=limited` seems to be a good idea.

mnalis:
> In theory it could be possible for rare cases where say toilet is at level 0 and restaurant in general upstairs/downstairs but it is extremely rare.

Well, over here it is actually quite common (especially in cafes/pubs, but restorants too) that restaurant is at ground level, but toilets are underground (and in majority of cases only accessible by steps).

> Limiting this quest to wheelchair=limited seems to be a good idea.

I agree that `toilets:wheelchair=no` must mean the whole place can **not** be `wheelchair=yes` ! 

But perhaps (especially given target audience who might've tagged restaurant as `wheelchair=yes` without specifically surveying toilets area) it might be better to **always** ask for toilet wheelchair accessibility, and then warn (like we do for mismatched `tracktype`/`surface` way) or even automatically fix `wheelchair=yes` to `wheelchair=limited`? 

That should produce safer and more correct data.

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
```diff
@@ -28,7 +28,10 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccess>() {
 
     override fun createForm() = AddWheelchairAccessToiletsForm()
 
-    override fun applyAnswerTo(answer: WheelchairAccess, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
-        tags.updateWithCheckDate("wheelchair", answer.osmValue)
+    override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        when (answer) {
+            is WheelchairAccessToilets -> answer.applyTo(tags)
+            is NoToilet -> tags.updateWithCheckDate("toilets", "no")
+        }
```
Helium314, line 35: Is it wanted that this now uses `toilets:wheelchair` instead of `wheelchair` for `amenity=toilets`?
Also adding `toilets=no` here doesn't really make sense.

matkoniecz: > Edit: toilets:wheelchair [should not be used here](https://wiki.openstreetmap.org/wiki/Tag:amenity=toilets#Toilets_within_places)

Are you sure? 

> Facilities such as shops and pubs often have toilets inside them for customers, and sometimes free for the public

> In general please do not map toilets that are inaccessible to the public. Many buildings have toilets inside for workers or owners

toilets:wheelchair seems applicable to customer-only toilets

Helium314: Still it might be a good idea to not go against the vast majority of current tagging:
`amenity=toilets` with `wheelchair`: 74 887
`amenity=toilets` with `toilets:wheelchair`: 7 124

arrival-spring: No. That was introduced here when applyTo was introduced. I'll revert back to before that this evening

arrival-spring:
Here are some relevant numbers of elements
`toilets=yes`, `toilets:wheelchair=no`, `wheelchair=yes` : 423
`toilets=yes`, `toilets:wheelchair=no`, `wheelchair=limited` : 562
`toilets=yes`, `toilets:wheelchair=yes`, `wheelchair=*` : 3284
`toilets=yes`, `wheelchair=yes` : 9100
`toilets=yes`, `wheelchair=limited` : 2695

'Incorrect' tagging of `toilets:wheelchair=no`, `wheelchair=yes` is very few elements, so not sure if it is worth implementing something to check for that. Also, I wouldn't be sure where the check should be made as there would be different tags that need changing depending on the actual situation

PR #4871, opened 2023-03-12
Allow buildings to be tapped for shop overlay

dbdean:
Often shops are in separate buildings that are already mapped, and it would be more correct to tag the entire building with the shop tagging than add a node in the building.

This PR allows buildings to be tapped in the shop overlay, and then get a shop tagging applied on top of the existing building tagging.

When a non-shop-tagged building is tapped, it treats it as if it was creating a new feature, so it shows the blank feature input form, and no information from the existing building is shown, changed or removed (except the name, if tagged).

I have attempted to implement this a few different ways (see the commit history), and I think this implementation would make the most sense. I have tested it in real world mapping and it appears to work as intended.


matkoniecz:
> Often shops are in separate buildings that are already mapped, and it would be more correct to tag the entire building with the shop tagging than add a node in the building.

No, that is worse, and less correct. As shop (business operating in a building) and building are separate object they should be tagged separately. 

It also makes processing data much easier  in many cases. For example many named POIs are in buildings without own names - in case of tagging them as a single object then  you get tricky issues when you try to display building names (there is some OSM Carto about this problem)

westnordost:
I agree with @matkoniecz there. We certainly do not want to encourage adding POIs to buildings, but this PR would have exactly this effect. As far as I know, in cases where there is already a POI mapped on a building, it is possible to change the shop type. So, the overlay already works as intended here.

dbdean:
> No, that is worse, and less correct. As shop (business operating in a building) and building are separate object they should be tagged separately.

I don't agree with this at all. If a business tags up an entire building it absolutely should be mapped as the entire area, not just as a POI. I guess ideally you could map two areas that share the same nodes, but that seems a bit excessive and difficult to maintain.

dbdean:
And [taginfo shows](https://taginfo.openstreetmap.org/tags/building%3Dretail#combinations)that 20% of building=retail are also tagged with shop=something. amenity=something adds another 5%. 

This isn't just me - this is an well-accepted way of tagging, and at least in my opinion - the correct way of tagging when a single shop occupies an entire building. Only mapping a node is losing information.

westnordost:
You are not incorrect, however 

>  We certainly do not want to encourage adding POIs to buildings

still applies.

dbdean:
I disagree - based on existing tagging practice - adding shop tagging to buildings is the correct thing to do when the building only has one shop, and should not be discouraged. SC should be matching the current practice unless there is a strong reason not to.

Is there anywhere where your advice that tagging shop information on a building should be discouraged? I can't find it.

matkoniecz:
>  If a business tags up an entire building it absolutely should be mapped as the entire area, not just as a POI. 

In such case it is less obviously a bad area, but such change would encourage doing this also in cases where shop takes part only of small part of building, or only ground level.

Also, even in cases where shop takes 100% of building (fairly rare, in nearly all cases shop POI is only part and there is part that is used by business and is not shop POI) it should be mapped as area separate from a building.

> Is there anywhere where your advice that tagging shop information on a building should be discouraged is documented? I can't find it.

Is it somewhere documented as preferred tagging for shop in buildings? Especially in cases where shop is the sole business in building but takes only part of space?

matkoniecz:
I guess that it could be discussed at https://community.openstreetmap.org/ if you want to start a thread or on tagging mailing list.

dbdean:
Your example above is not of a shop that occupies the entire building, as there 4 levels or residential above - so it isn't the case that intended this functionality for. 

As for the matter of non-public areas of the shops, I don't think there is a serious problem with those being in the shop=something area concurrent with the building. I wouldn't go to the effort in most cases in making such a distinction between the front and back areas of shops.

I think I'm used to mapping in a less built up area where multi-story buildings and buildings with many shops are relatively rare. I can see how this approach could lead to people tagging entire buildings like your example as a single shop when they are more than just one shop.

While I think this functionality would be very useful to me, and probably others in similar situations to me, I can see that there would be potential downsides with poor and incorrect mapping in more built up areas. I'm happy to close this PR, and continue mapping buildings as shops where appropriate outside of SC in my area.

mnalis:
> And [taginfo shows](https://taginfo.openstreetmap.org/tags/building%3Dretail#combinations) that 20% of building=retail are also tagged with shop=something. amenity=something adds another 5%.

Which would also mean that 75% of `building=retail` are **not** tagged with shop/amenity, right?


> Is it somewhere documented as preferred tagging for shop in buildings? 

https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element does list under `Examples of bad situations` example of:

> Shop in a [building](https://wiki.openstreetmap.org/wiki/Key:building) mapped as a single object with building=* and [shop](https://wiki.openstreetmap.org/wiki/Key:shop)=* and [name](https://wiki.openstreetmap.org/wiki/Key:name)=* 

The obvious problem is what does `name` applies to? Building name? Shop name? What if we added [start_date=*](https://wiki.openstreetmap.org/wiki/Key:start_date) - is it the date the shop opened, or the date building was constructed (esp. note the situations that new shops often reuse old buildings, instead of razing them to the ground and building new ones). What about `operator=*`? Operator of the shop, or of the building (many shops rent buildings instead of buying them outright or building from scratch). `wikipedia`/`wikidata` ? etc...

So if the building and shop are merged in single element, you can not simply add/edit any tags without checking each and every existing or new tag if it is appropriate for **both** building and a shop, or do you need to split them before adding/updating tags. So much **simpler and less time consuming  and less prone to error to keep them separate** in the first place, even in situations when _initially_ building+shop tagged on same element might be non-ambiguous.

matkoniecz:
> https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element does list under Examples of bad situations example of:

Though note that person who added it is me, so it is not exactly an independent confirmation.

matkoniecz:
> Your example above is not of a shop that occupies the entire building, as there are 4 levels of residential above - so it isn't the case that I intended this functionality for.

Yes, I know - but I expect that it will be used also for such cases even if it is not intended.

PR #4874, opened 2023-03-14
make roof:levels mandatory

### LINKED ISSUE #4661: Nachfrage nach Anzahl der Dachgeschosse
In der letzten Zeit hatte ich mehrmals den Fall, dass ich einen Gebäudetyp eingegeben habe und SC dann sofort nach dem Dachtyp fragte.
Nachforschungen ergaben, dass bei diesen Gebäuden bereits ein building:level gesetzt war, jedoch ein roof:level fehlte.
SC sollte in solchen Fällen nach der Anzahl der Dachgeschosse fragen.

### LINKED ISSUE #4821: ask about roof:levels where unset, start filling roof:levels=0 in countries where roof shape is not usually flat
Make `roof:levels` into a mandatory field in the building levels quest and re-ask building level quest for buildings that do have a `roof:shape` that is not `flat` but no`roof:levels` set. But:
- on new entry, prefill the roof levels field with "0". When tabbing into it / tapping on it, the whole text is selected so it can easily be replaced
- make the `roof:levels` field only mandatory (and prefill with 0 etc) in country where the roof shape is not usually flat (https://github.com/streetcomplete/countrymetadata/blob/master/data/roofsAreUsuallyFlat.yml)

this is copy of https://github.com/streetcomplete/StreetComplete/issues/4661#issuecomment-1431185989 and summary of https://github.com/streetcomplete/StreetComplete/issues/4661

westnordost:
Fixes #4821 (also fixes #4661)

- Ask building levels quest again for buildings that have no `roof:levels` but a `roof:shape` that is not `flat`
- make specifying `roof:levels` mandatory in countries where [roofs are not usually flat](https://github.com/streetcomplete/countrymetadata/blob/master/data/roofsAreUsuallyFlat.yml) and in cases where `roof:shape` has already been supplied and it is not `flat`

@biketeur

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
@@ -27,10 +27,6 @@ class AddBuildingLevelsForm : AbstractOsmQuestForm<BuildingLevelsAnswer>() {
         AnswerItem(R.string.quest_buildingLevels_answer_multipleLevels) { showMultipleLevelsHint() }
     )
 
-    private val roofsAreUsuallyFlat get() = countryInfo.roofsAreUsuallyFlat
-    private val hasNonFlatRoofShape get() =
-        element.tags.containsKey("roof:shape") && element.tags["roof:shape"] != "flat"
-
     private val levels get() = binding.levelsInput.intOrNull?.takeIf { it >= 0 }
     private val roofLevels get() = binding.roofLevelsInput.intOrNull?.takeIf { it >= 0 }
 
@@ -91,8 +87,11 @@ class AddBuildingLevelsForm : AbstractOsmQuestForm<BuildingLevelsAnswer>() {
         }
     }
 
-    override fun isFormComplete() =
-        levels != null && (roofsAreUsuallyFlat && !hasNonFlatRoofShape || roofLevels != null)
+    override fun isFormComplete(): Boolean {
+        val hasNonFlatRoofShape = element.tags.containsKey("roof:shape") && element.tags["roof:shape"] != "flat"
+        val roofLevelsAreOptional = countryInfo.roofsAreUsuallyFlat && !hasNonFlatRoofShape
+        return levels != null && (roofLevelsAreOptional || roofLevels != null)
+    }
 }
 
 private class LastPickedAdapter(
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -54,7 +58,14 @@ class AddBuildingLevelsForm : AbstractOsmQuestForm<BuildingLevelsAnswer>() {
...
         super.onViewCreated(view, savedInstanceState)
 
-        binding.levelsInput.requestFocus()
+        if (savedInstanceState == null) {
+            binding.levelsInput.setText(element.tags["building:levels"])
+            binding.roofLevelsInput.setText(element.tags["roof:levels"])
+        }
+        val focussedInput = if (levels == null) binding.levelsInput else binding.roofLevelsInput
+        focussedInput.requestFocus()
+        focussedInput.selectAll()
```
FloEdelmann, line 67:
```suggestion
        val focusedInput = if (levels == null) binding.levelsInput else binding.roofLevelsInput
        focusedInput.requestFocus()
        focusedInput.selectAll()
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -81,9 +92,7 @@ class AddBuildingLevelsForm : AbstractOsmQuestForm<BuildingLevelsAnswer>() {
     }
 
     override fun isFormComplete() =
-        // levels must be an int >= 0. IF roof levels is specified, it must also be an int >= 0
-        levels?.let { it >= 0 } ?: false
-        && (roofLevels == null || roofLevels?.let { it >= 0 } ?: false)
+        levels != null && (roofsAreUsuallyFlat && !hasNonFlatRoofShape || roofLevels != null)
```
FloEdelmann, line 95: I'd always avoid mixing `&&` and `||` on the same level, because you always have to think twice which one takes precedence. So either, I'd add parentheses:

```suggestion
        levels != null && ((roofsAreUsuallyFlat && !hasNonFlatRoofShape) || roofLevels != null)
```

Or, I'd extract a variable:

```kt
private val roofLevelsAreOptional get() = roofsAreUsuallyFlat && !hasNonFlatRoofShape
// …
override fun isFormComplete() =
    levels != null && (roofLevelsAreOptional || roofLevels != null)

westnordost: Really, you have to think twice? Come on, what about `5 + 3 * 2`?

westnordost: But anyway, I like the suggestion with the variable, it is more descriptive. Will do

PR #4875, opened 2023-03-15
New quest: barbecue cover.

### LINKED ISSUE #4832: New quests: BBQ
### General
We can add 3 quests for [amenity=bbq](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbbq)

1. [covered](https://wiki.openstreetmap.org/wiki/Key:covered)
- Question asked: Is this place protected from rain? 
- Answers: yes/no

2. [fuel](https://wiki.openstreetmap.org/wiki/Key:fuel)
- Question asked: What is this grill powered by?
- Answers: wood/electric/charcoal

3. [wood_provided](https://wiki.openstreetmap.org/wiki/Key:wood_provided) (in case of fuel=wood)
- Question asked: Is wood provided by the operator of this bbq?
- Answers: yes/no

### Checklist

roptat:
This is my first attempt at creating a new quest. I think I chose one that's relatively easy :)

It partially fixes #4832 about barbecue.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -16,6 +16,7 @@ import de.westnordost.streetcomplete.quests.air_pump.AddBicyclePump
 import de.westnordost.streetcomplete.quests.atm_cashin.AddAtmCashIn
 import de.westnordost.streetcomplete.quests.atm_operator.AddAtmOperator
 import de.westnordost.streetcomplete.quests.baby_changing_table.AddBabyChangingTable
+import de.westnordost.streetcomplete.quests.bbq_cover.AddBbqCover
 import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_installation.AddBicycleBarrierInstallation
 import de.westnordost.streetcomplete.quests.barrier_bicycle_barrier_type.AddBicycleBarrierType
 import de.westnordost.streetcomplete.quests.barrier_type.AddBarrierOnPath
@@ -244,6 +245,8 @@ fun questTypeRegistry(
 
     10 to AddBenchBackrest(),
     11 to AddPicnicTableCover(),
+
+    157 to AddBbqCover(),
 
     12 to AddBridgeStructure(),
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bbq_cover/AddBbqCover.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bbq_cover/AddBbqCover.kt
@@ -1,0 +1,39 @@
+package de.westnordost.streetcomplete.quests.bbq_cover
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.OUTDOORS
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddBbqCover : OsmFilterQuestType<Boolean>() {
+
+    override val elementFilter = """
+        nodes with
+          amenity = bbq
+          and access !~ private|no
+          and !covered
+          and (!seasonal or seasonal = no)
+    """
+    override val changesetComment = "Specify whether barbecues are covered"
+    override val wikiLink = "Key:covered"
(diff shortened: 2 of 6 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -244,252 +245,253 @@ fun questTypeRegistry(
 
     10 to AddBenchBackrest(),
     11 to AddPicnicTableCover(),
+    12 to AddBbqCover(),
 
-    12 to AddBridgeStructure(),
+    13 to AddBridgeStructure(),
```
matkoniecz, line 250: note: you should not change numbers identifying other quests, see for example https://github.com/streetcomplete/StreetComplete/commit/c60c1bf8fbdbf677419b3b62e0dbd743bf0f8cc9

This numbers are supposed to be constant for each quest to keep presets saved as QR codes working

matkoniecz: 156 is taken already, 157 seems free and never used

roptat: oh, I thought they were for organizing the quests by priority, let me fix that quickly ^^'

matkoniecz: priority is decided by order of quests in list

have you used some documentation? (if yes, then it may need updating)

arrival-spring:
Not so much a question for @roptat but may need to be looked at before a new string gets added here. There seems to be duplication in strings, surely we could have one string "Is this covered", rather than separate ones for each of bicycle parking, motorcycle parking, picnic table and BBQ?

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
roptat: I was following https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#add-the-quest-to-the-list-of-active-ones which left it to my imagination what those numbers were. "Ordered by priority" the number was the priority. Also "As this point" -> "At this point" :). Note that I force-pushed the branch, it's now number 157.

FloEdelmann: @roptat Would you fancy opening another PR to improve that part in the documentation? Otherwise, I can do that.

roptat: OK, I tried to write something in #4878

roptat:
@arrival-spring, I'm a bit skeptical. If you mean something like "Is this %s covered", then you get a problem with languages whose grammar require a different form for some of the words depending on what could be covered.

An example could be French (I'm a native French speaker) where "this", "it" and "covered" need to agree in gender with the noun. So you get "Ce barbecue est-il couvert" but "**Cette** table de picnic est-**elle** couvert**e**"

mnalis:
> Not so much a question for @roptat but may need to be looked at before a new string gets added here. There seems to be duplication in strings, surely we could have one string "Is this covered", rather than separate ones for each of bicycle parking, motorcycle parking, picnic table and BBQ?

It would make sense to me, like we did for `lit` - it would reduce translation efforts.
E.g. before we had `Is this street lit?` /  `Is this path lit?` / `Is this cycleway lit?` / `Are those steps lit?` etc.
then we changed it just to `Is it lit here?` and then in second line below we listed what it is about `street`, `path`, `cycleway`, `steps`...

westnordost:
Yeah I also wonder if the `AddPicnicTableCover` couldn't instead be modified to have a different more generic wording ("Is this covered?") and the filter extended to also include other things. Then, no icon is needed for every further thing to add and also no extra string + translations.

westnordost:
Would you do this, @roptat, instead of a new quest? (I guess it would be a new PR then)

matkoniecz:
@roptat And sorry that it appeared only now, on review! This PR was still useful as it revealed this not really needed splitting it into multiple quests.

roptat:
Allight, I understand. I'll work on that this week.

roptat:
Actually, I've started implementing this. I have this filter (not sure about the language, so it might not even work ^^'):

```
nodes with                                                              
          (leisure = picnic_table                                                               
           or amenity = bicycle_parking                                         
           or amenity = motorcycle_parking                                      
           or amenity = bbq)                                                    
          and access !~ private|no                                              
          and !covered                                                          
          and bicycle_parking !~ shed|lockers|building                          
          and motorcycle_parking !~ shed|garage_boxes|building
```

This should select the same elements as all four quests together. Is that what you meant? I can work with the rest of the class, except for this:

```
override val achievements = listOf(OUTDOORS)
```
Since for motorcycle, it would be CAR and for bicycle it would be BICYCLIST. Is there a way to select the exact achievement this quest counts towards? Or should I keep them separate and only have a common quest for picnic tables and barbecue?

Not that the "is this lit" quests are in a similar situation, there are still three different quests.

matkoniecz:
Maybe keep separate one for bicycle parking and maybe motorcycle_parking and have separate for picnic_table/bbq/other new values?

bicycle parking one is quite popular and makes faster to handle it because you know what type of feature will be asked about - and is frequent enough to make it useful (and is implemented already)

westnordost:
Superseded by #4905

PR #4878, opened 2023-03-15
Improve contributing instructions

roptat:
Since I was confused by the numbers, here is a short sentence so others don't get confused in the same way, and a typo fix in the same section.


mnalis (approved):

CONTRIBUTING_A_NEW_QUEST.md
```diff
@@ -140,9 +140,11 @@ See [this step](https://github.com/matkoniecz/StreetComplete_quest_creation_tuto
 
 Adjust [QuestsModule.kt](app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt) file. It contains a big list of active quests, ordered by priority. Read [what governs their priority](app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt#L172-L195) but do not worry too much, it can be tweaked later.
 
+Each quest is associated with a number in this list. These numbers are used to identify the quest uniquely and can be used to save presets as QR codes. When adding a new quest, use a new number that is not yet in the list. Put the quest in order of priority, even if it means the numbers are not sorted.
```
westnordost, line 143:
```suggestion
Each quest is associated with a number in this list. These numbers are used to identify the quest uniquely and can be used to save presets as QR codes. When adding a new quest, use the next consecutive number that is not yet in the list. Put the quest in order of priority, even if it means the numbers are not sorted.
```

PR #4879, opened 2023-03-16
Ask toilet part wheelchair access only if wheelchair=limited

Helium314:
See https://github.com/streetcomplete/StreetComplete/pull/4868#issuecomment-1467439875 and below, where @matkoniecz and @mnalis came to the conclusion this would make sense.

In short: the quest is unnecessary if `wheelchair=yes`, and the answer is very likely to be _no_ if `wheelchair=no`.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToiletsPart.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToiletsPart.kt
@@ -11,8 +11,8 @@ class AddWheelchairAccessToiletsPart : OsmFilterQuestType<WheelchairAccessToilet
 
     override val elementFilter = """
         nodes, ways with
-         wheelchair = limited
-         (
+          wheelchair = limited
+          and (
            toilets = yes
            or !toilets and (
              amenity ~ restaurant|pub|bar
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToiletsPart.kt
```diff
@@ -11,6 +11,7 @@ class AddWheelchairAccessToiletsPart : OsmFilterQuestType<WheelchairAccessToilet
 
     override val elementFilter = """
         nodes, ways with
+         wheelchair = limited
```
matkoniecz, line 14: have you tested it? From looking without testing I suspected that `and` is missing

matkoniecz (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToiletsPart.kt
Helium314: Thanks! I indeed did test it, but not this exact change...
This happens when you lazily type changes into github from memory

westnordost:
Uh, does this overlap with @arrival-spring new PR? How about reverting the previously merged PR to start of with a blank slate again first?

Helium314:
> Uh, does this overlap with @arrival-spring new PR?

Not with #4881, which focuses on issues with the answers. This PR is only about element selection.

arrival-spring:
> How about reverting the previously merged PR to start of with a blank slate again first?

That might not be a bad idea, but I think mine is now fine

PR #4880, opened 2023-03-16
Request map render after setting pins and elements

### LINKED ISSUE #2780: Quest icons not shown until zooming out (or maybe other map interaction), happened only on emulator so far
Scan for quests in afd4dcedb
Wait long time. Wait longer.
Zoom out - and quests finally appear.

**Versions affected**
afd4dcedb

Peek 2021-04-21 19-58

Helium314:
fixes #2780

This fixes rare cases of the map not being updated e.g. after switching overlays.
Usually the map is re-drawn frequently as consequence of user interaction, position update or view direction update. But if none of this happens, the effects are as described in #2780.

I found no noticeable performance impact (on S4 mini).


asmaael33 (approved):
gg

PR #4881, opened 2023-03-16
Fix wheelchair toilet tagging

arrival-spring:
Thanks @Helium314 for spotting this

See https://github.com/streetcomplete/StreetComplete/pull/4868#discussion_r1137291635

In reducing duplication, it incidentally changed it so that `amenity=toilets` would get tagged with `toilets:wheelchair` rather than `wheelchair`.

This reverts that change.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
@@ -31,7 +31,7 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAns
     override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
         when (answer) {
             is WheelchairAccessToilets -> tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
-            is NoToilet -> answer.applyTo(tags)
+            is NoToilet -> throw IllegalStateException()
         }
     }
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
@@ -30,10 +30,7 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAns
 
     override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
         when (answer) {
-            is WheelchairAccessToilets -> {
-                tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
-                tags["toilets"] = "yes"
-            }
+            is WheelchairAccessToilets -> tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
             is NoToilet -> answer.applyTo(tags)
         }
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
```diff
@@ -28,6 +29,12 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAns
     override fun createForm() = AddWheelchairAccessToiletsForm()
 
     override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
-        answer.applyTo(tags)
+        when (answer) {
+            is WheelchairAccessToilets -> {
+                tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
+                tags["toilets"] = "yes"
```
FloEdelmann, line 35: Is adding `toilets = yes` on `amenity = toilets` deliberate?

arrival-spring: Ah, no, this hasn't gone so well. We shouldn't offer a no toilet option when asking about `amenity=toilets` anyway. I'll fix this

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
```diff
@@ -28,6 +29,9 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAns
     override fun createForm() = AddWheelchairAccessToiletsForm()
 
     override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
-        answer.applyTo(tags)
+        when (answer) {
+            is WheelchairAccessToilets -> tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
+            is NoToilet -> answer.applyTo(tags)
```
arrival-spring, line 34: We shouldn't ever reach this point as the NoToilet answer will not be shown for toilets, so is it best to do something else?

FloEdelmann: Maybe throw instead?

```suggestion
            is NoToilet -> throw IllegalStateException()
```

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/WheelchairAccessToiletsAnswer.kt
```diff
@@ -7,12 +7,6 @@ sealed interface WheelchairAccessToiletsAnswer
...
-    is WheelchairAccessToilets -> {
-        tags.updateWithCheckDate("toilets:wheelchair", access.osmValue)
-        tags["toilets"] = "yes"
-    }
-    NoToilet -> {
-        tags.updateWithCheckDate("toilets", "no")
-    }
+fun NoToilet.applyTo(tags: Tags) {
+    tags.updateWithCheckDate("toilets", "no")
 }
```
westnordost, line 12: just inline this and remove the function

app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessToilets.kt
```diff
@@ -28,6 +29,9 @@ class AddWheelchairAccessToilets : OsmFilterQuestType<WheelchairAccessToiletsAns
     override fun createForm() = AddWheelchairAccessToiletsForm()
 
     override fun applyAnswerTo(answer: WheelchairAccessToiletsAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
-        answer.applyTo(tags)
+        when (answer) {
+            is WheelchairAccessToilets -> tags.updateWithCheckDate("wheelchair", answer.access.osmValue)
+            is NoToilet -> throw IllegalStateException()
```
westnordost, line 34: Hm on second thought, if `NoToilet` is not a valid choice, but only `WheelchairAccessToilets` is, then this should also be the generic type of this class.

westnordost:
I've done the changes I deemed necessary myself on master

arrival-spring:
Thanks, sorry I hadn't got round to it yet

PR #4884, opened 2023-03-18
Fix Crash: GridLayoutManager already attached to RecyclerView

### LINKED ISSUE #4869: Crash: GridLayoutManager already attached to RecyclerView
I received a SCEE crash report about `GridLayoutManager` being already attached to a `RecyclerView` (in code that is unchanged from StreetComplete):

       java.lang.IllegalArgumentException: LayoutManager androidx.recyclerview.widget.GridLayoutManager@42dde9a is already attached to a RecyclerView: androidx.recyclerview.widget.RecyclerView{676e3cb VFED..... .......D 0,53-980,2207 #7f0901d1 app:id/list}, adapter:de.westnordost.streetcomplete.view.image_select.GroupedImageSelectAdapter@40907a8, layout:androidx.recyclerview.widget.GridLayoutManager@42dde9a, context:de.westnordost.streetcomplete.screens.MainActivity@d280748

in
https://github.com/streetcomplete/StreetComplete/blob/4c01d0ff5a84ce71983b1fed0db26594aa00ee0b/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestForm.kt#L65

This is might be caused by `onViewCreated` being called twice after the form is created, which apparently can happen: https://stackoverflow.com/questions/23593965/when-would-fragments-ondestroyview-be-called-but-it-wouldnt-be-destroyed

Probably the adapter should be removed from the view in `onDestroyView`, or be created in `onCreateView` or `onViewCreated` (which would require a lot of quest forms to be adjusted, as they use `imageSelector` in `onCreate`).
Same issue likely applies to `AImageListQuestForm`.

westnordost:
fixes #4869

@Helium314

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestForm.kt
@@ -87,13 +87,6 @@ abstract class AGroupedImageListQuestForm<I, T> : AbstractOsmQuestForm<T>() {
         binding.list.adapter = imageSelector
     }
 
-    override fun onDestroyView() {
-        super.onDestroyView()
-        /* must be cleared because the layout manager holds a reference to the imageSelector
-           (through the spanSizeLookup) which exists outside of the view lifecycle */
-        binding.list.layoutManager = null
-    }
-
     private fun scrollTo(index: Int) {
         val item = binding.list.layoutManager?.findViewByPosition(index) ?: return
         val itemPos = IntArray(2)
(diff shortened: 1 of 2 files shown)


Helium314:
Definitely looks more to the point than my ideas.
I'm don't have much experience in (intentionally) finding / fixing memory leaks, but I guess it's fine now.

The `layoutManager` is now cleared `onDestroyView` as it's only referenced in the binding, right?

westnordost:
No, it is not cleared, I removed this again in a subsequent commit. But the `RecyclerView` is cleared (automatically, because the view root is dropped), and thus also its `layoutManager`. Since the `layoutManager` is not owned (i.e. is a field of or is used in any lambda owned by another object that is not cleared), it should be fine.

PR #4891, opened 2023-03-20
Use collectSurfaceDescriptionIfNecessary in AddPathSurfaceForm

Helium314:
This appears to be an oversight, but not 100% sure.


matkoniecz:
have you tested it? or is it requiring a test is it working?

Helium314:
I tested, and behavior looks the same: no difference when answering a path surface quest, and no difference in applied tags.
All other surface quests already use `collectSurfaceDescriptionIfNecessary`.

PR #4892, opened 2023-03-20
Ask sidewalk quest if nearby paths are not foot accessible

Helium314:
Ways not accessible by foot should be excluded from `maybeSeparatelyMappedSidewalksFilter`, as they are most likely not sidewalks.

I don't know how to find a situation where this is relevant, so I only tested in SCEE where I can change the tags of nearby paths for creating such a situation.


westnordost:
True, good idea! Could you format the filter into two lines for better readability?

westnordost:
Btw, such small enhancements that do not need to be discussed you can directly commit on master no problem. @matkoniecz also does that often

PR #4897, opened 2023-03-23
Store tangram geometry in pin

Helium314:
Currently a tangram `Point` is created for each pin every time `PinsMapComponent.set` is called.

By storing the tangram geometry in the pin, this can be avoided, which improves performance of `PinsMapComponent.set` by roughly 30%, and reduces need for garbage collection.
Actually the change is only used when using `QuestPinsManager.updateQuestPins` (mostly after answering / hiding a quest), other calls to `PinsMapComponent.set` always use new `Pin`s. Still it's an improvement, as solving quests happens quite often.

(a similar change is possible for `StyleableOverlayMapComponent`)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/PinsMapComponent.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/PinsMapComponent.kt
@@ -21,18 +21,7 @@ class PinsMapComponent(private val ctrl: KtMapController) {
 
     /** Show given pins. Previously shown pins are replaced with these.  */
     fun set(pins: Collection<Pin>) {
-        pinsLayer.setFeatures(pins.map { pin ->
-            // avoid creation of intermediate HashMaps.
-            val tangramProperties = listOf(
-                "type" to "point",
-                "kind" to pin.iconName,
-                "importance" to pin.importance.toString()
-            )
-            val properties = HashMap<String, String>()
-            properties.putAll(tangramProperties)
-            properties.putAll(pin.properties)
-            Point(pin.position.toLngLat(), properties)
-        })
+        pinsLayer.setFeatures(pins.map { it.tangramPoint })
     }
 
     /** Clear pins */
@@ -51,4 +40,17 @@ data class Pin(
     val iconName: String,
     val properties: Collection<Pair<String, String>> = emptyList(),
     val importance: Int = 0
-)
+) {
+    val tangramPoint by lazy {
+        // avoid creation of intermediate HashMaps.
+        val tangramProperties = listOf(
+            "type" to "point",
+            "kind" to iconName,
+            "importance" to importance.toString()
+        )
+        val props = HashMap<String, String>()
+        props.putAll(tangramProperties)
+        props.putAll(properties)
+        Point(position.toLngLat(), props)
+    }
+}


westnordost:
This sounds good, the implementation is also good.

Two ideas for further improvement:
1. The constructor of `com.mapzen.tangram.geometry.Point` expects a `java.util.Map` but then the only thing it does with that is to iterate through its `entrySet()` to put it into an array. Implementing a `ArrayMap` that implements `Map` but just consists of a `List` of key-value pairs would take up less space in memory (probably doesn't really matter) and remove the necessity of calculating the `hashCode` for every element inserted into the map.
2. As the size of the map is known at construction time, it could be passed into the constructor to avoid that the map needs to grow when the items are inserted.

Helium314:
> I dimly remember that something similar already came up in the past - ... maybe it was not worth it or there was no discernible gain at all?

I remember that too. There was no measureable performance improvement, and the memory improvement is probably not worth it: using [this info](https://stackoverflow.com/questions/20175225/hash-map-memory-overhead) we may save somthing like 8 bytes per entry, and 16 bytes per map. For a high density area with 10k pins in view, this would be ~720 kB.
We could likely save more memory (and improve performance, and have something that remains after migration to MapLibre) by not creating pins that are hidden by others (same position lower priority quest).

But if it's really simple to do... (I didn't look at it)

> As the size of the map is known at construction time, it could be passed into the constructor to avoid that the map needs to grow when the items are inserted.

Though for quests there are 7 entries, and default initial capacity is 16. Does unused capacity use memory?

westnordost:
> Though for quests there are 7 entries, and default initial capacity is 16. Does unused capacity use memory?

Yes, of course. But the more important thing is that it doesn't grow.

westnordost:
> We could likely save more memory (and improve performance, and have something that remains after migration to MapLibre) by not creating pins that are hidden by others (same position lower priority quest).

This would help performance in areas with high density of quests, but would require putting the quests/quest pins into a HashMap<LatLon, Quest(Pin)> first

Helium314:
> a similar change is possible for StyleableOverlayMapComponent

I was about to [start a PR](https://github.com/Helium314/SCEE/commit/635bdffa8d98c17524dab76fd11f8ac3c5f12d9b) when I noticed that this is might need further changes, as the road colors [depend on theme](https://github.com/streetcomplete/StreetComplete/blob/37633303a29dd3ac97ad18432fa80d111e6bec62/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/StyleableOverlayMapComponent.kt#L71-L72).
So I'll test whether it works anyway (not sure what exactly happens on light/dark switch), and if there are issues I think it's too much work.

westnordost:
You could not set it in the `lazy` function and let `StyleableOverlayMapComponent::set` supply it if `style.stroke == null && (style.strokeLeft != null || style.strokeRight != null)`

westnordost:
Ah nvm, not that easy as can't access the props in the tangram geometry once ist has been constructed

Helium314:
I added some logging and waited for an automatic theme switch. `StyleableOverlayManager` is destroyed on auto-switch, so the `StyledElement`s are created again after the switch anyway.
When switching manually, one has to enter the settings. And when doing so, `onStop` is called and thus `StyledElement`s are cleared.

westnordost:
Sounds like this could be done then

Helium314:
>  I dimly remember that something similar already came up in the past - ... maybe it was not worth it or there was no discernible gain at all?

Found it: https://github.com/streetcomplete/StreetComplete/pull/4152/files
For a dense area with several thousand quests this saves only 200 kB.

What I found  to save a lot more is string interning when loading elements / tags from database: 4 MB saved for 38k elements, at the cost of ~5% slower loading.

westnordost:
Hmm... yeah, that sounds like it make sense. On the other hand, memory is usually in ample supply so that memory-optimization at the cost of CPU is rarely worth it. 
Also, I am a bit skeptic if it is a good idea to intern every trash value that can be found in the OSM database. Maybe just the keys?

Anyway, if you are looking into interning, the first thing that comes to my mind is to intern the string elements (keys, values, but probably not the regex stuff) of the filter syntax during parsing.

Helium314:
Some more detailed results (averaged over a few runs, memory use after GC):
* Current state
  * fetch nodes: 11.4 s
  * fetch DataWithGeometry: 23.6
  * memory use: 39.4 MB
* Creating `HashMap`s with with deserialized tags (this is necessary for actually using interning)
  * fetch nodes: 10.9 s
  * fetch DataWithGeometry: 22.8
  * memory use: 38.4 MB
* Intern keys only
  * fetch nodes: 11.9 s
  * fetch DataWithGeometry: 24.1
  * memory use of java: 36.4 MB
* Intern keys and values
  * fetch nodes: 12.6 s
  * fetch DataWithGeometry: 25.6
  * memory use of java: 34.8 MB

I have no clue why creating an additional `HashMap`s is actually faster than just keeping the map returned by `decodeFromString`...
But anyway, I don't have a strong opinion on what to do here, or whether to do anything.

westnordost:
What are deserialized tags?

westnordost:
I think it is not really worth it.

Helium314:
Alright, then let's leave it as it is

PR #4905, opened 2023-03-26
Factorize picnic table and barbecue cover quests

roptat:
After discussion in #4875 we discussed the possibility of grouping "covered" quests together. This PR implements that for grouping picnic table and barbecue cover quests. I haven't changed the icon, but of course it should be changed.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -13,6 +13,7 @@ import de.westnordost.streetcomplete.quests.address.AddHousenumber
 import de.westnordost.streetcomplete.quests.air_conditioning.AddAirConditioning
 import de.westnordost.streetcomplete.quests.air_pump.AddAirCompressor
 import de.westnordost.streetcomplete.quests.air_pump.AddBicyclePump
+import de.westnordost.streetcomplete.quests.amenity_cover.AddAmenityCover
 import de.westnordost.streetcomplete.quests.atm_cashin.AddAtmCashIn
 import de.westnordost.streetcomplete.quests.atm_operator.AddAtmOperator
 import de.westnordost.streetcomplete.quests.baby_changing_table.AddBabyChangingTable
@@ -104,7 +105,6 @@ import de.westnordost.streetcomplete.quests.parking_access.AddParkingAccess
 import de.westnordost.streetcomplete.quests.parking_fee.AddBikeParkingFee
 import de.westnordost.streetcomplete.quests.parking_fee.AddParkingFee
 import de.westnordost.streetcomplete.quests.parking_type.AddParkingType
-import de.westnordost.streetcomplete.quests.picnic_table_cover.AddPicnicTableCover
 import de.westnordost.streetcomplete.quests.pitch_lit.AddPitchLit
 import de.westnordost.streetcomplete.quests.place_name.AddPlaceName
 import de.westnordost.streetcomplete.quests.playground_access.AddPlaygroundAccess
@@ -243,7 +243,7 @@ fun questTypeRegistry(
     9 to AddCarWashType(),
 
     10 to AddBenchBackrest(),
-    11 to AddPicnicTableCover(),
+    11 to AddAmenityCover(),
 
     12 to AddBridgeStructure(),
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/amenity_cover/AddAmenityCover.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/amenity_cover/AddAmenityCover.kt
@@ -1,4 +1,4 @@
-package de.westnordost.streetcomplete.quests.picnic_table_cover
+package de.westnordost.streetcomplete.quests.amenity_cover
 
 import de.westnordost.streetcomplete.R
 import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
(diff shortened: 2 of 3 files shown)


westnordost (changes requested):
Looks good! However, since you renamed this quest, you should add
```
"AddPicnicTableCover"                 to AddAmenityCover::class.simpleName!!,
```
to `typeAliases` in `AchievementsModule` so that quests solved so far of `AddPicnicTableCover` are correctly counted towards `AddAmenityCover` in the statistics

Helium314:
Renaming a quest will make the app crash if there are edits with the [old (now unknown) name](https://github.com/streetcomplete/StreetComplete/blob/37633303a29dd3ac97ad18432fa80d111e6bec62/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsDao.kt#L138-L139) in the database. 
Solution would be to do a db upgrade and rename the database entries.

westnordost:
Oh, I didn't know that.  So you have a stacktrace?

Am 27. März 2023 06:41:09 MESZ schrieb Helium314 ***@***.***>:
>Renaming a quest will make the app crash if there are edits with the [old (now unknown) name](https://github.com/streetcomplete/StreetComplete/blob/37633303a29dd3ac97ad18432fa80d111e6bec62/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsDao.kt#L138-L139) in the database. 
>Solution would be to do a db upgrade and rename the database entries.
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/4905#issuecomment-1484479917
>
>Message ID: ***@***.***>
-- 

westnordost:
> The previous db update was done for the same reason...

No, that was a different reason. The quest hadn't been renamed.

Anyway.

I will make it so that the aliases are handled in the questTypeRegistry itself, so that it will not be a problem.

westnordost:
> I will make it so that the aliases are handled in the questTypeRegistry itself, so that it will not be a problem.

Hm, not that straightforward for various reasons... It's also easier to deal with such quest type aliases just at their entry point to the app (i.e. when the stuff from the backend is parsed) then to carry it all through. The easiest solution is as you say to rename it or not change the name. Since @roptat already put in the work to change the name, I'll just do the DB upgrading.

westnordost:
Hmm, do you think this extra code is worth it, @Helium314 ? It's in quite a lot of tables, after all. The other alternative would be to not rename the quest type class (and I guess, generally in the future avoid it)

```diff
--- a/app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteSQLiteOpenHelper.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteSQLiteOpenHelper.kt
@@ -3,6 +3,7 @@ package de.westnordost.streetcomplete.data
 import android.content.Context
 import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteOpenHelper
+import androidx.core.content.contentValuesOf
 import de.westnordost.streetcomplete.data.download.tiles.DownloadedTilesTable
 import de.westnordost.streetcomplete.data.osm.created_elements.CreatedElementsTable
 import de.westnordost.streetcomplete.data.osm.edits.ElementEditsTable
@@ -186,7 +187,23 @@ class StreetCompleteSQLiteOpenHelper(context: Context, dbName: String) :
         if (oldVersion <= 7 && newVersion > 7) {
             db.delete(ElementEditsTable.NAME, "${ElementEditsTable.Columns.QUEST_TYPE} = 'AddShoulder'", null)
         }
+        if (oldVersion <= 8 && newVersion > 8) {
+            val old = "AddPicnicTableCover"
+            val new = "AddAmenityCover"
+            db.renameValue(ElementEditsTable.NAME, ElementEditsTable.Columns.QUEST_TYPE, old, new)
+            db.renameValue(OsmQuestTable.NAME, OsmQuestTable.Columns.QUEST_TYPE, old, new)
+            db.renameValue(OsmQuestsHiddenTable.NAME, OsmQuestsHiddenTable.Columns.QUEST_TYPE, old, new)
+            db.renameValue(VisibleQuestTypeTable.NAME, VisibleQuestTypeTable.Columns.QUEST_TYPE, old, new)
+            db.renameValue(OpenChangesetsTable.NAME, OpenChangesetsTable.Columns.QUEST_TYPE, old, new)
+            db.renameValue(QuestTypeOrderTable.NAME, QuestTypeOrderTable.Columns.BEFORE, old, new)
+            db.renameValue(QuestTypeOrderTable.NAME, QuestTypeOrderTable.Columns.AFTER, old, new)
+            db.renameValue(EditTypeStatisticsTables.NAME, EditTypeStatisticsTables.Columns.ELEMENT_EDIT_TYPE, old, new)
+        }
     }
 }

-private const val DB_VERSION = 8
+private const val DB_VERSION = 9
+
+private fun SQLiteDatabase.renameValue(table: String, column: String, oldValue: String, newValue: String) {
+    update(table, contentValuesOf(column to newValue), "$column = ?", arrayOf(oldValue))
+}
```

matkoniecz:
I think that it is not an outrageous amount of code and avoid confusing names persisting forever.

Helium314:
> I think that it is not an outrageous amount of code and avoids confusing names persisting forever.

I agree, having this is helpful in every time quests are renamed.

>> The previous db update was done for the same reason...
>
> No, that was a different reason. The quest hadn't been renamed.

I meant to avoid SC crashing when the name is not found in `questTypeRegistry`.

westnordost:
Alright, did this then.

westnordost (approved):

westnordost:
Will be included in the next release (not v52) because it contains new strings.

PR #4906, opened 2023-03-26
Store tangram geometries in styled element 

Helium314:
This is essentially the same as #4897, but geometry is stored in a var instead of created by lazy because creating the geometries requires `resources`.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/StyleableOverlayMapComponent.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/StyleableOverlayMapComponent.kt
@@ -154,5 +154,7 @@ data class StyledElement(
     val geometry: ElementGeometry,
     val style: Style
 ) {
+    // geometries may contain road color, which depends on current theme
+    // however, storing is not an issue as styled elements are cleared on theme switch (both automatic and manual)
     var tangramGeometries: List<Geometry>? = null
 }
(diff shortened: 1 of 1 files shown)


westnordost:
Wouldn't you find it more consistent if the creation of the tangram geometries was done in a function within StyledElement, just like for the quest pins? I.e.
```
class StyledElement {
    private var tangramGeometries: List<Geometry>? = null

    fun getTangramGeometries(resources: Resources): List<Geometry> {
        // TODO: Uh by the way, does this need to be thread-safe? Because this isn't
        if (tangramGeometries == null) {
            tangramGeometries = createTangramGeometries(resources)
        }
        return tangramGeometries 
    }
}
```

Helium314:
> Wouldn't you find it more consistent if the creation of the tangram geometries was done in a function within StyledElement, just like for the quest pins?

Probably it would be. But for this to work the `StyledElement` needs access to `darkenedColors` and `transparentColors`.

> Uh by the way, does this need to be thread-safe? Because this isn't

`StyledElement`s are always created and set synchronized to `mapDataInView`, so I think it's fine the way it is.

westnordost:
> I think it's fine the way it is.

Does this also refer to the suggestion about the consistency? If yes, do you not find it important that it is consistent or do you find it more logical to have the tangram-geometry-creation code not in the data class in general (i.e. also for the pins then)

Helium314:
> Does this also refer to the suggestion about the consistency?

No, this was only about whether it needs to be thread-safe.

> Do you not find it important that it is consistent or do you find it more logical to have the tangram-geometry-creation code not in the data class in general (i.e. also for the pins then)

Fo me it doesn't matter that much. The difference is that the geometry for a pin only depends on what is in the pin, while the geometry for a styled element also depends on `resources` (and the color hashmaps, albeit only for performance reasons).
When I wrote the code I found that the difference justifies some inconsistency, but if you want it gone, then it would be easier to adjust `PinsMapComponent`.

westnordost:
Ok, it's fine then. But I moved the creation of the List<Geometry> into an own function now so that the `set` function does not do two things (lazy-logic + creation).

Helium314:
> I moved the creation of the List into an own function now so that the `set` function does not do two things (lazy-logic + creation).

Right, that definitely makes it easier to see what's going on.

PR #4909, opened 2023-03-29
Hairdresser quest

mnalis:
Closes: https://github.com/streetcomplete/StreetComplete/issues/4833

TODO:
- [x] Discuss: does it need resurvey?
- [x] Discuss: better English  for question than _"What are the signed customers of this hairdresser?"_ ? Perhaps _"Which customers are indicated on the sign for this hairdresser?"_ ?
- [x] verify that everything works
  yes, debug APK at https://github.com/mnalis/StreetComplete/actions/runs/4600306640
- [x] uses icon made from existing scissors.svg


mnalis:
Help wanted:
- do people think this quest needs resurvey? Over here I don't think so, it seems static... If it ever changes, most probably the owner, name etc. changed anyway, so it is _"other shop"_
- suggestions for better English for quest title (as indicated in the PR summary) ?

westnordost:
I think it does not need resurvey.

For the icon, you can just use scissors. There is already a scissors SVG in the resources, you can reuse that.

mnalis:
I'm still struggling with good title for the quest being asked:

- **What are the signed customers of this hairdresser?** 
  (current from [#4833](https://github.com/streetcomplete/StreetComplete/issues/4833#issuecomment-1463752038), but sounds somewhat strange to me)
- Which customers are indicated on the sign for this hairdresser?
- What customers are signed on this hairdresser?
- What customers are featured on this hairdresser's sign?

The idea is to somehow indicate that user should enter what is marked on the sign / doors of the hairdresser, and not to imply that other customers (e.g. nonbinary) are forbidden/unwelcome  (see https://github.com/streetcomplete/StreetComplete/issues/4833)

so solutions like:
- Which customers does this hairdresser serve?
- What customers are serviced here?

are probably less then ideal.

Any suggestions, especially from native English speakers?
@Discostu36 @peternewman @dbdean 

(other than this wording issue, this PR is complete)

westnordost:
I don't quite understand why you need the "is signed as..." wording. I think it should be Somerton along the lines of "for whom is hairdresser?" ... with answers like :
-  men
-  women
-  anyone / not signed

After all, it doesn't have to be always signed, it could be clear from the context, logos or names. E.g. a hairdresser with "barber" in its name is likely for men only (that is,  for men's *hairstyles*) and a hairdresser that advertises with "beauty salon, fabulous nails and hairstyles" is something that is rather only targeted at women.

Am 30. März 2023 01:00:56 MESZ schrieb Matija Nalis ***@***.***>:
>I'm still struggling with good title for the quest being asked:
>
>- **What are the signed customers of this hairdresser?** 
>  (current from [#4833](https://github.com/streetcomplete/StreetComplete/issues/4833#issuecomment-1463752038), but sounds somewhat strange to me)
>- Which customers are indicated on the sign for this hairdresser?
>- What customers are signed on this hairdresser?
>- What customers are featured on this hairdresser's sign?
>
>The idea is to somehow indicate that user should enter what is marked on the sign / doors of the hairdresser, and not to imply that other customers (e.g. nonbinary) are forbidden/unwelcome  (see https://github.com/streetcomplete/StreetComplete/issues/4833)
>
>so solutions like:
>- Which customers does this hairdresser serve?
>- What customers are serviced here?
>
>are probably less then ideal.
>
>Any suggestions, especially from native English speakers?
>@Discostu36 @peternewman @dbdean 
>
>(other than this wording issue, this PR is complete)
>
>-- 
>https://github.com/streetcomplete/StreetComplete/pull/4909#issuecomment-1489449254
>
>Message ID: ***@***.***>
-- 

mnalis:
> I think it should be Somerton along the lines of "for whom is hairdresser?"

OK. ChatGPT however says :smiley_cat: :

> While the phrase "For whom is this hairdresser?" may be technically correct grammatically, it does not sound natural or idiomatic in English. A more natural way to ask this question would be "Who does this hairdresser serve?" or "Which customers does this hairdresser provide services for?"

And gives more alternatives, of which I've settled for short and simple _**"Which customers visit this hairdresser?"**_ 

Unless someone feels the need to suggest something better,  I'll mark this PR as ready tomorrow.

app/src/main/res/drawable/ic_quest_hairdresser.xml
```diff
@@ -0,0 +1,35 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android" xmlns:aapt="http://schemas.android.com/aapt"
```
westnordost, line 1: You should include the SVG source graphic for the quest too

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/Hairdresser.kt
```diff
@@ -0,0 +1,8 @@
+package de.westnordost.streetcomplete.quests.hairdresser
+
+enum class Hairdresser(val isMale: Boolean, val isFemale: Boolean) {
+    NOT_SIGNED(false, false),
+    ONLY_FEMALE(false, true),
+    ONLY_MALE(true, false),
+    MALE_AND_FEMALE(true, true),
+}
```
westnordost, line 8: Why isMale=false, isFemaile=false for "not signed"?

You should use a sealed interface here, i.e. 
sealed interface HairdresserAnswer
enum class Hairdresser : HairdresserAnswer { ... }
object HairdresserNotSigned : HairdresserAnswer

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/AddHairdresser.kt
```diff
@@ -0,0 +1,46 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.IS_SHOP_OR_DISUSED_SHOP_EXPRESSION
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddHairdresser : OsmFilterQuestType<Hairdresser>() {
```
westnordost, line 14: Hm, the name doesn't really fit. The quest is not about adding hairdressers.

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/AddHairdresser.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    override val elementFilter = """
+        nodes, ways with
+          (
+              shop = hairdresser
+              and !female and !male and !unisex
```
westnordost, line 20: I though the outcome of the discussion was that when unisex is specfied, it is ignored due to it being ambiguous and always re-asked?

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/AddHairdresser.kt
mnalis: ok, renamed to HairdresserCustomers instead

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/Hairdresser.kt
mnalis: Hm, because those flags are about what **is** signed? 

So if only "for male" is signed on the hairdresser, only `isMale=true` (as it was signed), and `isFemale=false` (as it was **not** signed). Same logic applies if nothing is signed: both `isMale=false` and `isFemale=false`, because none of them were signed.

But if you insist, I'll investigate how to use that `sealed interface` thingy and use that (only for NOT_SIGNED or for all of them?)

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/AddHairdresser.kt
mnalis: Hm, I asked which of the two choices you preferred in https://github.com/streetcomplete/StreetComplete/issues/4833#issuecomment-1464557177 and from your answer https://github.com/streetcomplete/StreetComplete/issues/4833#issuecomment-1483809818 I concluded that you prefer option (1) _"permanently ignore any hairdressers which have unisex=* tag set (thus preserving any ambiguous meaning that original mapper may have intended)"_ because _"otherwise we'd be fighting against iD windmills"_ (i.e. we'd be removing `unisex=yes` and adding `male=yes`+`female=yes`; whilst iD would be adding `unisex=yes` and removing `male=yes`+`female=yes` ? 

Or did I misunderstood and you prefer option (2) _"quest should be asked regardless of whether unisex=* is set, and when user has chosen current state of the hairdresser set male=* / female=* tags and remove unisex=* (thus fixing ambiguity as confirmed by new mapper)"_? 

Or did you want some third (unmentioned) option?

app/src/main/res/drawable/ic_quest_hairdresser.xml
mnalis: added

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/AddHairdresser.kt
westnordost: The "agree to your statement" referred to 

> preferable to make situation unambiguous when we have the chance with mapper on the ground, rather then to simply "sweep the problem under the rug"

Regarding the remark about fighting against windmills, my hope would be that how iD currently handles these things would be checked and a PR/ticket be opened at iD to straighten this out (if necessary) in order that iD does not overwrite/delete unambiguous data with ambiguous data.

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/Hairdresser.kt
westnordost: Well, if you think it is fine this way, we can leave it at that. However, since you haven't used sealed interfaces so far, maybe you should look at some other quests that use it and only decide then whether that approach would not be cleaner.

app/src/main/java/de/westnordost/streetcomplete/quests/hairdresser/AddHairdresser.kt
mnalis: Ok! I'll then implement `(2)`  tonight (and try to look into iD later)

mnalis: implemented replacing `unisex` tag. I'll lookup iD later (but it is not holding up this PR, which should be ready now)

westnordost:
I understand that you wanted to create an issue or (better) a PR over at the iD tagging schema repo if the way how iD tags it conflicts with this quest?

mnalis:
> I understand that you wanted to create an issue or (better) a PR over at the iD tagging schema repo if the way how iD tags it conflicts with this quest?

Yes, I'm wading through existing iD gender issues currently, so I have background and can reference them in new issue (or contribute to existing one with same problem, if present) 

iD uses radio button so only one of fields `male` / `female` / `unisex` can be selected, but it does indicate by background color when multiple (e.g. `male`+`female`) are selected:

mf2

So while not actually conflicting, it definitely does not handle it in the best possible way...

mnalis:
While iD does show multiple gender tags, it does it in _subtle_ way that is (IMHO) somewhat likely to be missed by its users.

The suggestion for UI improvement (i.e. use checkboxes instead of radiobuttons, like JOSM and Vespucci do) has been suggested at https://github.com/openstreetmap/iD/issues/9591

amandasaurus:
I agree about not using “whom”. Many people who learn English as a second language, use “whom” because in the past that was Proper Grammer™, however it sounds very old fashioned, and unnatural.

> And gives more alternatives, of which I've settled for short and simple "Which customers visit this hairdresser?"

A problem with this is that it refers to the type of customers, which regular visit that place. Someone standing on the street, can't know what happens regularly, they can only look at the sign. They cannot know the regular occurances.

I think “Who is this hairdresser for?” (while showing the gender options) is pretty clear & better.

Issue #4928, opened 2023-04-03
"Drinking Water" quest should not trigger on disused/historic wells


BenWiederhake:
StreetComplete asks "Is the water potable here?" even on wells that are already marked as `disused=yes, historic=yes`, so marking them as non-potable water seems unhelpful.

**How to Reproduce**
Step 1: Be in the vicinity of a disused/historic well, and search for quests.
Step 2: Observe that StreetComplete marks the well as a quest item, even though adding such information is not helpful.

The [Overpass Turbo **Wizard**](http://www.overpass-turbo.eu/) makes it easy to combine the quest trigger (taken from `AddDrinkingWater.kt`) and add `and (disused = yes or historic = yes)` to find all the places where StreetComplete will trigger a quest on a disused/historic well.

Here's the **Wizard** query, since the full Overpass query is long to paste here:
`(man_made = water_tap or man_made = water_well or natural = spring) and access !~ "private|no" and indoor != yes and drinking_water is null and "drinking_water:legal" is null and amenity != drinking_water and (seasonal is null or seasonal = no) and (disused = yes or historic = yes)`

**Expected Behavior**
StreetComplete should not ask whether the non-existent water in a historic well is potable.

**Versions affected**
I'm using SC 51.1, but since the query in `master` still doesn't mention `disused` or `historic`, I believe that also the latest version is affected.

EDIT: Original feature see #549

matkoniecz:
`historic=* man_made=water_well` does not really mark inactive water well...

`ruins=yes` `disused=yes` are also quite nasty in this context and ideally would not be used but maybe skipping such items would make sense (but in such case that does not mean that this tagging scheme is a good idea)

mnalis:
I agree that `disused=yes` + `man_made=water_well` (or others) is bad tagging. 
(It should've been `disused:man_made=water_well` etc.)

But I would definitely keep it in SC Quest, because it gives the ability to at least mark it as `drinking_water=no`.

If anything, one might also add `Other answer` of `It is disused` that would replace the matched `tag` with `disused:tag`; but that is probably too much work for too little gain (with any luck, `drinking_water=no` itself would equally well discourage people for going there searching for drinking water)

matkoniecz:
In related news, I am already making a systematic edit (not a bot edit) to handle such tagging on drinking water, see https://gitlab.com/matkoniecz/osm-automation/-/blob/master/script_assisted_cleanup/remove_trolltags_from_drinking_water.py

[...]

PR #4929, opened 2023-04-03
Handle disused/ruined springs

BenWiederhake:
Closes: #4928 

This prevents StreetComplete from asking whether the (usually non-existent) water in a disused/historic/ruined well is potable.

I can't easily test this, sorry. :(

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
@@ -20,6 +20,8 @@ class AddDrinkingWater : OsmFilterQuestType<DrinkingWater>() {
         and access !~ private|no and indoor != yes
         and !drinking_water and !drinking_water:legal and amenity != drinking_water
         and (!seasonal or seasonal = no)
+        and (!disused or disused = no)
+        and (!ruins or ruins = no)
     """
     override val changesetComment = "Specify whether water is drinkable"
     override val wikiLink = "Key:drinking_water"


matkoniecz:
`historic=* man_made=water_well` does not really mark inactive water well...

`ruins=yes` `disused=yes` are also quite nasty in this context and ideally would not be used but maybe skipping such items would make sense (but in such case that does not mean that this tagging scheme is a good idea)

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
```diff
@@ -20,6 +20,9 @@ class AddDrinkingWater : OsmFilterQuestType<DrinkingWater>() {
         and access !~ private|no and indoor != yes
         and !drinking_water and !drinking_water:legal and amenity != drinking_water
         and (!seasonal or seasonal = no)
+        and (!disused or disused = no)
+        and (!historic or historic = no)
```
matkoniecz, line 24: For start, thanks for a PR!

`historic=yes` does not mean that say water well is inactive - only that it is old/has historic value/etc. Can you remove it?

Skipping `disused=yes` makes sense, even if such tagging is a really bad idea.

Not sure about `ruins=yes` - there could be ruined working objects... And none of examples from #4928 have it. Not sure here.

matkoniecz (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
BenWiederhake: I agree that lifecycle prefixes would make far more sense, but that's not what I'm trying to change here. Instead, I'd like to suggest that SC should not ask a question if it doesn't apply. :)

I'll remove the `historic=yes` part.

I forgot to include ruins=yes in the overpass query, but there are matches, for example:
- https://www.openstreetmap.org/node/2624118790
- https://www.openstreetmap.org/way/967657623
- https://www.openstreetmap.org/node/2689111638 (only historic and ruins, but not "disused")

So I'm in favor of keeping the condition in there.

BenWiederhake: By the way, is there a reason why StreetComplete only considers this quest for water-well *nodes*, but not ways?

mnalis:
As noted in the issues itself, **I do not think this PR is the way to go**:

---

I agree that `disused=yes` + `man_made=water_well` (or others) is bad tagging. 
(It should've been `disused:man_made=water_well` etc.) 
Also as noted `historic=yes` does not mean that there is no drinking water there.

But I would definitely keep it in SC Quest, because it gives the ability to mark it as `drinking_water=no` if there is no drinking water!

If anything, one might add `Other answer` of `It is disused` that would replace the matched `tag` with `disused:tag`; but that is probably too much work for too little gain (with any luck, `drinking_water=no` itself would equally well discourage people for going there searching for drinking water)

app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
mnalis: > By the way, is there a reason why StreetComplete only considers this quest for water-well nodes, but not ways?

Probably not. I think it should be asked on ways too.

mnalis (changes requested):
While it would be better if the tagging was fixed by [automated edit](https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct) or something  (e.g. `man_made=water_well`+`ruins=yes` should be `ruins:man_made=water_well` etc.), it does not change the fact that it is better that SC keeps asking there - user on the ground can then confirm that the drinking water source is no longer available (by setting **very useful** `drinking_water=no` there!) or they could leave a note with picture explaining the situation in more detail. Or the situation might have changed and `disused=yes` water source have become fixed and is now `drinking_water=yes`!

So, I would not change existing behaviour there. (idea about asking on `ways` as well as `nodes` is a good one though, you might want to do PR for that)

mnalis:
> I can't easily test this, sorry. :(

As a side note, you probably can. 

In [your fork](https://github.com/BenWiederhake/StreetComplete/tree/dev-no-disused-wells), click on the `Actions` tab, then `Build debug APK`, then `Run workflow` and choose your branch you want to build. In some dozen minutes, you'll have `debug-apk.zip` with your `.apk` file ready to install on phone and test (it is debug build, so it can be run in parallel to upstream StreetComplete without ill effects)

matkoniecz:
On looking at it again, I think that makes sense to stop asking here.

At least in the current form it is simply confusing/pointless.

> user on the ground can then confirm that the drinking water source is no longer available (by setting very useful drinking_water=no there!) or they could leave a note with picture explaining the situation in more detail.

Not so useful in this case, and `drinking_water=no` is a bit confusing when there is no water at all.

> Or the situation might have changed and disused=yes water source have become fixed and is now drinking_water=yes!

That just makes tagging and situation even more confusing.

I will merge as is and propose systematic retagging this.

matkoniecz:
And other improvements (maybe adding support for asking on ways) are also welcome.

Maybe better handling it somehow for disused water wells? Though not sure what would be the best solution and how many objects benefit from this.

PR #4936, opened 2023-04-06
Also check ways when tag is not always a node

BenWiederhake:
Inspired by discussion in #4929, debug apk see https://github.com/BenWiederhake/StreetComplete/actions/runs/4623509091.

Barriers, bus stops, water wells, and recycling containers occupy significant amounts of space. A quick look at the wiki suggests that indeed, more than 1% of them are tagged as ways/areas, and not as nodes:

https://wiki.openstreetmap.org/wiki/Tag:barrier=yes
https://wiki.openstreetmap.org/wiki/Tag:public_transport=platform
https://wiki.openstreetmap.org/wiki/Tag:man_made=water_well
https://wiki.openstreetmap.org/wiki/Tag:recycling:glass=yes

Therefore, StreetComplete should also check ways, and not only nodes.

---

This means some "new" quests markers should pop up that have been skipped before, example:

Before this PR:

Screenshot_20230406-122736_censored

After this PR:

Screenshot_20230406-122720

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/AddBusStopName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/AddBusStopName.kt
@@ -11,7 +11,7 @@ import de.westnordost.streetcomplete.osm.applyTo
 class AddBusStopName : OsmFilterQuestType<BusStopNameAnswer>() {
 
     override val elementFilter = """
-        nodes with
+        nodes, ways with
         (
           public_transport = platform
           or (highway = bus_stop and public_transport != stop_position)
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierType.kt
```diff
@@ -14,7 +14,7 @@ import de.westnordost.streetcomplete.osm.Tags
 class AddBarrierType : OsmFilterQuestType<BarrierType>() {
 
     override val elementFilter = """
-        nodes with barrier = yes
+        nodes, ways with barrier = yes
```
matkoniecz, line 17: note that it is deliberate: `barrier=yes` line may be fence, wall, retaining wall... And retaining walls are directional, so supporting them would require supporting way reversing which is not implemented and tricky to implement.

matkoniecz (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_type/AddBarrierType.kt
BenWiederhake: Changed it.

Issue #4934, opened 2023-04-05
What form of traffic control is present at this intersection?


1ec5:
### General
Affected tag(s) to be modified/added: [`highway=traffic_signals`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_signals), [`highway=stop`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop), [`highway=give_way`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dgive_way), [`junction=uncontrolled`](https://wiki.openstreetmap.org/wiki/Tag:junction%3Duncontrolled)
Question asked: **What form of traffic control is present at this intersection?**
Or maybe more naturally: **How does traffic wait at this intersection?**

This quest should be limited to the United States and probably Canada. In these countries, there is no concept of a priority road and virtually every intersection has traffic lights, stop signs, yield signs, or the equivalent markings (“intersection control devices”). It should be limited to intersections where these tags are absent (and not on a nearby node representing a stopping point).

### Checklist
Checklist for quest suggestions (see [guidelines](https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md)):
- [x] 🚧 To be added tag is established and has a useful purpose
   - Three of these tags are de facto established and uncontroversial. `junction=uncontrolled` is rarer because it’s rarer in the real world.
   - Many renderers, including those used in turn-by-turn navigation, mark the locations of traffic lights, and some also mark the locations of stop signs and yield signs: mapbox/mapbox-navigation-ios#4185 ZeLonewolf/openstreetmap-americana#796.
   - Many routers apply a penalty to each traffic light or stop sign. Though small, this penalty adds up over successive blocks to contribute significantly to a short route. It also allows car routers to favor through streets more than stop-and-go neighborhood streets, where through traffic poses a safety hazard to cyclists and a nuisance to residents.
   - Ideally, a navigation application would use these tags to tell the user how many more blocks to go before the next turn, because many users don’t find distances very intuitive on surface streets: mapbox/mapbox-navigation-ios#426. “Blocks” are normally defined as controlled intersections, such as signalized intersections in urban areas. This is a major part of “natural-language guidance” in navigation systems, but OSM-based routers are currently unable to rely on these counts because of sparse coverage.
   - These tags will make it easier for mappers to classify pedestrian crossings as `crossing=traffic_signals` versus something else.
[...]

PR #4942, opened 2023-04-11
Create vertices experiment2

westnordost:
just using the github ui to review what this branch was about again and what's missing...


westnordost:
Hmm, one issue I see...:

1. split way 1 up into two ways: way 1 and way -1
2. add new node in the middle of way -1.
 
The `CreateNodeAction` now refers to a way with a temporary id. This means:

- on undo of the split way, all edits that are based on any elements created by this action should be deleted too. But the new node creation is not counted as dependent on the split way, hence it is not deleted (but it should!)

- when the split way is uploaded first, the new node creation still refers to way -1 and thus will automatically run into a conflict on upload because way -1 does not exist anymore by that time

westnordost:
By the way, the reason why I found it important that a node can be added to several ways also for the use case of StreetComplete is for cases where two areas are glued:

E.g. I want to add a door to a building. Now, this building could directly border the residential landuse, or a pedestrian area. I think it is common practice that the new door would become part of both ways (if only because it is kind of enforced by JOSM which likes to glue). Or, there is a fence in between two landuses and I want to add a gate to that - now the gate should probably be added to all three.
Things like that.

Though, on the other hand the question is whether
- people would actually find it bad if the door/gate node is only added to the building/wall and not the bordering area, whatever it may be
- whether for some cases, the node should **never** be added to the other way as well, i.e. it would be bad practice to do so. In that case, SC should probably not do it for any way to be on the safe side and then the question is if it is worth it to do this change

I understand that you, @Helium314 , implemented the adding-node-to-way by defining the way the node is added to as the "original element" of the node creation?

matkoniecz:
> whether for some cases, the node should never be added to the other way as well, i.e. it would be bad practice to do so.

adding gate directly under power line that crosses road?

Inverse case: adding gate to fence and road that were previously not sharing node (though in such case SC adding gate only to road would not make data worse)

matkoniecz:
Another case: adding entrance to multilevel mall building with shops mapped as areas inside. Is it necessary to avoid editing `level=-1` `level=1` shop areas when entrance at `level=0` is being added?

Helium314:
> You recently concerned yourself with this matter, do you understand where I saw the problem?

Looks like you answered that yourself already, I can't think of any further issues.

> I understand that you, @Helium314 , implemented the adding-node-to-way by defining the way the node is added to as the "original element" of the node creation?

Yes, that's how I did it.

> whether for some cases, the node should never be added to the other way as well

I guess there is no way to avoid maintaining lists of cases where node needs to be added to the other way, and where it should never be added...

westnordost:
> Looks like you answered that yourself already

Right, thanks for being my [rubber duck](https://en.wikipedia.org/wiki/Rubber_duck_debugging) :-)

westnordost:
Hm, I implemented it a bit more... to summarize the steps taken:

- remove `Element`, type and id from `ElementEdit` and the associated table `ElementEditsTable`
- instead, each `ElementEditAction` has as field for the `Element`(s) it needs to perform the action
- each `ElementEditAction` has a new method which must return the element key(s) it is based on. For e.g. the `UpdateElementTagsAction` it is just the element modified, for `SplitWayAction` it is the way being split, for `CreateNodeAction` it is the way(s) the node is added to, if any 
- there is a new table called `EditElementsTable` which is populated with all the element keys each edit is based upon and which is updated when these element ids change
- for this, each `ElementEditAction` has a new method in which it must return a new version of itself where the supplied `List<ElementIdUpdate>` has been applied
- when deleting an edit what is already done is that all edits that are based on an element that this edit created are deleted too (recursively). The `EditElementsTable` is used to find these edits.

**So, in a nutshell:**
`ElementEdit` does not anymore have exactly one element it is based on but instead has any number of elements it could be based on, depending on the `ElementEditAction`.

westnordost:
I am still contemplating whether the change is worth it, since the current use cases can well be covered with the current architecture.

**The good**
1. Previously, only the element id in the `ElementEditsTable` was updated but not the element itself because the latter lives in a serialized form, i.e. cannot be updated by a simple SQLite statement. It would be an error to use the `originalElement.id`, instead the `elementId` needed to be used as only that would be updated by the SQLite statement. This was a little ugly and is not the case anymore.
2. Previously, all edits *had to* be based on an element. Edits based on the `CreateNodeAction` however had none, which led to a small hack that had to be employed

**The bad**
1. Since the db structure somewhat changed, it is necessary to drop all local changes on database update because it would be too complicated to implement a migration

**The ugly**
1. Per point 1 in "the good": However, this makes it necessary for every `ElementEditAction` to implement a method that returns a copy of itself with the ids switched out. This is a lot of boilerplate that looks very similar in most actions and somewhat of a performance hit

matkoniecz:
> I am still contemplating whether the change is worth it, since the current use cases can well be covered with the current architecture.

I guess the question is what you think about potential future multiobjects edits? Like "is there a bridge over waterway or stream here". Or "do you confirm that such building exists here" - that would be useful for some datasources usable for import - where geometries are reliable but existence of buildings from the dataset is not and sometimes supposed buildings does not match reality.  Or things like #4934

If you see future for that then it makes sense to add it to enable it, if no and any such quests would be anyway rejected then benefits are far more dubious.

1ec5:
> By the way, the reason why I found it important that a node can be added to several ways also for the use case of StreetComplete is for cases where two areas are glued:

That makes sense. However, there’s still a class of potential quests that would add directional nodes, which are must be part of only one of the overlapping ways. StreetComplete’s inability to join the multiple ways using the new node could be a blessing in disguise, not only for the stop signs in #4934 but also for other kinds of directional features that the project might consider in the future.

For example, if the stop sign needs to be added as a separate node at the stop bar, and the roadway happens to be glued to a landuse area, the landuse area should remain untouched. Not only would the stop sign be meaningless in the context of the landuse area, but also the `direction` tag would very likely go in the opposite direction as the stop sign. Often it isn’t possible to reverse the landuse way’s direction, as that would turn around a different stop sign node. Even back when I was staunchly in favor of gluing landuse areas, I rarely glued them at stop signs because of the potential for misleading directions.

westnordost:
I don't know if you did know that, but if you undo an edit that creates element(s) while auto-sync is off (or when offline), all edits based on that first edit are currently deleted, too.
So if for example split a way at two points and then specify the surfaces for all three parts, but then decide that this was wrong and undo the split way (without first undoing the surface-edits), the surface edits will be found automatically and also be removed.

Now, in this PR, when one edit is able to edit multiple elements, it makes this functionality somewhat more complicated to realize. In fact, a database table (see `ElementEditsTable` and `ElementEditsDao`) must be added and its data be maintained to enable this functionality. On every upload of an edit that creates elements, edits that are based on any those elements must be fished out from the table, instantiated, a copy of that edit with the newly assigned element ids assigned created (see `ElementEditAction.idsUpdatesApplied`) and then persisted back to the database table. Performance-wise, this is probably not as bad as it sounds like, because usually few edits are affected by this, even if you have thousands of edits.

But complexity-wise I find it hard to justify (-> more code, more potential for bugs, much more test code that needs to be maintained). After all, nothing will break if SC does *not* delete also the edits based on an element the former edit created, the edit will just remain in the undo history and latest on upload, a conflict will be detected (due to trying to modify an element that doesn't exist) and subsequently dropped.

So, what do you think about removing that feature? I think @Helium314 has with by a long shot the best insight into the intricacies of the code here, but I'd also like to hear what you think of that from the user point of view.

1ec5:
In the first place, StreetComplete’s random-access undo is more sophisticated than any other software I can recall using – and I’ve used a lot of software. It’s far more common to treat undo as a stack. So I very much appreciate the care that you’re giving to this functionality.

From my perspective, if undoing something that has dependencies results in orphaned edits, it might not be surprising to see some undo actions that do nothing when saving, as long as they don’t cause any problems when you try to undo those actions themselves.

Your example with splitting sounds difficult to get right anyways, because the subsequent edits could have made the surface tags inconsistent among the formerly split ways.

Helium314:
> So, what do you think about removing that feature?

This basically means removing the [`getEditsBasedOnElementsCreatedByEdit`](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsController.kt#L141) step when deleting an edit?

I think this would be mostly fine and only affect very few users, as most will use auto-upload anyway. Of those who don't, probably most will never encounter this situation... I don't think I ever undid a split after editing any of the resulting ways. For those users where it happens, the discareded edits are probably just a minor annoyance, _if_ they notice them.
Btw editing the element with original id after a split, and then undoing the split results in an edit showing the wrong geometry (of the split, not the whole way), and I think no one ever complained.

Though, would it be possible to notice the (now) invalid edits e.g. as part of `MapDataWithEditsSource.applyEdit`?
This is triggered when deleting a edit (via `rebuildLocalChanges`), and should return null if the original element does not exist (due to the `ConflictException`). In this case the invalid edit could be removed.

Helium314:
> From my perspective, if undoing something that has dependencies results in orphaned edits, it might not be surprising to see some undo actions that do nothing when saving, as long as they don’t cause any problems when you try to undo those actions themselves.

I agree, though I guess it will still surprise enough users (if they notice it).

westnordost:
> Btw editing the element with original id after a split, and then undoing the split results in an edit showing the wrong geometry (of the split, not the whole way), and I think no one ever complained.
> [Edit: this only happens for some ways it seems, due to the isGeometrySubstantiallyDifferent thing. But the discarded edit remains in the history, and again no one ever complained. This might actually be fixed if the thing below works.]

Hm I see, the setting of the surface is not deleted because it is not based on an element created by the split. But at the same time, it should not be applied either because due to the `isGeometrySubstantiallyDifferent` check, a conflict is detected.

> Though, would it be possible to notice the (now) invalid edits e.g. as part of MapDataWithEditsSource.applyEdit?
This is triggered when deleting a edit (via rebuildLocalChanges), and should return null if the original element does not exist (due to the ConflictException). In this case the invalid edit could be removed.

So, remove any edits that fail to apply onto the local data. I considered that once, I don't remember anymore why I didn't do that. 
Can you think of a case where an edit would fail to apply to local data (throw conflict), but would not throw a conflict on upload? You know, applying a split locally or on upload is different because StreetComplete ignores any route relations locally.

Helium314:
> Can you think of a case where an edit would fail to apply to local data (throw conflict), but would not throw a conflict on upload?

Can't think of anything, but maybe the reason was different?
Could be performance though, because deleting the edit then triggers another call to `onDeletedEdits`. This calls `rebuildLocalChanges`, which was rather slow before introduction of the map data cache. And if there is another conflict when applying an edit during `rebuildLocalChanges`, we're back at `onDeletedEdits` once again...

westnordost:
Hm right, that will be the reason. `MapDataWithEditsSource` is not supposed to control, i.e. change anything (hence the name), just replay the edits on top of the OSM API map data.

westnordost:
And it would be a weird circular dependency if the `ElementEditsController` would listen on some kind of callback (`onTheFollowingEditsCouldNotBeAppliedDueToAConflict`) from the `MapDataWithEditsSource`.

Without a circular dependency, would probably have to be implemented like that there is something like a `VisibleElementEditsSource` that collects the information from both `ElementEditsSource` and `MapDataWithEditsSource`. Only edits enumerated by this class would be shown in the undo history, making it look like to the user that these edits were deleted (when actually they were only not shown because they could not be applied locally  ). But this sounds too complicated overall, so better just have orphaned edits in such a case...🤷

westnordost:
Oh no.... forget all that from yesterday.... I've made an error in thinking: Only because IntelliJ didn't report any other uses than the undo function, it doesn't mean that updating the ids after upload is not absolutely necessary. Otherwise, a simple thing as this would fail on upload:

1. create node -1
2. answer opening hours for node -1

I.e. edit 2 must update the id of the node it holds when edit 1 has been uploaded but for the controller to find which edits need updating, the `EditElementsTable` is necessary. The actual "delete edits based on this deleted edit" undo function just adds something like 30 lines of code to that.

westnordost:
Work continues in different branch

PR #4945, opened 2023-04-13
Make note edit text (and some more) selectable

### LINKED ISSUE #4924: Undo dialog: Make text selectable
**Use case**
After creating a note, I often realise there's more to add - it would be nifty if I could edit the note rather than undoing and retyping it to make my edit.

<strike>**Proposed Solution**
Perhaps not-yet-uploaded notes could be shown over the map in a different colour so they can be tapped again to edit?</strike>

**EDIT by @westnordost**:
Accepted solution is described here: https://github.com/streetcomplete/StreetComplete/issues/4924#issuecomment-1497361215

Helium314:
fixes #4924 

Makes text created by `UndoDialog.createTextView` selectable. This is used for all edits except for the actions `UpdateElementTagsAction` and `CreateNodeAction`.

Only question I have is about consistency: should all actions lead to selectable text? Or only notes? Or is current PR fine?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/edithistory/UndoDialog.kt
@@ -155,6 +155,7 @@ class UndoDialog(
         val txt = TextView(context)
         txt.layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, WRAP_CONTENT)
         txt.setText(text)
+        txt.setTextIsSelectable(true)
         return txt
     }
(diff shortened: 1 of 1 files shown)


westnordost:
Cool!

I think it would make sense to have it selectable regardless of the kind of edit.

westnordost:
Incidentally, did you test it?

Helium314:
I did test (using SCEE), and noticed that making text selectable breaks the wiki links for edited tags. So I think it would be better to have only have `createTextView` selectable.

westnordost:
I see, that's fine then. Thank you!

PR #4946, opened 2023-04-13
grant asphalt (etc) a different color from paving stones (etc)

westnordost:
I think it is more important to be able to differentiate asphalt etc. from paving stones etc. than gravel etc. from metal etc.

Not because the latter would be more similar to each other, but because
- the former surfaces are used much much more often
- the latter surfaces usually do not appear close to each other / are used in the same area / place

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/surface/SurfaceColorMapping.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/surface/SurfaceColorMapping.kt
@@ -38,24 +38,27 @@ import de.westnordost.streetcomplete.overlays.Color
  *   by extremely strong association between surface and colour
  */
 val Surface.color get() = when (this) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/overlays/surface/SurfaceColorMapping.kt
```diff
@@ -38,24 +38,27 @@ import de.westnordost.streetcomplete.overlays.Color
...
     COMPACTED, FINE_GRAVEL
                        -> Color.TEAL
     SAND               -> Color.ORANGE
     GRASS              -> Color.LIME
     DIRT, SOIL, EARTH, MUD, GROUND, WOODCHIPS
                        -> Color.GOLD
-    GRAVEL, PEBBLES, ROCK
+    GRAVEL, PEBBLES, ROCK,
+    // very different from above but unlikely to be used in same places, i.e. below are usually on bridges
+    WOOD, METAL, METAL_GRID
```
matkoniecz, line 57: Occasionally happens on hiking routes and similar cases, but overall it is an improvement.

westnordost:
image

PR #4947, opened 2023-04-14
Use older locations to check whether user has been on site

### LINKED ISSUE #4727: Less zealous "checked on-site" dialogue
When using Streetcomplete from a passenger seat I often find I am prompted as to whether I have really checked something on-site despite having just seen it.

**How to Reproduce**

Use Streetcomplete to answer questions from a moving vehicle, only selecting the quest after you pass it. Often despite having just seen a feature a prompt will open to interrogate the user about whether they checked in person.

Typical examples: 

- user has just passed a street that was lit, but is shown as unknown in the overlay
- user accidentally tapped the wrong quest and had to cancel and reselect the correct quest prompting the "you're too far" sreen

**Expected Behavior**

Now that a trace is shown for recent history I would expect all features within range of the last ~2-5s of travel to be within the boundary for "user has probably just seen this" and not trigger a prompt.

**Versions affected**

Versions to 50.1.

Helium314:
fixes #4727

Keeps a list of recent locations (within a minute of most recent location update), and uses them for `checkIsSurvey`.
The most recent one is checked first, so I would not expect any noticeable performance impact for most users.
If a user has been on site recently, a slower check is still considerably faster than clicking ok in the dialog.

Some performance optimizations are possible, but likely not worth the effort. It only could make things worse for users who haven't been on site recently, but even then: the check isn't _that_ slow.

This is _not yet tested_ for the actual intended use case, will do if the PR is acceptable otherwise.

`displayedLocation` is now unused and could be removed from the listeners.


matkoniecz:
marked as draft per "This is not yet tested"

matkoniecz:
In general it looks like a nice idea. Maybe even more than a minute would be nice? But that can be changed later.

-------------------------------------------------------------------

For performance: How many positions are there typically?  1000? Millions? AKA, how often onLocationChanged is called?

I think that bigger danger is constant maintaining such list of recent locations. But it is kept for GPS trail anyway.

That makes me ask: why separate list of locations is maintained here? Would it be possible to reuse what is kept to draw GPS trail? Or would it be more complicated.

Helium314:
> How many positions are there typically?

My device produces at best one GPS location per second, so it would not be more than 60. But maybe more modern devices can do faster updates. Easiest would be ignoring locations close to the previous one as part of `isWithinSurveyDistance`.

> Would it be possible to reuse what is kept to draw GPS trail?

No, because the trail requires accuracy of 20m.

matkoniecz:
> so it would not be more than 60

Then worrying about performance impact of checking all of them seems not needed

westnordost:
Hmm, a few comments regarding this:

1. this list of recent locations doesn't belong into the the UI code (into the fragments). When these fragments are recreated, boom, it's gone. Saving and restoring the state would only workaround this issue. The next alternative is not to persist the data in the database, I think that would be over the top, but to have a "static" class (i.e., a _singleton_, we use dependency injection after all) that just keeps a list of recent locations. This very class could also include the logic described in point 2

2. As noted earlier, sixty seconds is a bit little. It will already be difficult to exit the `isWithinSurveyDistance` within 60 seconds. The class I wrote about in point 1 could also contain the logic that disregards points fed to it that are too close to itself. (Now, often surveyors go in circles, or a bit back and forth etc., it would be ideal if these could be filtered out as well, but I have no idea how this could be done well, `LatLonRaster` only works on pre-set bounds)

(3. `Location` is an android specific data structure but the whole `isWithinSurveyDistance` is application logic that _should_ be independent of Android. But I realize that the whole implementation here is currently very much Android dependent, so nevermind. Maybe a task for the future.)

Helium314:
Hmm I see, so there is quite a bit more work necessary than I had expected...
Currently I'm not willing to spend the necessary time on this, so I'll continue a little later, maybe in a few days or maybe weeks or so.

(if anyone is interested, feel free to take over)

Helium314:
>  The class I wrote about in point 1 could also contain the logic that disregards points fed to it that are too close to itself. (Now, often surveyors go in circles, or a bit back and forth etc., it would be ideal if these could be filtered out as well, but I have no idea how this could be done well, `LatLonRaster` only works on pre-set bounds)

Current ideas (assuming we want at least 10 m between locations)
1. when adding a location, remove the previously added location if it's within 10 m of the current one: this is not good if all locations are within 10 m of each other, as then _only_ the most recent one is stored
2. when fetching recent locations, provide a sequence where a location is added only if it's at least 10 m away from the previous entry: this is better, but will not detect going forth and back
3. when fetching recent locations, provide a sequence where a location is added only if it's at least 10 m away from all previous entries: best result, but will be slow if there are many locations. May significantly profit from using the approimation https://en.wikipedia.org/wiki/Geographical_distance#Spherical_Earth_projected_to_a_plane, which I think is acceptable here
4. something like 3., but remove discarded locations from the recent locations list for performance reasons: this might have unintended side effects similar to 1... maybe?

westnordost:
> 3

Uhh, rather not. Or if at all, then everywhere. And that doesn't change that adding a new latlon would still be O(n) time complexity. Actually,  what would work to make time complexity I think O(log n) is to use a quadtree to sort the coordinates (i.e. go one level deeper if more than X points are in the same leaf). But,  I don't think it is worth the time it to implement this.

I think the most straightforward solution would be to not add a new latlon to the list of recent locations if it is closer than max-survey-distance/2. Going forth and back - so be it.

Helium314:
Some performance test with the current version:
Stores locations for 10 minutes, and returns locations that are at least 40 m and 5 sec apart from each other.
Usually only the first location is checked, which is very fast anyway.
With more than 100 locations returned, and checking against a complicated multipolygon building that is too far away (so really all locations are checked), the worst case was ~100 ms.

Helium314:
The maximum number of locations for the survey distance check is currently 120 (locations of the last 600 seconds, and at least 5 sec between locations).

endolith:
This confirmation dialog is the most annoying part of StreetComplete by far; I wish I could turn it off permanently. 

A simpler solution for most of the false positive warnings would be to trigger the warning based on the location I *clicked* the quest instead of the location I have moved to when I *completed* the quest.

mnalis:
@endolith (and other people of course) if you wish to try out how well does this PR solves the issue (to give feedback), you can try this debug APK https://github.com/mnalis/StreetComplete/actions/runs/5101411056 (my fork also changes a few quests slightly to be shown in little more places, but should not affect the experience regarding this issue).

Atrate:
Isn't "a list of recent locations" already kind of a thing with the GPS trace that's recorded? https://github.com/streetcomplete/StreetComplete/issues/2209

riQQ:
Any reason this is still a draft? All remarks / request for change seem to be resolved.

Helium314:
> Isn't "a list of recent locations" already kind of a thing with the GPS trace that's recorded? #2209

There was some reason this wasn't usable, maybe  because this only records high accuracy locations, iirc 15 m and better. On a train I usually have accuracy much worse than that.

> Any reason this is still a draft? All remarks / requests for change seem to be resolved.

"This is _not yet tested_ for the actual intended use case" still applies (I would only ever need that PR on a car or train ride, and still didn't test it there), and that was the reason to convert it to a draft.

Helium314:
Though @endolith maybe you tested @mnalis' build?

Going through the changes again, I guess there might be issues if you travel faster than ~100 km/h.

endolith:
> Though @endolith maybe you tested @mnalis' build?

Yeah it seems to be working well.

Helium314:
@matkoniecz do you consider this tested enough, so this PR can leave draft status?
@westnordost anything that needs to be changed?

app/src/main/java/de/westnordost/streetcomplete/screens/main/RecentLocationStore.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlin.math.abs
+import kotlin.math.cos
+
+class RecentLocationStore {
+    private val recentLocations: MutableList<Location> = mutableListOf()
+
+    /** returns a sequence of recent locations, with some minimum time and distance from each other */
+    fun get(): Sequence<Location> = synchronized(recentLocations) {
+        var previousLocation: Location? = null
+        recentLocations.asReversed().asSequence().filter {
```
westnordost, line 15: What is synchronized is the creation of the sequence, but the sequence is based on a non-copied version of `recentLocations` which means that `recentLocations` can be iterated through outside of the synchronized block. So, the synchronized block does not actually protect against race conditions. In fact, I believe this can lead to concurrent modification exceptions when the list is iterated while new locations are added.

I understand that you do not want to iterate the whole list and calculate `isTooCloseTo` on it, but this approach won't work for the above reason.
Couldn't you do this check in `add` instead?  (Should be faster, too)

app/src/main/java/de/westnordost/streetcomplete/screens/main/RecentLocationStore.kt
```diff
@@ -0,0 +1,49 @@
...
+
+// ~ 5 times faster than using SphericalEarthMath distanceTo, could be 20 times when using approximate cos
+private fun Location.isTooCloseTo(location: Location): Boolean {
+    // see https://en.wikipedia.org/wiki/Geographical_distance#Spherical_Earth_projected_to_a_plane
+    val dLat = (latitude - location.latitude) * PI / 180.0
+    val dLon = (longitude - location.longitude) * PI / 180.0
+    val cosDLon = cos((latitude + location.latitude) * PI / 180.0 / 2) * dLon // actually cos could be approximated using 1 - x^2 / 2 + x^4 / 24 - x^6 / 720: 4 times faster and good enough
+    val distanceSquared = EARTH_RADIUS * EARTH_RADIUS * (dLat * dLat + cosDLon * cosDLon) // no need for sqrt
+    return distanceSquared < MAX_DISTANCE_TO_ELEMENT_FOR_SURVEY * MAX_DISTANCE_TO_ELEMENT_FOR_SURVEY / 4
+}
```
westnordost, line 46: Hmm:

1. Are you sure that this isn't premature optimization, i.e. actually a performance bottleneck? Also, I'd prefer if "FlatEarthMath" (if it is really necessary) wasn't scattered in the code in some private functions but be in a `FlatEarthMath.kt` and subsequently also tested properly (e.g. 180th longitude)

2. (There is actually [`Location.distanceTo`](https://developer.android.com/reference/android/location/Location#distanceTo(android.location.Location)) but it likely does the same as the SphericalEarthMath function.)

westnordost (changes requested):
I am sorry, it seems I missed that you requested my review.

`RecentLocationStore` should probably rather go into `data/` as it has nothing to do directly with the UI (and so does part of `CheckIsSurvey` but this is not up to you to clean that up 🤷 ).

Helium314:
I only can remember that I did some performance tests, but not what exactly.
Will go through it again, once I have more time (in a week or two).

app/src/main/java/de/westnordost/streetcomplete/screens/main/RecentLocationStore.kt
Helium314: The sequence should always start with the most recent location, which is not guaranteed when checking in `add`. In worst case the user could be close to the location, but still get the popup.

westnordost: That doesn't seem to be a big problem. Just increase the maximum survey distance constant by a bit.

app/src/main/java/de/westnordost/streetcomplete/screens/main/RecentLocationStore.kt
Helium314: I only did some tests in Android Studio, here the overall check whether a simple 3 point way is within survey distance is 10% faster than with `LatLon.distanceTo`.
So probably not worth the optimization.

Actually a faster (or approximate) `SphericalEarthMath.angularDistanceToArc` would be much more helpful.

Helium314: I added `FaltEarthMath.kt`, and some tests whether the results are reasonably close to the `SphericalEarthMath` equivalents.
It's tested for 0.03° (up to ~3 km, depending on latitude) and here the results are within 0.5 m of the spherical functions.

Maybe there could be some additional functions that use the approximate distance calculations if the points are close, and the spherical ones otherwise.

Helium314:
> RecentLocationStore should probably rather go into data/...

Just in `data/`, or some subfolder?

matkoniecz:
>do you consider this tested enough, so this PR can leave draft status?

Yes

westnordost:
Hm, did you not read my [comment](`https://github.com/streetcomplete/StreetComplete/pull/4947#discussion_r1290286383)?

With the current implementation, not only is the whole list of recorded position copied each time a change is made (even for positions that are basically on top of each other), but these calculations need to be done on each making an edit.

It should have its own folder in `data/`.

Also, I noted another possible performance issue. Once the first locations are old enough for 

```kotlin
        recentLocations.removeAll {
            it.elapsedRealtimeNanos <= location.elapsedRealtimeNanos - LOCATION_STORE_TIME_NANOS
        }
````

to do anything, it will be done on every adding of a new location (each second?). I.e. each second, the whole array is being re-created (because removing element(s) not at the end of the array does that).

Re-allocation of a new array is probably cheap, but it could be avoided entirely by changing from a time-based cut-off (10 minutes) to a number-of-locations-based cut-off and using a ring buffer as the underlying data structure instead of an arraylist.

**Comment by Helium314** [comment]
> Hm, did you not read my comment?

Which one? The previous commit was done to address:

> Are you sure that this isn't premature optimization, i.e. actually a performance bottleneck? Also, I'd prefer if "FlatEarthMath" (if it is really necessary) wasn't scattered in the code in some private functions but be in a FlatEarthMath.kt and subsequently also tested properly (e.g. 180th longitude)

Now the whole `checkIsSurvey`, which is suspending because distanceToArcs is slow, should be about 10 times faster. So the worst check time I found (100 ms, and this is already acceptable) should be reduced to 10 ms (did not test on the phone). Or maybe let's say 20 ms if copying the list and other things are really slow.
So if this is the comment you're referring to, I don't understand necessity of further optimization of `get`. You even questioned the attempt of reducing the 100 ms by ~10% as premature optimization.

Or do you mean the testing part? Isn't it enough that for the planned use case the difference to the spherical counterpart is always less than 1 m?

Or do you mean 

> Couldn't you do this check in `add` instead? (Should be faster, too)

I could do this, but then again this strikes me as premature optimization, as you already said the measured 100 ms are acceptable (done without copying the list using `reversed`, but the time spent copying/reversing shouldn't nearly come close to 10 ms).

> With the current implementation, not only is the whole list of recorded position copied each time a change is made (even for positions that are basically on top of each other), but these calculations need to be done on each edit. And this on the UI thread IIRC, or at least, blocking further interaction on the UI thread, which should come down to the same.
With time the complexity being O(n²), I'd want to keep the n as small as possible.

This again is something I don't really understand. Performance even for complex geometries and on old phones is acceptable, but now should be optimized more?
Also I'm not sure which part you're referring to with O(n²)

> Also, I noted another possible performance issue. Once the first locations are old enough for [...]

While I agree that things could be optimized, again there is the question why, as the performance was acceptable already before optimizing the by far slowest part.
Though I changed the `removeAll` to only check the oldest location(s), as this is 20 times faster according to tests, and at the same time `add` is called much more often than `get`.


So in summary, I'm quite confused by the reason for demanding further optimizations. Do you plan to make changes like increasing `LOCATION_STORE_TIME_NANOS`, or other things where those optimizations actually become relevant?

Anyway, all this feels like it's going to to eat more time, and like it's not going to end anytime soon... maybe @mnalis or someone else who is actually interested in this feature wants to take over?

**Comment by westnordost** [comment]
Sorry, it looks like you poured much more thought and time into it than I. I wasn't aware that you already tested the different approaches in terms of time. Also, it seems it came across as demanding you change it. That was not my intention, I was just not sure whether you saw the comment I linked to because I answered after you collapsed the thread as resolved.

With 100 positions you came out with the 100ms, so for a session that runs 10 minutes or more, it will be ~120ms. This would be a noticeable delay for the user but I understand the move to the flat earth math made something like 10 times faster, so this is completely fine, then.

> Also I'm not sure which part you're referring to with O(n²)

More like O(n*m), Checking if a way of n segments is within survey distance of last m positions.

Will review now.

File: app/src/main/java/de/westnordost/streetcomplete/util/math/FlatEarthMath.kt [hunk 3]
```diff
@@ -0,0 +1,86 @@
...
+import de.westnordost.streetcomplete.util.ktx.asSequenceOfPairs
+import kotlin.math.PI
+import kotlin.math.abs
+import kotlin.math.sqrt
+
+/** Calculate stuff assuming an (almost) flat Earth. The main exception from assuming a completely
+ *  flat earth is use of cos(lat) to take into account decreasing longitude distance at
+ *  high latitudes.
+ *  Optimized for performance with precision within 1 m of spherical functions for up to
+ *  0.03° difference between points (several km at common latitudes). */
```
westnordost, line 15: Does the math take into account the wrapping at 180th longitude? It should probably be mentioned in the doc comment.

(Also it would be nice if you added the doc comments from their sphericalearthmath siblings to the public functions. Not a merge-blocker.)

app/src/main/java/de/westnordost/streetcomplete/data/location/RecentLocationStore.kt
```diff
@@ -0,0 +1,40 @@
+package de.westnordost.streetcomplete.data.location
+
+import android.location.Location
+import de.westnordost.streetcomplete.util.ktx.toLatLon
+import de.westnordost.streetcomplete.util.math.flatDistanceTo
+import kotlin.math.abs
+
+class RecentLocationStore {
+    private val recentLocations: MutableList<Location> = mutableListOf()
```
westnordost, line 9: Could you change the type to [`ArrayDeque`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-array-deque/)?
```suggestion
    private val recentLocations = ArrayDeque<Location>()
```

It really puzzles me that in `add()`, `recentLocations.removeFirst()` in a loop  is faster than `removeAll` with a filter lambda. Anyway, `ArrayDeque` is optimized for adding and removing elements at the front and at the end.

Helium314:
> More like O(n*m), Checking if a way of n segments is within survey distance of last m positions.

Thanks for clarifying, I had the impression you meant n² with n positions. Large number of segments definitely is what has most potential to slow down things, but it should be much better now.

> With 100 positions you came out with the 100ms

Actually this is (of course) on my S4 Mini, and happens only for complex geometries (many arcs for `distanceToArcs`) and if all checks result in "too far away". I don't have the logs any more, but single digit ms was the common case.
I'll do another "real life" test soon, including logging details like number of segments.

> Also, it seems it came across as demanding you change it

That was my feeling, thank you for clarifying.

Assuming the test doesn't reveal any issues, the current approach should be fine for the planned 10 minute storage. But in case you want it extended for longer times / only limit to a number of locations, your suggestion is definitely preferable.

app/src/main/java/de/westnordost/streetcomplete/util/math/FlatEarthMath.kt
Helium314: > Does the math take into account the wrapping at 180th longitude

Not yet, will update and add documentation.

Helium314: > Does the math take into account the wrapping at 180th longitude

Things got a bit too complicated for for `flatDistanceToArc`, so it will fall back to the spherical version in that case.

app/src/main/java/de/westnordost/streetcomplete/data/location/RecentLocationStore.kt
Helium314: > It really puzzles me that in add(), recentLocations.removeFirst() in a loop is faster than removeAll with a filter lambda

There was a huge but reproducible difference in a test I did, I also hadn't expected this.

Helium314:
The survey distance check for 100 locations and a building with 70 segments takes around 35-50 ms. Not as much improvement as I had expected, but I don't know which building I had tested on previously. And maybe the debug version is slower than the release version.

app/src/main/java/de/westnordost/streetcomplete/util/math/FlatEarthMath.kt
```diff
@@ -45,15 +52,21 @@ fun LatLon.flatDistanceToArcs(polyLine: List<LatLon>, globeRadius: Double = EART
...
+ *  Takes into account wrapping at 180th longitude */
 private fun flatAngularDistance(φ1: Double, λ1: Double, φ2: Double, λ2: Double): Double {
     // https://en.wikipedia.org/wiki/Geographical_distance#Spherical_Earth_projected_to_a_plane
     val δφ = φ1 - φ2
-    val δλ = λ1 - λ2
-    val cosδλ = approxCos((φ1 + φ2) / 2) * δλ
+    var δλ = abs(λ1 - λ2)
+    if (δλ > PI)
+        δλ = 2 * PI - δλ
+    val cosδλ = approximateCos((φ1 + φ2) / 2) * δλ
```
westnordost, line 63: This won't work if λ1 is e.g. than 5 * PI or -5 * PI. The longitude of a `LatLon` is not necessarily normalized. If you want to guarantee the value to be in a certain co-domain (e.g. -PI .... PI), have a look at `normalizeLongitude` (but that one works with degrees, not radians).

Also, the `abs(...)` looks suspect, because you truncate half of the co-domain of the result. Or was this a bug before that you corrected? (Better make sure all is covered in tests.)

Helium314: >  The longitude of a LatLon is not necessarily normalized

I thought it must be:
https://github.com/streetcomplete/StreetComplete/blob/14b80d7c5e1335d859c219e1735a238f854215e2/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/Element.kt#L85-L89

> Also, the abs(...) looks suspect [...]

Since only the square of `δλ` is actually used, we don't need to care about the sign.
I added the `abs` instead of `else if (δλ < -PI) δλ = -2 * PI + δλ`

> Better make sure all is covered in tests

What kind of further tests do you have in mind? Other than delivering very similar results to the spherical functions if distance is small...

westnordost: Oh, you are right. 

> What kind of further tests do you have in mind?

I meant if the `abs` thing was a bugfix. But I just didn't understand it, so that's fine.

westnordost (approved):

Issue #4686, opened 2022-12-15
Button for changing overlays on mainscreen 


Katzenstreu:
Hey,

I'm using the overlay quests a lot. But I can't recognise fast which kind of overlay is active without switching in the overlay drop down list.

First 
I wish an symbol, an indicator, which overlay is active.

Second
I suggest an easy method to change some own predefind overlay quests. I think most people are using some overlay quest types rarely far. Maybe we could activate some own defined overlay quests via checkbox. If we are outside it would be nice to change the most common own defined overlays fast with one click on the main screen. I've painted a mockup.

![Screenshot_20221215_081633_StreetComplete](https://user-images.githubusercontent.com/1107883/207801982-424a171d-41c2-4dcf-b088-f1c1bd632a23.jpg)

A long click would start the known dialog to activate and change the kind of overlay. But a short click would activate the next own predifined overlay.

In my case I would activate Straßenbeleuchtung, Bürgersteige, Parkstreifen and Fahrradwege because this are kind of quests. 

Best regards
Tim

westnordost:
A quick-switcher for overlays is currently not planned. (Sorry, I am very opinionated when it comes down to the number of buttons (permanently) visible on the main map view.)

mnalis:
(related rejected request for quick-switch of presets in https://github.com/streetcomplete/StreetComplete/issues/3827#issuecomment-1058071697.) 

Note that quick switch of overlays is much more important that of presets, as it happens all the time for many mappers. I.e. as a bicycle-mapping-oriented person, I would like to use `sidewalks overlay`, `bike paths overlay`, `surface overlay` and `street lighting overlay`. Unfortunately, I can only choose one, as switching between them by 4-clicks-action is way too annoying and time consuming (and I'd be switching them at least every 100 meters or so!).

I'd love 1-click "mode switch" as it exists in EveryDoor (its even 0-click sometimes - e.g. I can add amenity while in micromapping mode too!) . Even 2-clicks might be workable solution. But 4-clicks is way too much, so I end up using overlays very rarely. :cry: 

Another alternative would be to allow **some screen gesture**  for switching. That would avoid evil always-visible button (but that is likely more problematic to implement, and less discoverable)

PR #4949, opened 2023-04-14
Remove confirmation step from overlay selection

tapetis:
This PR changes the `OverlaySelectionDialog` to require one less tap to change overlays by replacing the "OK" button with a "cancel" button and directly switching the active overlay when a row is tapped. The updated dialog looks like this:

image

As a button on the main screen to change overlays is currently not planned (see #4686), I think this change is still a step in the right direction to allow faster switching between overlays.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/overlays/OverlaySelectionAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/overlays/OverlaySelectionAdapter.kt
@@ -22,6 +22,8 @@ class OverlaySelectionAdapter : RecyclerView.Adapter<OverlaySelectionAdapter.Vie
 
     var selectedOverlay: Overlay? = null
         set(value) {
+            // To match other preference dialogs, also invoke callback when nothing changed to allow
+            // dismissing the selection dialog when the active overlay is selected again.
             onSelectedOverlay?.invoke(value)
             if (field == value) return
             field = value
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/overlays/OverlaySelectionAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/overlays/OverlaySelectionAdapter.kt
@@ -22,9 +22,9 @@ class OverlaySelectionAdapter : RecyclerView.Adapter<OverlaySelectionAdapter.Vie
 
     var selectedOverlay: Overlay? = null
         set(value) {
+            onSelectedOverlay?.invoke(value)
             if (field == value) return
             field = value
-            onSelectedOverlay?.invoke(value)
             notifyDataSetChanged()
         }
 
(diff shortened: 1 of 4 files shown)


westnordost:
Hmm, this UI is inconsistent with expectation. You wouldn't expect a radio button select group to immediately return. 

Better would be if it looked like a context menu / selection dialog. E.g. like a selection dialog as used for selecting the surface in the surface overlay: No OK or Cancel button, no radio button graphics and the whole cell has the ripple animation on press down.

I don't know how to make a custom context menu that is anchored at any view on screen, but at least having a selection dialog as described should be possible.

westnordost:
Hmm, ypu ate right,  on both points... 

Well,  then it's fine like this.

However,  I think the whole row should then update the ripple, because the whole row acts as a button

westnordost:
If the radio button in the checked text view is a state drawable, you could just use that drawable in a linear layout that shows all of the radio button, the text and the icon and make that layout clickable.

app/src/main/java/de/westnordost/streetcomplete/screens/main/overlays/OverlaySelectionAdapter.kt
```diff
@@ -22,9 +22,9 @@ class OverlaySelectionAdapter : RecyclerView.Adapter<OverlaySelectionAdapter.Vie
 
     var selectedOverlay: Overlay? = null
         set(value) {
+            onSelectedOverlay?.invoke(value)
             if (field == value) return
             field = value
-            onSelectedOverlay?.invoke(value)
```
westnordost, line 27: huh? Why invoke anything if nothing is changed?

westnordost (commented):
Cool, so you found a way! One comment though:

app/src/main/java/de/westnordost/streetcomplete/screens/main/overlays/OverlaySelectionAdapter.kt
westnordost: I see, can you add that explanation as a comment?

westnordost:
Thank you!

westnordost:
By the way, do you happen to know how to make a dialog like this behave like a context menu?

westnordost:
Yeah, that one.  I.e. a dialog that is anchored in a certain view and is animated from that view

westnordost:
Thanks! 

I was asking because people keep requesting to have the overlay switcher as a button in the main view. Since the overlays make up an ever-increasing part of StreetComplete and thus quite a lot of functionality is "hidden" in an entry in the menu, maybe such a button could be added after all. But then, it would look more naturally if not a full-screen dialog would pop up to select the overlay but a view that is anchored in that button.

Helium314:
I had played with such a switcher as part of SCEE quick settings, but encountered issues with icons taking large amount of horizontal space for whatever reason.
Though that was in a sub-menu of the popup menu, maybe that's a relevant difference.

Issue #4303, opened 2022-08-20
recheck recycled materials-quest is non-intuitive to mappers when several containers are mapped close to each other


Lukas458:
See https://www.openstreetmap.org/node/7609573820 , touched in https://www.openstreetmap.org/changeset/123716033

There exist two recycling containers in osm, nearly at the same place. 
The first one all the recycled materials which are open to be recycled there "24/7" and the second one only contains the glass bottles, because of the opening hours. Because the two nodes are nerly at the same place, the user will be unable to distiinguish which one is now tagged as the glass_bottles-container and whch one is for the other materials.

**How to Reproduce**
You need at least two mapped recycling containers which are very close to each other. This constellatin exists often in Germany, for example, because containers where glass bottles are recycled usually have opening times, which are maped with opening_hours. The opening_hours are valid only for the glass bottles container and nor for other containers which recycle other materials, so the glass bottles container is mapped separately.

**Expected Behavior**
A container which has recycling:*=... and opening_hours=* shuld not get any recycling:...=yes tags added for materials, for which the opening_hours do not apply. 

In principle, however, this has nothing to do with opening_hours, but generally with separately mapped containers. With separately mapped containers, it is always the case that at least one of the containers that exist in osm only recycles exactly **one** specific material out of a total of several materials which are accepted at this location.

Sometimes containers are also mapped separately because depending on the material accepted, the operator differs.

The user needs an information about which container he's aked about. Maybe the information, that THIS container which he's asked about, was mapped as the glass-bottles container until now, or which materials are tagged there until now? Maybe we can find a solution together for this. 
**I think simply showing two icons is not enough. Most mappers understand this in such a way that a container represents an entire location and that he/she is asked for the recycled materials of the ENTIRE location.**

**Versions affected**
version 45.0

westnordost:
>  I think simply showing two icons is not enough. Most mappers understand this in such a way that a container represents an entire location and that he/she is asked for the recycled materials of the ENTIRE location.

Why not, though? Assuming that the recycling containers are placed correctly, you should be able to see which is which simply by looking at the icons on the map, i.e. how they are oriented towards each other.

Helium314:
Another example of this issue, very close separately mapped glass and clothes containers.
The original glass container https://www.openstreetmap.org/node/5413118293 was changed by one user for both materials, apparenly unaware of the second container.
[...]

Issue #3084, opened 2021-07-20
Show FIXME tags and allow for additional information


mnalis:
**Use case**

If a node/way contains [fixme=*](https://wiki.openstreetmap.org/wiki/Key:fixme) tag, it denotes that the item needs fixing; often (or always if string contains `#surveyme`, as in notes) because it needs surveying. SC currently handles Notes, but not the `fixme` tag which fulfills similar purpose. It would be nice if it supported that too.

It could also help with https://github.com/streetcomplete/StreetComplete/discussions/3083 or https://github.com/streetcomplete/StreetComplete/discussions/2477#discussioncomment-439407 and similar

**Proposed Solution**

If the node with `fixme` tag exists and OSM note does not exist at the same place; it should be shown in the same way as the existing OSM Note, eg. with with question `Can you contribute anything to this fixme?` and then showing the content of `fixme` tag. It should also use filtering that Notes use (eg. adapt to `Show all notes` preference, and to existence of `#surveyme` in tag) 

- if the user selects `No, hide`, it should be hidden in the same way as the note is (entry in local ignore-DB I guess?)
- if the user types text / adds a picture and clicks `OK`, new note should be created at that location with a context (ex. if there was `fixme=has this supermarket reopened?` and user answers `yes, opening hours are now xxxx` the new OSM Note created would say `user responded to "fixme=has this supermarket reopened?" with: yes, opening hours are now xxxx`). Creation of the note would turn it into regular note quest (and as explained at the top, `fixme` would be ignored on locations which have **both** `fixme` tag and open OSM Note)
- (optionally, we could add additional button `this has already been fixed`, which would remove `fixme` tag, but it is probably safer and simpler to let the user leave note and let somebody else updates it if it makes sense).

(Also note previous/closed related issue https://github.com/streetcomplete/StreetComplete/issues/1775; this provides new idea that addresses shortcomings of previous suggestion)

matkoniecz:
Note that many fixme values are cryptic or do not require a survey. See https://taginfo.openstreetmap.org/keys/fixme#values

I display `fixme` in my fork and it is typically usable only to trigger switch to Vespucci.

For example `fixme=continue?` requires context of where such node is placed.

mnalis:
> Note that many fixme values are cryptic or do not require a survey. See https://taginfo.openstreetmap.org/keys/fixme#values

[...]

PR #4956, opened 2023-04-19
Skip recycling quest on elements with "note" tag, as those are likely to be problematic 

mnalis:
Recycling quest should be skipped on elements having `note=* `tag (as its existence likely indicates the issue is too complicated for AI, and human is required to evaluate specific issues detailed in freeform text of `note`)

(see https://github.com/streetcomplete/StreetComplete/issues/4303#issuecomment-1507617359 and around for  context)

(Tested in https://github.com/mnalis/StreetComplete/actions/runs/4737627521 on e.g. [n2486015492](https://www.openstreetmap.org/node/2486015492), [n2486015499](https://www.openstreetmap.org/node/2486015499) and others)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterials.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterials.kt
@@ -26,6 +26,7 @@ class AddRecyclingContainerMaterials : OsmElementQuestType<RecyclingContainerMat
           amenity = recycling
           and recycling_type = container
           and access !~ private|no
+          and !note and !recycling:note and !note:recycling
     """.toElementFilterExpression() }
 
     override val changesetComment = "Specify what can be recycled in recycling containers"


westnordost:
>  note=* tag (as its existence likely indicates the issue is too complicated for AI, and human is required to evaluate specific issues detailed in freeform text of note)

Wouldn't this apply to any element? Or maybe not, maybe the note could also inform about that the feature isn't visible yet on aerial imagery? A note could include all sorts of stuff, relevant or not relevant.
If someone is on-site however, most if not all relevancy a note might have is void, because the mapper is checking the ground truth on site with his own eyes.

mnalis:
TL;DR: it would make this quest **safer** by avoiding **potentially** problematic recycling containers, while **retaining its full usefulness** (as only a very small percentage of objects would be skipped).

---

> Wouldn't this apply to any element? Or maybe not, maybe the note could also inform about that the feature isn't visible yet on aerial imagery? A note could include all sorts of stuff, relevant or not relevant.

Agreed, it does depends on the note.  It is still a huge indicator that human review of the note is needed. However, as there were only about `1 400` of such combinations (of `recycling_type=container` + `*note*=*`) worldwide, it seemed to me:
- that it would be playing it safe by avoiding such _**potentially** problematic_  elements
- that it would not seriously reduce usefulness of this quest, as in vast majority of the cases it would continue to work as it did before

> If someone is on-site however, most if not all relevancy a note might have is void, because the mapper is checking the ground truth on site with his own eyes.

That is a possibility. But to be sure, one would have to display such `note` to the user, and ask for their feedback if that is now resolved (so SC can remove `note` tag),  or if not, if they want to edit such `note` to add more information they now know as a on-ground surveyor. Such advanced feature would however be:

- somewhat of an overkill given relatively low number of such elements (now, if say 30% of recycling containers had such notes, _or_ if it would considered to apply such feature to **all quests**, it might be worthwhile!)
- maybe not fully in line with SC "be simple" philosophy
- determining whether it would be applicable to all elements would be much bigger task in itself (compared to just checking it in `recycling_type=container` case where it was shown to be the problem in linked issue)
- beyond my capabilities to implement anyway

So I've rather tried to keep it safe and simple instead.

westnordost:
A `note` is not a `fixme`. A `note` could stay forever and also not even hint at a problem, i.e. there is absolutely no necessity to resolve it. If it was a problem, it was a `fixme`.

I get where you are coming from, but my point of view is that:
- it would be inconsistent to only ignore all things tagged with anything `*note*` for this specific quest
- see my last comment, only because **potentially** the presence of a `note` could mean that in one way or another, the data on which the SC quest is based on does not apply is IMO too weak of a reason to generally ignore anything that has a `note`
- You are basically saying that anything `note` should be interpreted as "*do not edit unless you read this!*" and I don't really subscribe to this interpretation. If this interpretation is going to be established, it will just be another thing that makes editing the map somewhat more complicated, being up to the "experts"
- finally, I remember the issue that spawned this PR. And I have to say, the case that was explained (location of separately mapped containers mapped very closely to each other changes regularly), ... well, I see the issue more with how it was mapped, i.e. that they are mapped as separate containers in the first place even though they change/switch location regularly.

mnalis:
> A note is not a fixme. A note could stay forever and also not even hint at a problem, i.e. there is absolutely no necessity to resolve it. 

Yes. But note for example two nodes in example at the top: [n2486015492](https://www.openstreetmap.org/node/2486015492), [n2486015499](https://www.openstreetmap.org/node/2486015499) - those are two nearby containers, and note (in Italian) details them by color. 

If user solving the SC quest was asked for those containers, there is 50% chance that they'd answer the wrong one, thus making that `note` not only "not helpful", but actively incorrect / confusing, creating more problems down the line for next mappers. So it would be better (if far less than ideal) if that `note` was removed in those cases.

So perhaps to fix those we should just remove the `note` when we answer the quest? But that would break other cases like [n4537681043](https://www.openstreetmap.org/node/4537681043) or [n7098727064](https://www.openstreetmap.org/node/7098727064) where the `note` should stay regardless of how the quest was answered.

> If it was a problem, it was a fixme. (Uh, are we ignoring anything fixme? Should we?)

My opionion was (and is) that we should be fixing them instead :smile: e.g.:
- https://github.com/streetcomplete/StreetComplete/issues/3084

(but that one has been implemented in [SCEE](https://github.com/Helium314/SCEE) after being rejected here)

> it would be inconsistent to only ignore all things tagged with anything *note* for this specific quest

We're already looking at notes in other quest like AddPathSurface, AddPitchSurface, AddRoadSurface, AddCyclewayPartSurface, AddFootwayPartSurface... 

And as explained, **I've opted to implement it only for this quest because cost/benefit is clear here**. If I suggested to deploy that it in all 150+ quests, that would probably require me to do hundred and fifty times more cost/benefit analyses for all of them (and like here without guarantee it would not be for nought), and the issues have not been brought up for other quests yet where it would help (while this is at least second issue with recycling containers IIRC), so one might not have much of the justification to invest all that effort.

> see my last comment, only because potentially the presence of a note could mean that in one way or another, the data on which the SC quest is based on does not apply is IMO too weak of a reason to generally ignore anything that has a note

I'm aware. As you probably too remember (from that [KEYS_THAT_SHOULD_BE_REMOVED_WHEN_SHOP_IS_REPLACED](https://github.com/mnalis/StreetComplete-taginfo-categorize) episode) that I'm _**much more opposed to damaging (even small amounts of) data by negligence**_ and would much prefer not to change some part of the map at all, than to (even **if just potentially**) damage it instead of improving it.

Especially if the **cost of avoiding such damage is so tiny** (so the few thousand recycling containers won't be having a SC quest)

> You are basically saying that anything note should be interpreted as "do not edit unless you read this!" and I don't really subscribe to this interpretation. If this interpretation is going to be established, it will just be another thing that makes editing the map somewhat more complicated, being up to the "experts"

But would it really be so horrible **if SC was to show 1% less quests** in order to much reduce potential damage it might lead its users to unknowingly do (even if that damage is just a small percentage of all SC edits)? To me it seems that SC skips many quests that it could be asking (in fact it was one of main reasons for my own fork), so **SC should not really be so averse to reducing that number of quests asked by a tiny amount more for a good cause**.

But anyway, that was my pitch. If you decide to skip all quests with elements that have `note` globally,  great! If you decide to accept just this PR, plenty good. If you decide to close it, I'll (probably :smiling_face_with_tear:) survive too, with NHF.

westnordost:
Any comments about this from other contributors?

matkoniecz:
I actually considered at some point skipping all objects with `note` `fixme` etc where risk of unexpected or already broken data is higher.

Though it never went further than "maybe it is worth investigating is it going to be helpful".

> it would be inconsistent to only ignore all things tagged with anything *note* for this specific quest

I agree, either it should be skipped for all quests (and defined in one place), definitely not implemented by each quest on its own.

>  the issue that spawned this PR. And I have to say, the case that was explained (location of separately mapped containers mapped very closely to each other changes regularly), ... well, I see the issue more with how it was mapped, i.e. that they are mapped as separate containers in the first place even though they change/switch location regularly amongst each other.

In such case SC just revealed already present bad mapping. Mapping each container from group as own `amenity=recycling` is dubious, when container often exchange positions it is simply incorrect.

-------------------

On the other hand, some `note` objects are going to be legitimately tricky and SC has higher chance to damage them.

----------------

> You are basically saying that anything note should be interpreted as "do not edit unless you read this!" and I don't really subscribe to this interpretation. 

I think I at least partially subscribe to this interpretation, but I am not really sure.

-------------------

Overall, I still have no strong preferences but I am not planning on implementing it and I would be against doing it for a single quest.

westnordost:
I will close this then. Sorry, @mnalis .

PR #4961, opened 2023-04-21
Fix keyboard not being dismissed in SearchFeaturesDialog

### LINKED ISSUE #4952: "What's here now instead?": confirmation button covered by keyboard is not reachable
When the alert is shown, after using the keyboard to search for an item in the shop type list, if the keyboard hides the "ok" button, as shown in the picture, there is no way to access it afterwards: trying to use the "back" phone button, as well as the provided one to hide the keyboard, in the bottom right of the screen, only makes the "What's here now instead?" alert to just close, making not possible to commit the chosen item from the shop types list, going back to the "Is this still here?" quest.

Nothing changes when selecting "It's vacant..." or "Something else (leave note)", or when pressing the "Fatto" ("Done", in Italian) blue keyboard button on the bottom right of it.

**How to Reproduce**

1. Open a "Is it still there?" quest
2. Press "No"
3. Select "A(n)..." then search for an item from the list by using the keyboard (e.g., type "pizza")

**Expected Behavior**

Expected the keyboard to hide, then letting the user press the "ok" button.

**Versions affected**

52.0

Android 13, Samsung Galaxy A51, default system keyboard.

tapetis:
Fixes #4952 by manually closing the keyboard when the `SearchFeaturesDialog` is dismissed. The additional parameter `dismissKeyboardOnClose` is used so that in the `ShopsOverlayForm` the keyboard remains open to be able to enter the shop name without having to tap the `EditText` again. This is necessary because I was not able to reliably dismiss the keyboard in any other way but by overwriting the `dismiss` function in the dialog.

I tested this change on an API 24 emulator, where I could reproduce the issue, and on API 33, where I was unable to reproduce the issue.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
@@ -55,7 +55,8 @@ class ShopGoneDialog(
                 countryCode,
                 featureCtrl.feature?.name,
                 ::filterOnlyShops,
-                ::onSelectedFeature
+                ::onSelectedFeature,
+                true
             ).show()
         }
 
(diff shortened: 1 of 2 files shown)


westnordost:
Cool, thank you for the analysis and the fix!

Issue #4968, opened 2023-04-23
AddEntrance quest not asked for multipolygon buildings


Helium314:
Currently the quest only asked if the building is a way, see [AddEntrance.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt#L44-L46)
But at the same time `buildingWaysFilter` contains `relations` too, so I'm not sure whether this is intended.

Anyway, this results in inconsistent behavior, as the quest is asked for some ways touching a building, but not for others, and there is no immediately obvious reason.

matkoniecz:
I can try looking into it and change if doable without massive effort. Can you link some example location?

Helium314:
It's not asked for https://www.openstreetmap.org/relation/7750171, but it was asked for the neighboring building.

I guess the fix would be adding nodes of all ways that are member of the relation.

matkoniecz:
that is because https://www.openstreetmap.org/relation/7750171 has single entrance and https://www.openstreetmap.org/way/31097877 also has single entrance, skipping buildings with a single entrance is deliberate

Helium314:
Umm... ok?
So should I open another issue that for non-multipolygon buildings like .https://www.openstreetmap.org/way/31097877 the entrance quest is (was) asked even if it has only one entrance?

matkoniecz:
oh right, that is adding entrance - not adding entrance ref.

matkoniecz:
@Helium314 Do you want to make a PR if you wrote code already?

deevroman:
> Umm... ok? So should I open another issue that for non-multipolygon buildings like [openstreetmap.org/way/31097877](https://www.openstreetmap.org/way/31097877) the entrance quest is (was) asked even if it has only one potential entrance?

if this problem was solved in this PR (I did not find a another issue), then it has not been fixed yet and got into 52.1. For example, the quest is asked for https://openstreetmap.org/way/23917326 and https://openstreetmap.org/way/141570523

PR #4971, opened 2023-04-25
Ask AddEntrance for multipolygon buildings

Helium314:
see #4968


matkoniecz:
Have you tested it? What about nonmultipolygon relation members? See http://overpass-turbo.eu/s/1ucd

Helium314:
It will also find members of non-multipolygon relation buildings, as I did not change the filter

> Have you tested it? 

yes

matkoniecz:
I guess that ideally there would be also extra filter to get rid of unwanted relations (I will add it unless you figure out how to access tags in that steps - or maybe nonmultipolygons relations are not downloaded/stored by SC anyway).

Helium314:
Adding `and (!type or type = multipolygon)` to the filter should work, as relations are supposed have a type.

matkoniecz:
Again , feel free to add one if you want (or I will add after I dig out of current TODO pile) - but maybe a test case in tests would be nice, with specifically a relation?

Helium314:
btw it would be the test file is not found automatically by Android Studio because of the name "AddBuildingEntranceTest", same for entrance ref quest.

matkoniecz:
> btw the test file is not found automatically by Android Studio because of the name "AddBuildingEntranceTest", same for entrance ref quest.

fixed

> same for entrance ref quest.

I also added functionality for this one, based on your code.

PR #4972, opened 2023-04-25
Show steps dashes in overlays

Helium314:
Currently highlighted steps use the same style as other ways. This is confusing, as it's not immediately clear that they are steps.
With these changes, step dashes show up using outline color.
(the map style changes should also go to the mapstyle repo, but this makes things awkward for anyone wanting to test this PR)

### DIFF
--- a/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
+++ b/app/src/main/assets/map_theme/jawg/streetcomplete.yaml
@@ -65,6 +65,10 @@ styles:
     map_data-lines:
         base: lines
         blend: translucent
+    map_data-steps-dashes:
+        base: lines
+        blend: translucent
+        dash: [0.6,0.4]
     map_data-lines-dashed:
         base: lines
         blend: translucent
@@ -321,6 +325,14 @@ layers:
                           color: function() { return feature.strokeColor; }
                           width: 1m
                           order: function() { return 702 + Number(feature.layer)*5; }
+        steps-dashes:
+            filter: { type: [line], color: true, steps: true }
+            draw:
+                map_data-steps-dashes:
+                    order: function() { return 702 + Number(feature.layer)*5; }
+                    cap: butt
+                    color: function() { return feature.strokeColor; }
+                    width: function() { return feature.width; }
         line-left:
             filter: { type: [line], colorLeft: true, dashedLeft: false }
             draw:
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/StyleableOverlayMapComponent.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/StyleableOverlayMapComponent.kt
@@ -78,6 +78,8 @@ class StyleableOverlayMapComponent(private val resources: Resources, ctrl: KtMap
                     if (style.stroke.dashed) props["dashed"] = "1"
                     props["color"] = style.stroke.color
                     props["strokeColor"] = getDarkenedColor(style.stroke.color)
+                    if (element.tags["highway"] == "steps")
+                        props["steps"] = "1"
                 } else if (style.strokeLeft != null || style.strokeRight != null) {
                     // must have a color for the center if left or right is defined because
                     // there are really ugly overlaps in tangram otherwise


westnordost (changes requested):
This code doesn't belong into the `StyleableOverlayMapComponent` though but it is up to the individual overlays to define how to style the data.

Helium314:
You could say the same for width

westnordost:
I guess you could...

To my defense, I found that since tapping on a way should feel like that way as it is shown in the background map is just selected, it should always have the same width as the way on the map. I.e. if you select a motorway, it would not match the user expectation if the highlighted line is the same width as the highlighted line would be on a footway (doesn't cover the motorway from the background map).

In contrast, whether the highlighted line is dashed or not is already a property that is defined in the `StrokeStyle` (as opposed to the width of the stroke, which is not present there because it is determined automatically). Hence, since whether the stroke is dashed or not is defined by the individual overlay, the user expectation is that the dashed-ness carries a certain meaning, e.g. that a cycle lane is only an advisory cycle lane. It would be confusing if any steps would have a dashed styling regardless.

westnordost:
However, in line 84+ (when left/right is specified but not the center stroke, the overlay mimics drawing the stroke as if it was invisible (because actually making it invisible doesn't work properly). So in this case, making the steps look like steps would make sense even with my line of argumentation. Although, no situation comes to my mind where a `highway=steps` would have a highlighting for left and right. (`StepsHandrailOverlay`? ;-) )

Helium314:
> a way highlighted with a color should just feel like it is the same way as is shown in the backgroun

That was actually my line of thinking when I added the steps dashes

(Another thing like this is the outline dashes for private ways disappearing in overlays. I would expect them to show up the same way as on the background map)

> the user expectation is that the dashed-ness carries a certain meaning

Definitely. But the dashes introduced in this PR are
* not instead, but additional to the line (like normal steps dashes). You get `color` (line) & `strokeColor` (dashes) compared to `color` (dashes) & transparent (line) for other dashed overlay lines
* same distance as normal steps dashes, which is much shorter than dashed overlay lines

So the way simply looks like a colored version of what the user already knows as steps.
Did you try it and think it would really be confusing?

westnordost:
Huh? Then maybe I misunderstood this from reading the code. Could you post a screenshot?

Helium314:
steps
(the weirdly bent dashes are also visible without the overlay)

westnordost:
So how would that look like if the blue has actually a dashed stroke style? (AFAIK currently there is no overlay that uses dashed strokes for the main way.)

Anyway, regarding private ways: That would make sense because you cannot style the `StrokeStyle` to have a stroked outline.

Helium314:
> So how would that look like if the blue has actually a dashed stroke style

Well, it's not the "normal" dashed style, but
```
    map_data-steps-dashes:
        base: lines
        blend: translucent
        dash: [0.6,0.4]
```

With the normal dashed style, the dashes are much longer.

westnordost:
Well, and they done with a shader, which is why they might look different than expected

westnordost:
If it looks something like this, it's fine

grafik

If it looks something like this, it's not:

grafik

Helium314:
Dashes on center line look horrible in general, steps or not.
dashes

westnordost:
Hmm, this looks unexpected... I guess it is not possible to control that the outline and the stroke dashed line have the same phase and frequency...

I think it would look best if
- whether the stroke style is dashed or not never applies to the outline. The outline is always a continuous line
- since then the outline stroke style cannot be styled programmatically for each overlay, it can be styled to always follow the dashed-if-private-and-otherwise-continuous rule
- as for steps... well, I guess it could indeed be as suggested in this PR, as the steps use the outline color while the shader uses the stroke color...

Helium314:
Should the non-dashed outlines be part of this PR?
Because currently it's not used, and it could be that by the time there is an overlay making use of it, SC has already switched to maplibre.

westnordost:
Who knows what all is possible with maplibre and what is not. I suspect many other things need to be changed too when migrating to maplibre. I think it would fit well in this PR. But I guess it doesn't **have** to be in this PR.

Helium314:
> I think it would fit well in this PR. But I guess it doesn't have to be in this PR.

I agree that it would fit, and it would be good to have.
But I already spent more time on this than on steps dashes, and still without result. So I'd prefer to maybe give it another try later.

westnordost:
Fair enough, then this is ready to merge. I'll leave this unmerged and open until when I release the next version though, in case you find the time to work on it in the meantime.

westnordost (approved):

Helium314:
Waiting really helped... I just tried again, and adding dashes for private ways just worked. Not sure what I had done wrong previously.

westnordost:
Uh, I am about to release v52.1... do you want to slip this in real quick? Then I'll wait with the release

Helium314:
No, failed at 

> whether the stroke style is dashed or not never applies to the outline. The outline is always a continuous line

If the line has a stroke, either the stroke applies to outline too, or there is no outline at all. May need more time to fix this, as it still isn't high priority.

westnordost:
ok

PR #4973, opened 2023-04-27
Add steps answers to lit and surface overlays

### LINKED ISSUE #4938: Allow changing path to steps in surface overlay
**Use case**

When mapping with the surface overlay you may come across a section mapped as a path which is actually steps. There is no way to change it at the moment, unlike with the surface quest. If the surface is already set then even disabling the overlay makes it impossible to change to steps (except via e.g. the smoothness quest).

My workaround was to change the section to a wrong surface which would trigger the smoothness quest, use the smoothness quest to say that it was steps and then set the surface correctly so that the steps quests would trigger.

**Proposed Solution**

Add an other answer to allow changing to steps when a path or footway is selected.

arrival-spring:
Fixes #4938 

Mentioned there was also "This is indoors" answer. Currently that only seems to exist for path surface quest, so I left it there

Tested and working as expected

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Steps.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Steps.kt
@@ -1,0 +1,21 @@
+package de.westnordost.streetcomplete.osm
+
+import de.westnordost.streetcomplete.data.osm.edits.update_tags.StringMapChangesBuilder
+
+val KEYS_THAT_SHOULD_BE_REMOVED_WHEN_CHANGING_TO_STEPS = listOf(
+    "smoothness"
+).map { it.toRegex() }
+
+fun StringMapChangesBuilder.changeToSteps(tags: Map<String, String>) {
+    for (key in keys) {
+        if (KEYS_THAT_SHOULD_BE_REMOVED_WHEN_CHANGING_TO_STEPS.any { it.matches(key) }) {
+            remove(key)
+            removeCheckDatesForKey(key)
+        }
+    }
+
+    for ((key, value) in tags) {
+        this[key] = value
+    }
+    this["highway"] = "steps"
+}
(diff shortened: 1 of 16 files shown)


westnordost:
Hmm, AbstractOverlayForm would be the wrong place to implement this functionality. Inheritance should not be used for sharing functionality, you can just have global helper functions for that.

matkoniecz:
> Inheritance should not be used for sharing functionality,

would it be OK to add exactly this wording to https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#code-style ?

arrival-spring:
It seemed to match the other "other answers", so I put it together with those for both overlays and quests.

Where should it go instead?

westnordost:
It should go into the individual overlays. Common functionality can go into some utility functions or if more complex into utility classes, but I see you already did that.

>> Inheritance should not be used for sharing functionality,

> would it be OK to add exactly this wording to https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#code-style ?

Not really, I guess, because this is not really specific to this project but a principle of clean code. I am bad with citing the design pattern names / statements found in programming literature, but regarding inheritance, I remember that inheritance should not be used to share code. That's not really what inheritance is for. Keyword "composition over inheritance", you may also want to look up an essay / rant titled something like "inheritance considered harmful". Rust for example pretty much did away with inheritance - you cannot extend structs (classes), you can only implement traits (interfaces) (the reasoning for that design decision is here: https://doc.rust-lang.org/book/ch17-01-what-is-oo.html ).

arrival-spring:
Ok, I've put the answers into the individual overlays, and simplified the way that the answers are added in the quests.

Again, tested and working

westnordost (approved):

westnordost:
Thanks! LGTM

PR #4976, opened 2023-04-29
curb extensions in parking overlay + add entrances in building overlay

westnordost:
# Main features

One can now add and modify curb extensions in the parking overlay

image

These are added on vertices of roads or if no vertex is near, a new one is created. The cursor in the UI snaps to nearby roads and also snaps to nearby vertices, such as in the example above where it snapped to the crossing.

Same with entrances on building outlines:

https://user-images.githubusercontent.com/4661658/235497181-6aa55fad-9957-4cde-9b7d-4b830c57686d.mp4

## Why?

Why specifically allow adding and modifying traffic calmings in the parking overlay, what has one to do with the other? It's actually even more specific than that: It's only about traffic calmings that (may) narrow the road in one way or another - mostly about curb extensions.
This is important because when mapping street parking, there is somewhat of a fluent transition between `parking:<side>=lane` (street parking) and `parking:<side>=street_side`. See https://wiki.openstreetmap.org/wiki/Street_parking#Distinction_between_street_side_and_lane_parking

# Further changes

- Data structure of element edits changed. Local edits database is cleared on update - a warning must be posted about that in a prior update.

# Todo

All the fundamental changes are done, now just remain some minor UI improvements. Review of the done parts is highly appreciated (also after merge).

- Test this vertex-creation stuff with real data - and also reverting it


app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/create/CreateNodeFromVertexAction.kt
```diff
@@ -0,0 +1,82 @@
...
+/** Action that transforms a way vertex into a node, i.e. adds tags to a vertex in a way.
+ *
+ *  A node in a way (a vertex) without any tags has no meaning in itself but just serves to define
+ *  the geometry of a way. Hence, if we add tags to such a node, it is like creating a new node -
+ *  not on the technical level, but on the semantic level.
+ *
+ *  So, when this is edit is applied, it should be made sure that the vertex didn't change its
+ *  meaning in the definition of the geometry: I.e. its position and the ways it is part of.
+ *
+ *  So, a conflict will be thrown if:
```
Helium314, line 29: Probably it should also count as a confict if the vertex now has tags, because "A node in a way (a vertex) without any tags has no meaning in itself but just serves to define the geometry of a way" isn't correct any more.

westnordost: Hm well, the action exists because the assumption is that the node did not have any intrinsic meaning other than for geometry before. But there is no reason why it couldn't also have had an intrinsic meaning before. The important part is the part about the geometry.

For example, a node that is also a `highway=crossing`, like in the screenshot, actually. Or an entrance of a building.

Helium314: If one user adds an entrance to a building node, and meanwhile someone else has added `amenity=post_box` to the same node, should adding `entrance=yes` to the post box really be uploaded?

westnordost: Well, nevermind someone adding `amenity=post_box` in the meantime - what should happen if it already **has** (such) tags? After all, being able to add `traffic_calming=choker` to a crossing or an address to a `entrance=yes` is certainly something that should be possible. And `CreateNodeFromVertexAction` is the right action to choose because the node is selected due to its geometry, not due to it having certain tags.

Helium314: Then the node had tags before, and

> A node in a way (a vertex) without any tags has no meaning in itself but just serves to define  the geometry of a way. Hence, if we add tags to such a node, it is like creating a new node - not on the technical level, but on the semantic level.

is not correct, because it's not like creating a new node

westnordost: I understand your point. The sentence you cite refers to a vertex without tags.  But a vertex *with* tags still retains the property that its meaning is **also** to define geometry of the way. Anyway, I'll amend the doc comment to mention that.

westnordost:
So from my point of view, this is ready to merge. With a caveat, see #4980

westnordost:
See #4980: ways that are partly outside the bbox are not recognized. The bbox is the position of the cursor plus 100m in each direction. For building+addresses, it should be less of a problem. Can you confirm that is indeed the issue?

westnordost:
Good finding, thank you! My last commit should have fixed that.

westnordost:
It would make sense to not show the three dots when there are no options, but this is not specific to this PR. As to why no "leave note" is displayed, I don't know.

westnordost:
Ah right, the other answer options are only available for when there is already an element.

app/src/main/java/de/westnordost/streetcomplete/overlays/AbstractOverlayForm.kt
```diff
@@ -201,6 +201,7 @@ abstract class AbstractOverlayForm :
             element?.let { getNameAndLocationLabel(it, resources, featureDictionary) }
         )
 
+        binding.moreButton.isGone = assembleOtherAnswers().isEmpty()
```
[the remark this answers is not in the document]
westnordost: Oh, hmm...

app/src/main/java/de/westnordost/streetcomplete/util/math/SnapToWayUtils.kt
```diff
@@ -0,0 +1,113 @@
...
+                continue
+            }
+            val point = positions[i]
+
+            val distance = distanceTo(point)
+            if (distance < minDistance && distance <= maxDistance) {
+                minDistance = distance
+                nearestNodeId = nodeId
+                nearestPoint = point
+                nearestWayIds = mutableSetOf()
```
Helium314, line 71: Shouldn't this be `mutableSetOf(way.id)`?
I noticed that I often get `VertexOfWay` with an empty list of way ids, which doesn't really make sense, and this seems to be the cause.

westnordost: You are right, thank you! Though, it should not lead to any bug, because the ways of the node are fetched independently on creating the action.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsController.kt
```diff
@@ -69,18 +73,27 @@ class ElementEditsController(
...
+            elementUpdates.idUpdates.flatMapTo(editIdsToUpdate) {
+                editElementsDB.getAllByElement(it.elementType, it.oldElementId)
+            }
+            for (id in editIdsToUpdate) {
+                val oldEdit = editsDB.get(id) ?: continue
+                val updatedEdit = oldEdit.copy(action = oldEdit.action.idsUpdatesApplied(idUpdatesMap))
+                editsDB.put(updatedEdit)
+                // must clear first because the element ids associated with this id are different now
+                editElementsDB.delete(id)
+                editElementsDB.put(id, updatedEdit.action.elementKeys)
```
Helium314, line 91: Is there a reason why `oldEdit` is not deleted?
It shouldn't matter much because the upload will fail anyway, but it "doesn't feel right"

Helium314: some more possible issues (or just me not understanding what's going on?):
* should `editElementsDB.put(id, updatedEdit.action.elementKeys)` rather use `updatedEdit.id`?
* should `elementIdProviderDB` also be updated to use the new edit id?

westnordost: 1. the `oldEdit` is supposed to be overwritten with `editsDB.put(updatedEdit)`. The `id` is the primary key. Will add a test for that.

PR #4984, opened 2023-05-02
Add support for the predictive back gesture

tapetis:
This PR adds support for the the predictive back gesture introduced in Android 13 (see [here](https://developer.android.com/guide/navigation/predictive-back-gesture) for details). The predictive back gestures use an ahead-of-time model for back event handling (implemented using the AndroidX `onBackPressedDispatcher` and `FragmentManager`) instead of the deprecated `onBackPressed` callback that is only called once the back button was already pressed. According to the developer documentation, supporting this functionality will be mandatory for a future Android version.

The following video shows this feature in action on an Android 14 device with "Predictive back animations" enabled in the developer settings:

https://user-images.githubusercontent.com/6892794/235721284-7ccdce95-48a8-4094-8943-4c43700125dd.mp4

I tested that the back button still works correctly with the following features (including after recreation of the `Activity` or `Fragment`):
- Map screen: bottom sheets, messages (hints, unread messages, change log, achievements), and history side bar
- Settings and about screens: detail fragments and debug menu
- User screen: achievements, edit stats, and login

I hope this covers all the places where the back button is used.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -29,6 +29,7 @@
     <application
         android:name="de.westnordost.streetcomplete.StreetCompleteApplication"
         android:allowBackup="false"
+        android:enableOnBackInvokedCallback="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/BackPressedListener.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/BackPressedListener.kt
@@ -1,6 +1,0 @@
-package de.westnordost.streetcomplete.screens
-
-interface BackPressedListener {
-    /** Return true to consume the event */
-    fun onBackPressed(): Boolean
-}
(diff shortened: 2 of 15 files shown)


westnordost:
Wow, thank you! A shitty move by google to require things like that, this drives up the maintenance cost of apps and hence small (FOSS) projects will fall off the plate. So, thank you for doing that.

Further places where back press is used:

- Quest forms
- overlay forms
- note creation form
- faux dialogs in profile screen that show quest details, country details and achievement details
- quest selection, preset selection etc. screens in settings screen
- changelog, credits etc. fragments in the about screen
- special screens like the split way fragment, move node fragment

westnordost (commented):
I've scrolled through the changes. I'd not be able to point out any mistakes as the API is new to me. So, code looks fine.

westnordost:
I wonder if at some point Google makes Material 3 mandatory. I've not upgrading to it so far because I found it looked ugly / made many things look broken in the app.

westnordost (approved):
I've tried all places in the app I can think of to use the Android 13 back function. All of it works without issue.

By the way, aren't the gesture controls much older than Android 13 already?

Issue #4980, opened 2023-05-01
`MapDataController::getMapDataWithGeometry` returns incomplete Ways


westnordost:
I think as part of @Helium314 's extensive caching PR or even a bit earlier than than, we made bounding box queries to local map data mimic the behavior of the OpenStreetMap API:

1. get all nodes in bounding box
2. get all ways that use any of these nodes
3. get all relations that use any of these nodes or any of these ways

However, we have overlooked something: The OpenStreetMap API never returns incomplete ways. In other words, if a way is included in the result set, it always also includes all nodes of the way, even if some of these nodes are not in the given bounding box. Currently, `MapDataController::getMapDataWithGeometry` returns all ways that have nodes within the given bounding box, but omits nodes of these ways that are not within the given bounding box.

This inconsistency has not been deliberate, but an oversight (at least from my part). I am pretty sure this is the source of #4925, maybe others. It definitely affects #4976.

A fix would look like this in the `ElementDao` (plus tests)...

```diff
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/ElementDao.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/ElementDao.kt
@@ -45,12 +45,19 @@ class ElementDao(
 
     fun getAll(bbox: BoundingBox): List<Element> {
         val nodes = nodeDao.getAll(bbox)
-        val nodeIds = nodes.map { it.id }
+        val nodeIds = nodes.map { it.id }.toSet()
         val ways = wayDao.getAllForNodes(nodeIds)
         val wayIds = ways.map { it.id }
-        val relations = relationDao.getAllForElements(nodeIds = nodeIds, wayIds = wayIds)
-        val result = ArrayList<Element>(nodes.size + ways.size + relations.size)
+        val additionalWayNodeIds = ways
+            .asSequence()
+            .flatMap { it.nodeIds }
+            .filter { it !in nodeIds }
+            .toList()
+        val additionalNodes = nodeDao.getAll(additionalWayNodeIds)
+        val relations = relationDao.getAllForElements(nodeIds = additionalWayNodeIds + nodeIds, wayIds = wayIds)
+        val result = ArrayList<Element>(nodes.size + additionalNodes.size + ways.size + relations.size)
         result.addAll(nodes)
+        result.addAll(additionalNodes)
         result.addAll(ways)
         result.addAll(relations)
         return result
```

[...]

PR #4985, opened 2023-05-02
Fix issue with cache

Helium314:
see https://github.com/streetcomplete/StreetComplete/issues/4980#issuecomment-1531960544

No tests yet, so far I failed to reproduce the bug in a unit test.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt
@@ -6,6 +6,7 @@ import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
 import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometryEntry
 import de.westnordost.streetcomplete.data.osm.geometry.ElementPointGeometry
 import de.westnordost.streetcomplete.util.SpatialCache
+import de.westnordost.streetcomplete.util.math.contains
 
 /**
  * Cache for MapDataController that uses SpatialCache for nodes (i.e. geometry) and hash maps
@@ -373,6 +374,8 @@ class MapDataCache(
             // get dropped when the caches are updated
             // duplicate fetch might be unnecessary in many cases, but it's very fast anyway
 
+            // get(bbox) for tiles not in spatialCache calls spatialCache.fetch, but this is still
+            // safe as tiles are replaced and properly filled as part of the following update
             nodes = HashSet<Node>(spatialCache.get(bbox))
             update(updatedElements = elements, updatedGeometries = geometries, bbox = fetchBBox)
 
@@ -383,8 +386,13 @@ class MapDataCache(
                 return result
             }
 
-            // get nodes again, this contains the newly added nodes, but maybe not the old ones if cache was trimmed
-            nodes.addAll(spatialCache.get(bbox))
+            // add newly fetched nodes from elements
+            // getting nodes from spatialCache can cause issues, as tiles in the bbox may now be removed unexpectedly
+            // see https://github.com/streetcomplete/StreetComplete/issues/4980#issuecomment-1531960544
+            for (element in elements) {
+                if (element !is Node) continue
+                if (element.position in bbox) nodes.add(element)
+            }
         } else {
             nodes = spatialCache.get(bbox)
         }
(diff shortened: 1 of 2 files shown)


westnordost:
TBH I understand neither the problem nor the solution. But this is not a requirement, unless you need feedback from me.

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt
```diff
@@ -383,8 +386,13 @@ class MapDataCache(
                 return result
             }
 
-            // get nodes again, this contains the newly added nodes, but maybe not the old ones if cache was trimmed
-            nodes.addAll(spatialCache.get(bbox))
+            // add newly fetched nodes from elements
+            // getting nodes from spatialCache can cause issues, as tiles in the bbox may now be removed unexpectedly
```
westnordost, line 390: how/when though? This is all synchronized, isn't it?

Helium314:
What happens:
Tile 1 is cached from a previous fetch, then `getMapDataWithGeometry` is called for a bbox (exactly) containing tile 1 and _adjacent_ tile 2.
[`tilesToFetch`](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt#L360) will contain tile 2 only, because tile 1 is already cached.
In `nodes = HashSet<Node>(spatialCache.get(bbox))`, the `SpatialCache` will use [its own `fetch`](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/util/SpatialCache.kt#L117-L120), but `SpatialCache.fetch` [always returns an empty list](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt#L29) for tile 2. This adds tile 2 to the `SpatialCache`, containing the empty list returned by `SpatialCache.fetch`.
At this point there is no issue, because the tile being empty is expected, and we replace the fetched empty tile 2 with the actual data as part of the [immediately following `update`](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt#L64).
But during the `update`, the `tilesRect.asBoundingBox(zoom).enclosingTilesRect(zoom)` bug(?) sometimes kicks in, and the unexpectedly [extended larger tiles rect](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/util/SpatialCache.kt#L86) suddenly contains tile 1 and tile 2, instead of tile 2 only. Tile 1 is then seen as [incomplete](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/util/SpatialCache.kt#L87), because it's not fully in the bbox that was supposed to contain tile 2 only.
Since tile 1 is supposed to be replaced with incomplete data, tile 1 is removed from `SpatialCache`.

Next, `MapDataCache` fetches the nodes added to `SpatialCache` in `update`, and for shorter code it was done using [`SpatialCache.get`](https://github.com/streetcomplete/StreetComplete/blob/59e44eaae360181eb649df19b63ae775d4d8476d/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt#L387). I had seen this as safe, but now we know it's not.
As described above, when now calling `get` for the full bbox of tile 1 and tile 2, the previously removed tile 1 is fetched using `SpatialCache.fetch`, so it contains no nodes. But this time `MapDataCache` isn't aware of this, and the empty tile is seen as properly cached. On this fetch, the returned data is still correct. On further fetches in tile 1, nothing will be returned.

With this PR, the second `SpatialCache.fetch` is not done, and thus if the `tilesRect.asBoundingBox(zoom).enclosingTilesRect(zoom)` bug occurs, the removed tile 1 is simply fetched again from DB.

app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt
Helium314: Updating the cache may remove tiles, see comment below for an explanation.

Helium314:
added a test that fails without the change to `MapDataCache`

Helium314:
I'll merge this, as it works, and with the PR merged I'll continue #4980 (might actually be finished already, assuming the issues I observed were really caused by this bug only).

westnordost:
Thank you for the explanation. 

It makes sense to not rely on `tilesRect.asBoundingBox(zoom).enclosingTilesRect(zoom) == tilesRect` because when one data structure is inclusive and the other exclusive to its bound, it is somewhat difficult or impossible to have a lossless conversion between the two.

Issue #4970, opened 2023-04-25
address numbers are not shown in address overlay view


JZersche:
Addresses aren't being displayed while in "Address Overlay" view mode, despite the fact that the area has full address information available. Navigating to an area where addresses haven't been filled in, and then while in "Address Overlay" mode, you can start adding the house numbers, and they'll show up. But if you stay in address overlay mode and go to the area where you know addressing information is available, the addresses aren't visible until you leave the address overlay mode, in which case, navigating back to the original location where you just filled in addresses, they are no longer visible. 

So the house numbers don't show up while in overlay mode, unless you add them manually, in which case the area may already be properly addressed leading to unnecessary edits..

arrival-spring:
Can you link to the area where this happens?

House numbers are meant to be shown in the overlay.

House numbers will also be shown on the map, when not in an overlay, when you're zoomed in a lot. However if you have only just added them then the map won't have updated yet.

JZersche:
Baltimore: https://www.openstreetmap.org/?mlat=39.27499&mlon=-76.61317#map=17/39.27498/-76.61317

Pretty much all of Baltimore City, Maryland is completely mapped out in OSM, **but the addresses will only show you have you have overlay set to none.**

Setting the overlay to addresses will show addresses for any areas outside of Baltimore assuming there is data, actually I've found that anywhere I've looked there seems to be no address information other than Baltimore, but Cedar Hill Lane is an area outside of Baltimore where they started building new houses recently about 2 years ago and it's been filled in with data and works with address overlays enabled, and with overlays set to none.

https://www.openstreetmap.org/?mlat=39.21104&mlon=-76.61051#map=19/39.21104/-76.61051


Another area that has address information mapped out that only shows addresses if overlay is set to None is located here:
https://www.openstreetmap.org/?mlat=39.16620&mlon=-76.73059#map=19/39.16620/-76.73059

I can see addresses on those houses if I set overlays to none, but if I set it to addresses they disappear.

Helium314:
Works nicely for me. With the address overlay, many more house numbers are shown than without.
Did you actually download the map data (means: you see quests) in the area where you have the issue?

JZersche:
> Works nicely for me. With the address overlay, many more house numbers are shown than without. Did you actually download the map data (means: you see quests) in the area where you have the issue?

Can you give me a location so I can test it on my end and eliminate the potential that my recent addressing isn't the issue? Yes, I've tried updating and downloading new data.

[...]

PR #4986, opened 2023-05-02
Visualize downloaded area

westnordost:
fixes the other thing for #4970

Turned out to be relatively straightforward after all. The further you zoom out, the less visible the hatching will be.

When one is zoomed out very far when the app initializes the map, I could reproduce that the entire world is hatched, even the downloaded tiles. The display is only updated after one downloads a tile or tabs out of the app and in again. I checked, it is definitely not an issue with the data but apparently with the rendering 🤷 
Not sure what to do about that. But otherwise, it works.

Screenshot_20230502_220341
Screenshot_20230502_220009
Screenshot_20230502_220126
Screenshot_20230502_220158
Screenshot_20230502_220213

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/data/download/tiles/DownloadedTilesDaoTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/data/download/tiles/DownloadedTilesDaoTest.kt
@@ -16,79 +16,74 @@ class DownloadedTilesDaoTest : ApplicationDbTestCase() {
     }
 
     @Test fun putGetOne() {
-        dao.put(r(5, 8, 5, 8), "Huhu")
+        dao.put(r(5, 8, 5, 8))
 
-        assertEquals(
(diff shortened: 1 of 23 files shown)


Helium314:
I didn't test, but does this show that nothing is downloaded when exiting team mode? (because then `removeAll` is called)

westnordost:
Yes

westnordost:
But yeah, that's not so nice. I'll make it so that instead of `removeAll`, the download time is set to 0, so it is always considered out of date.

westnordost:
Done.

westnordost:
Also, worked around that tangram bug

Helium314:
> But yeah, that's not so nice. I'll make it so that instead of `removeAll`, the download time is set to 0, so it is always considered out of date.

Thanks!
(At some point I had tried to show a warning in SCEE when creating nodes and notes outside the downloaded area, but often got seemingly incorrect warnings because tiles had been removed.)

PR #4991, opened 2023-05-03
Improve layout of history items

tapetis:
This PR improves the layout of history items in the sidebar on both small and large screens. As can be seen from the screenshots below, the texts and images are now both centered correctly and the dividers always span the entire width of the history sidebar.

| Old phone  | Old tablet | New (phone & tablet) |
| ------------- | ------------- | ------------- |
|  |  |   |

I also removed some unnecessary XML attributes (`ImageView`s are visible and not clickable by default) and aligned the divider height with other dividers in the app.


westnordost:
Cool, I didn't notice that the layout was broken on tablets before. Thank you!

PR #4992, opened 2023-05-03
Use FragmentContainerView in more places

tapetis:
This PR replaces FrameLayouts that are used as fragment containers with [FragmentContainerViews](https://developer.android.com/reference/androidx/fragment/app/FragmentContainerView). This ensures that animations of fragment transactions are shown correctly on all API levels. During testing, I did not notice any issues with these changes.

There are also some RelativeLayouts used as fragment containers that I did not replace, because many inner fragments rely on `layout_alignParent*` to be displayed on the left screen side even for RTL languages. These places are now all correctly annotated in the layout files. Theoretically, however, it should also be possible to use FragmentContainerViews in these cases, but then `layout_alignParent*` would have to be replaced by `layout_gravity` everywhere.


westnordost:
Looks good to me, thank you!

PR #5003, opened 2023-05-07
Tactile steps

### LINKED ISSUE #3534: [Quest] Do these steps have tactile paving at the top and bottom?

arrival-spring:
Fixes #3534 now that it is not marked as blocked, follows on from work in #3762

Again, my first attempt at an icon (combining tactile paving and steps icons)

There are three new strings here as the current string for incorrect tactile paving is "Only on one side" and this doesn't make sense for steps, so it needed at least a string "Only at one end" and to me it makes sense to have separate options for top and bottom.

Tested and working and tests all pass


res/graphics/quest/steps_tactile_paving.svg
westnordost, line 1: The right end of the steps are a bit too close to the edge.

In Inkscape I usually add a grid (cell size 4x4) and try to align the drawings to that, for better visibility when scaled down.

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingSteps.kt
```diff
@@ -0,0 +1,39 @@
...
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.surface.ANYTHING_PAVED
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+
+class AddTactilePavingSteps : OsmFilterQuestType<TactilePavingStepsAnswer>() {
+
+    override val elementFilter = """
+        ways with highway = steps
+         and surface ~ ${ANYTHING_PAVED.joinToString("|")}
+         and (!indoor or indoor = no)
```
westnordost, line 16: Hmm, why not indoor, though? Maybe rather not conveying?

app/src/main/java/de/westnordost/streetcomplete/quests/tactile_paving/AddTactilePavingSteps.kt
```diff
@@ -0,0 +1,39 @@
...
+         and access !~ private|no
+        and (
+          !tactile_paving
+          or tactile_paving = unknown
+          or tactile_paving = no and tactile_paving older today -4 years
```
westnordost, line 21:
```suggestion
          or tactile_paving ~ no|partial|incorrect and tactile_paving older today -4 years
```
?

westnordost (commented):
Thank you, looks good. Just a few remarks

res/graphics/quest/steps_tactile_paving.svg
arrival-spring: I've adjust it, how's that?

westnordost (approved):

res/graphics/quest/steps_tactile_paving.svg
westnordost: It's better, thanks

PR #5008, opened 2023-05-10
Improve English wording

arrival-spring:
"part of the year" sounds awkward, we'd say "time of year"

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -108,7 +108,7 @@ The info you enter is directly added to OpenStreetMap in your name, without the 
     <string name="default_disabled_msg_no_ar">You need a tape measure to solve this quest because measuring with the camera is not supported on this device.</string>
     <string name="default_disabled_msg_difficult_and_time_consuming">This quest type is disabled by default because it is either time-consuming or difficult to make the right choice. Only enable this quest type if you are really committed to adding this kind of data.</string>
     <string name="default_disabled_msg_overlay">This quest type is disabled by default because there is already an overlay with which it is more effective and efficient to enter and update this data.</string>
-    <string name="default_disabled_msg_seasonal">"This quest type is disabled by default because it may require re-checking the situation at a specific time of year."</string>
+    <string name="default_disabled_msg_seasonal">"This quest type is disabled by default because it may require re-checking the situation at different times of the year."</string>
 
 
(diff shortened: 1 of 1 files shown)


westnordost:
The intention is to convey that it might be only in summer or only in winter or something.

"specific time of year" to my ears sounds a bit like "christmas" or "easter", i.e. more specifically more granular.

app/src/main/res/values/strings.xml
```diff
@@ -108,7 +108,7 @@ The info you enter is directly added to OpenStreetMap in your name, without the
     <string name="default_disabled_msg_no_ar">You need a tape measure to solve this quest because measuring with the camera is not supported on this device.</string>
     <string name="default_disabled_msg_difficult_and_time_consuming">This quest type is disabled by default because it is either time-consuming or difficult to make the right choice. Only enable this quest type if you are really committed to adding this kind of data.</string>
     <string name="default_disabled_msg_overlay">This quest type is disabled by default because there is already an overlay with which it is more effective and efficient to enter and update this data.</string>
-    <string name="default_disabled_msg_seasonal">"This quest type is disabled by default because it may require re-checking the situation at a specific part of the year."</string>
+    <string name="default_disabled_msg_seasonal">"This quest type is disabled by default because it may require re-checking the situation at a specific time of year."</string>
```
mnalis, line 111:
```suggestion
    <string name="default_disabled_msg_seasonal">"This quest type is disabled by default because it may require re-checking the situation at different seasons."</string>
```

arrival-spring: rather "in different seasons"

arrival-spring:
> The intention is to convey that it might be only in summer or only in winter or something.
> 
> "specific time of year" to my ears sounds a bit like "christmas" or "easter", i.e. more specifically more granular.

Hmm, a little bit, something like "at different times of the year" could also work or "different seasons", as suggested.

westnordost:
"at different times of the year" would be fine for me (because it is most open/generic)

westnordost:
Where are you from, by the way? What variant of English is your mother tongue?

arrival-spring:
I'm from the UK so British English is my mother tongue

mnalis (approved):

peternewman (approved):
Sounds like an improvement to me @arrival-spring !

PR #5013, opened 2023-05-12
Skip Building Level quest with existing building:min_level

HolgerJeromin:
If the building has an existing `building:min_level` tag, this object is too complicated without having more information.

I got it on a miss tagged roof, but there are other elements which would hit

https://www.openstreetmap.org/way/245685307
https://www.openstreetmap.org/way/851896988

https://overpass-turbo.eu/s/1uS4


matkoniecz:
have you tested is it working?

HolgerJeromin:
> have you tested is it working?

Yes. Tested with the first exampe way.

mnalis (approved):

Issue #2694, opened 2021-03-29
Asking the colour of buildings.


Mbodin:
### General
Affected tag(s) to be added: [`building:colour`](https://wiki.openstreetmap.org/wiki/Key:building:colour) (and [`roof:colour`](https://wiki.openstreetmap.org/wiki/Key:roof:colour) in a similar but separate quest)
Example of items with these tags: https://www.openstreetmap.org/way/272006373 and https://www.openstreetmap.org/way/62278743

The colour of buildings and roofs is mainly used in 3D renderings like [OSMBuildings](https://osmbuildings.org)—it could also in theory be used for 2D maps, but I haven’t seen any using these tags.
Some neighbourhoods are famous for their colourful houses, and such a quest can help noticing them in maps.
It can also be useful to help spreading the use of OSM: one feedback I often get when sharing OSMBuildings links to non-OSM people is that its buildings are colorless. It could also be helpful when planning walks to choose colourful streets, which are not always known.
Finally, colours might be useful as they provide additional reference points (e.g., “let’s meet next to the red building on this street”).

### Checklist
Checklist for quest suggestions (see guidelines):
- [X] 🚧 To be added tag is established and has a useful purpose
- [ ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one): all building have colors. However, there is no `many` value for the tag `building:colour`: there might be some corner cases in which the user would just want to say that there are many colors without picking any particular.
- [X] 🐿️ Easily answerable by everyone from the outside but a survey is necessary: see discussion below for how to pick colours. The roof colour may be taken from aerial imagery instead of a survey.
- [X] 💤 Not an overwhelming percentage of elements have the same answer (No spam): this may depend on the region. I guess that this is similar than the building-level quest.
- [X] 🕓 Applies to a reasonable number of elements (Worth the effort).

### Ideas for implementation

**Element selection:** 
As this key is currently mainly used by 3D renderer, I suggest only asking for buildings with a number of levels already referenced.

For `roof:colour`, I suggest selecting all building with a `roof:shape` key already filled in, filtering out high buildings and flat roofs.

These selections would also have the advantage that these building quests would be late, thus focussing first on the important properties of buildings (building type and number of levels) before more aesthetic properties (like roof shape and building colours—although roof shape has non-aesthetic applications).

**Ideas for the GUI:**
[...]

Issue #1358, opened 2019-03-24
Building colour and material


jluttine:
### General

First, thanks a lot for StreetComplete! It is just so much fun completing the quests!

Affected tag(s) to be modified/added: [building:colour](https://wiki.openstreetmap.org/wiki/Key:building:colour)
Question asked: **What is the colour of the building?**

### Checklist
Checklist for quest suggestions (see guidelines):
- [ ] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

Though the colour of the building isn't very useful in ordinary OSM usage, it is very nice information for [OSM Buildings](https://osmbuildings.org/). Marking building colours is very tedious otherwise but with StreetComplete it could be very simple. I think it goes a bit to the same category as building levels, as far as I understand.

ENT8R:
Is there anybody else who could use that specific kind of data?
I don't think so... That's just a very specific use case and a massive flood of quests IMO...

jluttine:
@ENT8R Yep, I'm not aware of others. And I definitely agree with your view. At least it should be at the bottom of the quest priority list.

Building material could also be queried. But yes, I'm not sure how useful it would be. Would be nice to know where and how color and/or material information is used - if anywhere except OSM Buildings.

westnordost:
Sorry, I find color and material (which is the surface material, btw) too unimportant to be added on StreetComplete.

jluttine:
I understand! :+1: By the way, do you know where building roof type and building levels are used as they have quests? Levels of course is quite useful information in general, but does the roof type have some use cases in some map renderers or elsewhere, do you know? Just trying to understand how these metadata are utilized.

westnordost:
Roof type probably only 3d maps, building levels could f.e. additionally be used to estimate population density.

jluttine:
[...]

PR #5018, opened 2023-05-13
New quest: Roof colour

hangy:
Adds a basic quest asking for roof colours. 

image
image

As my first attempt at adding new quests to StreetComplete, this is heavily based on the existing AddRoofShape quest. The conditions that define whether the quest is shown, are the same as the original. This quest currently uses the example colours from [key:roof:colour](https://wiki.openstreetmap.org/wiki/Key:roof:colour), but changing the colour palette is simple enough.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -122,6 +122,7 @@ import de.westnordost.streetcomplete.quests.religion.AddReligionToPlaceOfWorship
 import de.westnordost.streetcomplete.quests.religion.AddReligionToWaysideShrine
 import de.westnordost.streetcomplete.quests.road_name.AddRoadName
 import de.westnordost.streetcomplete.quests.road_name.RoadNameSuggestionsSource
+import de.westnordost.streetcomplete.quests.roof_colour.AddRoofColour
 import de.westnordost.streetcomplete.quests.roof_shape.AddRoofShape
 import de.westnordost.streetcomplete.quests.seating.AddSeating
 import de.westnordost.streetcomplete.quests.segregated.AddCyclewaySegregation
@@ -491,6 +492,7 @@ fun questTypeRegistry(
     150 to AddBuildingType(),
     151 to AddBuildingLevels(),
     152 to AddRoofShape(countryInfos, countryBoundariesFuture),
+    200 to AddRoofColour(countryInfos, countryBoundariesFuture),
 
     153 to AddStepCount(), // can only be gathered when walking along this way, also needs the most effort and least useful
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/roof_colour/AddRoofColour.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/roof_colour/AddRoofColour.kt
@@ -1,0 +1,71 @@
+package de.westnordost.streetcomplete.quests.roof_colour
+
+import de.westnordost.countryboundaries.CountryBoundaries
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
+import de.westnordost.streetcomplete.data.meta.CountryInfos
+import de.westnordost.streetcomplete.data.meta.getByLocation
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BUILDING
+import de.westnordost.streetcomplete.osm.BUILDINGS_WITH_LEVELS
+import de.westnordost.streetcomplete.osm.Tags
+import java.util.concurrent.FutureTask
+
+class AddRoofColour(
+    private val countryInfos: CountryInfos,
+    private val countryBoundariesFuture: FutureTask<CountryBoundaries>,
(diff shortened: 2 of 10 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -491,6 +492,7 @@ fun questTypeRegistry(
     150 to AddBuildingType(),
     151 to AddBuildingLevels(),
     152 to AddRoofShape(countryInfos, countryBoundariesFuture),
+    200 to AddRoofColour(countryInfos, countryBoundariesFuture),
```
hangy, line 495: Index/sort order is currently a placeholder. I'm not sure where this should be placed in practice. 🤷

app/src/main/java/de/westnordost/streetcomplete/quests/roof_colour/AddRoofColour.kt
```diff
@@ -0,0 +1,71 @@
...
+    override val wikiLink = "Key:roof:colour"
+    override val icon = R.drawable.ic_quest_roof_colour
+    override val achievements = listOf(BUILDING)
+    override val defaultDisabledMessage = R.string.default_disabled_msg_roofColour
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_roofColour_title
+
+    override fun createForm() = AddRoofColourForm()
+
+    override fun getApplicableElements(mapData: MapDataWithGeometry) =
```
hangy, line 42: `getApplicableElements`, `isApplicableTo`, and `roofsAreUsuallyFlatAt` are straight up copied from `AddRoofShape`. According to the guidelines, using inheritance for this would be discouraged. Should the methods be moved somewhere in the `de.westnordost.streetcomplete.util` package, or is that kind of duplication wanted for readability of the quests?

app/src/main/res/drawable/ic_quest_roof_colour.xml
hangy, line 1: Please do consider this a placeholder. I modified `ic_quest_roof_shape` a bit to have more than one colour, but it wouldn't surprise me if there were better suggestions. 😅

westnordost:
I am sorry, but building and roof color quests are not going to be included in StreetComplete. See #2694, in particular https://github.com/streetcomplete/StreetComplete/issues/1358#issuecomment-595232339 and following comments.

Apart from that, roof color is really pretty difficult to ascertain from the ground. It would be much more efficient and effective to use a pipette tool on the aerial pictures for that, i.e. tag it in JOSM / iD, if you still think that it makes sense after reading the link mentioned above.

HolgerJeromin:
Perhaps this is something for @Helium314 
and his fork https://github.com/Helium314/SCEE/

For roof color I would be ok-ish to add it to SCEE (with a few less red colors).

hangy:
> I am sorry, but building and roof color quests are not going to be included in StreetComplete. See #1358, in particular [#1358 (comment)](https://github.com/streetcomplete/StreetComplete/issues/1358#issuecomment-595232339) and following comments.
> 
> Apart from that, roof color is really pretty difficult to ascertain from the ground. It would be much more efficient and effective to use a pipette tool on the aerial pictures for that, i.e. tag it in JOSM / iD, if you still think that it makes sense after reading the link mentioned above, that is.

I think it can be quite useful for 3D renderers, but can totally respect your decision.

> Perhaps this is something for @Helium314 and his fork https://github.com/Helium314/SCEE/
> 
> For roof color I would be ok-ish to add it to SCEE (with a few less red colors).

Will have a look, thanks!

Helium314:
I actually had thought about adding a (building) color quest to SCEE, but didn't because I don't see much use of `colour`, and thus didn't want to spend much time here.

However, if you're interested in having the quest in SCEE that would be fine. Maybe with some adjustments, like preferring (extended) CSS color names over RGB values.

mnalis:
> but didn't because I don't see much use of [colour](https://wiki.openstreetmap.org/wiki/Key:colour)

Neither did I, but I guess https://wiki.openstreetmap.org/wiki/Streets_GL **_is_** highly addictive (now, my neighborhood has differently colored buildings, so if that went in, I'd be quite tempted to reuse that code to start tagging them. Just the _roof_ colors is not that big a motivator for me, though).

> preferring (extended) CSS color names over RGB values

I agree that would be better

westnordost:
By the way, I totally get the hype regarding streets gl and wanting to have more quests for that. It **is** awesome. However, a guideline for StreetComplete is that only things should be recorded with the app that have also use case other than rendering.

So,  if you want to make your city pretty in 3D, how about trees?

There could be a tree overlay in which you could add trees and e.g. specify the trunk circumference to make them render in different sizes. Maybe, ... more advanced, but... maybe it is possible to also record the tree species with the help of "PlantNet" - maybe there is an API for that. 

A tree overlay is something I'd gladly merge.

hangy:
> > but didn't because I don't see much use of [colour](https://wiki.openstreetmap.org/wiki/Key:colour)
> 
> Neither did I, but I guess https://wiki.openstreetmap.org/wiki/Streets_GL **_is_** highly addictive

streets.gl was definitely a motivator for this PR 😆

> (now, my neighborhood has differently colored buildings, so if that went in, I'd be quite tempted to reuse that code to start tagging them. Just the _roof_ colors is not that big a motivator for me, though).

I was going to add a `building:colour` quest (and potentially a `building:material`) next, but just wanted to start with a simple one first.

> > preferring (extended) CSS color names over RGB values
> 
> I agree that would be better

I might just use of the top used values from taginfo, and add some outliers. To completely fair, neither roof nor building colour will always be super accurate, and will usually only serve to differentiate buildings a little bit. The perceived colour will depend on the lighting and [time of day](https://www.researchgate.net/figure/Color-temperature-variation-throughout-a-single-day_fig3_328726901). Temperature would probably also affect colour detection from a photo (which might be used like in StreetMeasure). In a perfect world we might light each building in a neutral temperature, and use photogrammetry to get fully textured models of buildings, but I fear that's not gonna happen on a larger scale 😉

Issue #5005, opened 2023-05-07
Crash when adding node in parking overlay


Helium314:
**How to Reproduce**
* split a long way road near one end
* go to the other end, enable parking overlay and click the plus button
* see a crash:
```
java.lang.NullPointerException
at de.westnordost.streetcomplete.overlays.street_parking.LaneNarrowingTrafficCalmingForm$initCreatingPointOnWay$1.invoke(LaneNarrowingTrafficCalmingForm.kt:89)
```

**Expected Behavior**
no crash

**Versions affected**
current latest, unreleased: https://github.com/streetcomplete/StreetComplete/commit/a460271dc55ef45c77ee1c2d86d5c1df209e14a9

Crash possibly happens because `modifyBBoxMapData` only adds nodes inside bbox
https://github.com/streetcomplete/StreetComplete/blob/a460271dc55ef45c77ee1c2d86d5c1df209e14a9/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt#L387

westnordost:
Oh, splitting ways... yeah, `MapDataWithEditsSource` probably needs to be edited too.

Helium314:
I only checked that 
```
            if (element is Way) {
                element.nodeIds.forEach { nodeId ->
                    val nodeKey = ElementKey(NODE, nodeId)
                    updatedElements[nodeKey]?.let { mapData.put(it, updatedGeometries[key]) }
                }
            }
```
in `modifyBBoxMapData` helps in the specific case.
But I didn't do more than that, and I don't think this is complete.

PR #5026, opened 2023-05-15
Fix crash when adding node after modifying a node in the same way

Helium314:
closely related to #5005
When editing a node near one end of a long way, and trying to insert a node near the other end, there is a crash because the edited node is not in map data.

This is because nodes outside the bbox are removed in `modifyBBoxMapData`, and only re-added if they are part of a modified way.
But in this case the way is not modified, so the node is missing.

This fix checks before removing a node whether it is part of a way in `mapData`. Since we want updated ways, the loop over `(key, geometry)` is done for ways before nodes.
This might not be optimal for performance, as for each updated node `getWaysForNode` is called.
An alterative idea would be looping over all nodes of all ways in `mapData`, which could be clearly faster or much slower (depending on the number of edits and size of map data).

@westnordost 
If you have some alternative solution, feel free to update the PR, or scrap it and implement it some different way.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -386,10 +386,10 @@ class MapDataWithEditsSource internal constructor(
 
     private fun modifyBBoxMapData(bbox: BoundingBox, mapData: MutableMapDataWithGeometry) = synchronized(this) {
         val addWays = ArrayList<Way>()
-        for ((key, geometry) in updatedGeometries) {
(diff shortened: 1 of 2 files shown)


westnordost:
Sorry for replying so late. I noticed the PR earlier but I needed some time to wrap my head around this.

Looking at `getWaysForNode`, I think the second solution you propose is actually better / more performant because 

- for every `getWaysForNode` call, a new hash map is created, all `updatedElements` are iterated through and all nodes of all updated ways are iterated through

- this is done for every single modified node (regardless of it is actually part of a way or not - it must be checked after all)

- on the other hand, collecting all node ids of all ways within the bounding box is naturally bound to the size of the bounding box. I.e. it cannot grow into infinity but remains the same more or less. And, a `Set` is only created once here, not for every single node.

PR #5040, opened 2023-05-22
Fix crash when adding node after modifying a node in the same way

westnordost:
supersedes #5026 . See that ticket for a description.

---

I hope I thought of everything (this time).


Helium314:
That's definitely more efficient than what I had thought of.

> I hope I thought of everything (this time).

I think it should be fine now

Helium314 (approved):

PR #5060, opened 2023-06-09
remove geometry filtering from sidewalk and cycleway quest

### LINKED ISSUE #5000: Reduce filtering of ways for sidewalk quest
As mentioned in #4915, there are quite a few cases where the sidewalk quest is not asked for ways for which I would expect it to be asked, because of too strict filtering. I understood that there is a filter that ensures that the sidewalk quest is not asked for ways that are believed to have separately mapped sidewalks already by looking for nearby footways that are running in the same direction as the way. Examples mentioned include way https://www.openstreetmap.org/way/862062530 that is not quested because of https://www.openstreetmap.org/way/1056791202 that is of insignificant length, or way https://www.openstreetmap.org/way/23147663 that is not quested because of https://www.openstreetmap.org/way/499159632 that is a path that runs in continuation of it, and https://www.openstreetmap.org/way/673415848 that has a separately mapped sidewalk only on one side (if there is a sidewalk on the other side is unknown).

I think this filter should be reviewed to see:

1. if it's necessary at all. Now that `sidewalk:*=separate` can easily be added by the app, it's not such a waste of time any more. It's actually a good thing to add it, because (as I understood) pedestrian routing apps would like to have info on a separately mapped sidewalk and the road it is related to (I remember reading somewhere the suggestion to add the road name to the sidewalk too, to clarify they are related). https://wiki.openstreetmap.org/wiki/Sidewalks#Sidewalk_as_separate_way states that "the associated street/highway should also be tagged with `sidewalk=separate`" so shouldn't we help doing that with SC?
2. if it can be relaxed or refined, without too much effort, so that fewer roads are excluded from sidewalk questing. One could for instance only take footways tagged with `footway=sidewalk` into account instead of all paths, and demand that separetely mapped sidewalks are present on both sides of the road before excluding it. Determining if a (series of) sidewalk(s) is truly running parallel to a (series of) road(s) for a significant length is probably quite difficult?

westnordost:
fixes #5000


Helium314:
SC doesn't display whether a nearby way is accessible for foot or bicycle. I think highlighting nearby ways would be a useful addition to avoid confusion e.g. whether the nearby way is a cycleway.

Proper highlighting would then require `MainFragment.showHighlightedElements` to be adjusted to use at least the bbox of the `element`.
While this could potentially be really slow (for very long roads), I've been using this in SCEE for ~half a year without noticing issues.

westnordost:
MainFragment.showHighlightedElements doesn't show at least the bbox of the element?

Helium314:
No, it's the center and surrounding 20 or 30 meters

westnordost:
> No, it's the center and surrounding 20 or 30 meters

Oh. I guess it should be the bbox enlarged by 20 or 30 meters instead. Since the cache, this should not be so devastating to performance anymore, I guess.

Helium314:
> Oh. I guess it should be the bbox enlarged by 20 or 30 meters instead. Since the cache, this should not be so devastating to performance anymore, I guess.

That's also my impression. Even in bad cases where it takes a few seconds to get the data, afterwards everything nearby is cached.

For the highlighted elements filter something like
```
        ways with
          highway ~ path|footway
          or (highway ~ cycleway|bridleway and foot ~ yes|designated)
          and foot !~ no|private
          and access !~ no|private
```
should be good (similar for cycleways)

westnordost:
Thank you for the filter

PR #5064, opened 2023-06-11
Enable cycleways/tactile paving quests in Israel

elad661:
Tactile paving is pretty common in Israel, having been mandatory in new developments and street rennovations for at least 5 years, if not more. Bike paths are also common in the flat areas of Central Israel and the bike path network is slowly being expanded.

I think enabling these quests make sense as it will help gather this kind of data. There's no reason to outright disable these quests in Israel.


westnordost:
Quests should only be shown when there is an a chance that the answer will not be the same (in this case, "no") for the vast majority of cases. I think previously, we said something like that in at least 10% of cases, the answer should differ. Would you say that this is fulfilled for both?
Otherwise, you know that there is the cycleway overlay too, right? This you can use everywhere.

westnordost:
No reply, closing.

PR #5071, opened 2023-06-14
Don't set newer tile timestamp when invalidating

### LINKED ISSUE #5068: Area shown as downloaded on map, but no data on device
When invalidating downloaded tiles (e.g. on version upgrade), they are set to an age of 12 hours, to have auto-download starting on next occasion.
This has a negative effect on the (new) functionality of showing non-downloaded areas on the map: After an upgrade, the tiles will keep showing as dowloaded for two weeks (minus 12 hours), even if the map data has already been deleted.

This is a minor issue, but it partially defeats the purpose for showing the downloaded area.
My suggestion is not updating the timestamp of a tile if it's already older than the date to be set.

Helium314:
fixes #5068

Tested by logging tile timestamps before and after invalidating: works as intended.

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/data/download/tiles/DownloadedTilesDaoTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/data/download/tiles/DownloadedTilesDaoTest.kt
@@ -67,7 +67,7 @@ class DownloadedTilesDaoTest : ApplicationDbTestCase() {
 
     @Test fun updateTime() {
         dao.put(r(0, 0, 0, 1))
-        dao.updateTime(TilePos(0, 0), 0)
+        dao.updateTimeIfNewer(TilePos(0, 0), 0)
         assertEquals(
             listOf(TilePos(0, 1)),
             dao.getAll(1)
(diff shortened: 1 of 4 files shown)


app/src/androidTest/java/de/westnordost/streetcomplete/data/download/tiles/DownloadedTilesDaoTest.kt
```diff
@@ -67,7 +67,7 @@ class DownloadedTilesDaoTest : ApplicationDbTestCase() {
 
     @Test fun updateTime() {
```
westnordost, line 68: Should update the names of the test functions too.

app/src/main/java/de/westnordost/streetcomplete/data/download/tiles/DownloadedTilesDao.kt
```diff
@@ -31,16 +31,20 @@ class DownloadedTilesDao(private val db: Database) {
...
-            where = "$X = ? AND $Y = ?",
-            args = arrayOf(tile.x.toString(), tile.y.toString()),
+            where = "$X = ? AND $Y = ? AND $DATE > ?",
+            args = arrayOf(tile.x.toString(), tile.y.toString(), time.toString()),
         )
     }
 
-    fun updateAllTimes(time: Long) {
-        db.update(NAME, values = listOf(DATE to time))
+    fun updateAllTimesIfNewer(time: Long) {
```
westnordost, line 42: Sounds odd. Wouldn't `updateAllTimesNewerThan` be more descriptive? (Same for the other function, `updateTimeIfNewerThan` or something.)

westnordost (changes requested):
Since the function does something different now, it would make sense to add two tests that test the new behavior: I.e. that only the times are updated that are older than X.

Helium314:
I renamed the functions and added a test.
But for whatever reason, I can't run the android tests on my device, so I can't actually check whether the test succeeds...

westnordost (approved):

PR #5073, opened 2023-06-14
Don't delete old nodes that are part of a way

### LINKED ISSUE #5066: Cleanup of old data may result in incomplete ways
related to #4980

I received a crash report for SCEE about a NullPointerException in https://github.com/streetcomplete/StreetComplete/blob/7dbfc8867e03b06bc361777a4b4ea53322b6fd5d/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt#L415
which can only occur if not all nodes are available.
It's SCEE, but I'm quite sure the same crash can happen in StreetComplete too, so it should be fixed here.

My suspicion is that old nodes got cleaned (more than 14 days old), but the way and at least one node was still there. As far as I understand, this can happen when both were edited and uploaded. Then the edited elements will have an updated `LAST_SYNC` timestamp after upload, but the non-edited nodes will keep the old timestamp.

Details leading to this idea were provided by the user:
* Before the crash, the overlay was not shown, despite being active (i.e. no highlighted data)
* The crash came up repeatedly until the user downloaded the same area again, so I think the issue is not connected to the map data cache
* The user downloads data manually

**How to Reproduce**
I did __not__ try to reproduce it, but I assume it could be reproduced as follows:
* download an area
* wait 7 days
* edit a way and a node of that way and upload the edits
* wait more than 7, but less than 14 days
* edit the way again
* do something that calls `getMapDataWithGeometry` with the edited node inside the bbox

Helium314:
should fix #5066

This changes `deleteOlderThan` to first delete ways, and then deletes only nodes that are not part of any way in the database.

_Not_ yet tested, and not unit tests.
This can be solved in a bunch of different ways and I don't want to work out every detail before knowing the basic approach is ok.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/ElementDao.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/ElementDao.kt
@@ -85,15 +85,6 @@ class ElementDao(
         wayDao.clear()
         nodeDao.clear()
     }
-
-    fun getIdsOlderThan(timestamp: Long, limit: Int? = null): List<ElementKey> {
-        val result = mutableListOf<ElementKey>()
-        // get relations first, then ways, then nodes because relations depend on ways depend on nodes.
-        result.addAll(relationDao.getIdsOlderThan(timestamp, limit?.minus(result.size)).map { ElementKey(RELATION, it) })
-        result.addAll(wayDao.getIdsOlderThan(timestamp, limit?.minus(result.size)).map { ElementKey(WAY, it) })
-        result.addAll(nodeDao.getIdsOlderThan(timestamp, limit?.minus(result.size)).map { ElementKey(NODE, it) })
-        return result
-    }
 }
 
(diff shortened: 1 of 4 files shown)


westnordost (commented):
Looks good to me. 

Except maybe deleting the relations and ways before updating the cache may be a problem?

It looks like `ElementDao.getIdsOlderThan` can be deleted then.

Helium314:
> Except maybe deleting the relations and ways before updating the cache may be a problem?

I don't think so, but there's no problem in making it safe and deleting them from cache before database.

westnordost:
Ready for review?

Helium314:
I just tested it with some detailed logging and looks like it's fine: all nodes of a modified building remain in database, while all the nearby old data was deleted.

Ways and relations are now deleted from cache before deleting from database, `ElementDao.getIdsOlderThan` is removed and a test is added.

westnordost (approved):

PR #5090, opened 2023-06-24
Fix crash when reverting CreateNodeAction

Helium314:
Fixes a crash when creating a node, uploading and reverting immediately.

The reason for the crash is that `getIdProvider(edit.id)` will return an empty `ElementIdProvider` after `delete(edit)` instead of a provider containing the node to be deleted.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsController.kt
@@ -109,10 +109,11 @@ class ElementEditsController(
         if (edit.isSynced) {
             val action = edit.action
             if (action !is IsActionRevertable) return false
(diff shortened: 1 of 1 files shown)


westnordost:
Cool, that you find all those issues! And fix them! :-)

westnordost:
Could it be that this has been an issue since the create node action (i.e. since the shop overlay was introduced) existed? And could it be that it **always** crashed, even if not 

> reverting immediately.

?

westnordost (approved):

Helium314:
I guess I find them because SCEE users tend to use the app differently, especially creating more nodes, so I end up with the crash reports.

> Could it be that this has been an issue since the create node action (i.e. since the shop overlay was introduced) existed?

yes, I think so too

> And could it be that it always crashed

I _think_ I didn't have a crash when uploading (using some debug-fake upload in SCEE), restarting and then reverting.
But looking at the code now I'm not even sure whether this is actually possible.

PR #5104, opened 2023-06-28
add crossing kerbs quest

### LINKED ISSUE #3239: Kerbs for existing crossings
In the most recent beta 34.0 #2999 is fixed. That means crossing nodes are added for non-crossing type ways meeting a highway. The quests which trigger then involve saying whether the crossing has an island, what type of kerbs it has, what type of crossing it is and whether it has tactile paving.

Three of those four quests also trigger for crossing nodes on crossing ways where they meet highway ways. However the quest to set a kerb type on this crossing node never triggers. To me this should trigger for all highway=crossing nodes, regardless of whether they are for a crossing type way meeting a highway type way or for a non-crossing type way meeting a highway.

westnordost:
fixes #3239


app/src/main/res/values/strings.xml
```diff
@@ -927,6 +927,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_construction_steps_title">Are these steps completed?</string>
     <string name="quest_construction_generic_title">Is this path completed?</string>
 
+    <string name="quest_crossing_kerb_height_title">What’s the height of the curbs of this crossing?</string>
```
FloEdelmann, line 930: Maybe this is a better wording? @arrival-spring 

```suggestion
    <string name="quest_crossing_kerb_height_title">What’s the height of the curbs at this crossing?</string>
```

arrival-spring: "at" is better than "of", but because curbs is plural, it should be "What are the heights of the curbs at this crossing?"

westnordost: But it is asking for a single height, i.e. the user should be not sure what to answer if the height**s** differ (-> leave note)

arrival-spring: Yes, I see. Flo's suggestion is fine, height is singular, so "what's" matches

PR #5107, opened 2023-06-29
Support Not operator in filter expressions

riQQ:
- Support Not operator in filter expressions
- Add element filter tests

I'm looking for some feedback if this is going in the right direction and if there is anything missing. Are more unit tests required?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/BooleanExpression.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/BooleanExpression.kt
@@ -12,7 +12,7 @@ abstract class Chain<I : Matcher<T>, T> : BooleanExpression<I, T>() {
 
     val children: List<BooleanExpression<I, T>> get() = nodes.toList()
 
-    fun addChild(child: BooleanExpression<I, T>) {
+    open fun addChild(child: BooleanExpression<I, T>) {
         child.parent = this
         nodes.add(child)
     }
@@ -55,7 +55,7 @@ abstract class Chain<I : Matcher<T>, T> : BooleanExpression<I, T>() {
         val it = nodes.listIterator()
         while (it.hasNext()) {
             val child = it.next() as? Chain ?: continue
-            if (child.nodes.size == 1) {
+            if (child.nodes.size == 1 && child !is Not<I, T>) {
                 replaceChildAt(it, child.nodes.first())
                 it.previous() // = the just replaced node will be checked again
             } else {
@@ -69,6 +69,9 @@ abstract class Chain<I : Matcher<T>, T> : BooleanExpression<I, T>() {
         val it = nodes.listIterator()
         while (it.hasNext()) {
             val child = it.next() as? Chain ?: continue
+            if (child is Not<I, T>) {
+                continue;
+            }
             child.mergeNodesWithSameOperator()
 
             // merge two successive nodes of same type
@@ -92,6 +95,28 @@ class AllOf<I : Matcher<T>, T> : Chain<I, T>() {
 class AnyOf<I : Matcher<T>, T> : Chain<I, T>() {
     override fun matches(obj: T) = nodes.any { it.matches(obj) }
     override fun toString() = nodes.joinToString(" or ") { "$it" }
+}
+
+class Not<I : Matcher<T>, T> : Chain<I, T>() {
+    override fun addChild(child: BooleanExpression<I, T>) {
+        if (nodes.size > 0)
+        {
+            throw IllegalStateException("Adding a second child to '!' (NOT) operator is not allowed")
+        }
+        super.addChild(child)
+    }
+
+    override fun matches(obj: T) = !nodes.first().matches(obj)
+
+    override fun toString() : String {
+        val childString = when (val firstNode = nodes.firstOrNull()) {
+            is Chain -> "($firstNode)"
+            null -> ""
+            else -> "($firstNode)"
+        }
+
+        return "!$childString"
+    }
 }
 
 interface Matcher<in T> {
--- a/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/BooleanExpression.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/BooleanExpression.kt
@@ -99,24 +99,13 @@ class AnyOf<I : Matcher<T>, T> : Chain<I, T>() {
 
 class Not<I : Matcher<T>, T> : Chain<I, T>() {
     override fun addChild(child: BooleanExpression<I, T>) {
-        if (nodes.size > 0)
-        {
-            throw IllegalStateException("Adding a second child to '!' (NOT) operator is not allowed")
-        }
+        check(nodes.isEmpty()) { "Adding a second child to '!' (NOT) operator is not allowed" }
         super.addChild(child)
     }
 
     override fun matches(obj: T) = !nodes.first().matches(obj)
 
-    override fun toString() : String {
-        val childString = when (val firstNode = nodes.firstOrNull()) {
-            is Chain -> "($firstNode)"
-            null -> ""
-            else -> "($firstNode)"
-        }
-
-        return "!$childString"
-    }
+    override fun toString() = "!(${nodes.firstOrNull() ?: ""})"
 }
 
 interface Matcher<in T> {
--- a/app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/ElementFiltersParserTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/ElementFiltersParserTest.kt
@@ -1,5 +1,7 @@
 package de.westnordost.streetcomplete.data.elementfilter
 
+import de.westnordost.streetcomplete.osm.MAXSPEED_TYPE_KEYS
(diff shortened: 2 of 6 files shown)


westnordost:
Hmm, interesting.

I remember having have attempted this too in the past but gave up for some reason, maybe it becoming too complex. Incidentally, did you choose `not` instead of `!` for this reason? I definitely chose `!` as operator because I also find it more consistent to write `and !(key1 or key2)` than `and not (key1 or key2)`.

I also see that you do not allow more than one item on `Not`. But why, then, is it a `Chain`? Seems inconsistent to me.

Anyway, I find it not very useful to support tag filters like `not key != value`. Rather, I'd find it more useful if tag filters like `!(condition1 and condition2 or condition3)` were supported.

riQQ:
> Incidentally, did you choose not instead of ! for this reason.

Yes, I chose `not` for that reason. Otherwise, how would you distinguish which not is meant with e.g. `!key`?


> I also see that you do not allow more than one item on Not. But why, then, is it a Chain? Seems inconsistent to me.

Yeah, I'm not that happy about this. I tried it without `Not` being a chain, but it required a lot more changes. Nevertheless I can implement it the other way if you want.

> Anyway, I find it not very useful to support tag filters like not `key != value`. Rather, I'd find it more useful if tag filters like `!(condition1 and condition2 or condition3)` were supported.

Have you seen the tests? This is supported.

westnordost:
> Yes, I chose not for that reason. Otherwise, how would you distinguish which not is meant with e.g. !key?

Because I guess it would always be followed by a `(`.

riQQ:
> Because I guess it would always be followed by a `(`.

That would mean you have to know that a `(` is always necessary when writing filters for StreetComplete even if isn't necessary in standard boolean logic. That's why I don't like mandating braces.

Have you seen the other quotes with my replies in my previous comment? Any input there?

westnordost:
> That would mean you have to know that a ( is always necessary when writing filters for StreetComplete even if isn't necessary in standard boolean logic. That's why I don't like mandating braces.

What do you mean?

Where would it not be necessary? If you want to negate a sub-expression, you'd also have to add a `!(...)`, except for negating single leafs: `not highway != residential` etc.  - but that doesn't look like it would make a lot of sense. Can anything useful be done with this, considering that the above expression is the same as `highway = residential`? I mean, is there a use case for negating a single leaf?

westnordost:
> Yeah, I'm not that happy about this. I tried it without Not being a chain, but it required a lot more changes. Nevertheless I can implement it the other way if you want.

Hmm, it certainly is not clean. Inheriting from `Chain` also inherits the whole flattening of the tree logic. Though, looking shortly at the code right now, I am not sure how to well fit it into the current structure. Do you have an idea?

Maybe the usage of `Chain` could be replaced with a sealed interface which can be one of `AnyOf`, `AllOf`, `Not` ( and `BracketHelper`, I guess)

riQQ:
> > That would mean you have to know that a ( is always necessary when writing filters for StreetComplete even if isn't necessary in standard boolean logic. That's why I don't like mandating braces.
> 
> What do you mean?
> 
> Where would it not be necessary? If you want to negate a sub-expression, you'd also have to add a `!(...)`, except for negating single leafs: `not highway != residential` etc. - but that doesn't look like it would make a lot of sense. Can anything useful be done with this, considering that the above expression is the same as `highway = residential`? I mean, is there a use case for negating a single leaf?

Yes, I'm talking about single leafs not requiring braces. I don't have the complete picture on everything that is possible with the StreetComplete filtering syntax. If you can't think of anything then there's probably no use case.

riQQ:
> > Yeah, I'm not that happy about this. I tried it without Not being a chain, but it required a lot more changes. Nevertheless I can implement it the other way if you want.
> 
> Hmm, it certainly is not clean. Inheriting from `Chain` also inherits the whole flattening of the tree logic. Though, looking shortly at the code right now, I am not sure how to well fit it into the current structure. Do you have an idea?
> 
> Maybe the usage of `Chain` could be replaced with a sealed interface which can be one of `AnyOf`, `AllOf`, `Not` ( and `BracketHelper`, I guess)

The problem is that `Not` also needs to run `Flatten` to get rid of `BracketHelper`s below it. So there's not much left over from `Chain` that `Not` doesn't need.

westnordost:
> Yes, I'm talking about single leafs not requiring braces. I don't have the complete picture on everything that is possible with the StreetComplete filtering syntax. If you can't think of anything then there's probably no use case.

I do think that. Have a look at the cheatsheet in ElementFilterExpression.

> The problem is that Not also needs to run Flatten to get rid of BracketHelpers below it. So there's not much left over from Chain that Not doesn't need.

Ah right, okay. Then, let's just keep it a chain.

westnordost (commented):
I think the original implementation is better than the other approach d2fefa6.

I'd strongly prefer if the not operator was `!` (and only worked in conjunction with `!(...)`) because as explained already
- `not x=y` (etc) does not make sense, the only use case for this is to negate a chain anyway
- I find it more consistent not to have two different operators for negation

Is there speaking anything against using the `!` operator instead?

riQQ:
> Is there speaking anything against using the `!` operator instead?

I first thought it would be more difficult to implement, but I found an easy way.

Two more questions:
- Should spaces be allowed between `!` and `(`?
- Should `!(x = y)` be allowed?

westnordost:
I think spaces should be allowed.

I don't understand, why should `!(x = y)` not be allowed? How would you even prevent that? Why that in particular, what about `!(x < y)`?

riQQ:
> I think spaces should be allowed.

They are allowed now.

> I don't understand, why should `!(x = y)` not be allowed? How would you even prevent that? Why that in particular, what about `!(x < y)` etc.?

You said you didn't want `! <leaf-expression>` to be supported and I thought that meant you also don't want `!(<leaf-expression>)` to be support. Just a misunderstanding.


-------

I addressed all of your feedback and this is ready for review.

app/src/test/java/de/westnordost/streetcomplete/data/elementfilter/ElementFiltersParserTest.kt
```diff
@@ -497,6 +501,134 @@ class ElementFiltersParserTest {
...
+          and (
+            !maxspeed
+            or maxspeed > 20
+            or $notInZone30OrLessNew
+          )
+          and surface !~ ${ANYTHING_UNPAVED.joinToString("|")}
+    """.toElementFilterExpression()
+        return expression
+    }
+
```
westnordost, line 631: This looks out of place for me. Why would this test be necessary when you already test every variant? If you really think that this test makes sense, it should go into the cycleway quest.

Otherwise, concise and easy to read tests!

app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/BooleanExpression.kt
```diff
@@ -94,6 +97,28 @@ class AnyOf<I : Matcher<T>, T> : Chain<I, T>() {
     override fun toString() = nodes.joinToString(" or ") { "$it" }
 }
 
+class Not<I : Matcher<T>, T> : Chain<I, T>() {
+    override fun addChild(child: BooleanExpression<I, T>) {
+        if (nodes.size > 0)
+        {
+            throw IllegalStateException("Adding a second child to '!' (NOT) operator is not allowed")
+        }
```
westnordost, line 105: The Kotlin way is 

```suggestion
        check(nodes.isEmpty()) { "Adding a second child to '!' (NOT) operator is not allowed" }
```

app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/BooleanExpression.kt
```diff
@@ -94,6 +97,28 @@ class AnyOf<I : Matcher<T>, T> : Chain<I, T>() {
...
+        val childString = when (val firstNode = nodes.firstOrNull()) {
+            is Chain -> "($firstNode)"
+            null -> ""
+            else -> "($firstNode)"
+        }
+
+        return "!$childString"
+    }
+}
+
```
westnordost, line 121: Looks like this can be simplified to

```suggestion
    override fun toString() = "!(${nodes.firstOrNull() ?: ""})"
```

westnordost (commented):
Wow, I wouldn't have thought that the final implementation required so little code.

westnordost:
@bryceco FYI this will soon be merged and as a result, element filters in StreetComplete may include this syntax in the future. The syntax is `! ( ... )`, i.e. negating whatever is in the brackets.

bryceco:
@westnordost  Thanks for the notice. Although I started down the path of parsing SC filters I ultimately abandoned that approach. This and future breaking changes won't affect me.

riQQ:
> Wow, I wouldn't have thought that the final implementation required so little code.

It didn't require much code, but it took some time to understand the logic.

app/src/test/java/de/westnordost/streetcomplete/quests/cycleway/AddCyclewayTest.kt
```diff
@@ -22,6 +22,9 @@ import org.junit.Test
 import org.mockito.ArgumentMatchers.anyDouble
 import java.util.concurrent.FutureTask
 
+private const val HIGHWAY = "highway"
+private const val RESIDENTIAL = "residential"
+
```
westnordost, line 27:
```suggestion
```

westnordost (approved):

PR #5147, opened 2023-07-15
Don't animate global rank changes to rank -1

### LINKED ISSUE #5141: No -1 in the world after log-in
image

I changed the username, password and email address of my osm account and after logging in with the new info the app claims that I'm number -1 in the world (which is a pretty bad place to be imo)

**How to Reproduce**
Change your OSM profile so that you have to renew your StreetComplete log-in.

Log-in to street complete

**Expected Behavior**
I am placed where I should be

**Versions affected**
53.2

tapetis:
Should fix #5141. See https://github.com/streetcomplete/StreetComplete/issues/5141#issuecomment-1636712069 for an explanation.


westnordost:
Thanks!

PR #5153, opened 2023-07-18
make bboxes created from TilesRect snap to a certain precision

westnordost:
Supersedes #5091

cc @Helium314 : How does this look to you? You obviously have been testing with a larger set of test data. Would you test this out with your data? Does it make sense to add this as unit tests?

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/download/tiles/TilesRect.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/download/tiles/TilesRect.kt
@@ -7,6 +7,7 @@ import kotlinx.serialization.Serializable
 import kotlin.math.PI
 import kotlin.math.asinh
 import kotlin.math.atan
+import kotlin.math.pow
 import kotlin.math.sinh
 import kotlin.math.tan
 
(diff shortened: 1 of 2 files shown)


Helium314:
For the tests I don't use a fixed set of points, I simple change `points` to a mutable list and do
```kotlin
        repeat(1000000) {
            points.add(p(Random.nextDouble(-85.05, 85.05), Random.nextDouble(-180.0, 180.0)))
        }
```
latitude is in limited range as in https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#X_and_Y

`TilesRectTest` fails for many `LatLon`s, e.g. `LatLon(latitude=67.21044, longitude=-42.19332)`

westnordost:
> TilesRectTest fails for many LatLons, e.g. LatLon(latitude=67.21044, longitude=-42.19332)

Uff...

With this code...

```kotlin
    @Test
    fun `convert bbox to tiles rect and back results in same bbox`() {
        val tries = 100_000_000
        var fail = 0
        for (i in 0..tries) {
            val p = p(Random.nextDouble(-85.05, 85.05), Random.nextDouble(-180.0, 180.0))
            val tile = p.enclosingTilePos(15)
            val bbox = tile.asBoundingBox(15)
            val r = bbox.enclosingTilesRect(15)
            val bbox2 = r.asBoundingBox(15)
            if (bbox != bbox2) fail++
        }
        assertTrue("Failed ${100.0*fail/tries}%", false)
    }
```

I get...

- ca. 8% for master

- ca. 7% for master with the `PI` moved just after `180` at `tile2lat` 

- ca. 0.3% for #5091

- ca. 3% for #5153

- ca. 3% for #5091 + #5153

---

This... really puzzles me. Your PR only changes the values of the calculation by a value close to 1e-15 while this PR cuts off basically half of the digits...

Helium314:
> This... really puzzles me

Such weird thing when diving into this a bit more was the reason why I wanted to close #5091.
Improvement for cache preformance just doesn't justify the apparently necessary amount of time to get it right...

westnordost:
I checked which bounds exactly cross over to the next cell. It's the top (=max latitude)

image

So, then it makes sense that this PR does not reduce the problem that much because it only rounds up or down at the seven digits. But when the result of the calculation is e.g. already slightly above `x.xxxxxxx5` while it shouldn't be (due to error introduced due to rounding imprecision which even goes up all the way to digit 7, apparently ) this PR does not solve the problem.

Anyway, recognizing that the latitude calculation in code de-facto seems to be off by a tiny amount (in ~8% of cases), I'll just apply your PR then.

PR #5162, opened 2023-07-26
Ask AddCrossing for sidewalk and crossing ways

### LINKED ISSUE #5160: Ask AddCrossing for sidewalks
I recently found a place where _AddCrossing_ unexpectedly wasn't asked, and then stumbled upon https://github.com/streetcomplete/StreetComplete/blob/321ee53cd4502b1f24570818627266d4ac6f9300/app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossing.kt#L33-L36

Would it be ok to just hide the `crossing=no` answer instead of not asking at all? Or maybe open a dialog asking to open a note and describe the situation?
This would enable the _AddCrossing_ and all the follow-up quests for many more ways.

Helium314:
fixes #5160

Hides the prohibited answer for `footway=sidewalk` and `footway=crossing`, so the user has to leave a note in that case.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingForm.kt
@@ -11,37 +11,38 @@ import org.koin.android.ext.android.inject
 class AddCrossingForm : AListQuestForm<CrossingAnswer>() {
     private val mapDataSource: MapDataWithEditsSource by inject()
 
-    override val items get() = listOf(
+    override val items = listOf(
         TextItem(YES, R.string.quest_crossing_yes),
         TextItem(NO, R.string.quest_crossing_no),
-        TextItem(PROHIBITED, R.string.quest_crossing_prohibited)
+        TextItem(PROHIBITED, R.string.quest_crossing_prohibited),
     )
 
-    /* PROHIBITED is neither possible for sidewalks nor crossings (=separately mapped sidewalk infrastructure)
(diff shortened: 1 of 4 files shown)


westnordost:
So atm I would prefer a solution as outlined here: https://github.com/streetcomplete/StreetComplete/issues/5160#issuecomment-1660047197

... because of the remarks in the comments before that.

Helium314:
The text may need adjustment, otherwise it shoud be ok.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingForm.kt
```diff
@@ -1,15 +1,47 @@
...
+        TextItem(PROHIBITED, R.string.quest_crossing_prohibited)
     )
+
+    /* PROHIBITED is neither possible for sidewalks nor crossings (=separately mapped sidewalk infrastructure)
+    *  because a "no" answer would require to also delete/adapt the crossing ways, rather than just
+    *  tagging crossing=no on the vertex.
+    *  This situation needs to be solved in a different editor, so we ask the user to leave a note.
+    *  See https://github.com/streetcomplete/StreetComplete/pull/2999#discussion_r681516203
+    *  and https://github.com/streetcomplete/StreetComplete/issues/5160 */
+    override fun isFormComplete(): Boolean {
```
westnordost, line 26: You made this function have side-effects. The code should go into onClickOk. But dont worry, I will do this then.

westnordost (approved):

Issue #5182, opened 2023-08-08
Public transport platform quests aren't available if platform is a relation


burrscurr:
Public transport platform quests (is there a bin, bench, shelter at this platform and similar) are not displayed for `public_transport=platform`s that are relations (typically `type=multipolygon`). Island platforms are often mapped as multipolygons, for instance, because the waiting area has an elevators, stairs to a passenger tunnel or similar within its perimeter. (I therefore consider this a bug rather than a feature request or a change).

This affects the following quests:

 - `AddBenchStatusOnBusStop.kt`
 - `AddBusStopName.kt`
 - `AddBinStatusOnBusStop.kt`
 - `AddTactilePavingBusStop.kt`
 - `AddBusStopLit.kt`
 - `AddBusStopShelter.kt`
 - maybe `AddBusStopRef.kt` (this one just filters for nodes, maybe it also should for ways and even relations?)

I drafted a pull request extending the quests on relations: #5183.

**How to Reproduce**

1. Consider [this relation](https://www.openstreetmap.org/relation/10362716) which is a railway platform in the main station of a fairly large city. Note that this relation was last edited 2 years ago, and does not have `lit`, `bench`, `bin` or `shelter` tags.
2. Open the location of the relation in SC and download the area's data.
3. There are no quests displayed that refer to the platform relation.

**Expected Behavior**
SC should display quests related to platforms even if those platforms are multipolygon relations.

**Versions affected**

I checked this with v53.1.

Issue #5132, opened 2023-07-08
Platform vs stop quest (name/shelter/waste basket) in tram/bus stops


AgaFra:
It's very tricky to respond in the correct way in quests as "Is there a shelter at this stop? Train Platform" or "What's the name of this stop? Train platform" for trams or buses. The question suggests that I should write the name of the stop. But the quest is about platform... All is clear in terms of rail, or bigger interchange. But I’m lost when there is only one stop and one platform - I don't know if I should write name under stop quest or under platform quest. And there is no option "platform doesn't have separate name". In case of shelter (or waste basket) - are those quests relate for this same object? I believe such situation happens when stop is a node and platform as [public_transport]=[platform].

mnalis:
> But I’m lost when there is only one stop and one platform - I don't know if I should write name under stop quest or under platform quest

Now _I_ am confused: you have a platform for waiting/boarding the train/bus which has **one** name on its sign, but a stop position for same train/bus at that location has **different** name? Can you give picture/example how that looks like? It makes no sense to me that it should exist; over here a bus/train stop is just a point alongside waiting platform where vehicle will stop - it is basically just a "part of/point alongside" the passenger platform  

> And there is no option "platform doesn't have separate name". In case of shelter (or waste basket) - are those quests relate for this same object?

About being asked for duplicate information, that is likely because data is duplicate in OSM already (so you should also enter the same information twice), see related discussions:
- https://github.com/streetcomplete/StreetComplete/issues/4690#issuecomment-1356919286
- https://github.com/streetcomplete/StreetComplete/issues/4341#issuecomment-1358927732 and below

AgaFra:
Thank you for the references to the discussions. Indeed, my problem is caused by double tagging. It asked me one time for a node, and another time for a way.

> > But I’m lost when there is only one stop and one platform - I don't know if I should write name under stop quest or under platform quest
> 
> Now _I_ am confused: you have a platform for waiting/boarding the train/bus which has **one** name on its sign, but a stop position for same train/bus at that location has **different** name? Can you give picture/example how that looks like? It makes no sense to me that it should exist; over here a bus/train stop is just a point alongside waiting platform where vehicle will stop - it is basically just a "part of/point alongside" the passenger platform (or I misunderstood that part)

Platform doesn't have a name at all, only a bus stop has. So I didn't know if I should put the name under the node or way quest.  
[...]

PR #5183, opened 2023-08-08
Enable platform quests on relations

burrscurr:
This commit also considers `public_transport=platform`s that are relations (e.g. multipolygons). See #5182 for more details.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
@@ -12,7 +12,7 @@ import de.westnordost.streetcomplete.util.ktx.toYesNo
 class AddBenchStatusOnBusStop : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
-        nodes, ways with
+        nodes, ways, relations with
         (
           public_transport = platform
           or (highway = bus_stop and public_transport != stop_position)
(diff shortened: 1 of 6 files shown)


mnalis:
Have you tested this pull request? 
(i.e. have you tested *in practice* that it does fix the problem you note, without breaking other cases)?

burrscurr:
I did not test the changes so far - this was more intendet to outline which quests are (probably) affected and the changes that I think solve the issue.

Should have made that more clear, though. I will change this to a draft.

burrscurr:
I now have checked that the bench, bin, is lit, shelter and name quests are now also displayed for relation platforms and are answerable. The quests also (still) come up if the platform is a node or way. (Note: I did not change anything on the bus stop ref quest, because I am not sure whether that one intentionally just searches for nodes.) I'm not sure what else could be affected by the change, so if anyone with more insight into SC internals can point me to other functionality that  requires testing, I'd be very grateful.

mnalis:
> I now have checked that the bench, bin, is lit, shelter and name quests are now also displayed for relation platforms and are answerable. The quests also (still) come up if the platform is a node or way

That seems to be about it. StreetComplete Quests are self-contained, so no other quest should behave differently except the quests you changed. As you tested them they work OK for both old and new behaviour, it looks good to me. Feel free to mark it as `ready for review`.

> (Note: I did not change anything on the bus stop ref quest, because I am not sure whether that one intentionally just searches for nodes.)

Good question. I recall some issues with problems when some transport information is mapped both as node and way (see e.g. https://github.com/streetcomplete/StreetComplete/issues/5132 and linked issues), so I'm not sure if there is a reason for that.

But that one can be always be added as separate PR if that turns out to be safe to do (and if not safe, another PR documenting in the code why it only looks for nodes would be welcome.)

mnalis (approved):

westnordost:
Hmm, I think nothing speaks against this...

The bus stop quests especially in regard to railway platforms have been subject of various discussions in the past, but I guess since railway platforms are now also included in the quest, it makes sense to also look for relations (i.e. railway platforms with a hole to accommodate stairs).

Looking at git blame, other than me, @FloEdelmann has some lines of code there, so I am just going to ping him if he has any concerns.

FloEdelmann:
I don't think anything speaks against this :)

PR #5184, opened 2023-08-09
Downloading instead of scanning

arrival-spring:
The menu item was renamed from "scan for quests here" to "download data here", so the error message should match that.


app/src/main/res/values/strings.xml
```diff
@@ -267,10 +267,10 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
 
     <string name="no_gps_no_quests">"To scan for quests around you automatically, turn on location"</string>
```
arrival-spring, line 268: Should this also be changed to "download data"?

westnordost: Hmm, I guess it makes sense to have consistent naming for the action everywhere.

arrival-spring:
Found all of the rest of the places where scanning is mentioned and changed to downloading as well

PR #5193, opened 2023-08-14
Use kotlin.test function to check for exception in test

riQQ:
Use kotlin.test function `assertFailsWith<T>` instead of self-made function to check for exception in tests.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -120,6 +120,7 @@ dependencies {
     androidTestImplementation("androidx.test:rules:1.5.0")
     androidTestImplementation("org.mockito:mockito-android:$mockitoVersion")
     androidTestImplementation("org.assertj:assertj-core:3.23.1")
+    androidTestImplementation(kotlin("test"))
 
     // dependency injection
     implementation("io.insert-koin:koin-android-compat:3.4.1")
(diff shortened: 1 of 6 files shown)


westnordost:
Oh, good! Is this the only place where this is used? I thought this construct is used in more places. For consistency, it should all be replaced in one go.

riQQ:
I only found one more file in the tests folder (see latest update).

Other than that, some classes use `@Test(expected = <ExceptionClass>::class)`. Should this also be replaced?

Example:
https://github.com/streetcomplete/StreetComplete/blob/3bca19fda5e7a11d4b5c8048b813d6cc96f3dfa7/app/src/test/java/de/westnordost/streetcomplete/data/ObjectTypeRegistryTest.kt#L8-L11

westnordost:
> Other than that, some classes use @Test(expected = <ExceptionClass>::class). Should this also be replaced?

Oh yeah, I'd like that! Not required for this PR to be merged, though.
Always didn't like that API. I think they also corrected that awkward API in JUnit 5. (But if we migrate, we rather migrate fully to `kotlin.test` rather than Junit 5)

riQQ:
> There is `assertThrows` in some test files, maybe this could be replaced too

This is done.

> > Other than that, some classes use `@Test(expected = <ExceptionClass>::class)`. Should this also be replaced?
> 
> Oh yeah, I'd like that! Not required for this PR to be merged, though. Never liked that API. I think they also corrected that awkward API in JUnit 5. (But if we migrate, we rather migrate fully to `kotlin.test` rather than Junit 5).
> 
> If you do that, consider whether it makes sense to group such test functions into one test function (i.e. they were only really in separate functions because of that API before).

I will do this in another PR.

----

Ready for review.

westnordost:
Nice, looks good.

FloEdelmann:

Issue #4448, opened 2022-10-01
Missing meta data / context in wheelchair quest


HolgerJeromin:
https://www.openstreetmap.org/node/1423239695

entrance : yes
shop : estate_agent

![rps20221001_181016_971](https://user-images.githubusercontent.com/2410353/193418294-edae4451-9dab-4610-8e6a-ff36df96f104.jpg)

SC  v48.0-alpha1

westnordost:
This is because `shop=estate_agent` is not recognized by the iD presets. The iD presets support the much more widely used `office=estate_agent` instead. 

It is mentioned here, too:
https://github.com/openstreetmap/id-tagging-schema/issues/529

I guess the reason why iD does not propose to "upgrade" `shop=estate_agent` to `office=estate_agent`, i.e. recognize the former as "deprecated" is because that tag is actually not deprecated.
In any case, this is an upstream issue.

HolgerJeromin:
If SC knows nothing we should perhaps present the raw (main) tag so the user get a clue that the quest is not for drachenpoint (in the screenshot).
At least for shop=/office= elements.

westnordost:
No, not making exceptions to not confronting the users with the tags. Drachenpoint is shown as a separate node.

HolgerJeromin:
Skip the quest because how should the user be able to answer the quest?

westnordost:
In #4241 (@Helium314 ) it was changed that this quest is also asked for nameless properties. I guess this was done under the assumption that for every feature asked, a feature name can be found through the id editor presets.

A few quests, like the opening hours quest, explicitly check for whether the object can be named and only show it if that is the case

https://github.com/streetcomplete/StreetComplete/blob/4a8f743ffc55eb23a266a7f0c073f3d76fb270fc/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt#L176-L182

...but to do that for every (shop-POI) quest would be somewhat expensive to do I think - not to mention the code duplication.

So while the current situation is not favourable (because inconsistent), i.e. some quests are shown even if the feature cannot be referred to by name and some have these checks, the better long term solution is to "simply" improve the id presets to cover more shop types rather than dealing with it in the way as cited above.
Maybe we should (just) return to checking `name or brand or name:signed=no` for all shop-type quests (and trusting that `name:signed=no` is only set if it has a feature name as the `AddPlaceName` quest checks for that?)

@FloEdelmann @matkoniecz

[...]

PR #5198, opened 2023-08-16
prevent really unclear level quests with no detected object type

matkoniecz:
require object having known name or known object type otherwise mapper has no chance to answer question
similar to say https://github.com/streetcomplete/StreetComplete/commit/14f004013d1b321edbddac106c479bcaf0849e90

prompted by https://www.openstreetmap.org/node/6213197233

Without this change the level number quest may be asked without giving any info whatsoever what it is being asked about.

screen

Note that it will NOT block objects of known type with no name tags.

Similar filtering is used already by opening hours quest.


westnordost:
This has been discussed in particular in https://github.com/streetcomplete/StreetComplete/issues/4448#issuecomment-1264727891

westnordost:
So, it looks like the chosen solution should rather mimic what https://github.com/streetcomplete/StreetComplete/commit/a4c6d320a372de3298024bcc89060623cca94506 did for consistency.

matkoniecz:
That may fail in unusual cases, say `name=unusual_shop_value name:signed=no` but it is really unlikely as `name:signed=no` is added basically only by StreetComplete.

westnordost:
Yeah

PR #5200, opened 2023-08-17
set and enforce a maximum tilt

### LINKED ISSUE #5177: Crash when trying to add address in overlay mode
**How to Reproduce**
1. Turn on address overlay
2. Tilt the map
3. Tap +
4. Crash

https://github.com/streetcomplete/StreetComplete/assets/25708359/c60f9dd1-4534-4ca5-bbb0-97d1f653c5ad

**Versions affected**
StreetComplete 53.3 Android 13 OneUI 5.1 Samsung Galaxy A52

westnordost:
fixes #5177

Set maximum tilt to 60° (PI/6) and enforce it also during the shove gesture.

@Helium314 tested that Tangram is not able to find a position at 60.47° tilt and above. So if the crash persists on other devices, we could just set the max tilt to 45° or something.


Helium314 (approved):
Looks good, and doesn't crash for me any more
(I actually had thought setting a maximum tilt would require a little less change)

westnordost:
Maybe in MapLibre it will be more straightforward. I really hope you (or someone else) will see the migration to the end, we have so many workarounds for tangram-es already.. :-/

Helium314:
> Maybe in MapLibre it will be more straightforward

I think there is a simple setting for maximum tilt.

PR #5202, opened 2023-08-17
test: use assertFailsWith instead of expected

riQQ:
As promised in https://github.com/streetcomplete/StreetComplete/pull/5193#issuecomment-1678464000
> > > Other than that, some classes use `@Test(expected = <ExceptionClass>::class)`. Should this also be replaced?
> > 
> > 
> > Oh yeah, I'd like that! Not required for this PR to be merged, though. Never liked that API. I think they also corrected that awkward API in JUnit 5. (But if we migrate, we rather migrate fully to `kotlin.test` rather than Junit 5).
> > If you do that, consider whether it makes sense to group such test functions into one test function (i.e. they were only really in separate functions because of that API before).
> 
> I will do this in another PR.


westnordost (approved):
LGTM!

FloEdelmann:

Issue #2536, opened 2021-01-29
Further reduce the number of cycleway quests for maxspeed=30 ways


cyclingcat:
Hi there,

it's me again, still wondering why there are villages with many cyclingway quests in maxspeed=30 zones where other villages don't show them anymore. So I tried to find a pattern behind this and my current suspect is the following combination of tags:

```
maxspeed=30
zone:traffic=DE:urban
```
While this combination isn't widespread in general, there seem to be clusters where it occurs quite often. See [this Overpass turbo query](http://overpass-turbo.eu/s/12Vd).

However, this tag combination is necessary but not sufficient for the superfluous cycleway quests. See for example this screenshot:

![cycleway](https://user-images.githubusercontent.com/32601012/106252931-1f126080-6217-11eb-8acb-0c86389f825a.png)

For way 1 (= [OSM way 31774388](https://www.openstreetmap.org/way/31774388)) the cycleway quest is displayed while for way 2 (= [OSM way 31774385](https://www.openstreetmap.org/way/31774385)) it's not. If you compare the tags of these ways, the only difference is the "name" tag.

(BTW: Investigations like this are an additional use case for profiles! I wished I had a possibility to use a profile with no quests at all in order to deliberately activate just a single quest which would remove cluttering the display (and the screenshot) with quests irrelevant for this issue). 

(See also issues #2251 and #2486.)

The cycling cat

westnordost:
It is not showing for Weisse Gasse because the footway branching off from Lindenstraße leading into the park is both almost parallel to the street for a few meters and close enough to the street that the algorithm decided that it might be a separately mapped sidewalk. The cycleway quest is thus not shown.

---

Regarding the `zone:traffic`:

I've looked at the code, the part of the filter that is relevant is 
```
highway = residential and (
  maxspeed > 30
  or (maxspeed ~ ".*mph" and maxspeed !~ "([1-9]|1[0-9]|20) mph")
  or $NOT_IN_30_ZONE_OR_LESS
)
```
and I guess what you would like it to be is
```
highway = residential and not (
  maxspeed <= 30
  and (maxspeed !~ ".*mph" or maxspeed ~ "([1-9]|1[0-9]|20) mph")
  and $IN_30_ZONE_OR_LESS
)
```

However, the `not` operator is not supported in the filter syntax. So far there hasn't been the necessity for it, plus back when the statements were still translated to overpass QL, it would have increased complexity by a lot.

[...]

PR #5210, opened 2023-08-22
feat: filter out more max-speed zone ways in add-cycleway quest

riQQ:
Filter out more ways that couldn't be handled previously because the not operator was not supported yet.
See https://github.com/streetcomplete/StreetComplete/issues/2536#issuecomment-769824333 for some discussion.


westnordost:
> Filter out more ways that couldn't be handled previously because the not operator was not supported yet.

Oh, it could!

`~key !~ value`

westnordost:
Anyway, without having read too closely, it looks like I mentioned a different filter than the one you coded.

riQQ:
The main point is to work around the problem, that `zone:traffic` has the documented value `DE:urban` and the undocumented value `DE:zone30`, so it's not clear if `DE:urban` means that it's not in a max-speed 20 / 30 zone. By changing the filter as proposed here it is enough that any of relevant tags indicates a max-speed zone to exclude the way from the add cycleway quest.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -164,7 +164,7 @@ private val untaggedRoadsFilter by lazy { """
 """.toElementFilterExpression() }
 
 private val notInZone30OrLess = """
-   ~"${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}" ~ ".*:(urban|rural|trunk|motorway|nsl_single|nsl_dual)"
+   !(~"${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}" ~ ".*:(30|20)")
```
westnordost, line 167: The filter should be like in `AddRoadWidth` or `CyclewayOverlay`:

`".*:(zone)?:?([1-9]|[1-2][0-9]|30)"`

westnordost (commented):
The same change to the filter as is being done to `AddCycleway` should be done to `AddLanes` for consistency, I'd say. (Take care that `!(key ~ value)` returns true when the `key` does not exist, too)

riQQ:
> The same change to the filter as is being done to `AddCycleway` should be done to `AddLanes` for consistency, I'd say. (Take care that `!(key ~ value)` returns true when the `key` does not exist, too)

I think I found a good solution for this, see the latest force push.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCyclewayForm.kt
```diff
@@ -34,11 +34,14 @@ import kotlinx.serialization.json.Json
 class AddCyclewayForm : AStreetSideSelectForm<CyclewayAndDirection, LeftAndRightCycleway>() {
 
     override val buttonPanelAnswers get() =
-        if (isDisplayingPrevious) listOf(
+        if (isDisplayingPrevious) {
+            listOf(
             AnswerItem(R.string.quest_generic_hasFeature_no) { isDisplayingPrevious = false },
             AnswerItem(R.string.quest_generic_hasFeature_yes) { onClickOk() }
         )
```
FloEdelmann, line 41: Indentation:

```suggestion
        if (isDisplayingPrevious) {
            listOf(
                AnswerItem(R.string.quest_generic_hasFeature_no) { isDisplayingPrevious = false },
                AnswerItem(R.string.quest_generic_hasFeature_yes) { onClickOk() }
            )
```

riQQ: I mistakenly pushed this file. This is fixed with 46a24ef3c9d4190cffb92209cfd5e9317ce9ef7b.

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
```diff
@@ -0,0 +1,12 @@
+package de.westnordost.streetcomplete.quests
+
+import de.westnordost.streetcomplete.osm.MAXSPEED_TYPE_KEYS
+
+val anyMaxSpeedTagKey = "~${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}"
+
+val onlyInANoneMaxSpeed30OrLessZone = """
```
westnordost, line 7: Typo: **_Non_**, not _None_.

The common name for these things is just [30 zone](https://en.wikipedia.org/wiki/30_km/h_zone) (or 20 mph zone). Since the actual speed limit is variable, depending on the sign, in StreetComplete, you'll find that in some places it is called "slow zone".

So, I'd prefer "notInSlowZone", or better, "inSlowZone" and negate it in usage.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
```diff
@@ -156,7 +157,7 @@ private val untaggedRoadsFilter by lazy { """
       and (
         !maxspeed
         or maxspeed > 20
-        or $notInZone30OrLess
+        or $onlyInANoneMaxSpeed30OrLessZone
```
westnordost, line 160: What about the `notInZone30OrLess` variable now?

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
```diff
@@ -0,0 +1,12 @@
...
+  (
+    $anyMaxSpeedTagKey !~ ".*:(zone)?:?([1-9]|[1-2][0-9]|30)"
+    and $anyMaxSpeedTagKey ~ ".*:(urban|rural|trunk|motorway|nsl_single|nsl_dual)"
```
westnordost, line 10: I thought a reason to do this at all was to get rid of the hardcoded list of maxspeed type values (since it can be somewhat of a free-form value depending on what the local community decides what values there should be, see GB community with their nsl stuff.

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
```diff
@@ -0,0 +1,12 @@
+package de.westnordost.streetcomplete.quests
```
westnordost, line 1: Such things should go into the osm package, e.g. into MaxspeedZone.kt or Maxspeed.kt.

riQQ: I placed it there initially but somehow convinced myself it would be better in quests. But you're right. This could be used by other stuff as well, e.g. overlays.

app/src/main/java/de/westnordost/streetcomplete/quests/cycleway/AddCycleway.kt
riQQ: What do you mean? Is this about the name, too? Or about something else?

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
riQQ: The only other thing I can think of right now that works with all unit tests is the following:

```
val onlyInANoneMaxSpeed30OrLessZone = """
  (
    $anyMaxSpeedTagKey
    and !(maxspeed ~ "walk|(\d+( ?mph)?)" and ${(MAXSPEED_TYPE_KEYS).joinToString(" and ") { "!$it" }})
    and $anyMaxSpeedTagKey !~ ".*:(zone)?:?([1-9]|[1-2][0-9]|30)"
  )
"""
```

westnordost: Why wouldn't `$anyMaxSpeedTagKey ~ ".*:(zone)?:?([1-9]|[1-2][0-9]|30)"` suffice?

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
riQQ: But it's more than that. It's also not a simple maxspeed value.

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
riQQ: Because it doesn't capture the assumption that something that is only tagged with the `maxspeed` tag and no other maxspeed tags is assumed to be in a max-speed zone if the speed is low enough.

With following it works and is more explicit:
```
  !(
    maxspeed ~ "walk|(\d+( ?mph)?)" and !(~${(MAXSPEED_TYPE_KEYS).joinToString("|")})
    or $explicitlyInSlowZone
  )
```

app/src/main/java/de/westnordost/streetcomplete/osm/Maxspeed.kt
```diff
@@ -10,6 +10,18 @@ val MAXSPEED_TYPE_KEYS = setOf(
     "zone:traffic"
 )
 
+val anyMaxSpeedTagKey = "~${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}"
+
+val explicitlyInSlowZone = """$anyMaxSpeedTagKey ~ ".*:(zone)?:?([1-9]|[1-2][0-9]|30)""""
```
westnordost, line 15:
```suggestion
val explicitlyInSlowZone = """$anyMaxSpeedTagKey ~ ".*:(zone:?)?([1-9]|[1-2][0-9]|30)""""
```

app/src/main/java/de/westnordost/streetcomplete/osm/Maxspeed.kt
```diff
@@ -10,6 +10,18 @@ val MAXSPEED_TYPE_KEYS = setOf(
...
+
+// This assumes that elements only tagged with 'maxspeed' and no other maxspeed tags are in a slow zone.
```
westnordost, line 17: Hmm... that's rather bold, and **eurocentric**. And even within (Germany), there are plenty of tempo-30 roads that are not 30 zones.

Slow zones as a concept do not exist in that many countries, and hence, the associated legislation does also not exist. I.e. in this case, that the local traffic authority may only designate roads as within 30 zones if there is no compulsory cycleways.

Incidentally, the source:maxspeed tagging in Germany (and Europe) is most complete when compared to the rest of the world, so I think even the assumption that if maxspeed=30 it must be a 30-zone will often be wrong even in Germany / Europe.

riQQ: So the naming and comment is incorrect. Or do you propose to change the filter and the tests?
I'm not sure how to proceed.

westnordost: Well, I wouldn't like this behavior: That simply a `maxspeed=30` is assumed to be a 30-zone and this means that surely there'll never be bike paths then.

app/src/main/java/de/westnordost/streetcomplete/quests/SharedFilters.kt
riQQ: If I leave out the part about having only the `maxspeed` tag and no other maxspeed tags, the following test fails. That means the quest would be shown for more elements than desired:

https://github.com/streetcomplete/StreetComplete/blob/e300eaab151c413c387e9a9e936e2b9013ad5dab/app/src/test/java/de/westnordost/streetcomplete/quests/lanes/AddLanesTest.kt#L55-L59

app/src/main/java/de/westnordost/streetcomplete/osm/Maxspeed.kt
riQQ: My comment is wrong, so I removed it.

riQQ:
Please also see my answer here https://github.com/streetcomplete/StreetComplete/pull/5210#discussion_r1326134203

app/src/main/java/de/westnordost/streetcomplete/osm/Maxspeed.kt
```diff
@@ -10,6 +10,17 @@ val MAXSPEED_TYPE_KEYS = setOf(
     "zone:traffic"
 )
 
+val anyMaxSpeedTagKey = "~${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}"
+
+val explicitlyInSlowZone = """$anyMaxSpeedTagKey ~ ".*:(zone:?)?([1-9]|[1-2][0-9]|30)""""
+
+val notInSlowZoneAndNotPlainMaxspeedTagging = """
+  !(
+    maxspeed ~ "walk|(\d+( ?mph)?)" and !(~${(MAXSPEED_TYPE_KEYS).joinToString("|")})
```
westnordost, line 19:
```suggestion
    maxspeed ~ "walk|(\d+( ?mph)?)" and !~${(MAXSPEED_TYPE_KEYS).joinToString("|")}
```

app/src/main/java/de/westnordost/streetcomplete/osm/Maxspeed.kt
```diff
@@ -10,6 +10,17 @@ val MAXSPEED_TYPE_KEYS = setOf(
     "zone:traffic"
 )
 
+val anyMaxSpeedTagKey = "~${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}"
+
+val explicitlyInSlowZone = """$anyMaxSpeedTagKey ~ ".*:(zone:?)?([1-9]|[1-2][0-9]|30)""""
+
+val notInSlowZoneAndNotPlainMaxspeedTagging = """
+  !(
+    maxspeed ~ "walk|(\d+( ?mph)?)" and !~${(MAXSPEED_TYPE_KEYS).joinToString("|")}
```
westnordost, line 19: uhm, but what about `source:maxspeed=sign` then?

westnordost: (Also, is there a test that additionally checks for this condition? I am having still problems undestanding why this is there in the first place, and a test would also serve as documentation.)

westnordost: > If I leave out the part about having only the `maxspeed` tag and no other maxspeed tags, the following test fails. That means the quest would be shown for more elements than desired:

> https://github.com/streetcomplete/StreetComplete/blob/e300eaab151c413c387e9a9e936e2b9013ad5dab/app/src/test/java/de/westnordost/streetcomplete/quests/lanes/AddLanesTest.kt#L55-L59

How can it fail, though? The filter currently looks like this:

```
or highway = residential and (
             maxspeed > 33
             or ~"${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}" ~ ".*:(urban|rural|trunk|motorway|nsl_single|nsl_dual)"
           )
```

... see the maxspeed > 33

riQQ: The current test doesn't fail.
But it fails, if I leave out `maxspeed ~ "walk|(\d+( ?mph)?)" and !~${(MAXSPEED_TYPE_KEYS).joinToString("|")}`  the Because it's not explicitly in a slow zone `!($explicitlyInSlowZone)` becomes true.

westnordost:
So, I did review, had trouble understanding, and just changed it myself that it is more clearly understandable. Also, added tests to test that the different filter fragments actually do what we want.

@riQQ **HOWEVER**, there is now a failing test. And, it is not the fault of the filter but it seems to indicate a critical bug in the filter parser itself(!!). I didn't investigate deeply yet, but hopefully it is only related to the new `!(...)` syntax.

Basically, it looks like this expression from `AddLanes?`

```
maxspeed > 33
or (
  !maxspeed
  or ~source:maxspeed|zone:maxspeed|maxspeed:type|zone:traffic|maxspeed ~ "implicit|([A-Z-]+:.*)"
)
and !(~source:maxspeed|zone:maxspeed|maxspeed:type|zone:traffic|maxspeed ~ "[A-Z-]+:(zone:?)?([1-9]|[1-2][0-9]|30)")
```

is broken down during parsing to

```
maxspeed > 33
or !maxspeed
or ~source:maxspeed|zone:maxspeed|maxspeed:type|zone:traffic|maxspeed ~ "implicit|([A-Z-]+:.*)"
and !(~source:maxspeed|zone:maxspeed|maxspeed:type|zone:traffic|maxspeed ~ "[A-Z-]+:(zone:?)?([1-9]|[1-2][0-9]|30)")
```

or in short

```
a or (b or c) and !(d)
```
to 
```
a or b or c and !(d)
```

westnordost:
I added a (failing) test in master. Would you be able to have a look at it on short notice? It is definitely related to the new `!(...)` syntax but I am not sure how critical it is, i.e. it wouldn't be that critical if it only afflicted filters that actually use this syntax because I think this PR will be the only place so far where it is used,.

westnordost:
image

westnordost:
Okay, it looks like your change had nothing to do with this. It looks like this bug in the filter has been there all along... what the fuck, how did this never surface? Did we by chance write all the filter expressions in a way that this bug would never trigger?? :-o !

westnordost (approved):
But maybe *you* want to review it now.

app/src/test/java/de/westnordost/streetcomplete/quests/cycleway/AddCyclewayTest.kt
```diff
@@ -208,4 +208,18 @@ class AddCyclewayTest {
...
+    @Test
+    fun `not applicable to residential road with maxspeed 30`() {
+        val residentialWayWithMaxspeed30 = way(1L, listOf(1, 2, 3), mapOf(
+            "highway" to "residential",
+            "maxspeed" to "30",
+        ))
+
+        val mapData = TestMapDataWithGeometry(listOf(residentialWayWithMaxspeed30))
+
+        // a residential way with maxspeed=30 and no other maxspeed tags is assumed to be in a max-speed 30 zone
```
riQQ, line 221: What about the following way:
```
highway=residential
maxspeed=30
zone:traffic=DE:urban
```

Right now it will be applicable. Is this intended?

riQQ:
> Okay, it looks like your change had nothing to do with this. It looks like this bug in the filter has been there all along... what the fuck, how did this never surface? Did we by chance write all the filter expressions in a way that this bug would never trigger?? :-o !

Nice find! Or maybe it wasn't that obvious? Hard to tell without analyzing all the filters.

> But maybe _you_ want to review it now.

One comment, rest looks good.

app/src/test/java/de/westnordost/streetcomplete/quests/cycleway/AddCyclewayTest.kt
westnordost: Yes, because (in Germany) only in 30 **zones** the traffic agency is not allowed to establish (benutzungspflichtige) bike paths. (Or actually, the other way round, it can only be a 30 zone if there are no benutzungspflichtige bike paths)

PR #5211, opened 2023-08-23
Bbq fuel quest (Issue #4832)

qugebert:
My first attempt of implementing this quest from Issue [#4832](https://github.com/streetcomplete/StreetComplete/issues/4832)


app/src/main/res/values-de/strings.xml
```diff
@@ -296,6 +296,12 @@ Andernfalls kannst du eine alternative Tastatur im App Store herunterladen. Beli
     <string name="quest_powerPolesMaterial_wood">"Holz"</string>
     <string name="quest_powerPolesMaterial_metal">"Stahl"</string>
     <string name="quest_powerPolesMaterial_concrete">"Beton"</string>
+
+    <string name="quest_bbq_fuel_title">Womit wird dieser Grill betrieben?</string>
```
matkoniecz, line 300: translations are only imported from translation site, any data here will be overwritten by such import

if you can translate it: it is really welcome to add it to translation site after PR is merged and new version is being prepared

app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBBQFuel.kt
```diff
@@ -0,0 +1,39 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.OUTDOORS
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddBBQFuel : OsmFilterQuestType<BBQFuel>() {
```
FloEdelmann, line 12: Not sure what the convention in this repository is, but I always prefer "camel-casing" acronyms, i.e. `AddBbqFuel`, `BbqFuel`, etc. instead of `AddBBQFuel`, `BBQFuel`, etc.

app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBBQFuel.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    override val elementFilter = """
+        nodes, ways with
+        (
+          amenity = bbq
+          and !fuel
+        )
+        and access !~ no|private
```
FloEdelmann, line 20:
```suggestion
        nodes, ways with
          amenity = bbq
          and !fuel
          and access !~ no|private
```

app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBBQFuel.kt
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+
+    override val changesetComment = "Specify what grill is powered by"
```
FloEdelmann, line 24:
```suggestion
    override val changesetComment = "Specify barbecue fuel"
```

app/src/main/res/values/strings.xml
```diff
@@ -1284,6 +1284,13 @@ If there are no signs along the whole street which apply for the highlighted sec
...
     <string name="quest_powerPolesMaterial_concrete">Concrete</string>
 
+
+    <string name="quest_bbq_fuel_title">What is this grill powered by?</string>
+    <string name="quest_bbq_fuel_wood">Wood</string>
+    <string name="quest_bbq_fuel_electric">Electric</string>
+    <string name="quest_bbq_fuel_charcoal">Charcoal</string>
+
+
     <string name="quest_railway_crossing_barrier_title2">What barriers are used at this railway crossing?</string>
```
FloEdelmann, line 1294:
```suggestion
    <string name="quest_powerPolesMaterial_concrete">Concrete</string>

    <string name="quest_bbq_fuel_title">What is this grill powered by?</string>
    <string name="quest_bbq_fuel_wood">Wood</string>
    <string name="quest_bbq_fuel_electric">Electric</string>
    <string name="quest_bbq_fuel_charcoal">Charcoal</string>

    <string name="quest_railway_crossing_barrier_title2">What barriers are used at this railway crossing?</string>
```

app/src/main/res/values-de/strings.xml
qugebert: Thank you, i will skip this step next time

app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBBQFuel.kt
qugebert: Hast recht, in den ATM-Quests ist es auch camel-case. Habs geändert.

westnordost (commented):
Code looks flawless and you thought of every little thing (even attribution etc).

The monochrome icon is pretty boring though, if not ugly. It will be some time till this is merged anyway so I'd take care of this then.

One thing I noticed, generally, while looking at the PR, is that it is kind of really involved to create a new quest as one needs to touch so many files.

Looking at what **could** be changed, I wonder if the whole _PinIcons_ stuff should be dropped and instead always the icons of the fitting iD presets should be used instead.

Also, I wonder if it was more neatly arranged if the model (in this case, BbqFuel) went into the same file as the quest, as it is somewhat related. I mean, generally, in the future, not for this PR.

What do you think?

westnordost (approved):

westnordost:
(needs icon)

PR #5212, opened 2023-08-24
Fix element highlighting in amenity_coverage

qugebert:
In the AddAmenityCover-Task,

it is asked for coverage of various amenitys (picnic_table,bbq, maybe more in the future), but the getHighlightedElements(..) was still hardcoded to highlight picnic_tables, but should highlight elements of the same type than surveyed to avoid misunderstandings, which element should be checked.

This was mostly copy&paste from the logik used in the CheckExistence-quest.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -246,7 +246,7 @@ fun questTypeRegistry(
     9 to AddCarWashType(),
 
     10 to AddBenchBackrest(),
-    11 to AddAmenityCover(),
+    11 to AddAmenityCover(featureDictionaryFuture),
 
     12 to AddBridgeStructure(),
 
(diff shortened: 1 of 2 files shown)


westnordost:
That's true, it is an issue and thank you very much for the fix!

westnordost (commented):
Hmm, you made some extra changes that weren't in the title. Why not simply also highlight bbqs? I am not a fan of introducing complexity where it is not needed.

qugebert:
> Hmm, you made some extra changes that weren't in the title. Why not simply also highlight bbqs? I am not a fan of introducing complexity where it is not needed.

I'm not sure, if i understand this right. 
I thought, this would be the easiest way, as this was already implemented.
And if someone adds other types of amenities in the future, the highlighting doesn't need to be changed again as it always searches for the neighbour of the same type.
How should i do this instead? Sorry, i'm really new to kotlyn/android programming so thank you for any suggestins how i can do this better.

westnordost:
I guess you are right. In the future, the interface could be changed to be more straightforward to have less code in the individual quests, e.g.

`class MyQuest(private val getFeature: (tags: Map<String, String>) -> Feature?) { ... }`

westnordost (approved):

westnordost (approved):

PR #5214, opened 2023-08-25
Add openUri extension functions

tapetis:
This PR adds two `openUri` extension functions that extract common functionality from 11 classes opening URLs in different ways. I tested these changes and everything still works as before.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/about/AboutFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/about/AboutFragment.kt
@@ -1,12 +1,10 @@
 package de.westnordost.streetcomplete.screens.about
 
-import android.content.Intent
 import android.os.Bundle
 import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.DrawableRes
 import androidx.appcompat.app.AlertDialog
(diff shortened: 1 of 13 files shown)


app/src/main/java/de/westnordost/streetcomplete/screens/about/AboutFragment.kt
```diff
@@ -86,15 +72,20 @@ class AboutFragment : TwoPaneListFragment(), HasTitle {
...
-        return true
+    private fun setUrlClickListener(preferenceKey: String, url: String) {
+        setUrlClickListener(findPreference(preferenceKey), url)
+    }
+
+    private fun setUrlClickListener(preference: Preference?, url: String) {
+        preference?.setOnPreferenceClickListener {
+            openUri(url)
+            true
+        }
```
westnordost, line 83: I think this would be better / more readable:

```
    private fun Preference.setClickOpensUrl(url: String) {
        setOnPreferenceClickListener {
            openUri(url)
            true
        }
```

or even drop these completely.

westnordost (commented):
Thanks, looks good! I find the convenience function in the about fragment not so great, though.

westnordost (approved):

PR #5262, opened 2023-09-23
Don't show ok button in AddBuildingType if category value is null

Helium314:
Currently when selecting e.g. "for cars" category in `AddBuildingType`, the ok button is shown, but when clicking ok the user is asked to select a more specific value.
With this change, the ok button is not shown any more when selecting such a category. This should avoid unnecessary attempts on selecting some categories, while at the same time providing a subtle hint that categories can actually be used for answering the quest too.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AGroupedImageListQuestForm.kt
@@ -134,5 +134,5 @@ abstract class AGroupedImageListQuestForm<I, T> : AbstractOsmQuestForm<T>() {
 
     abstract fun onClickOk(value: I)
 
-    override fun isFormComplete() = selectedItem != null
+    override fun isFormComplete() = selectedItem?.value != null
 }


matkoniecz (approved):

mnalis:
If I understand correctly, it would only apply to `For cars`, `On a farm` and `Other` categories, while other categories (`Residential`, `Commercial building`, `Civic building`, `Religious building`) would still retain `OK` button and be selected? Good idea!

Also, if `OK` button is not shown in such cases, would it make `quest_generic_item_invalid_value` text and related code unneeded?

Helium314:
> If I understand correctly, it would only apply to For cars, On a farm and Other categories, while other categories [...] would still retain OK button and be selected?

Exactly

> Also, if `OK` button is not shown in such cases, would it make `quest_generic_item_invalid_value` text and related code unneeded?

I don't think so. It's been a while, but I think it is/was somehow possible to press the ok button when it shouldn't be visible at all. Possibly related to the appear/disappear animation.
Though maybe the button showing condition is checked again when pressing the ok button, didn't look this up.

westnordost:
Well, whether or not that string could be removed now (and the code that makes it show) would be something for a different PR. Thanks @Helium314

PR #5263, opened 2023-09-23
Use Android's built-in XmlPull implementation

tapetis:
This PR fixes a build error with AGP version 8 or greater, where the previously ignored proguard / R8 warning for including a program (i.e., built-in Android) class in a library is now considered to be an error. For details, see [here](https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes) and [here](https://issuetracker.google.com/issues/247066506).

The app still works fine with these changes on my device. All XML classes used by the [osmapi library](https://github.com/westnordost/osmapi) are part of Android since API level 1.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -101,6 +101,7 @@ configurations {
     all {
         // it's already included in Android
         exclude(group = "net.sf.kxml", module = "kxml2")
+        exclude(group = "xmlpull", module = "xmlpull")
     }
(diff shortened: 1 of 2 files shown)


westnordost:
Oh, that's super useful! If I remember correctly, I wanted to upgrade AGP several times but found myself in unsolveable (to me) build problems after, so I reverted the upgrade!

PR #5264, opened 2023-09-23
Improve profile layout on wide screens & with large fonts

tapetis:
This PR improves the profile screen by:

1. Fixing the "osm profile" and "logout" buttons from overlapping with large fonts on small screens by then displaying them in two lines.
2. Allowing scrolling using the entire screen width when the screen is wider than 400 dp (see the scrollbar in the screenshots below).
3. Scaling the top two text views with the user-selected font size (sp instead of dp)
4. Removing unnecessary inner layouts. Most of the highlighted lines in the diff are due to the resulting change in indentation.

Before:

After:


riQQ:
As an aside appending `w=1` at the end of the URL makes the diff ignore whitespace changes:
https://github.com/streetcomplete/StreetComplete/pull/5264/files?w=1

westnordost:
Nice, thank you!

PR #5265, opened 2023-09-23
Move map lifecycle callbacks to onStart / onStop

tapetis:
This PR changes the lifecycle callbacks so that the tangram map is resumed in `onStart` and paused in `onStop` instead of in `onResume` and `onPause`. This causes the callbacks to be executed a little earlier or later, respectively.

This prevents the map from being completely black during the back gesture animation (when enabled) on Android 14 (see the screenshots below). This is also the recommended behavior on devices supporting multi-window but not multi-resume (see [here](https://developer.android.com/guide/topics/large-screens/multi-window-support#lifecycle)), where an activity could be visible to the user in its paused state.

| Old | New |
| --- | --- |
|  |  |

I have been using the app with these changes for some time now and have not experienced any related issues.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/MapFragment.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/MapFragment.kt
@@ -163,15 +163,11 @@ open class MapFragment :
             delay(50)
             sceneMapComponent?.loadScene()
         }
-    }
-
-    override fun onResume() {
-        super.onResume()
         binding.map.onResume()
     }
 
-    override fun onPause() {
-        super.onPause()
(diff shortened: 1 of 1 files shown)


westnordost:
Hm well, but still `com.mapzen.tangram.MapView.onResume` is called. I digged into that and in the end, it calls `android.opengl.GLSurfaceView.onResume`. Unless it is specifically documented (that it is fine) otherwise, I would not like to break `GLSurfaceView`s expectation that this method is called on `onResume`.

westnordost:
Oh wow, that shall suffice. I wouldn't have thought that exactly the thing I mentioned would directly be mentioned in the documentation for that very function, lol.

Thank you!

PR #5266, opened 2023-09-23
Remove unnecessary dontnotes

tapetis:
The linked issue has been marked as fixed in 2018. I also do not see any related notes printed to the console when building without specifying these three proguard dontnotes.


westnordost:
ah ,cool

PR #5267, opened 2023-09-24
Update AGP to v8.1.1 and Gradle to v8.3

tapetis:
This PR updates the AGP and the Gradle wrapper (`gradle wrapper --gradle-version 8.3`) to the latest stable versions. See the results of the 'Validate Gradle Wrapper' GitHub task for the validity of the updated gradle-wrapper.jar.


westnordost:
Thanks! I did it myself now but set you as author. 

https://github.com/streetcomplete/StreetComplete/commit/5cc034cf6b9c47168b0a3e6ae94d51051ac0d8d5

Commits that push binaries in the repo I rather do myself

westnordost:
This is only tangentially related, but do you have any idea why changing `SomeEnum.values()` to `SomeEnum.entries` is still marked as error because it is missing the "experimental api" annotation in Android Studio even after updating to Kotlin 1.9.10?

PR #5268, opened 2023-09-24
Exclude own package from map app installed check

tapetis:
As StreetComplete itself can also open geo URIs, the "No other map application installed" toast was not displayed before even if no other map app was actually installed. This PR fixes the behavior by excluding the SC package name from the resolve check. I have tested on an emulator to confirm that the new implementation now works as expected both when another map app is installed and when not.


westnordost:
Makes sense, thank you!

Issue #5241, opened 2023-09-09
Crash


opk12:
It's a while I did not use StreetComplete. I opened it, it was at a very low zoom (showed Europe, Africa, part of America) and I zoomed a bit [around South Egypt](https://www.openstreetmap.org/#map=9/24.1242/32.8656). The app crashed.

```
java.lang.RuntimeException: Tried to perform an operation on an invalid pointer! This means you may have used a MapData that has already been removed.
at com.mapzen.tangram.MapData.checkPointer(MapData.java:116)
at com.mapzen.tangram.MapData.setFeatures(MapData.java:36)
at de.westnordost.streetcomplete.screens.main.map.components.DownloadedAreaMapComponent.set(DownloadedAreaMapComponent.kt:32)
at de.westnordost.streetcomplete.screens.main.map.DownloadedAreaManager$update$1.invokeSuspend(DownloadedAreaManager.kt:51)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:793)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:697)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:684)
Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@6afc11a, Dispatchers.Default]
```

**Versions:**
StreetComplete 53.3 from F-droid on Android 7.1

opk12:
> Currently I'm trying to have the `set` `synchronized`...

Regardless of this issue: Multi-threaded access to shared resources must always be inside `synchronized` (or the field must be `volatile`, which is effectively a shorthand) because the JVM (this is not Kotlin-specific) can legally
* do things non-atomically (writing an object reference into a field in two steps)
* reorder instructions (next line is executed before previous line)
* have each thread work on its local copy of the heap and not see updates from other threads

[...]

PR #5269, opened 2023-09-24
Fix crash #5241

Helium314:
Doing `DownloadedAreaMapComponent.set` synchronized avoids the situation described in https://github.com/streetcomplete/StreetComplete/issues/5241#issuecomment-1722423742.
With this change the crash is not reproducible any more, while without it is reproducible: https://github.com/streetcomplete/StreetComplete/issues/5241#issuecomment-1722410706

I can't guarantee all occurrences of this crash will be fixed by this change, but it fixes the crashes on startup I sometimes had.


westnordost:
Okay, cool! :-)

Issue #4916, opened 2023-03-31
No label / feature for post boxes


Helium314:
(works with iD presets 5.2.1)
Post boxes don't have a label (text below the question) any more, because the feature search returns nothing. This matters for the check existence quest, which requires a feature to be asked.

Search works when either using iD presets v5.2.1, or when searching by country (though results may obviously depend on country). Seems in iD presets 6.0.0 it's assumed post boxes are only a local thing for some reason...

Anyway, similar things might be happening for other features, just no one noticed. Maybe feature search should always specify a country?

matkoniecz:
Seems to be about `amenity=post_box`

Helium314:
https://github.com/streetcomplete/StreetComplete/blob/bb5a34485bd258b0fa898761264c068df0002af2/app/src/main/java/de/westnordost/streetcomplete/util/NameAndLocationLabel.kt#L74-L85
returns `null` for `amenity=post_box`.
But it works when providing a (any?) country using `.inCountry(countryOrSubdivisionCode)`.

Apparently this is because of `locationSet` that excludes GB and US in iD presets.
So in iD presets there is no generic world wide post box feature, and thus I see the point of osmfeatures not returning anything without specifying a country.

westnordost:
That it doesn't work correctly in osmfeatures may have something to do with that finding features is done based on indices rather than rummaging through the whole and ever-growing list of presets (in particular, for NSI) - for performance reasons. Now, the indices are based on per-country-basis only for NSI features, not normal presets, because the assumption is that an OSM feature that can exist in one country can exist in another too.

And this assumption is still true: Postboxes in the US are not something different from postboxes elsewhere. They are only a separate feature in the iD presets now to shoehorn an additional suggested field into the preset - whether it is drive-through or not.
But I think this is semantically wrong, and also from an architectural point of view, a dead-end: Postboxes in the UK are not something else altogether from postboxes in other countries only because OpenStreetMappers like to tag the `royal_cipher` on it. Semantically, it is not the preset itself that is country-specific, but the individual fields are.
[...]

PR #5273, opened 2023-09-25
Simplify feature dictionary handling

### LINKED ISSUE #5219: Simplify `featureDictionaryFuture` interface in quests
Currently, quests that want to use the feature dictionary, accept a `FutureTask<FeatureDictionary>`, e.g.

https://github.com/streetcomplete/StreetComplete/blob/80781a7f45cddf19223c46ed5ff1a35e3964bb0e/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt#L249

https://github.com/streetcomplete/StreetComplete/blob/80781a7f45cddf19223c46ed5ff1a35e3964bb0e/app/src/main/java/de/westnordost/streetcomplete/quests/amenity_cover/AddAmenityCover.kt#L16-L18

As suggested in https://github.com/streetcomplete/StreetComplete/pull/5212#issuecomment-1693484783 by @westnordost, this interface could be simplified to something like:

```kt
class MyQuest(private val getFeature: (tags: Map<String, String>) -> Feature?) { ... }
```

@qugebert Would you fancy giving it a try?

FloEdelmann:
Resolves #5219.

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
@@ -1,4 +1,4 @@
-package de.westnordost.streetcomplete.quests.existence
+package de.westnordost.streetcomplete.util
 
 import android.content.res.Configuration
 import android.content.res.Resources
@@ -7,28 +7,22 @@ import de.westnordost.osmfeatures.AndroidFeatureDictionary
 import de.westnordost.osmfeatures.FeatureDictionary
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.data.osm.mapdata.Node
-import de.westnordost.streetcomplete.util.getNameAndLocationLabel
 import java.util.Locale
-import java.util.concurrent.FutureTask
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
-class CheckExistenceLabelTest {
-    private var featureDictionaryFuture: FutureTask<FeatureDictionary>
+class NameAndLocationLabelTest {
+    private var featureDictionary: FeatureDictionary
     private var englishResources: Resources
-    private var questType: CheckExistence
 
     init {
         val context = InstrumentationRegistry.getInstrumentation().targetContext
-        featureDictionaryFuture = FutureTask { AndroidFeatureDictionary.create(context.assets, "osmfeatures/default", "osmfeatures/brands") }
-        featureDictionaryFuture.run()
+        featureDictionary = AndroidFeatureDictionary.create(context.assets, "osmfeatures/default", "osmfeatures/brands")
 
         val conf = Configuration(context.resources.configuration)
         conf.setLocale(Locale.ENGLISH)
         val localizedContext = context.createConfigurationContext(conf)
         englishResources = localizedContext.resources
-
-        questType = CheckExistence(featureDictionaryFuture)
     }
 
     // https://github.com/streetcomplete/StreetComplete/issues/2512
@@ -90,6 +84,6 @@ class CheckExistenceLabelTest {
         getNameAndLocationLabel(
             Node(0, LatLon(0.0, 0.0), tags),
             englishResources,
-            featureDictionaryFuture.get()
+            featureDictionary
         )?.toString()
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/shops/ShopsOverlay.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/shops/ShopsOverlay.kt
@@ -16,6 +16,7 @@ import de.westnordost.streetcomplete.quests.place_name.AddPlaceName
 import de.westnordost.streetcomplete.quests.shop_type.CheckShopType
 import de.westnordost.streetcomplete.quests.shop_type.SpecifyShopType
 import de.westnordost.streetcomplete.util.getNameLabel
+import de.westnordost.streetcomplete.util.ktx.getFeature
 import java.util.concurrent.FutureTask
 
 class ShopsOverlay(private val featureDictionaryFuture: FutureTask<FeatureDictionary>) : Overlay {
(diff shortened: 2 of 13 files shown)


app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
```diff
@@ -7,28 +7,22 @@ import de.westnordost.osmfeatures.AndroidFeatureDictionary
...
-        featureDictionaryFuture = FutureTask { AndroidFeatureDictionary.create(context.assets, "osmfeatures/default", "osmfeatures/brands") }
-        featureDictionaryFuture.run()
+        featureDictionary = AndroidFeatureDictionary.create(context.assets, "osmfeatures/default", "osmfeatures/brands")
 
         val conf = Configuration(context.resources.configuration)
         conf.setLocale(Locale.ENGLISH)
         val localizedContext = context.createConfigurationContext(conf)
         englishResources = localizedContext.resources
-
-        questType = CheckExistence(featureDictionaryFuture)
```
FloEdelmann, line 31: Since this test does not depend on the `CheckExistence` quest anymore, but rather the `getNameAndLocationLabel` function, I renamed the test from `CheckExistenceLabelTest` to `NameAndLocationLabelTest`.

app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
FloEdelmann, line 1: I noticed that the `postBox` test (based on #2640) is failing both on the `master` branch and in this PR:

```
expected:<Deutsche Post[ (Mail Drop Box)]> but was:<Deutsche Post[]>
```

Maybe since #3870? Can we do something about this?

FloEdelmann: Hmm apparently it was introduced later: https://github.com/streetcomplete/StreetComplete/pull/3870#discussion_r825439861

riQQ: See #4916

FloEdelmann: Ah, thanks for finding and linking that issue. I'll resolve this thread then, as there is already a tracking issue.

app/src/test/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistenceTest.kt
```diff
@@ -5,14 +5,14 @@ import de.westnordost.streetcomplete.quests.TestMapDataWithGeometry
...
 import de.westnordost.streetcomplete.testutils.node
 import de.westnordost.streetcomplete.testutils.on
-import java.util.concurrent.FutureTask
+import de.westnordost.streetcomplete.util.ktx.getFeature
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
 class CheckShopExistenceTest {
-    private val questType = CheckShopExistence(mockOfFeatureDictionary())
+    private val questType = CheckShopExistence { mockOfFeatureDictionary().getFeature(it) }
```
westnordost, line 13: We don't need any mocking anymore in the tests. It's just a function now.

E.g. 

```
if (tags["shop"] == "greengrocer") return Feature(...)
else return null
```

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -246,7 +247,7 @@ fun questTypeRegistry(
     9 to AddCarWashType(),
 
     10 to AddBenchBackrest(),
-    11 to AddAmenityCover(featureDictionaryFuture),
+    11 to AddAmenityCover { featureDictionaryFuture.get().getFeature(it) },
```
westnordost, line 250: Looks like the function should be passed as a parameter of `questTypeRegistry`, too.

app/src/main/java/de/westnordost/streetcomplete/overlays/shops/ShopsOverlay.kt
```diff
@@ -41,10 +42,7 @@ class ShopsOverlay(private val featureDictionaryFuture: FutureTask<FeatureDictio
         mapData
             .filter(IS_SHOP_OR_DISUSED_SHOP_EXPRESSION)
             .map { element ->
-                val feature = featureDictionaryFuture.get()
-                    .byTags(element.tags)
-                    .isSuggestion(false) // no suggestions because we just want the icon
-                    .find().firstOrNull()
+                val feature = featureDictionaryFuture.get().getFeature(element.tags)
```
westnordost, line 45: The feature dictionary here is used the same way as in the quests. So, instead of the future, just pass the function in the constructor of ShopsOverlay

westnordost (commented):
Looks good!

app/src/test/java/de/westnordost/streetcomplete/quests/shop_type/CheckShopExistenceTest.kt
FloEdelmann: Ah, you're right. That actually simplifies mocking a lot!

Fixed in [`ea6264b`](https://github.com/streetcomplete/StreetComplete/pull/5273/commits/ea6264b06a2c758089a79ab121a398ab69088069).

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
FloEdelmann: Done in [`af75a4c`](https://github.com/streetcomplete/StreetComplete/pull/5273/commits/af75a4ca3e1dc7f58ab61a2a9c64c6867c039dc8).

app/src/main/java/de/westnordost/streetcomplete/overlays/shops/ShopsOverlay.kt
FloEdelmann: Done in [`2bd3207`](https://github.com/streetcomplete/StreetComplete/pull/5273/commits/2bd32072a828d5c47b630e48a53b814e343ec3f5).

westnordost:
Looks good to me now, I made a small change.

Issue #5237, opened 2023-09-07
Quests that maybe should be shown in Slovenia


CTJoriginal:
**Does this stop have tectile paving?**
- This is quite common, especially on bus stops in cities and on a lot of train stops

**Is there tactile paving at the top and bottom of these steps?**
- this isnt quite common on regular stairs but is really common on stairs neae train stations (if it's possible to activate quests just for the vicinity of specific places, this could be a good solution)

**Is there tactile paving on this curb here?**
- this feature is really common in cities

**Is there a bicycle lane or path here? What type?**
- We do have many cycling lanes on roads, especially in cities
I found this law about cycling paths: 
http://www.pisrs.si/Pis.web/pregledPredpisa?id=PRAV11505
Section 33
- the country Metadata that defines whatever there are or aren't cycle lanes contains comment with street view pointing directly at example of cycling lane

~~**What track number is here?**~~
~~- If I understood wiki correctly, this feature shows track number on train stations~~
~~https://github.com/streetcomplete/StreetComplete/assets/48687492/5f0b7760-426c-4a8f-a2d2-20ec016a962c~~
~~We have similar track numbering on nearly all train stations~~

I think it's best if these features are disabled by default, but it really bugs me that I'm unable to map things that clearly are there and could help many blind people

I can provide further information if necessary

mnalis:
- In my (limited) experience in Slovenia, enabling `tactile paving` quests probably make sense 
- `Is there a bicycle lane or path here? What type?` can you be more specific what you think needs changing? This line https://github.com/streetcomplete/countrymetadata/blob/2b2011b6fe79eaad0f18c3a5492bf373ac9c4430/data/exclusiveCycleLaneStyle.yml#L127 ?
- `What track number is here?` - Unless I'm wrong, isn't that [SCEE](https://wiki.openstreetmap.org/wiki/SCEE) quest which should be reported [there](https://github.com/Helium314/SCEE/issues)?

_(also, it is preferable to have **one-issue-per-request**. Ok, multiple tactile paving quests perhaps could be bundled together as they are very similar, but not all different issues like cycleways, trains etc. Because, it makes it hard to track - e.g. one might do a PR that would enable tactile pavings in Slovenia, but unless **all** the different suggestions in the issue are resolved, the issue cannot be closed, thus it would become big and hard to follow with several different discussion interspersed)_

CTJoriginal:
Okay, I'll open multiple issues next time

[...]

PR #5274, opened 2023-09-25
Added Slovenia to list of countries where tectile paving is common based on discussion in #5237

CTJoriginal:
Added Slovenia to list of countries where tectile paving is common based on discussion in #5237.


CTJoriginal:
I also updated the cycleway quests discussed in the same issue.

PR #5277, opened 2023-09-27
Change enabled quests in New Zealand

FloEdelmann:
* disable `AddPostboxRoyalCypher`
* enable `AddBusStopRef`

See https://wiki.openstreetmap.org/wiki/Talk:StreetComplete/Quests#Quests_in_New_Zealand for more context. Note that I put a permalink to that page version in the code comments.


westnordost:
Cool, thanks!

PR #5280, opened 2023-09-27
Pass `getCountryInfoByLocation` and `getCountryCodeByLocation` to quests/modules

FloEdelmann:
Follow-up to #5273. This PR implements similar changes for `CountryBoundariesFuture`. Thus, more implementation detail is abstracted from the quests/overlays and the Java dependency is reduced a bit.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/OverlaysModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/OverlaysModule.kt
@@ -3,7 +3,10 @@ package de.westnordost.streetcomplete.overlays
 import de.westnordost.countryboundaries.CountryBoundaries
 import de.westnordost.osmfeatures.Feature
 import de.westnordost.osmfeatures.FeatureDictionary
+import de.westnordost.streetcomplete.data.meta.CountryInfo
 import de.westnordost.streetcomplete.data.meta.CountryInfos
+import de.westnordost.streetcomplete.data.meta.getByLocation
+import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.data.overlays.OverlayRegistry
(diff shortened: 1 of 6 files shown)


westnordost:
Nice! But what is actually needed? Is the whole `CountryBoundaries` needed, or just a function that returns in which country a particular position is? Or even (better), a function that returns the `CountryInfos` for a particular position?

westnordost:
By the way, the reason why I deem these changes useful is basically:

1. keeps interface (of dependencies) small
2. removes implementation-specifics and Java-specifics from interface (Futures, ...)

(roughly https://en.wikipedia.org/wiki/Interface_segregation_principle  )

FloEdelmann:
* In `AddressOverlay`: get country code of a specific position
* In `CyclewayOverlay`: get `CountryInfo` for a specific position
* In `AddCycleway`: get `CountryInfo` for a specific position
* In `AddRoofShape`: get `CountryInfo` for a specific position

The country code is also included in the `CountryInfo` object, so we could rewrite the `AddressOverlay` to a `getCountryInfoByLocation` function, but it would be a little inefficient (create `CountryInfo` object from country code, only to access its country code property). But it would make the interfaces clean and consistent again.

westnordost:
Hmm, why do you think that the interfaces need to be consistent when the different classes have different needs (dependencies)?

FloEdelmann:
Hmm, indeed, I guess I just wanted to keep the number of different parameter types low. I'll implement one `getCountryInfoByLocation` and one `getCountryCodeByLocation` then.

westnordost (approved):
Very fine!

Looking at `AddRoofShape`, I guess it would even be better to not even have the whole of CountryInfos on the interface, but well, can always be done later, too.

If you like to look more into some small clean-up tasks, I noticed yesterday that this mcc resources hack is still used at a few places. I'd like to get rid of that, instead have it in code. I think you already did that for one quest, but it would be nice to get rid of it completely.

PR #5284, opened 2023-10-01
Enable searching by English quest titles

FloEdelmann:
This might make it easier to find quests where you don't know the exact translation. E.g. the [quest list in the OSM Wiki](https://wiki.openstreetmap.org/wiki/StreetComplete/Quests) is only in English.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
@@ -137,7 +137,7 @@ class QuestSelectionAdapter(
         }
 
         override fun onQuestTypeOrdersChanged() {
-            // all/many quest orders have been changed - reinit list
+            // all/many quest orders have been changed - re-init list
             viewLifecycleScope.launch { questTypes = createQuestTypeVisibilityList() }
         }
     }
@@ -319,7 +319,7 @@ class QuestSelectionAdapter(
                 AlertDialog.Builder(compoundButton.context)
                     .setTitle(R.string.enable_quest_confirmation_title)
                     .setMessage(item.questType.defaultDisabledMessage)
-                    .setPositiveButton(android.R.string.ok) { _, _ -> applyChecked(b) }
+                    .setPositiveButton(android.R.string.ok) { _, _ -> applyChecked(true) }
                     .setNegativeButton(android.R.string.cancel) { _, _ -> compoundButton.isChecked = false }
                     .setOnCancelListener { compoundButton.isChecked = false }
                     .show()
--- a/app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
@@ -2,6 +2,7 @@ package de.westnordost.streetcomplete.screens.settings.questselection
 
 import android.content.Context
 import android.content.SharedPreferences
+import android.content.res.Configuration
 import android.view.LayoutInflater
 import android.view.MotionEvent
 import android.view.ViewGroup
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
```diff
@@ -80,15 +89,22 @@ class QuestSelectionAdapter(
...
 
+    private fun questTypeMatchesSearchWords(questType: QuestType, words: List<String>): Boolean {
+        val question = genericQuestTitle(context.resources, questType).lowercase()
+        if (question.containsAll(words)) {
+            return true
+        }
+
+        val englishQuestion = genericQuestTitle(englishResources, questType).lowercase()
+        return englishQuestion.containsAll(words)
+    }
```
westnordost, line 100:
```suggestion
    private fun questTypeMatchesSearchWords(questType: QuestType, words: List<String>) =
        genericQuestTitle(context.resources, questType).lowercase().containsAll(words)
        || genericQuestTitle(englishResources, questType).lowercase().containsAll(words)
```

I think this is well readable, too

app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
```diff
@@ -303,7 +319,7 @@ class QuestSelectionAdapter(
                 AlertDialog.Builder(compoundButton.context)
                     .setTitle(R.string.enable_quest_confirmation_title)
                     .setMessage(item.questType.defaultDisabledMessage)
-                    .setPositiveButton(android.R.string.ok) { _, _ -> applyChecked(b) }
+                    .setPositiveButton(android.R.string.ok) { _, _ -> applyChecked(true) }
```
westnordost, line 322: What's this?

westnordost (commented):
Yeah, makes sense! Two comments:

app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
FloEdelmann: Yes, that's actually much nicer than my version!

app/src/main/java/de/westnordost/streetcomplete/screens/settings/questselection/QuestSelectionAdapter.kt
FloEdelmann: That was a minor IDE lint warning that I fixed while editing the file: The variable `b` is guaranteed to be true here, since we explicitly check for `!b` a few lines above.

matkoniecz: I was never convinced by such lint rule - that is asking for bugs if b will stop being checked, while for now is not really causing any problems

FloEdelmann: Well, if we change the condition above, we have to change the "ok" and "cancel" button behavior anyway, so I don't think we have to prepare for that. And `applyCheck(true)` is much easier to understand than `applyCheck(!b)` (What is the meaning of `b`? And what is it's current value when clicking the button?).

matkoniecz: I guess that it is matter of preferences, and hard to prove superiority...

Note that it is `applyChecked(b)` not `applyChecked(!b)`

also, it makes it symmetric to `applyChecked(b)` above

westnordost: Could you revert it to as it was before?

FloEdelmann: Restored in [`0d90dfb`](https://github.com/streetcomplete/StreetComplete/pull/5284/commits/0d90dfbc279b4b7a8a055b88b872b68e4d861b19).

PR #5285, opened 2023-10-01
Improve quest preset UX

FloEdelmann:
* Add hint to preset name input field when renaming or duplicating preset
* Auto-focus preset name input field when opening the dialog
* Automatically select newly added/duplicated preset


westnordost:
Seems to make sense, thanks! Code looks fine.

PR #5298, opened 2023-10-10
Drop MCC hack for layouts

FloEdelmann:
As suggested in https://github.com/streetcomplete/StreetComplete/pull/5280#pullrequestreview-1663594560, this PR removes the mobile country code hack that provides different layouts depending on the country the quest is located in. A similar change has been made for the fire hydrant quest in #4369.
Now, no more `layouts-mccXXX` folders are left :tada: 

Still left to remove in a future PR:

* The `traffic_sign_background` color (used in vector drawables) is still localized for `SE`, `FI` and `IS` via `values-mccXXX/colors.xml` files.
* The `vibrating_button_illustration` drawable image (used in `quest_traffic_lights_vibration` layout) is still localized for `GB` and `AU` via `drawable-mccXXX-YYYdpi` folders.
* The MCC logic is still included in `AbstractQuestForm`.

I'm not sure how to approach the color and drawable localizations. @westnordost do you have a suggestion?


westnordost (approved):
Yeah, very nice!

westnordost:
> I'm not sure how to approach the color and drawable localizations. @westnordost do you have a suggestion?

The `Drawable`s are a bit of a black box in code, i.e. I don't think it would be feasible (not sure if possible) to modify their properties (i.e. some color) at runtime.

So, the `traffic_sign_background` would need to be removed from these drawables and instead new drawables like `background_maxspeed_sign_yellow.xml` need to be created. Which drawables are used for the layout then needs to be set prammatically.

matkoniecz:
Thanks! I remember how confused I was by this mechanism (and still not entirely sure how it worked)

PR #5299, opened 2023-10-11
Fix failing unit tests build

neonowy:
Was interested in contributing to the project, but stumbled upon compiler errors after trying to run `testDebugUnitTest`.
Seems like some tests were not updated in #5280.

BTW, would you be interested in a separate PR for running test job for every commit on `master`? Seems like the existing test job definition is outdated/not used.

### DIFF
--- a/app/src/test/java/de/westnordost/streetcomplete/QuestsOverpassPrinter.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/QuestsOverpassPrinter.kt
@@ -8,7 +8,7 @@ import de.westnordost.streetcomplete.testutils.mock
 
 fun main() {
 
-    val registry = questTypeRegistry(mock(), mock(), mock(), mock(), mock(), mock())
+    val registry = questTypeRegistry(mock(), mock(), mock(), mock(), mock())
 
     for (questType in registry) {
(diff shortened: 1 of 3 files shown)


FloEdelmann:
Oh thanks, I indeed forgot to check the tests!

riQQ:
> BTW, would you be interested in a separate PR for running test job for every commit on `master`? Seems like the existing test job definition is outdated/not used.

See #2522 :(

matkoniecz:
> See https://github.com/streetcomplete/StreetComplete/issues/2522 :(

It seems that

> Also, currently some Android tests always fail, see https://github.com/streetcomplete/StreetComplete/issues/2028 so the CI would just report for everything that tests fail. That's not useful. It would be useful to do something about https://github.com/streetcomplete/StreetComplete/issues/2028 or its successor-ticket.

is no longer applying

riQQ:
But you left out the first line of westnordost's post:
> I don't see the necessity for this. I do run Android lint, the tests etc before each release.

matkoniecz:
Maybe that was overintepretation... I initially thought that it was expressing that it is not strictly necessary AND blocked.

Though "I don't see the necessity for this" can also mean that it is simply unwanted at all.

PR #5301, opened 2023-10-11
Drop MCC hack completely

FloEdelmann:
Follow-up to #5298. Please review each commit individually, see the respective commit messages.


westnordost (changes requested):
You're fast...

Regarding the `R.layout...._sign_fi`, I'd very much prefer if the background color was just set programmatically instead of duplicating the whole layout.

westnordost:
Because:
- duplicating layouts in general is problematic because it is a big source of error: The layouts need to be identical as far as their IDs are concerned after all for the code to handle it correctly.
- it is a lot of duplicated code which could be avoided, i.e. it needs to be maintained twice
- it is not clear why it was duplicated, so later viewers of the code may assume that it is a completely distinct layout such as the US ones are

westnordost (approved):
Nice!! Code looks good to me. Did you test it? (You can test it by scrolling to e.g. Finland on the map, then going into the settings and opening the respective form

FloEdelmann:
For reference, the last two commits were reverted in ebe7dfce8418481ef90e7a065e679fef3cbd5127 to fix #5384.

PR #5306, opened 2023-10-13
Update CONTRIBUTING_A_NEW_QUEST.md

qugebert:
Just few minor edits of CONTRIBUTING_A_NEW_QUEST.md 

- AddIsDefibrillatorIndoor was deleted, so i choosed another quest as example for simple yes/no quest
- There was another reference to AddIsDefibrillatorIndoor where choosing a different quest would make not so much sense, so i set a permalink to this file in the StreetComplete_quest_creation_tutorial
- Editing PinIcons.kt for adding new amenitys is not neccessary any more, so no reason to distract people by mentioning this.


matkoniecz:
This another AED reference also should be somehow edited out, but this is already a clear improvement. Thanks!

PR #5308, opened 2023-10-13
Added quest for mtb:scale

### LINKED ISSUE #1850: Ask for MTB difficulty for paths
### General
Affected tag(s) to be modified/added: [mtb:scale](https://wiki.openstreetmap.org/wiki/Key:mtb:scale)
Question asked: **What is the difficulty level for mountain biking on this path?**

### Checklist

### Ideas for implementation
**Element selection:** 
Select only highway=path that don't have surface=paved|asphalt or similar, and that does not have the tag mtb:scale.

**Proposed GUI:**
Image examples for each value, preferably with a split image for each value to display two different kinds of paths to be tagged with that value (like half the image from a mountain area with stones and another half from a forest area where obstacles are more like roots etc.

Values to display together with the image can be same as for iD:
0: Solid gravel/packed earth, no obstacles, wide curves
1: Some loose surface, small obstacles, wide curves
2: Much loose surface, large obstacles, easy hairpins
3: Slippery surface, large obstacles, tight hairpins
4: Loose surface or boulders, dangerous hairpins
5: Maximum difficulty, boulder fields, landslides
6: Not rideable except by the very best mountain bikers

There also needs to be an alternative for **Not Applicable**

mtb:scale is very valuable to determine the difficulty of a path, not only for MTB, but also for trail running and hiking and thus very useful for data consumers using this information for route planning.

torhovland:
Fixes #1850.

The icon can be improved. Also, the photos for value 2-5 are low resolution and had to be upscaled. Better photos would be good.

image


matkoniecz:
Looks nice! Though I would be unable to distinguish "some trial techniques" and "very few can ride here"

torhovland:
Yeah, this is based on the [Wiki categorisation](https://wiki.openstreetmap.org/wiki/Key:mtb:scale), but obviously shortened to fit. I am happy to adjust the text and photos if somebody has any suggestions.

Having said that, I think it is fine as it is. There are quite a few riders who have "some trials technique", i.e. enough to lift the rear wheel through a tight turn, while the next category indicates that even this is not enough.

westnordost (changes requested):
Heya, I wrote a longer comment in the ticket. In a nutshell, I think it should be implemented as an overlay, instead.

From a cursory look, from the technical perspective it looks very fine. A few comments, though:

- The pictures for scale 2-5 are super-blurry, I'd like to very much encourage you to find pictures with at least a resolution of 384x384 pixels. So, these miniature-pictures are from the wiki(?) - maybe update the wiki with proper pictures then, too? It can only help to improve the overall quality of the data collected - with or without StreetComplete. 
  Current pictures: As a layman, the jump between scale 1 and scale 2 seems really large to me (the loose rock on scale 2 picture looks dangerous,.... if he would go in the gravel in-between, I'd understand). Also, I do not see a difference between scale 3 and 4. The jump between scale 5 and 6 is probably deliberate though(?)

- translations are done exclusively in POEditor, strings in `values-XX` will be overwritten by an import from POEditor

- I am not content with the descriptions. "advanced riding skills", "very good skills",... etc. are absolutely subjective. The descriptions should not describe how advanced you gotta be (on a scala from 0-6) but should try its best to nail down some things the user can see, with his eyes. See also https://github.com/streetcomplete/StreetComplete/issues/1850#issuecomment-632113430 .

- Note that if the descriptions turns out to be longer, and it is already quite long, I suggest to use the list style as used by the smoothness quest, where the pictures are on the left and the description is on the right

- The icon is much too detailed! You know how small of an icon it is displayed as. Lines should not be thinner than 4px (when assuming a size of 128x128px). Suggestion: Use the bicycle icon `bicycle.svg` and the bump from `car_bumpy.svg`, rotate the bicycle a bit and/or make the bump look more like a rock, done. Also, I think a green background is better here - it is about bicycles, yes, but it is about mountainbiking which is a sport, so the "leisure/amenity" color seems to be more fitting here. Alternatively, leave the icon to me and just use `bicycle.svg` as a placeholder.

mnalis:
Overlay might indeed be  only workable solution. E.g. this [overpass query](http://overpass-turbo.eu/s/1CyO) shows what current query does in the middle of the urban environment (where there ideally should be _zero_ quests about `mtb:scale=*`):

Screenshot 2023-10-28 at 12-29-38 overpass turbo

Upon taking a closer look at random ways selected, I do not think that refining the query for Quest might improve situation significantly. Requiring the [incline=*](https://wiki.openstreetmap.org/wiki/Key:incline) would help with reducing the spurious quests in citires, but then it would miss flat parts of MTB paths in woods too, and many inclined ways do not have `incline=*` tagged anyway. 
And I [do not think](https://github.com/streetcomplete/StreetComplete/issues/1850#issuecomment-1760350326) that StreetComplete can recognize and automatically deselect urban areas currently (e.g. `landuse=residental` etc)?

Thus, converting this to Overlay seems to me like best solution

westnordost:
No feedback since 5+ months, closing.

This would need to be re-implemented as an overlay anyway, there is little which could be borrowed from the current state of this PR. So, for an overlay, a new PR should be created.

PR #5314, opened 2023-10-15
Allow searching in debug "Show quest forms" view

### LINKED ISSUE #3229: Make quest findable in debug menu "Show quest form"
**Use case**
Make easier to debug specific quest without scrolling for lightyears

**Proposed Solution**
The same code as for main quest list would be hopefully applicable

FloEdelmann:
The more quests we add, the harder it gets to find the quest in the debug "Show quest forms" list. This PR adds a simple search filter, just like we have in the quest selection view. It also uses basically the same implementation as in the quest selection view.

Fixes #3229.


mnalis:
- Thanks @FloEdelmann, this should actually implement https://github.com/streetcomplete/StreetComplete/issues/3229 which I missed greatly!

westnordost (approved):
LGTM

westnordost:
Also, thank you! Happened to me often enough that I spent too much time looking through that list to find the right quest type I wanted to test...

Issue #2146, opened 2020-10-08
New quest: defibrillator:location


Cj-Malone:
### General
Affected tag(s) to be modified/added: [defibrillator:location](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Ddefibrillator)
Question asked: **Where is this defib located?**

*On the outside wall of x opposite y*

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [ ] 🐿️ Easily answerable by everyone from the outside but a survey is necessary
 - You may need to filter by `indoor=no` if you want to limit this to outside surveys only, but there is still value to asking this question inside.
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

**Proposed GUI:**
I guess the UI is just a text area with a max length, and some examples of good descriptions.

westnordost:
Hmm, I think this would be quite problematic because of the free text input. Don't you think that adding this info through this app could negatively affect the data quality of this tag? That the data quality of this tag is good is very important - better not set at all than not accurate.

westnordost:
@matkoniecz did we talk about this at the start of September?

Cj-Malone:
> Don't you think that adding this info through this app could negatively affect the data quality of this tag?

I think any issues that could be caused by people adding it through StreetComplete are equally as plausible to be caused by somebody mapping with JOSM.

I guess it comes down to a QA problem, maybe StreetComplete should reject certain values, like under x chars. And we should have another QA tool monitoring OSM data to ensure this tag has quality data in it.

> That the data quality of this tag is good is very important - better not set at all than not accurate.

The inverse is also true, a defib with no context (other than cords) could be worse than not having an accurate description of it's location, especially in areas that aren't as well mapped.

matkoniecz:
> @matkoniecz did we talk about this at the start of September?

No, it has not appeared in discussion (I looked at this tag, it seemed to not be obviously fitting SC so I have not proposed this)

qugebert:
> maybe StreetComplete should reject certain values, like under x chars.

What do you think would be an acceptable minimum of chars for a useful description?

Cj-Malone:
[...]

PR #5328, opened 2023-10-21
Defibrillator Location (Issue #2146)

qugebert:
This Issue #2146 

Minimum description length should be discussed.

Screenshot_20231021-153210_1

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
@@ -9,7 +9,7 @@ import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
 import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.LIFESAVER
 import de.westnordost.streetcomplete.osm.Tags
 
-class AddDefibrillatorLocation : OsmFilterQuestType<DefibrillatorLocation>() {
+class AddDefibrillatorLocation : OsmFilterQuestType<String>() {
 
     override val elementFilter = """
         nodes with
@@ -30,9 +30,8 @@ class AddDefibrillatorLocation : OsmFilterQuestType<DefibrillatorLocation>() {
 
     override fun createForm() = AddDefibrillatorLocationForm()
 
-    override fun applyAnswerTo(answer: DefibrillatorLocation, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
-        tags["defibrillator:location"] = answer.location
+    override fun applyAnswerTo(answer: String, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        tags["defibrillator:location"] = answer
     }
 }
 
-data class DefibrillatorLocation(val location: String)
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
@@ -14,7 +14,8 @@ class AddDefibrillatorLocation : OsmFilterQuestType<DefibrillatorLocation>() {
     override val elementFilter = """
         nodes with
         emergency = defibrillator
-        and !location and !defibrillator:location"
+        and !location and !defibrillator:location
+        and access !~ private|no"
     """
     override val changesetComment = "Specify defibrillator location"
     override val wikiLink = "Tag:emergency=defibrillator"
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
@@ -1,0 +1,37 @@
+package de.westnordost.streetcomplete.quests.defibrillator
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.LIFESAVER
+import de.westnordost.streetcomplete.osm.Tags
+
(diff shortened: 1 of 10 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocationForm.kt
```diff
@@ -0,0 +1,28 @@
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding.locationInput.doAfterTextChanged { checkIsFormComplete() }
+    }
+
+    override fun onClickOk() {
+        applyAnswer(DefibrillatorLocation(location!!))
+    }
+
+    override fun isFormComplete() = location != null && location!!.length >= 10
```
qugebert, line 27: Minimum description length should be discussed.

app/src/main/res/values/strings.xml
```diff
@@ -1058,6 +1058,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_internet_access_no">No connection</string>
 
     <string name="quest_is_amenity_inside_title">Is this inside a building?</string>
+    <string name="quest_defibrillator_location">Where is this defibrillator located?</string>
+    <string name="quest_defibrillator_location_description">Please give a short but precise description of the position of this defibrillator (e.g. "in the porter's lounge")</string>
```
westnordost, line 1062: "Short but precise" would be "concise", to my understanding.

Also, "of... of" does not sound so good.

```suggestion
    <string name="quest_defibrillator_location_description">Please give a concise description of this defibrillator's position (e.g. "in the porter's lounge")</string>
```

Also, you are kind of duplicating the question here. It is already clear that this is about the defi location.

Plus, it is already displayed in the hint that this is about a defi.

app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
```diff
@@ -0,0 +1,38 @@
...
+        getMapData().filter("nodes with emergency = defibrillator")
+
+    override fun createForm() = AddDefibrillatorLocationForm()
+
+    override fun applyAnswerTo(answer: DefibrillatorLocation, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        tags["defibrillator:location"] = answer.location
+    }
+}
+
+data class DefibrillatorLocation(val location: String)
```
westnordost, line 38: Any reason why this is not just a string?

app/src/main/res/values/strings.xml
qugebert: Sorry, my foreign language skills are not soo good, i still didn't know 'concise' but yes, this sounds better.

I'm not sure about extending to other amenities, as they use 'location', not 'defibrillator:location'

westnordost: > I'm not sure about extending to other amenities, as they use 'location', not 'defibrillator:location'

Which ones, for example?

app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocationForm.kt
westnordost: And maximum length. Allowed in OSM are only 255 unicode character points

app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocation.kt
qugebert: Not really. I see, this might work with simple String.

app/src/main/res/values/strings.xml
qugebert: Ok, after some quick research:
[location](https://wiki.openstreetmap.org/wiki/Key:location) is not used for free text input at all, only some *:location-tags.
[taginfo search](https://taginfo.openstreetmap.org/search?q=%3Alocation#keys) for :location only lists a few common :location tags
42 236 [defibrillator:location](https://taginfo.openstreetmap.org/keys/defibrillator%3Alocation)
5 402 [fire_hydrant:location](https://taginfo.openstreetmap.org/keys/fire_hydrant%3Alocation)
2 692 [generator:location](https://taginfo.openstreetmap.org/keys/generator%3Alocation)
2 397 [changing_table:location](https://taginfo.openstreetmap.org/keys/changing_table%3Alocation)
1 227 [post_box:location](https://taginfo.openstreetmap.org/keys/post_box%3Alocation)
389 [description:location](https://taginfo.openstreetmap.org/keys/description%3Alocation)

with multiple variants of defibrillator:location:[language-code]

app/src/main/java/de/westnordost/streetcomplete/quests/defibrillator/AddDefibrillatorLocationForm.kt
qugebert: Maximum length is limit by android:maxLength="253"

app/src/main/res/values/strings.xml
westnordost: I see, to defibrillator is the only place where this kind of tag is used.

app/src/main/res/layout/quest_location.xml
```diff
@@ -0,0 +1,22 @@
...
+    android:labelFor="@+id/commentInput"/>
+
+<EditText xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/locationInput"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:inputType="text"
+    android:maxLength="253"
+    />
```
westnordost, line 21: Wouldn't it make sense to make it a multiline edit text?

westnordost (commented):
Looks good, just one small suggestions

app/src/main/res/layout/quest_location.xml
qugebert: Sure, multiline looks better, i changed that now.

Screenshot_20231021-214557_1

westnordost (approved):

app/src/main/res/layout/quest_location.xml
westnordost: I notice that you put "in the porter's lounge" but these do not show up in the string in the app.

To fix that, I think what should work is to put the entire string in "..." and escape the \"in the porter's lounge\"

qugebert: Oops, i didn't notice the missing quotation marks. 

> To fix that, I think what should work is to put the entire string in "..." and escape the \"in the porter's lounge\"

Thank you, this worked fine.
Screenshot_20231022-000318(1)

westnordost:
Perfect

PR #5331, opened 2023-10-25
Add santiary dump station fee quest (Needs Icon)

catdogmat:
This quest applies to ~50% of all sanitary dump stations on OSM. This quest will ask "Does this Sanitary Dump Station require a fee?"

This PR needs a icon.


qugebert:
why not just extend the general_fee quest?

catdogmat:
thanks for the heads up

HolgerJeromin:
New PR in #5332

catdogmat:
Reopen as people said this is a better solution

FloEdelmann:
see https://github.com/streetcomplete/StreetComplete/pull/5332#issuecomment-1782643189

PR #5332, opened 2023-10-25
Extend General fee with Sanitary Dump Stations

catdogmat:
This quest applies to ~50% of all sanitary dump stations on OSM. I extended the General Fee Quest, and added Sanitary Dump Stations.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
@@ -23,7 +23,11 @@ class AddGeneralFee : OsmFilterQuestType<Boolean>() {
     override val icon = R.drawable.ic_quest_fee
     override val achievements = listOf(CITIZEN)
 
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_generalFee_title2
+    override fun getTitle(tags: Map<String, String>) =
+     if (tags["amenity"] == "sanitary_dump_station")
+         R.string.quest_generalFee_title
+     else
+         R.string.quest_generalFee_title2
 
     override fun createForm() = YesNoQuestForm()
 
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
@@ -12,7 +12,7 @@ class AddGeneralFee : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
         nodes, ways with
-         (tourism = museum or leisure = beach_resort or tourism = gallery or tourism = caravan_site)
+         (tourism = museum or leisure = beach_resort or tourism = gallery or tourism = caravan_site or amenity = sanitary_dump_station)
          and access !~ private|no
          and !fee
     """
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt
```diff
@@ -12,7 +12,7 @@ class AddGeneralFee : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
         nodes, ways with
-         (tourism = museum or leisure = beach_resort or tourism = gallery or tourism = caravan_site)
+         (tourism = museum or leisure = beach_resort or tourism = gallery or tourism = caravan_site or amenity = sanitary_dump_station)
```
FloEdelmann, line 15:
```suggestion
         tourism ~ museum|gallery|caravan_site
         or leisure = beach_resort
         or amenity = sanitary_dump_station
```

catdogmat:
I think this is ready to merge.

westnordost (changes requested):
Hmm, but the question posed is "Do you need to pay to enter here?"

The wording doesn't really fit for sanitary dump stations. Can a wording be found that would fit all these places?

catdogmat:
How about "Is there an entrance fee or cost associated with this location?"

Helium314:
Couldn't there just be a separate title for sanitary dump stations, or maybe generalized to places / things one does not (want to) enter?

westnordost:
Yes, that would be possible. I would also prefer such a solution. The new string could then say something like whether it costs something to use it.

catdogmat:
Ok sounds good, closing then.

westnordost:
Hu, why? I think there has been a misunderstanding. The PR is fine, but we'd need a second question text.

Notice how the `getTitle` function in line 26 has the tags as a parameter? You could check whether it is a sanitary dump station and in that case instead return a different string, something like "Do you ahve to pay ... to **use**...?"

qugebert:
@catdogmat this should work
   ```
        if (tags["amenity"] == "sanitary_dump_station")
            R.string.quest_generalFee_title
        else
            R.string.quest_generalFee_title2
```
and add a second title to app/src/main/res/values/strings.xml

**Comment by catdogmat** [comment]
Alright, I'll add this now

**Comment by catdogmat** [comment]
Thanks, @qugebert i've added it

**Comment by catdogmat** [comment]
Alright, @westnordost I think this is ready for a merge!

File: app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt [hunk 2]
```diff
@@ -12,7 +12,9 @@ class AddGeneralFee : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
         nodes, ways with
-         (tourism = museum or leisure = beach_resort or tourism = gallery or tourism = caravan_site)
+         tourism ~ museum|gallery|caravan_site
+         or leisure = beach_resort
+         or amenity = sanitary_dump_station
```
westnordost, line 17:
```suggestion
         (
           tourism ~ museum|gallery|caravan_site
           or leisure = beach_resort
           or amenity = sanitary_dump_station
         )
```

missing brackets

westnordost (changes requested):

catdogmat:
Ok, @westnordost I fixed the brackets

westnordost (approved):
Thanks, looks fine now!

catdogmat:
I think this is ready to merge, @westnordost

catdogmat:
Oh wait does @FloEdelmann need to review first

FloEdelmann (approved):
I'm fine with the changes, too :slightly_smiling_face:

westnordost:
It will be merged for the next main release (not for a bugfix release, which I expect the next release to be)

PR #5335, opened 2023-10-26
Add in-app log viewer

### LINKED ISSUE #3597: Add in-app log viewer
**Use case**
To help debug issues, in-app log viewer would be useful.

**Proposed Solution**
It has been mentioned several times by @westnordost, and that @matkoniecz might want to do it and that the ticket for it would be created, but is seems to have slipped, so I'm opening this one.

See for example https://github.com/streetcomplete/StreetComplete/issues/3290#issuecomment-946175907, https://github.com/streetcomplete/StreetComplete/issues/1892#issuecomment-848214943, https://github.com/streetcomplete/StreetComplete/issues/3406#issuecomment-949252071

neonowy:
Closes #3597.

Feel free to point out how I can improve the UI here, as I don't use Android day-to-day and I'm not familiar with its UI guidelines etc.

## Summary of changes:
- Three implementations of the new `Logger` interface:
  - `AndroidLogger` for writing messages to Android logs (visible in Logcat)
  - `DatabaseLogger` for writing message to the database
  - `Log` object that replaces all the existing `android.util.Log` calls
- New link in the About screen to see the list of log messages:
  - Every row shows message colored by log level and message timestamp (without the date part)
  - Filter by log level, message content (contains), timestamp (newer/older than)
  - Share button for exporting (filtered) log messages to clipboard/email/etc
- `Cleaner` now removes logs older than 3 days from the database (set via `DELETE_OLD_LOG_AFTER` constant)
- `CrashReportExceptionHandler` attaches the last 100 lines of logs to the email crash report (set via `MAX_LOG_LINES_TO_ATTACH_TO_CRASH_REPORT` constant)

## Screenshots:
| Light theme | Dark theme |
|--------|--------|
| Screenshot_20231026_204356 | Screenshot_20231026_204353 |

## Demo:
https://github.com/streetcomplete/StreetComplete/assets/5042328/2401333c-19ec-4481-9ba5-dddc5afa62e0

## Sample log copied via Share button:
```
2023-10-26T20:57:52.887: [QuestAutoSyncer] All downloaded in radius of 284 meters around user (6 tiles)
2023-10-26T20:57:52.789: [MapDataController] Fetched 5857 elements and geometries in 2907ms
2023-10-26T20:57:51.943: [Preloader] Preloading data took 8.6s
2023-10-26T20:57:51.940: [Preloader] Loaded features dictionary in 8.6s
2023-10-26T20:57:49.866: [QuestAutoSyncer] Checking whether to automatically download new quests at 52.2296750,21.0122283
2023-10-26T20:57:48.196: [Preloader] Loaded country boundaries in 4.8s
```


westnordost:
Wow, awesome! We could clone you a few times and we'll have a full iOS version in less than a year! 😆

So, it is really cool that you implemented this logging wrapper and it is even cooler that you even implemented this whole log viewer thing! @Helium314 , could it be that there is also a log viewer in SCEE? If so, would you say that if this is merged to SC it would be sufficient for SCEE too, or would you not replace it because something is missing from that which you deem important?

Before I review it, the main use case for the log viewer at all is to help find bugs. So in this respect, it would be very handy to have something in the UI that facilitates making use of this. For example

- automatically attaching the last log lines from before a crash to a crash report or
- being able to easily find and copy a range of logs, even if the log is from days (or weeks??) ago because a developer might ask for the logs (days, weeks) after a bug ticket was posted.

I haven't looked at the PR yet and what you did or didn't implement, I just wanted to first explain for what the feature would (from my point of view) be used and thus what would be the primary use case for this. The use case dictates how the UI should work.

app/src/main/java/de/westnordost/streetcomplete/util/CrashReportExceptionHandler.kt
```diff
@@ -109,4 +118,9 @@ class CrashReportExceptionHandler(
     private fun deleteCrashReport() {
         appCtx.deleteFile(crashReportFile)
     }
+
+    private fun readLogFromDatabase(): String = logsController
+        .getLogs(LogsFilters())
+        .take(ApplicationConstants.MAX_LOG_LINES_TO_ATTACH_TO_CRASH_REPORT)
```
westnordost, line 124: As noted before, I think it makes more sense to limit them by time rather than total number. Or at least, *additionally* by time.

app/src/androidTest/java/de/westnordost/streetcomplete/data/logs/LogsDaoTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    @Test fun getAll() {
+        val m1 = createMessage("1", timestamp = 1)
+        val m2 = createMessage("2", timestamp = 100)
+        val m3 = createMessage("3", timestamp = 10)
+
+        listOf(m1, m2, m3).forEach { dao.add(it) }
+
+        // sorted by timestamp descending
+        assertEquals(listOf(m2, m3, m1), dao.getAll())
+    }
```
westnordost, line 25: How about this?

```suggestion
    @Test fun getAll_sorts_by_timestamp_descending() {
        val m1 = createMessage("1", timestamp = 1)
        val m2 = createMessage("2", timestamp = 100)
        val m3 = createMessage("3", timestamp = 10)

        listOf(m1, m2, m3).forEach { dao.add(it) }

        assertEquals(listOf(m2, m3, m1), dao.getAll())
    }
```
(etc.)

I think the tests are not written completely consistent, but most tests have as names what behavior they test

app/src/main/java/de/westnordost/streetcomplete/util/logs/DatabaseLogger.kt
```diff
@@ -0,0 +1,56 @@
...
+import de.westnordost.streetcomplete.data.logs.LogLevel.*
+import de.westnordost.streetcomplete.data.logs.LogMessage
+import de.westnordost.streetcomplete.data.logs.LogsDao
+import de.westnordost.streetcomplete.util.ktx.nowAsEpochMilliseconds
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+
+class DatabaseLogger(private val logsDao: LogsDao) : Logger {
```
westnordost, line 12: It is quite consistently implemented in the data layer within StreetComplete that the only place a DAO is accessed from the outside is its associated `Controller` (if there is one). The reason is that if (in the future) there shall be a listener that observes updates on the log such as when a new log message is added - this listener would be on the `LogsController`. To be able to fire the even to its listeners, the `LogsController` must know about it whenever the log changes.*

So, for consistency, please use `LogsController` here as a dependency and implement the adding of a `LogMessage` also in `LogsController`.

* actually, for the UI it would make sense if there was such a listener - to have live updates of the log in the log view, no?

app/src/main/java/de/westnordost/streetcomplete/util/logs/DatabaseLogger.kt
```diff
@@ -0,0 +1,56 @@
...
+    private val coroutineScope = CoroutineScope(SupervisorJob() + CoroutineName("DatabaseLogger"))
```
westnordost, line 13:
```suggestion
    private val coroutineScope = CoroutineScope(SupervisorJob() + CoroutineName("DatabaseLogger") + Dispatchers.IO)
```

app/src/main/java/de/westnordost/streetcomplete/util/CrashReportExceptionHandler.kt
```diff
@@ -54,12 +59,16 @@ class CrashReportExceptionHandler(
     }
 
     private fun askUserToSendErrorReport(activityCtx: Activity, @StringRes titleResourceId: Int, error: String?) {
+        val logText = readLogFromDatabase()
```
westnordost, line 62: There is a slight problem with this implementation.

On crash, the crash report (stack trace, android version etc.) are all written to a file. Then, the next time the user opens the app, which might very well be a week after, he is asked to send the error. But by this time, the log may not actually exist in the database anymore.
Furthermore, (another small point,) these functions are executed on the UI thread.

So, how about this:
- Log is fetched from the database table and written into the crash report file at the time of the crash
- For `askUserToSendErrorReport(activityCtx: Activity, @StringRes titleResourceId: Int, e: Exception)`, either pass the log lines as parameter or retrieve them automatically in that function

Also, I note now that I look at the file, and it is not your fault, that the error reporting is inconsistent: A full text wth app version, device name etc. is attached on an uncaught exception, but when `askUserToSendErrorReport` is called from application code, it is not. Right? Or am I misreading something?

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsController.kt
```diff
@@ -0,0 +1,44 @@
...
+            Log.v(TAG, "Deleted $deletedCount old log messages")
+        }
+    }
+
+    companion object {
+        private const val TAG = "LogsController"
+    }
+}
+
+data class LogsFilters(
```
westnordost, line 32: This data class is curious. Why not pass the filter parameters directly? I cannot think of a reason why, other than that this data class is maybe used in the UI, a view model, if you will? In which case it might be slightly better to move this data class to the UI code.
It would be more consistent with the rest of the codebase.

westnordost (commented):
Here's my first review. I only reviewed the data part of the PR so far, i.e. nothing from the UI. Should give you enough to look into. I don't want to steal your thunder by letting you wait for long for a review.

First of all, very well done, I love how well you made this PR to be consistent with the code style and overall architecture of the app. The things I noted in this review are all quite minor, but anyway, here it is:

app/src/main/java/de/westnordost/streetcomplete/util/CrashReportExceptionHandler.kt
Helium314: > Right? Or am I misreading something?

Right. On errors during download or upload, `askUserToSendErrorReport` is called and results in a report without device details etc. (very recently noticed that in SCEE)

Helium314:
> @Helium314 , could it be that there is also a log viewer in SCEE? If so, would you say that if this is merged to SC it would be sufficient for SCEE too, or would you not replace it because something is missing from that which you deem important?

There is a simpler logger and viewer in SCEE. The logger just stores the most recent (up to 12k) lines in a list.
I did not want the logger to write to the database, as I was afraid it might affect performance or cause issues if ~200 lines pile up when downloading data / creating quests.
The SCEE viewer just reproduces the logcat output and is much simpler. Only thing it has over this PR is the reverse button that can sometimes be handy.

I plan to remove the SCEE logger, or maybe have an option to use the old list-logging instead of writing to database.

westnordost:
Reverse, you mean reverse the order of the log lines displayed in the list? (Why do you find it handy?)

mnalis:
> Reverse, you mean reverse the order of the log lines displayed in the list? (Why do you find it handy?)

Because one does not have to scroll 12000 lines to the bottom of the list in order to see what was _the error that just happened_? Instead (in reversed order) the latest error is first line shown, which is far more convenient.

westnordost:
Ah of course, but then you don't need a reverse button. The view should just by default show the latest log messages, just like an email app would.

mnalis:
> The view should just by default show the latest log messages, 

If you're **only interested in the very last message**, sure, that's what `reverse mode` is there for!

But if you want to **follow several (dozen?) messages** (to find causality), **regular chronological order** (i.e. oldest at top, newest at bottom, just as it would be if you were writing a journal) is much **more useful**. 
(Especially as multiple newlines inside one message will always be sorted top-to-bottom, it quickly gets hard to follow when messages themselves are in reversed i.e. bottom-to-top order.)

This is from general Latin European perspective; I don't know how it would work in languages which are regularly written left-to-right but bottom-to-top (or if any exist, or that matter?)

> just like an email app would.

<details>

<summary>I'm not sure what you mean here... (click to expand)</summary>

I guess the intended meaning there depends on what email app you use :smile: and I don't even know what popular solutions look today (either because I try hard to avoid google and its stuff like gmail; and I don't use microsoft platforms at all so have no idea how outlook express (or whatever is used there those days) looks like by default. When I do see/wait on people using them, it looks and feels oh-so-horribly-slow-and-inefficient so I don't pay much attention to details of their design).
 
In my preferred [mutt(1)](https://en.wikipedia.org/wiki/Mutt_(email_client)), I constantly switch between different sort modes (e.g. when reading e-mail threads from mailing lists and similar I highly prefer oldest message at top, with rest of them forking down each in their own sub-thread; each subthread sorted chronologically from oldest message to newest. Trying to read e-mail threads in plain newest-message-at-top order would be horrible experience, and I'd completely lose the thread who is replying to what - especially combined  with fact that many people don't properly trim&quote paragraphs to which they are replying to)

</details>

westnordost:
Mhm. 

Well, towards @neonowy , as I wrote, I didn't review the UI yet, but I would expect that the messages are either ordered by time ascending with by default showing the latest messages (unless you scroll up manually) or alternatively by time descending. In any case, unless you interact with the UI, the latest messages should be in view.

app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -27,6 +27,11 @@ object ApplicationConstants {
      *  database if not used anymore and have not been refreshed in the meantime  */
     const val DELETE_OLD_DATA_AFTER = 14L * 24 * 60 * 60 * 1000 // 14 days in ms
 
+    /** the duration after which logs will be deleted from the database */
+    const val DELETE_OLD_LOG_AFTER = 3L * 24 * 60 * 60 * 1000 // 3 days in ms
+
+    const val MAX_LOG_LINES_TO_ATTACH_TO_CRASH_REPORT = 100
+
```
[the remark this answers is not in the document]
neonowy: I went too strict here, 14 days sound like a good limit 👍 

Out of curiosity, I tried to estimate the storage required and looks like it should take around 1MB for every 10,000 messages.

<details><summary>Details if you're curious:</summary>
<p>

Started with some guesstimates based on data types:

```
level: max 7 chars x 1B = 7B
tag: ~20 chars x 1B = 20B
message: ~80chars x 1B = 80B
timestamp = max 8B

total = 115B
```
(This of course doesn't take into account overhead from SQLite internals etc.)

Running `sqlite3_analyzer` on a database with ~400 log messages gives similar results:

```
*** Table LOGS ****************************************************************

Percentage of total database......................   0.58%   
Number of entries................................. 437       
Bytes of storage consumed......................... 40960     
Bytes of payload.................................. 31759       77.5% 
Bytes of metadata................................. 2204         5.4% 
B-tree depth...................................... 2         
Average payload per entry......................... 72.68     
Average unused bytes per entry.................... 16.01     
Average metadata per entry........................ 5.04      
Average fanout.................................... 9.00      
Non-sequential pages.............................. 3           33.3% 
Maximum payload per entry......................... 113       
Entries that use overflow......................... 0            0.0% 
Index pages used.................................. 1         
Primary pages used................................ 9         
Overflow pages used............................... 0         
Total pages used.................................. 10        
Unused bytes on index pages....................... 4022        98.2% 
Unused bytes on primary pages..................... 2975         8.1% 
Unused bytes on overflow pages.................... 0         
Unused bytes on all pages......................... 6997        17.1% 
```

</p>
</details> 

Also agree with changing the log lines limit to be time-based. I think this log lines limit idea came from my misunderstanding of the `CrashReportExceptionHandler` class, which you explained in the later comment 😄

app/src/main/java/de/westnordost/streetcomplete/data/logs/LogLevel.kt
```diff
@@ -0,0 +1,20 @@
+package de.westnordost.streetcomplete.data.logs
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.logs.LogLevel.*
+
+enum class LogLevel {
```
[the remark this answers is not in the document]
neonowy: Went with the third option as it's the most consistent and least fragile in case of future changes.

app/src/main/java/de/westnordost/streetcomplete/data/logs/LogsTable.kt
```diff
@@ -0,0 +1,25 @@
...
+        const val LEVEL = "level"
+        const val TAG = "tag"
+        const val MESSAGE = "message"
+        const val ERROR = "error"
+        const val TIMESTAMP = "timestamp"
+    }
+
+    const val CREATE = """
+        CREATE TABLE $NAME (
+            ${Columns.ID} int PRIMARY KEY,
```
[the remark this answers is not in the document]
neonowy: Removed the primary key as suggested.

Regarding an additional index on the timestamp: My reasoning was that this table will be very much write-heavy with very sporadic reads (few users will sporadically use the log viewer to troubleshoot bugs). This means that slowing frequent writes that will affect every user might not be worth a slightly faster filtering in the log viewer. I have not run any benchmarks to prove this, though. If you think an extra overhead for writes is not an issue, then I can add the index.

app/src/main/java/de/westnordost/streetcomplete/util/CrashReportExceptionHandler.kt
neonowy: Solved both issues by extracting creation of a full report text to a separate `createErrorReport` function, so both uncaught and caught errors generate similar report with app version, logs, etc.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsController.kt
neonowy: You're correct that this is kind of a view model that I've used in the UI code.
Moved it alongside the UI code where it's used.

neonowy:
> (Especially as multiple newlines inside one message will always be sorted top-to-bottom, it quickly gets hard to follow when messages themselves are in reversed i.e. bottom-to-top order.)

This is a good point. I think in that case a reverse button like in SCEE can be useful. We could show the latest message on the top by default (as it does now) and allow the user to reverse the order with the reverse button in the menu.

app/src/main/java/de/westnordost/streetcomplete/data/logs/LogsTable.kt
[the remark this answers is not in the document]
westnordost: A sound reasoning. You are right about that it is not accessed frequently. On the other hand, accessing any range of logs within the table without such an index requires SQLite to iterate the whole table each time. The table could be quite large, as you've upped the time limit to 2 weeks now. But to be honest, I have no idea how many log lines to expect in two weeks of usage.
I also don't have an idea how big the performance overhead of an index would be on insertion. However, a primary key is also an index and (pretty sure) implemented in the same way. In other words, any table with a primary key (pretty much every table) will have the same overhead - so I expect that this is somewhat streamlined. Insertion into a B-Tree shouldn't be a particularly costly operation.

So, gut feeling, I'd go with an index. But I'll merge this PR either way, so without further information on insertion overhead vs access overhead, I leave this up to you.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFiltersDialog.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    private fun updateNewerThanInput() {
+        binding.newerThanClearButton.visibility = if (filters.timestampNewerThan == null) View.GONE else View.VISIBLE
+        binding.newerThanTextDate.setText(filters.timestampNewerThan?.toString() ?: "")
+    }
+
+    private fun updateOlderThanInput() {
+        binding.olderThanClearButton.visibility = if (filters.timestampOlderThan == null) View.GONE else View.VISIBLE
+        binding.olderThanTextDate.setText(filters.timestampOlderThan?.toString() ?: "")
```
westnordost, line 101: Timestamps should be properly formatted according to the user's locale. Err... not sure if `kotlinx.time` has something for that, after all, `Locale` is a Java-only class. For Java's old time-date classes, there was `SimpleDateFormat`, but maybe there is (at least) something for the new data-time classes?

westnordost (commented):
The UI code is also very well done!

I commented on some smaller things. I also noticed that some code on the logger interface could be reduced.

How does the UI behave when something is being logged while in the screen but you are scrolled down? When you are not scrolled to top, it shouldn't automatically reset to the top position because if the user is looking through the log, it shouldn't be interrupted by new logs appearing.

Actually, looking at some other logging views, the most usual behavior and layout seems to be:
- new log lines appear at the bottom (logs sorted by date ascending)
- log view is always scrolled to the bottom, where the new log lines appear, by default
- when the log view is not scrolled to the bottom, the scroll position stays where it is when new log lines appear
- filters only filter, i.e. do not change any of the above behavior
- no reverse button

app/src/main/java/de/westnordost/streetcomplete/util/logs/Logger.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun w(tag: String, message: String)
+
+    /** Send WARNING log message and exception */
+    fun w(tag: String, message: String, e: Throwable)
+
+    /** Send ERROR log message */
+    fun e(tag: String, message: String)
+
+    /** Send ERROR log message and exception */
+    fun e(tag: String, message: String, e: Throwable)
```
FloEdelmann, line 23: This builds upon @westnordost's [suggestion](https://github.com/streetcomplete/StreetComplete/pull/5335/files#r1376893189) but also uses KDoc's [links to elements](https://kotlinlang.org/docs/kotlin-doc.html#links-to-elements):

```suggestion
    /** Send VERBOSE log [message] */
    fun v(tag: String, message: String)

    /** Send DEBUG log [message] */
    fun d(tag: String, message: String)

    /** Send INFO log [message] */
    fun i(tag: String, message: String)

    /** Send WARNING log [message] with optional [exception] */
    fun w(tag: String, message: String, exception: Throwable? = null)

    /** Send ERROR log message with optional [exception] */
    fun e(tag: String, message: String, exception: Throwable? = null)
```

app/src/main/java/de/westnordost/streetcomplete/util/logs/DatabaseLogger.kt
```diff
@@ -0,0 +1,99 @@
...
+        coroutineScope.launch {
+            logsController.add(LogMessage(
+                level = ERROR,
+                tag = tag,
+                message = message,
+                error = e.toString(),
+                timestamp = nowAsEpochMilliseconds()
+            ))
+        }
+    }
```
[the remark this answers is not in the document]
neonowy: Nice! FYI I've simplified it even further by declaring default values only in the interface definition.

app/src/main/res/layout/row_logs_filters_checkbox.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
neonowy: Removed altogether, as I've used programmatically defined [`Chip`s](https://developer.android.com/reference/com/google/android/material/chip/Chip) instead of checkboxes.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFiltersDialog.kt
```diff
@@ -0,0 +1,163 @@
...
+                ) { hour, minute ->
+                    val dateTime = LocalDateTime(year, month, dayOfMonth, hour, minute)
+                    callback(dateTime)
+                }.show()
+            },
+            initialDateTime.year,
+            initialDateTime.monthNumber,
+            initialDateTime.dayOfMonth
+        ).show()
+    }
```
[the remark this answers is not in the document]
neonowy: Looks great, thanks for the tip!

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFiltersDialog.kt
neonowy: Unfortunately i18n seems to be out of scope for [`kotlinx-datetime`](https://github.com/Kotlin/kotlinx-datetime#design-overview):
> The library is based on the ISO 8601 international standard, other ways to represent dates and times are out of its scope. Internationalization (such as locale-specific month and day names) is out the scope, too.

I've used `java.time.format.DateTimeFormatter` for now.

Just to be clear, I've only changed formatting of date time filter inputs, but left time labels from the logs list as-is (ISO 8601 formatted). Regardless of user's locale, `HH:mm:ss.SSS` format seems to be the only one that fits limited space and is precise up to milliseconds, don't you think?

app/src/main/java/de/westnordost/streetcomplete/util/logs/Logger.kt
neonowy: I'm pretty new to Kotlin, so thanks for the tip! 😄

app/src/main/java/de/westnordost/streetcomplete/data/logs/LogsTable.kt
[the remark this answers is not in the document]
neonowy: Added an index after all. Might go back to benchmarking this after there's some real-world usage data (e.g. number of log messages generated per day), but right now it would all be premature optimization probably :)

neonowy:
Thanks for all the feedback! ❤️
 I think I've resolved/replied to all the reviews comments. I've also made two major changes to the UI:

1) Reversed the sorting order for the logs list (i.e. newest messages are now shown at the bottom of the screen). It's both easier to read multiline messages this way (as mentioned by @mnalis) and scroll view has no way of jumping up and down because of a new item added to the top of the list. It works exactly as @westnordost described the usual layout for similar UIs:
> - new log lines appear at the bottom (logs sorted by date ascending)
> - log view is always scrolled to the bottom, where the new log lines appear, by default
> - when the log view is not scrolled to the bottom, the scroll position stays where it is when new log lines appear
> - filters only filter, i.e. do not change any of the above behavior
> - no reverse button

2) Redesigned filters dialog a little, by adding [`TextInputLayout`](https://github.com/material-components/material-components-android/blob/master/docs/components/TextField.md) (as suggested by @westnordost) and filter [`Chip`s](https://github.com/material-components/material-components-android/blob/master/docs/components/Chip.md#filter-chip). Here's how it looks right now:

image

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFiltersDialog.kt
westnordost: 😥

Looks like we need to look for something for this in the future, then. There is certainly the need for that.

> Regardless of user's locale, HH:mm:ss.SSS format seems to be the only one that fits limited space and is precise up to milliseconds, don't you think?

It should be well understandable at least. And I would think that times are specified like this everywhere anyway (except maybe the ms part).

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFragment.kt
```diff
@@ -0,0 +1,145 @@
...
+    private fun showLogs() {
+        viewLifecycleScope.launch {
+            val logs = getLogs()
+            adapter.messages = logs
+            binding.toolbar.root.title = getString(R.string.about_title_logs, logs.size)
+            binding.logsList.scrollToPosition(logs.lastIndex)
+        }
+    }
+
+    private suspend fun getLogs(): List<LogMessage> = withContext(Dispatchers.IO) {
```
westnordost, line 131: By the way, this means that on opening the log view, in absence of any filters, it will load *all logs* from the database, i.e. logs spanning over several days. The index doesn't help here.

A workaround might be to add some sort of internal pagination that uses `LIMIT $limit` for the first query and on scrolling, for `newerThan`/`olderThan` is passed to the DAO equal to the first/last item in the list. (However, this solution would have the caveat, that the scroll bar always has the wrong size)

But this seems to be a bit of work and adds some complexity. And if/when the UI code is changed to Jetpack Compose / Multiplatform Compose, the implementation would need to look different anyway. Maybe there is even a facilitation for this, as this seems to be a common use case. So in any case, as long as this does not turn to be a huge problem, I think it is fine to leave it as it is now, without any sort of internal pagination/buffering.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFragment.kt
```diff
@@ -0,0 +1,145 @@
...
+                    onClickFilter()
+                    true
+                }
+                else -> false
+            }
+        }
+    }
+
+    private fun onClickShare() = viewLifecycleScope.launch {
+        val shareIntent = Intent(Intent.ACTION_SEND).apply {
```
westnordost, line 90: By the way, I usually avoid `.apply` and other functions whose block redefines `this` at least within classes, as one often needs the IDE's code hints to avoid the confusion what `this` refers to.

But anyway, not a hard rule and certainly not done consistently throughout the codebase, so I don't mind this here.

westnordost (approved):
The image in your last post doesn't display, so I made one:

Screenshot_1699205435

It looks really pretty now and everything works like a charm! 

I wonder, are you following any material design guidelines with these chips? My gut feeling is, that any checked chips should have the log level color as background color, while any unchecked chips should have grey as their background color. 
But IMO the Material 3 design guidelines (which SC does not currently follow, it uses Material 2) kind of broke the whole chips-design, as any button also has this same rounded shape as chips too, if I remember correctly.

(Anyway, no need to put any more effort in the UI, it looks already as polished as it could be)

Screenshot_1699205428

This is the "share" prompt on an Android 5.1.1 emulator. Disappointed that the share prompt didn't have the "copy to clipboard" option back then. I hope it was introduced in an Android version soon after.

westnordost:
After merge I tested this on an API 25 emulator and it shows this weird behavior (can't make a video due to some error):

Screenshot_1701018799

All checkmarks have the same color, then if I disable and enable another log level, **all** checkmarks get the color of the last checked one. Not reproducible on an Android API 34 emulator and also wasn't able to fix it playing around with how the filter chip is initialized, so I guess it must be a bug in some older version of Android.

(So unless someone has an idea what we do wrong here and/or if there is a workaround, it shall be a will not fix. I didn't test it, but if this issue occurs on relatively recent versions of android too, we should consider to just use the standard, less fancy, style for chips then https://m2.material.io/components/chips#filter-chips )

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFragment.kt
westnordost: Actually, a really easy workaround came to my mind: The default filters when entering the fragment are only the logs from today.

```kotlin
    init {
        val startOfToday = LocalDateTime(systemTimeNow().toLocalDate(), LocalTime(0, 0, 0))
        filters = LogsFilters(timestampNewerThan = startOfToday)
    }
```

FloEdelmann:

PR #5338, opened 2023-10-27
Upgrading minor dependencies

mccartney:
**What**: Upgrading 3 minor dependencies: zxing, kaml and OpeningHoursParser to their latest available versions.

**Why**: For no specific reason other than staying up-to-date

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -175,17 +175,17 @@ dependencies {
     implementation("org.jbox2d:jbox2d-library:2.2.1.1")
 
     // sharing presets/settings via QR Code
-    implementation("com.google.zxing:core:3.5.0")
+    implementation("com.google.zxing:core:3.5.2")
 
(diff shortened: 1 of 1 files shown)


matkoniecz:
Have you run tests?

Have you tested related functionality (say QR codes)?

mccartney:
Testing performed:
- [x] `./gradlew test`
- [x] running the app on my physical phone using Wireless Debugging, going to Settings | Manage Presets | dots | share and a QR code appeared
- [x] using the same StreetComplete Dev opening a quest of "Are these opening hours still correct" and observing parsed hours

Would you be so kind to suggest any other forms of testing I should apply here?

westnordost:
That should be sufficient. One trap to fall in is also if a library uses reflection and the classes which need to be kept for ProGuard/R8 have changed in the new version of the library. But I think kaml doesn't use reflection.

PR #5346, opened 2023-10-30
Get rid of CopyOnWriteArrayList

neonowy:
Replaces `java.util.concurrent.CopyOnWriteArrayList` with a lightweight wrapper around `HashSet` for storing listeners in a thread-safe way. (See https://github.com/streetcomplete/StreetComplete/issues/1892#issuecomment-1671074967)


westnordost (approved):
Looks good, thank you!

I figure, all this

```kotlin
    fun addListener(listener: Listener) {
        listeners.add(listener)
    }
    fun removeListener(listener: Listener) {
        listeners.remove(listener)
    }
```

stuff could be removed now too and replaced with the `listeners` field to be public, now that it does not expose the whole of a mutable list. (I.e. users call `someController.listeners.add(...)`)

To be clean, it would be necessary however to change `Listeners` into an interface and make a class like `HashSetListeners` or whatever that implements `Listeners` so that users of `Listeners` don't have access to the `forEach` method.

So, not sure if it would be worth it 🤷, this PR is approved, anyway.

FloEdelmann:

PR #5349, opened 2023-11-01
Removing assertj dependency

mccartney:
**What**: Removing `assertj` dependency in tests.

**Why**: In order to make a tiny step towards #1892

**Testing performed**
- [x] just a `./gradlew test`

I suggest a careful review as a bug in tests might cause us miss some issues in the future.

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/quests/oneway_suspects/TrafficFlowSegmentsApiTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/quests/oneway_suspects/TrafficFlowSegmentsApiTest.kt
@@ -5,8 +5,9 @@ import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
 import de.westnordost.streetcomplete.quests.oneway_suspects.data.ONEWAY_API_URL
 import de.westnordost.streetcomplete.quests.oneway_suspects.data.TrafficFlowSegment
 import de.westnordost.streetcomplete.quests.oneway_suspects.data.TrafficFlowSegmentsApi
-import org.assertj.core.api.Assertions.assertThat
 import kotlin.test.Test
+import kotlin.test.assertContains
+import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 
 class TrafficFlowSegmentsApiTest {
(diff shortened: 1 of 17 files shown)


app/src/androidTest/java/de/westnordost/streetcomplete/quests/oneway_suspects/TrafficFlowSegmentsApiTest.kt
```diff
@@ -27,7 +28,7 @@ class TrafficFlowSegmentsApiTest {
             1L to listOf(TrafficFlowSegment(LatLon(2.0, 1.0), LatLon(6.0, 5.0))),
             2L to listOf(TrafficFlowSegment(LatLon(4.0, 3.0), LatLon(8.0, 7.0)))
         )
-        assertThat(result).containsAllEntriesOf(expected)
+        assertMapsAreEqual(expected, result)
```
westnordost, line 31: Actually, shouldn't `assertEquals(expected, result)` work?

(`TrafficFlowSegment` and `LatLon` are both data classes, the rest being basic types)

app/src/test/java/de/westnordost/streetcomplete/osm/LocalizedNamesKtTest.kt
```diff
@@ -133,6 +133,5 @@ internal class LocalizedNamesKtTest {
 private fun verifyAnswer(tags: Map<String, String>, answer: List<LocalizedName>, expectedChanges: Array<StringMapEntryChange>) {
     val cb = StringMapChangesBuilder(tags)
     answer.applyTo(cb)
-    val changes = cb.create().changes
-    Assertions.assertThat(changes).containsExactlyInAnyOrder(*expectedChanges)
+    assertSetsAreEqual(expectedChanges.toSet(), cb.create().changes)
```
westnordost, line 136: Same here: `assertEquals(expectedChanges.toSet(), cb.create().changes)`

westnordost (commented):
Thank you for looking into it!

mccartney:
Just a status report: I remember about this PR and will update as per the feedback given. I am swamped with something else with higher prio.

FloEdelmann:
(Just assigning people who work on the ticket so that https://github.com/orgs/streetcomplete/projects/1/views/1 looks nicer :-) )

westnordost:
I am currently rewriting all those tests that used to have the `verifyAnswer` function into something that is IMO more readable. A side-effect of that is that the `AssertThat` stuff is not used anymore. So, continue to work on this PR does not make sense.

PR #5354, opened 2023-11-04
RFC: opening_hours: prefer "open" to deceptive "24/7"

lynxlynxlynx:
both imply the same meaning, but they are not the same. So prefer the unambiguous "open".

opening_hours.js warns that 24/7 is actually 24/5:

>   WARN: 24/7 is handled as a synonym for 00:00-24:00, so Mo-Fr 24/7 (though
>   not really correct, because of that you should avoid it or replace it
>   with "open".

I didn't test it, but it does address all uses of `AlwaysOpen`, while leaving 24/7 parsing and logic intact.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -167,7 +167,7 @@ class AddOpeningHours(
         } else {
             val openingHoursString = when (answer) {
                 is RegularOpeningHours  -> answer.hours.toString()
-                is AlwaysOpen           -> "24/7"
+                is AlwaysOpen           -> "open"
                 is DescribeOpeningHours -> "\"" + answer.text.replace("\"", "") + "\""
                 NoOpeningHoursSign      -> throw IllegalStateException()
             }
--- a/app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt
+++ b/app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt
@@ -75,7 +75,7 @@ class AddOpeningHoursTest {
     @Test fun `apply always open answer`() {
         questType.verifyAnswer(
             AlwaysOpen,
-            StringMapEntryAdd("opening_hours", "24/7")
+            StringMapEntryAdd("opening_hours", "open")
         )
     }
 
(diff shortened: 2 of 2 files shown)


Cj-Malone:
Is there any consensus that "24/7 is actually 24/5"?

It's certainly not wha I mean when I tag `24/7`

lynxlynxlynx:
That's the motivation for this PR, once I noticed the documentation. OSM wiki links to the js lib as a spec.

Helium314:
See https://github.com/opening-hours/opening_hours.js#time-ranges
The description is very confusing, I read is as "It means what you think it means, and should only be used for that. But we interpret it as differently (no reason given)."

matkoniecz:
https://openingh.ypid.de/evaluation_tool/?EXP=24%2F7&lat=48.7769&lon=9.1844&mode=0&DATE=1699089420000 - `24/7` works

It is only `Sa-Su 24/7` and similar misuse that fails. See https://openingh.ypid.de/evaluation_tool/?EXP=Sa-Su%2024%2F7&lat=48.7769&lon=9.1844&mode=0&DATE=1699089420000

matkoniecz:
Maybe that documentation should be improved? But it already has

> In cases where a facility is really open 24 hours 7 days a week thats where this value is for.

that describes our use.

Cj-Malone:
I don't agree with the 24/5 thing, but is there a similar question about public holidays? Does 24/7 include them or not?

I'm now wondering if SC "Always open" should fill the UI with "Mo-Fr 00:00-24:00", allow the user to edit (Add PH, remove Sa Su, etc) then tag the more explicit value.

Helium314:
The 24/7 thing may be triggered by a recent edit in the wiki, also promoting the `open` value that has only 89 uses

matkoniecz:
I opened https://github.com/opening-hours/opening_hours.js/pull/453

from

> - **WARN:** 24/7 is handled as a synonym for `00:00-24:00`, so `Mo-Fr 24/7` (though not really correct, because of that you should avoid it or replace it with "open". A warning will be given if you use it anyway for that purpose) will be handled correctly

to

> - **WARN:** 24/7 is handled as a synonym for `00:00-24:00`, so it can be misused like `Mo-Fr 24/7` and still interpreted (but it is not really correct, you should avoid it or replace it with "open". A warning will be given if you use it anyway for that purpose)

matkoniecz:
> The 24/7 thing may be triggered by a recent edit in the wiki, also promoting the open value that has only 89 uses

link?

Helium314:
Looks like it was just reverted: https://wiki.openstreetmap.org/w/index.php?title=Key:opening_hours&action=history

lynxlynxlynx:
Ok, it's easier if the docs are just bad, though `open` is a valid choice here. 
I wonder if there's any value in discerning places that are always open (like a walled park or in what prompted this investigation, one mall in Slovenia) and those that are marketed as 24/7 (like some petrol stations or shops outside Europe).

matkoniecz:
`open` is less clear than`24/7` though `Mo-Fr 00:00-24:00` proposed in #5355 may make sense

PR #5357, opened 2023-11-04
Wrap SharedPreferences behind Preferences interface

neonowy:
Abstracted Android dependency `SharedPreferences` by introducing `Preferences` interface, as suggested in https://github.com/streetcomplete/StreetComplete/issues/1892#issuecomment-1671074967.

The new `AndroidPreferences` implementation uses non-blocking `apply()` everywhere, for consistency/simplicity. This also guarantees you can safely write to preferences from the UI thread. According to [Android docs](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()), changing the existing `commit()` calls to use `apply()` should be safe:
> As [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) instances are singletons within a process, it's safe to replace any instance of [commit()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#commit()) with [apply()](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()) if you were already ignoring the return value.
>
> You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states.

### Breaking changes:
- `MapFragment` and `LocationAwareMapFragment` preferences now use the default shared preferences instance (stored in `shared_prefs/de.westnordost.streetcomplete.preferences.xml`) instead of one scoped to the `MainActivity` (stored in `shared_prefs/de.westnordost.streetcomplete.screens.MainActivity.xml`).
  **AFAIK resetting these values is not a huge problem, since user often changes them by interacting with the map anyway?** If you prefer backwards compatibility over consistency in accessing preferences using a single abstraction, I can leave these unchanged, as Fragments code won't be shared across platforms anyway.


westnordost:
⚡

app/src/main/java/de/westnordost/streetcomplete/util/prefs/Preferences.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun putFloat(key: String, float: Float)
+    fun putDouble(key: String, double: Double)
+    fun putString(key: String, string: String?)
+
+    fun getBoolean(key: String, defaultValue: Boolean): Boolean
+    fun getInt(key: String, defaultValue: Int): Int
+    fun getLong(key: String, defaultValue: Long): Long
+    fun getFloat(key: String, defaultValue: Float): Float
+    fun getDouble(key: String, defaultValue: Double): Double
+    fun getString(key: String, defaultValue: String?): String?
```
westnordost, line 20: I think it would be nice to have the second parameter optional and have reasonable defaults (documented as part of the signature). E.g. 

```suggestion
    fun getBoolean(key: String, defaultValue: Boolean = false): Boolean
    fun getInt(key: String, defaultValue: Int = 0): Int
    fun getLong(key: String, defaultValue: Long = 0L): Long
    fun getFloat(key: String, defaultValue: Float = 0f): Float
    fun getDouble(key: String, defaultValue: Double = 0.0): Double
    fun getString(key: String, defaultValue: String? = null): String?
```

Not sure about this, but maybe actually make it all nullable (so that default is always `null`)?

You've combed through the source code, what do you think would be the most convenient (and self-explanatory) interface?

westnordost (commented):
Did you have a look at [Multiplatform-Settings](https://github.com/russhwolf/multiplatform-settings)? (Does it have the same guarantees regarding (UI) thread safety as your implementation?)

Reading the readme.md, it looks quite well done. The first stable release has been released almost a year ago. This `flow` stuff could be useful for Jetpack Compose.

westnordost:
Note; The iOS version of `SharedPreferences` is [`NSUserDefaults`](https://developer.apple.com/documentation/foundation/nsuserdefaults).  If the interface exposes functionality that this class doesn't have, this could be a problem. Didn't check.

westnordost:
Alright, so I will merge this now and create a ticket for the thing mentioned later

FloEdelmann:

PR #5359, opened 2023-11-05
Disable restoring hidden quests when there are none to unhide.

jmizv:
The option to unhide hidden quests should be disabled when there are none to unhide.

grafik

grafik


app/src/main/java/de/westnordost/streetcomplete/screens/settings/SettingsFragment.kt
```diff
@@ -261,4 +280,32 @@ class SettingsFragment :
...
+
+    private fun setHiddenQuestsSummary() {
+        val amountOfHiddenQuests = countHiddenQuests()
+        viewLifecycleScope.launch {
+            val pref = findPreference<Preference>("quests.restore.hidden")
+            pref?.summary = calculateHiddenQuestsSummary(amountOfHiddenQuests)
+            if (amountOfHiddenQuests == 0L) {
+                pref?.isEnabled = false
+            }
+        }
```
westnordost, line 292:
```suggestion
        val pref = findPreference<Preference>("quests.restore.hidden")
        viewLifecycleScope.launch {
            val amountOfHiddenQuests = withContext(Dispatchers.IO) { countHiddenQuests() }
            pref?.summary = calculateHiddenQuestsSummary(amountOfHiddenQuests)
            pref?.isEnabled = amountOfHiddenQuests > 0
        }
```

1. countHiddenQuests is an IO ooperation
2. preference needs to be enabled when amountOfHiddenQuests > 0

app/src/main/res/values/strings.xml
```diff
@@ -549,6 +549,9 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="pref_summaryOff_show_notes_not_phrased_as_questions">"Ignoring notes not phrased as questions"</string>
 
     <string name="pref_title_quests_restore_hidden">Restore hidden quests</string>
+    <string name="pref_title_quests_restore_hidden_summary_none_hidden">You don’t have any hidden quests.</string>
+    <string name="pref_title_quests_restore_hidden_summary_one_hidden">1 quest is currently hidden.</string>
+    <string name="pref_title_quests_restore_hidden_summary_multiple_hidden">%1$d quests are currently hidden.</string>
```
westnordost, line 554: I'd like to reduce the number of necessary strings (=thus, effort for translators)

How about `Number of hidden quests: %d`

westnordost (changes requested):
Nice! But, hmm, ... why? Why is it useful to know beforehand how many quests have been hidden?

app/src/main/res/values-de/strings.xml
```diff
@@ -503,6 +503,9 @@ Teilweise bedeutet, dass man die Toilette schon im Rollstuhl benutzen kann, aber
     <string name="quest_motorcycleParkingCoveredStatus_title">"Ist der Motorradstellplatz überdacht (regengeschützt)?"</string>
     <string name="default_disabled_msg_maxspeed">"Dieser Aufgabentyp ist standardmäßig deaktiviert, weil du normalerweise die ganze Straße nach entsprechenden Schildern absuchen musst, nicht nur den hervorgehobenen Abschnitt. Das kann manchmal viel Arbeit für eine einzelne Aufgabe sein."</string>
     <string name="pref_title_quests_restore_hidden">"Versteckte Aufgaben wiederherstellen"</string>
+    <string name="pref_title_quests_restore_hidden_summary_none_hidden">Du hast keine versteckten Aufgaben.</string>
+    <string name="pref_title_quests_restore_hidden_summary_one_hidden">Eine Aufgabe ist aktuell versteckt.</string>
+    <string name="pref_title_quests_restore_hidden_summary_multiple_hidden">%1$d Aufgaben sind aktuell versteckt.</string>
```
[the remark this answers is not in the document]
jmizv: Yes, I know. But does it any harm to have it here? I can, of course, clean this XML before merging.

app/src/main/java/de/westnordost/streetcomplete/data/Database.kt
```diff
@@ -73,6 +73,10 @@ interface Database {
         args: Array<Any>? = null
     ): Int
 
+    fun countAll(
+        table: String
+    ): Long
+
```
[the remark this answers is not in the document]
jmizv: Okay, I will change this.

jmizv:
> Nice! But, hmm, ... why? Why is it useful to know beforehand how many quests have been hidden?

Maybe the focus of this PR should be: "Disable the option to unhide hidden quests when there are none". And the information about the currently hidden quests is just a nice-to-have. So it is not useful per se but in combination with the disabling of the resetting it is nice to see why it is disabled (=no hidden quests).

app/src/main/res/values-de/strings.xml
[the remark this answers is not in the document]
westnordost: Yes, it doesn't do any harm. It is just wasted effort.

peternewman:
> Maybe the focus of this PR should be: "Disable the option to unhide hidden quests when there are none". And the information about the currently hidden quests is just a nice-to-have. So it is not useful per se but in combination with the disabling of the resetting it is nice to see why it is disabled (=no hidden quests).

I always thought it seemed a bit odd that it told you after you'd unhidden them how many would be unhidden, but not before you made the choice. It seems a nice to have, is it just doing the ten from earlier today or 200 from a few weeks/month?

jmizv:
Conflicts in `.../values-de/strings.xml` can be ignored, right?
@westnordost I've made the requested changes. Can you please take a look again? Thanks.

HolgerJeromin:
> Conflicts in .../values-de/strings.xml can be ignored, right?

Your PR should not include additions to translated files. The strings will be translated later.

jmizv:
> > Conflicts in .../values-de/strings.xml can be ignored, right?
> 
> Your PR should not include additions to translated files. The strings will be translated later.

Okay, conflicts and changed for that file have been removed.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/SettingsFragment.kt
```diff
@@ -65,6 +67,8 @@ class SettingsFragment :
     private val questTypeRegistry: QuestTypeRegistry by inject()
     private val visibleQuestTypeSource: VisibleQuestTypeSource by inject()
     private val questPresetsSource: QuestPresetsSource by inject()
+    private val osmQuestsHiddenDao: OsmQuestsHiddenDao by inject()
+    private val noteQuestsHiddenDao: NoteQuestsHiddenDao by inject()
```
westnordost, line 71: It is uncommon (if not: not done at all) in this codebase to directly access the Daos from view code. Fragments and Activities instead access *Controllers or *Source classes to update or get data. The update of the view subsequently should be reactive to changes in the data, i.e. the quest summary should not update after having tapped the button, but instead the tapping of the button changes the data and because the data is changed, the label is updated.

I wanted to write the details but that got so involved that I rather do it myself unless you know now exactly what to do. I'll do that then otherwise shortly before the next update.

westnordost (commented):
Looks good, except for one thing, which is a trifle on its own, but I want to avoid that this kind of thing creeps into the codebase:

app/src/main/java/de/westnordost/streetcomplete/screens/settings/SettingsFragment.kt
jmizv: I'd be happy if you could work on this as I don't think I'm expert enough to do it myself. Thank you!

westnordost:
Alright, did this now, I also updated that one overview of classes and how they are connected, this is why the change seems to be so big counted in lines of code. In a nutshell, what I did was:

- don't access any Dao from anywhere else than the *Controller classes
- make the OsmQuestController and the OsmNoteQuestController implement new interfaces that are just for handling quest visibility
- use these interfaces in places where only getting/counting/hiding/unhiding the quests made invisible are of concern - e.g the settings fragment but also the EditHistoryController
- in the SettingsFragment, update the summary ("x quests hidden") not after clicking that button, but after the data has changed - i.e. the fragment listens now to the changes in the data

westnordost:
Feel free to look at my changes, you can see them here: https://github.com/streetcomplete/StreetComplete/pull/5359/commits/0ab82c5fad95b0b4baf0813587c95889eb271a73

PR #5363, opened 2023-11-06
New Quest: Does this caravan area have a sanitary_dump_station?

### LINKED ISSUE #5233: Does this caravan area have a sanitary_dump_station ?
This is a proposition to tag objects tagged with this tag : [caravan_site](https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dcaravan_site)

### General
Affected tag(s) to be modified/added:   [sanitary_dump_station](https://wiki.openstreetmap.org/wiki/Key:sanitary_dump_station)
Question asked: **Does this caravan area have a  [sanitary_dump_station](https://wiki.openstreetmap.org/wiki/Key:sanitary_dump_station)?**
response : yes/no/custumors 

same could be done for theses tags  :
- [water_point](https://wiki.openstreetmap.org/wiki/Key:water_point)
- [power_supply](https://wiki.openstreetmap.org/wiki/Key:power_supply)

### Checklist
- [ x ] 🚧 To be added tag is established and has a useful purpose
- [ x ] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [ x ] 🐿️ Easily answerable by any pedestrian from the outside but a survey is necessary
- [ x ] 💤 Not an overwhelming percentage of quests have the same answer (No spam)
- [ x ] 🕓 Applies to a reasonable number of map data (Worth the effort)

### Ideas for implementation

**Element selection:** 

**Metadata needed:** 

**Proposed UI:**

qugebert:
Closes #5233 

Screenshot_20231106-144822


app/src/main/java/de/westnordost/streetcomplete/quests/sanitary_dump_station/AddSanitaryDumpStation.kt
```diff
@@ -0,0 +1,35 @@
...
+        nodes, ways with
+         (
+           tourism = caravan_site
+           or tourism = camping_site and caravans = yes and !backcountry
+         )
+         and access !~ private|no
+         and !sanitary_dump_station
+    """
+
+    override val changesetComment = "Specify, if there is a sanitary dump station at camp or caravan site"
```
FloEdelmann, line 23:
```suggestion
    override val changesetComment = "Specify if there is a sanitary dump station at camp or caravan site"
```

app/src/main/res/authors.txt
```diff
@@ -225,6 +225,10 @@ recycling_centre.jpg           Public Domain     https://commons.wikimedia.org/w
 recycling_container.jpg        CC0               https://commons.wikimedia.org/wiki/File:Vitoria_-_Contenedores_de_reciclaje_en_la_Avenida_de_Gasteiz.jpg
 recycling_container_undergr... CC0               https://commons.wikimedia.org/wiki/File:San_Fernando_de_Henares_-_reciclaje_de_residuos_urbanos_04.JPG
 
+sanitary_dump_station_illus... CC BY-SA 4.0 Deed https://commons.wikimedia.org/wiki/File:Karlslunde_%C3%B8st_sanitary_dump_grey.jpg (Elgaard)
+                               CC BY-SA 3.0 Deed https://wiki.openstreetmap.org/wiki/File:Entsorgung_Camping_IMG_1510.JPG (Nightflyer)
+                               
+
 sliding_envelope.wav           CC0               https://freesound.org/people/MTJohnson/sounds/444431/
```
FloEdelmann, line 232:
```suggestion
                               CC BY-SA 3.0 Deed https://wiki.openstreetmap.org/wiki/File:Entsorgung_Camping_IMG_1510.JPG (Nightflyer)
                               
sliding_envelope.wav           CC0               https://freesound.org/people/MTJohnson/sounds/444431/
```

westnordost:
The pictures for the different resolutions all have the same resolution.

But apart from this, I am really not sure if these pictures are that helpful. When not looking (very) closely, they just look like photos of a generic motorway lay-by with some signs.

What do we have? We have

- a question that mentions the keyword "sanitary dump station" - most important
- a description that explains what a "sanitary dump station" is - also important, because there may be synonyms
- a quest icon that imitates how the sign for this may look like (in France)

So, to me, the photos do not add a value, because (very tiny) you see the same sign again, and for the right picture, I don't even know where the sanitary dump station is supposed to be.

Realisticially, such a facility would either be indicated by name, or by a well-known sign with an icon. So, what is missing to make sure that a surveyor with 0 knowledge about this can ascertain whether a place like this exists, is to show him how (else) this facility could be indicated, not some photos of how the place in general would look like.

Not sure if there is a an official Vienna-convention sign for this, if there is, this should be shown instead. Just like it is done for example in the maxspeed quest when the user selects that it is a living street, see here:
https://github.com/streetcomplete/StreetComplete/tree/master/res/graphics/living%20street

If no standardization exists, then might as well leave out the picture because not sure if showing many variations of how it may look like would be too helpful.

qugebert:
Ok, so i can remove the image, just added this because of the comments in the issue that we need further explanation. I have no idea how we can make a better hint, but i asked the issue author.

qugebert:
Ok, so like this?
Screenshot_20231110-155744

westnordost:
Hm, two times basically the same sign, a bit German(?)-centric. But that's okay, I might draw up some own icons myself after merge.

qugebert:
Sorry, ich habe überhaupt keinen Sinn für gute Grafiken / Design.
Ich könnte das hier noch als zweites Schild anbieten (eben selbst in Inkscape skizziert, angelehnt an Ergebnisse aus der Bildersuche), ansonsten würde ichs euch überlassen.

rect1510-7

qugebert:
Stimmt, das mit der Übersetzung hatte ich nicht bedacht, ich lasse den Text weg. 
Ich habe diese Version mal mit abgelegt, wenn du willst passe es gerne noch an oder nehmt was komplett anderes, ich kann so etwas wie gesagt einfach schlecht einschätzen.

westnordost:
So today I set out to draw some nicer generalized signs than shown in the last screenshot, including something like the brown sign, only to find out that you already changed the picture to include the brown sign.

Anyway, I think my redrawings are still a bit of an improvement, at least they fit the graphics style of the app more

westnordost (approved):

qugebert:
Dankeschön :)

sun-geo:
Thank you for the new quest.

Please update the query under:
https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/sanitary_dump_station/AddSanitaryDumpStation.kt
line 17:
from:
 or tourism = **camping_site** and caravans = yes and !backcountry
to:
 or tourism = **camp_site** and caravans = yes and !backcountry

In the wiki there is only 
https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dcamp_site
available.

FloEdelmann:
Thanks for finding this @sun-geo! Would you like to update this yourself and send a pull request?

sun-geo:
I updated it, see request: https://github.com/streetcomplete/StreetComplete/pull/5407

PR #5365, opened 2023-11-06
Exclude motor_vehicle=no from oneway quest

arrival-spring:
I found a note that showed that the oneway quest triggered for a road tagged as `motor_vehicle=no`, and no other access tags. It also had bollards mapped on it. This seems like a valid way to map such a road, and clearly oneway does not apply in such cases, so the quest should not be asked.

I haven't tested this yet, apologies.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt
@@ -27,7 +27,7 @@ class AddOneway : OsmElementQuestType<OnewayAnswer> {
         ways with highway ~ living_street|residential|service|tertiary|unclassified
          and width <= 4 and (!lanes or lanes <= 1)
          and !oneway and area != yes and junction != roundabout
-         and (access !~ private|no or (foot and foot !~ private|no))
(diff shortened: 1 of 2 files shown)


mnalis:
But your PR (unless I'm misreading it) does not seems to trigger **only** on _"road tagged as motor_vehicle=no, and no other access tags"_. I too haven't tested it, but it seems it would trigger for example on `motor_vehicle=no` + `bus=yes` (or  `taxi=yes`, `psv=yes` etc); for which it is relatively common that they are oneway, but only specific type of motor vehicle is allowed (and all other motor vehicles are prohibited).


e.g. http://overpass-turbo.eu/s/1D0P

arrival-spring:
Ah yes, access tags are always more complicated than you first think. Is it best to say that this is rare enough of a situation that we ignore it (and close this PR). Or do we make the filter more complicate somehow?

mnalis:
I'd just close this PR, as worst that will happen without it is that in some rare situation mapper would be asked if the way is oneway, when the answer might be mostly obvious. Not a biggie. Much better than adding extra complexity in the code (and tests for it) IMHO.

arrival-spring:
Ok, agreed

PR #5371, opened 2023-11-12
Grill fuel quest -- add OtherAnswer that this is just a firering, not a Grill

mnalis:
[amenity=bbq](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbbq) requires that there is (usually metal) cooking surface (grill/grate/plate) for it to be a Grill/Bbq. However, it is incorrectly tagged for lots of places which only have fire-containment walls, and not a cooking surface i.e. fire rings / firepits (that are tagged with [leisure=firepit](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfirepit) instead).

This PR adds `OtherAnswer` to indicate this is not a grill/bbq but instead a firering/firepit only. [Tested](https://github.com/mnalis/StreetComplete/actions/runs/6840861319) and works fine.

See https://community.openstreetmap.org/t/amenity-bbq-without-grill-grate/9220 for discussion about tagging itself.

(Suggestions for better English texts are always welcome)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBbqFuelForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBbqFuelForm.kt
@@ -2,24 +2,18 @@ package de.westnordost.streetcomplete.quests.bbq_fuel
 
 import androidx.appcompat.app.AlertDialog
 import de.westnordost.streetcomplete.R
-import de.westnordost.streetcomplete.quests.AbstractOsmQuestForm
-//import de.westnordost.streetcomplete.quests.AListQuestForm
+import de.westnordost.streetcomplete.quests.AListQuestForm
 import de.westnordost.streetcomplete.quests.AnswerItem
-//import de.westnordost.streetcomplete.quests.TextItem
+import de.westnordost.streetcomplete.quests.TextItem
 import de.westnordost.streetcomplete.quests.bbq_fuel.BbqFuel.WOOD
 import de.westnordost.streetcomplete.quests.bbq_fuel.BbqFuel.ELECTRIC
 import de.westnordost.streetcomplete.quests.bbq_fuel.BbqFuel.CHARCOAL
 
-class AddBbqFuelForm : AbstractOsmQuestForm<BbqFuelAnswer>() {
-//    override val items = listOf(
-//        TextItem(WOOD, R.string.quest_bbq_fuel_wood),
-//        TextItem(ELECTRIC, R.string.quest_bbq_fuel_electric),
-//        TextItem(CHARCOAL, R.string.quest_bbq_fuel_charcoal),
-//    )
-    override val buttonPanelAnswers = listOf(
-        AnswerItem(R.string.quest_bbq_fuel_wood) { applyAnswer(WOOD) },
-        AnswerItem( R.string.quest_bbq_fuel_electric) { applyAnswer(ELECTRIC) },
-        AnswerItem(R.string.quest_bbq_fuel_charcoal) { applyAnswer(CHARCOAL) },
+class AddBbqFuelForm : AListQuestForm<BbqFuelAnswer>() {
+    override val items: List<TextItem<BbqFuelAnswer>> = listOf(
+        TextItem(WOOD, R.string.quest_bbq_fuel_wood),
+        TextItem(ELECTRIC, R.string.quest_bbq_fuel_electric),
+        TextItem(CHARCOAL, R.string.quest_bbq_fuel_charcoal),
     )
 
     override val otherAnswers = listOfNotNull(
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBbqFuelForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBbqFuelForm.kt
@@ -2,18 +2,24 @@ package de.westnordost.streetcomplete.quests.bbq_fuel
 
 import androidx.appcompat.app.AlertDialog
 import de.westnordost.streetcomplete.R
-import de.westnordost.streetcomplete.quests.AListQuestForm
+import de.westnordost.streetcomplete.quests.AbstractOsmQuestForm
+//import de.westnordost.streetcomplete.quests.AListQuestForm
 import de.westnordost.streetcomplete.quests.AnswerItem
-import de.westnordost.streetcomplete.quests.TextItem
+//import de.westnordost.streetcomplete.quests.TextItem
 import de.westnordost.streetcomplete.quests.bbq_fuel.BbqFuel.WOOD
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/BbqFuelAnswer.kt
```diff
@@ -1,7 +1,11 @@
...
-enum class BbqFuel(val osmValue: String) {
+sealed interface BbqFuelAnswer
+
+enum class BbqFuel(val osmValue: String) : BbqFuelAnswer {
     WOOD("wood"),
     ELECTRIC("electric"),
     CHARCOAL("charcoal")
 }
+
+object NOT_BBQ : BbqFuelAnswer
```
westnordost, line 11: Convention is `IsNotBbqAnswer`

app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/AddBbqFuelForm.kt
```diff
@@ -1,14 +1,31 @@
...
+        TextItem(WOOD, R.string.quest_bbq_fuel_wood),
+        TextItem(ELECTRIC, R.string.quest_bbq_fuel_electric),
+        TextItem(CHARCOAL, R.string.quest_bbq_fuel_charcoal),
+    )
 
-    override val items = listOf(
-        TextItem(BbqFuel.WOOD, R.string.quest_bbq_fuel_wood),
-        TextItem(BbqFuel.ELECTRIC, R.string.quest_bbq_fuel_electric),
-        TextItem(BbqFuel.CHARCOAL, R.string.quest_bbq_fuel_charcoal),
+    override val otherAnswers = listOfNotNull(
```
westnordost, line 19:
```suggestion
    override val otherAnswers = listOf(
```

app/src/main/res/values/strings.xml
```diff
@@ -1298,6 +1298,8 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_bbq_fuel_wood">Wood</string>
     <string name="quest_bbq_fuel_electric">Electric</string>
     <string name="quest_bbq_fuel_charcoal">Charcoal</string>
+    <string name="quest_bbq_fuel_not_a_bbq">Not a grill - lacks a cooking surface</string>
+    <string name="quest_bbq_fuel_not_a_bbq_confirmation_title">Is this really just a fire ring instead of full grill?</string>
```
westnordost, line 1302: Other answer text is too long and likely to be truncated on most phones. Also, do you "cook" on a bbq grill?

Also, "fire ring" doesn't seem to be a common word. Are you sure that it will be understand what is meant?

https://www.youtube.com/watch?v=5WyLhwYFgmk

westnordost (commented):
Although answering that it is not a grill is not a direct answer to the question, we have other instances where such "other answers" are provided that are similar, I think. So, this PR is fine, only the wording seems off to me.

app/src/main/java/de/westnordost/streetcomplete/quests/bbq_fuel/BbqFuelAnswer.kt
mnalis: updated.

app/src/main/res/values/strings.xml
mnalis: That's why I'd appreciate comments from native English speakers, too. 

As for the `cook(ing)`, my understanding is that it is correct term, e.g. wikipedia [Grill disambiguation page](https://en.wikipedia.org/wiki/Grill) says (among other things):

> [Barbecue grill](https://en.wikipedia.org/wiki/Barbecue_grill), a device or surface used for **cooking** food, usually fuelled by gas or charcoal, or the part of a cooker that performs this function
[Flattop grill](https://en.wikipedia.org/wiki/Flattop_grill), a **cooking** device often used in restaurants, especially diners

But perhaps there are better words to use. What is important is that BBQ/Grill does not always have a grill/grate, but **might have a solid metal plate** (thus allowing food cooked with liquids), as noted [here](https://community.openstreetmap.org/t/amenity-bbq-without-grill-grate/9220/4). 

What differentiates `amenity=bbq` from `leisure=firepit` is that BBQ always provides _**some kind**_ of ***"cooking surface"***, which I've tried to convey there. Perhaps enumerating them (_"grill, grate, plate or similar"_)  might be better solution? But then the text might get even longer :man_shrugging: 

As for [fire ring](https://en.wikipedia.org/wiki/Fire_ring), dunno. [Fire pit](https://en.wikipedia.org/wiki/Fire_pit) might be more common, but I'm afraid people might decide _"well in my case it is not a hole in the ground (pit) so this cannot be an answer either"_. Maybe _"fire enclosure"_ could be more generic and less confusing term?

mnalis: I've changed strings to:
- _"Just a fire enclosure, without cooking surface"_ 
- _"Is this really just a fire ring / fire pit?"_

to try to use positive instead of a negative, and shorten a little while trying to use different words.

Better suggestions welcome!

westnordost: "Just a fire enclosure, without cooking surface"

Too long. "It's a fire enclosure". Then, in the dialog, there can be more text. E.g. "Are you sure?" - "A fire enclosure differs from ... by ..."

mnalis: Something like this?

small_Screenshot_20231117_051325_de westnordost streetcomplete mn debug

app/src/main/res/values/strings.xml
```diff
@@ -1298,6 +1298,8 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_bbq_fuel_wood">Wood</string>
     <string name="quest_bbq_fuel_electric">Electric</string>
     <string name="quest_bbq_fuel_charcoal">Charcoal</string>
+    <string name="quest_bbq_fuel_not_a_bbq">"It's a fire enclosure"</string>
+    <string name="quest_bbq_fuel_not_a_bbq_confirmation">"Fire enclosure doesn't have cooking surface (e.g. grate/plate), as opposed to grill."</string>
```
FloEdelmann, line 1302: 1. Please use typographic apostrophes (`’` instead of `'`), then the double quotes around the string are also not needed.
2. I suggest this little message improvement, but I welcome any native speakers to suggest further improvements!

```suggestion
    <string name="quest_bbq_fuel_not_a_bbq">It’s a fire enclosure</string>
    <string name="quest_bbq_fuel_not_a_bbq_confirmation">Fire enclosures don’t have a cooking surface (e.g. grate/plate), as opposed to grills.</string>
```

westnordost (approved):

app/src/main/res/values/strings.xml
mnalis: Added improved wording by @FloEdelmann , hopefully is good now

Issue #5152, opened 2023-07-18
Shop overlay does not support tourism=hotel and friends


mss:
**Use case**
Hotels, Bed & Breakfasts, etc. can't be added (and verified) via the app. Admittedly they aren't really shops but neither are some other amenities like kindergartens. Small guest houses are often located where small shops could be and it would be useful if one could just add them like the rest without switching apps.

**Proposed Solution**
I think these types of amenities should be available:

* tourism=hotel
* tourism=motel
* tourism=hostel
* tourism=guest_house
* maybe others

westnordost:
Related https://github.com/streetcomplete/StreetComplete/issues/5065#issuecomment-1589524501 and following comments.

So, in that regard, since hotels, motels, ... generally do not occupy shop spaces but rather either are purpose-built (many hotels) or occupy otherwise residential/office space, they are not included in the list. It should not be surprising to not find these POIs in the shop overlay, given its name. Retirement homes, hospitals, schools etc. are also not included.

mss:
Fair enough, I can continue using EveryDoor for these POIs.

But I'd like to point out that nowadays Churches are also included in the Shop Overlay. As are Kindergartens/Preschools and Corporate Offices. None of these are Shops and follow similar occupation patterns in dense metropolitan areas as Guest Houses, BnBs, and small hotels.

An example where this would have been useful is the changeset https://www.openstreetmap.org/changeset/138065994 in which I switched to EveryDoor to verify a Guest House and removed another. In the latter case the space was repurposed as a shop again. I added the seemingly missing shop (and created a duplicate POI) and only later discovered that it already existed but just wasn't shown in SC when I checked that street.

westnordost:
Hm, well as written, a line has to be drawn which POIs to include and which not _somewhere_, right?

mnalis:
> line has to be drawn which POIs to include and which not somewhere, right?

Probably, but it does needs some work to draw that line, as OSM tagging is a kind of a mess. Probably it would be easiest to copy it from project that has already put in the work to do it? 

For example, Every Door has `amenities` mode (similar to `Shop overlay` in SC) and `micromapping` mode (similar to proposed [street furniture](https://github.com/streetcomplete/StreetComplete/issues/4912) overlay in SC). The line drawn there seems pretty good to me.

[...]

PR #5373, opened 2023-11-14
street furniture overlay

### LINKED ISSUE #4912: Overlay for street furniture (etc.)
I figure there could be an overlay with which one could add (delete, move) street furniture, i.e. benches, waste baskets, bicycle stands, trees, hydrants, maybe even traffic signs, single lanterns etc.

The most difficult of this overlay would be to think of all those POIs that could be collected. And at the same time not... open a can of worms. Certainly, much more so than for shops, there is an inexhaustible number of things that **could** be added. So, we'd need to come up with a reasonable guideline what should be able to be added with this overlay and what not (either deserving an own overlay or not displayable+addable at all). Of course, only things that can be added with this overlay are also shown on the map.

Anyway, I think maybe the (long) initial phase we could do collaboratively:

https://docs.google.com/document/d/1tcLm0EDWoCBdkq1N0pKebtksJcF2vMtUgKc9h96fuF0/edit?usp=sharing

Contributions welcome! As far as I know, there is no comprehensive list of "street furniture" in the wiki, so we'd need to dig through it ourselves. Digging through the [iD presets](https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets) may help, too.

matkoniecz:
fixes #4912


app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
```diff
@@ -0,0 +1,42 @@
+package de.westnordost.streetcomplete.osm
+
+import de.westnordost.streetcomplete.data.elementfilter.toElementFilterExpression
+
+fun isStreetFurnitureFragment(prefix: String? = null): String {
+    val amenities = listOf(
+        "bicycle_parking", "bicycle_rental", "bench", "longer", "bbq", "grit_bin", "toilets",
```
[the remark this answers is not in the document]
matkoniecz: not in iD presets

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
```diff
@@ -0,0 +1,42 @@
...
+        or ${p}man_made ~ water_tap|obelisk|cross|monitoring_station|flagpole|carpet_hanger|planter
+        or ${p}tourism ~ viewpoint|artwork
+        or (${p}tourism ~ information and information ~ guidepost|board|map|terminal)
+        or ${p}historic ~ memorial|monument|wayside_shrine|wayside_cross|boundary_stone
+        or ${p}highway ~ milestone|street_lamp|emergency_access_point
+        or ${p}emergency ~ fire_hydrant|life_ring|phone|defibrillator|siren|lifeguard|assembly_point|access_point
+        or ${p}advertising
+        or ${p}leisure = pitch and sport ~ table_tennis|sport=chess
+        or ${p}natural ~ tree
+        or ${p}man_made = street_cabinet and street_cabinet != postal_service
```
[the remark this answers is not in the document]
matkoniecz: Because post_box/letter box cannot be added so I worried that it may be mistakenly added instead by people searching for these

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
goldfndr, line 7: Typo: lounger not longer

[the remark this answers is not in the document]
westnordost: (this in itself is not a reason to exclude it from the list though, because it could always be added to iD later)

westnordost:
Regarding what should be in the overlay and what not:

StreetComplete affords to be more opinionated than iD (presets) when it comes down to which things to support displaying and adding. iD potentially has everything in the presets for which there is a clear documentation in the wiki, no disputes (and someone created a PR for that).

For StreetComplete on the other hand, the quest and overlay guidelines continue to apply. For example, storm drains (`manhole=drain`) would be both spammy (because there are so many) and not serve a useful purpose, maybe more. There are a couple of more things whose useful purpose (and spammyness) may be questionable, such as (from the top of my head) free-standing bollards that are not actually vertices on roads and ways.

Furthermore, certain things should be not included in the overlay that may qualify as street furniture but should rather go into an own, dedicated overlay. One example of these would be `traffic_sign=*`.

A feature which fulfills the StreetComplete guidelines that fits into the overlay but there is no iD preset yet, can still be added to the overlay. It may not be displayed and not be addable for now, but as soon as an iD preset is added, it would be.

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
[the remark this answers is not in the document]
westnordost: Comments for the exclusion would be good, so that this can be changed once the underlying reason for it is solved.

app/src/test/java/de/westnordost/streetcomplete/overlays/street_furniture/FlagpoleMatching.kt
```diff
@@ -0,0 +1,29 @@
+package de.westnordost.streetcomplete.overlays.street_furniture
```
westnordost, line 1: as it tests `de.westnordost.streetcomplete.osm.IS_STREET_FURNITURE_INCLUDING_DISUSED_EXPRESSION`, should be renamed and moved to osm/StreetFurnitureTestKt.kt

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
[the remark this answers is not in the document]
matkoniecz: I guess it can be added now to query so that it will become supported in future if iD starts to support it?

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
```diff
@@ -0,0 +1,159 @@
...
+            SearchFeaturesDialog(
+                requireContext(),
+                featureDictionary,
+                element?.geometryType,
+                countryOrSubdivisionCode,
+                featureCtrl.feature?.name,
+                ::filterOnlyStreetFurniture,
+                ::onSelectedFeature,
+                listOf(
+                    // ordered by popularity, skipping trees as there are multiple variants of them
```
westnordost, line 89: should be ordered by occurrence (not necessarily the current popularity). Hence, makes sense that street lamp is at the top, but so should `natural=tree`. What variants?

(Whether or not both trees and street lamps should be moved to their own layer - in the latter case, the street lighting layer? - is independent of that.)

(Also, this should probably rather be a private val at top of the class)

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlay.kt
```diff
@@ -0,0 +1,66 @@
...
+import de.westnordost.streetcomplete.overlays.Color
+import de.westnordost.streetcomplete.overlays.Overlay
+import de.westnordost.streetcomplete.overlays.PointStyle
+import de.westnordost.streetcomplete.overlays.PolygonStyle
+import de.westnordost.streetcomplete.util.getNameLabel
+
+class StreetFurnitureOverlay(private val getFeature: (tags: Map<String, String>) -> Feature?) : Overlay {
+
+    override val title = R.string.overlay_street_furniture
+    override val icon = R.drawable.ic_quest_apple // TODO
```
westnordost, line 22: I think the bench icon would fit best.

westnordost (commented):
A preliminary review:

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
matkoniecz: Broadleaf, needlelaved and so on seems to have own entries in addition to general tree entry.

If someone is mapping trees at all then specific variants are better than general ones

westnordost: Why does it need to have own categories? Could one not simply have `natural/tree` here? I think this should be possible with the iD presets. Also, I guess if one **had** to choose, it would be the broadleaf one, as these are much more common as street trees.

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
```diff
@@ -0,0 +1,159 @@
...
+        checkIsFormComplete()
+    }
+
+    override fun hasChanges(): Boolean =
+        originalFeature != featureCtrl.feature
+
+    override fun isFormComplete(): Boolean = featureCtrl.feature != null
+
+    override fun onClickOk() {
+        viewLifecycleScope.launch {
```
westnordost, line 130: Why this `launch` and why is `createEditAction` a suspend function?

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
```diff
@@ -0,0 +1,159 @@
...
+
+        val element = element
+        originalFeature = element?.let {
+            if (IS_DISUSED_STREET_FURNITURE_EXPRESSION.matches(element)) {
+                DummyFeature(
+                    "street_furniture/unknown_disused",
+                    requireContext().getString(R.string.unknown_disused_street_furniture),
+                    "ic_preset_maki_marker_stroked",
+                    element.tags
+                )
```
westnordost, line 48: should be moved to the bottom, i.e.

```
originalFeature = featureDictionary.findTheFeature(...)
    ?: ( if (is disused) dummyDisused else dummy )
```
because it should be shown correctly when in the future, iD has some presets for that.

---

Actually... even better would be the following, as we can not expect that iD presets will ever contain (all the) disused features. Maybe then also the same for the shop overlay:

1. Copy the element tags and strip any disused prefix
2. Find the feature with these tags
3. If the feature matches the disused expression, wrap the returned feature in a `data class DisusedFeature(private var feature: Feature)` which changes the name etc. to e.g. "Bench (disused)"

And voila, we have disused features support.

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
```diff
@@ -0,0 +1,159 @@
...
+                    // but interesting and worth promoting
+                    "emergency/defibrillator",
+                    )
+            ).show()
+        }
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+    }
```
westnordost, line 112:
```suggestion
```

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        // title hint label with name is a duplication, it is displayed in the UI already
+        setTitleHintLabel(element?.tags?.let { getLocationLabel(it, resources) })
+        setMarkerIcon(R.drawable.ic_quest_apple)
```
westnordost, line 73: I think a bench icon would be confusing here. So, should probably use a neutral white plus on a a green (amenity) background.

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
matkoniecz: `natural/tree` can go here but that works encourage mappers to use that instead of more precise ones (they may be even unaware of them and be stuck with general entry now as latest used one)

Adding broadleaf deciduous one only can cause them to use it also for other cases.

Adding all variants would take too much space.

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
matkoniecz: Heh, I just went here to write that I  plan to implement first part of this while waiting at airport.

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
matkoniecz: Yes, that is leftover from shop overlay (this one started as a direct copy of it,)

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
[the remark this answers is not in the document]
matkoniecz: added to list

to activate it: propose iD preset to support it and/or make PR if accepted and/or map more of such POIs to enable it

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
westnordost: Some data:

- About 4 million street lanterns are mapped.
- About 23 million trees are mapped.
- Only 18% (4 million) trees are tagged with `leaf_type=*`. However, of these, 92% (4 million) are tagged with `broadleaved`
- iD presets exist for all of *tree*, *broadleaved tree* and *needleleaved tree*, each named properly. The icon for the latter is a needleaved tree, the icons for the former are (two different!) broadleaved trees

westnordost:
Generally, in terms of completeness, I understand you so far

- [x] went through the suggestions so far in the linked ticket and https://docs.google.com/document/d/1tcLm0EDWoCBdkq1N0pKebtksJcF2vMtUgKc9h96fuF0/edit?pli=1#heading=h.v3kovgbui5i2 and noted in a code comment or so if/why any of these were not included after all
- [x] checked implementation of "micromapping POIs" in Every Door and considered any that were explicitly mentioned there
- [ ] went through iD presets to see if there are any more that should be included
- [ ] went through taginfo to see if there are any more that should be included

Is this about right? Just wanting to document what has been looked at so far so that others see an avenue to further contribute to this PR.

matkoniecz:
> Generally, in terms of completeness, I understand you so far

Yes. I have not planned at all to go through iD presets, I guess it would be duplicated by looking at taginfo.

matkoniecz:
Also, I have tested it while going home: and this overlay works better than I expected.

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
```diff
@@ -0,0 +1,44 @@
...
+        or (${p}amenity = recycling and recycling_type = container)
+        or ${p}leisure ~ picnic_table|firepit
+        or ${p}man_made ~ water_tap|obelisk|cross|monitoring_station|flagpole|carpet_hanger|planter
+        or ${p}tourism ~ viewpoint|artwork
+        or (${p}tourism ~ information and information ~ guidepost|board|map|terminal)
+        or ${p}historic ~ memorial|monument|wayside_shrine|wayside_cross|boundary_stone
+        or ${p}highway ~ milestone|street_lamp|emergency_access_point
+        or ${p}emergency ~ fire_hydrant|life_ring|phone|defibrillator|siren|lifeguard|assembly_point|access_point
+        or ${p}advertising
+        or ${p}leisure = pitch and sport ~ table_tennis|sport=chess
```
FloEdelmann, line 30:
```suggestion
        or ${p}leisure = pitch and sport ~ table_tennis|chess
```

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
```diff
@@ -0,0 +1,44 @@
...
+        // is also disabled to avoid bad data being added
+    )
+    val p = if (prefix != null) "$prefix:" else ""
+    return ("""(
+        ${p}amenity ~ ${amenities.joinToString("|")}
+        or (${p}amenity = recycling and recycling_type = container)
+        or ${p}leisure ~ picnic_table|firepit
+        or ${p}man_made ~ water_tap|obelisk|cross|monitoring_station|flagpole|carpet_hanger|planter
+        or ${p}tourism ~ viewpoint|artwork
+        or (${p}tourism ~ information and information ~ guidepost|board|map|terminal)
```
FloEdelmann, line 25:
```suggestion
        or (${p}tourism = information and information ~ guidepost|board|map|terminal)
```

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
```diff
@@ -0,0 +1,44 @@
...
+        or ${p}historic ~ memorial|monument|wayside_shrine|wayside_cross|boundary_stone
+        or ${p}highway ~ milestone|street_lamp|emergency_access_point
+        or ${p}emergency ~ fire_hydrant|life_ring|phone|defibrillator|siren|lifeguard|assembly_point|access_point
+        or ${p}advertising
+        or ${p}leisure = pitch and sport ~ table_tennis|sport=chess
+        or ${p}natural ~ tree
```
FloEdelmann, line 31:
```suggestion
        or ${p}natural = tree
```

matkoniecz:
anyone can help with these, let me know if you looked through taginfo:

TODO: is any other popular (>1000 uses) `amenity` that should be added? Maybe looks at >100 ones...
TODO: is any other popular (>1000 uses) `leisure` that should be added? Maybe looks at >100 ones...
TODO: is any other popular (>1000 uses) `tourism` that should be added? Maybe looks at >100 ones...
TODO: is any other popular (>1000 uses) `man_made` that should be added? Maybe looks at >100 ones...
TODO: is any other popular (>1000 uses) `historic` that should be added? Maybe looks at >100 ones...
TODO: is any other popular (>1000 uses) `highway` that should be added? Maybe looks at >100 ones...
TODO: is any other popular (>1000 uses) `emergency` that should be added? Maybe looks at >100 ones...


TODO: go through comments
TODO: test untested code supporting display of disused featured (and retagging them - disallow it or handle extra tags present on them)

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
matkoniecz: as more natural can be added in theory, I would stay with this one I think

matkoniecz: add  `natural=tree_stump`

westnordost:
Oh, regarding my previous comment that someone who is excited about this overlay (I am also excited, but have not time in the coming days or even coming week) could look through taginfo to see if more POIs should be included in the overlay:

While doing that, keep in mind #5152 - the plan is to extend the "shop overlay" into a "places overlay" which shall complement this "street furniture overlay" to basically cover most POIs there are. So, while looking through taginfo for this PR, you could also add things that should otherwise go into the "places overlay" to a second list.
(It would be ideal if the extension of the "shop overlay" would go live in the same update as this PR)

westnordost:
A few more comments / TODOs:

- [ ] How does the overlay handle POIs that are vertices? Can POIs that are vertices be added? Can they be changed? (For example `barrier=bollard` as a vertex - can it be changed to a traffic light or crossing? Can a `barrier=bollard` that is a free-floating node be changed into a traffic light?) - does the behavior as observed make sense?
- [ ] Did you consider https://github.com/streetcomplete/StreetComplete/pull/5373#discussion_r1393438424 ? The more I think about the "even better" solution, the more I like it. The filter should be somewhat faster, easier to read, easier to understand for users (because they'll know _what_ disused amenity / POI it is) and any `disused` or even any other lifecycle tag can be supported this way at one central place.
- [ ] Did you test what happens when you select a new POI, tab out so that the Activity is destroyed and then go back in? Is the selection retained? (I noticed there is no saved instance state)

qugebert:
If i understand this right, you're using the id-tagging-schema, so could we maybe make the "term"-field searchable too?
E.g. if i enter "AED" here, defibrillator should be a suggestion, because the id-tagging-schema has it [listed as search term](https://github.com/openstreetmap/id-tagging-schema/blob/a7efca734551b65b1386fec857d4a95caf2606f1/data/presets/emergency/defibrillator.json#L18)

Screenshot_20231118-122325

matkoniecz:
@qugebert - is it possible that it is not appearing as it is not search term in German translation? Is it found by iD when language is set to German?

Is UAE flag untranslated to German in iD/NSI presets? (or maybe NSI is not being translated?)

Helium314:
For other things the term field is already used, but also in English neither `AED` nor `aed` find a defibrillator. Maybe there is some issue with _AED_ being in capital letters?

matkoniecz:
or maybe there is minimal length before term gets matched?

Helium314:
> or maybe there is minimal length before term gets matched?

Searching for `atm` finds [payment terminal](https://github.com/openstreetmap/id-tagging-schema/blob/a7efca734551b65b1386fec857d4a95caf2606f1/data/presets/amenity/payment_terminal.json) (term `atm`
Searching for `eap` finds [emergency assembly point](https://github.com/openstreetmap/id-tagging-schema/blob/a7efca734551b65b1386fec857d4a95caf2606f1/data/presets/emergency/assembly_point.json) (term `EAP`)

So I conclude neither length nor capitalization cause the AED to be not found (anyway, that should actually go in a separate issue)

qugebert:
> @qugebert - is it possible that it is not appearing as it is not search term in German translation? Is it found by iD when language is set to German?

mea culpa. Your implementation already searches for the search terms, it is just a bad translation of the presets, so just ignore my comment.
Thanks for the hint, i will try to add it to the translation as aed is a common name in german too.

matkoniecz:
Well, once you edit translation you will fix bug source, and it was actually affecting StreetComplete and it was found before deployment so thanks for testing! If you are stuck with fixing it let us know

matkoniecz:
help would be helpful:

if there are `man_made=lamp` enthusiasts here I encourage them to answer https://wiki.openstreetmap.org/wiki/Talk:Tag:man_made%3Dlamp and maybe improve also that page

https://github.com/openstreetmap/id-tagging-schema has 76 PRs open right now, but maybe someone wants to take a chance and submit PRs (or issues first?) for `amenity=ski_rental` `amenity=device_charging_station`  `amenity=dog_toilet`  `man_made=snow_cannon` `highway=cyclist_waiting_aid` (if not submitted already)

for example there is `man_made=insect_hotel` has https://github.com/openstreetmap/id-tagging-schema/pull/1020 - and not sure what can be done to speedify review

and maybe PR https://github.com/ideditor/schema-builder/issues/94 if someone wants to take chance with making PR for ignored issue

--------------------------------

new todos

TODO: in current version `man_made=planter` will not glue itself to ways and bollards are not supported at all. Supporting gluing things to ways, for some things, is annoying at best to implement (more annoying if tram stops are also supported as these glue itself to different things....)

TODO: test is adding surveillance features working well

TODO: `amenity ~ bicycle_rental|boat_rental|ski_rental`  - first two were in `isShopExpressionFragment` already but commented out as "usually outside, could be automatic too", added `ski_rental` there - also commented  out.
I am considering including them in both layers, but keeping them also in one works
not sure where `amenity=photo_booth` should go

matkoniecz:
and thanks to @mcliquid for review, if anyone else wants to make such review feel free to (I think that at this point I can mark review of taginfo as done so thanks again for doing this!)

matkoniecz:
> How does the overlay handle POIs that are vertices? Can POIs that are vertices be added?

no, and my initial attempt to support this (tram stops and bollards) revealed it to be highly annoying to support

matkoniecz:
should `natural=spring` be supported here? I cannot decide for now so taking note for now.

matkoniecz:
Also, I want to bump

> keep in mind https://github.com/streetcomplete/StreetComplete/issues/5152 - the plan is to extend the "shop overlay" into a "places overlay" which shall complement this "street furniture overlay" to basically cover most POIs there are. So, while looking through taginfo for this PR, you could also add things that should otherwise go into the "places overlay" to a second list.

https://github.com/streetcomplete/StreetComplete/pull/5373#issuecomment-1816990139

(so if there is some taginfo digger then making such extension list for #5152 and posting it there is helpful, maybe even making direct PR if someone is adventurous)

westnordost:
Exactly, if there is a group of similar POIs and one or some are included, all of these should be included to avoid wrong/imprecise taggings.

matkoniecz:
TODO: problem was found in testing

some entries have no icon (like `advertising=totem`), therefore are invisible and not tappable on map

solution: inherit more generic icon (or, in theory, ensure that iD items have all dedicated icons but that is unlikely to be viable)

westnordost:
Why does it not have an icon? Does the preset have no icon? I thought for overlays, points that have no icon are shown as a (pink) dot instead, or something?

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
peternewman: You're not worried about it looking a lot like first aid?
https://en.wikipedia.org/wiki/First_aid#/media/File:ISO_7010_E003_-_First_aid_sign.svg

Maybe a tick or paste icon? Or a pin with a plus?

westnordost: No. If it is not thick enough, the plus will not look like that.

FloEdelmann: The icon was changed to [`ic_preset_maki_marker_stroked`](https://commons.wikimedia.org/wiki/File:Maki7-marker-stroked.svg) in [`cbfa267`](https://github.com/streetcomplete/StreetComplete/pull/5373/commits/cbfa267cd25a33a7d12f57a17ef790ed603f8321).

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/advertising/totem.json has no icon

has no dedicated icon in iD either, see https://www.openstreetmap.org/edit?editor=id#map=21/50.07387/20.01560 - gets some catchall icon (though distinct from point icon...)

screen

westnordost:
Oof, I am sorry, I contributed a few small changes but then I pressed the wrong button. I rebased it on current master, instead of merging current master!

westnordost:
There, fixed it.

app/src/main/java/de/westnordost/streetcomplete/overlays/street_furniture/StreetFurnitureOverlayForm.kt
westnordost: I made a list of further things that could be in the quick select.

```
tree				23.0M
broadleaved tree		 4.0M
needleleaved tree		 0.3M

waste basket			 0.7M (fast growth)

recycling container		 0.4M

artwork				 0.2M

charging station		 0.1M (fast growth)
```

It is most convenient, if things that would often be added by users was in this list.

For this reason I suggest to remove the defibrillator (only at commercial buildings) and picnic table (only in parks) and add both the broadleaved tree and the needleleaved tree. Not sure about the others and how much space there is (on small screen)

matkoniecz:
All current TODOs

TODO:  Did you test what happens when you select a new POI, tab out so that the Activity is destroyed and then go back in? Is the selection retained? (I noticed there is no saved instance state)

TODO: there should be support for deleting objects, probably

TODO: maybe show type of disused features?

TODO: in current version man_made=planter will not glue itself to ways and bollards are not supported at all. Supporting gluing things to ways, for some things, is annoying at best to implement (more annoying if tram stops are also supported as these glue itself to different things....)

TODO: amenity ~ bicycle_rental|boat_rental|ski_rental - first two were in isShopExpressionFragment already but commented out as "usually outside, could be automatic too", added ski_rental there - also commented out.
I am considering including them in both layers, but keeping them also in one works
not sure where amenity=photo_booth should go

Helium314:
Some observiations from playing with the current state:
* in some cases (not always) I am able to convert bicyle parking ways to something that should be a node e.g. a street lamp
* changing objects is dangerous, it keeps many tags that should be removed (e.g. in above case the street lamp will keep the `capacity`)
* motorcycle parking is not shown, but probably should be (becuase there is a kind of motorcycle parking that is very similar to bicycle parking)
* sometimes where there are many elements together, nothing is shown on medium zoom, when I would expect at least one of the elements (possibly not fixable and a tangram issue, minor anyway)
* some non-applicable elements can be added, e.g. _Bank of Baroda_ (cash machine), but that's also in iD
* elements can be added outside the downloaded area (though not really related to this overay)

> there should be support for deleting objects, probably

definitely

> Did you test what happens when you select a new POI, tab out so that the Activity is destroyed and then go back in? Is the selection retained? (I noticed there is no saved instance state)

selection is lost

westnordost:
> in some cases (not always) I am able to convert bicyle parking ways to something that should be a node e.g. a street lamp

I think for street furniture, it only makes sense in very rare cases to convert one POI type into another. So, what do you think about that when opening the form of a POI that is already mapped, unlike for shops, one cannot change the type at all? (Instead, one would need to delete the old and add the new in such a rare case as "it is not a bench but a picnic table!" or whatever)

> possibly not fixable and a tangram issue, minor anyway

@Helium314  You might be happy to hear that I have been looking into creating a MapLibre StreetComplete style (light, dark) in the last days (and boy is the style definition much more cumbersome to work with than with tangram-es!). Though, there's still no new release of MapLibre since February this year.

Helium314:
Allowing only certain POI type conversions might be a lot of work, also with the potential issues about why-this-and-not-that.
I haven't thought through everything, but currently I agree that not allowing type change would be a good idea.

Btw deletion of ways could also be allowed, at least if the way and its nodes are not connected to anything else.

> and boy is the style definition much more cumbersome to work with than with tangram-es!

Oh indeed. Not sure how far you got, but maybe [this page](https://docs.stadiamaps.com/custom-styles/) could help.

> Though, there's still no new release of MapLibre since February this year.

At least development appears to be very active

matkoniecz:
Underground hydrants are confusing with their pillar icon - maybe not implementing direct deletion will be good enough? Separate icon makes far less sense in iD so not sure that suggesting or implementing it makes sense

westnordost:
Why does it not make sense as an iD preset? (underfloor hydrant)

westnordost:
I'll take care of #5152 and look through taginfo in the frame of https://github.com/streetcomplete/StreetComplete/pull/5373#issuecomment-1816817730 as I go along.

westnordost:
Alright, I have completed my journey through

`amenity`, `leisure`, `tourism`, `man_made`, `historic`, `highway`, `emergency`, `natural`, `attraction`, `advertising`, `boundary`, `healthcare`, `craft`, `club`, `office`, `shop`, `seamark`, `public_transport`, `golf`, `disc_golf`, `military`

 and updated both `Shop.kt`  (to be renamed to `Place.kt`) and `StreetFurniture.kt`. A lot of things were added to the latter, to the former, also some.

The new descriptions/comments for "places" and for "street furniture" are...

# `Shops.kt`

> ### **Map features like shops or amenities that usually have a name and can be entered.**
>
>  To demarcate this category from others, the following are not included:
> - things that are rather structures or buildings than places (e.g. communication towers, train station)
> - landuses, outside-things (golf courses, zoos, marketplaces, car parking)
>
>  **Note**: When this function is modified, please follow update instructions in:  https://github.com/mnalis/StreetComplete-taginfo-categorize/blob/master/README.md
>
>  **Important**: The contribution policy here is that only things may be added if an iD preset for it exists. This policy exists in order to reduce effort to maintain this list, i.e. we don't want to check, weigh and balance requests in parallel to iD maintainers (in terms of notability, it being unambiguous, consensus etc.)

So, while the list of things included in this overlay has been extended by things that are purpose-built - schools, universities, stadiums, hotels, guest houses, hostels, motels, police stations, fire brigade stations, bowling alleys, hospitals, cinemas, conference centers, townhalls, etc., what is still omitted are things that are rather like landuses, such as parks, gardens, golf courses, but also marketplaces, miniature golf courses, zoos, theme parks, military bases etc. 

So, some things that may be called "places" are still missing from this overlay even though they can be entered, but I found it weird to add it into this category, or what do you think? Here's a list of things I skipped:

- **leisure=** beach_resort, disc_golf_course, garden, golf_course, horse_riding, marina, miniature_golf, nature_reserve, park, pitch, playground, resort, schoolyard, shooting_ground, sports_centre, summer_camp, swimming_area, track, water_park
- **tourism=** camp_pitch, camp_site, caravan_site, picnic_site, theme_park, zoo
- **amenity=** animal_breeding, boat_storage, building_yard, bus_station, driver_training, ferry_terminal, festival_grounds, grave_yard , marketplace, parking, traffic_park, recycling,   recycling_type=centre, refugee_site , waste_transfer_station, waste_dump_site
- **military=** airfield, base, academy, ...

See changes here: https://github.com/streetcomplete/StreetComplete/pull/5373/commits/6eb6ddf9c278e66d913bb8ed1beb48017fdb7cc0

# `StreetFurniture.kt`

> ### **Small map features that are often mapped as points and usually cannot be entered.**
>
>  To demarcate this category from others, the following are not included:
>  - structures that are big enough that they would rather be classified alongside buildings
>    (e.g. water towers)
>  - permanent features of the landscape (e.g. geysers, peaks, springs)
>  - things that are features on other map features that wouldn't make sense on their own (e.g.
>    crossings, traffic lights, entrances, golf holes)
>  - things that should not be mapped as points - even if allowed - and landuses (e.g. car parks,
>    dog parks, playgrounds)
>
>  Additionally, these things are generally not included:
>  - things that are intangible / difficult to verify on-site (e.g. fishing spots)
>  - things that would just be madness to comprehensively map (e.g. manhole covers), maybe what's
>    madness to map or not is in the eye of the beholder, in this case, in the eyes of app
>    maintainers
>
>  **Important**: The contribution policy here is that only things may be added if an iD preset for it exists. This policy exists in order to reduce effort to maintain this list, i.e. we don't want to check, weigh and balance requests in parallel to iD maintainers (in terms of notability, it being unambiguous, consensus etc.)

Now, (when you look at the commit) you'll see that it includes a lot more than what could conventionally considered "street furniture", but we really lack a better for these kind of smallish POIs.

Also, what results to be the most difficult to properly separate from street furniture are structures. Where to draw the line? Small structures like shelters are already included. But specifically `man_made` is full of things that are smallish but also kind of structures.

Currently not included are `dolphin`, `dovecote` and `monitoring_station`. But many other kind-of-like structures are included, e.g. `hunting_stand`, `bird_hide`, `bandstand` (because shelters are included), `stage` (because bandstand is included), and more, see the commit.
So, what do you think?

See changes here: https://github.com/streetcomplete/StreetComplete/pull/5373/commits/ab4aed421827b08b942296eb8afd2176ce08aeec

westnordost:
Remaining TODOs:

- [ ] discuss element selection (see previous comment)
- [ ] support deleting POIs
- [ ] it should **_not_** be possible to change the type of an existing POI. Either delete it or (but I doubt its usefulness,) mark it as disused.
- [ ] "should bench icon be used on object selection? Or the same as during adding?" (_I don't understand this TODO_)

matkoniecz:
> "should bench icon be used on object selection? Or the same as during adding?" (I don't understand this TODO)

when you select already added object marker has bench icon

matkoniecz:
+1 to taking some decision on split here

> <string name="overlay_places">Places</string>

It is kind of confusing for people used to `place=` tag in OSM, but I have no better name :(

I think I would like to have `monitoring_station` to be included but I may be unusually interested in them.

westnordost:
"Places" could also be named "Shops & Amenities". But this would be slightly incorrect as it includes more and I think, the association with `place` only exists for OSM veterans.

Also, since "Street Furniture" does not work 100% for the things included plus more importantly, I expect it will be difficult and/or awkward to translate, I am considering just using the word "Things".

So there are then Places, Things, (Structures, Roads).

westnordost:
So what do you think about the rename to "Things"?

westnordost:
>> "should bench icon be used on object selection? Or the same as during adding?" (I don't understand this TODO)

> when you select already added object marker has bench icon

Hm, right, well, this is somewhat of a technical limitation. So, maybe use the plus icon for all? Or maybe have a white dot (inside the green circle) as an icon for the overlay?

peternewman:
> > ### **Map features like shops or amenities that usually have a name and can be entered.**
> > To demarcate this category from others, the following are not included:

Pedant, does that mean you can't map kiosks as a shop? As an end user, you don't enter them. I can't think of a better simple wording though (apart from usually can be entered).

> what is still omitted are things that are rather like landuses, such as parks, gardens, golf courses, but also marketplaces, miniature golf courses, zoos, theme parks, military bases etc.
> 
> So, some things that may be called "places" are still missing from this overlay even though they can be entered, but I found it weird to add it into this category, or what do you think?

It seems a bit of a shame we can't just drop a pin on the map to say this is here. I appreciate the footprint is probably more important with these than a building.

>     * **tourism=** camp_pitch, camp_site, caravan_site, picnic_site, theme_park, zoo

It seems a huge shame and rather inefficient that the camp ones (which would trigger additional SC quests) can't be added to the map from within SC.

>     * [ ]  it should **_not_** be possible to change the type of an existing POI. Either delete it or mark it as disused (but I doubt the usefulness of that).

It seems a bit odd I can do bench->picnic table in the quest, but not in this new overlay, or e.g. phonebox re-purposed to sharing library.

Likewise I've found disused post boxes and phoneboxes quite often. The thing is still there, but it doesn't serve its usual purpose.

> So what do you think about the rename to "Things"?

It seems a reasonable compromise to me, again I can't think of a different all-encompassing name and it fits well with places.

> > Currently not included are `dolphin`, `dovecote` and `monitoring_station`.
> 
> I'd like to see `man_made = dovecote` being included, I don't know much about the other two.

I'm pretty certain I've seen dolphins, but didn't know they were called that.

app/src/main/java/de/westnordost/streetcomplete/osm/StreetFurniture.kt
peternewman, line 1: Will this and the other files get renamed when a name is settled on?

westnordost: yes

westnordost:
Yeah, you cannot enter a kiosk, but just like (street) food stalls that only have outdoor seating (if at all) I would still rather allocate them to places than things.

Regarding your remarks: It is unfortunate, but do you appreciate the reasoning behind that?

PR #5374, opened 2023-11-14
Ask crossing height if either `kerb:left` or `kerb:right` is missing

FloEdelmann:
I just noticed https://github.com/streetcomplete/StreetComplete/commit/9f501df0a3b2fea7b9d6dd0f54143465e748ba34. The commit message ("do not ask for crossing kerb height if kerb:left ***and*** kerb:right are set", emphasis mine) does not match its implementation. If the current implementation is correct, please just close this PR. Otherwise, this PR changes the implementation to match the commit message.


westnordost:
Oh, you are right! Thank you, @FloEdelmann for this fix!

PR #5375, opened 2023-11-15
Ask "Does this excrement bag dispenser have a bin?"

### LINKED ISSUE #4370: Quest suggestion: Does this poop bag dispenser have a trash can?
### General
Affected tag(s) to be modified/added: [`bin`](https://wiki.openstreetmap.org/wiki/Key:bin)
Question asked: **Does this poop bag dispenser have a trash can?**

### Checklist

### Links

* https://taginfo.openstreetmap.org/tags/vending=excrement_bags#combinations
* https://wiki.openstreetmap.org/wiki/Tag%3Avending%3Dexcrement_bags

### Ideas for implementation

**Element selection:**

```
nodes with
  vending = excrement_bags
  and (!bin or (bin = no and bin older today -12 years))
```

**Metadata needed:** None

**Proposed UI:** Simple Yes/No quest

qugebert:
Closes #4370


app/src/main/java/de/westnordost/streetcomplete/quests/excrement_bag_dispenser/AddExcrementBagDispenserHasBin.kt
```diff
@@ -0,0 +1,64 @@
...
+          and access !~ private|no
+    """.toElementFilterExpression() }
+
+    private val nearbyBinsFilter by lazy { """
+        nodes with
+          (
+          amenity = waste_basket
+          or bin = yes
+          )
+          and access !~ private|no
```
FloEdelmann, line 31:
```suggestion
        nodes with
          (amenity = waste_basket or bin = yes)
          and access !~ private|no
```

app/src/main/res/values/strings.xml
```diff
@@ -1043,6 +1043,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
 
     <string name="quest_gritBinSeasonal_title">Is this also here outside of winter?</string>
 
+    <string name="quest_excrement_bag_dispenser_bin_title">Does this excrement bag dispenser have a bin?</string>
```
FloEdelmann, line 1046: Maybe being a bit more colloquial is okay/preferable?

```suggestion
    <string name="quest_excrement_bag_dispenser_bin_title">Does this poop bag dispenser have a bin?</string>
```

qugebert: Sure, this is fine for me too

westnordost:
Hm sorry, the linked issue ticket slipped under my radar. Generally, only features and new quests with `new quest` or `enhancement` have the "stamp of approval" in the sense that all doubts have been cleared and a PR would be accepted. I will look through the linked issue now and comment there, first.

westnordost:
This PR can stay open until the discussion is over. However, I would not work on it further until then.

PR #5382, opened 2023-11-21
[Quest] How power line is attached on this support?

### LINKED ISSUE #4825: Line attachments on power poles/towers
### General
Affected tag(s) to be modified/added: [line_attachment](https://wiki.openstreetmap.org/wiki/Key:line_attachment)
Question asked: **How power line is attached on this support?**

### Checklist
- [X] 🐿️ Easily answerable by any pedestrian from the outside but a survey is necessary

### Ideas for implementation

**Element selection:** Attachments may be complex and sometimes a single support may have different attachments leading to complex tagging notation like line_attachment=(suspension|pin|suspension)|(anchor). That's not what we expect for streetcomplete quest.
We should only select supports with:
* no line_attachment=* key.
* only one line passing by power supports

**Metadata needed:** This quest is worldwide and no local metadata applies.

**Proposed UI:**
Drawings are took from wiki and could be enhanced/redrawn to meet streetcomplete's standards. We could use photos as well.
streetcomplete_attachment

Jean-BaptisteC:
Fixes #4825

Todo:
- [x] Improve translations (en and fr)
- [x] Add graphics (photos or SVG)


app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
```diff
@@ -0,0 +1,42 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BUILDING
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddPowerAttachment : OsmFilterQuestType<PowerAttachment>() {
+
+    override val elementFilter = """
+        nodes with
+          (power = tower or power = pole or power = insulator)
```
FloEdelmann, line 16:
```suggestion
          power ~ tower|pole|insulator
```

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val wikiLink = "Key:line_attachment"
+    override val icon = R.drawable.ic_quest_power
+    override val achievements = listOf(BUILDING)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_powerAttachment_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry): Sequence<Element> {
+        val mapData = getMapData()
+        // and also show the (power) lines themselves
+        return mapData.filter("nodes with power = tower or power = pole or power = insulator") +
```
FloEdelmann, line 29:
```suggestion
        return mapData.filter("nodes with power~tower|pole|insulator") +
```

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/PowerAttachmentItem.kt
```diff
@@ -0,0 +1,21 @@
...
+import de.westnordost.streetcomplete.quests.powerpoles_material.PowerAttachment.ANCHOR
+import de.westnordost.streetcomplete.quests.powerpoles_material.PowerAttachment.PIN
+import de.westnordost.streetcomplete.view.image_select.Item
+
+fun PowerAttachment.asItem() = Item(this, iconResId, titleResId)
+
+private val PowerAttachment.titleResId: Int get() = when (this) {
+    SUSPENSION ->     R.string.quest_powerAttachment_suspension
+    ANCHOR ->    R.string.quest_powerAttachment_anchor
+    PIN -> R.string.quest_powerAttachment_pin
```
FloEdelmann, line 14:
```suggestion
    SUSPENSION -> R.string.quest_powerAttachment_suspension
    ANCHOR -> R.string.quest_powerAttachment_anchor
    PIN -> R.string.quest_powerAttachment_pin
```

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/PowerAttachmentItem.kt
```diff
@@ -0,0 +1,21 @@
...
+}
+
+private val PowerAttachment.iconResId: Int get() = when (this) {
+    SUSPENSION ->     R.drawable.power_attachment_suspension
+    ANCHOR ->    R.drawable.power_attachment_anchor
+    PIN -> R.drawable.power_attachment_pin
```
FloEdelmann, line 20:
```suggestion
    SUSPENSION -> R.drawable.power_attachment_suspension
    ANCHOR -> R.drawable.power_attachment_anchor
    PIN -> R.drawable.power_attachment_pin
```

app/src/main/res/values/strings.xml
```diff
@@ -1287,6 +1287,11 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_postboxRoyalCypher_title">What’s the royal cypher on this postbox?</string>
     <string name="quest_postboxRoyalCypher_type_none">No royal cypher is visible</string>
     <string name="quest_postboxRoyalCypher_type_scottish_crown">The Crown of Scotland</string>
+	
+	<string name="quest_powerAttachment_title">How power line is attached on this support?</string>
+	<string name="quest_powerAttachment_suspension">Suspension</string>
+	<string name="quest_powerAttachment_anchor">Anchor</string>
+	<string name="quest_powerAttachment_pin">Pin</string>
```
FloEdelmann, line 1294: Please use spaces for indentation, not tabs. Also, I suggest this wording for the question:

```suggestion
    <string name="quest_powerAttachment_title">How is this power line attached?</string>
    <string name="quest_powerAttachment_suspension">Suspension</string>
    <string name="quest_powerAttachment_anchor">Anchor</string>
    <string name="quest_powerAttachment_pin">Pin</string>
```

matkoniecz:
I still have serious doubts about value of collecting this data (like for `power=pole` materials) but per https://github.com/streetcomplete/StreetComplete/issues/4825#issuecomment-1820961906 westnordost seems convinced.

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
qugebert, line 29: the power=line should get highlighted too.

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
```diff
@@ -0,0 +1,42 @@
...
+class AddPowerAttachment : OsmFilterQuestType<PowerAttachment>() {
+
+    override val elementFilter = """
+        nodes with
+          (power = tower or power = pole or power = insulator)
+          and !line_attachment
+    """
+    override val changesetComment = "Specify line_attachment power support"
+    override val wikiLink = "Key:line_attachment"
+    override val icon = R.drawable.ic_quest_power
```
qugebert, line 21: This quest should get a own icon.

Jean-BaptisteC: Why?
SC has similars quests with same the icon.
Do you have a better idea?

Jean-BaptisteC:
@flacombe any news?

Jean-BaptisteC:
I want finish this PR
These graphics is enough recognizable for you ?
- https://wiki.openstreetmap.org/wiki/File:Example_attachment_suspension.png
- https://wiki.openstreetmap.org/wiki/File:Example_attachment-anchor.png

mnalis:
>  These graphics is enough recognizable?

At desktop they do seem recognizable to me, but I'm not sure how they'd look scaled down to just a small part of (already small) mobile phone screen. Although looking at thumbnails at https://wiki.openstreetmap.org/wiki/Key:line_attachment  they might probably be fine.

westnordost:
> These graphics is enough recognizable?

Theoretically yes, after all the picture could be cropped a little to focus on what's important.
However, the minimum size is 384x384 (that's for 3x display density displays, the thumbnail would have a size of 128x128 pixels on a desktop) and the pictures are smaller than that.

Also, it would be preferable if anything but the supports was about the same on the pictures, i.e. similar sky background, similar perspective.

Jean-BaptisteC:
I have found better examples with good quality:
- [suspension](https://commons.wikimedia.org/wiki/File:Overhead_Power_-_geograph.org.uk_-_2037509.jpg)
- [suspension](https://commons.wikimedia.org/wiki/File:Power_tower_and_mobile_phone_base_station_2.jpg)
- [anchor](https://commons.wikimedia.org/wiki/File:Power-_tower.jpg)
- [anchor](https://commons.wikimedia.org/wiki/File:Power_tower_-_geograph.org.uk_-_2274819.jpg)

I continue to search an example for line_attachment=pin

@flacombe any suggestions?

Jean-BaptisteC:
I have take a new pictures during my trip in Canada

Pin


Anchor


Someone can help me to finish this PR?

flacombe:
Hello @Jean-BaptisteC and thank you for pictures.

To me it's all good and recognizable.
I'm not the most puzzled by such pictures though. How do others feel about them?

westnordost:
The difference is definitely recognizable. The first are some UFOs, the other are some springs. It's also nice that everything else about the pics is similar (the sky, the kind of pole), it highlights the differences.

Jean-BaptisteC:
The PR is ready, I have added pictures from my personal albums,  I have only added photos about power pole because I think the anchor is more easy to understand on pole than on big tower (on a too tiny photos)

app/src/main/res/values/strings.xml
```diff
@@ -1352,6 +1352,11 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_postboxRoyalCypher_title">What’s the royal cypher on this postbox?</string>
     <string name="quest_postboxRoyalCypher_type_none">No royal cypher is visible</string>
     <string name="quest_postboxRoyalCypher_type_scottish_crown">The Crown of Scotland</string>
+	
+    <string name="quest_powerAttachment_title">How is this power line attached?</string>
+    <string name="quest_powerAttachment_suspension">Supported from above</string>
+    <string name="quest_powerAttachment_anchor">Fixed horizontally</string>
+    <string name="quest_powerAttachment_pin">Supported from under</string>
```
FloEdelmann, line 1359:
```suggestion
    <string name="quest_powerAttachment_pin">Supported from below</string>
```

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachmentForm.kt
```diff
@@ -0,0 +1,13 @@
+package de.westnordost.streetcomplete.quests.power_attachment
+
+import de.westnordost.streetcomplete.quests.AImageListQuestForm
+
+class AddPowerAttachmentForm : AImageListQuestForm<PowerAttachment, PowerAttachment>() {
+
+    override val items = PowerAttachment.values().map { it.asItem() }
```
westnordost, line 7:
```suggestion
    override val items = PowerAttachment.entries.map { it.asItem() }
```

this is recommended in place of values() since Kotlin 1.9. (Don't know why)

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
westnordost: Because all quests have each a unique icon. But right now, I am unable to come up with a good idea for an icon.

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry): Sequence<Element> {
+        val mapData = getMapData()
+        // and also show the (power) lines themselves
+        return mapData.filter("nodes with power ~ tower|pole|insulator") +
+            mapData.filter("ways with power ~ line|minor_line")
+    }
+
+    // map data density is usually lower where there are power poles and more context is necessary
+    // when looking at them from afar
+    override val highlightedElementsRadius get() = 100.0
```
westnordost, line 35: How far apart are power towers usually (and at most)? If the radius is below that, it would mean that the power line wouldn't be highlighted in such cases.

westnordost (commented):
In `/app/src/main/res` there is a file named `authors.txt`. Could you please add the source, license and author in the form as they are supplied for the other graphics used within this app?

westnordost:
Since the photos are (in part) from your personal albums, you can choose a license by yourself, anything between PD, CC-0, CC-BY and CC-BY-SA would be fine. It would be convenient if the original photos were available somewhere, e.g. in Wikimedia commons or in the OpenStreetMap Wiki.

Finally, one more note: I noticed that one of the three pictures shows a power pole that is more zoomed out. It would be best if they had all the same zoom on the line attachment. The kind of line attachment is better to see, then.

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
Jean-BaptisteC: We have around 500 meters between each power tower

Jean-BaptisteC:
I have added photos on wikimedia and I have update authors.txt.
I have tried to resize suspension photo, but I didn't find the right solution to resize without too much crop top of the pole.

westnordost:
https://commons.wikimedia.org/wiki/File:Power_pole_anchor.jpg is a different picture than the one you use in StreetComplete.

> I have tried to resize suspension photo, but I didn't find the right solution to resize without too much crop top of the pole.

Hm? Why not like this?

imagen

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
Jean-BaptisteC: In such of case, we have 50m between poles, I have reused values from power material quest.

Jean-BaptisteC:
I have tried this solution but with hdpi and mdpi resolution, the sides of pole is too much cutted.

westnordost:
I don't get it. Sides of the pole too much cutted? Can you post a picture? Maybe then I understand what you mean.

I create the drawables by, in GIMP:
1. crop the original image to the square I need
2. resize the image to 384x384, save to xxhdpi, undo, resize to 256x256, save to xhdpi, undo, etc...

Jean-BaptisteC:
Files updated
Thanks for the tips, my solution is very too much complex 😅

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
westnordost: okay, so 100m is fine then.

westnordost:
Alright, note also this:

> https://commons.wikimedia.org/wiki/File:Power_pole_anchor.jpg is a different picture than the one you use in StreetComplete.

Jean-BaptisteC:
Already fixed on wikimedia commons 🙂
Ready to merge?

app/src/main/java/de/westnordost/streetcomplete/quests/power_attachment/AddPowerAttachment.kt
goldfndr: Perhaps a circle/disk/line at the bottom (and/or top) of the lightning bolt, making contact?

mnalis:
On small local [power=pole](https://wiki.openstreetmap.org/wiki/Tag:power=pole) nodes, I get lot of situations  like this: https://www.openstreetmap.org/note/4544696 which I am unsure how to map?

Jean-BaptisteC:
Select Fixed horizontally, these cases is not the most of case in OSM

PR #5383, opened 2023-11-23
Authorization with OAuth 2

### LINKED ISSUE #5322: Migrate to OAuth 2.0
**Use case**
OAuth 1.0a will be deprecated.
https://www.openstreetmap.org/user/pnorman/diary/401157

**Proposed Solution**
OAuth 2.0 is a lot simpler to implement, requires less maintenance and is generally stress free.

westnordost:
Implement authorization with OAuth 2. Fixes #5322

Blocked by decision to be made in https://github.com/openstreetmap/openstreetmap-website/issues/4360

Maybe @tsmock would review this ([OAuthAuthorization.kt](https://github.com/streetcomplete/StreetComplete/pull/5383/files#diff-6ae4dc39ffba677a18272a2ac850836f533b244d7b1f732aefb90fe6ca0547a2))?


app/src/main/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorization.kt
```diff
@@ -0,0 +1,203 @@
...
+ *  4. retrieveAccessToken(authorizationCode) with the retrieved authorizationCode
+ *
+ *
+ *  @param authorizationUrl the OAuth2 server authorization endpoint
+ *  @param accessTokenUrl the OAuth2 server token endpoint
+ *  @param clientId the OAuth2 client id
+ *  @param scopes the scopes (aka permissions) to request
+ *  @param redirectUri the redirect URI (aka callback URI) that the authorization endpoint should
+ *                     call when the user allowed the authorization.
+ * */
```
FloEdelmann, line 38:
```suggestion
/**
 * One authorization with OAuth. For each authorization request, a new instance must be created
 *
 * Authorization flow:
 *
 * 1. createAuthorizationUrl() and open it in a browser (or web view)
 * 2. let user accept or deny the permission request in the browser. Authorization endpoint will
 *    call the redirect URI (aka callback URI) with parameters in either case.
 * 3. check if the URI received is for this instance with itsForMe(uri) and then
 *    extractAuthorizationCode(uri) from that URI
 * 4. retrieveAccessToken(authorizationCode) with the retrieved authorizationCode
 *
 *
 * @param authorizationUrl the OAuth2 server authorization endpoint
 * @param accessTokenUrl the OAuth2 server token endpoint
 * @param clientId the OAuth2 client id
 * @param scopes the scopes (aka permissions) to request
 * @param redirectUri the redirect URI (aka callback URI) that the authorization endpoint should
 *                    call when the user allowed the authorization.
 */
```

see https://kotlinlang.org/docs/kotlin-doc.html

app/src/main/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorization.kt
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /**
+     * Extracts the authorization code from a parameter of the callback uri
+     *
+     * @throws AuthorizationException if the URI does not contain the authorization code, e.g.
+     *                                   the user did not accept the requested permissions
+     *  @throws ConnectionException if there has been an error that is the server's fault (try again
+     *                              later or open a bug report at openstreetmap-website if it
+     *                              persists)
```
FloEdelmann, line 91:
```suggestion
     * @throws ConnectionException if there has been an error that is the server's fault (try again
     *                             later or open a bug report at openstreetmap-website if it
     *                             persists)
```

app/src/main/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorization.kt
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     *  Retrieves the access token, using the previously retrieved [authorizationCode]
+     *
+     *  @throws IOException if an I/O exception occurs.
+     *  @throws AuthorizationException if there has been an OAuth authorization error
+     *  @throws ConnectionException if there has been an error that is the server's fault (try again
+     *                              later or open a bug report at openstreetmap-website if it
+     *                              persists)
+     */
```
FloEdelmann, line 117:
```suggestion
    /**
     * Retrieves the access token, using the previously retrieved [authorizationCode]
     *
     * @throws IOException if an I/O exception occurs.
     * @throws AuthorizationException if there has been an OAuth authorization error
     * @throws ConnectionException if there has been an error that is the server's fault (try again
     *                             later or open a bug report at openstreetmap-website if it
     *                             persists)
     */
```

westnordost:
The final shape this PR shall take depends on what decision is being made on https://github.com/openstreetmap/openstreetmap-website/issues/4360

If the checkboxes will return for the OAuth 2.0 implementation, will need to read and handle the `scope` response in the redirect uri. If not, I have to decide whether to always request "write gpx" permission or to dynamically request it when first tapping on the "create gpx track" menu item.

app/src/main/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorization.kt
```diff
@@ -0,0 +1,190 @@
...
+    private val clientId: String,
+    private val scopes: List<String>,
+    private val redirectUri: String
+) {
+    /** For the code challenge as specified in RFC 7636
+     *  https://www.rfc-editor.org/rfc/rfc7636
+     *
+     *  and required in the OAuth 2.1 draft
+     *  https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-09
+     */
```
FloEdelmann, line 51:
```suggestion
    /**
     * For the code challenge as specified in RFC 7636
     * https://www.rfc-editor.org/rfc/rfc7636
     *
     * and required in the OAuth 2.1 draft
     * https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-09
     */
```

app/src/test/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorizationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    @Test fun extractAuthorizationCode() {
+        assertEquals(
+            "my code",
+            createOAuth().extractAuthorizationCode(URI("localhost://oauth?code=my%20code"))
+        )
+    }
+
+    // it's not properly possible to test retrieveAccessToken in isolation because the http client
+    // is not injected (passed in the constructor)
```
tsmock, line 112: Have you looked at https://wiremock.org/ ?
JOSM uses it to "mock" network responses for tests.

app/src/main/java/de/westnordost/streetcomplete/data/user/UserModule.kt
```diff
@@ -1,34 +1,31 @@
...
-import se.akerfeldt.okhttp.signpost.OkHttpOAuthConsumer
-import se.akerfeldt.okhttp.signpost.OkHttpOAuthProvider
 
-private const val BASE_OAUTH_URL = "https://www.openstreetmap.org/oauth/"
-private const val CONSUMER_KEY = "NV4cEoqQ94Kuoowh8qGJvUJLnbts40WiNykyeC1T"
-private const val CONSUMER_SECRET = "r68v1Bd7RewTixAp0dMdCwn3w5iQvmpk4HlJcH2y"
-private const val CALLBACK_SCHEME = "streetcomplete"
-private const val CALLBACK_HOST = "oauth"
+const val OAUTH2_TOKEN_URL = "https://www.openstreetmap.org/oauth2/token"
+const val OAUTH2_AUTHORIZATION_URL = "https://www.openstreetmap.org/oauth2/authorize"
```
tsmock, line 6: This wasn't present when I wrote the initial implementation for JOSM, but the OSM website now supports [rfc8414](https://datatracker.ietf.org/doc/html/rfc8414) (see https://www.openstreetmap.org/.well-known/oauth-authorization-server ).

If I were writing an implementation today, I _would not_ hardcode the OSM OAuth URLs. And I'm kind of planning on removing the hardcoded OAuth URLs in JOSM.

tsmock (commented):
It looks good to me -- I don't see anything that jumps out as obviously wrong, anyway.

app/src/test/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorizationTest.kt
westnordost: I haven't, thanks for the link!

app/src/main/java/de/westnordost/streetcomplete/data/user/UserModule.kt
westnordost: But why would you not hardcode them? I get that when writing a generic OAuth client, it would be very wise to support rfc8414 but since this only speaks with OSM, it's basically one hardcoded URL (https://www.openstreetmap.org) vs two (auth url + token url).

(Also, at the time of implementation, the testing instance didn't have the .well-known file. Not sure if it changed.)

tsmock: > Also, at the time of implementation, the testing instance didn't have the .well-known file. Not sure if it changed.

The testing instance does have a .well-known file now ( https://api06.dev.openstreetmap.org/.well-known/oauth-authorization-server ).

> But why would you not hardcode them?

As a general principle, I try to avoid hardcoding things that can be obtained through standardized means (assuming the standard is reasonable). As an example, website <foo> might decide that they want to move the `/oauth2` endpoints to `/oauth` when they remove OAuth 1.0 support. I _don't_ think OSM will do that, so it probably isn't a "big" issue.

westnordost: I see, thanks!

By the way, https://github.com/openstreetmap/openstreetmap-website/issues/4360 may affect JOSM implementation, too:

According to the specification, a `scope` parameter is returned from the redirect URI which contains the actually granted scopes by the user. Meaning, according to the specification, a user may be allowed to not grant all permissions of what has been requested by the client. The app can/should handle that somehow, either rejecting the access token if it does not have all required scopes granted and/or dealing with the missing permission in the UI.
(That is, if OSM website maintainers decide that they will consider making use of this in the future. But I understand you like future-proofing JOSM for possible RFC compliant changes to the OAuth 2 API on OSM)

westnordost:
By the way, thank you @dirkmueller from SUSE for sponsoring this, I also mentioned you in the changelog for the upcoming version.

PR #5387, opened 2023-11-25
Don't ask pitch surface for high jump or pole vault

arrival-spring:
Prompted by https://community.openstreetmap.org/t/how-to-tag-a-high-jump-mattress/106241

Looking at https://taginfo.openstreetmap.org/keys/athletics#values I think the rest make sense to ask for surfaces if such pitches exist.

I just made the change through the browser, so untested.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
@@ -29,6 +29,7 @@ class AddPitchSurface : OsmFilterQuestType<SurfaceAndNote>() {
          and sport ~ "(^|.*;)(${sportValuesWherePitchSurfaceQuestionIsInteresting.joinToString("|")})($|;.*)"
          and (access !~ private|no)
          and indoor != yes and (!building or building = no)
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddPitchSurface.kt
```diff
@@ -29,6 +29,7 @@ class AddPitchSurface : OsmFilterQuestType<SurfaceAndNote>() {
          and sport ~ "(^|.*;)(${sportValuesWherePitchSurfaceQuestionIsInteresting.joinToString("|")})($|;.*)"
          and (access !~ private|no)
          and indoor != yes and (!building or building = no)
+         and (athletics !~ high_jump|pole_vault)
```
matkoniecz, line 32: do we need brackets here?

matkoniecz (approved):
though I have not run the code

matkoniecz:
@arrival-spring Have you run the code to confirm it works?

HolgerJeromin:
> I just made the change through the browser, so untested.

You can start a build in github gui and download an apk for test on a device.

PR #5391, opened 2023-11-27
Add hint locales for localized name input

### LINKED ISSUE #5351: Use hint locales for multi-language name input
**Use case**
When entering names in multiple languages, it may be preferable or even necessary to switch keyboard language. StreetComplete could help here by providing the desired language, so the keyboard can switch automatically.

**Proposed Solution**
When the name is for a specific locale, provide [hint locales](https://developer.android.com/reference/android/widget/TextView#setImeHintLocales(android.os.LocaleList)).
When the name is international or definitely requires latin characters, there is a [FORCE_ASCII](https://developer.android.com/reference/android/view/inputmethod/EditorInfo#IME_FLAG_FORCE_ASCII) flag that can be set.

Helium314:
fixes #5351 

The behavior is already described in https://github.com/streetcomplete/StreetComplete/issues/5351#issuecomment-1826262994
`input.imeHintLocales` is only available starting at API24.
`restartInput` is necessary for the changed hint locales to get applied immediately.
`locale.toString().isEmpty()` is used for checking whether a valid `Locale` was created (using `Locale.forLanguageTag`). Maybe there are better ways to test this...

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/LocalizedNameAdapter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/LocalizedNameAdapter.kt
@@ -2,12 +2,17 @@ package de.westnordost.streetcomplete.quests
 
 import android.content.Context
 import android.graphics.Typeface
+import android.os.Build
+import android.os.LocaleList
+import android.text.InputType
 import android.view.LayoutInflater
 import android.view.Menu.NONE
 import android.view.View
 import android.view.ViewGroup
+import android.view.inputmethod.InputMethodManager
 import android.widget.EditText
 import android.widget.TextView
+import androidx.annotation.RequiresApi
 import androidx.appcompat.widget.PopupMenu
 import androidx.core.view.isGone
 import androidx.core.view.isInvisible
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/LocalizedNameAdapter.kt
```diff
@@ -343,6 +351,16 @@ class LocalizedNameAdapter(
...
         }
+
+        @RequiresApi(Build.VERSION_CODES.N)
+        private fun updateHintLocales(locale: Locale) {
+            if (locale.toString().isEmpty()) {
+                input.imeHintLocales = null
+            } else {
+                input.imeHintLocales = LocaleList(locale)
+            }
+            context.getSystemService(InputMethodManager::class.java).restartInput(input)
```
westnordost, line 362: Doesn't
```suggestion
            context.getSystemService<InputMethodManager>().restartInput(input)
```
work?

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/LocalizedNameAdapter.kt
Helium314: Sure it does.

Issue #2074, opened 2020-09-04
New Quest: Check if an object is still there


westnordost:
Question asked: **Is this X still here?**

So with the maintenance quests in place, the next step could be to have a or have several quests where the user is asked to confirm if an object still exists.
StreetComplete does currently not support deleting elements, so this must be implemented in the frame of this issue or before that.

The interface will be a simple yes/no button. If it still exists, `check_date` or maybe `check_date:existance` should be tagged. It should be discussed which of the two tags should be used.

Till the quest is ready to be implemented, feel free to add more ideas of objects whose existance should be rechecked every now and then.

### For the following objects

very often because it is temporary by definition (1 year?):
- `landuse=construction`

often because it is small and not a big thing to remove (2 years?):
- `amenity=post_box`
- `amenity=telephone`
- `amenity=clock`
- `amenity=vending_machine` (less often for ticket vending machines)
- `amenity=bench`
- `amenity=atm`
- `leisure=picnic_table`
- `amenity=waste_basket`
- `tourism=information` + `information=board` or `terminal` or `map`

often because it is important that it is up-to-date
- `emergency=phone`
- `emergency=defibrillator` (only for those where it is known where they are - inside or outside, on which level)
- `emergency=fire_extinguisher` (only for those where it is known where they are - inside or outside, on which level)

a little less often (4 years?):
- `amenity=bicycle_parking` (only for stands, wall_loops etc)
- `amenity=motorcycle_parking`
- `leisure=pitch` + `sport=table_tennis`
- `advertising=column` or `board` or `poster_box`
- `highway=path` with ground surface and bad visibility. How to determine bad visibility though?
- most `shop`s
- `amenity=pharmacy` maybe more often because it is more important

**Post here if you have more ideas!**

matkoniecz:
`amenity=restaurant` `=pub` `=cafe` `=fast_food` `=bank` `=car_sharing` `=bicycle_rental`

Maybe it would make to sense to ask about `disused:amenity` whatever this objects are now fully gone? https://taginfo.openstreetmap.org/keys/disused%3Aamenity

It also shows what tends to disappear.

FloEdelmann:
When `landuse=construction` is answered as “does not exist anymore”, instead of deleting the node/area, wouldn't it make sense to only update the `landuse` tag to the value of the `construction` tag?

[...]

PR #5400, opened 2023-12-07
Ask if amenity=car_sharing is still there

### LINKED ISSUE #5399: Ask if `amenity=car_sharing` is still there
It has been briefly mentioned in the original "is it still there" issue (#2074) but hasn't been implemented or discussed any further.

I assume it would be quite easy to add it to the "Is x still there" quest.

> I've just opened a few notes for car sharing stations which are probably gone. Would be useful to add this imo.

_Originally posted by @dreua in https://github.com/streetcomplete/StreetComplete/issues/2074#issuecomment-1841807400_

catdogmat:
It has been briefly mentioned in the original "is it still there" issue (https://github.com/streetcomplete/StreetComplete/issues/2074) but hasn't been implemented or discussed any further.

> I've just opened a few notes for car sharing stations which are probably gone. Would be useful to add this imo.

Originally posted by @dreua in https://github.com/streetcomplete/StreetComplete/issues/2074#issuecomment-1841807400
This PR fixes #5399

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
@@ -35,6 +35,7 @@ class CheckExistence(
             or amenity = post_box
             or leisure = picnic_table
             or amenity = bbq
+            or amenity = car_sharing
             or leisure = firepit
             or (leisure = pitch and sport ~ table_tennis|chess|table_soccer|teqball)
             or leisure = fitness_station


dreua:
Ha, I had no idea it would be *that* simple!

westnordost:
But why check every 4 years? I think it should be moved to line 68.

catdogmat:
> But why check every 4 years? I think it should be moved to line 68.

They come and go quite alot where I live, as not many people use them.

westnordost:
Well, fair enough. I was thinking that this is the same class of things as e.g. a taxi stand or a motorcycle parking, which is in the 6-year category. But, let's not count peas.

matkoniecz:
(also, it is possible to look at how people answered "is it still there" quests, look how old objects were at time of quest and tweak quest configuration based on that)

PR #5403, opened 2023-12-16
Improve wording of contribute quest docs

arrival-spring:
I've read through and adjusted anything that didn't sound right, as a native English speaker.

Mostly minor thing like missing articles or word order.

### DIFF
--- a/CONTRIBUTING_A_NEW_QUEST.md
+++ b/CONTRIBUTING_A_NEW_QUEST.md
@@ -2,9 +2,9 @@ This file describes in detail the process of creating a new quest.
 
 For code style and more general info - see [CONTRIBUTING file](CONTRIBUTING.md#development).
 
-If you want to contribute code to StreetComplete, then making a new quest is one of the easiest programming tasks. It is quite easy to implement a quest where layout design matches an existing quest. For example one more yes/no quest or where user taps one of displayed images.
+If you want to contribute code to StreetComplete, then making a new quest is one of the easiest programming tasks. It is quite easy to implement a quest where the layout design matches an existing quest. For example, one more yes/no quest or where the user taps a displayed image.
 
-Contributions like that are highly welcomed and you would make mapping one more thing in OSM much easier! You can implement it also if you never used Kotlin or implemented anything for Android. Being highly experienced programmer is not necessary here.
+Contributions like that are highly welcomed and you would make mapping one more thing in OSM much easier! You can also implement it if you have never used Kotlin or implemented anything for Android. Being a highly experienced programmer is not necessary here.
 
 Reading the text below is not necessary to create a new quest. Duplicating an existing quest and modifying its code may be sufficient - people were creating new quests before this documentation existed. This material simply helps you get a better understanding of it.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING_A_NEW_QUEST.md
```diff
@@ -64,21 +64,21 @@ Use <kbd>Ctrl</kbd>+<kbd>N</kbd> to go to a certain class. <kbd>Ctrl</kbd>+ mous
...
 
-Are you trying to implement a quest that will have simple yes/no answer? Take [`AddBusStopLit`](app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_lit/AddBusStopLit.kt) quest as a base. Or [`AddTracktype`](app/src/main/java/de/westnordost/streetcomplete/quests/tracktype/AddTracktype.kt) where the mapper will be selecting one of presented images.
+Are you trying to implement a quest that will have a simple yes/no answer? Take the [`AddBusStopLit`](app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_lit/AddBusStopLit.kt) quest as a base. Or [`AddTracktype`](app/src/main/java/de/westnordost/streetcomplete/quests/tracktype/AddTracktype.kt) where the mapper will be selecting from one of some presented images.
 
-Is it going to be asked for POIs and should be disabled by default? [`AddWheelchairAccessBusiness`](app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt) may be a good base.
+Is it going to be asked for POIs and it should be disabled by default? [`AddWheelchairAccessBusiness`](app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt) may be a good base.
```
riQQ, line 77:
```suggestion
Is it going to be asked for POIs and should it be disabled by default? [`AddWheelchairAccessBusiness`](app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt) may be a good base.
```

arrival-spring: Actually I don't like the "it" I added now, I think I should revert that line change.

mnalis: > Actually I don't like the "it" I added now, I think I should revert that line change.

Agreed, although as a non-native speaker I always find such multiple-rules (_"is it XXX and should YYYY be ZZZZ?"_)  in question form (instead of if/then form) somewhat hard to parse. 

I.e. something like:
```suggestion
If it is going to be asked for POIs and should be disabled by default, then [`AddWheelchairAccessBusiness`](app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt) may be a good base.
```

or even better (if the emphasis is supposed to be on disabled-by-default part):

```suggestion
If the quest which is going to be asked for POIs should be disabled by default, then [`AddWheelchairAccessBusiness`](app/src/main/java/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt) may be a good base.
```

would be more readable to me.

westnordost:
LGTM. Tell me when I shall merge this

CONTRIBUTING_A_NEW_QUEST.md
arrival-spring: First one sounded better to me, so I've gone with that

arrival-spring:
Ready from my point of view

matkoniecz:
thanks for improving my English!

PR #5407, opened 2023-12-19
Fix `AddSanitaryDumpStation` quest filter

sun-geo:
Fix line 17 ("camping_site" to "camp_site")


FloEdelmann (approved):
Thanks! :slightly_smiling_face: 

For reference: https://github.com/streetcomplete/StreetComplete/pull/5363#issuecomment-1862133354

PR #5428, opened 2024-01-01
Update AddBikeParkingCapacity.kt to exclude bicycle_parking=informal

bxl-forever:
Users created notes because they could not assess capacity for informal bicycle stands (e.g. long metal bar, street furniture or fences).

Purpose of this pull request: if a device has "bicycle_parking=informal", the quest will not show up.


matkoniecz:
note: value is documented at https://wiki.openstreetmap.org/wiki/Key%3Abicycle_parking

merging this PR should not be treated as endorsing mapping such objects as a good idea

Issue #5378, opened 2023-11-18
Speed limit on trunk road in urban area


rhhsm:
Today I was surveying in Datça, Turkey, the start of the 2000 km trunk road D400. This [part of](https://www.openstreetmap.org/way/1163126410/history) the trunk road is within city limits, so it should get tagged `maxspeed:type=TR:urban` (50 km/h) when there are no speed limit signs. However, SC appears to assume that trunk roads never pass through urban areas, and sets `maxspeed:type=TR:trunk` (110 km/h) without asking if the road is in an urban area. It also doesn't ask the lit and sidewalk quests. I can understand the second, but it seems quite likely that at least parts of trunk roads may be lit.

SC should also ask if trunk roads are in an urban area to set the speed limit correctly.

westnordost:
I guess this may depend on the country. But generally I think you are right. Many/most legislations only bestow to motorways special speed limit regulations that are independent of whether the road is within an urban area or not.

PR #5430, opened 2024-01-02
Simplify max speed quest logic

FloEdelmann:
Since #5378, `ROADS_WITH_DEFINITE_SPEED_LIMIT` and `ROADS_WITH_DEFINITE_SPEED_LIMIT_GB` are identical. Thus, the more complicated implementation of #2811 and #2812 is no longer necessary and can be simplified.

@arrival-spring could you please confirm this?


arrival-spring:
Yes, this looks good to me

PR #5432, opened 2024-01-03
Also ask for opening hours of various indoor amenities

### LINKED ISSUE #5099: What are the opening hours of this parcel locker?
### General
Affected tag(s) to be modified/added: [opening_hours](https://wiki.openstreetmap.org/wiki/Key:opening_hours)
Question asked: **What are the opening hours of this parcel locker?**

### Checklist

I'm not entirely sure on this one, since some parcel lockers are located inside of malls or shops and require the user to check the respective shop's opening hours. It may not be obvious to all users so they might enter 24/7.

### Ideas for implementation

**Element selection:** 
`amenity=parcel_locker + !opening_hours`

**Metadata needed:** 
None

**Proposed UI:**
Similar to the quest for opening hours of shops.

qugebert:
This closes #5099

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -104,7 +104,7 @@ class AddOpeningHours(
           )
           or (
             (!opening_hours or opening_hours older today -1 years)
-            and indoor = yes
+            and (indoor = yes or addr:floor or level:ref or level)
             and ( """+
 
         // amenities inside a building (might depend on opening hours of the containing building),
@@ -132,7 +132,7 @@ class AddOpeningHours(
         and (
           name or brand or noname = yes or name:signed = no
           or barrier
-          or indoor = yes
+          or (indoor = yes or addr:floor or level:ref or level)
           or amenity ~ toilets|bicycle_rental
         )
         and opening_hours:signed != no
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -103,6 +103,23 @@ class AddOpeningHours(
             and (!opening_hours or opening_hours older today -1 years)
           )
           or (
+            (!opening_hours or opening_hours older today -1 years)
+            and indoor = yes
+            and ( """+
+
+        // amenities inside a building (might depend on opening hours of the containing building),
+        // changes made on this list should probably be adopted to the Element Selection of the AddIsAmenityIndoor-Quest
+        mapOf(
+            "amenity" to arrayOf(
+                "atm", "telephone", "parcel_locker", "luggage_locker", "locker", "post_box", "public_bookcase","give_box","ticket_validator","vending_machine"
+            ),
+            "emergency" to arrayOf(
+                "defibrillator", "fire_extinguisher", "fire_hose"
+            ),
+        ).map { it.key + " ~ " + it.value.joinToString("|") }.joinToString("\n or ") + "\n" +
+        """)
+          )
+          or (
             opening_hours older today -1 years
             and (
               leisure = park
@@ -115,6 +132,7 @@ class AddOpeningHours(
         and (
           name or brand or noname = yes or name:signed = no
           or barrier
+          or indoor = yes
           or amenity ~ toilets|bicycle_rental
         )
         and opening_hours:signed != no
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -108,7 +108,7 @@ class AddOpeningHours(
             and ( """+
 
         // amenities inside a building (might depend on opening hours of the containing building),
-        // changes made on this list should probably be adopted to the Element Selection of the AddIsAmenityIndoor-Quest
+        // changes made to this list should probably also be adopted in the AddIsAmenityIndoor quest's element selection
         mapOf(
             "amenity" to arrayOf(
                 "atm", "telephone", "parcel_locker", "luggage_locker", "locker", "post_box", "public_bookcase","give_box","ticket_validator","vending_machine"


app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -102,6 +102,23 @@ class AddOpeningHours(
             )
             and (!opening_hours or opening_hours older today -1 years)
           )
+          or (
+            (!opening_hours or opening_hours older today -1 years)
+            and (indoor = yes or addr:floor or level:ref or level)
+            and ( """+
+
+        // amenities inside a building (might depend on opening hours of the containing building),
+        // changes made on this list should probably be adopted to the Element Selection of the AddIsAmenityIndoor-Quest
```
FloEdelmann, line 111:
```suggestion
        // changes made to this list should probably also be adopted in the AddIsAmenityIndoor quest's element selection
```

westnordost:
Hmm, I just re-read #5099 and then read the implementation.

In #5099 I noted that it should only be asked for indoor facilities to avoid confusion. However, after contemplation, I noticed that there is space for confusion for both alternatives (only showing outside <> only showing inside facilities):

**Showing outside facilities:**
- maybe spammy: usually they are open 24/7.
- usually there is no sign. The very question may be confusing.

**Showing inside facilities:**
- usually there is **also** no sign and hence the very question may also be confusing. It actually is even worse for inside facilities because answering "always open" will actually be mistake!
- it is more difficult to ascertain: you have to go back out and look for the opening hours there. Depending on the size (of the mall) it is also somewhat more cumbersome. So, when asked to confirm if the opening hours are still correct, users may feel lazy and just say "yeah, whatever, probably it is", also potentially creating wrong data.

So actually, if not mitigated, asking for inside facilities seems to be actually more problematic than for outdoor ones.

So, mitigation. Maybe it can be solved in UI. When something is `indoor`, a short note could be shown in the quest, informing that the opening hours of that facility are likely ultimately the opening hours of the building it is contained within and that the user needs to check there.
This does not mitigate the point about user laziness, but then again, at least he is warned, so he may think twice to just tap it away.

westnordost:
Suggestion/draft for text: "Note: The opening hours may be linked to the opening hours of the building it is contained in"

mnalis:
Hm, I'm not actually sure it is a good idea. 

- Are `indoor=yes` + `emergency=fire_hose` and similar really to be tagged with `opening_hours`? It was my understanding that fireman will extinguish fires even outside working hours, breaking doors if needed for access to needed fire hoses / outlets / etc. 

Also, for the rest (non-emergency-services related, e.g. things like `amenity=vending_machine`), I would assume that in vast majority of situations their `opening_hours` is the same as the space they are enclosed in (e.g. `shop=mall` polygon or whatever). Re-tagging those some opening hours for each and every element inside is bad match for SC **Quest** because:

- it's not being very considerate of mapper's time (i.e. _"💵 Valuable Surveyors"_ tenet), especially as it is not simple `yes/no` quest, but significantly more involving
- in my experience, in vast majority of the cases, the opening_hours is implied by the shop/mall/whatever it is situated in, so this fails _"💤 No spam"_ too
- more trouble than it is worth, esp. taking into consideration the maintenance of the data. 
  E.g. in Croatia, only few working Sundays are allowed per year for shopping centers, so `opening_hours` of a `shop=mall` are relatively easily edited few times per year as those Sundays become known by using e.g. EveryDoor on location. If every locker, vending machine, ATM etc. inside must be rettaged also, it becomes much bigger ordeal, with much more potential of stale data.

---

What I would suggest instead, is simply **skipping the `opening_hours` quest if the amenity is inside** (`indoor = yes or addr:floor or level:ref or level`). That way, the problem of users inadvertently adding `opening_hours=24/7` is avoided (the crux of https://github.com/streetcomplete/StreetComplete/issues/5099 as I see it), but abovementioned problems are not introduced. 

If there are some few rare cases where there is e.g. `amenity=atm` or `amenity=vending_machine` situated inside `shop=mall` halls (or whereever) but having different `opening_hours` from it, I'd say EveryDoor is much better solution for such uncommon situations (or hypothetical future StreetComplete OpeningHours Overlay)  than StreetComplete Quest.

westnordost:
Hm, I don't like to go back on a decision I made earlier when there has already been work spent that is based on it, but at this point I have to admit that @mnalis has a point. So to summarize your suggestion, it would be: only show for `indoor != yes` and do not show for emergency equipment?

mnalis:
> Hm, I don't like to go back on a decision I made earlier when there has already been work spent that is based on it, but at this point I have to admit that @mnalis has a point

I too would much prefer if I managed to spot this earlier to avoid wasted effort, but my backlog at the moment is too big to cope properly will significant part of discussions where I'm not explicitly mentioned; sorry people :cry:  

> So to summarize your suggestion, it would be: only show for indoor != yes and do not show for emergency equipment?

Actually, to the best of my understanding, the whole idea should probably be `wontfix`ed; because:

- if the `amenity=parcel_locker` (or `amenity=atm` etc.) is **outside**, than as noted in https://github.com/streetcomplete/StreetComplete/issues/5099#issuecomment-1610909406 in vast majority of the cases it will have implied `opening_hours=24/7`, so it is counterproductive to tag it explicitly (wastes valuable surveyor time / is spammy).
- if the `amenity=parcel_locker` (or `amenity=atm` etc.) is **inside** (`indoor=yes`), then as noted [above](https://github.com/streetcomplete/StreetComplete/pull/5432#issuecomment-1891247588) in vast majority of the cases it will have same implied `opening_hours=*` as the amenity it is situated in (e.g. `shop=mall` or whatever).

  In that case, it is also counterproductive to tag it explicitly, as it not would not only waste valuable surveyor time / be spammy (as it is for the **outside** case), but **in addition** would also significantly increase maintenance effort and chances of outdated opening hours -- for no particularly good reason [^1]

The `emergency=fire_hose` and friends are just additional thing where it IMHO never makes sense to have `opening_hours` ; but since two points above make it IMHO undesirable to ask for opening_hours for both indoor and outdoor instances of _any_ of the POIs mentioned, it does not even need special-casing.

Perhaps further discussion might reveal some niche situations in which it would bring more advantages than disadvantages, but I'm not sure if it would be worth persuing (especially in StreetComplete).

[^1]: there is a case of making life somewhat easier for data consumers: e.g. if it were accepted globally to microtag `opening_hours` on _every_ POI inside a polygon which has `opening_hours`, they would only need to check `opening_hours` of a point and not of point+polygon-it-is-in; but the drawbacks seems to be much too big. Having the data consumer fetch the data themselves, either directly in the app or during preprocessing step seems much preferred to me as it gets all advantages and no disadvantages)

---

**TL;DR**: it feels to me like this PR and related issue are for `close` / `wontfix` resolution as I don't see a way that would help more than it would harm; sorry

qugebert:
Ok, so should i add a hint now or would you just close this?

PR #5433, opened 2024-01-05
Show name and ref as label for roads

### LINKED ISSUE #5427: Just show name and ref as description for roads
For roads, I think it would be better if the `NameAndLocationLabel` showed the `name` and` ref` in preference to the `operator`.

E.g. here, everyone would know this as the "M6" (`ref`) so it should be shown as "Preston Bypass [M6]"
Screenshot_20240101-192620~2

And here simply "M6" instead:
Screenshot_20240101-192340~2

So for roads I think it should be
`name [ref]` or `name` (if no ref) or `ref` (if no name) and never show `operator`.

I'm asking rather than just opening a PR because I don't know if the operator of roads would actually be helpful information to show in any other countries. But even so, I would think that ref is more important than operator for roads.

arrival-spring:
Fixes #5427  as that seems to have support, and no comments against.

Now:
Screenshot_20240105-123040~2

And:
Screenshot_20240105-123111~2

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
@@ -80,6 +80,28 @@ class NameAndLocationLabelTest {
         )))
     }
 
+    @Test fun roadWithName() {
+        assertEquals("Main Street (Residential Road)", getQuestLabel(mapOf(
+            "highway" to "residential",
+            "name" to "Main Street",
+        )))
+    }
+
+    @Test fun roadWitRef() {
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
FloEdelmann, line 1: I think the test cases should include the `operator`, so it's actually tested that the generated label doesn't contain that operator.

arrival-spring: Yes, of course, thanks. I've changed it now

FloEdelmann (approved):
Looks good to me :)

matkoniecz (approved):

Issue #5415, opened 2023-12-20
Multiplatform config file parsing


westnordost:
In various places, the app loads and parses some simple YAML files for configuration:

- a dictionary of street name abbreviations with their expansions (`abbreviations.yml`)
- a dictionary of version number to html fragments that are the changelog (`changelog.yml`)
- a list of main contributors (`credits_main.yml`)
- a list of side project contributors (`credits_projects.yml`)
- a list of all code contributors on github (`credits_contributors.yml`)
- a list of all art contributors (`credits_art.yml`)
- a dictionary of languages -> translator name -> number of translated words (`credits_translators.yml`)
- a dictionary of flag alignments by country code (`flag_alignments.yml`)
- a dictionary of various country properties (`<country code>.yml`)

#### keep YAML

Currently, [`com.charleskorn.kaml`](https://github.com/charleskorn/kaml) is used which is a plugin for `kotlinx-serialization`, however, no native targets are supported currently and noone is working on it at the time of writing (see https://github.com/charleskorn/kaml/issues/232). 

Another kotlin multiplatform yaml plugin exists, [`net.mamoe.yamlkt`](https://github.com/Him188/yamlkt), but according to the readme, it is (still) in alpha state. Multiline strings, which are currently used in `changelog.yml`, are "partially supported" since v0.13.0, not sure if this means whether the way multiline strings are used in this project is supported. (`changelog.yml` could be changed to be a pure HTML file, though, or one file for every version.)

If none of these libraries have proper multiplatform native support at the time this ticket is tackled, the task is to convert the listed files to another config file format like JSON or TOML and parse them as these files instead.

Some of these files are generated automatically, such as `credits_contributors.yml`, so some build scripts need to be adapted, too. Below some notes regarding each:

#### convert to TOML

- \+ well-human-readable format, easy to edit

- \- kotlinx-serialization plugin [`com.akuleshov7-ktoml-core`](https://github.com/akuleshov7/ktoml) is multiplatform but implementation is incomplete. In particular, "Array of Tables" feature is missing as the time of writing, which is necessary for `credits_translators.yml`

- \+ kotlinx-serialization plugin [`net.peanuuutz.tomlkt`](https://github.com/Peanuuutz/tomlkt) is multiplatform and is actually complete, looks well tested and has a useful interface. It's rather fresh at the time of writing, though (meaning, we don't know how well tested it is in production and how committed the author is to maintain it).

[...]

Issue #3355, opened 2021-10-09
Changelog info text in F-droid


aha999:
**Use case**

Currently, there is no visible window with the latest changes when you open your app's page in F-Droid.

**Proposed Solution**

Add changelog text when you open the fdroid app repo

**Some tips:**

> F-Droid looks for metadata in the app’s source repo. The supported formats for this metadata are borrowed from popular projects that automate uploading this metadata to Google Play Store.

>The one in your example corresponds to Fastlane Supply and is documented [here](https://docs.fastlane.tools/actions/supply/#changelogs-whats-new). All formats supported by F-Droid are listed [here](https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-applications-source-repository)

matkoniecz:
I think that it is a bug caused by F-Droid and should be reported on F-Droid repository.

Is there some reason to suspect that it is bug caused by StreetComplete itself? This works well when installed from Google Play, it works well when sideloaded.

Is F-Droid modifying StreetComplete in some way before releasing it?

aha999:
Maybe @eighthave would know how to answer this ?

FloEdelmann:
Wo could do it the other way around: Manually create one file per version in that location, and then combine them into the all-in-one file with a Gradle task.

westnordost:
And how to get from version code to version name then?

FloEdelmann:
Just leave it as a HTML heading in the individual files.

westnordost:
Hmm, that would be possible. I wonder what F-Droid does with `*.txt` files that are actually html (snippets) though.

westnordost:
Also, do you know an example app for which this changelog stuff is used? I would like to see how it is displayed because I am still not sure if this is worth it at all.

FloEdelmann:
HTML Markup seems to be possible in at least the app description: https://forum.f-droid.org/t/f-droid-support-fastlane-full-description-in-html-or-md-instead-of-txt/9395

westnordost:
What exactly are you planning to do and why?

westnordost:
> Imo markdown is the best option here. It's the easiest to update and copy-paste to different places (e.g. GitHub releases)

<h3>This</h3> is <i>HTML</i> and it
<ul>
<li>works</li>
</ul>
no problem

westnordost:
[...]

PR #5439, opened 2024-01-11
Move changelog from YAML to Markdown

FloEdelmann:
… and automatically convert them in a Gradle task to HTML strings in a JSON file.

* Part of #5415
* See also #3355


westnordost:
Oh, that's cool! The markdown also makes it more convenient to write changelogs from now on. The task should be part of the `updateSTreetCompleteData` job, that's the one I execute every time before release.

I wonder if it made more sense to have the changelog simply as a giant HTML and the changelog dialog just cuts off when it finds <h1>$older version number</h1>?

westnordost (commented):
Did you convert the changelog to markdown by hand or did you write a script? Can you post the source code?

I think it would be slightly more convenient (for adding a new version, but also for reading) if there was just one giant changelog.md source file that gets appended. That changelog.md could then live in the repo root, as is quite common, I think? And if it gets processed anyway for in-app display, it doesn't matter if it is files or one file.
(Though, yesterday night I proposed to keep it as a giant HTML file and not take it apart. I stand to it insofar as I do not think it will affect performance (or maybe even positively), plus, in the changelog fragment, the scrollbar would behave not erratically, which is also a plus)

The script looks fine.

FloEdelmann:
> The task should be part of the `updateSTreetCompleteData` job, that's the one I execute every time before release.

Done in [`d112d36`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/d112d3690e266b272fcb55ed6fb16d7866b91deb).

> Did you convert the changelog to markdown by hand or did you write a script? Can you post the source code?

I wrote a small Node.js script using an [existing module](https://www.npmjs.com/package/node-html-markdown):

```js
#!/usr/bin/env node

import { readFileSync, writeFileSync } from "fs";
import { NodeHtmlMarkdown } from "node-html-markdown";

const changelog = JSON.parse(readFileSync("changelog.json", "utf8"));

for (let [version, changes] of Object.entries(changelog)) {
    changes =
        NodeHtmlMarkdown.translate(changes, {
            bulletMarker: "-",
        }) + "\n";

    writeFileSync(`${version}.md`, changes);
}
```

Then I went over the generated files and adjusted them manually. I found some syntax errors and typos there, which I corrected in the `master` branch in https://github.com/streetcomplete/StreetComplete/commit/425836ad09ecd63eb0c2ad1a26a2edc326842714.

> That changelog.md could then live in the repo root, as is quite common, I think?

You're right, I also like this approach. I'll implement that now.

> Regarding F-Droid changelogs

Thanks for the info. I only linked that issue for reference though; I don't intend to generate those in this PR.

westnordost:
> Thanks for the info. I only linked that issue for reference though; I don't intend to generate those in this PR.

Right, okay, I also do not care that much. (But for anyone who does care, it could be easier now)

FloEdelmann:
I combined all the Markdown release notes files into one root-level `CHANGELOG.md` file now, which the Gradle task then converts to HTML. Unfortunately, rendering that HTML in a TextView is really bad performance-wise, so I used a WebView instead. Styling that is a bit difficult, but I think I managed to match the previous look.

westnordost:
>  Unfortunately, rendering that HTML in a TextView is really bad performance-wise

Oh, that was unexpected. Hm, well, I guess what is most important is that it looks "native", i.e. the font and sizes are the same as in the rest of the app

buildSrc/src/main/java/UpdateChangelogTask.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.intellij.markdown.parser.MarkdownParser
+import java.io.File
+
+/** Convert a markdown changelog to HTML. */
+open class UpdateChangelogTask : DefaultTask() {
+
+    @get:Input lateinit var sourceFile: String
+    @get:Input lateinit var targetFile: String
+
+    private val markdownFlavour = CommonMarkFlavourDescriptor()
```
westnordost, line 16: Can we use Github flavour? I post the current changelog also for each result, so it would be nice if I could use all the markdown that is supported by github

buildSrc/src/main/java/UpdateChangelogTask.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val json = Json { prettyPrint = true }
```
westnordost, line 17: unused
```suggestion

```

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
```diff
@@ -52,20 +50,21 @@ class WhatsNewDialog(context: Context, sinceVersion: String) : AlertDialog(conte
...
-            var currentVersionIndex = fullChangelog.indexOfFirst { it.title == sinceVersion }
-            // if version not found, just show the last one
-            if (currentVersionIndex == -1) currentVersionIndex = 1
-            val changelog = fullChangelog.subList(0, currentVersionIndex)
-
-            binding.changelogList.adapter = ChangelogAdapter(changelog)
+            var sinceVersionIndex = fullChangelog.indexOf("<h2>$sinceVersion</h2>")
+            if (sinceVersionIndex == -1) {
+                // if version not found, just show the last one
+                sinceVersionIndex = fullChangelog.indexOf("<h2>", 4)
```
westnordost, line 63: This implementation seems a bit unstable too me. If I understand correctly, it intends to finds the second `<h2>` in the document and does this by leaving the first 4 characters out of the search. This breaks if the HTML changelog file does not immediately start with the header of the most current changelog. (Whitespaces?)

A more stable way would be to find the first occurrence, then search again from the first occurrence. Or, maybe more generally, maybe a helper extension function `fun String.indicesOf(): Sequence<Int>` could help to make it better readable?

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
```diff
@@ -75,33 +74,57 @@ class WhatsNewDialog(context: Context, sinceVersion: String) : AlertDialog(conte
...
+            h4 { font-size: ${h4Size}px; }
+
+            h2:not(:first-child) {
+                border-top: 1px solid $dividerColor;
+                padding-top: 1rem;
+            }
+
+            @media (prefers-color-scheme: dark) {}
+        </style>
+    """.trimIndent()
```
westnordost, line 127: Too bad that you have to do this extra work to make this look native. If you are already doing that, I think it may be better to actually create a kind of valid html and put the `html` content in `<body></body>` and the above in `<head></head>`?

Anyway, did you test:
- it works fine both on light and dark mode
- text scales with the system setting as other text in the system does, too
?

westnordost (commented):
I guess, `TextView` always renders as a whole, so if the text is very large, a lot of it is rendered off-screen and then discarded. `WebView` inevitably needs optimizations for that case, otherwise the web would be just incredibly slow.

Anyway. The resulting CHANGELOG.md looks very clean and will be smooth to extend. You committed some clean-ups of the changelog on master, so I trust that it is fine as is now.

Not too important, but I would like if the resulting changelog html had a comment at the beginning that advises that it should not be edited because it is created automatically from XYZ and if easily possible I would prefer if it was formatted in a human readable manner.

buildSrc/src/main/java/UpdateChangelogTask.kt
FloEdelmann: Yes, it's easy to switch to `GFMFlavourDescriptor`, but it only changes one thing in the output: `www.AddisMap.com` (from [v33.2](https://github.com/streetcomplete/StreetComplete/releases/tag/v33.2)) is rendered as a link instead of plaintext, because GFM linkifies URLs, while CommonMark does not.

buildSrc/src/main/java/UpdateChangelogTask.kt
FloEdelmann: Good catch! Removed in [`51d0b31`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/51d0b312d1b6e6b02a877b3dc8a81a33d7708f75).

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
FloEdelmann: Yeah, you're right. I like the `indicesOf` idea and implemented it in [`1811ddc`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/1811ddc410f494c205bb24bef812476cef1100a5).

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
FloEdelmann: It doesn't seem to make a difference, but creating valid HTML might be a good idea anyway. I changed that in [`41b89e0`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/41b89e09177669780f74a8c0ffff9dcabd099ec3).

I tested (on Android 13) that the WebView behaves correctly in both light and dark mode and with various text and display size settings, without restarting the app.

FloEdelmann:
> I would like if the resulting changelog html had a comment at the beginning that advises that it should not be edited because it is created automatically from XYZ

Done in [`2c35760`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/2c35760b560065b092a75e4831ba70441160f674).

> if easily possible I would prefer if it was formatted in a human readable manner

I would also have liked that, but unfortunately it's not easily possible in https://github.com/JetBrains/markdown. We could switch to a different Markdown to HTML converter though.

buildSrc/src/main/java/UpdateChangelogTask.kt
westnordost: That's fine, it's for the future, too

app/src/main/java/de/westnordost/streetcomplete/util/ktx/String.kt
```diff
@@ -4,3 +4,7 @@ fun String.truncate(length: Int): String =
     if (this.length > length) substring(0, length - 1) + "…" else this
 
 fun String.containsAll(words: List<String>) = words.all { this.contains(it) }
+
+/** Returns the indices of all occurrences of [substring]. */
+fun String.indicesOf(substring: String) =
+    generateSequence(indexOf(substring)) { indexOf(substring, it) }
```
westnordost, line 10: Uh, wait a moment... that works?

What about -1? If the substring cannot be found, the sequence should be empty, not consist of unlimited -1s.

And, shouldn't the offset be `it+1`?

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
```diff
@@ -52,20 +51,21 @@ class WhatsNewDialog(context: Context, sinceVersion: String) : AlertDialog(conte
...
-            var currentVersionIndex = fullChangelog.indexOfFirst { it.title == sinceVersion }
-            // if version not found, just show the last one
-            if (currentVersionIndex == -1) currentVersionIndex = 1
-            val changelog = fullChangelog.subList(0, currentVersionIndex)
-
-            binding.changelogList.adapter = ChangelogAdapter(changelog)
+            var sinceVersionIndex = fullChangelog.indexOf("<h2>$sinceVersion</h2>")
+            if (sinceVersionIndex == -1) {
+                // if version not found, just show the last one
+                sinceVersionIndex = fullChangelog.indicesOf("<h2>").elementAt(1)
```
westnordost, line 64: throws exception if there is no element at 1?

app/src/main/java/de/westnordost/streetcomplete/util/ktx/String.kt
westnordost: (using `sequence` with `yield` may result to be more correct and readable, maybe?)

westnordost:
Oh, and, what happens if you click on a link in the webview? It should always be opened in the system browser and never in the WebView itself.

westnordost:
Can you gauge when you will be able to finish this? I would like to do a release soon, but want to avoid creating additional work for you by modifying the current changelog.yml.

FloEdelmann:
Sorry, I didn't have much time the last days. I'll address your comments now.

> what happens if you click on a link in the webview? It should always be opened in the system browser and never in the WebView itself.

That's exactly what happens (at least on my phone with Android 13).

buildSrc/src/main/java/UpdateChangelogTask.kt
FloEdelmann: Okay, changed to GitHub Flavored Markdown in [`f20d12b`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/f20d12b5ce78213e9ec9e21844f88aafd726bcf6).

app/src/main/java/de/westnordost/streetcomplete/util/ktx/String.kt
FloEdelmann: You're right, I didn't test it properly. I rewrote the implementation with `yield` and added a unit test: [`99aba20`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/99aba20c10349a4f1e16bc5cf99aa2a7eeec32bf)

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
FloEdelmann: Now it does, see [`99aba20`](https://github.com/streetcomplete/StreetComplete/pull/5439/commits/99aba20c10349a4f1e16bc5cf99aa2a7eeec32bf).

westnordost: Actually, my point was that it should not throw an exception in case there are not two `<h2>` headlines in the file. But well, not so important because there are more than two headlines in the file.

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/screens/about/ChangelogFragment.kt
westnordost: And, I guess that was the behavior before, too, so my fault

PR #5440, opened 2024-01-11
Add hint label to note text

### LINKED ISSUE #5398: Add more context to notes
**Use case**
Some notes that users leave when unable to answer a quest are missing the context of the question, but users assume that this context is understood by other people. For example, when the quest asks whether a shops opening hours are still correct, users may assume that other people know which shop the question was asked for. Notes often look like this:

    Unable to answer "Are these opening hours still correct?" for https://osm.org/node/<nodeid> via StreetComplete 55.1:
    
    It is not there anymore.

This is problematic for objects that have changed since, for example if another user as already corrected the node without seeing the note. I know that I can always look at the object and its history and see which state it was in when the note was left, but this is a bit cumbersome.

**Proposed Solution**
If an object has a name, it would be helpful to include the name in the note automatically. There probably are other context clues one could include for different quests, but none come to mind right now.

    Unable to answer "Are these opening hours still correct?" for "Shop A" (https://osm.org/node/<nodeid>) via StreetComplete 55.1:
    
    It is not there anymore.

I think this could be helpful for all quests where the relevant object has a name. It would make the decision whether a note is still relevant a lot easier in those situations.

Helium314:
fixes #5398

This changes the note text as discussed in #5398.
I chose to use `<` and `>` for the link, as I suspect that links are parsed by different consumers (osm.org, notesReview, and probably more that I don't think of) and ideally none of them should produce broken links.

Though I'm not sure how nice this is in all cases, e.g. with level:
_Unable to answer "Are these opening hours still correct?" for on level 2: Mobile Service (Mobile Phone Shop) <https://...>_

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
@@ -245,7 +245,7 @@ abstract class AbstractOsmQuestForm<T> : AbstractQuestForm(), IsShowingQuestDeta
         val leaveNoteContext = if (hintLabel.isNullOrBlank()) {
             "Unable to answer \"$questTitle\""
         } else {
-            "Unable to answer \"$questTitle\" for $hintLabel"
+            "Unable to answer \"$questTitle – $hintLabel\""
         }
         listener?.onComposeNote(osmElementQuestType, element, geometry, leaveNoteContext)
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
@@ -245,7 +245,7 @@ abstract class AbstractOsmQuestForm<T> : AbstractQuestForm(), IsShowingQuestDeta
         val leaveNoteContext = if (hintLabel.isNullOrBlank()) {
             "Unable to answer \"$questTitle\""
         } else {
-            "Unable to answer \"$questTitle – $hintLabel\""
+            "Unable to answer \"$questTitle\" – $hintLabel"
         }
         listener?.onComposeNote(osmElementQuestType, element, geometry, leaveNoteContext)
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
@@ -241,7 +241,12 @@ abstract class AbstractOsmQuestForm<T> : AbstractQuestForm(), IsShowingQuestDeta
 
     protected fun composeNote() {
         val questTitle = englishResources.getQuestTitle(osmElementQuestType, element.tags)
-        val leaveNoteContext = "Unable to answer \"$questTitle\""
+        val hintLabel = getNameAndLocationLabel(element, englishResources, featureDictionary)
+        val leaveNoteContext = if (hintLabel.isNullOrBlank()) {
+            "Unable to answer \"$questTitle\""
+        } else {
+            "Unable to answer \"$questTitle\" for $hintLabel"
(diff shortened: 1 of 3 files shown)


westnordost:
Well, to be absolutely sure, I guess one could also use spaces. E.g. ( https://www.dot.at ). Looks fine, too. But, your decision. LGTM

westnordost (approved):

FloEdelmann:
> Though I'm not sure how nice this is in all cases, e.g. with level:
> _Unable to answer "Are these opening hours still correct?" for on level 2: Mobile Service (Mobile Phone Shop) [...](https://...)_

Another option would be `Unable to answer "$question $hint" for $element`, e.g. *Unable to answer "Are these opening hours still correct? on level 2: Mobile Service (Mobile Phone Shop)" for osm.org/…*

Helium314:
Thanks, I think that looks better and it avoids the parentheses.

Helium314:
Oh, that brings me to the next issue: It doesn't work for `In context of "$overlayTitle" overlay for $element`, e.g. _In context of "Street Parking Service Road" overlay for http..._ 
Maybe
* `In context of overlay "$overlayTitle $hintTitle" for $element`
_In context of overlay "Street Parking Service Road" for http..._ 
* `In context of overlay "$overlayTitle ($hintTitle)" for $element`
_In context of overlay "Street Parking (Service Road)" for http..._ 
* `In context of "$overlayTitle" overlay ($hintTitle) for $element`
_In context of "Street Parking" overlay (Service Road) for http..._ 

I'd go for the second one if there is no objection or better idea.

FloEdelmann:
I guess there is no easy way to append the hint text after the element URL? That would be consistent and clean:

* `Unable to answer "$question" for $element ($hintTitle)`
  *Unable to answer "Are these opening hours still correct?" for http… (on level 2: Mobile Service (Mobile Phone Shop))*
* `In context of "$overlayTitle" overlay for $element ($hintTitle)`
  *In context of "Street Parking" overlay for http… (Service Road)*

westnordost:
Maybe without the "for" but instead with a "-"?

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
```diff
@@ -241,7 +241,12 @@ abstract class AbstractOsmQuestForm<T> : AbstractQuestForm(), IsShowingQuestDeta
 
     protected fun composeNote() {
         val questTitle = englishResources.getQuestTitle(osmElementQuestType, element.tags)
-        val leaveNoteContext = "Unable to answer \"$questTitle\""
+        val hintLabel = getNameAndLocationLabel(element, englishResources, featureDictionary)
+        val leaveNoteContext = if (hintLabel.isNullOrBlank()) {
+            "Unable to answer \"$questTitle\""
+        } else {
+            "Unable to answer \"$questTitle – $hintLabel\""
```
westnordost, line 248: Wrong placement of `"`?

Should it really be

> Unable to answer "Are these opening hours still correct? - on level 2: Mobile Service (Mobile Phone Shop)"

and not rather

> Unable to answer "Are these opening hours still correct?" - on level 2: Mobile Service (Mobile Phone Shop)

?

Helium314: I don't really have a preference here, if you like like the second variant more I'll change it.

westnordost: Yeah, I like the second option more, or actually, the full string would be 

> Unable to answer "Are these opening hours still correct?" - on level 2: Mobile Service (Mobile Phone Shop) for https://osm.org/node/....

wouldn't it? Actually, my suggestion was to not use "for" at all, but use "-", i.e.

> Unable to answer "Are these opening hours still correct?" - on level 2: Mobile Service (Mobile Phone Shop) - https://osm.org/node/....

westnordost (approved):

PR #5442, opened 2024-01-12
Multiplatform JSON parsing (kotlinx.serialization)

starsep:
At first I was using `allowTrailingComma` which was introduced in kotlinx.serialization 1.6.1.
It required upgrade of kotlinx.serialization, Kotlin, and using `ExperimentalSerializationApi` annotation.

I checked that `TrafficFlowSegmentsApi` returns proper JSON without trailing commas.
Example url: https://www.westnordost.de/streetcomplete/oneway-data-api/?bbox=18.0000000,-34.0000000,19.0000000,-33.0000000
I decided to modify the tests instead.

I added `val wayId: Long` to `TrafficFlowSegment`. 
Alternatively I can create `data class` only for serialization and create `TrafficFlowSegment` out of it without `wayId`.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/StreetCompleteImageUploader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/StreetCompleteImageUploader.kt
@@ -14,7 +14,10 @@ import java.net.URLConnection
 
 
 @Serializable
-private data class PhotoUploadResponse(@SerialName("future_url") val futureUrl: String)
+private data class PhotoUploadResponse(
+    @SerialName("future_url")
+    val futureUrl: String
+)
 
 /** Upload and activate a list of image paths to an instance of the
  * <a href="https://github.com/exploide/sc-photo-service">StreetComplete image hosting service</a>
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/StreetCompleteImageUploader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/StreetCompleteImageUploader.kt
@@ -2,18 +2,27 @@ package de.westnordost.streetcomplete.data.osmnotes
 
 import de.westnordost.streetcomplete.ApplicationConstants
 import de.westnordost.streetcomplete.data.download.ConnectionException
-import org.json.JSONException
-import org.json.JSONObject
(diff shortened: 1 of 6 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/StreetCompleteImageUploader.kt
```diff
@@ -2,18 +2,27 @@ package de.westnordost.streetcomplete.data.osmnotes
...
+import kotlinx.serialization.json.Json
 import java.io.File
 import java.io.IOException
 import java.net.HttpURLConnection
 import java.net.URL
 import java.net.URLConnection
 
+
+@Serializable
+private data class PhotoUploadResponse(@SerialName("future_url") val futureUrl: String)
```
FloEdelmann, line 17:
```suggestion
private data class PhotoUploadResponse(
    @SerialName("future_url")
    val futureUrl: String
)
```

app/src/main/java/de/westnordost/streetcomplete/quests/oneway_suspects/data/TrafficFlowSegment.kt
```diff
@@ -1,5 +1,7 @@
 package de.westnordost.streetcomplete.quests.oneway_suspects.data
 
 import de.westnordost.streetcomplete.data.osm.mapdata.LatLon
+import kotlinx.serialization.Serializable
 
-data class TrafficFlowSegment(val fromPosition: LatLon, val toPosition: LatLon)
+@Serializable
+data class TrafficFlowSegment(val wayId: Long, val fromPosition: LatLon, val toPosition: LatLon)
```
FloEdelmann, line 7:
```suggestion
data class TrafficFlowSegment(
    val wayId: Long,
    val fromPosition: LatLon,
    val toPosition: LatLon,
)
```

FloEdelmann (commented):
Nice to see that the code is not only more expressive than before, but also shorter :slightly_smiling_face: 

I have two little readability improvement suggestions, otherwise this looks good to me.

westnordost (approved):
LGTM, thank you!

I didn't have a deep look at the modified logic in StatisticsParser, but thankfully we have a test, so If the test runs through, I think it must be fine!

westnordost:

PR #5448, opened 2024-01-18
Optimize parsing of element filter

westnordost:
I refreshed the implementation for the element filters parser to make it more performant and easier readable (IMO).

According to my measurements, parsing should be about 3x faster now. It was slow before mainly because finding the word length (tag key, tag value length) involved iterating each to the end of the string. Also, regexes are slow, relatively speaking.

Sounds a lot, but it actually really isn't that much of an improvement. On my phone, parsing of all quest filters took ~350ms before, now it takes ~100ms - on a single thread. With element filters being parsed in parallel, we'll end up with something like 50ms faster first quests download.
(Quest filter parsing does not affect startup time because quest filters are generally parsed lazily, i.e. only the first time they are used)

I think it is still useful to merge this because I overall improved the readability of the element filters parser.


westnordost:
Oh damn, I committed these changes to the wrong branch... well, I'll just review this myself and merge it rather than picking these out again...

FloEdelmann:
I just had a quick look over this PR and it looks good to me, so merging was fine :slightly_smiling_face:

Issue #4965, opened 2023-04-22
Marker outside oddly shaped target area


tobiasgradl:
![image](https://user-images.githubusercontent.com/52239795/233801665-fd77221e-4284-4484-a1c0-ae739a93dae0.png)

For the "J"-shaped [forest area](https://osm.org/way/375479412) in the above screenshot, Street Complete asks for forest's the leaf type. The marker for the questions is located on a meadow outside the area (probably at its geometrical center).

**How to Reproduce**
Go to [N 51,0356643, E 7,3105947](https://www.openstreetmap.org/#map=18/51.0356643/7.3105947) in the StreetComplete app

**Expected Behavior**
The question's map marker is located inside the corresponding area.

**Versions affected**
Android, v52.0.

rhhsm:
I've seen this a few times, afaik always for the leaf type quest, but it didn't bother me because it was clear enough (so far) which oddly shaped forest it was asked about.

westnordost:
The code to calculate the center of is here:

https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/util/math/SphericalEarthMath.kt#L362-L386

tobiasgradl:
It's a minor issue. I spotted it for the first time. It got aggravated in this rare case by the area being very narrow, i.e., hardly visible on the map. In most other cases the user can probably guess which area the question belongs to.

Thanks for providing a pointer to the current algorithm! Maybe I'll find time to create a PR.

westnordost:
I'll release a new minor update in a week or so, so now is a good timeframe to implement this when you want to see this change live quickly. If you are not interested anymore in implementing this, I will mark this as "help wanted".

tobiasgradl:
Unfortunately, I'm not able to spend time on this issue, at the moment. Feel free to mark it as "help wanted"!

westnordost:
Correction: "[centroid](https://en.wikipedia.org/wiki/Centroid#Properties)" is by definition not necessarily within the boundary of the polygon.

MapBox implemented an algorithm finding the polygon pole of inaccessibility, the most distant internal point from the polygon outline in JavaScript: https://github.com/mapbox/polylabel

There is not one port of it to Java, ... but three??
- https://mvnrepository.com/artifact/org.geotools/gt-polygon-labeller (source code unknown, probably somewhere in the geotools repo)
[...]

PR #5461, opened 2024-01-26
Buildings overlay

### LINKED ISSUE #4741: Add 'previous choice' option to Street lit overlay
**Use case**
Currently if you're doing a bunch of the "is this street lit?" Entries; it is requiring more clicks than I feel like it should be doing

**Proposed Solution**
I think similar to the quest where you have to select every Street has a cycle way; there should be a button when you're trying to answer the streetlights quest in the bottom left corner with the last answer you gave. That would half the amount of taps

westnordost:
https://github.com/streetcomplete/StreetComplete/assets/4661658/95a9fd9f-f70f-42c4-970a-e811c1d926f4

### Building type overlay features:

- buildings are roughly categorized according to function and get their highlight color from there. Additionally, when zooming in, the icon of each building type is displayed.

- some not supported building types are treated as "synonyms" to supported building types. E.g. `building=livestock` is treated as synonym to `building=farm_auxiliary`. These buildings will be displayed as that building type in the app and if the in-app building type is not changed to another building, the "synonym" is retained on saving changes.

- other not supported building types are highlighted in grey, together with "other" buildings, clicking on them and changing them is still possible

- deprecated building types are highlighted in red, i.e. as if the building type hasn't been specified yet. However, if the building has a tag that potentially already describes the building type (e.g. `amenity`, `shop`, `emergency`, etc...), it is not highlighted in red

- the last selected buildings are displayed as buttons below the building selector

- added more selectable buildings: tower, container, tent, outbuilding, tomb also for the quest

- changed the categorization of the buildings: There is now a new category named "outbuilding" (which is also selectable) that contains all those smaller secondary buildings, like garages, sheds, garden allotment buildings, service buildings etc.

### Additionally

https://github.com/streetcomplete/StreetComplete/assets/4661658/9ceac671-8d51-4abe-81c2-7ff291f262c6

- added last-picked buttons to surface overlay (only for main surface), lit overlay, cycleway overlay (for separate cycleways) (fixes #4741 and more)


westnordost:
Oh, and of course, I disabled the building type quest now by default.

To be honest, this was the main reason why I even started the implementation for this overlay. The quest is just so spammy.

Helium314:
Some minor issues I found in a short test:
* The icon can be outside the building outline. This is the same as for quests, but for the overlay it "feels different". But I remember reading that correct placement inside the building can be tricky in some cases, so probably too much work.
* The icons for recently selected values are somewhat too wide for my display, so the rightmost icon has ~10-20% cut off. The whole line is scrollable, but it still seems a little weird.
* Some icons have weird lines at the bottom (they change color when zooming):
lines

westnordost:
Thanks for trying it out!

1. Yes, that's #4965 . It would be nice, but I do not see it as required for this PR to be merged, although the implementation would roughly be just 150 lines of code (+tests)

2. Hmm... it is difficult to find a right max width that will not look much too little on some devices. Can you add a screenshot, both in portrait mode and in landscape mode? For the building levels quest, you do not have this problem, as there are those buttons, too?

3. Yeah, I'm pretty sure that's tangram-specific. I don't really want to concern myself with tangram issues.

Helium314:
> For the building levels quest, you do not have this problem? After all, there are buttons like these, too?

The building levels quest only has 5 buttons, so there it's not an issue.
Maybe the number of recent values could be reduced if not all fit on screen?

Screenshot_20240131-204904_Street�Complete_Dev
Screenshot_20240131-204913_Street�Complete_Dev

westnordost:
In general it would of course be best if one could actually scroll the buttons, and that **is** possible, but the default UI for that is not good. I experimented a bit with overlaying a gradient at the beginning and end (like done in the tutorial) but gave up because it turned out to be too much work. And in light of me wanting to migrate to Jetpack Compose as UI framework soon-ish, I don't want to put too much complexity in (new) layouts.

Anyway, maybe it would suffice to just make the buttons just a little bit narrower, I will have a look.

westnordost:
2dp smaller should do the trick, right?

mnalis:
> Oh, and of course, I disabled the building type quest now by default.

> To be honest, this was the main reason why I even started the implementation for this overlay. The quest is just so spammy.

Uhh, while I agree that building quest was spammy and of relatively low value (see https://github.com/streetcomplete/StreetComplete/issues/2464), doesn't that (disabling building quest by default) mean that solving (among most important of the quests, IMHO) **Address quest is now significantly less likely to be triggered**?

I'm afraid that it would make address quest much less discoverable and usable.

E.g. when there are lots of buildings tagged just with `building=yes` (as is most common, and happens in [~80% of all cases](https://taginfo.openstreetmap.org/keys/building#values)), am I correct in assuming that now Address quests won't ever be asked by default, unless user enables Building overlay and starts tagging buildings with more specific values? (or of course enables Address overlay and tags housenumbers there directly)?


---

Perhaps now would be a good time to revisit https://github.com/streetcomplete/StreetComplete/issues/2464#issuecomment-782962059 (which was postponed at the time)  and extended housenumber quest to asks also for generic `building=yes`?

While I have used Address Overlay in the past, it was sporadic (and mostly to correct off-by-one numbering). In all other cases, I've found Address Quest to be significantly more useful and easier to use. It would be real shame to basically lose it in 80% of the cases...

matkoniecz:
And power users can reenable this quest type and/or use overlay - so they are not impacted negatively.

mnalis:
Thanks for the detailed answer @westnordost , I really appreciate it! 

I definitely agree that nice and smooth UX is a must for targeting newbies, and that default config should not be targeted towards advanced users (so I am OK with your decision here - I was just unsure if the collateral damage was overlooked or intended).

I've opened https://github.com/streetcomplete/StreetComplete/discussions/5472 as discussion to identify issues and suggest possible fixes to make SC more fun and less work for users (in order not to pollute this PR)

PR #5466, opened 2024-01-31
Add some achievement links

### LINKED ISSUE #5462: Open vege map app is discontinued
**How to Reproduce**

Complete kind to animals and there is a link to open vege map. Which is now discontinued https://wiki.openstreetmap.org/wiki/OpenVegeMap

**Expected Behavior**

Just have the badge display upon completion

**Versions affected**

56.0

westnordost:
fixes #5462


app/src/main/res/values/strings.xml
```diff
@@ -429,16 +429,19 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
     <string name="link_vespucci_description">Advanced OSM editor for Android</string>
     <string name="link_ideditor_description">Simple OSM editor for the browser. Optimized for the desktop or large tablets</string>
+    <string name="link_every_door_description">Powerful POI editor for Android and iOS</string>
+    <string name="link_mapcomplete_description">Collection of thematic OSM maps that each show and let you edit one kind of feature</string>
     <string name="link_umap_description">Create maps with custom data quickly and embed them in your site</string>
     <string name="link_opnvkarte_description">A map of public transport routes and stops</string>
     <string name="link_pic4review_description">Contribute to OSM simply by looking at pictures of your city</string>
     <string name="link_wiki_description">The wiki is your starting point to everything related to OpenStreetMap</string>
+    <string name="link_forum_description">Need help? Want to read what’s going on? Come and mingle, it’s bustling with activity from communities all over the globe!</string>
```
matkoniecz, line 443: Would it make sense to mention that it is (1) an official forum (2) you can login with OSM account?

westnordost: Well, it's called "OpenStreetMap Community Forum", so the expectation for both is "yes". The wiki link also does not have any description that it is official.

app/src/main/java/de/westnordost/streetcomplete/data/user/achievements/AchievementsModule.kt
```diff
@@ -239,14 +272,6 @@ private val links = listOf(
         R.drawable.ic_link_wheelmap,
         R.string.link_wheelmap_description
     ),
-    Link(
-        "openvegemap",
-        "https://openvegemap.netlib.re",
```
peternewman, line 244: Why not just replace it with a link to https://www.veggiekarte.de/ as mentioned here:
https://github.com/Rudloff/openvegemap/issues/134#issuecomment-1516185005

Seems to do the same thing in a similar way.

westnordost: Hmm, good idea actually. Didn't know this existed and must have overread it in case it was mentioned.

peternewman: I mean it still seems a shame to me that someone hasn't just made a generic open diet map which you could toggle between say vegetarian/vegan, halal, gluten free etc, but there we go.

westnordost: done.

Well, I think it makes most sense to do that in mapcomplete?

PR #5468, opened 2024-01-31
Use AppCompat ImageView instead ImageView

Jean-BaptisteC:
This PR replaces android:src by app:srcCompat
That improves the quality of drawables on older devices because AppCompat library does a better scaling that Android library -> more information here https://stackoverflow.com/questions/34936590/why-isnt-my-vector-drawable-scaling-as-expected
No impact in Kotlin code -> It's the same code to set icons with Android library and AppCompat library.

### DIFF
--- a/app/src/main/res/layout-land/fragment_achievement_info.xml
+++ b/app/src/main/res/layout-land/fragment_achievement_info.xml
@@ -20,7 +20,7 @@
 
     <ImageView
         android:id="@+id/shineView1"
-        android:src="@drawable/shine"
+        app:srcCompat="@drawable/shine"
         android:layout_width="800dp"
         android:layout_height="800dp"
         app:layout_constraintTop_toTopOf="@id/dialogContainer"
@@ -31,7 +31,7 @@
 
     <ImageView
         android:id="@+id/shineView2"
-        android:src="@drawable/shine"
+        app:srcCompat="@drawable/shine"
         android:layout_width="800dp"
         android:layout_height="800dp"
         android:rotation="180"
--- a/app/src/main/res/layout/cell_image_thumbnail.xml
+++ b/app/src/main/res/layout/cell_image_thumbnail.xml
@@ -19,7 +19,7 @@
         android:layout_width="32dp"
         android:layout_height="32dp"
         android:padding="0dp"
-        android:src="@drawable/ic_close_24dp"
+        app:srcCompat="@drawable/ic_close_24dp"
         android:scaleType="center"
         app:tint="#fff"
         android:background="#9333"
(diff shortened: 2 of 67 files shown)


westnordost:
> more information here https://stackoverflow.com/questions/34936590/why-isnt-my-vector-drawable-scaling-as-expected

Mh, at least the first comment(s) don't mention `srcCompat` and those that do mention it only as part of some more involved workaround. Plus the whole discussion is 8 years old. Impossible to say if nothing has changed.

Do you have a device that runs on Android Lollipop and have actually been experiencing issues?

Jean-BaptisteC:
My explanation is not fully correct, we already use AppCompatImageView, because ImageView is auto inflated in AppCompatImageView.
And we use already app:tint property, a property from AppCompat -> https://developer.android.com/reference/kotlin/androidx/appcompat/widget/AppCompatImageView
Why do not use new properties from AppCompat for better drawables support? And specially in this case look of drawables is better on older devices with low dpi -> I have doing some tests with another app on old tablets, we see a real difference, drawables is more clean.

westnordost:
I just want to make sure we are not trying to solve a problem that does not exist. I don't know what other changes using `srcCompat` may bring. Will it only replace behavior on Android versions that are known to have issues, or will it also replace behavior on versions that don't (, adding overhead)? Are there actually any old Android versions still supported by the app that do have issues with vector drawable scaling?

I know of one appcompat thing the Android linter also keeps nagging about that is actually not even relevant for the app anymore as its min SDK version is above the version that used to cause issues. I had to turn off that linter warning in the build.gradle.kts to have my peace.

Jean-BaptisteC:
I understood your position, srcCompat replace behavior on all Android versions.
Scaling problem appears on devices with Android 5 and low dpi. Here you can find example in other app -> https://github.com/organicmaps/organicmaps/issues/3556 

I have planned to doing somes tests on older device and add table before/after

westnordost:
> I have planned to doing somes tests on older device and add table before/after

If I knew this PR solves an actual problem, that would make me feel somewhat more comfortable. Alternatively, knowing what exactly srcCompat does (on all Android versions) would also help.

Jean-BaptisteC:
I have done some tests on SC (On Lenovo Yoga Tab 2 - Android 5) and I agree, the difference is very little
Here is an example with screenshots tutorials
|Before|After|
|-|-|
|||

I have found more information about vectors support with AppCompat -> https://medium.com/androiddevelopers/appcompat-v23-2-age-of-the-vectors-91cbafa87c88#.pf6wmj3oz
I'm not an expert on the topic, but from my experience, I use this property on exodus app project from app has been reworked in Kotlin with any problems.

westnordost:
Alright, let's merge it then. (I've also read the article, it mentions that the compat drawing is only used on versions below API 21, which can't be entirely true because your tablet is API 21)

Thank you for bearing with me!

PR #5471, opened 2024-02-02
Redo crossing quests

### LINKED ISSUE #4886: Prefer tagging crossing:markings=* and crossing:signals=* over crossing=*
Here's an example change: https://www.openstreetmap.org/changeset/111677927 
You can see that currently it just adds `crossing:unmarked` but could also add `crossing:markings=no`

**Edit by @westnordost:** The accepted solution is described here https://github.com/streetcomplete/StreetComplete/issues/4886#issuecomment-1485667735

westnordost:
Fixes #4886

There are now three quests, asked in this order:

1. Does this crossing have traffic signals? - yes/no
2. Does this crossing have a traffic island? - yes/no
3. Does this crossing have markings on the ground? - yes/no

They each tag `crossing:signals`, `crossing:island` and `crossing:markings`.

To avoid quest spam, the three quests are not asked for some situations:

- **Traffic signals** is not asked if any `crossing`-value other than `island` is set because it is assumed that if the crossing had traffic signals, it would have been tagged with `crossing=traffic_signals` rather than some marked or unmarked value.

- **Markings**  is not asked if any `crossing`-value other than `island` is set because either it has already been tagged as marked/zebra/uncontrolled/unmarked, or as having traffic signals. If it has traffic signals (either via `crossing=traffic_signals` or `crossing:signals`), it is not asked because it is considered spammy - the answer would usually be "yes".

- Conversely, **Traffic Island** is of not asked if `crossing=island` is set, but otherwise always

There are some more restrictions for all three, but these restrictions are not new.

---

The PR further adds support for `crossing:signals` in other quests, e.g. the quest about whether the traffic lights have a vibrating button are also asked if `crossing:signals=yes` is set, not only when `crossing=traffic_signals` is set.

---

While both `crossing:island` and `crossing:markings` are approved tags, `crossing:signals` is currently only in-use but not approved. It however has substantial usage (40k+), substantial enough that it would make sense for data consumers to also check that tag and not only `crossing=traffic_signals`.
Since the latter is unambiguous and the `crossing` value will not be touched by this app, I think adding `crossing:signals` is fine. People who insist on tagging `crossing=traffic_signals` can just do so based on the information added by StreetComplete users as they see fit.


app/src/main/java/de/westnordost/streetcomplete/quests/crossing_signals/AddCrossingSignals.kt
```diff
@@ -0,0 +1,62 @@
...
+          and !crossing:signals
+          and (!crossing or crossing = island)
+    """.toElementFilterExpression() }
+    /* only looking for crossings that have no crossing=* at all set because if the crossing had
+     * signals, it would have been tagged with crossing=traffic_signals */
+
+    private val excludedWaysFilter by lazy { """
+        ways with
+          highway and access ~ private|no
+          or highway = service
```
1ec5, line 32: FYI, this would exclude things like entrances to shopping malls and some residential subdivisions at signalized intersections, for example:

https://www.openstreetmap.org/node/4146380881
https://www.openstreetmap.org/node/5860157357
https://www.openstreetmap.org/node/11419081726

However, I can see how such crossings would get very spammy because of the legacy of openstreetmap/iD#8463. Do you have a way of checking whether the crossing is one node away from a `highway=traffic_signals` node? I think that would be a much better indicator of whether signalization is likely or not.

res/graphics/quest/traffic_lights_button.svg
1ec5, line 1: As an American, my first impression of this icon is that it depicts a vehicular traffic signal (since our pedestrian signals are orange and white, not red and green) and a raised hand telling the pedestrian to stop. I think most Americans would think this icon is about whether the crossing has a crossing guard ([`supervised`](https://wiki.openstreetmap.org/wiki/Key:supervised)).

A more intuitive icon for a pedestrian call button quest would be something that depicts a finger pressing a button. The button’s shape and surrounding panel looks very different from region to region, but if you show it in profile, then the raised button will be more broadly recognizable.

res/graphics/quest/blind_traffic_lights.svg
1ec5, line 1: I think this change will be OK, but it’s very close to looking like the green light is crossed out, rather than the stick figure holding a cane.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_markings/AddCrossingMarkings.kt
```diff
@@ -11,40 +11,38 @@ import de.westnordost.streetcomplete.osm.Tags
...
+          and !crossing:markings
+          and (!crossing or crossing = island)
+          and crossing:signals != yes
     """.toElementFilterExpression() }
+    /* only looking for crossings that have no crossing=* at all set because if the crossing was
+     * - if it had markings, it would be tagged with "marked","zebra" or "uncontrolled"
+     * - if it hadn't, it would be tagged with "unmarked"
+     * - and in case of "traffic_signals", we currently assume that when there are traffic signals
+     *   it would be spammy to ask about markings because the answer would almost always be "yes".
+     *   Might differ per country, research necessary. */
```
1ec5, line 32: I realize you aren’t keen on _applying_ `crossing:signals=separate` or `shared`, especially since [this proposal](https://wiki.openstreetmap.org/wiki/Proposal:Crossing_signalization) is still a draft. However, if your intention is to exclude signalized crossings, then the filter would be more effective if it also excludes the [1,412 nodes](https://taginfo.openstreetmap.org/keys/crossing%3Asignals?filter=nodes#values) with these values.

I agree that the “spamminess” differs markedly by country. Hopefully there will be some flexibility in the future to ask about markings for some obvious regions like the U.S. (outside of the largest cities perhaps). But even if not, markings are much easier to armchair-map than signalization.

app/src/main/res/values/strings.xml
```diff
@@ -1288,7 +1283,11 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_parkingType_underground">Underground garage</string>
     <string name="quest_parkingType_multiStorage">Parking garage</string>
 
-    <string name="quest_pedestrian_crossing_island">"Does this pedestrian crossing have an island?"</string>
+    <string name="quest_pedestrian_crossing_signals">"Does this crossing have traffic signals?"</string>
```
1ec5, line 1286: This would be a good place to clarify that either vehicular or pedestrian signals would count as a yes. Otherwise, without any additional qualification, users will respond inconsistently: some users will think “traffic signals” only means vehicular signals and say no even if there are pedestrian signals, while others will think a crossing controlled only by vehicular signals “has” no signals at all, because no signals are dedicated only to the crossing. Meanwhile, some users will mistakenly think warning lights count as traffic signals, even though we tag them as [`flashing_lights=yes`](https://wiki.openstreetmap.org/wiki/Key:flashing_lights).

app/src/main/res/values/strings.xml
```diff
@@ -1288,7 +1283,11 @@ If there are no signs along the whole street which apply for the highlighted sec
...
+
+    <string name="quest_pedestrian_crossing_island">"Does this crossing have an island?"</string>
```
1ec5, line 1288: In case someone doesn’t understand the meaning of “island” in this context, how about something more descriptive like:

> Can a pedestrian cross halfway and wait at an island for the next walk signal?

1ec5 (commented):
Thank you for this vote of confidence in the `crossing:signals=*` key!

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_signals/AddCrossingSignals.kt
westnordost: You mention the legacy of that iD issue, but this hasn't been the reason for excluding `highway=service`, but rather the deliberation that

1. many service roads that are actually driveways are not (yet) tagged with `service=driveway`
2. it is generally really unlikely that a crossing with a service road is signalized. 

My gut estimation is that it would be too spammy. That's because the crossing island one is also not asked for any `highway=service`. Yeah, that's all gut estimation, but to topple that estimation, an argument backed with some numbers would be necessary (and I would say it is hard to get this information).

Point 1 is a big a problem. It would be much less spammy if this could be classified as solved, for example by a quest that asks for the type of service road, but IIRC that attempt failed because we determined that there are cases where the type of service road cannot be defined any more detailed than `highway=service` (i.e. no sub-type).

> Do you have a way of checking whether the crossing is one node away from a highway=traffic_signals node?

It is somehow possible, but it would be very complex because in the data structure, the vertices of ways do not link back to the way(s) they are contained within. Too complex to find out this detail, it also covers only a very small number of cases: Those cases where someone mapped the vehicular traffic lights but did forget to add the crossing type.

res/graphics/quest/traffic_lights_button.svg
westnordost: Hm yeah, the icons are in other places also quite European / Vienna-Convention style.

Since it is only an icon, and not the question itself, I think it is fine.

I did try to create an icon with a finger, but I failed to make it look recognizable. A single finger is not recognizable as one  (just a white line), a white fist with a finger 👆 is difficult to recognize when the icon is shown small, it looks like a white blob with a long thing sticking out at the top. But the icon is shown relatively small.

So anyway, I tried, and I failed, so I let it be.

res/graphics/quest/blind_traffic_lights.svg
westnordost: You are right. Let's see what I can do about that.

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_markings/AddCrossingMarkings.kt
westnordost: Right, changed it to `and (!crossing:signals or crossing:signals = no)`

app/src/main/res/values/strings.xml
westnordost: Sounds good, but also so complicated that I think it might be bordering on actually being confusing (which happens if you circumscribe a common thing instead of naming it). I've asked in the US #language chat and it seems to me that at least when the context is given (a crossing), "Island" is generally understood)

app/src/main/res/values/strings.xml
westnordost: Heh. Well, but is there a possible wording that is not massively complicated (and thus confusing enough that it also can lead to misunderstandings and hence wrong taggings)?

westnordost:
Thanks for the review, @1ec5 !

app/src/main/java/de/westnordost/streetcomplete/quests/crossing_signals/AddCrossingSignals.kt
1ec5: > Those cases where someone mapped the vehicular traffic lights but did forget to add the crossing type.

That’s fair, most `crossing:signals` ([69.67% of them](https://taginfo.openstreetmap.org/keys/crossing:signals#combinations)) are already being combined with `crossing:markings`, although the proportion will decrease over time thanks to this quest.

res/graphics/quest/traffic_lights_button.svg
1ec5: > A single finger is not recognizable as one (just a white line), a white fist with a finger 👆 is difficult to recognize when the icon is shown small, it looks like a white blob with a long thing sticking out at the top.

Interesting, I wouldn’t have expected that, given the prevalence of pointing hands as mouse cursors (especially as hyperlink cursors). But maybe that’s a peculiarity of the artistic style used here.

app/src/main/res/values/strings.xml
1ec5: For someone in Spain or Australia, the quest’s icon already makes the intention very clear, because the signal resembles a pedestrian signal and vehicular signals are presumably irrelevant. I was kind of hoping the icon could very by country so that more countries like Argentina, Canada, or the U.S. could benefit from the same cue. But since this isn’t feasible, would it be possible to display a small icon inside or alongside the question, perhaps varying by region?

> “Are there traffic signals like [two-phase signal](https://commons.wikimedia.org/wiki/File:Japanese_pedestrian_traffic_signal.svg) or [vehicular signal](https://commons.wikimedia.org/wiki/File:Traffic_lights_dark_green.svg) at this crossing?

> “Are there traffic signals like [walk signal](https://commons.wikimedia.org/wiki/File:MUTCD_Ped_Signal_-_Walk.svg) or [vehicular signal](https://commons.wikimedia.org/wiki/File:Traffic_lights_dark_green.svg) at this crossing?”

If that’s too complicated, I would pose a much simpler question:

> “Do you have to wait for a traffic signal to cross here?”

I realize you [aren’t keen on requiring any knowledge of traffic laws](https://wiki.openstreetmap.org/wiki/Proposal_talk:Crossing_signalization#Verifiability_problem_with_shared/separate) to solve quests. However, if the user, a pedestrian, doesn’t know what a crossing or intersection signal looks like, then this quest is probably putting them in immediate danger. 😬 More likely, the user knows what the various signals look like, but they don’t know that a flashing warning light doesn’t count as a “yes” for the purpose of this quest, because it’s worded so vaguely.

westnordost: > “Do you have to wait for a traffic signal to cross here?”

Oh, that's good, I will use that, then.

vmartyanov:
I have doubts about the question: "Do you have to wait for a traffic signal to cross here?" which sounds weird. What is wait? 5 seconds? Or ten? Is it a question about traffic lights timing? I suppose no :-) I really don't know what was asked and how to answer...

If the question is about existence of traffic lights - it should be "Does this crossing have traffic signals?" like other questions about crossings: "have island", "have markings" etc.

app/src/main/res/values/strings.xml
vmartyanov: New wording is confusing to me: what is wait? Wait for what? etc.

matkoniecz:
> If the question is about existence of traffic lights - it should be "Does this crossing have traffic signals?" like other questions about crossings: "have island", "have markings" etc.

Note that some crossing have warning traffic lights (blinking light) that do not regulate when you can cross it.

(though now I remember than in some places red light does not mean you need to wait either, just that you have no right of way)

westnordost:
Hm, why did I change the naming to that again... it was something related to the US. @1ec5 , do you know?

vmartyanov:
> Note that some crossing have warning traffic lights (blinking light) that do not regulate when you can cross it.

Yep, maybe it's a subject for further tags development? crossing:signals=yes + crossing_signals=...

1ec5:
`crossing:signals` is about traffic signals that _require_ the pedestrian or cyclist to wait for other traffic before crossing. It is not for the mere presence of any kind of signal. There are already other ways to map that, such as `flashing_lights=yes` `button_operated=yes` for warning lights that the pedestrian can activate for road users to see. But it does need to account for vehicular signals that also apply to the pedestrian going in the same direction. (I have [a draft proposal](https://wiki.openstreetmap.org/wiki/Proposal:Crossing_signalization) that would make this distinction more explicit.)

In suggesting the current wording, I was assuming a typical user would not think they would answer the question differently if they are unlucky enough to arrive at the intersection a few seconds too late, given that so many of StreetComplete‘s quests focus on recording physical infrastructure. However, if we’re concerned about a user mistaking StreetComplete for a sort of [_dérive_](https://en.wikipedia.org/wiki/Dérive) facilitator, logging your personal interactions with the environment as you move through it, then this clunkier alternative would leave less room for interpretation:

> Are there signals of any kind that would require you to wait before crossing the street at this crossing?

PR #5473, opened 2024-02-03
Replay edits done during quests download

### LINKED ISSUE #4258: Quest is asked again immediately after solving it
This is again reproduced https://github.com/streetcomplete/StreetComplete/issues/4124 (and before that, https://github.com/streetcomplete/StreetComplete/issues/4086). See them for more details and history of the problem.

I've managed to do both things requested in https://github.com/streetcomplete/StreetComplete/issues/4124#issuecomment-1173187068:
- try with 45.x version
- verify history that edits were made 

**How to Reproduce**
- answered the cycleway quest with `no` at `2022-07-29` about `14:04 CEST` (or minute before).
- noticed at `14:04 CEST` that I'm asked if the situation is still the same
- started screen recording
- verified in the `undo` menu that there is recorded answer `Added cycleway:right=no`
- answered `no` to the `is this still the cycleway situation here?`
- answered `displayed separately on the map` to the `is this still the cycleway situation here?`
  (note that this answer does not match reality, but I choose different answer than `no` to be sure to get multiple changes in changeset, so I can prove the bug is not only in my mind :smile:)
- verified in the `undo` menu that the new answer generated `Updated to cycleway:right=separate`
- verified in the `undo` menu that there is also still recorded old answer `Added cycleway:right=no`
- stopped screen recording
- started uploading all quests immediately after the video ends (i.e. `14:05/14:06 CEST`)

Video:

https://user-images.githubusercontent.com/156656/182037500-aeb8c647-8e17-46d9-8b61-c936509c68f6.mp4

I expect it is this changeset: https://www.openstreetmap.org/changeset/124228420#map=18/46.41630/16.69096
and this way specifically:
https://www.openstreetmap.org/way/670948488/history (`v8` and `v9`)

**Expected Behavior**
 - `is this still the cycleway situation here?` quest should not have triggered, as the cycleway quest was answered seconds (and not months/years) ago.

**Versions affected**

45.0-based ([my fork](https://github.com/mnalis/StreetComplete/actions/runs/2679787218)), Android 10 (on Huawei P30 Pro).
But note [previous repetition of bug](https://github.com/streetcomplete/StreetComplete/issues/4124) that looked exactly the same was in vanilla **_mainstream SC_** (not in my fork), so I hope the bug is the same, and so that this bugreport still helps.

westnordost:
fixes #4258

This actually does not only concern quests, but also overlays, which is why the solution has been done in `MapDataWithEditsSource` rather than in `OsmQuestController`.

Rather than making it really complicated for edge cases to deal with possible parallelism in `onReplacedForBBox`, I made it so that it cannot be executed in parallel. This doesn't change anything for the current configuration because OSM downloads can not happen in parallel (and it is very unlikely that this is ever changed).

---

Note that while testing this solution, I noticed that I could not reproduce the issue in the first place. This was because `Listeners.kt` replaced the previous `CopyOnWriteArrayList`-based listeners and `Listeners` just synchronizes **all** access:

I.e. any `Listeners.forEach(action)` was synchronized, meaning that a parallel 
```listeners.forEach { it.onReplacedForBBox(bbox, mapDataWithGeometry) }```
after map data download which triggers creating anew the quests for the downloaded map data and 
```listeners.forEach { it.onUpdated(updated, deleted) }```
after a single update of the data via an edit which triggers checking quests for just that one quest was not possible anymore. They were always executed in order.

**So, when trying to solve a quest while the osm data is being updated (=the quests are created), the UI would just seemingly freeze. Argh! The very thing I wanted to avoid.** I fixed this now, I hope my fix is fine. Second or third pair of eyes appreciated.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -66,10 +66,10 @@ class MapDataWithEditsSource internal constructor(
     private val onReplacedForBBoxLock = Any()
 
     // access to isReplacingForBBox is atomic (didn't want to pull in kotlinx-atomicfu dependency just for this)
-    private var isReplacingForBBoxLock = Any()
+    private val isReplacingForBBoxLock = Any()
     private var isReplacingForBBox: Boolean = false
 
-    private var updatesWhileReplacingBBox = MapDataWithGeometryUpdates()
+    private val updatesWhileReplacingBBox = MapDataWithGeometryUpdates()
 
     private val mapDataListener = object : MapDataController.Listener {
 
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -19,6 +19,7 @@ import de.westnordost.streetcomplete.data.osm.mapdata.MapDataController
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataRepository
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataUpdates
 import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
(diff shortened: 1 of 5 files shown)


Helium314:
Looks good to me.
The only very minor things I noticed is that `isReplacingForBBoxLock` and `updatesWhileReplacingBBox` could be `val`s

PR #5474, opened 2024-02-03
Small changeset bboxes

### LINKED ISSUE #4582: Keep changeset bbox small
**Use case**
A user has auto-upload off and keeps solving a lot of quests offline in a lot of different places, e.g. on a vacation. Then later, he uploads them all at once. All the quests get sorted into changesets that span very large areas.

This is somewhat inconvenient for people keeping track of changes in their area because changesets are sorted by bbox in the osm.org UI for looking at local changesets. I.e. the list of changesets displayed for a local area always also includes changesets whose bbox spans the whole continent because e.g. one edit in the same changeset was made on the westernmost verge and the other on the easternmost verge of the continent.

This sometimes leads to clashes between StreetComplete users and those monitoring the changes. The former are oblivious about how this works internally. See also https://news.ycombinator.com/item?id=33035946

**Proposed Solution**
When adding another edit to a changeset, check if it would enlarge the bounding box of the current changeset over a certain threshold (5 km?). If yes, open another changeset instead and close the previous one.

**However, do this only if the edit made is guaranteed to have been made on-site! I.e. 1. location *on* and 2. the app did not need to ask the user whether he has really checked it on-site.**

This requires also persisting this information on creating an edit. The reason for this is that we do **not** want to hide misuse of this app that should only be used for on-site edits by users who solve quests all over the place while not really being there.

westnordost:
Fixes #4582

### DIFF
--- a/app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsDaoTest.kt
+++ b/app/src/androidTest/java/de/westnordost/streetcomplete/data/osm/edits/ElementEditsDaoTest.kt
@@ -243,7 +243,8 @@ private fun updateTags(
             StringMapEntryModify("c", "d", "e"),
             StringMapEntryDelete("f", "g"),
         ))
-    )
+    ),
+    false
 )
 
 private fun revertUpdateTags(timestamp: Long = 123L, isSynced: Boolean = false) = ElementEdit(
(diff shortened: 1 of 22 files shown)


westnordost:
And, tested:

- SC asks about whether I am really is on-site, I lied  -> puts all into one changeset https://www.openstreetmap.org/changeset/147103648#map=10/53.6817/9.3542
- SC is sure that one is on-site because I spoofed my GPS location -> puts into several small changesets https://www.openstreetmap.org/changeset/147103632 + https://www.openstreetmap.org/changeset/147103633

Issue #5070, opened 2023-06-13
Use ViewModels


westnordost:
So far, the data in the view layer has been kept in the `Fragment`s and `Activity`s. Since these can be destroyed and recreated when the Android system decides to do this, the serialization and deserialization into a `savedInstanceState` must be done manually. Moving to ViewModels has the following advantages:

1. It's less error prone, possibly also less code. The data is not bound to the view lifecycle and hence does not have to be manually re-created

2. It provides a better separation between view logic and view data. Thus, the latter is likely mostly platform independent and could be re-used on other platforms with less effort. (View model patterns exists for iOS too, there is at least one kotlin multiplatform library in development that one might switch to in the future).

3. It will likely simplify some of the "talks to parent via listener" pattern, because some, or most of it(?) can probably be done with shared view models instead (i.e. a view model that is shared between a fragment and its parent fragment / activity).

4. It's what the cool kids use! Joking. It is actually pretty standard to use since many years and the more standard things are used in StreetComplete, the easier it is for new contributors to join

Doomsdayrs:
While it is nearing the end of the summer for me.

I have experience with implementing view models into android apps.

Would you care for some help in regards to this?

pikzen:
@westnordost Jetpack Compose can absolutely be used with ViewModels! In the case of a pure Compose architecture, various navigation solutions allow you to have proper scoping. 

However, a complete Compose rewrite of StreetComplete feels like a pipe dream, a branch that'll just live on the side and hopefully, when it's complete, it releases and more bugs have to be resolved. Not to mention some fun things, like map integrations. Having access to a MapView in a Compose-friendly manner requires _lots_of work (see https://github.com/googlemaps/android-maps-compose, where you need quite a bit of Compse-specific knowledge to make an implementation of it. It could be accessed with an AndroidView, but even then, that's quite a bit of work.

A happy middle ground that can slowly be released and tested out would be keeping Fragments, moving all the state into ViewModels, and using Compose for the UI (returning a ComposeView in onCreateView). 

Funnily enough, I saw this issue as I checked the repo wondering if some help with Compose or code cleanups would be needed, perfect timing :D

As for Compose Multiplatform, I don't believe it'd be of much use for now. Ignoring the fact that I don't believe there is StreetComplete for iOS, the implementation is a bit dodgy at the moment. Also, many composables would be missing for other platforms like desktop, etc.

westnordost:
[...]

PR #5482, opened 2024-02-11
ViewModels (a start)

westnordost:
A start for #5070 

So, ultimately we want to end up having a `ViewModel` for every `Fragment` that isn't purely UI. Until the switch to Jetpack Compose is complete, we can limit that to those places where it is worthwhile to switch, i.e. Fragments that have a lot of state (and would otherwise need to re-fetch this state from database or save it otherwise on `onDestroy`).

Now, this draft PR is just the first foray into that. Finding the patterns that fit best and can be applied _consistently_ throughout the app and are convenient to use. It is draft because it is open for comments. For this, I tried to create two ViewModels for two Fragments, as an example.

### Comments welcome

Our mid term goal is to migrate the UI also to Jetpack Compose, so we should keep that in mind to not necessitate another migration later. Also, the data layer in StreetComplete is blocking, i.e. nothing with coroutines or StateFlow. Hence, in the current architecture, it is up to the ViewModels to wrap everything up into `launch` et cetera. 

### ViewModels crash course

`ViewModels` hold and manage view state of a fragment/activity. They make it easier to hold the UI state because they transcend the view lifecycle. Furthermore, they manage access to the data layer, so the only thing that remains in fragments is controlling the UI.
`ViewModels` usually have a range of properties wrapped in `StateFlow`s, these are basically (coroutine friendly) observable properties, i.e. the fragment can do something if any `StateFlow` changes.


westnordost:
The main part is
- [ProfileFragment.kt](https://github.com/streetcomplete/StreetComplete/pull/5482/files#diff-a4897d401da7827ccc0867cd49c938294ee4c49d795d6a3173017c9ecbcf03b9) and [ProfileViewModelImpl.kt](https://github.com/streetcomplete/StreetComplete/pull/5482/files#diff-ac51a2906deb64a3c4cd86fcfe2bf4de650d36b7b8383140f3395136e4f45590)
- [LogsFragment.kt](https://github.com/streetcomplete/StreetComplete/pull/5482/files#diff-017e8a0406e9e2e79971fdc38a0d98b926a30a9a338d4f92e5a2e3adf57938c4) and [LogsViewModelImpl.kt](https://github.com/streetcomplete/StreetComplete/pull/5482/files#diff-503f59e1fa6d5a26f79900e39d0494952d5eac419f3184870e2c9f9b7d9aaf76)

For the logs, I also made a data model immutable and a few other minor things, but this is not really too relevant on the ViewModel stuff.

app/src/main/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsController.kt
```diff
@@ -178,11 +178,11 @@ class StatisticsController(
         val today = systemTimeNow().toLocalDate()
         val lastUpdateDate = Instant.fromEpochMilliseconds(lastUpdate).toLocalDate()
         lastUpdate = nowAsEpochMilliseconds()
+        activeDatesDao.addToday()
```
Doomsdayrs, line 181: Why was this moved? It was not explained in the commit message.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFragment.kt
```diff
@@ -97,7 +62,7 @@ class LogsFragment : TwoPaneDetailFragment(R.layout.fragment_logs) {
     }
 
     private fun onClickShare() = viewLifecycleScope.launch {
-        val logText = getLogs().format()
+        val logText = viewModel.logs.value.format()
         val logTimestamp = LocalDateTime.now().toString()
         val logTitle = "${BuildConfig.APPLICATION_ID}_${BuildConfig.VERSION_NAME}_$logTimestamp.log"
```
Doomsdayrs, line 67: Replacement with a ViewModel function of `getLogText` & `getLogTitle` would make this cleaner for the UI.

This is business code anyways (producing a state)

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModel.kt
```diff
@@ -0,0 +1,12 @@
+package de.westnordost.streetcomplete.screens.about
+
+import androidx.lifecycle.ViewModel
+import de.westnordost.streetcomplete.data.logs.LogMessage
+import de.westnordost.streetcomplete.data.logs.LogsFilters
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+abstract class LogsViewModel : ViewModel() {
+    abstract val filters: MutableStateFlow<LogsFilters>
```
Doomsdayrs, line 10: Any reason in particular to allow the UI modification to the filters instead of having a `setFilters` function?

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+
+class LogsViewModelImpl(
+    private val logsController: LogsController
+) : LogsViewModel() {
+
+    private val logsControllerListener = object : LogsController.Listener {
+        override fun onAdded(message: LogMessage) {
+            if (filters.value.matches(message)) {
+                // TODO this is hugely inefficient (log list is copied every time a single entry is added!!!)
+                logs.update { it + message }
```
Doomsdayrs, line 27: To solve this, we can replace `logs` with a `MutableSharedFlow` with a replay of 1 and have a private `ArrayList` as the backing. Whenever the `ArrayList` is appended to, you can `emit` the `ArrayList` into `MutableSharedFlow` (which does not care about equation of the emitted values)

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+            // get logs anew whenever filters changed
+            filters.collect { f ->
+                logs.update { withContext(Dispatchers.IO) { logsController.getLogs(f) } }
+            }
+        }
+    }
+
+    override fun onCleared() {
+        logsController.removeListener(logsControllerListener)
+    }
```
Doomsdayrs, line 53: Utilize a [Callback Flow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/callback-flow.html) instead for this.

```kotlin
logs = filters.transformLatest { f -> // reloads list with filter change
 emit(logController.getLogs(f)) // emit initial list
// emit based on subscription
 emitAll(
  callbackFlow {
   // put listener creation here
   logsController.addListener(listener)
   awaitClose {
    logsController.removeListener(listener)
   }
}.flowOn(Dispatchers.IO);
```

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+}
+
+private fun LogsController.getLogs(filters: LogsFilters) =
+    getLogs(
+        levels = filters.levels,
+        messageContains = filters.messageContains,
+        newerThan = filters.timestampNewerThan?.toEpochMilli(),
+        olderThan = filters.timestampOlderThan?.toEpochMilli()
+    )
```
Doomsdayrs, line 62: If this won't be used else where, why is it not an internal ViewModel function?

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
```diff
@@ -114,33 +52,60 @@ class ProfileFragment : Fragment(R.layout.fragment_profile) {
...
+        observe(viewModel.editCount) { count ->
+            binding.editCountText.text = count.toString()
+        }
+        observe(viewModel.editCountCurrentWeek) { count ->
+            binding.currentWeekEditCountText.text = count.toString()
+        }
+        observe(viewModel.achievementLevels) { levels ->
+            binding.achievementLevelsContainer.isGone = levels <= 0
+            binding.achievementLevelsText.text = levels.toString()
+            binding.achievementLevelsText.background.level = min(levels / 2, 100) * 100
```
Doomsdayrs, line 76: calculations done in UI, slow downs. Better to be done in view model via a custom data object and flows.

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
```diff
@@ -114,33 +52,60 @@ class ProfileFragment : Fragment(R.layout.fragment_profile) {
...
-            updateDaysActiveText()
-            updateGlobalRankTexts()
-            updateLocalRankTexts()
-            updateAchievementLevelsText()
-            updateDatesActiveView()
+        observe(viewModel.userName) { name ->
+            binding.userNameTextView.text = name
+        }
+        observe(viewModel.userAvatarFile) { file ->
+            val avatar = if (file.exists()) BitmapFactory.decodeFile(file.path) else anonAvatar
```
Doomsdayrs, line 64: IO Business code in UI, let VM return a `Bitmap?` which is null if the decoding fails, then default to anon avatar in UI.

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
```diff
@@ -114,33 +52,60 @@ class ProfileFragment : Fragment(R.layout.fragment_profile) {
...
+                context.dpToPx(18),
+                context.dpToPx(2),
+                context.dpToPx(4),
+                context.resources.getColor(R.color.hint_text)
+            ))
+        }
+        observe(viewModel.daysActive) { daysActive ->
+            binding.daysActiveContainer.isGone = daysActive <= 0
+            binding.daysActiveText.text = daysActive.toString()
+            binding.daysActiveText.background.level = min(daysActive + 20, 100) * 100
```
Doomsdayrs, line 96: Calculation on UI instead of on VM, see previous comment.

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
```diff
@@ -156,181 +121,90 @@ class ProfileFragment : Fragment(R.layout.fragment_profile) {
...
-        if (statistics?.countryCode != previousStatistics?.countryCode ||
-            previousStatistics?.rank == null || rank > previousStatistics.rank) {
+
+        if (previousRank == null || previousRank < rank) {
             updateRank(rank)
         } else {
-            animate(previousStatistics.rank, rank, container, updateRank)
+            animate(previousRank, rank, circle, updateRank)
         }
     }
```
Doomsdayrs, line 206: The entire update rank etc seems like VM code?

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileViewModel.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    abstract var lastShownGlobalUserRank: Int?
+    abstract var lastShownGlobalUserRankCurrentWeek: Int?
+    abstract var lastShownLocalUserRank: CountryStatistics?
+    abstract var lastShownLocalUserRankCurrentWeek: CountryStatistics?
+
+    abstract fun logOutUser()
+}
+
+data class DatesActiveInRange(val datesActive: List<LocalDate>, val range: Int)
```
Doomsdayrs, line 37: Move to its own file?

Doomsdayrs (changes requested):
- Comments left on what I noticed.
- Suggestible to ensure View Models & Models are in their own directory to keep things organized and easy to locate.

app/src/main/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsController.kt
westnordost: It was an unrelated bugfix, committed to master now

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsFragment.kt
westnordost: I see your point, but I'd like to keep view models platform agnostic and this method to share the logs is specific to Android.

I'm not sure what to do yet with platform specific code in the `Fragment`s, as they certainly have no place in Compose code either.
I think I am going to postpone thinking about this when the time actually comes. You're probably right that in the end, it will end up one way or another in the ViewModel, with some `expect` / `actual`.

Now that you made me think about it... for starting an Activity (to send an Intent), you _need_ a `context` but at the same time, docs say that the activity etc. should not be a member in a ViewModel, so I do wonder how it would have to look in the end.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
westnordost: What do you mean? It is already private

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
westnordost: Ah, nice idea. I didn't foray into `SharedFlow` yet, but it seems quite straightforward.

So, like this?

https://github.com/streetcomplete/StreetComplete/pull/5482/commits/314f05f75df208a543a3566a5c81f6a31731ae8d

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
westnordost: (Ah well, the general pattern in StreetComplete code is to define static functions that don't have anything to do with the state of the class outside of that class, even if they are used only for that one class.)

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileViewModel.kt
westnordost: This is okay consistent to other code in this repo. Data classes somewhat related to another interface/class are often kept together in the same file or even within the interface/class, especially if it is just a one-liner like that.

westnordost:
Regarding the comments on view logic in the UI code (rather than ViewModel):

Well, currently I envisioned ViewModels to have the following responsibilities:
- hold UI state to transcend UI lifecycle (as a matter of course)
- manage access to data layer
- (UI) business logic

An example of what I understand as (a state in) UI business logic would be 
```kotlin
val isLoggedIn: Boolean
```
while an example of what I understand as (a state in) pure UI logic would be
```kotlin
val isLogoutButtonVisible: Boolean
//or
val rankViewBackgroundTint: Color
```

Now, there is obviously a smooth transition between the two. So far, I pretty much left out anything from the ViewModel that would be considered view logic. 
Are there any arguments for a certain middle ground or one or the other extreme?

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileFragment.kt
westnordost: `Bitmap` is Android specific, though, and my aim would be to keep ViewModels free of any platform specific code. Compose Multiplatform maybe brings their own `Bitmap` (like) + File system access classes (probably not), but then the ViewModels depend on a specific UI framework.
I'd like to avoid that. Anyway, as with the other thing with Intent/Activity, I tend towards postponing thinking about that.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModelImpl.kt
westnordost: Hm, and this also does not solve the issue really. Yes, the array list is not copied in the view model anymore, but in the LogsFragment it is. The `LogsAdapter` needs to know whether just one item was added at the end or the whole list was replaced, because, well, that's how these `RecyclerView.Adapter`s work, with `notifyItemInserted` etc.

Maybe this isn't an issue in Compose, I don't know.

(Anyway, I will revert the linked commit for now)

Doomsdayrs:
I am going to make a branch with a few commits, as I feel like that would be faster to explain what I notice.

app/src/main/java/de/westnordost/streetcomplete/screens/about/LogsViewModel.kt
Doomsdayrs: It is a bit hard to explain what I am thinking.

Essentially, by having a `setFilters` function instead, it makes it clear where I/O between the VM and UI is occurring. This can be enhanced with the VM handling logging of function calls.

Doomsdayrs:
The more I look at this code the more horrified I am by the duplicated code and its efforts.

For example, the `Dao` is not an actual Room `DAO`, but a hand crafted SQL wrapper. Room already handles generation of this code, but it also provides functionality to directly observe the database changes.
https://medium.com/androiddevelopers/room-flow-273acffe5b57

This then leads to using `LogController` to create its own listener system to handle feedback.

Which then leads to horrific memory duplication in the UI layer.

westnordost:
Yes, the app doesn't use Room, it is older than Room, modern architecture (Jetpack) components etc.. (development started 9 years ago)

It is also not a possibility to migrate to Room because it is not multiplatform. (Subsequently) the data layer is also different from what one finds in modern guides. I think I noted that here or in the other issue ticket yesterday.

It is not out of question to refactor this one day to be more in line with modern architecture design guides, but it is important to do such refactorings one step at a time and then each consistently finishing that refactoring step before continuing onto the next one, otherwise one will end up with an unmaintainable mess of inconsistently applied patterns. For now, what's on the menu is to consistently introduce ViewModels (- and after that or partly intermixed with that use Jetpack Compose for UI.) - this means that it falls to the ViewModels to stream the data into flows because the data layer is not already flow/coroutine-based.

westnordost:
(In case you are interested, here is a bird's view overview over the data layer only of the app, it also includes explanations about the responsibilities of the different class types: https://raw.githubusercontent.com/streetcomplete/StreetComplete/master/res/documentation/overview_data_flow.svg . As you see, it is quite extensive, and actually, the `Logs` stuff is missing from this, because it is pretty new)

Doomsdayrs:
It took me a bit to test due to the snow storm here in New York.

The only think I wonder is how to get a log to occur while in the logs view.

How does 1cce1a0 look to you @westnordost ?

westnordost:
Oh yeah, I see its circulating in the news right now. In Hamburg, on the other hand, it's just mud season (40°F, heavy clouds, bit of rain).

Easiest way to make logs appear while being in the log view is to start a download (menu -> download map here) and then go to the logs view. The download usually takes a bit (depending on the download map size and data density) and logs a lot of things. (Since the log view is rarely opened and things are rarely logged while in that view, it would actually not be that bad if updating the live logs was hugely inefficient)

Ah, yes, I remember I've looked at the documentation of callback flow (+ transformLatest, stateIn) yesterday and was unable to wrap my head around it. Maybe I need to need to look through the flow documentation in general again first I'll try again tomorrow.

Doomsdayrs:
> Oh yeah, I see its circulating in the news right now. In Hamburg, on the other hand, it's just mud season (40°F, heavy clouds, bit of rain).

The weather for the Winter was so far Muddy at a consistent 10°C~.

> Easiest way to make logs appear while being in the log view is to start a download (menu -> download map here) and then go to the logs view. The download usually takes a bit (depending on the download map size and data density) and logs a lot of things. (Since the log view is rarely opened and things are rarely logged while in that view, it would actually not be that bad if updating the live logs was hugely inefficient)

Corrected a small issue that I remembered, but did not remember at the time of writing with 0ffddeb5e557bc560038c93e788cdfe8fa021c6f.

Also optimized the view while I was at it with b2e90eec7affc84c25df383e0a71d0df442f3f9c.
 
> Ah, yes, I remember I've looked at the documentation of callback flow (+ transformLatest, stateIn) yesterday and was unable to wrap my head around it. Maybe I need to need to look through the flow documentation in general again first I'll try again tomorrow.

If you would like, we can skill share via a voice call & screen share? I should be free after 15:00 EST.

Doomsdayrs:
> However, it's nice and all that this is idiomatic (experimental?) 

They mark it experimental, it hasn't changed for years now. It's just a liability thing for themselves.

> Kotlin flow-code, but doesn't [0ffddeb](https://github.com/streetcomplete/StreetComplete/commit/0ffddeb5e557bc560038c93e788cdfe8fa021c6f) make it hugely inefficient again, as the whole list is copied for every single log entry added again?

Yes, which is annoying. But it happens in a flow now, and should not block the UI. And the JVM will optimize the routine after the first 2~ iterations. From testing, I saw no visible lag. Do you see any lag?

Aside the greater lag came from using `notifyDataSetChanged` on the UI thread :wink: .

> Uff, to use your words, it's horrible to see how this is getting more and more complicated. The more complicated the code, the more difficult to maintain. 

We currently face a code base that was built without proper architecture & abstractions. Technical debt is a debt, one has to pay eventually.

> In prospect of the UI to be reimplemented in Compose, I'd also have not bothered with implementing that, 

Mix implementation is possible, but do note the application will bloat in size slightly holding two UI stacks.

Keep things as organized as you have, slow and steady implementation can be achieved.

> but I am fine with merging your improvement. It is good that in very few places in the app, there is such a live-stream of new data that should be displayed.

I can later spend some time quickly making a PR that implements `DiffUtil` throughout the application. 

A separate effort should be made to replace the current fake "Daos" with real Room Daos, since that would help with proper live-streams from the DB without the duplication that currently occurs.

> Anyway, feel free to create a PR that merges into this PR, it is maybe easier to talk about the improvements then.

Alright, will do that soon.

> Regarding the video call, rather not for now, flows are very new to me and I am slow on the uptake to understand new concepts, I don't think a quick explanation would help.

Not a problem, Just offering a sturdier hand if needed.

westnordost:
Hmm, well, thank you for the review and all those suggestions. This PR is a sort of test how the viewmodels should look after the refactor. I think I will roughly follow these guidelines:

- no UI (framework) specific code in ViewModels, no Android-specific stuff in the ViewModels (it is TBD where that code should go, for the time being, it stays in the Fragments)
- only use `StateFlow`s when it is actually needed, i.e. values that are displayed and expected to change, otherwise just normal fields to keep complexity down. For data that is fetched asynchronously (from DB), there doesn't seem to be another choice, though.
- bundling observable data into a data structure as done in Android examples is okay for related data (from same data source). If it leads to more readable and/or less code and is not performance critical, bundling it all into an ui state data class would be fine, too. Compose does not really care anyway.
- have an accompanying interface for the ViewModel in order to not expose `MutableStateFlow` on its interface (but just the `StateFlow`)
- prefer not exposing `MutableStateFlow` on ViewModel interface (use setters instead), for cleaner interface
- keep directly view related data in the view rather than in the view model (`isLoggedIn` in in the viewmodel, something like `rankBackgroundGradient` in the view). May change later during refactor to Jetpack Compose.

Doomsdayrs:
Isn't dc7768f more complicated then 5dcc124 ?

westnordost:
CopyOnWriteArrayList is JVM-only. Plus, it does not solve the performance issue, because... on every write, the array is still copied, just like the name says :-)

Regarding `replayCache[0]`, I find it relatively ugly to have that on the interface. From the point of view of any observer, it is an implementation detail that internally, `logs` is a SharedFlow.

PR #5483, opened 2024-02-11
Switch from java.net.URL to Ktor client

### LINKED ISSUE #5408: Multiplatform HTTP Client (use Ktor Client)
Use a multiplatform HTTP client. Currently, we are using `java.net.URL` for these.

### HTTP connections in the app are used for

- any communication with OSM servers, i.e. uploading map data, notes and gpx tracks, downloading map data, notes, user data, ...

- querying a [simple text file](https://westnordost.de/streetcomplete/banned_versions.txt) ([`VersionIsBannedChecker`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+VersionIsBannedChecker.kt&type=code)) 

- communicating with [an API to upload and activate photos](https://github.com/exploide/sc-photo-service) ([`StreetCompleteImageUploader`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+StreetCompleteImageUploader.kt&type=code)) 

- querying user statistics from [a simple API](https://github.com/streetcomplete/sc-statistics-service) ([`StatisticsDownloader`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+StatisticsDownloader.kt&type=code)) 

- downloading OSM user avatars ([`AvatarsDownloader`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+AvatarsDownloader.kt&type=code)) 

- querying oneway traffic data from [another simple API](https://github.com/streetcomplete/oneway-data-api) ([`TrafficFlowSegmentsApi`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+TrafficFlowSegmentsApi.kt&type=code)) 

- during OAuth access token negociation ([`OAuthAuthorization`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+OAuthAuthorization.kt&type=code))

### HTTP Client Requirements

- multiplatform support for at least Java/Android and iOS

- stable, well-tested and ideally (semi-)official / well-maintained

- should be injectable, i.e. the http client can be passed as a (constructor) parameter to the function/class that uses it. Important in order to easily test the aforementioned classes

Below some comments on libraries I've looked at.

Note that of the candidates mentioned below, some are asynchronous clients, i.e. use coroutines, which is fine, but may necessitate more changes in the app architecture than simply replacing the java `URL` stuff: The upload and download process that uses the above-mentioned HTTP APIs currently expect these things to be synchronous and manages concurrency itself.

#### OkHttp

OkHttp is said to have basic [multiplatform support starting with 5.0](https://github.com/square/okhttp/blob/master/CHANGELOG.md), but v5.0 has been in alpha since 2021-01 and since one year, no new alpha of v5.0 has been released.

So, while OkHttp is a very very popular library, it is unknown when this ever will be stable. On the other hand, [according to the current main developer](https://github.com/square/okhttp/issues/7883#issuecomment-1837604213), only the APIs are unstable [, not the implementation].

#### Ktor Client

A candidate is [Ktor Client](https://ktor.io/docs/getting-started-ktor-client.html). 

This seems to be a bit of a rabbit hole, because it seems to be designed to be (used as) part of a larger framework, in particular in a project where the server and client(s) basically live in the same repository and share a good portion of the code, which is not the case for our app.
On the other hand, it seems [to plug-in well with kotlinx-serialization](https://ktor.io/docs/serialization.html) (for e.g. JSON and XML), so it seems to be the more integrated solution. 

The library is also featured in the [Compose Multiplatform Wizard](https://terrakok.github.io/Compose-Multiplatform-Wizard/)

#### Fuel

Another candidate is [Fuel](https://github.com/kittinunf/fuel). However, it does not look to be eligible because:

- it is only multiplatform starting with version 3, which is still alpha 
- little activity in repo, neither in response to a security vulnerability nor on v3 development. This does not inspire confidence that it is well-maintained
- all fields of the the HTTP response class are platform dependent (in the current alpha) - so, only multiplatform on paper.

#### Other

- khttp is dead. No commit since 2020. Also, not multiplatform anyway.

- kohttp is dead. No update since 2020. Also, not multiplatform anyway.

---

In general, both Ktor and OkHttp 5 seem okay, though I lean towards Ktor because OkHttp 5 is not stable yet and Ktor seems to be the solution that is more geared towards a Kotlin (MP) ecoystem.

adrianclay:
Part of the effort in making the codebase compatible with iOS.

Closes #5408


westnordost:
Wow, that's great!

This PR is marked as Draft, does this mean you still have some questions or doubts, or are still working on it?

app/src/main/java/de/westnordost/streetcomplete/quests/oneway_suspects/data/TrafficFlowSegmentsApi.kt
```diff
@@ -2,24 +2,31 @@ package de.westnordost.streetcomplete.quests.oneway_suspects.data
...
             bbox.min.latitude,
             bbox.max.longitude,
             bbox.max.latitude
         ).joinToString(",") { it.format(7) }
 
-        val url = URL("$apiUrl?bbox=$leftBottomRightTopString")
-        val json = url.openConnection().getInputStream().bufferedReader().use { it.readText() }
-        return parse(json)
+        val response = httpClient.get("$apiUrl?bbox=$leftBottomRightTopString") {
+            expectSuccess = true
```
westnordost, line 27: So what happens when it is not successful? Does it throw an exception? [The documentation is really not that helpful.](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins/expect-success.html)

app/src/main/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsDownloader.kt
```diff
@@ -1,31 +1,30 @@
...
+            header("User-Agent", ApplicationConstants.USER_AGENT)
+        }
+        when (response.status) {
+            HttpStatusCode.OK -> {
+                return statisticsParser.parse(response.body())
+            }
+            else -> {
+                val errorMessage = response.status.description
+                val errorDescription = response.body<String>()
+                throw Exception("${response.status.value} $errorMessage: $errorDescription")
```
westnordost, line 27: I know you just copied the behavior from before, but maybe it is possible to tell Ktor to just throw an exception by itself instead of creating it oneself from the information extracted prior from `HttpURLConnection`?

app/src/main/java/de/westnordost/streetcomplete/data/upload/VersionIsBannedChecker.kt
```diff
@@ -1,29 +1,33 @@
...
+    suspend fun get(): BannedInfo {
         try {
-            connection = (URL(url).openConnection() as HttpURLConnection)
-            connection.inputStream.bufferedReader().use { reader ->
-                for (line in reader.lineSequence()) {
-                    val text = line.split("\t".toRegex())
-                    if (text[0] == userAgent) {
-                        return IsBanned(if (text.size > 1) text[1] else null)
+            val response = httpClient.get(url)
+            if (response.status == HttpStatusCode.OK) {
```
westnordost, line 19: If `expectSuccess` does what I suppose it does, maybe use that instead of checking manually for status code?

In any case, it would be nice if this was consistent in the different downloader/uploader classes.

app/src/main/java/de/westnordost/streetcomplete/ApplicationModule.kt
```diff
@@ -15,4 +16,5 @@ val appModule = module {
     single { CrashReportExceptionHandler(androidContext(), get(), "streetcomplete_errors@westnordost.de", "crashreport.txt") }
     single { DatabaseLogger(get()) }
     single { SoundFx(androidContext()) }
+    single { HttpClient() }
```
westnordost, line 19: Is HttpClient threadsafe? I.e. is it recommended to use one instance of HttpClient for everything, or create new instances for different addresses?

In any case, the module is the perfect place to set a few settings on the HttpClient. Is it for example possible to set the default user agent here to be added to the header of every request?

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/AvatarsDownloader.kt
```diff
@@ -42,11 +47,14 @@ class AvatarsDownloader(
...
-    fun download(userId: Long, avatarUrl: String) {
+    suspend fun download(userId: Long, avatarUrl: String) {
         if (!ensureCacheDirExists()) return
         try {
             val avatarFile = File(cacheDir, "$userId")
-            URL(avatarUrl).saveToFile(avatarFile)
+            val response = httpClient.get(avatarUrl) {
+                expectSuccess = true
+            }
+            response.bodyAsChannel().copyAndClose(avatarFile.writeChannel())
```
westnordost, line 57: I've seem to remember that I read somewhere, that Ktor Client is based on / uses Okio because it is multiplatform.

Is it (not) possible to replace the dependency to `java.io.File` here and elsewhere? (E.g. the photo uploader also accesses the file system IIRC)

westnordost (commented):
Here's a partial preliminary review.

I see you also updated an outdated link somewhere in the comments, cool! (It's a small thing but I take from that that you work quite thoroughly)

Also, thank you very much for the additional tests! I didn't look at them yet, but it always bothered me that the URL-stuff was not well dry-testable.

**Edit**: I noticed that at least a few tests also still use Java `URL`, `UrlDecoder` etc. does Ktor have classes to replace those?

adrianclay:
Thanks for the partial review ❤️.

Raised as draft because I’m still plodding away, but wanted to give visibility of my progress publicly.

Will continue adding some tests at _some_ point, hopefully one evening this week, and then switch over to some of the improvements you suggested.

app/src/main/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorization.kt
```diff
@@ -39,12 +44,12 @@ import kotlin.io.encoding.ExperimentalEncodingApi
  *              several at once (using the same redirect URI)
  */
 @Serializable class OAuthAuthorization(
```
westnordost, line 46: Right, I should have kept in mind when I wrote this class that the HttpClient should be injected and hence the structure of this class cannot be kept that way.

I think it would be cleaner to make OAuthAuthorization a pure data class (maybe rename to OAuthAuthorizationParams) and move all logic into class that you named OAuthService.

app/src/main/java/de/westnordost/streetcomplete/quests/oneway_suspects/data/TrafficFlowSegmentsApi.kt
adrianclay: Good question, so `expectSucess` will throw a [few different types of exception](https://github.com/ktorio/ktor/blob/2.3.8/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultResponseValidation.kt#L53-L56). I've added some test cases to demonstrate `ClientRequestException` and `ServerResponseException`.

app/src/main/java/de/westnordost/streetcomplete/ApplicationModule.kt
adrianclay: From what I've read, the [HttpClient creates its own threads](https://ktor.io/docs/create-client.html#close-client), so based on that I'd assume that it is threadsafe. The guidance also recommend:

> Note that creating HttpClient is not a cheap operation, and it's better to reuse its instance in the case of multiple requests.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/AvatarsDownloader.kt
adrianclay: I was also thinking that removing `java.io.File` would be a useful thing to do, but I'm also concious that this PR is getting big, and replacing `File` usage could itself be an independent bit of work.

I'll keep the Java API for now, and will do some investigations into Okio separately.

app/src/main/java/de/westnordost/streetcomplete/data/user/oauth/OAuthAuthorization.kt
adrianclay: So, I've done a little bit of refactoring of this class. I've kept a few methods in the `OAuthAuthorizationParams` because they are highly coupled to the fields of that class, and nothing else, so it made sense to me to keep them in that class. Happy to do more moving around if you disagree though.

app/build.gradle.kts
```diff
@@ -153,6 +153,11 @@ dependencies {
     // scheduling background jobs
     implementation("androidx.work:work-runtime:2.9.0")
 
+    // HTTP Client
+    implementation("io.ktor:ktor-client-core:2.3.8")
+    implementation("io.ktor:ktor-client-cio:2.3.8")
```
adrianclay, line 158: Other [engines](https://ktor.io/docs/http-client-engines.html#requirements) exist. CIO is the "default" that Ktor talk about, but I see that OkHttp is already in use by `MapFragment.kt`.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/AvatarsDownloader.kt
westnordost: Right, ok

app/build.gradle.kts
westnordost: The OkHttp code in MapFragment is specific to tangram-es which is due to be removed, so don't bother with it.

app/src/test/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsParserTest.kt
```diff
@@ -1,80 +0,0 @@
-package de.westnordost.streetcomplete.data.user.statistics
-
-import kotlinx.datetime.Instant
-import kotlinx.datetime.LocalDate
-import kotlin.test.Test
-import kotlin.test.assertEquals
-
-class StatisticsParserTest {
```
westnordost, line 8: You removed this test. I think it is better if the `StatisticsParserTest` remained and tested the statistics parser, while the `StatisticsDownloaderTest` *only* tests the downloading. I.e. mocks an own `StatisticsParser` that just always returns a valid `Statistics` or whatever is required by the specific test.

westnordost (approved):
Looks very good! I always like it when a refactor actually reduces the amount of code.

I skimmed through the test cases you added. Thank you a lot for that, they LGTM except for one thing:

app/src/test/java/de/westnordost/streetcomplete/data/user/statistics/StatisticsParserTest.kt
adrianclay: Yeah, I think this is my opinions on software cohesion leaking out. 😓 In my head the code to communicate with an API, and the code to convert from the APIs representation to application domain representation make a lot of sense to stay together. I can't imagine the parser being used in any context other than with that API. I also suspected that the only reason for separate `StatisticsParser` and `StatisticsDownloader` classes was because the "parsing" behaviour was hard to test previously, and I'd actually push for combining the two classes 🤣 .

I'm happy to put my thoughts on software design in a box and split out the tests later this week.

westnordost: But wouldn't you agree that as long as the classes are separate, the tests should also be separate? Sure, they are closely related, that's why they are in the same package. There are several levels of how functionality that is closely related/interdependent can be separated - in the same package, in the same file, in the same class, in the same function, ...and, well, I guess different philosophies do exist. Separating parser, (generator,) and IO is however done quite consistently throughout the project IIRC and I like the code base to stay as consistent as possible.

Nevertheless, I would be interested to know where you would draw the line, when to separate and when not to separate, and how to separate, if you like to share that with me.

adrianclay: > But wouldn't you agree that as long as the classes are separate, the tests should also be separate?

I'm more of the opinion that test design and code design should be free to diverge independently. Tests exist to test behaviours of the system, not classes. In some cases, I may want to write several classes worth of tests, for a single production class. In some cases I may be able to define the behaviour of the software within one test class, but want to use several different production classes to implement said behaviour. See also [this video](https://youtu.be/EZ05e7EMOLM?feature=shared&t=1350).

> Separating parser, (generator,) and IO is however done quite consistently throughout the project IIRC and I like the code base to stay as consistent as possible.

Agree with consistency being important 👍🏻

> Nevertheless, I would be interested to know where you would draw the line, when to separate and when not to separate, and how to separate, if you like to share that with me.

When I'm thinking about testing, I'm usually thinking first about what boundaries do I have around my softwares components or want to have around my components. In this case, I see:
1. An API which provides a very hard boundary between once software component, and the mobile app. This being a boundary or not is outside of our control.
2. We also have this `data class Statistics` which somebody has sensibly designed, in a way where it is used across multiple different components of the app, and generated from one place. Allows statistics to be cached locally, or even calculated locally in some future without needing to change the design of other components. A useful boundary.

To me, with this mindset of software boundaries I'd then be wanting good confidence that a particular end to end boundary works, and being lazy I'd default to testing it all together as one. See also [Layered Architecture](https://ddd-practitioners.com/home/glossary/layered-architecture/).

I don't see the StatisticsParser as a valuable boundary. `StatisticsDownloader` has an explicit dependency on `StatisticsParser` via the constructor parameter. `StatisticsParser` (as far as I can imagine) would only ever be used by a `StatisticsDownloader` instance. These classes will likely always appear as a pair. Friends forever 😄.

westnordost: If tests are not coupled to classes, it becomes harder to ascertain the test coverage on a glance, though. But yeah, I recognize that this is a very imprecise measure anyway and there are better tools for that.

I've watched the talk (till min 40). Interesting. Though in general he starts from a pretty dogmatic tests-first / "100% test coverage" uhm... hellhole and I never bought into TDD that much that I could relate or experienced any of the issues he described with TDD - TDD has been the exception rather than the rule for me.

Anyway, what you write makes sense. In a nutshell, if you find it cleaner to merge them into one class, I'd be fine with that, but IMO if you do, it should happen both for the tests and the production code, i.e. the parser should not be injectable in the `StatisticsDownloader` in that case but be part of that class.

We say we make dependencies injectable for easier testing. But when the tests should ideally not concern themselves with interactions within a "module", i.e. interactions that are implementation details from a more ehm... birds eye view the tests should ideally be written from, then it also makes no sense to have these "dependencies" injectable.

I fear, if one follows that paradigma to the extreme again (like the speaker with TDD apparently at the beginning), one could end up with quite monolithic class giants that actually become more complex to test because they do many things at the same time.

I think, the beauty of tearing apart into separate classes functionalities even within one use case / module / ... is to make sure that ideally one class / file only does one thing, and that one thing can easily be tested. Whether or not that thing is only used at one place within the same module seems not to be too relevant to the test itself.

Anyway, I don't have such a strong opinion about it and I feel we are not actually that far away from each other opinion-wise. The `Statistics*` is a small thing and whether it is done one way or another ultimately doesn't really change anything. Where exactly the line is drawn (=when something is an own class, an own function, ...) is always subject of consideration

app/src/main/java/de/westnordost/streetcomplete/data/upload/Uploader.kt
```diff
@@ -27,7 +27,7 @@ class Uploader(
 
     private val listeners = Listeners<UploadProgressSource.Listener>()
 
-    private val bannedInfo by lazy { versionIsBannedChecker.get() }
+    private lateinit var bannedInfo: BannedInfo
```
westnordost, line 30: `lateinit` always looks like a hack in Kotlin to me. Is it not possible to keep it `Lazy` or use a similar data structure? Is there a suspending Lazy?

adrianclay: Agreed, I tried to find something nicer but this was the best I could find. I'm not a Kotlin person, so perhaps my research wasn't good enough 😞.

Happy to replace with anything you find!

westnordost: Hmm... I think [`Deferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/) with a `[CoroutineStart.LAZY](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-start/-l-a-z-y/index.html)` might be it, didn't try it out though.

Or maybe we are overthinking it, and it could be just a nullable field.

westnordost: On second though, I am not sure about it... using GlobalScope is certainly not correct there... but without it, it would need to be suspending but `suspend val bannedInfo` is not possible as far as I know.

But no matter, if you are not a Kotlin person, I will look into it myself. As far as your work is concerned, the PR is complete and approved.

westnordost:
Thank you again not only for implementing this, but also for replacing the usages of URL-related APIs (encode url etc.) - it certainly was an expansion of the scope of the initial task, so I am glad that you took care of it in one PR :-) 

Also, I love that you added all those unit tests for the previously untested HTTPS communication. I know I wouldn't have done that (out of laziness), so I am glad that we have someone as a contributor so committed to produce well-tested and high quality code like you! Reading that Kotlin is not even your home turf, I find it even more laudable. :-)

app/src/main/java/de/westnordost/streetcomplete/data/upload/Uploader.kt
westnordost: I've looked a bit more. Including the Kotlin documentation doesn't mention anything about lateinit var should be preferred only for legacy APIs or something like that, so I guess it is fine after all, especially there does not exist a proper suspending lazy.

PR #5488, opened 2024-02-15
Implementation ViewModel improvements

Doomsdayrs:
From #5482


westnordost:
Your suggestion regarding SharedFlow, to solve the array copying, is it (not) also possible here?

westnordost:
I mean, it is not that important, as you've tested. It did not lag before and it still does not lag, plus it is a screen that is rarely even visited. Live update of the logs is something of a nice-to-have-feature anyway.

Just curious.

Doomsdayrs:
> Your suggestion regarding SharedFlow, to solve the array copying, is it (not) also possible here?

I think it was simply time constraints and I forgot.

5dcc1247 does this

PR #5495, opened 2024-02-22
add 'gas' as an option for bbq fuel

k-yle:
#5211 added a new quest to tag the fuel of a barbeque. The current options are [`wood`](https://taginfo.osm.org/tags/fuel=wood), [`electric`](https://taginfo.osm.org/tags/fuel=electric), or [`charcoal`](https://taginfo.osm.org/tags/fuel=charcoal).

From my experience in Australia and New Zealand, public barbeques can sometimes be powered by gas. This could be LPG, gas mains, or BYO propane tank.

This PR adds [`gas`](https://taginfo.osm.org/tags/fuel=gas) as a new option.


matkoniecz:
Can you add it also on https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbbq then if it is one more valid value?

westnordost (changes requested):
quest_bbq_fuel_gas must be defined in app/src/main/res/values/strings.xml

matkoniecz (approved):

PR #5496, opened 2024-02-22
Add quest about tree leaf type

matkoniecz:
Side effect of testing new thing overlay

Thanks to showing nearby objects ideas from #83 has become more viable


app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/TreeLeafTypeItem.kt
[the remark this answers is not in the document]
matkoniecz: From what I remember @westnordost was not fan of inheritance/etc in such cases and there is barely any logic here anyway

app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/AddTreeLeafType.kt
```diff
@@ -0,0 +1,34 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.OUTDOORS
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddTreeLeafType : OsmFilterQuestType<TreeLeafType>() {
+    override val elementFilter = """
+        nodes with natural = tree and !leaf_type
```
westnordost, line 15: Should not be asked if another tag already gives a more precise information about the type of tree

matkoniecz:
Hmm, are people tagging palms as trees?

app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/AddTreeLeafType.kt
matkoniecz:
```
        and !species:wikidata and !species:wikipedia and !genus:wikidata and !genus:wikipedia
        and !genus and !species
```

anything else?

westnordost: genus:de etc?

Check out NameAndLocationLabel, there is some tree taxon stuff

matkoniecz: should be done now

app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/AddTreeLeafType.kt
```diff
@@ -0,0 +1,37 @@
...
+        and !species:wikidata and !species:wikipedia and !genus:wikidata and !genus:wikipedia
+        and !genus and !~genus:.* and !species and !~species:.* and !taxon and !~taxon:.*
+        and !taxon:species and !~taxon:species:.* and !taxon:genus and !~taxon:genus:.*
```
FloEdelmann, line 18: Some reordering to make it more readable:

```suggestion
        and !species and !species:wikidata and !species:wikipedia and !~species:.*
        and !genus and !genus:wikidata and !genus:wikipedia and !~genus:.*
        and !taxon and !~taxon:.*
        and !taxon:species and !~taxon:species:.*
        and !taxon:genus and !~taxon:genus:.*
```

app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/AddTreeLeafType.kt
```diff
@@ -0,0 +1,39 @@
...
+    override val wikiLink = "Key:leaf_type"
+    override val icon = R.drawable.ic_quest_leaf
+    override val achievements = listOf(OUTDOORS)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_leafType_tree_title
+
+    override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
+        getMapData().filter("""
+            nodes with natural=tree
+        """.toElementFilterExpression())
```
FloEdelmann, line 32:
```suggestion
        getMapData().filter("nodes with natural = tree".toElementFilterExpression())
```

westnordost (changes requested):
I believe this quest should be disabled by default because:

- 💤 it is very spammy: Vast majority of (street) trees are broadleaved plus they are very frequent for every place in which they are comprehensively mapped
- 💵 the information value is very low, there is little one could do with this information

---

Thanks for adding tests, it's a good practice overall! However, I think it does not really add value here to test a (simple) element filter expression like

```
nodes with natural = tree and !leaf_type and !~"(taxon|genus|species).*"
```

because

1. the element filter expression itself is much more expressive for documenting the intended behavior than the tests
2. there is no added logic beyond the element filter expression that is being tested

...weighted against the fact that all added code adds (ever so slightly) maintenance effort.

So, I'd suggest to remove the tests. If you are adamant about keeping them, please not only test some taggings randomly but make sure that the tests are complete, i.e. test every tag that should lead to the quest not being created.
Also, the readability should be improved, e.g.

```kotlin
    @Test fun `not applicable to trees with any taxon`() {
        assertFalse(isApplicableTo("natural" to "tree", "genus" to "Quercus"))
        assertFalse(isApplicableTo("natural" to "tree", "species:de" to "Stieleiche"))
        //... add all variations one wants to test
    }

    private fun isApplicableTo(vararg tags: Pair<String, String>): Boolean =
      questType.isApplicableTo(node(tags = mapOf(*tags)))
```

But, as I wrote, I think it would be better to let the element filter expression to speak for itself in this case.

matkoniecz:
> marked this pull request as ready for review

hmm, I tend to keep my PRs in draft state when there are known issues for me to resolve

matkoniecz:
> 💵 the information value is very low, there is little one could do with this information

what motivated me was that thing overlay show this trees differently - and to give people way of tagging detail without deleting/adding them (which is a bad idea for many reasons)

matkoniecz:
> Thanks for adding tests, it's a good practice overall! However, I think it does not really add value here to test a (simple) element filter expression like

in this case I created them because I was not sure whether regex works in keys and wanted to be sure that I have no silly mistakes - I had case before where I tested seemingly obvious regexp and tests were failing

I can remove them

app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/AddTreeLeafType.kt
```diff
@@ -0,0 +1,37 @@
...
+          natural = tree
+          and !leaf_type
+          and !~"(taxon|genus|species).*"
+    """
+    override val changesetComment = "Specify leaf types"
+    override val wikiLink = "Key:leaf_type"
+    override val icon = R.drawable.ic_quest_leaf
+    override val isDeleteElementEnabled = true
+    override val achievements = listOf(OUTDOORS)
+    override val defaultDisabledMessage = R.string.default_disabled_msg_difficult_and_time_consuming
```
westnordost, line 25: Hmm, this is not entirely correct, but we have no text for "it's spammy". So, I guess, it's fine.

westnordost (approved):

westnordost:
Looks good, ready to merge. (I'd say it should go into the next major release though, not the one currently in beta)

app/src/main/java/de/westnordost/streetcomplete/quests/leaf_detail/AddTreeLeafType.kt
matkoniecz: Actual message is rather `default_disabled_msg_difficult_or_time_consuming` ("or" not "and") and "time consuming" seems good match for spammy.

And I would not add extra message just for that

PR #5508, opened 2024-02-26
Rename ShopsOverlay → PlacesOverlay

FloEdelmann:
… to match other usages.

I've tested it by creating edits with the places overlay (without this PR's changes) and upgrading the debug app (with this PR's changes). The edits were still there and nothing crashed.

Only the currently selected overlay was cleared, as that is not saved in the database, but in the shared preferences. I think that's okay though.

Might there be other places which could cause a crash?


westnordost:
Thanks! I think that is okay. No additional places come to my mind.

westnordost (approved):

PR #5511, opened 2024-03-01
explain status of building quest

matkoniecz:


matkoniecz:
Not entirely sure about mentioning where "for cars" buildings went - I initially was confused by that and thought that I found a bug.

westnordost:
It's already mentioned in line 38.

matkoniecz:
hmmmmmmm

I guess that falls into "users will not actually read any documentation and messages"

PR #5514, opened 2024-03-01
Make question mark parsing code more readable

FloEdelmann:
See individual commits:

1. Add a test about this, so that the refactoring doesn't break anything.
2. Change string + large comment to a list of single-character strings with individual comments that are later joined back together.
3. Add another question mark character listed on the linked Wikipedia page. [Vai](https://en.wikipedia.org/wiki/Vai_syllabary) seems to be a language/script that is in use today, so it's probably worth adding.


westnordost:
Looks good to me

PR #5518, opened 2024-03-03
Use SoftwareKeyboardControllerCompat

Isira-Seneviratne:


westnordost:
Can you explain how this is an improvement? Overall, it is more code. Anything that `SoftwareKeyboardControllerCompat` does better than the current solution?

Isira-Seneviratne:
> Can you explain how this is an improvement? Overall, it is more code. Anything that `SoftwareKeyboardControllerCompat` does better than the current solution?

It calls some new APIs for opening the keyboard introduced in API level 30.

westnordost:
But how is that new API better?

Isira-Seneviratne:
The software keyboard controller API handles some edge cases for showing the keyboard: https://github.com/androidx/androidx/blob/155b4777a51e77095ca1609b1c514d37933d8828/core/core/src/main/java/androidx/core/view/SoftwareKeyboardControllerCompat.java#L97

westnordost:
Nice, that's good to hear.

app/src/main/java/de/westnordost/streetcomplete/util/ktx/Context.kt
```diff
@@ -51,16 +50,13 @@ fun Context.toast(@StringRes resId: Int, duration: Int = Toast.LENGTH_SHORT) {
 fun Context.hasPermission(permission: String): Boolean =
     ContextCompat.checkSelfPermission(this, permission) == PackageManager.PERMISSION_GRANTED
 
-fun View.showKeyboard(): Boolean? =
-    context?.inputMethodManager?.showSoftInput(this, SHOW_IMPLICIT)
+fun View.showKeyboard() = SoftwareKeyboardControllerCompat(this).show()
```
westnordost, line 53: Would you specify the return type for this function and hideKeyboard explicitly?

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/CreateNoteFragment.kt
```diff
@@ -140,7 +142,11 @@ class CreateNoteFragment : AbstractCreateNoteFragment() {
     override fun onComposedNote(text: String, imagePaths: List<String>) {
         /* pressing once on "OK" should first only close the keyboard, so that the user can review
            the position of the note he placed */
-        if (contentBinding.noteInput.hideKeyboard() == true) return
+        val rootInsets = ViewCompat.getRootWindowInsets(contentBinding.noteInput)
+        if (rootInsets?.isVisible(WindowInsetsCompat.Type.ime()) == true) {
```
westnordost, line 146: I think this should be a extension function, too then. `fun View.isKeyboardVisible(): Boolean`

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/util/ktx/Context.kt
Isira-Seneviratne: It doesn't return anything.

westnordost: Yes, so it should be `: Unit`

If not specified, the return type is set automatically to whatever the return type of the expression on the right is. For a reader without an IDE that can help him, it is then not obvious what the return type is.

app/src/main/java/de/westnordost/streetcomplete/screens/main/bottom_sheet/CreateNoteFragment.kt
```diff
@@ -14,6 +14,8 @@ import android.view.animation.BounceInterpolator
 import android.view.animation.TranslateAnimation
 import androidx.core.graphics.toPointF
 import androidx.core.os.bundleOf
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
```
FloEdelmann, line 18: Unused imports, right?

```suggestion
```

westnordost (approved):
Nice, looks good!

Issue #5412, opened 2023-12-20
Set up KMM Project Structure


westnordost:
Currently, the project setup is that of a native Android app. The task is to

1. change the project structure to that of a [Kotlin Multiplatform Mobile](https://kotlinlang.org/docs/multiplatform-discover-project.html) project
2. move all files that are already pure kotlin at the time this task is done (no Java or Android dependencies) into the common module, including its tests. (The implementation for iOS is obviously empty for now)
4. test build on Android

After this, work can start on a skeleton implementation of the iOS app.

This should ideally be done when there are as few PRs open as possible, as all PRs will result to have conflicts in every single file.

This should also ideally be done in concert with @Helium314 (the maintainer of the SCEE fork/addon), i.e. the fewer unmerged branches there are, the less work it is to rebase it on top / merge it with all the moved files.

HolgerJeromin:
git is able to track renames in rebases. But yes: It can not do magic. 
Rename files and change them (to match its new location) should be done in separate commits.

PR #5521, opened 2024-03-05
Switch from java.io.File API to okio

adrianclay:
Following on from [#5483](https://github.com/streetcomplete/StreetComplete/pull/5483#discussion_r1485675962)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -163,6 +163,10 @@ dependencies {
     implementation("io.ktor:ktor-client-cio:2.3.9")
     testImplementation("io.ktor:ktor-client-mock:2.3.9")
 
+    // File/Stream/IO API
+    implementation("com.squareup.okio:okio:3.8.0")
+    testImplementation("com.squareup.okio:okio-fakefilesystem:3.8.0")
+
     // finding in which country we are for country-specific logic
     implementation("de.westnordost:countryboundaries:2.1")
     // finding a name for a feature without a name tag
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/AvatarsDownloader.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/AvatarsDownloader.kt
@@ -7,17 +7,22 @@ import de.westnordost.streetcomplete.util.logs.Log
 import io.ktor.client.HttpClient
 import io.ktor.client.plugins.expectSuccess
 import io.ktor.client.request.get
-import io.ktor.client.statement.bodyAsChannel
-import io.ktor.util.cio.writeChannel
-import io.ktor.utils.io.copyAndClose
-import java.io.File
+import io.ktor.client.statement.readBytes
+import okio.FileSystem
+import okio.IOException
+import okio.Path
+
+interface AvatarStore {
+    fun cachedProfileImagePath(userId: Long): String?
+}
 
 /** Downloads and stores the OSM avatars of users */
 class AvatarsDownloader(
     private val httpClient: HttpClient,
     private val userApi: UserApi,
-    private val cacheDir: File
-) {
+    private val filesystem: FileSystem,
+    private val cacheDir: Path
+) : AvatarStore {
 
     suspend fun download(userIds: Collection<Long>) {
         if (!ensureCacheDirExists()) {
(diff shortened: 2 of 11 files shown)


app/build.gradle.kts
```diff
@@ -158,6 +158,10 @@ dependencies {
     implementation("io.ktor:ktor-client-cio:2.3.8")
     testImplementation("io.ktor:ktor-client-mock:2.3.8")
 
+    // File/Stream/IO API
+    implementation("com.squareup.okio:okio:3.8.0")
```
adrianclay, line 162: I notice we already have `implementation("org.jetbrains.kotlinx:kotlinx-io-core:0.3.0")` above. It appears that both these libraries have the same purpose, so perhaps I should use `kotlinx-io-core` instead?

westnordost: We use kotlinx-io-core for a certain functionality only, for reading a ByteArray. Otherwise, it is not used. Can kotlinx-io-core do everything okio can?

westnordost:
Awesome! :-D Let me know when you need feedback or I should review it

app/build.gradle.kts
westnordost: Anyway, since `kotlinx-io-core` is described experimental, I would like to avoid it at this time. If it was not experimental, I would probably slightly prefer it over okio because it is kind of official.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/NotesModule.kt
```diff
@@ -2,17 +2,18 @@ package de.westnordost.streetcomplete.data.osmnotes
...
 import org.koin.dsl.module
-import java.io.File
 
 val notesModule = module {
-    factory(named("AvatarsCacheDirectory")) { File(get<Context>().cacheDir, ApplicationConstants.AVATARS_CACHE_DIRECTORY) }
-    factory { AvatarsDownloader(get(), get(), get(named("AvatarsCacheDirectory"))) }
+    factory {
+        val avatarCacheDir = get<Context>().cacheDir.toOkioPath().resolve(ApplicationConstants.AVATARS_CACHE_DIRECTORY)
+        AvatarsDownloader(get(), get(), get(), avatarCacheDir)
+    }
```
westnordost, line 12: It seems unclean to me to remove the `AvatarsCacheDirectory` as injectable dependency. (As you likely noticed,) this leads to application components that have nothing to do with download of avatar images (and should not be able to trigger an avatar download) to depend on `AvatarDownloader`.

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/AvatarsDownloader.kt
```diff
@@ -47,20 +49,37 @@ class AvatarsDownloader(
...
-        val avatarFile = File(cacheDir, "$userId")
+        val avatarPath = avatarPath(userId)
         try {
             val response = httpClient.get(avatarUrl) {
                 expectSuccess = true
             }
-            response.bodyAsChannel().copyAndClose(avatarFile.writeChannel())
-            Log.d(TAG, "Downloaded file: ${avatarFile.path}")
+            filesystem.write(avatarPath) {
+                response.bodyAsChannel().copyTo(this.outputStream())
```
westnordost, line 58: `outputStream` and `copyTo` is a Java API.

`this` is a `okio.BufferedSink` but the response provides a `ByteReadChannel`. It is possible to write a `ByteReadChannel::copyTo(out: okio.BufferedSink)` oneself in case no adapter library for that exists, it is indeed not that much. However, compared with photo uploads, OSM user avatars are quite small, so in this case it would be 100% fine to just convert to ByteArray and stuff that ByteArray into the okio sink. (But maybe not that consistent if conversion in one direction is implemented but not the other - optimum anyway would be if there was an official library for that)

Soem details in the other comment.

westnordost (changes requested):
I see you inadvertently used some Java APIs. I guess that's an argument to do #5412 (move all stuff that is pure kotlin / multiplatform into "common" module) rather sooner than later.

Thanks for the tests, too!

Anyway, comments below.

westnordost:
I plan to do a release in the next days but would (like to) wait with that if you think you can get it done this week. There is of course no particular hurry, it's just that I would like to check off some obvious tasks before I do #5412 (=move all the multiplatform code into a common module)

adrianclay:
Hey @westnordost I had taken a look at your comments, the one which is taking the most time is the `ImageUploader`. More research/work required I'm afraid, I hope to have some time on Sunday.

westnordost:
Cool, alright! Then I'll release the update before, so that there is no rush.

westnordost:
Reviewing this, I changed my opinion about kotlinx-io. 

Sure, it is still experimental, but this just means that the API may change, which will not likely produce a lot of work.

Also, as you noted, we are already using kotlinx-io in the project, so I find it is a good idea to keep the number of (direct) dependencies low (and the number of dependences that do the same thing to 1).

As noted in my previous review, using okio together with ktor will necessitate writing wrappers between the data structures of these two ourselves. With kotlinx-io I am not sure, but the likeliness that something will be done there is much higher (as okio❤okhttp, not ktor).

I think using your PR as a source, I will try to do this in the second part of this week.

westnordost:
Ktor also migrates to kotlinx-io this year: https://kotlined.com/blog/the-ktor-roadmap-for-2024/

westnordost:
and this is in progress right now, see https://youtrack.jetbrains.com/issue/KTOR-6030/Migrate-to-new-kotlinx.io-library

I started with this in a separate branch, but I think it only makes sense to continue with it once the changed API has been released at least as an alpha.

One thing I noticed that is missing for kotlinx.io.files is a FakeFileSystem. So, tests would either need to use the host file system and then clear the created files after the test or a FileSystem must be mocked (which I am not sure right now if it is possible because currently `FileSystem` is a sealed interface).

In any case, I will close this PR, the little work so far (mostly the same, minus tests, what you did) I've done is in the `kotlinx-io-files` branch.

PR #5524, opened 2024-03-06
Use TypedValueCompat

Isira-Seneviratne:
Simplify pixel conversion code using `TypedValueCompat` methods.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/util/ktx/Context.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/util/ktx/Context.kt
@@ -14,20 +14,10 @@ import androidx.core.content.ContextCompat
 import androidx.core.content.getSystemService
 import androidx.core.location.LocationManagerCompat
 import androidx.core.net.toUri
-import androidx.core.util.TypedValueCompat
 import de.westnordost.streetcomplete.ApplicationConstants
 import de.westnordost.streetcomplete.R
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlin.coroutines.resume
-
-/** return the number of pixels for the given scalable pixels */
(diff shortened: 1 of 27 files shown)


FloEdelmann:
Could you check whether the conversion functions in `ChangelogFragment` can also be simplified using these extension functions?

westnordost (approved):
LGTM

app/src/main/java/de/westnordost/streetcomplete/util/ktx/Context.kt
```diff
@@ -14,28 +14,20 @@ import androidx.core.content.ContextCompat
...
 /** return the number of density independent pixels for the given pixels */
-fun Context.pxToDp(px: Int): Float = px / resources.displayMetrics.density
+fun Context.pxToDp(px: Int): Float = resources.pxToDp(px)
 
 /** return the number of pixels for the given density independent pixels */
-fun Context.dpToPx(dp: Int): Float = dp * resources.displayMetrics.density
-
-/** return the number of pixels for the given scalable pixels */
-fun Context.spToPx(sp: Int): Float = sp * resources.displayMetrics.scaledDensity
+fun Context.dpToPx(dp: Int): Float = resources.dpToPx(dp)
```
FloEdelmann, line 30: For consistency: Should we maybe move all those px/dp/sp-related extension functions from `Context` to `Resources` and use it there directly?

FloEdelmann (approved):
Looks good to me now. Thanks for incorporating all the feedback! Note that I haven't checked out the code locally, so it's only a code review.

westnordost:
LGTM too. (Also before.)

Note that having shortcuts like `fun Context.a() = resources.a()` is completely fine by me. But I don't mind so much either way, especially since the whole Android view stuff is due to be replaced with Jetpack compose. (Currently looking into setting it up.)

PR #5533, opened 2024-03-12
covered by building answer in AddIsAmenityIndoor

matkoniecz:


matkoniecz:
I was quite confused myself how to answer this.

"just leave note" is an option, but for quest specifically about being inside in building having a direct answer would be nice

IMG_20240307_063203

app/src/main/res/values/strings.xml
```diff
@@ -1747,5 +1747,6 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="label_log_older_than">To</string>
     <string name="label_log_message_contains">Message contains</string>
     <string name="title_logs_filters">Filter messages</string>
+    <string name="quest_isAmenityIndoor_outside_covered">Covered by building but not inside</string>
```
westnordost, line 1750: Seems a bit long. 
```suggestion
    <string name="quest_isAmenityIndoor_outside_covered">Not inside, but covered</string>
```
Is this better?

matkoniecz: I guess it can be `covered=yes` not only by building 

though it loses some info, but not so important info I guess

westnordost (approved):

Issue #5447, opened 2024-01-17
Set up Jetpack Compose


westnordost:
Other than adding the dependency, the (dark+light) theme, style, colors, fonts are defined in code in Jetpack Compose rather than in `res/values/*.xml` files.

So, as a start, the current theme etc should be duplicated there. (See any sample Compose app for an example).

The app uses a theme based on Material 2. If a migration to Material You (=3) should happen, the best time would be to do this before this ticket, as the theme definition also for compose looks a bit different between Material 2 and 3.

I've tried it before, but ended up not doing it, see #3650. (I have no really strong opinion on not using Material You - if someone else did the migration, I'd be okay with it)

westnordost:
#5607

Issue #5421, opened 2023-12-20
iOS - Planning


westnordost:
This ticket is something like the master ticket to coordinate development on an iOS port of StreetComplete. It replaces #1892 which also included a lot of discussion and research / observation work.

### TLDR: Have a look at the [Project Board](https://github.com/orgs/streetcomplete/projects/1) for a list of tasks. Contributions are welcome!

# Approach

The code base of this app is written in 100% Kotlin, a modern programming language that is quite similar to Swift. Kotlin code can be transpiled to JavaScript and also to machine code, just like Swift. Furthermore, many dependencies of this app by now are pure Kotlin libraries.

<img align="left" width="64" src="https://github.com/streetcomplete/StreetComplete/assets/4661658/b1c25b4c-2f60-463e-a587-12793db95961">

An iOS version of this app hence can be done with [Kotlin Multiplatform](https://www.jetbrains.com/kotlin-multiplatform/). 

The UI code can be shared using [Compose Multiplatform](https://www.jetbrains.com/lp/compose-multiplatform/), which is currently in alpha (= usable, but API may change). It is a fork of [Jetpack Compose](https://developer.android.com/jetpack/compose) with largely the same API. Jetpack Compose is a reactive UI framework in which developers define the UI completely in code (similar idea as SwiftUI). We'll have to re-create the entire UI (incrementally) in Compose, though.

This will keep the amount of code that needs to be platform dependent (Android / iOS) to a minimum.

This approach has the big advantage over others in that further maintenance of this project will not increase significantly, as there will continue to be just one code base.

# Steps in a nutshell

- Separate all platform specific code from application logic. Replace Android/Java dependencies for application logic with Kotlin multiplatform dependencies

- Incrementally migrate the UI code to Compose Multiplatform:

  1. <img align="right" width="64" src="https://github.com/streetcomplete/StreetComplete/assets/4661658/6f8ff785-140f-4360-9087-e811c71f7219">Migrate the UI code that currently uses Android XML layouts to Android [Jetpack Compose](https://developer.android.com/jetpack/compose). This can be done incrementally when done bottom-up (see [migration strategy guide](https://developer.android.com/jetpack/compose/migrate/strategy))

[...]

PR #5535, opened 2024-03-12
Handle LocationStateButton saved state using Bundle

Isira-Seneviratne:


westnordost (commented):
Why is this better?

Did you test this?

Isira-Seneviratne:
> Why is this better?

It removes the need for an extra class to store the additional state.

> Did you test this?

Yes, on an Android 5.0 emulator. There were no issues.

westnordost:
Hm, well, it's not less code at least. And it is a different API that I know, so to understand this code I'd need to look into that, too.

So, from my point of view, it is not an improvement because it replaces something that works with something that is mostly just different. I understand that if other custom views in this project were to save their instance state in that "old way" too, this API or even the Parcelize plugin may be an improvement, but I think there are on other custom views that do that, right?

I really do appreciate that you groom and modernize the codebase with your contributions. Did you read #5421, though? We plan to migrate the Android view code [step by step](https://developer.android.com/jetpack/compose/migrate/strategy) to Jetpack Compose. So, any improvements on the current Android view code will be somewhat moot or short-lived. I don't mind merging it if the improvement is apparent, but this is the reason why I am somewhat reluctant to concern myself with it to understand it.

And maybe it is also not worth your time, if you consider how relatively short lived your improvements will be. Now, on the Jetpack Compose stuff, if you want to and can help there, your work would really make a difference.
Compose is [somewhat of a paradigm shift](https://developer.android.com/jetpack/compose/mental-model), so if you wanted to help and don't know it yet, it might take some time to get acquainted with it. On the plus side, the whole industry is shifting to this kind of declarative UI paradigm, so it's a good idea to learn it too, anyway.

westnordost:
Oh by the way, I noticed you are a contributor to NewPipe! Cool stuff, thank you! Love it!

Isira-Seneviratne:
Yeah, I think making contributions in Compose would be better. I've gone through the tutorials, but hands-on experience would be very useful.

westnordost:
Cool! I am also going through the documentation currently... it is quite.. well, as I wrote, paradigm shift... it's a lot. I am also having a cold right now, so that may contribute towards why I am a bit overwhelmed with it. I started with setting it up (#5447) though I am not sure if it wouldn't be better (=cleaner, more efficient) to first migrate to ViewModels everywhere before.

On the other hand, if I want to encourage people contributing to this, I should enable it as soon as possible. And contributions by contributors are vital to the whole undertaking. So, I'll probably set it up end of this week.

Anyway. I've looked through the PR again and I have to admit that it is actually better readable even if it is not shorter. But well, just wanted to let you know that (future) improvements on Android view stuff will be quite short-lived.

PR #5558, opened 2024-03-29
Don't highlight roads in cycleway overlay where cycling is not allowed

burrscurr:
The cycleway quest already ignores roads with bicycle=no, but the cycleway overlay did not.

(Since this is a very minor change/fix, I directly created a PR.)


westnordost:
Right, thanks!

PR #5559, opened 2024-03-29
Don't highlight tunnels in parking overlay

burrscurr:
Parking in tunnels is very rare (can't think of any cases where parking in a road tunnel would be allowed, mainly for safety reasons). The parking overlay therefore should not highlight tunnels that don't have information on parking to avoid spam.

(Since this is a very minor change, I directly created a PR.)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/street_parking/StreetParkingOverlay.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/street_parking/StreetParkingOverlay.kt
@@ -81,6 +81,7 @@ private val streetParkingTaggingNotExpected by lazy { """
       highway ~ service|pedestrian
       or motorroad = yes
       or expressway = yes
+      or tunnel = yes
       or junction = roundabout
       or ~"${(MAXSPEED_TYPE_KEYS + "maxspeed").joinToString("|")}" ~ ".*:(rural|nsl_single|nsl_dual)"
       or maxspeed >= 70


westnordost:
Makes sense, thanks!

HolgerJeromin:
What about building_passages?

mnalis:
> What about building_passages?

Good point, I think whole `tunnel` key should be excluded, not just `tunnel = yes` (as [any more specifically mapped tunnel](https://taginfo.openstreetmap.org/keys/tunnel#values) is unlikely to have parking spots)

burrscurr:
I also think `tunnel=building_passage` is unlikely to have parking. However, given that building passages are typically overground and very short, not showing such sections in the overlay may be more confusing rather than just displaying the parking status. (I'm thinking of building passages like [this one](https://www.mapillary.com/app/?lat=51.035019904375&lng=13.719303645194&z=18.180647134821317&pKey=757244204981518&focus=photo&x=0.48881127613454745&y=0.5801640649081056&zoom=0), where one might think that the parking along the street continues under the building. Clarifying the parking status makes sense there, I'd say.) Regarding other `tunnel` values, maybe `avalanche_protector` could be excluded as well (very expensive structures, likely no parking). Most other common `tunnel` values (`culvert`, `flooded` etc.) don't apply to road tunnels, so excluding `tunnel=*` would seem a little to broad to me.

matkoniecz:
> excluding tunnel=*

At most excluding `tunnel=* and tunnel!=no`

PR #5567, opened 2024-04-01
New Quest: (Belgium) Moped access to cycleway information

PowerPlop:
Proof of concept implementation for https://github.com/streetcomplete/StreetComplete/issues/5565

2df9b92e-e940-4057-8c9d-585d44af8f14

It's my first pull request ever, so be gentle. :)
If fixing this pull request would require too much work, you may close it.

### Todo

- [ ] Translations
- [x] Icon
- [x] Decide on form
- [x] Extend elementfilter to have less options
- [x] Improve button formatting (3rd option 'mandatory' is shown on new line which is strange)

### Questions

- [x] Is it too Belgium specific? Is it allowed when filtering it in the element list on countries? If not, can we make it more general?
- [x] Is it spam? Every cycleway has to be tagged, you could filter by limiting on speed and regions, but is this okay?
       How can I easily test the elementFilter (only by using overpass turbo?)
- [x] Should I use a simple yes/no form or is it okay to create a specific form?
- [x] Should I include the signs in the form? Or should I make a `AImageListQuestForm` ? What is the best way to integrate sigs in a quest? Or better to keep it yes/no and assume the user can make the decision without photo?
- [x] Do I need to make a distinction between Designated and Yes? Should I use Forbidden instead of No as answer option?
- [x] Is the quest name okay? Should I use `ProhibitedForMoped `or better `MopedAccessAllowed`?

### Tests

- [x] Is the quest listed?
- [x] Look at the quest list in settings - does your quest appear there? If not - see [this step](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#add-the-quest-to-the-list-of-active-ones).
- [x] Is it shown for expected elements?
- [x] Can you tap on a quest marker to open the quest form?
- [x] Can you fill in the answer as expected?
- [x] Can you solve the quest?
- [] Is expected tagging being applied?
-> How to check?
- [x] Does the quest disappear after being solved?


PowerPlop:
@westnordost  I'm happy to receive some feedback for this quest!
Please let me know if this would be approved (previous quest suggestion was too broad), and give me some feedback on how to make the quest more consistent with other quests. (how to handle sign information, more generic, simple yes/no etc).

westnordost:
From a cursory look, most of it looks fine. A few remarks:
- the enum name is "prohibited"-something, yet, the "Yes"  option means that it is not prohibited. That's confusing.
- what about highway=path with bicyle=yes etc? A maxspeed < 70 on a highway=cycleway??

But before continuing here, let's first clear up the stuff in the linked issue ticket.

PowerPlop:
@westnordost I processed the feedback of https://github.com/streetcomplete/StreetComplete/issues/5565.

* No external forms, but simple radio buttons
* Removed signs and custom form
* Dutch translations
* changed answers to include sign visibility
* disabled by default
* changed filter

Todo:

* icon
* optimize element filter (there are a lot of cycleways, so I would like to make a useful selection?)

app/src/main/res/values-nl/strings.xml
```diff
@@ -1165,6 +1165,10 @@ Helaas mogen er vanwege een Google Play Store-beleid geen links naar donaties in
     <string name="quest_memorialType_stele_wooden">"Houten zuil (gedenkplaat)"</string>
     <string name="quest_memorialType_stele_stone">"Stenen zuil (gedenkplaat)"</string>
     <string name="quest_memorialType_sculpture">"Ander beeldhouwwerk"</string>
+    <string name="quest_moped_prohibited_title">Zijn bromfietsen (klasse b) toegelaten op dit fietspad?</string>
+    <string name="quest_moped_prohibited_designated">Ja, een gebodsbord is gekoppeld aan het "fietspad" bord.</string>
+    <string name="quest_moped_prohibited_allowed">Ja, maar er is geen verbodsbord zichtbaar bij het fietspad.</string>
+    <string name="quest_moped_prohibited_forbidden">Nee, er is een verbodsbord gekoppeld aan het "fietspad" bord.</string>
```
westnordost, line 1171: Any translated strings will be overwritten on each import from POEditor. So, this should rather be done after merge on POEditor.

app/src/main/res/values/strings.xml
```diff
@@ -1151,6 +1151,14 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_memorialType_stele_stone">Stone stele (memorial slab)</string>
     <string name="quest_memorialType_sculpture">Other sculpture</string>
 
+    <string name="quest_moped_prohibited_title">Are mopeds allowed on this cycleway?</string>
+    <string name="quest_moped_prohibited_designated">Yes, a mandatory sign is attached to the cycleway sign.</string>
+    <string name="quest_moped_prohibited_allowed">Yes,there is no prohibition sign is visible.</string>
+    <string name="quest_moped_prohibited_forbidden">No, a prohibition sign is attached to the cycleway sign.</string>
```
westnordost, line 1157: "cycleway" is British English. The default strings are in American English. Use "bike path" instead.

As described earlier, I would find it better if it just asked for a sign because that doesn't presuppose any expert knowledge, i.e. in this case the legal situation in Belgium. Yes, I see that you worded the "allowed" answer in a way that lectures the user, but in my opinion, it is unnecessary to explain the legal situation to the user, it just makes the text longer. Speaking of which;

I find that the text options are too long. It's key to accommodate the keywords that also appear on the signs in the answer options, but in your current version, e.g. the designated and prohibited options have 10 words but differ in only one word. How about something like... 

**Is there a sign concerning mopeds for this bike path?**
- No, there is no sign
- Yes, mopeds are prohibited here
- Yes, this side path is mandatory for mopeds

(this is just a draft / suggestion)

app/src/main/java/de/westnordost/streetcomplete/quests/moped/AddProhibitedForMoped.kt
```diff
@@ -0,0 +1,49 @@
...
+        tags: Tags,
+        geometry: ElementGeometry,
+        timestampEdited: Long,
+    ) {
+        when (answer) {
+            ProhibitedForMopedAnswer.ALLOWED -> tags["moped"] = "yes"
+            ProhibitedForMopedAnswer.FORBIDDEN -> tags["moped"] = "no"
+            ProhibitedForMopedAnswer.DESIGNATED -> tags["moped"] = "designated"
+
+        }
```
westnordost, line 47:
```suggestion
        tags["moped"] = when (answer) {
            ProhibitedForMopedAnswer.ALLOWED -> "yes"
            ProhibitedForMopedAnswer.FORBIDDEN -> "no"
            ProhibitedForMopedAnswer.DESIGNATED -> "designated"
        }
```

Can also be written like this.

app/src/main/java/de/westnordost/streetcomplete/quests/moped/AddProhibitedForMoped.kt
```diff
@@ -0,0 +1,49 @@
...
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddProhibitedForMoped : OsmFilterQuestType<ProhibitedForMopedAnswer>() {
+
+    override val elementFilter =
+        // filter all cycleways next to a road with a speed of more than 50, as Moped use is designated by law
+        """
+        ways with (
+            highway = cycleway
+            or (cycleway and bicycle ~ yes|designated and (maxspeed <= 30 or !maxspeed))
```
westnordost, line 17: 1. What about `highway = path` (or `footway`) + `bicycle = designated`? A bike path can also be mapped like this.
2. What's that `cycleway` doing there? It looks wrong.

westnordost (changes requested):
> Icon

What's the reason why you want this data in OSM, by the way? Does the prohibition of mopeds indicate a better, or worse, bike path? Do they have a nickname? Are bromfiets very loud? Maybe this could be an inspiration for a motif. Or do they often or always (in Belgium) have a certain feature that distinguishes them well from bicycles even from afar?

westnordost:
Maybe something like this, on the other hand, I think even simplified it may be too complex for an icon.

OIG3

westnordost:
Or these kind of moped helmets, don't know how they are called, but they look distinct from motorbike helmets and also from bicycle helmets.

app/src/main/java/de/westnordost/streetcomplete/quests/moped/AddProhibitedForMoped.kt
```diff
@@ -0,0 +1,49 @@
+package de.westnordost.streetcomplete.quests.moped
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.NoCountriesExcept
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddProhibitedForMoped : OsmFilterQuestType<ProhibitedForMopedAnswer>() {
```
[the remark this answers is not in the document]
PowerPlop: Fine by me, i'll change it!

app/src/main/java/de/westnordost/streetcomplete/quests/moped/AddProhibitedForMoped.kt
PowerPlop: 1. A moped is only allowed on a bike path. If it is shared with pedestrians, it isn't allowed
2. If I understand correctly, there are 2 types of bike path tags. https://wiki.openstreetmap.org/wiki/Key:cycleway 
Highway = cycleway -> separate way/road for bikes
cycleway = * (e.g. lane) -> part of road to be used by bikes

Therefore I added the 'cycleway'  tag on roads with a speed limit of 50.
Maybe I can remove this filter since in this case it doesn't matter whether you're riding on the cycleway or a road next to it?

westnordost: > A moped is only allowed on a bike path. If it is shared with pedestrians, it isn't allowed

Segregated foot and bike paths can be tagged like `highway=path` + `segregated=yes` + `foot=designated` + `bicycle=designated` or also `highway=footway` + `segregated=yes` + `bicycle=designated` or also `highway=cycleway` + `segregated=yes` + `foot=designated`.
Conversely, `highway=cycleway` alone does not guarantee that it is a path that is segregated from pedestrians, only if `segregated=yes` is set it is.

> cycleway = * (e.g. lane) -> part of road to be used by bikes

Yes, but it is much more complicated than that. First of all, there is also `cycleway:both`, `cycleway:left` and `cycleway:right`. Second, the situation can differ per side. So, this would require a different interface that lets you select each side individually. Plus, you would have to handle situations where it is defined for one side, but not for the other, etc.
This would be a lot of effort, I recommend just not asking for cycleways that are mapped as properties on roads. It looks like most cycleways in Belgium are mapped separately anyway.

westnordost:
> I could use the current bicycle and motorbike icon and put them next to each other?

that could work, probably there is only space on top of each other, though which then would look very similar to the capacity quest icon. I guess, just a pictogram of a moped would be fine, but a pictogram of a moped might be difficult to distinguish from a bicycle?

PowerPlop:
I'll try to push the changes next weekend!

PowerPlop:
> > I could use the current bicycle and motorbike icon and put them next to each other?
> 
> that could work, probably there is only space on top of each other, though, which then would look very similar to the capacity quest icon. I guess, just a pictogram of a moped would be fine, but a pictogram of a moped might be difficult to distinguish from a bicycle?

What do you think of this one?

image

I like it because it keeps the cycleway icon and adds the existing forbidden sign format. But maybe it's too much info?

westnordost:
It does look nice! However, you are correct that it is too complex for the size the pins are usually displayed. Also, is this actually a moped? Looks like a scooter. I thought mopeds always had, well, pedals.

What is also slightly problematic is that the quest asks first of all for a sign. People who don't read (i.e. most people) might then think the question is about whether they are forbidden. But since the answer options are not just _yes_, _no_, _mandatory_ but are half a sentence, this is only slightly problematic.

app/src/main/res/values/strings.xml
```diff
@@ -1151,6 +1151,14 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_memorialType_stele_stone">Stone stele (memorial slab)</string>
     <string name="quest_memorialType_sculpture">Other sculpture</string>
 
+    <string name="quest_moped_access_title">Is there a sign indicating access for mopeds on this bike path?</string>
```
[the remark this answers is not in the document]
westnordost: Hm, not sure, if "yes" and "no" are removed, it may feel like the answer options are not directly answers to the question asked?

PowerPlop:
> It does look nice! However, you are correct that it is too complex for the size the pins are usually displayed. Also, is this actually a moped? Looks like a scooter. I thought mopeds always had, well, pedals.
> 

If I look at the OSM tags, I would say 'mofa' is a scooter with pedals, while 'moped' is a regular scooter (but limited in speed).

> What is also slightly problematic is that the quest asks first of all for a sign. People who don't read (i.e. most people) might then think the question is about whether they are forbidden. But since the answer options are not just _yes_, _no_, _mandatory_ but are half a sentence, this is only slightly problematic.

You're correct, the icon better matches the original quest 'are mopeds allowed/forbidden on this bike path'.
I can redesign it so it resembles a traffic sign on a pole?

westnordost:
> You're correct, the icon better matches the original quest 'are mopeds allowed/forbidden on this bike path'.
> I can redesign it so it resembles a traffic sign on a pole?

Hm, I would suggest either:
- the 🚫🛵  only on top of the cycleway (i.e. as now but without the bicycle icon)
- only the 🛵 on top of the cycleway, i.e. same as the cycleway icon but with the moped on top

PowerPlop:
> > You're correct, the icon better matches the original quest 'are mopeds allowed/forbidden on this bike path'.
> > I can redesign it so it resembles a traffic sign on a pole?
> 
> Hm, I would suggest either:
> 
> * the 🚫🛵  only on top of the cycleway (i.e. as now but without the bicycle icon)
> * only the 🛵 on top of the cycleway, i.e. same as the cycleway icon but with the moped on top

2 options:

image
image

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
PowerPlop: I agree, an other option would be to rephrase the question to 'what (kind of) sign is indicating access for mopeds on this bikepath' ?

westnordost:
For me, it's fine, there is very little time I need to put into to review this.

Anyway, as I wrote, the one above would be too complex, the lower one is good, though you should center the scooter icon. Also, in general, the shadow is I think always exactly 4px straight down from any shape in those quest icons.

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
westnordost: Hm, I think this is the best: "Is there a sign indicating access for mopeds on this bike path?" with answers

- There is no sign
- There is a sign that prohibits mopeds
- There is a sign that designates this bike path for mopeds

westnordost:
Note that the final icon should be put into the /res/graphics/quest directory, too. In res/authors.txt you can put the author and license of that icon.

PowerPlop:
> Note that the final icon should be put into the /res/graphics/quest directory, too. In res/authors.txt you can put the author and license of that icon.

I used an icon of svgrepo with 'Public Domain' license, so I assume I don't have to add anything?

westnordost:
> I used an icon of svgrepo with 'Public Domain' license, so I assume I don't have to add anything?

It's still useful to give the source, to have proof that it is indeed public domain.

app/src/main/java/de/westnordost/streetcomplete/quests/moped/AddMopedAccess.kt
```diff
@@ -0,0 +1,48 @@
...
+        ways with (
+            highway = cycleway
+        )
+        and !moped
+        and (motor_vehicle != no or !motor_vehicle)
+        """
+    override val enabledInCountries = NoCountriesExcept("BE")
+    override val defaultDisabledMessage = R.string.default_disabled_msg_visible_sign_moped
+
+
```
westnordost, line 25:
```suggestion
    override val elementFilter = """
        ways with (
            highway = cycleway
            or highway = path and bicycle = designated
            or highway = footway and bicycle = designated
        )
        and !moped
        and (motor_vehicle != no or !motor_vehicle)
    """
    override val enabledInCountries = NoCountriesExcept("BE")
    override val defaultDisabledMessage = R.string.default_disabled_msg_visible_sign_moped
```

Separately mapped cycleways can also be mapped like `highway = path` + `bicycle = designated` etc.

Removed the comment, because it doesn't make sense. We don't assume this here in the quest.

PowerPlop:
> Not merging yet, see my previous comment.

Added author!

westnordost (approved):

westnordost:
And added strings to POEditor. You could add your translations there.

westnordost:
Congratulations on your first quest! 🥳

PowerPlop:
> Congratulations on your first quest! 🥳

Thanks for the patience, the help and building such a beautiful app!

Issue #4913, opened 2023-03-29
Bicycle Overlay: Distinguish between "signed bicycle allowed" and "not designated for cyclists"


westnordost:
Currently, the overlay only displays and allows to specify a foot-/cycle-/pathway as one of (tags in (...) are examples, more tags match the given categories):

- no cycling allowed (`bicycle=no`)
- path or trail, i.e. not designated for anyone in particular (`highway=path`)
- not designated for cyclists (`highway=footway`)
- shared-use path, i.e. designated to both pedestrians and cyclists, not segregated (`highway=cycleway + foot=designated` etc...)
- separated bike- and foot path (`highway=cycleway + foot=designated + segregated=yes` etc...)
- bike-only path (`highway=cycleway`)

The obvious omission here is "Cyclists are explicitly allowed on footway" (`highway=footway + bicycle=yes`).

There has been past discussions about this, the reason why it is not included is that it is not clear from these tags whether this is signed as such or the mapper was of the opinion that bikes are OK too. In the case that it isn't signed, it is not surveyable. In the worst case, StreetComplete users would think this tagging is wrong and must be corrected to "not designated for cyclists", removing the `bicycle=yes` when it indeed might be correct, due to some legislation or so that isn't signed.

The suggestion now is to use `bicycle:signed=yes` to denote that the access restriction (`bicycle=yes`) is indeed explicitly signed and thus surveyable.
`bicycle:signed=*` occured in related m-m-monster discussions in the forum several times (see https://community.openstreetmap.org/search?q=bicycle%3Asigned ) and as far as I remember, there was not really any opposition to it, only remarks about that it does not solve all the issues (discussed in that thread), e.g. that in Germany, if "bicycles allowed on footway" is signed, it means that bicycles must go at most at walking-speed.

I do no plan to implement this soon, I first wanted to drop this concrete suggestion here for comments.

melbournefan:
In Australia (at least AU-VIC) 'Shared Paths' which are usually non segregated and intended for bicycles and foot traffic are signed at the beginning of the path and at the end of the path however they can have markings on the surface of the path, this is more prevalent now that Victoria is now making 'e-scooters' legal to use. Bicycles aren't allowed on footpaths `(footway=sidewalk)` unless the rider is under 12 years of age (though this varies in other states).

[...]

PR #5575, opened 2024-04-06
Add "None, but a sign allows cycling on sidewalk" as an option for cycleway overlay/quest

wielandb:
This PR adds the possibility to tag that cycling is allowed on the sidewalk by a sign in the cycleway quest and overlay, and that a sign allows cycling on a footway in the separate cycleway overlay.

|  |  |
|:---:|:---:|
| photo_2024-05-02_11-09-02 | photo_2024-05-02_11-09-22 |

- Cycleway quest and overlay
  - The StreetSideSelect Form now offers an option to tag "None, but a sign allows cycling on sidewalk" which will tag `sidewalk:[side]:bicycle=yes`
  - The colour in the overlay for SIDEWALK_EXPLICIT has been changed to be a continious cyan line, so that SIDEWALK_OK can have the dashed cyan line
- Seperate Cycleway Form
  - It's now possible to tag a seperate way as explicitly allowed for bicycles by a sign


westnordost:
Linking to #4913

wielandb:
Linking https://github.com/streetcomplete/StreetComplete/issues/5596#issuecomment-2080839826, which talks about a change on how `highway=path` + `foot=no` should be parsed.

westnordost:
Hey Wieland, are you still working on your streetcompleteness thing? If yes, could you make it a heatmap of quest densities? That would be so cool! 😁

wielandb:
Would apprechiate some feedback on the current state. 

The features work as intended as far as I can see. I'm new to doing unit tests, I implemented some, but I'm unsure if they are enough/correct. 

I am also still a bit unsure about tagging, especially in which cases `bicycle=` should be removed. Should choosing PATH remove `bicycle=`? Should choosing one of the options that indicate a cycleway remove `bicycle:signed=`, or maybe even add it?

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/Cycleway.kt
```diff
@@ -180,16 +180,17 @@ fun getSelectableCycleways(
...
     val dir = direction?.takeUnless { it == BOTH } ?: Direction.getDefault(isRightSide, isLeftHandTraffic)
     val cycleways = mutableListOf(
-        NONE,
         TRACK,
         EXCLUSIVE_LANE,
         ADVISORY_LANE,
         UNSPECIFIED_LANE,
         SUGGESTION_LANE,
         SEPARATE,
+        NONE,
```
westnordost, line 189: NONE should continue to be on top.

app/src/main/res/drawable/ic_cycleway_sidewalk_ok.xml
```diff
@@ -0,0 +1,53 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
westnordost, line 1: new graphics should be committed also to /res/graphics/ as SVGs

x3

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway_separate/SeparateCyclewayItem.kt
```diff
@@ -20,7 +20,7 @@ private val SeparateCycleway.titleResId: Int get() = when (this) {
 
 private fun SeparateCycleway.getIconResId(isLeftHandTraffic: Boolean): Int = when (this) {
     PATH ->               R.drawable.ic_separate_cycleway_path
-    NOT_ALLOWED ->        R.drawable.ic_separate_cycleway_no
+    NOT_ALLOWED ->        R.drawable.ic_separate_cycleway_disallowed
```
westnordost, line 23: Strings that are not used anymore should be deleted from strings.xml

x2

westnordost (changes requested):
I just had a quick look first, this is not a complete review.

> I am also still a bit unsure about tagging, especially in which cases bicycle= should be removed. Should choosing PATH remove bicycle=? Should choosing one of the options that indicate a cycleway remove bicycle:signed=, or maybe even add it?

I think such things should be changed if keeping the current value would contradict the newly selected option. If it was e.g. `bicycle:signed=no` before but user selects "exclusive cycleway" now, then `bicycle:signed=no` should be removed.

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway_separate/SeparateCyclewayItem.kt
wielandb: Ran a check, found and removed `R.string.separate_cycleway_allowed`, but not the second one.

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt
```diff
@@ -94,6 +94,7 @@ private fun parseCyclewayForSide(
     val cycleway = tags[cyclewayKey]
     val cyclewayLane = tags["$cyclewayKey:lane"]
     val isSegregated = tags["$cyclewayKey:segregated"] != "no"
+    val isCyclingOkOnSidewalk = tags["sidewalk$sideVal:bicycle"] == "yes" || tags["sidewalk:both:bicycle"] == "yes"
```
westnordost, line 97: `sidewalk:<side>:bicycle=designated` has also some use. But what does it mean?

Probably the same as `cycleway:<side>=track` + `cycleway:<side>:segregated=no`  ()?

Or just the same as `sidewalk:<side>:bicycle=yes` ()? Some mappers consider `designated` to mean "it is signed that way" rather than "it is primarily intended for".

**Four options how to deal with this tagging:**
1.  (**_currently_**:) ignore that value -> interpret it as `NONE`
2. consider this as `SIDEWALK_OK` ()
3. consider this as `SIDEWALK_EXPLICIT`. ()
4. consider this as `INVALID` due to the above ambiguity

I... tend towards considering it invalid, i.e. force users to choose one definite option, due to it being entirely undocumented and ambiguous. I.e. in all places where it is tagged like that, data consumers (routers, editors, visualizers, ...) will have the above options to deal with this, resulting in somewhat inconsistent behavior how this data is treated and this is certainly undesirable.

Tagging @tordans as the main developer of the most detailed cycling map and analysis tool to date (Radverkehrsatlas), @Phyks and @Florimondable as the authors of the Cyclosm cartocss style and @matkoniecz generally for his opinion on tagging.

westnordost, line 97: > tags["sidewalk:both:bicycle"] == "yes"

Don't check for `both` specifically, the sides are expanded in a parsing step before, in `expandRelevantSidesTags`. Add `result.expandSidesTags("sidewalk", "bicycle", true)` there instead.

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayCreator.kt
```diff
@@ -185,6 +196,11 @@ private fun CyclewayAndDirection.applyTo(tags: Tags, isRight: Boolean, isLeftHan
     if (!touchedSegregatedValue) {
         tags.remove("$cyclewayKey:segregated")
     }
+    // no matter what option was chosen, if it's not SIDEWAY_OK, remove that cycling is signed and ok on sidewalk if it's there
+    if (cycleway != SIDEWALK_OK && tags.containsKey("sidewalk:$side:bicycle") && tags["sidewalk:$side:bicycle"] == "yes" && tags.containsKey("sidewalk:$side:bicycle:signed")) {
```
westnordost, line 200:
```suggestion
    if (cycleway != SIDEWALK_OK && tags["sidewalk:$side:bicycle"] == "yes" && tags["sidewalk:$side:bicycle:signed"] == "yes") {
```

(Or was there a reason why you are not checking the value of `sidewalk:$side:bicycle:signed`?)

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt
westnordost, line 97: At first it seems inconsistent that
- `sidewalk:both:bicycle:signed=yes` doesn't need to be present for a road to parsed as `SIDEWALK_OK`
- but selecting `SIDEWALK_OK` adds that tag
- and for separately mapped footways, `...signed=yes` must always be specified for it to be parsed as `ALLOWED_ON_FOOTWAY`

Is this an oversight?

westnordost (changes requested):
**Alright, first full review:**

---

Wouldn't it make sense, for completeness sake to have in the end three options?

- Footway, no signs about bicycles
- Footway, but a sign allows bicycles
- A sign prohibits bicycles

This would also allow all options that can be parsed to be displayed in the overlay.

Currently, e.g. `highway=footway+ bicycle=no + bicycle:signed=yes` results in it being displayed as _"Not designated for cyclists (cycling may still be allowed)"_.

The `NOT_ALLOWED` option would necessarily continue to not imply a footway, to cover taggings like `highway=path + bicycle=no + bicycle:signed=yes`

---

One thing we can also consider _then_ is to *remove* the `PATH` option to reduce complexity and number of choices - i.e. merge `PATH` and `NON_DESIGNATED` option. One one hand, this simplifies the selection options mentioned above, because no need to imply a footway, on the other hand, ways signed as footways () have implicit restrictions on bicycles in e.g. Germany as far as I know, so whether it is `foot=designated`  or just `foot=yes` probably should make a difference for cyclists when evaluating the data. In other words, to be able to make this distinction in SC might be useful for data consumers. (Although, `highway=footway` / `foot=designated` doesn't necessarily mean that there is a sign. E.g. the typical footways within a housing estate will likely be tagged that way, but there will be no signs.)

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt
matkoniecz: `sidewalk:both:bicycle` ? I would 

I would consider it as  equivalent of `bicycle=yes` on `highway=footway footway=sidewalk` 

So meaning for me would be `SIDEWALK_OK` - but I have no very high confidence it would be shared.

But if that is ambiguous then the same applies to `highway=footway footway=sidewalk bicycle=yes` anyway and to disambig it - one would need to tag specific traffic sign


Note that current use of `sidewalk:both:bicycle` is quite low - see https://overpass-turbo.eu/s/1Laf

westnordost: Read again. This is about `=designated`.

wielandb: From my gut feeling, I would interpret `sidewalk:<side>:bicycle=designated` as someone trying to tag `SIDEWALK_EXPLICIT`, because they seem to try to put the tagging you would give to a standalone path (`highway=path` + `foot=designated` + `bicycle=designated`) to the sidewalk by prefixing those tags with `sidewalk:<side>:`

So my Idea would be: 

- Interpret `sidewalk:<side>:foot=designated` + `sidewalk:<side>:bicycle=designated` as `SIDEWALK_EXPLICIT`
- Interpret only `sidewalk:<side>:bicycle=designated` as `INVALID`, because we can't be 100% sure what is meant

wielandb: Yeah, back from when it was not about beeing signed explicitly. Fixed that.

wielandb:
> **Alright, first full review:**
> 
> Wouldn't it make sense, for completeness sake to have in the end three options?
> 
> * Footway, no signs about bicycles
> * Footway, but a sign allows bicycles
> * A sign prohibits bicycles
> 
> This would also allow all options that can be parsed to be displayed in the overlay.
> 
> Currently, e.g. `highway=footway+ bicycle=no + bicycle:signed=yes` results in it being displayed as _"Not designated for cyclists (cycling may still be allowed)"_.
> 
> The `NOT_ALLOWED` option would necessarily continue to not imply a footway, to cover taggings like `highway=path + bicycle=no + bicycle:signed=yes`
> 

For now, I have implemented and enabled all three options in the overlay.

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt
wielandb: > From my gut feeling, I would interpret `sidewalk:<side>:bicycle=designated` as someone trying to tag `SIDEWALK_EXPLICIT`, because they seem to try to put the tagging you would give to a standalone path (`highway=path` + `foot=designated` + `bicycle=designated`) to the sidewalk by prefixing those tags with `sidewalk:<side>:`
> 
> So my Idea would be:
> 
> * Interpret `sidewalk:<side>:foot=designated` + `sidewalk:<side>:bicycle=designated` as `SIDEWALK_EXPLICIT`
> * Interpret only `sidewalk:<side>:bicycle=designated` as `INVALID`, because we can't be 100% sure what is meant

Implemented it like this, for now.

westnordost: > `sidewalk:<side>:foot=designated`

That tag doesn't make sense to me. `sidewalk` already implies it being designated for pedestrians.

wielandb: Hm.... you're right. I guess we should go with `INVALID` then for this case to be safe.

app/src/main/res/values/strings.xml
```diff
@@ -1660,7 +1661,8 @@ Partially means that a wheelchair can enter and use the restroom, but no handrai
     <string name="bicycle_boulevard_is_not_a">It’s not a %1$s</string>
     <string name="bicycle_boulevard">Bicycle boulevard</string>
     <string name="separate_cycleway_no">No cycling</string>
```
westnordost, line 1663: But now this string is not used anymore, right?

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/SeparateCyclewayForm.kt
```diff
@@ -21,7 +21,7 @@ import org.koin.android.ext.android.inject
 class SeparateCyclewayForm : AImageSelectOverlayForm<SeparateCycleway>() {
 
     override val items: List<DisplayItem<SeparateCycleway>> get() =
-        listOf(PATH, NON_DESIGNATED_ON_FOOTWAY, NON_SEGREGATED, SEGREGATED, EXCLUSIVE_WITH_SIDEWALK, EXCLUSIVE).map {
+        listOf(PATH, NON_DESIGNATED_ON_FOOTWAY, ALLOWED_ON_FOOTWAY, NOT_ALLOWED, NON_SEGREGATED, SEGREGATED, EXCLUSIVE_WITH_SIDEWALK, EXCLUSIVE).map {
```
westnordost, line 24: Now, this should be possible.

```suggestion
        SeparateCycleway.entries.map {
```

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt
wielandb: I guess "valid" tagging for cycleway should always hinge on `cycleway=` beeing present? So, if no `cycleway=` tag is present at all, but `sidewalk:right:bicycle=yes` & `sidewalk:right:bicycle:signed=yes` is present, should it evaluate to `SIDEWALK_OK` or to `null`? I guess it is `null` in this case, as we want to force the user to make _some_ decision for `cycleway`.

Same with  `sidewalk:<side>:bicycle=designated`. In othwer words, should the check for theese tags be placed as a condition for `"no"` or for `null`?

westnordost: Does your argument not work for just `sidewalk:<side>:bicycle=designated`, without requiring `sidewalk:<side>:foot=designated`? (Looked at taginfo, some sidewalks are indeed tagged like that, but I'd just say that this is superfluous, nothing more?)

wielandb: > Does your argument not work for just `sidewalk:<side>:bicycle=designated`, without requiring `sidewalk:<side>:foot=designated`? (Looked at taginfo, some sidewalks are indeed tagged like that, but I'd just say that this is superfluous, nothing more?)

Yes, after more consideration, I guess it does. I was afraid someone might tag a situation like [this](https://www.google.com/maps/@51.8432515,12.2236736,3a,21.8y,37.79h,91.2t/data=!3m6!1e1!3m4!1s_W3rP2SRvL8O2KTIc6AkvA!2e0!7i16384!8i8192?entry=ttu) where the "sidewalk" is an exclusive cycleway with `sidewalk:<side>:bicycle=designated`, but I guess that would be wrong then. 

If we assume a sidewalk is always `foot=designated`, `sidewalk:<side>:bicycle=designated` would imply `SIDEWALK_EXPLICIT` to me. Still I'm unsure if `cycleway=` needs to be set for this to be valid.

matkoniecz: > Read again. This is about =designated.

ops

`sidewalk:both:bicycle=designated` would be equivalent of  `bicycle=designated` on `highway=footway footway=sidewalk` 

So meaning for me would be `SIDEWALK_EXPLICIT` - but I have no very high confidence it would be shared.

But if that is ambiguous then the same applies to `highway=footway footway=sidewalk bicycle=designated` anyway and to disambig it - one would need to tag specific traffic sign

Note that current use of `sidewalk:both:bicycle` is quite low - see https://overpass-turbo.eu/s/1Laf

westnordost: Hm, so both of you are of the opinion that `sidewalk:both:bicycle=designated` are equivalent to `cycleway:both=track` + `cycleway:both:segregated=no`.

If this tagging should be supported by StreetComplete, then it would need to be treated in both the parser (i.e. return SIDEWALK_EXPLICIT in that case) and also in the creator (i.e. remove such tagging when another option is selected). + Tests.

Since this was mentioned basically as a side-note of your implementation and has nothing to do directly with it, I leave it up to you if you want to include implementation for that in this PR. It could also be added later.

westnordost (changes requested):
Alright, went through the code one more time and also had a closer look at the tests.

Two more general things:
- a few tests in SeparateCyclewayParserKtTest should be failing
- Tests for CyclewayParser are missing. I.e. what tagging leads to SIDEWALK_OK

Also, as noted in that prior discussion, it is up to you whether you want to implement understanding `sidewalk:<side>:bicycle=designated` as an alias for `cycleway:<side>=track` + `cycleway:<side>:segregated=no` in this PR or not. If you do...

> ...then it would need to be treated in both the parser (i.e. return SIDEWALK_EXPLICIT in that case) and also in the creator (i.e. remove such tagging when another option is selected). + Tests.

app/src/test/java/de/westnordost/streetcomplete/osm/cycleway_separate/SeparateCyclewayCreatorKtTest.kt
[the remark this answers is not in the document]
wielandb: Yeah, I don't know what happened here. My Android Studio seems to have changed the formatting for some reason. I will reset the file to it's former state.

app/src/test/java/de/westnordost/streetcomplete/osm/cycleway_separate/SeparateCyclewayCreatorKtTest.kt
```diff
@@ -717,16 +842,51 @@ class SeparateCyclewayCreatorKtTest {
...
+                StringMapEntryModify("highway", "footway", "path"),
+                StringMapEntryDelete("bicycle:signed", "yes"),
+            ),
+            PATH.appliedTo(
+                mapOf(
+                    "highway" to "footway",
+                    "bicycle" to "yes",
+                    "bicycle:signed" to "yes"
+                )
+            )
```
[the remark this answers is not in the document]
wielandb: Added the following tests:
- apply non-designated on footway also removes bicycle signed tag
- apply allowed on footway adds the bicycle signed tag
- apply disallowed adds bicycle signed tag
- apply exclusive with sidewalk sets bicycle signed to yes if its currently no

westnordost:
So, only these two things are missing:

> Two more general things:
> - a few tests in SeparateCyclewayParserKtTest should be failing
> - Tests for CyclewayParser are missing. I.e. what tagging leads to SIDEWALK_OK

wielandb:
I think everything should be ready now :)

westnordost:
Great, congratulations!

A thought came to my mind yesterday night:

**Should  imply tagging (`bicycle=no` +) **`bicycle:signed=yes`** ? In either case, does the wording of the "no sign / prohibited by sign" options match the answer to this question?**

Is a clarification text necessary? Because, if it is not crystal clear, it **will** come up as an issue. We absolutely want to avoid **`bicycle:signed=yes`** getting ambiguous through this!

The legal situation in Germany is that on ways signed with  all traffic except pedestrians is prohibited, which includes bicycles. I expect the legal situation to be the same elsewhere, due to the Vienna convention of traffic signs.
On sidewalks, there will usually not be such a sign, still, cycling is prohibited. In other countries, it may be allowed.

Hence, I would say it would make sense that  implies tagging (`bicycle=no` +) **`bicycle:signed=yes`** even though there is no bicycle displayed on the sign, much like  (in Germany) implies `motor_vehicle=no` **by signage** even though no car is displayed on the sign.

Opinions?

wielandb:
Oh my... Yeah I didn't think about that either.

On the one hand,  does clearly make a statement about whether bicycles are allowed or not, so it should be `bicycle:signed=yes`. 

On the other hand, this does bring us back to the problem about "expert knowledge" on the local legislation. If we say  is `bicycle:signed=yes`, we again need the user to have the needed knowledge to know that this sign showing pedestrians means something for bicycles, which is what we wanted to avoid in the first place by clearly only asking for signed situations. 

Tricky situation...

I would be in favour of defining `<vehicle>:signed=` as only if  `<vehicle>` is shown on the sign as a symbol. 

This would also avoid using the tag `*:signed` too much. If we say  means `bicycle=no` `bicycle:signed=yes` and is worth tagging, people could say well then it also means `horse=no`, `horse:signed=yes`, `motor_vehicle=no`, `motor_vehicle:signed=yes` etc. 

Or, we could draw a distinction between SC and wider OSM. We could say that  indeed _technically means_ `bicycle:signed=yes`, `bicycle=no` but do not give the SC user the option to tag such, only letting them tag cases where it's absolutely clear. I would not be in favour of this. 

> Is a clarification text necessary?

My idea would be to make the "Path" text more clear, or in general add text that says what an option does not mean.
E.g. "Path or trail (not explicitly designated for anyone)", "A sign prohibits cycling (but others may be allowed)" or some such. Unsure, but I think this might avoid some confusion?

wielandb:
Another thought I had that might be worth revisiting: I saw that it is possible to define region-specific resource-IDs. Maybe part of a solution could be to explicitly show e.g. the german traffic signs associated with the answers in germany?

westnordost:
That would mean we take responsibility to show the correct sign in every country in the world.

mnalis:
> I would be in favour of defining <vehicle>:signed= as only if <vehicle> is shown on the sign as a symbol.

It is a hard question. IMHO:

- one option is to only add `foot=yes` + `foot:signed=yes` + `bicycle=no`  in such case, which is compliant with given current tagging situation, but leaves the source of `bicycle=no` unknown, thus not really resolving ambiguities (which would be a shame since user already expanded the effort to determine exact situation!)

- alternatively, if `bicycle:signed=*` were to be tagged on that, I'd _avoid_ using just `bicycle:signed=yes`, and use some separate value like `foot=yes` + `foot:signed=yes` + `bicycle=no` + `bicycle:signed=implied_by_foot_sign` (or similar)

But perhaps we should have a wider discussion on this, as https://wiki.openstreetmap.org/wiki/Key:bicycle:signed is not documented yet, and https://wiki.openstreetmap.org/wiki/Key:*:signed is not too specific either

>  If we say means bicycle=no bicycle:signed=yes and is worth tagging, people could say well then it also means horse=no, horse:signed=yes, motor_vehicle=no, motor_vehicle:signed=yes etc.

Well, true, it does mean that.   I recall seeing lately that some cycleways in some countries are open to `mofa=yes`, for example (which is clearly `motor_vehicle=*`. More complex is `electric_bicycle=*` vs. `speed_pedelec=*`)

One might be tempted to simplify that with [access=no](https://wiki.openstreetmap.org/wiki/Key:access) + `foot=yes`, but that wouldn't really work either. E.g. depending on legislation, such footway sign might (or might not) allow for  `dog=yes`, `inline_skates=yes`, `ski=yes`, `kick_scooter=yes`, `wheelchair=yes`  to name the few, and `access=*` affects _all of them_...

matkoniecz:
`bicycle:signed=implied_by_foot_sign` is a bit complicated but I look this idea

westnordost:
I don't like that idea at all... writing a longer comment now...

westnordost:
Well, to make a few points for the sentiment that `bicycle:signed=yes` should just mean that there is a sign from which `bicycle=no` is derived:

1. _(in response to `bicycle:signed=implied_by_foot_sign`)_
   `*:signed` in general is used to denote that the referred value is derived from a sign. The linked wiki page also states that `source:*=sign` is an alternative way to express this because it's meaning is very similar or the same. The tag does not define through _what_ sign it is derived and even if we _wanted_ to also record that information, there is the `traffic_sign` tag already, which will just cover that. Inventing new values for a `yes`/`no` key always turns out to be somewhat problematic because it turns a boolean into effectively a free text field.

2. Regarding the fact that a bicycle is not displayed on the , I think it is not that relevant. You don't see a vehicle on  either. The argument about it being expert knowledge (more or less) still stands, but this might be solved by adapting the description accordingly.


3. The translated German description of  is 
    > **Restriction or prohibition**
    >
    > 1. traffic other than pedestrians may not use the sidewalk.
    >
    > 2. if additional signs permit the use of a footpath for another type of traffic, this must take pedestrian traffic into consideration. Pedestrian traffic must not be endangered or obstructed. If necessary, traffic must wait; it may only travel at walking speed.
    >
    > **Explanation**
    >
    > The sign indicates a footpath (Section 25 (1) sentence 1) where clarification is required.
    
    and I suppose it will be similar in other countries, because the blue circular sign always has the same meaning.

wielandb:
> That would mean we take responsibility to show the correct sign in every country in the world.

Oh no, I did not mean that. My idea would have been:

in germany -> show german signs
in country X -> show X signs
else -> show pictograms as they are now

So use the pictograms the way the are right now normally, but if an explicit resource showing a sign is available for that country, show that.

wielandb:
> Regarding the fact that a bicycle is not displayed on the , I think it is not that relevant. You don't see a vehicle on  either. The argument about it being expert knowledge (more or less) still stands, but this might be solved by adapting the description accordingly. E.g. "Signed footway for pedestrians only or bicycles are explicitly prohibited"

So, do you propose to update the current option "Not designated as cycleway (cycling may still be allowed)" or the option "A sign prohibits bicycles" to this new wording, or introduce a new option altogether?

mnalis:
> Sure, the blue circular sign means that it is designated for that traffic participant, i.e. foot=designated + foot:signed=yes is correct, but from *=designated alone it cannot be derived that a way is prohibited for all other traffic participants. However, that blue circular sign explicitly disallows other modes of transportation.

> traffic other than pedestrians may not use the sidewalk.
> and I suppose it will be similar in other countries, because the blue circular sign always has the same meaning.

_(doing some research)_

Well, I wouldn't be so sure... In nearby Croatia, for example, such sign indeed means it  is forbidden for any _vehicle_ (_"vozilo/a"_).

However [ZOSPNC 2.1.28.](https://www.zakon.hr/z/78/Zakon-o-sigurnosti-prometa-na-cestama) defines _"vehicle"_ somewhat differently, i.e. makes exception for _"children's means of transport, personal or motorized means of transport for persons with disabilities or the elderly, if they do not move at a speed higher than the speed of human walking"_

So, **kid on kids bicycle**, as well as **wheelchair user** (provided they move slow enough) are not considered vehicles by law and **are allowed on all such footways**. Contrast that to [OSM](https://wiki.openstreetmap.org/wiki/Key:access), where `wheelchair=yes` (regardless of its speed), as well as `bicycle=yes` (regardless of `bicycle:type=child`) are still considered `vehicle=yes`.

How the low defines in some further away country, I wouldn't dream to claim to comprehend.


So, while tagging it as `foot=designated` + `foot:signed=yes` are correct in Croatia (and I'd guess elsewhere), **`bicycle=no` is _actually incorrect_ for some bicycle users** (slow small kids) here, so it is more like `bicycle=yes` + `bicycle:conditional=none @ (speed>walk); none @ (bicycle:type != child)` + `wheelchair=no` + `wheelchair:conditional = yes @ (speed<=walk)` or some similar country-dependent horror.

(i.e. the issue seems somewhat akin to maxspeed issue with complex solution like https://westnordost.github.io/osm-legal-default-speeds)

---

**TL;DR:** I would rather prefer that SC would add only fully factually correct and verifiable data, but if it had to simplify with such partly-incorrect information like `bicycle=no`, I would **really** like add some tag to indicate that. 

You're absolutely right about `bicycle:signed` probably being wrong tag, though! How about `source:bicycle=implied_by_pedestrian_sign` (or something similar)? It is freeform text already, and it is even [documented on the wiki](https://wiki.openstreetmap.org/wiki/Key:source:bicycle), has dozen thousands of uses, and we already do freeform things like `surface:note=*` in SC.


---

wielandb:
> So, while tagging it as foot=designated + foot:signed=yes are correct in Croatia (and I'd guess elsewhere), bicycle=no is actually incorrect for some bicycle users (slow small kids) here,

Just for completeness sake, it's the same way in germany. But since this is universal for everywhere in germany, we don't tag it onto every sidewalk. (It's not recorded in https://www.openstreetmap.org/relation/8131479 (yet), but if at all, it should be recorded there.)

westnordost:
@wielandb  Right, right, I am absolutely for not opening cans of worms...

I thought your comment regarding `horse:signed=yes`, `motor_vehicle:signed=yes` etc. referred to discussions in the community. Yeah, for implementation of StreetComplete bicycle overlay, it is not an issue, but for people that would check every box in JOSM access presets (and/or people complaining about those people), it might. In the end, it will be a problem for StreetComplete, because what goes around comes around.

Maybe your initial suggestion is the best one, after all. And adapt the wording accondingly:

<table>
<tr><th>Wording</th><th>Tags</th></tr>
<tr><td><ins>Generic</ins> path or trail </td><td><code>highway=path</code> / <code>foot=yes + bicycle=yes</code></td></tr>
<tr><td><del>Not designated for cyclists (cycling may still be allowed)</del><br><ins>Footway, no sign about cycling<br>(whether it is allowed depends on legislation)</ins></td><td><code>highway=footway</code> / <code>foot=designated</code></td></tr>
<tr><td>A sign prohibits cycling</td><td><code>bicycle=no + bicycle:signed=yes</code></td></tr>
<tr><td>Footway, but a sign allows cycling</td><td><code>bicycle=yes + bicycle:signed=yes</code></td></tr>
<tr><td><ins>Signed</ins> Shared-use path</td><td><code>foot=designated + bicycle=designated</code></td></tr>
<tr><td>...</td><td>...</td></tr>
</table>

app/src/main/res/values/strings.xml
```diff
@@ -1659,20 +1660,21 @@ Partially means that a wheelchair can enter and use the restroom, but no handrai
...
+    <string name="separate_cycleway_non_segregated">Designated shared-use path</string>
     <!-- foot=designated + bicycle=designated + segregated=yes -->
-    <string name="separate_cycleway_segregated">Separated bike and foot path</string>
+    <string name="separate_cycleway_segregated">Designated separated bike and foot path</string>
     <!-- highway=cycleway / bicycle=designated -->
-    <string name="separate_cycleway_exclusive">Bike-only path</string>
+    <string name="separate_cycleway_exclusive">Designated bike-only path</string>
     <!-- highway=cycleway + sidewalk=* -->
-    <string name="separate_cycleway_with_sidewalk">Bike-only path with adjacent sidewalk (separated by a curb)</string>
+    <string name="separate_cycleway_with_sidewalk">Designated bike-only path with adjacent sidewalk (separated by a curb)</string>
```
wielandb, line 1677: Changed the strings

app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt
```diff
@@ -122,7 +124,14 @@ private fun parseCyclewayForSide(
...
         "separate" -> SEPARATE
-        "no", "opposite" -> NONE
+        "opposite" -> NONE
+        "no" -> when (isCyclingOkOnSidewalk) {
+            true -> SIDEWALK_OK
+            false -> when (isCyclingDesignatedOnSidewalk) {
+                true -> SIDEWALK_EXPLICIT
+                false -> NONE
+            }
+        }
```
westnordost, line 134: I was about to write "uh... that seems it does not work?" until I realized that it is just written in a confusing way. Let me correct that.

```suggestion
        "opposite" -> NONE
        "no" -> when {
            isCyclingOkOnSidewalk -> SIDEWALK_OK
            isCyclingDesignatedOnSidewalk -> SIDEWALK_EXPLICIT
            else -> NONE
        }
```

app/src/main/res/values/strings.xml
westnordost: Right, thanks, but I tent towards removing "Designated" from the last three options again.

For the shared-use path, it is important to have "designated" to distinguish it from the "path or trail".

For the others, I don't think it does add value as there is no risk of mix-ups with other values. To the contrary, I think it might actually **add** potential for confusion: Bike paths are _always_ designated for cyclists in one way or another, otherwise it would not be a bike path (but a path or trail), so no need to explicitly state it. However, using the word "designated" (=DE: _gekennzeichnet_) may invoke the notion in users that to choose this option, the path must be *signed* as a bike path, which is not required. E.g. many bike paths in Germany are not signed but are distinguishable from the footway by different (often red) paving - whether these bike paths are "_de: gekennzeichnet_" as such rather than just _recognizable_ as such is a topic people may butt their heads about. (After all, there indeed was a long discussion about it in the forum, whether `*=designated` means that it is signed in one way or another.)

westnordost (approved):
Great, let's put this into the release I am going to do today!

app/src/main/res/values/strings.xml
wielandb: Removed the last three "Designated"s

westnordost:
strings uploaded to POEditor now, will release v58 this evening.

westnordost:
I see you've been editing German translation the same time as I did. I am done now, feel free to review and see if you find better translations.

🎉

frankie2784:
Can I clarify the difference between:

"Footway, but a sign allows cycling	" and "Designated shared-use path"?

In Australia there is no distinction, legal or otherwise.

westnordost:
first is `highway=footway` / `foot=designated`, second is `foot=designated` + `bicycle=designated`.

I.e. the first would be designated as a footway, but there is some additional sign that _allows_ also bicycles (the latter obviously just being "guests").

The second would be designated as a foot+bike path. The picture you see on the [second last page in this document](https://hdp-au-prod-app-tam-yourvoice-files.s3.ap-southeast-2.amazonaws.com/7416/1887/6734/DRAFT_Active_Transport_Strategy_document.pdf) would be a shared-use path.

Depending on country and legislation, the difference between the two can indeed be marginal.

frankie2784:
That's fine. I just think having both options might end up being confusing
for your audience in countries where there is no meaningful distinction
between the two.

Perhaps having the "footway, cycling permitted" option available only in
relevant countries similar to, e.g. advisory bike lane, would be a good
idea? That way, users can request to opt in for their jurisdiction if
"designated shared use path" is deemed not sufficient.

westnordost:
It would be more problematic (and complex) to not display this option in certain countries, because unlike bike lanes, these are not subtags that could then just be ignored. I.e. a situation where in a country were (officially) one of these options does not exist would need to be handled gracefully, without potentially removing valid data.

So, for one thing, we would need to be extra careful which countries would be added to a presumed `bicycleOnFootwayIsOkSignDoesntExist.yml` list. But we also need some rules, agreed on by the community, how to handle cases where despite apparently an official sign missing, such tagging is still found in the wild in said country. And these rules should of course be globally valid because these tags should not mean different things in different countries.

frankie2784:
I'm still not entirely clear on what the legal or practical difference is
between the two categories.

Is it that pedestrians have right of way in the first case but not in the
second? Do bikes as "guests" have to ride slower or differently than if
"designated"?

And how will this be communicated to users of the app (through an icon or
other means) as to how to identify the right category if this distinction
doesn't exist in their location?

I ask because there is a risk that people may get confused and start
changing sections from bicycle "designated" to "yes", or vice versa, when
the two have different connotations in some routing apps (for example,
bicycle "yes" on a footpath in Victoria where I live is typically used to
identify where there is NO signage, but cyclists informally use it anyway -
so not legally, but do in practice).

westnordost:
What legal difference it makes depends on the legislation. In Germany, it does make a difference.

Your example sounds like an instance of, well, creative tagging usage. If it is not legal, then `bicycle=no` is wrong.

But anyway, there is also the option _"Footway, no explicit sign about cycling"_, and in fact, this is the option you are going to see in StreetComplete for `highway=footway + bicycle=yes` because the "Footway, but a sign allows cycling" option is only selected if `bicycle:signed=yes`.

As for icons, well, I just released it, so you can see for yourself. For the first option, a pedestrian is shown and on the lower corner you see "🚲 OK" in what looks like a sign or annotation. For the second, a pedestrian and a bicycle is shown together.

mnalis:
> That's fine. I just think having both options might end up being confusing
for your audience in countries where there is no meaningful distinction
between the two.

Yes, I was quite confused by that German distinction too... 
Perhaps we could add some small clarification in that recently-available "info button" :information_source: ?

> strings uploaded to POEditor now, will release v58 this evening.

I understand (and support!) the need for _"release early, release often"_, but with my translator hat on, I'd really _**appreciate if translators were given more than few hours notice**_ to do their work, especially on weekends (when people are sometimes even known to be offline!)... say, at least 2-3 days between publishing strings and doing a release?

frankie2784:
Is it only a thing in Deutschland?

mnalis:
> Is it only a thing in Deutschland?

Probably not only there (though I have not done research where else it might be the case); but it is definitely not a thing in Croatia (which is why it confused me).

westnordost:
Oh, shit, I should have checked that... I hope it is just a mistake in the tests!

wielandb:
Oh no! At least on my machine, all tests passed when making the final commit. Will check this ASAP.

westnordost:
I skimmed through the failing tests. Fortunately it just looks like some tests were not adapted to the new behavior, so need for a hotfix or anything.

Many tests are named like they test whether a certain tag is added/removed on certain combinations.
For this, a check like this could be used instead of `assertEqual` to make the test more focussed on what is actually tested. E.g.:

```kotlin
    @Test fun `apply allowed on footway adds the bicycle signed tag`() {
        assertTrue(
            ALLOWED_ON_FOOTWAY.appliedTo(mapOf(
                "highway" to "footway",
            )).contains( // or containsAll if it is several that should be checked.
                StringMapEntryAdd("bicycle:signed", "yes")
            )
        )
    }
``̀ 

(compare with current version)

westnordost:
Anyway, frankie2784, I only explained why it is more effort and complex to implement that. Something like this is certainly a valid and meaningful feature request but to make this a reality, first research is needed in which countries a distinction between those two options is made, either in legislation or by the local community.
A good place to start might be the forum.

wielandb:
> A good place to start might be the forum.

I created this post:
https://community.openstreetmap.org/t/does-your-country-have-the-legal-concept-of-designated-for-pedestrians-but-cyclists-allowed/114096

wielandb:
Should `apply allowed does not re-tag bicycle=permissive etc` still succeed? 
In other words, if we apply "A sign allows cycling", should a `bicycle=permissive` be turned into a `bicycle=yes` + `bicycle:signed=yes` or should it stay a `bicycle=permissive`? I guess that it should not become `bicycle=permissive` + `bicycle:signed=yes` because that would mean the _permissiveness_ is signed.

westnordost:
Yeah, I would say you are right. If there is a sign, permissive sounds wrong. After all, the surveyor can't and should not need know whether the signed access restriction was put there by the owner or by the municipality. On the other hand, if there is something like a hand-written sign like "hey bikers, feel free to use this shortcut through my property here!" or whatever, maybe indeed `permissive` would be a better fit but I doubt that this kind of thing is the usual situation one expects to encounter with `bicycle=permissive`.

wielandb:
> Since this was merged, a few tests in `SeparateCyclewayCreatorKtTest.kt` are failing (see https://github.com/streetcomplete/StreetComplete/actions/runs/9330540814/job/25684214512). @wielandb, could you take a look at that please?

Created #5672 to fix this

Issue #5569, opened 2024-04-01
Question for street width is asked endless again and again


ratti:
**How to Reproduce**
See node https://www.openstreetmap.org/way/1228191464 , roughly located in Hamburg, Bernadottestraße 242a
1. Go to that place in SC, answer question on road width (see screenshot) with "3"
2. Get asked about street width _again_ (Countercheck in JOSM: maxwidth: is already given as "3")
3. Enter "3"
...and again and again.

**Expected Behavior**
- not getting asked at all, since "3" is already given
- If I get this wrong and the attribut is not "maxwidth", but something else: Ask once, store it, don't ask again.

**Versions affected**
57.1 from Github
![Screenshot_20240401-212656](https://github.com/streetcomplete/StreetComplete/assets/2254123/a33efa7b-c25a-43bc-a6d5-39c789bc06b2)

westnordost:
Yes, very likely the same as #5545. I believe progress on tracking down the source can only be made after the next update, since I added some logging.

ratti:
I found some time to play around with this and now I believe this is not related to #5545 , but a “simple“ issue with attribute dependencies.

I found it suspicious I could tag the whole street, but not the 2 parts that are a “schikane“ (narrow road for speed reduction). Comparing a working part with a not working part I noticed a difference:
- Road parts having “maxwidth“  are asking repeatedly for maximum width
- Road parts having “width“ are fine

So I changed that not working parts in JOSM and checked what would happen in SC.

Road: https://www.openstreetmap.org/way/1228191464

Initially attributes:
```
cycleway:left:surface=compacted
cycleway:right:surface=compacted
cycleway=track
highway=residential
lanes=1
lit=yes
maxspeed=30
maxwidth=3
name=Bernadottestraße
sidewalk:left:surface=paving_stones
sidewalk:right:surface=compacted
sidewalk=both
surface=sett
traffic_calming=choker
traffic_sign=DE:274.1[30]
zone:maxspeed=DE:30
```

Streetview permanently asks: “Wie breit ist die Fahrbahn hier?“ (How wide is the road here?) , to which I reply “3“
—> symbol does not disappear, question is re-asked.

Using JOSM to change “maxwidth=3“ to “width=3“
Upload from JOSM
Refresh in SC

Two new different question appear in SC (Surface type and one-way-road, but those shouldn't matter here) ⟶ problem is gone. Never get asked again.

[...]

PR #5578, opened 2024-04-12
Do not ask width in endless loop on segments of road tagged with traffic_calming=choker

mnalis:
Fixes: https://github.com/streetcomplete/StreetComplete/issues/5569

--- 

Description:

If the road had `traffic_calming=choker` (or similar), we were tagging [`maxwidth` instead of `width`]( https://github.com/streetcomplete/StreetComplete/blob/38203c669a5448414f9965b95dcd8d6defee2e35/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddRoadWidth.kt#L74).
However, in element selection code, we only selected _ways_ with non-existent `width` tag. This PR adds also a check for `maxwidth` too if the road way segment has [ROAD_NARROWERS](https://github.com/streetcomplete/StreetComplete/blob/38203c669a5448414f9965b95dcd8d6defee2e35/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddRoadWidth.kt#L91).

I've also added [automated tests](https://github.com/mnalis/StreetComplete/actions/runs/8668878845/job/23774730168) for this use case, and also checked manually with [v57.1-based debug APK](https://github.com/mnalis/StreetComplete/actions/runs/8668430767) for [w1228191464](https://www.openstreetmap.org/way/1228191464) from the bug report, and it seems to work fine now (not asking width for that road segment anymore).

(we also [already have code](https://github.com/streetcomplete/StreetComplete/blob/38203c669a5448414f9965b95dcd8d6defee2e35/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddRoadWidth.kt#L21-L26) in that quest that handles traffic choker _nodes_; I didn't touch that part).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddRoadWidth.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/width/AddRoadWidth.kt
@@ -44,6 +44,7 @@ class AddRoadWidth(
         )
         and area != yes
         and (!width or source:width ~ ".*estimat.*")
+        and (traffic_calming !~ ${ROAD_NARROWERS.joinToString("|")} or !maxwidth or source:maxwidth ~".*estimat.*")
         and (surface ~ ${ANYTHING_PAVED.joinToString("|")} or highway ~ ${ROADS_ASSUMED_TO_BE_PAVED.joinToString("|")})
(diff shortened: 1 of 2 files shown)


westnordost:
Oh, thank you for the analysis and even the fix for the issue!

I've reviewed the code and it looks fine.

PR #5589, opened 2024-04-21
Do not ask payment methods if payment:others=no is set

urbalazs:
The [wiki](https://wiki.openstreetmap.org/wiki/Key:payment:*#Others) says that `payment:others=no` can be used if you have already indicated **all** methods of payment which are accepted. Do not need to ask whether cash or cards are accepted if this key is set.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/accepts_cards/AddAcceptsCards.kt
@@ -17,7 +17,7 @@ class AddAcceptsCards : OsmFilterQuestType<CardAcceptance>() {
           amenity ~ restaurant|cafe|fast_food|ice_cream|food_court|pub|bar
           or (shop and shop !~ no|vacant|mall)
         )
-        and !payment:credit_cards and !payment:debit_cards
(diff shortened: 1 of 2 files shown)


mnalis:
Make sense, although one should add checks for `no` value explicitly (there are few percent of other values for `payment:others`, like `yes`)

Helium314:
Shouldn't `yes` mean that _all_ other methods are accepted? (doesn't sound like something that could actually be correct though)

westnordost (changes requested):
Thanks for the PR, makese sense! As per the comments above, the filter should be `payment:others != no` though.

urbalazs:
I added only the key without value because

- `payment:others=no` means that no need to ask if missing cash or card method is needed since others is **no**
- `payment:others=yes` means that no need to ask if missing cash or card method is needed since others is **yes** - I know that this makes no sense, but StreetComplete is not a validator to decide whether a tag makes sense or not
- missing `payment:others=*` means that SC should ask if missing cash or card can be accepted

westnordost:
I think `yes` is the opposite of `no`, so it means "yes, there are other payment methods that may accepted" rather than "yes, ANY other payment method imaginable is accepted". As you and Helium314 noted, the latter would make no sense.

mnalis (approved):

Issue #5595, opened 2024-04-25
Replace ballpit quest-view in user profile with something else


westnordost:
We are currently using `org.jbox2d:jbox2d-library` for the physics simulation that is happening in the ballpit view in the user profile for the quest solved statistics (per-quest-type and per-country).

There is no Kotlin Multiplatform replacement for this.

While Box2D exists as a C++ library and thus it is theoretically possible to add bindings to that in Kotlin, I deem this far too much effort to preserve functionality that is essentially a fun gimmick. A gimmick that has clear downsides, too: With the number of quests we already have, the view is becoming more and more imperformant and most quest types are not properly clickable because they are just too small. Also, the ballpit never worked that well for the per-country-stats because there is usually one bubble that dominates all.

So, we need a replacement for this, _ideally_ something that is also fun and interesting to look at. But the effort or feasibility to implement it (in Jetpack Compose) needs to be taken into account, too.

**Proposed Solution**
For now, I'd like to collect some ideas what could be done. I created some sketches. 

![img20240425_13243071](https://github.com/streetcomplete/StreetComplete/assets/4661658/4f4444ec-5e9b-4024-8109-364fe846a156)

Do you have other ideas? Do you have ideas or hints how to implement one or the other?

Especially the tree map I find interesting, but not sure how to implement it or if the end result actually looks fine, given the lack of really hierarchical data. I would appreciate if someone interested would squeeze some actual quest statistics data through a program that is able to create such a visualization to be able to judge if the end result can look good.

FloEdelmann:
I had a first look into this. As a dataset, I used the 2024 counts from https://piebro.github.io/openstreetmap-statistics/#6773, divided by 1000, to have a rough estimate. Here is a CSV file with those numbers for use in data visualization software: [streetcomplete_fake-2024-counts.csv](https://github.com/streetcomplete/StreetComplete/files/15120422/streetcomplete_fake-2024-counts.csv)

Then I tried to visualize these numbers with [Datawrapper](https://www.datawrapper.de/) (a tool I can definitely recommend for data visualization!), but unfortunately, there are no really interesting plots. These are the ones worth sharing:

- [Bar chart](https://www.datawrapper.de/_/mctpb/)
- [Pie chart with outer labels](https://www.datawrapper.de/_/PJKH9/) (max. 8 sectors → 7 quest types plus one "Other" sector)
[...]

PR #5607, opened 2024-04-29
First steps in Jetpack Compose (reimplement parts of the user screen)

westnordost:
I re-implemented the UI of parts of the user screen in compose and set up compose in general (define the default theme etc.). When this is merged, others can contribute re-implementations of parts of the UI in compose.

<details>
<summary>
I initially planned to re-implement the whole user screen, but I came upon two issues which made me postpone this. (Click for more info)
</summary>

1. physics-ballpit needs to be replaced with something more boring. On the slim chance that someone with know-how pops up or when I return back to this, I have become a Compose-wizard, I'll leave this open for now (#5595)

2. Login currently uses the WebView, for which there is no official Compose replacement. There is a [library by a third party](https://github.com/KevinnZou/compose-webview-multiplatform), but it misses at least one crucial part of the API to be used for OAuth authorization -  `shouldOverrideUrlLoading` - used for retrieving the authorization code from the callback URL. So, maybe it is better to use the browser for OAuth authorization instead of a webview (which is also recommended in OAuth 2 anyway) but in any case I will postpone this because of that.
</details>

**@ Contributors and people thinking about contributing in the future:** Feel free to look through this and learn from it.

To be honest, I have to admit that it's quite fun to use Compose, the API, at least to do the standard things, is super straightforward. The previewing tools are excellent. But since the design is being done in code, it leaves more opportunities to turn the layout code into unreadable blurb, which is why more care is necessary splitting up components in the UI into own reusable composables for readability (and reusability). After all, this wasn't really done at all before, as all design was locked away in the layout XMLs (with lots of duplication inside). One reason for the almost -1000 lines of code on this PR.

**@ People who know Jetpack Compose:** Feel free to look through this and suggest improvements.

---

Since having viewmodels is almost something like a requirement to effectively convert the UI to compose and sometimes it is a bit more complex, I will continue with that after this PR is merged.


Jean-BaptisteC:
Is it possible to have an APK to test these changes?

westnordost:
I also changed some behavior (slightly) because it was easier to do in Compose while others were mere complicated to do.

 @Jean-BaptisteC Why would you want to test the changes? It should work as before, except the some small things I changed. And it is just three screens.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/debug/ShowLinksActivity.kt
```diff
@@ -1,33 +0,0 @@
...
-import de.westnordost.streetcomplete.R
-import de.westnordost.streetcomplete.databinding.FragmentShowLinksBinding
-import de.westnordost.streetcomplete.screens.BaseActivity
-import de.westnordost.streetcomplete.screens.user.links.GroupedLinksAdapter
-import de.westnordost.streetcomplete.util.ktx.openUri
-import de.westnordost.streetcomplete.util.viewBinding
-import org.koin.androidx.viewmodel.ext.android.viewModel
-
-/** activity only used in debug, to show all achievement links */
-class ShowLinksActivity : BaseActivity() {
```
westnordost, line 15: @FloEdelmann I removed this because it is more convenient to preview the links directly in the IDE, see LazyLinksColumn

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileViewModel.kt
```diff
@@ -75,37 +68,6 @@ class ProfileViewModelImpl(
     override val biggestSolvedCountCountryStatistics = MutableStateFlow<CountryStatistics?>(null)
     override val biggestSolvedCountCurrentWeekCountryStatistics = MutableStateFlow<CountryStatistics?>(null)
 
-    override var lastShownGlobalUserRank: Int?
```
westnordost, line 78: @matkoniecz I removed this because I found it to be a bit complicated or at least not worth the amount of code to keep it in Compose. Now, the badges just always animate, which is nicer anyway. Sure, the sense of progression is lost, but that feature was too well hidden anyway, i.e. it was barely noticeable what changed exactly because the animation was played automatically , so one would have to look really quick anyway to see what the value before was. And making it more prominent would on the other hand usually just look too disruptive. (Don't need a "New achievement unlocked"-type of dialog every time my rank is changed towards the better)

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/LaurelWreath.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.compose.ui.graphics.drawscope.inset
+import androidx.compose.ui.graphics.drawscope.withTransform
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.tooling.preview.Preview
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.ui.theme.LeafGreen
+
+@Composable
+fun LaurelWreath(
```
westnordost, line 21: @matkoniecz I re-did the laurel wreath in compose. Feel free to look through the code to understand what I changed.

westnordost:
Changes to current behavior:

- badges have double the size, nice and smooth animation always when entering the screen

- dates active table is bigger, i.e. fills screen width, so especially on tables etc. it is bigger. Also, added support for right-to-left layouts.

- there is no appear-animation on opening the links screen or the achievement screen. (Nothing conveniently built-in in the API for that yet, and I didn't want to add cumbersome code just to preserve this somehow)

- the achievement info dialog should work much better on a variety of different screen sizes and orientations now

- the achievement info dialog is much more plain than before: No fancy pop-in animation that animates from the clicked view (no idea at all how to implement that in compose) and before I add some custom appear-animations to some (i.e. the first I re-implement) dialog, I think I better look into how maybe all appear-animations of all dialogs could be made to look consistent.

Helium314:
Did you try the screen in dark mode?
In SCEE I have dark text on dark background (didn't test SC yet, though).

westnordost:
Hmm... I thought I did. I will check.

app/src/main/java/de/westnordost/streetcomplete/screens/user/profile/ProfileViewModel.kt
matkoniecz: After encountering it several times since them I want to say - this was actually a significant improvement in how nice this menu is!

PR #5612, opened 2024-05-02
Info button and hints

### LINKED ISSUE #1913: Add clarifications / details for quests

westnordost:
Kind of fixes #1913. Many suggestions in #1913 are not actionable because the definition of certain tags is vague in the wiki and I cannot simply make something up. (E.g. `lit` status on bus stops.)

This PR mostly moves the hints into an area that is expanded when pressing the info button that is shown on the right side of the question, however, adds a few hints where I found they make sense. I did not move all texts that appear as hints into the info area because with this implemented, we now have three places in which things can be clarified and each are for different use cases.

<table>
  <tr><th>Tool</th><th>Use case</th><th>Example</th></tr>
  <tr>
    <td>Info hint text & images</td>
    <td>Doubts on the question itself (terms used, etc.), <b>how to</b> answer or how to find the answer.</td>
    <td>Explanation of what is halal, what is tactile paving, what is a sanitary dump station, when a level counts as a roof level and when as a normal level etc., explanation of the street side puzzle UI</td>
  </tr>
  <tr>
    <td>Answer option (including in "uh...")</td>
    <td>Doubts on how to answer a specific situation.</td>
    <td>Buildings with several uses, e.g. retail and residential</td>
  </tr>
  <tr>
    <td>Label text for an input field</td>
    <td>Instruction on how to answer, i.e. in what format the answer is expected. These should always be shown and be short.</td>
    <td>"Pick one"; that road names should not be abbreviated; that on the vegetarian quest one should only answer "yes" if there are proper options...</td>
  </tr>
</table>


westnordost:
If anyone wants to review this, the most interesting part is the strings.xml. I added some strings and modified some strings.

westnordost:
By the way I didn't add an animation because it is somewhat cumbersome to do. Since sooner or later the UI is redone in Jetpack Compose, I didn't want to put an unreasonable amount of effort in it:
- `animateLayoutChanges`=true seems to be bugged when used with bottom sheets. The animation looked fine, but it also made the bottom sheet form go to the top of the screen (wth)
- a normal object animator animation with `scaleY` will literally just scale the content, which looks odd. Animating the layout height on the other hand had the effect that the layout children tried to scale themselves to fit the mandated height during animation (i.e. images weirdly grow in size)

FloEdelmann:
Had a look at the code and strings now, they look fine!

FloEdelmann:
Should we add the table from the PR description to some more permanent place, like the quest docs?

app/src/main/java/de/westnordost/streetcomplete/quests/foot/ProhibitedForPedestriansAnswer.kt
```diff
@@ -3,6 +3,5 @@ package de.westnordost.streetcomplete.quests.foot
 enum class ProhibitedForPedestriansAnswer {
     YES,
     NO,
-    IS_LIVING_STREET,
```
FloEdelmann, line 6: I guess this was removed on purpose? Now a test is failing, see https://github.com/streetcomplete/StreetComplete/actions/runs/8960282103/job/24606606961:

> app/src/test/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestriansTest.kt:57:39 Unresolved reference: IS_LIVING_STREET

westnordost: oops. Thanks!

westnordost:
> Should we copy the table from the PR description to some more permanent place, like the quest docs?

To "contributing a quest?". I don't know, I don't really maintain this, that's @matkoniecz project.

app/src/main/java/de/westnordost/streetcomplete/quests/foot/ProhibitedForPedestriansAnswer.kt
FloEdelmann: Fixed in https://github.com/streetcomplete/StreetComplete/commit/1fe1f79925c1b72592212f64d5b31609a03ac288

Helium314:
Finally had time for a look. Very nice, thank you!

mnalis:
Can somebody check if the info button is visible for them in **dark mode**?

small_Screenshot_20240523_085403_StreetComplete Dev small_Screenshot_20240523_084956_StreetComplete Dev

(I've merged the latest master to my fork for some testing, so I cannot easily say if it is due to my local changes until it is released)

westnordost:
Right, fixed it.

HolgerJeromin:
ref 7a40b562cc6251a03eda951b6319cd5c170243ea

PR #5615, opened 2024-05-04
Update Sponsors in README.md

danieldegroot2:
Adds Prototype Fund grant StreetComplete for iOS.
Also, sorts and simplifies Sponsors based on existing content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,23 +49,22 @@ This software is released under the terms of the [GNU General Public License](ht
 
 ## Sponsors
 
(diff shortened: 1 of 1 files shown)


westnordost:
Oh, thank you, I have to admit that I forgot about it! I will rather do this myself though. The sponsors are sorted by rough amount.

westnordost:
Alright, I updated this section now, replaced some graphics etc. (Patreon has a new logo... again!)

Thanks for the reminder!

Issue #5470, opened 2024-02-02
Unclear fire hydrant diameter quest in Austria


Helium314:
In SC 56.1 the `AddFireHydrantDiameter` quest got enabled in Austria. I checked 2 hydrants, but none of them had the necessary sign (not on the hydrant, and not nearby). I found some numbers on the hydrants (like `A100 - 16 - 1.50/E`), but none explicitly stated any kind of diameter.

The commit https://github.com/streetcomplete/StreetComplete/commit/e0e1b0476eee80221bfbf992bfd898dab355c333 added a link to images, including https://commons.wikimedia.org/wiki/Category:Fire_hydrant_signs_in_Austria.
Only on one of the three images there is a sign that looks like in the one displayed in the quest form.
On two signs on that page I cannot find a diameter at all. One of the 2 hydrants I checked had such a blue sign in vicinity, however I'm not sure whether it was actually related to the hydrant.

Is it expected that many/most hydrants will get the _not signed_ answer?

westnordost:
> Is it expected that many/most hydrants will get the not signed answer?

At this point, neither expected or unexpected. I don't know how many hydrants would be signed in Austria.

Before this quest was introduced, I thought that every fire hydrant in Germany has this sign. Now, gut feeling, about 50% are signed.

The blue signs in Germany are as far as I know not hydrant signs but for the water supply.

Maybe if you asked in the Austrian community forum, you will get more helpful answers from Austrians? I have no problem with disabling the quest again for Austria.

westnordost:
That's about numbering, not the hydrant diameter.

westnordost:
Anyway, I would like to leave this in your hand. If you think it does not make sense to ask in Austria, just create a PR to disable it again. It would be nice if you asked around for more input, but in any case, both of you are more knowledgeable than me about this topic so whatever you decide has more weight than the decision I made solely base on a few pictures on wikimedia.

hungerburg:
There might be some misunderstandings: The sign is used a lot, but not for the pillar type hydrants, instead for the underground hydrants. I made a picture showing both the sign and the hydrant and added an explanation, how to read the sign to learn the position of the hydrant. https://commons.wikimedia.org/wiki/File:Unterflurhydrant_Wegweiser.jpg

PS: The diameter (DN) of the underground pipe is of no interest for the fire brigades. They care about volume flow (l/min) and pressure. Pipe DN is not signed on over-ground hydrants. Never seen that.

[...]

PR #5617, opened 2024-05-05
Added Hungarian sign to quest fire_hydrant_diameter

dnet:
Since [the German sign for fire hydrants](https://commons.wikimedia.org/wiki/File:Bachs_Garten,_Arnstadt.jpg) is very similar to [the Hungarian one](https://wiki.openstreetmap.org/wiki/Hungary/Jel%C3%B6l%C3%A9si_p%C3%A9ld%C3%A1k/T%C5%B1zcsap), I forked that one and modified the `H` to `T` (for _Tűzcsap_). I compiled a debug build from this version and tested it in the emulator, it seemed to work well, see screenshot below.

Screenshot of debug build running in Android Studio emulator

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_diameter/AddFireHydrantDiameter.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/fire_hydrant_diameter/AddFireHydrantDiameter.kt
@@ -33,6 +33,7 @@ class AddFireHydrantDiameter : OsmFilterQuestType<FireHydrantDiameterAnswer>() {
         "DE", "BE", "LU",
         // not "AT", - see https://community.openstreetmap.org/t/streetcomplete-quest-zu-hydrantendurchmesser-in-osterreich/108899
         "GB", "IE",
+        "HU",
         "PL",
         "FI",
         "NL"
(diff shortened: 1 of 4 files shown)


westnordost:
Ah, I see, cool! Code looks fine, however are the quest guidelines fulfilled for asking for fire hydrant signs in Hungary? Most notably

> 💤 No spam: It must be possible to determine whether the quest should reasonably be asked. A quest which is to be answered in 99% of the time with the same answer is not a good quest, don't bore the users. I.e. asking for any road if it is a one-way road, is silly, because the vast majority of all roads will not be one-ways.

Just recently, I noticed that the hydrant sign in Austria is the same as in Germany, so I added Austria to the list of countries in which to ask for what this sign says. Only to be notified by Austrian user(s) that while that sign may exist, practically no hydrants in Austria do have a sign. See #5470

dnet:
> are the quest guidelines fulfilled for asking for fire hydrant signs in Hungary?

In most cases, yes, fire hydrants (at least on publicly accessible locations) have this sign in Hungary.

westnordost (approved):

westnordost:
Cool! Thanks for the addition!

(Looking at this addition and the necessary copy-pasta duplication involved, I yearn towards migrating this stuff to Jetpack Compose)

PR #5623, opened 2024-05-08
Add automatic per-app languages preferences generation

Arthur-GYT:
All is in the title ^^.

https://developer.android.com/guide/topics/resources/app-languages

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -9,6 +9,10 @@ plugins {
 }
 
 android {
+    androidResources {
+        generateLocaleConfig = true
+    }
+
     compileOptions {
         sourceCompatibility = JavaVersion.VERSION_11
         targetCompatibility = JavaVersion.VERSION_11
--- a/app/src/main/res/resources.properties
+++ b/app/src/main/res/resources.properties
@@ -1,0 +1,1 @@
+unqualifiedResLocale=en-US


westnordost:
Oh, that's interesting.

Currently, there is an in-app selection option for that. Should that not be shown on Android  13+ then? Have you tested how the in-system-settings version of it interacts with when setting it in the app?

Arthur-GYT:
> Oh, that's interesting.
> 
> Currently, there is an in-app selection option for that. Should that not be shown on Android 13+ then? Have you tested how the in-system-settings version of it interacts with when setting it in the app?

I have tested and per-app system has priority (by which I mean that the in-app system does not take the change into account). I think we can make it so that when we click on the segment, it returns directly to the application settings when this is supported. What you think about this idea @westnordost ?

westnordost:
Yes, I think if this feature was to be enabled, then for Android 13+, that option in the StreetComplete settings should lead the user directly to the Android system settings or even not be displayed at all, including deleting/ignoring the preference if set before.

However, I do remember that one reason why this language selection setting was introduced in the first place because users were simply unable to select certain languages in the system language selection. So, are all languages in which StreetComplete is translated to even available in the system language selection? If not, then this new Android 13+ feature does not solve the use case for which the in-app language-selector was introduced in the first place.

westnordost:
So, to move this forward, someone would need to check on Android 13+ all languages selectable in the app are also selectable from the system selector. Actually, I found this file in the Android source code. I think these are the languages that can be selected in the Android system locale screen. Didn't look through / compare it with out list yet. However, it looks like at least Esperanto is missing.

https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-13.0.0_r83/core/res/res/values/locale_config.xml

riQQ:
On Android 14 I checked another app that offers the language setting and there's no Esperanto.

westnordost:
Yes, that system language picker dialog only displays the languages that are actually supported **by the app**. Question is, **if** a language is supported by the app but not by the system at large, will it be displayed in that per-app system language picker?

westnordost:
I looked into it.

E.g. Esperanto will **not** be displayed in the language picker provided for the app in the system settings, even if it is specified as a supported language. This means that in any case, the in-app language picker would need to remain. Support for the language picker provided for the app in the system settings can only be **an additional feature** (for Androids API 33 and above).

However, if it is just an additional feature, this means that we must use the appcompat API [`AppCompatDelegate::setApplicationLocales`](https://developer.android.com/reference/androidx/appcompat/app/AppCompatDelegate#setApplicationLocales(androidx.core.os.LocaleListCompat)) (etc.) on any API version to set the app's language. This API is lacking in several regards:

1. there does not seem to be any method to retrieve the **available** languages for display in a picker, despite the generated `locale_config.xml` containing this information. Hence, we must (continue to) use our own languages.yml (created by a custom gradle task)

2. on API versions below API 33, the selected language is **not** stored automatically. The options to handle it (described in the documentation of the linked `setApplicationLocales`) are awkward - either add some service to the manifest that throws some Exception which needs to be ignored somehow plus set the language *once* manually, or set the current language manually in `attachBaseContext` of each Activity - _basically what we are doing right now_.

So, not really any saved code.

Worse, the automatically generated `locale_config.xml` which includes all the languages that should be selectable (when `generateLocaleConfig = true` is specified in the build.gradle) includes all languages for which **some** localized resources exist. Seems smart, but doesn't work for our use case, as we have a gradle build task that generates "localized" `translation_info.xml`s also for languages that are not selectable. The `translation_info.xml` simply contains an integer that indicates by how much percent the translation is completed and that is displayed in the about menu. E.g. Vietnamese translation is completed by 40% but its strings will only be added and thus the language should only be selectable when translation is almost complete. But that language is included in the automatically generated `locale_config.xml` and hence selectable in the system app language picker.
This can be solved either by not generating `locale_config.xml` automatically or by choosing a different solution for this `translation_info` stuff, but either way, this is going to be extra effort.

All in all, I do not think this is worth the effort of pursuing further. Especially in light of that soon™, we'll switch to Compose Multiplatform as UI framework and there, resources management will be different anyway.

PR #5632, opened 2024-05-10
Exclude `intermittent=yes` water source from drinking water quest

esilja:
`seasonal=yes` was already excluded, but `intermittent=yes` is another such attribute: https://wiki.openstreetmap.org/wiki/Key:intermittent.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
@@ -19,6 +19,7 @@ class AddDrinkingWater : OsmFilterQuestType<DrinkingWater>() {
         )
         and access !~ private|no and indoor != yes
         and !drinking_water and !drinking_water:legal and amenity != drinking_water
+        and (!intermittent or intermittent = no)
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/drinking_water/AddDrinkingWater.kt
```diff
@@ -22,6 +22,7 @@ class AddDrinkingWater : OsmFilterQuestType<DrinkingWater>() {
         and (!seasonal or seasonal = no)
         and (!disused or disused = no)
         and (!ruins or ruins = no)
+        and (!intermittent or intermittent = no)
```
matkoniecz, line 25: maybe grouping it with seasonal and put it into next line after it would make sense?

or is it ordered in some other way?

esilja: I don't know, I just added it at the end. I changed it to follow your suggestion.

mnalis (approved):

matkoniecz:
Thanks for the first contribution! It will be available in app once release is published.

And you are invited to contribute more if you would be interested!

PR #5638, opened 2024-05-13
Add Parcel Locker Brand quest

### LINKED ISSUE #4959: What's the brand of this parcel locker?
### General
Affected tag(s) to be modified/added: [brand](https://wiki.openstreetmap.org/wiki/Key:brand)
Question asked: **What's the brand of this parcel locker?**

### Checklist

### Ideas for implementation

Empty input field, can be populated the same way that the ATM field is populated.

**Element selection:** 
`amenity=parcel_locker + !operator + !brand`

**Metadata needed:** 
Brands for each country (could be from NSI or from OSM data).

**Proposed UI:**
Same as ATM operator quest.

Arthur-GYT:
All is in the title ^^.

Close #4959


matkoniecz:
> For Country Metadata file, do I need to do something ?

See https://github.com/streetcomplete/StreetComplete/blob/8dbb9e664ad5c3b75a5ed1dddec94e0522940e7b/build.gradle.kts#L37 for existing tasks to fetch them, similar new one should be created (and likely run, to generate file for https://github.com/streetcomplete/StreetComplete/tree/8dbb9e664ad5c3b75a5ed1dddec94e0522940e7b/res/country_metadata )

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -525,4 +526,6 @@ fun questTypeRegistry(
 
     /* at the very last because it can be difficult to ascertain during day. used by OsmAnd if "Street lighting" is enabled. (Configure map, Map rendering, Details) */
     154 to AddWayLit(),
+
+    167 to AddParcelLockerBrand(),
```
matkoniecz, line 530: likely it can be put higher than at least some. Maybe below AddBikeParkingCapacity ?

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
```diff
@@ -0,0 +1,31 @@
...
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.osm.mapdata.filter
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddParcelLockerBrand : OsmFilterQuestType<String>() {
+
+    override val elementFilter = "nodes with amenity = parcel_locker and !brand"
```
matkoniecz, line 14: maybe we should exclude ones without name tags?

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
```diff
@@ -0,0 +1,31 @@
...
+    override val changesetComment = "Specify Parcel Locker brand"
```
matkoniecz, line 15:
```suggestion
    override val changesetComment = "Specify parcel locker brand"
```

nouns are not capitalised by default in English

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
FloEdelmann: I guess you mean "exclude ones *with* name tags", right?

matkoniecz: yes, definitely

Arthur-GYT: Not really relevant for Amazon lockers. They have their own name.

app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
Arthur-GYT: I'm going to add new quests related to parcel lockers in other PR's ^^

FloEdelmann: That's cool! But please see

https://github.com/streetcomplete/StreetComplete/blob/8dbb9e664ad5c3b75a5ed1dddec94e0522940e7b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt#L214-L244

TL;DR: The quests shouldn't be ordered by topic, but how easy it is to solve them.

matkoniecz: In other words, this quest can be put higher and should not be hidden by say maxspeed/lit quest (if these got enabled)

Arthur-GYT: I missed this line ^^. It has been modified

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
Arthur-GYT: For me, it wouldn't make sense for Amazon lockers, for example.

Arthur-GYT: Fixed.

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
```diff
@@ -0,0 +1,31 @@
...
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddParcelLockerBrand : OsmFilterQuestType<String>() {
+
+    override val elementFilter = "nodes with amenity = parcel_locker and !brand and !name"
+    override val changesetComment = "Specify parcel locker brand"
+    override val wikiLink = "Tag:amenity=parcel_locker"
+    override val icon = R.drawable.ic_quest_notes
+    override val isDeleteElementEnabled = true
+    override val achievements = listOf(CITIZEN)
```
FloEdelmann, line 19: Maybe this should rather count towards the `POSTMAN` achievement?

Arthur-GYT: The description of the quest is not very much in tune : `You helped to locate %d addresses!\nThis data is essential not only for the mail carrier but also for any navigation system.`

FloEdelmann: Hmm good point. The `CITIZEN` achievement kind of turns into a "catch all" achievement though. Should we maybe expand the `POSTMAN` achievement to other post-related quests, and adjust its description accordingly? (Question to all)

Arthur-GYT: I think we can create a new achievement that can be used for parcel lockers and post partners.

https://wiki.openstreetmap.org/wiki/Tag:post_office%3Dpost_partner

FloEdelmann: see also https://github.com/streetcomplete/StreetComplete/discussions/3009

FloEdelmann: > create a new achievement

Note that this is relatively much effort, as you need to create a nice picture, find a good name and description, translate those texts, and have enough quests that count towards the new achievement.

See also #3009.

Helium314: > The description of the quest is not very much in tune : `You helped to locate %d addresses!\nThis data is essential not only for the mail carrier but also for any navigation system.`

Postman is also used for mailbox collection times, royal cypher and road name quests, so I think it would be ok to use here.

westnordost: Yeah, let's use POSTMAN and maybe change the description a bit. (Anyone wants to create a PR for that?)

Arthur-GYT: It's done ^^.

FloEdelmann (approved):
Looks good from my side :)

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
```diff
@@ -0,0 +1,31 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.POSTMAN
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddParcelLockerBrand : OsmFilterQuestType<String>() {
+
+    override val elementFilter = "nodes with amenity = parcel_locker and !brand and !name"
+    override val changesetComment = "Specify parcel locker brand"
+    override val wikiLink = "Tag:amenity=parcel_locker"
+    override val icon = R.drawable.ic_quest_notes
```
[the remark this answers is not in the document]
Arthur-GYT: What do you think about that ?

parcel_locker

app/src/main/assets/country_metadata/HR.yml
```diff
@@ -14,6 +14,7 @@ noParkingSignStyle: vienna
 noStoppingSignStyle: vienna
 officialLanguages: [hr]
 orchardProduces: [grape, olive, walnut, hazelnut, apple, plum, cherry, orange, peach, pear, almond, chilli_pepper, sweet_pepper, fig, tomato, blueberry]
+parcelLockerBrand: [Paket24, GLS, BoxNow]
```
mnalis, line 17:
```suggestion
parcelLockerBrand: [Paket24, GLS, BoxNow, DPD]
```

there are a [lot of DPD parcel lockers in Croatia](https://www.google.com/maps/d/viewer?mid=1-SldYNLgTpq5qwH-qq7TLO1NAfBFTFQ&ll=44.862215647602454%2C15.653692520409685&z=8) (not many are currently tagged in OSM, though, which is why this quest would help here) and it is listed at [nsi.guide](https://nsi.guide/index.html?t=brands&k=amenity&v=parcel_locker&cc=hr).

FloEdelmann: 1. NSI suggests using "DPD Pickup Station" as brand (and "NSI" as operator).
2. I don't think it makes sense to manually change this file, as it will be overwritten anyway by the Gradle task before the next release.

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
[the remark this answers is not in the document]
FloEdelmann: Looks nice! Some suggestions/remarks:

* Try to use colors that have already been used in other quest icons. E.g. brown colors from the `railway.svg`, `picnic_table_cover.svg`, `leaf.svg` or `street_width.svg` could be used. If that doesn't look good, leave it as is though.
* Foreground objects should have a "shadow" (half-transparent, no blur, no spread, 2px straight down).
* Those cyan stripes look a bit weird. Try making them dark brown or removing them completely.
* If you used some icon, please add it to [`res/graphics/authors.txt`](https://github.com/streetcomplete/StreetComplete/blob/master/res/graphics/authors.txt).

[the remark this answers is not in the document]
Arthur-GYT: Tell me what you think ^^.

parcel_locker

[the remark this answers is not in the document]
FloEdelmann: I think the shadow looks a bit dark. Also, there are small green borders around the dark stripes at the bottom right.

But regardless, feel free to commit the file as is, it's better than the note quest icon already, and we can clean it up later. I can give it a go if you like :)

app/src/main/assets/country_metadata/HR.yml
matkoniecz: This is autogenerated - either some POIs in Croatia miss this info and should be mapped or Gradle task need to be changed or service used here has some data problems.

This file definitely should not be edited manually.

mnalis: > NSI suggests using "DPD Pickup Station" as brand (and "NSI" as operator).

You mean "DPD" as operator? Yes, I guess that was done to distinguish them. Either name would be fine with me...

> I don't think it makes sense to manually change this file, as it will be overwritten anyway by the Gradle task before the next release.

That is problematic then if it gets automatically overwritten, indeed. I guess the script ignores nsi.guide and relies on usage counts only? Could script be made to **_merge_** **both** "usage count" data **and** nsi.guide?  

The small annoyance is that it that might produce duplicates (e.g. theoretical _"DPD_ vs. _"DPD Pickup Station"_ in that example), but that seems better then not offering anything (especially when whole point of such quest is to add that data to the map!)
 
And "usage data" approach (sadly) produces much duplicates already (e.g. looking at ATM operators countrydata  we have in Croatia _"OTP Bank"_, _"OTP banka"_, _"OTP Banka"_, _"OTP"_, _"OTP banka d.d."_ etc.)

Perhaps NSI data could be put at the front of that merged list, as it more likely to be correct and more detailed (having `*:wikidata` etc). Should I open separate issue for that?

matkoniecz: > And "usage data" approach (sadly) produces much duplicates already (e.g. looking at ATM operators countrydata we have in Croatia "OTP Bank", "OTP banka", "OTP Banka", "OTP", "OTP banka d.d." etc.)

This is just symptom of fragmented data in OSM - if one version is obviously preferable then bot edit can be applied. (if there is a clear consensus in community then running edit itself is easy, feel free to ping me in the community thread - but it is also easily doable with JOSM or level0)

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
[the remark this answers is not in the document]
Arthur-GYT: It's added ^^.

app/src/main/assets/country_metadata/HR.yml
mnalis: @matkoniecz yeah it is quite easy to _actually do_ with overpass  & JOSM, it is just that the whole bureaucratic process of doing mechanical edits is so unappealing that I keep postponing it for [more two and a half years](https://github.com/streetcomplete/StreetComplete/pull/3352#issuecomment-939199676) now...

_And_ it does not solve this particular problem about parcel lockers operators, for which one would need to do an import (whose process is even more unappealing) or dedicate whole day or few targeted just to survey enough of them for script to pick them up (which is better used for other things)

Having the update script also **use (and ideally, prefer) NSI** in addition to taginfo usage would however be **much more preferable**, and would fix multiple of those issues (For example; I've already fixed ATM/banks in NSI: https://github.com/osmlab/name-suggestion-index/pull/9432, https://github.com/osmlab/name-suggestion-index/pull/9444, and improved parcel lockers in https://github.com/osmlab/name-suggestion-index/pull/9369) by itself...

matkoniecz: For simple regional ones it as simple as asking on community forum (maybe also some other channel) and waiting two weeks whether anyone objects.

(I make also wiki pages even for obvious edits but that is easy to just copy past and amend some specific cases, though many people admittedly skip this part - and as long as you as ask before doing edit this is not so terrible, though I would not advise it)

app/src/main/java/de/westnordost/streetcomplete/quests/parcel_locker_brand/AddParcelLockerBrand.kt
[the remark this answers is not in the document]
Arthur-GYT: My bad. It's pushed ^^.

app/src/main/assets/country_metadata/HR.yml
westnordost: Hm, this discussion is moot. At @matkoniecz noted:

> This is autogenerated

Nothing the PR author could do about without completely changing how this list is generated.

westnordost:
Well done! I have only a few small remarks.

app/src/main/res/values/strings.xml
```diff
@@ -1273,6 +1273,8 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="produce_brazil_nuts">Brazil nuts</string>
     <string name="produce_tung_nuts">Tung nuts</string>
 
+    <string name="quest_parcel_locker_brand">"What's the brand of this parcel locker?"</string>
```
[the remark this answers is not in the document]
FloEdelmann: It actually slipped my eyes :sweat_smile:

westnordost (approved):

FloEdelmann (approved):

res/graphics/authors.txt
```diff
@@ -366,6 +366,7 @@ quest/
     oneway.svg
     opening_hours.svg
     opening_hours_signed.svg
+    parcel_locker_brand.svg          Arthur-GYT, parts from https://emojipedia.org/ (License: CC0 1.0)
```
westnordost, line 369:
```suggestion
    parcel_locker_brand.svg          Arthur-GYT, parts from Google Noto Color Emoji: U+1F4E6
```

PR #5658, opened 2024-05-25
Add support for Jetpack compose version of quest Form

Arthur-GYT:
WIP

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
@@ -7,13 +7,8 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
 import androidx.annotation.AnyThread
-import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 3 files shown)


westnordost:
Uy, that's cool!

The quest form however, that's a quite central component. Are you sure you are not taking on something bigger than you can chew?

The [migration guide](https://developer.android.com/jetpack/compose/migrate/strategy) recommends a bottom-up approach for migration. So, taking another in fact quite large quest as an example - `AddMaxWeight`, the migration should be done in this order:

interop-animation


1. reimplement the different maxweight signs (maxweight, MGV, axle load, tandem axleload and its country-specific variants) in Jetpack Compose. This is already a large undertaking. It includes also reimplementing the sign illustrations because shape drawables are not supported by Jetpack Compose (instead, users should write this in `onDrawBehind` drawing blocks in code)

2. reimplement other components, such as the image select dialog in compose

3. then, one level up, reimplement the whole UI of that quest in compose. For this step, it is quite recommended to actually have a ViewModel in place, for the abstract quest form, too.

Without dabbling with ViewModels etc., I would recommend to for now only reimplement UI at the lower levels, i.e. nothing `Fragment` wide (nothing quest *`Form`-wide), but views and viewgroups used like "widgets" in these forms.

Would you be interested in doing that? In that case, I can look up what could be a good starter for this, to get into Jetpack Compose coding.

PR #5663, opened 2024-05-30
PowerPoleMaterial: Additional answer for power=terminal

### LINKED ISSUE #5657: Additional answer (power=terminal) for PowerPolesMaterial-Quest
### General

From time to time i come across notes like 'This is not a power pole; the line is anchored to the tower.' ([Example Note](https://www.openstreetmap.org/note/4255525))
So this is not power=pole but [power=terminal](https://wiki.openstreetmap.org/wiki/Tag%3Apower%3Dterminal)
Apparently, there are some mappers who record these as power poles. Therefore, I would like to suggest adding another option with a picture that sets power=terminal.

Possible CC0-Picture from OSM-Wiki

### Checklist
 --> Adds an answer that is currently not covered
- [X] 🐿️ Easily answerable by any pedestrian from the outside but a survey is necessary

### Ideas for implementation

**Element selection:** 

**Metadata needed:** 

**Proposed UI:**

qugebert:
Closes #5657 

Adds an answer in the 'Uhh' Menu "The power line is anchored to a building" which sets power=terminal instead of power=pole
Screenshot_20240530-140408

If anyone has a suggestion for a better description that's short enough to fit in the 'other answers' menu, feel free to share ;)


app/src/main/res/values/strings.xml
```diff
@@ -1332,6 +1332,7 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_powerPolesMaterial_wood">Wood</string>
     <string name="quest_powerPolesMaterial_metal">Steel</string>
     <string name="quest_powerPolesMaterial_concrete">Concrete</string>
+    <string name="quest_powerPolesMaterial_is_terminal">The power line is anchored to a building</string>
```
westnordost, line 1335:
```suggestion
    <string name="quest_powerPolesMaterial_is_terminal">Power line is anchored to building</string>
```

A little shorter

westnordost (changes requested):
Regarding implementation, please solve it like for e.g. the `AddPostboxRef` quest: Don't shoehorn the other answer into the regular PowerPolesMaterial enum. "terminal" is not a power poles material.

qugebert:
> Regarding implementation, please solve it like for e.g. the `AddPostboxRef` quest: Don't shoehorn the other answer into the regular PowerPolesMaterial enum. "terminal" is not a power poles material:
> 
> ```kotlin
> sealed interface PowerPolesMaterialAnswer
> enum class PowerPolesMaterial(...) : PowerPolesMaterialAnswer { ... }
> data object PowerLineAnchoredToBuilding : PowerPolesMaterialAnswer
> ```
Thanks for the hint.  I changed this now.

app/src/main/java/de/westnordost/streetcomplete/quests/powerpoles_material/PowerPolesMaterial.kt
westnordost, line 1: For consistency, this should now be renamed to the name of the sealed interface.

westnordost (commented):
Thanks! One more tiny thing, after that, it's perfect!

app/src/main/java/de/westnordost/streetcomplete/quests/powerpoles_material/PowerPolesMaterial.kt
qugebert: Ok, i renamed it

PR #5665, opened 2024-05-30
Don't shoehorn information=map in BoardType-Quest

qugebert:
For consistency reasons and because "map" is not a board type


riQQ:
`information=map` is also [discouraged in the wiki](https://wiki.openstreetmap.org/wiki/Key:board_type):
> board_type=map - Consider to use [tourism](https://wiki.openstreetmap.org/wiki/Key:tourism)=[information](https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dinformation) + [information](https://wiki.openstreetmap.org/wiki/Key:information)=[map](https://wiki.openstreetmap.org/wiki/Tag:information%3Dmap) for tourism maps.

westnordost:
@riQQ: typo? Anyway, irrelevant, StreetComplete does not tag `board_type=map`.

PR #5666, opened 2024-06-01
Add Drinking Water Fee Quest

Arthur-GYT:
All is in the title ^^.


matkoniecz:
In which areas drinkable water has decent chance to be paid? I never have seen one in Poland and Europe in general. And also in my (limited) trips outside Europe.

westnordost:
Never seen that either... Also, arguably, if the water costs something, then the question is whether it should really be tagged as `drinking_water` rather than as a vending machine that sells water.

Helium314:
I also never saw a paid drinking water... thing, but according to taginfo, 1% of all `amenity=drinking_water` is tagged with `fee=yes`

Helium314:
A good portion of the paid ones seem to be indoor or near caravan sites

matkoniecz:
> indoor or near caravan sites

and probably should be `access=customers` instead

matkoniecz:
There is another class: mineral water similar to https://uzdrowisko-iwonicz.com.pl/pijalnia-wod-mineralnych/ but these are exceedingly rare

PR #5672, opened 2024-06-04
Fix failing SeparateCyclewayCreator tests

wielandb:
This PR fixes some tests by re-writing them as the test condition only needing _some specific_ changes to occur opposed to all changes matching a predetermined set of changes.

It also changes the behaviour of applying ALLOWED_ON_FOOTWAY in the separate cycleway overlay to change some yes-like values to yes.


westnordost:
Great, thank you!

PR #5673, opened 2024-06-04
Add new quest for bench armrests

bompstable:
Resolves streetcomplete/StreetComplete#5664

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -31,6 +31,7 @@ import de.westnordost.streetcomplete.quests.barrier_type.AddBarrierType
 import de.westnordost.streetcomplete.quests.barrier_type.AddStileType
 import de.westnordost.streetcomplete.quests.bbq_fuel.AddBbqFuel
 import de.westnordost.streetcomplete.quests.bench_backrest.AddBenchBackrest
+import de.westnordost.streetcomplete.quests.bench_armrest.AddBenchArmrest
 import de.westnordost.streetcomplete.quests.bike_parking_capacity.AddBikeParkingCapacity
 import de.westnordost.streetcomplete.quests.bike_parking_cover.AddBikeParkingCover
 import de.westnordost.streetcomplete.quests.bike_parking_type.AddBikeParkingType
@@ -266,6 +267,7 @@ fun questTypeRegistry(
     9 to AddCarWashType(),
 
     10 to AddBenchBackrest(),
+    170 to AddBenchArmrest(),
     11 to AddAmenityCover(getFeature),
 
     12 to AddBridgeStructure(),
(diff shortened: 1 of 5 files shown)


westnordost:
Hm. I am still kind of on the edge whether this should be added. A few people on the forums seemed to be excited about mapping armrests, though. There's some time till the next feature release anyway, so it will change nothing if I leave this PR open for a while in which I can monitor the forums etc.

Regardless of whether it will be merged or not, if it is not merged, then I am quite sure it could be merged into SCEE, so here's a little review:
- each quest should have an own icon, so should this. Maybe... an arm?
- it seems unnecessary to me to provide the answer option that it is a picnic table, as before, the backrest quest has already been answered in which if it was a picnic table, it should have been selected, no?
- I find it unnecessary to have a clarification text. "any armrests" should be fine, no? The answer to that should be "yes" regardless whether there is one, two, three... armrests on that bench.

bompstable:
> Hm. I am still kind of on the edge whether this should be added.

I hope we can persuade you. :-) What would help? We should probably take this discussion to the issue though.

> * each quest should have an own icon, so should this. Maybe... an arm?

This not my skill area but I will see if anyone is interested in creating one.

> * it seems unnecessary to me to provide the answer option that it is a picnic table, as before, the backrest quest has already been answered in which if it was a picnic table, it should have been selected, no?

I was in two minds whether to include this and don't have strong opinions either way. My logic was that someone new could be doing the armrest quest and a reasonable amount of time could have passed since the backrest question was answered. It's possible that someone answered backrest=yes on a bench without mentioning that it's actually a picnic table so this is another opportunity for data clean up.

In the action for this answer I did consider dropping the backrest tag but `backrest` is [in use](https://taginfo.openstreetmap.org/tags/leisure=picnic_table#combinations) and [documented ](https://wiki.openstreetmap.org/wiki/Tag:leisure=picnic_table)on `leisure=picnic_table`. If someone has said there is a backrest, this still applies if its actually a picnic table.

> * I find it unnecessary to have a clarification text. "any armrests" should be fine, no? The answer to that should be "yes" regardless whether there is one, two, three... armrests on that bench.

I want to make sure that someone who is unsure how to answer can get a little bit more help. There are some unusual armrest combinations out there - for example wheelchair accessible benches with a single armrest at one end. I think someone could be confused as to whether "any armrests" applies to such a bench.

westnordost:
I contemplated this PR for a long time and actually discussed this with several people but I ultimately decided that it should not be a quest in StreetComplete, for now. (Also, a dedicated icon is missing.)

I am sorry that it took so long to decide, but it could maybe fit nicely into SCEE, so hopefully no work is lost.

The main reason is that I still doubt its usefulness even after you posted those UK mobility guides in #5664. I summarized what these guides say about armrests and commented it in https://github.com/streetcomplete/StreetComplete/issues/5664#issuecomment-2142710355.

I understand how armrests on a chair or a bench where every seat has armrests can help really stiff and weak people lower themselves onto the seat, as they can use their arms to support kneeling down. But this is not what is asked and tagged here. It is asked whether a **bench** has ***any*** armrest.
An armrest on one side (the other is too far away on a bench) only is not really a help. (I tried it a bit) On the contrary, it might be even easier to slide onto the bench from the side if there is *no* armrest, especially if you have a rollator or walking stick.

Finally, in case I didn't bring this up before, I have issues with accepting "it is for statistics" as a valid 🚧 _useful purpose_ on it's own. You can do statistics with a lot of things. But some statistics can also end up not being meaningful, if e.g. the data chosen is not meaningful. And this is the point I have issues with, that the data raised is of little value and not that meaningful (see paragraph above).

PR #5677, opened 2024-06-05
Resolve button text confusion

### LINKED ISSUE #5671: Text used for quest_generic_otherAnswers2 can be mistaken for an acronym and causes confusion
Quests that use the `quest_generic_otherAnswers2` display answers such as "UH…", "ER…" "OH…". To new users, it is not always obvious that these are actually words and furthermore can sometimes be mistaken for acronyms. It is therefore not clear that people are meant to select them when they are not sure what the answer is.

For example, a friend thought that the EN-GB value of "ER..." was actually short for "Error" and thought it meant that something had gone wrong within the app. This was caused by the use of uppercase text and the three dots being misinterpreted as representing a cut off message rather than a person hesitating to answer the question.

Whilst this issue can be partly improved through better translations, the issue of the text not looking like natural words is caused by the fact that  the answer text on quests such as bench backrests, postbox collection times, bus stop shelters etc is upper-cased even though the original strings use mixed case. Presenting "Uh..." instead of "UH..." would make it easier for users to read the strings as they are intended. Furthermore, using "Yes" and "No" instead of "YES" and "NO" for the other answers would make them easier to read and recognise. In general, mixed case is easier and quicker to read than uppercase and is arguably more attractive.

I think the answer options should be presented in their original case rather than being converted to upper case before being displayed.

bompstable:
- Use mixed-case in all application buttons
- Use Uh... instead of Er... in en-GB to avoid confusion with "Error"

Resolves #5671

### DIFF
--- a/app/src/main/res/values-en-rGB/strings.xml
+++ b/app/src/main/res/values-en-rGB/strings.xml
@@ -1385,7 +1385,7 @@ Additionally, some overlays allow you to add new data at the position of a displ
 You can find donation information on our website or project home, Google Play does not allow us to provide the link within the app."</string>
     <string name="quest_isAmenityIndoor_outside_covered">"Not inside, but covered"</string>
     <string name="lit_value_unsupported">"Specific lit status that is not selectable in this app"</string>
-    <string name="quest_generic_otherAnswers2">"Er…"</string>
+    <string name="quest_generic_otherAnswers2">"Uh…"</string>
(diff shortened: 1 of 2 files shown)


westnordost:
Great! So, is this ready for review / merge?

bompstable:
Sure. What about commit 2beec1b, though? Shall I just drop it?

westnordost:
You can, but you don't have to.

PR #5682, opened 2024-06-10
Optimization of elementFilter in AddAcceptsCards quest

jmizv:
The filter is already checking that the tag `brand` is *not* set, so the check for having it in line 23 is probably obsolete.


matkoniecz:
Currently it is not adding anything and probably consumes some CPU time.

I guess it could have its place in case we stop filtering out branded ones, or if someone will base its code on this query...

westnordost:
The performance impact is practically zero. But well, 🤷‍♂️, no point in arguing in over something that tiny.

PR #5686, opened 2024-06-13
Multiplatform OSM API client

### LINKED ISSUE #5410: Multiplatform OSM API client
Requires #5408 to be done first.

The [OSM API](https://wiki.openstreetmap.org/wiki/API_v0.6) is a HTTP REST API that communicates chiefly in XML. This must be ported to Kotlin multiplatform. 

Currently, the Java library [`de.westnordost.osmapi`](https://github.com/westnordost/osmapi) is used for any communication with the OSM API 0.6, i.e. uploading map data, notes and gpx tracks, downloading map data, notes, user data, etc.

The dependency to `osmapi` must be replaced with an implementation based on

- ideally the http client from #5408 and

- a multiplatform XML parser / writer, which is stable, well-tested and ideally (semi-)official / well-maintained and *ideally* but not necessarily compatible with `kotlinx-serialization`. This is also important for other XML parsing in the app, i.e. #5369

- using `kotlinx-datetime` to parse timestamps (already a dependency)

See interfaces [`TracksApi`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+TracksApi.kt&type=code), [`MapDataApi`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+MapDataApi.kt&type=code) and [`NotesApi`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+MapDataApi.kt&type=code)

### XML Parser / Writer requirements

- multiplatform support for at least Java/Android and iOS

- stable, well-tested and ideally (semi-)official  / well-maintained

#### xmlutil

There is currently one xml serializer compatible with kotlinx-serialization: [`io.github.pdvrieze.xmlutil`](https://github.com/pdvrieze/xmlutil). It is also mentioned in the Ktor docs, i.e. is the XML serializer/deserializer used in Ktor.

I didn't find any other multiplatform XML parsers/writers.

westnordost:
Resolves #5410

---

Also took the opportunity to make the implementations and naming for all the clients that use the multiplatform Ktor-Client for Http (which is everything now) consistent.

---

I did a bit of performance testing. Unfortunately, the results are **very** unsatisfying. 

### Downloading city center of Hamburg (~35000 elements)

(Done on a Samsung S10e, release APK)

performance

**Edit: The download+parse time for ktor+xmlutil has now been reduced to about 5.2s for this particular test**

The previously used Java library osmapi...
- uses Java's [HttpUrlConnection](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html) to do the Http request
- the response [InputStream](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html) is consumed by a [streaming parser](https://developer.android.com/reference/org/xmlpull/v1/XmlPullParser) which produces one OSM element after another
- which in turn each is copied into the app's own data structure 

With this PR, we use
- Ktor-client with CIO engine  
- response body is copied into a string
- the string is parsed by [xmlutil](https://github.com/pdvrieze/xmlutil) into a (DOM) data structure
- we copy the data structure into our own
 
It is clear that the second approach must be slower, especially for large data, because we don't stream anything (not really supported by ktor and xmlutil **just yet**) but always copy the whole data. On the other hand, getting the response body as a string (rather than as a stream) and copying all the data in the end seems to be not be the cause for why it is so slow.
I am (even more) surprised how slow Ktor-client (and/or CIO) is. I see on the log that the garbage collector is quite busy. What is it doing?


westnordost:
I tested with other engines for ktor-client:
- CIO: ~7s
- Android: ~4s
- OkHttp: ~3s

That's still slower than before, but a measurable difference. On iOS, another engine could be used.

westnordost:
And even though I marked it as blocked, you're welcome to review it! It is not a work-in-progress, it is done.

westnordost:
Hmm, with a minimal ... uh... "benchmark", I can not reproduce the issue with Ktor client.

```kotlin
import io.ktor.client.*
import io.ktor.client.engine.cio.*
import io.ktor.client.request.*
import io.ktor.client.statement.*
import kotlinx.coroutines.runBlocking
import java.net.HttpURLConnection
import java.net.URL

fun main() {
    ktorClientCio()
    httpUrlConnection()
}

fun ktorClientCio() = runBlocking {
    val client = HttpClient(CIO)
    print("Ktor-client (CIO): ")
    val t = System.currentTimeMillis()
    val response = client.get(url)
    val text2 = response.bodyAsText()
    println("Read ${text2.length} bytes in ${System.currentTimeMillis() - t} ms")
}

fun httpUrlConnection() {
    print("HttpUrlConnection: ")
    val t = System.currentTimeMillis()
    val connection = (URL(url).openConnection() as HttpURLConnection)
    val inputStream = connection.inputStream
    val text = inputStream.bufferedReader().readText()
    inputStream.close()
    println("Read ${text.length} bytes in ${System.currentTimeMillis() - t} ms")
}

val url = "https://api.openstreetmap.org/api/0.6/map?bbox=9.98,53.53,10.0,53.55"
```

Or rather, the times fluctuate so much that it looks like it is rather caused by how long the server reply takes.

And at least, when comparing reading each all incoming bytes into a string. (I guess, at least theoretically, if the parsing does not take longer than the transmission over the wire, parsing the stream into the XML is "free" in the sense as it can take place in parallel to the data arriving.)

westnordost:
Tried (JVM only) streaming while keeping kotlinx-serialization into data structure to see if it would improve anything:

```diff
-    fun parseMapData(osmXml: String, ignoreRelationTypes: Set<String?>): MutableMapData =
-        xml.decodeFromString<ApiOsm>(osmXml).toMapData(ignoreRelationTypes)
+    fun parseMapData(osmXml: ByteReadChannel, ignoreRelationTypes: Set<String?>): MutableMapData =
+        xml.decodeFromReader<ApiOsm>(xmlStreaming.newReader(osmXml.toInputStream().bufferedReader())).toMapData(ignoreRelationTypes)
```

Unfortunately, it still consistently takes about 7.5 seconds for the data tested with, about one second slower actually as when I decode from string.

(The data tested with is the response from `https://api.openstreetmap.org/api/0.6/map?bbox=9.9920654,53.5435716,10.0085449,53.5566260`)

westnordost:
Posted about that the byte stream does not seem to become available before it has been fully read in the ktor issue tracker. Let's see what they say. https://youtrack.jetbrains.com/issue/KTOR-7103/

westnordost:
Implemented manual parsing/serialization using xmlutil's XmlParser / XmlWriter now (it is actually less code) and replaced the CIO engine with the platform default (which uses HttpUrlConnection under the hood).

Download and parsing for that particular test is now down ~5.2s as opposed to ~2.7s. And this is caused, I am pretty sure, by the thing I stated in my last comment.

The rest of the download task (download and parse notes, create quests, persist all the data) for that particular test takes about ~4.5s, so I think a performance loss of 25% or so on download is bearable. However, there is no rush to merge this, I will still wait for KTOR-7103 and see what they say. Maybe it is just a misconfiguration on my part.

westnordost:
I was informed that Ktor is working as intended and streaming is possible via another interface documented [here](https://ktor.io/docs/client-responses.html#streaming). For that to work (multiplatform) properly, I'll have to wait first that both xmlutil and ktor are based on the same IO library (kotlinx-io).

Helium314:
I just did a comparison with my old S4 Mini, instead of the A3 I had with me the past few days.
for 57k elements, dowload is about 25 s on `master`, and 52 s on `osmapi`. Difference is much bigger than on A3.

westnordost:
Thanks for testing! In debug mode?

Helium314:
Yes, debug mode.

westnordost:
>  I'll have to wait first that both xmlutil and ktor are based on the same IO library (kotlinx-io).

Beta of ktor is now based on kotlinx-io. However, it does not expose any `Source` that could be parsed by parsers. Not sure how they intend their interface should be used, so I asked https://youtrack.jetbrains.com/issue/KTOR-7224

pdvrieze:
I had a look at the parsing speed with the optimized library (11813680 bytes read). When parsing the document from a string it parses in 351ms. When reading from inputstream it parses in 2000ms (2829ms on first parse). The manual parsing took about 200ms from a string). Obviously the manual parsing does less (it doesn't check unexpected items) - btw. it is better to loop through the attribute indices rather than look them up by name - the latter is repetitive.

pdvrieze:
I checked out your project, but wrote my own little test class. Reading from inputstream appears to be network constrained. What I mean is yes, that pseudocode should be faster (although somewhat lax, it doesn't check namespaces nor duplication).

westnordost:
> What I mean is yes, that pseudocode should be faster

...because there are less iterations. I rewrote it and tested the performance. There is almost no difference, though, actually, it is about 10% slower.

My guess why this is marginally slower is because for every element parsed, I have to initialize a few new local variables, these are allocated on the heap, which is not for free. (See pseudo-code above).

westnordost:
I intend to merge this now into v59.0.

I have been waiting on this because I am assuming that download+parsing speeds could be improved by letting the XML parsing library (xmlutils) parse the stream of bytes as it arrives through the network, i.e. do the parsing in parallel to the IO and not after. Currently, in this branch, first al bytes are streamed into a string, then that string is parsed. This is potentially slow for large data.

There is no Kotlin Multiplatform replacement for Java's `InputStream` in the standard library. The light at the end of the tunnel is `Source` from the kotlinx-io multiplatform library, which is just that. kotlinx-io is on its way to be stabilized and then form part of the set of official standard libraries maintained by JetBrains (creators of Kotlin).

For this to work, both xmlutil needs to be able to consume `Source` and Ktor (multiplatform HTTP library) needs to produce a `Source` representing the HTTP response. Ktor now migrated to use kotlinx-io under the hood, so things were looking promising.

However, long story short, Ktor uses kotlinx-io only under the hood, i.e. it does not expose a `Source` that could be consumed by a parser on its API. This is deliberate, as was explained to me in issue [KTOR-7224](https://youtrack.jetbrains.com/issue/KTOR-7224), as Ktor's API is suspending while `Source` is a blocking API.
I posted a comment in [kotlinx-io#163](https://github.com/Kotlin/kotlinx-io/issues/163), the ticket in kotlinx-io in which a possible future async API is discussed, suggesting Ktor devs exchange ideas with kotlinx-io devs.  But I don't think things will move quickly there, as an async API seems to be still somewhat in the brainstorm phase and I suspect that their priorities are first to stabilize the API that's already there.
And this is the reason why I merge this now, cause this could otherwise just as well lie around for another year and generate maintenance work (merging, keeping it up to date in parallel to current code).

The caveat of using this code now, is, as I previously mentioned, that there is a performance loss for downloads of large areas. Since part of the download process is also the persisting in the database and the creation of quests, it comes down to a performance loss of up to 25%.

westnordost:
https://youtrack.jetbrains.com/issue/KTOR-7327/Support-conversion-between-byte-channel-interfaces-and-kotlinx-io-primitives

PR #5693, opened 2024-06-17
Replace Tangram with Maplibre

westnordost:
by @Helium314 and me.

This PR is the same as https://github.com/Helium314/SCEE/pull/516 only that it merges to vanilla StreetComplete.

#### Visible differences to Tangram-ES
- improved style, e.g. private roads, rails and paths look better now, some issues with the old style were solved and bridges / tunnels render more correct now
- can zoom in much further
- quest dots are clustered on low zoom
- no 3D building except in the buildings overlay. (They don't look so good with MapLibre)
- icons in building overlay have no white outline anymore (technical limitation)
- color of icons in places and things overlay adapts to day/night mode
- color of text in overlays adapt to day/night mode
- animation on selecting (quest) pin
- (better) animation on selecting elements in an overlay
- improved memory consumption
- location accuracy circle doesn't get blurry when it is large
- (of course, no tangram bugs. For example those flickering quest pins when zooming back out, that was horrible)
- (... anything more?)

#### Blocked by
- https://github.com/maplibre/maplibre-native/issues/2410
- (https://github.com/maplibre/maplibre-native/issues/2206)

#### Known bugs
- location pointer pin sometimes points to the wrong direction, caused by
  https://github.com/maplibre/maplibre-native/issues/2255
- sometimes crashes in the background (i.e. not noticed by user), caused by
  https://github.com/maplibre/maplibre-native/issues/2432
- steps dashes and private-dashes look a bit funny while zooming, caused by
  https://github.com/maplibre/maplibre-native/issues/2433

#### Remarks
- displaying overlay and quests in the currently visible map section is implemented manually. Could potentially be removed when https://github.com/maplibre/maplibre-native/issues/2262 is fixed
- location dot, accuracy circle and view direction has been implemented manually because MapLibre's own LocationComponent didn't look good and felt jerky. If that changes in the future, the own rendering could be replaced


Helium314:
Regarding the track dots
> But anyway, if it looks weird that way when zooming

You can give it a try, but I think the dot / dashes change is not really an issue.

westnordost:
> dot / dashes change is not really an issue.

👍

Helium314:
I noticed that user track is hidden by overlays (e.g. roads in surface overlay). @westnordost you probably have a better overview where to place the layer to make it work properly.

westnordost:
It shouldn't be hidden by the overlay?

Helium314:
I'll check how it is currently, but I never noticed this.

westnordost:
Don't mind how it is currently, more important how it **should** be.

westnordost:
Would it be possible to completely remove the map cache size option or would the cache then grow without borders? Contemplating this setting, it would actually be better if the map tiles were more closely tied to the OSM data downloaded. I.e., the map tiles for any given bbox should **always** be available as long as the OSM data for it is. In other words, the map tiles data should only be cleared (after) when the OSM data is cleared (after two weeks).

Helium314:
> In other words, the map tiles data should only be cleared (after) when the OSM data is cleared (after two weeks).

This is already happening. When map tiles are part of an `OfflineRegion`, they will not be deleted as long as the `OfflineRegion` exists. Offline regions stay until they are removed, either by [cleanOldOfflineRegions](https://github.com/Helium314/SCEE/blob/maplibre/app/src/main/java/de/westnordost/streetcomplete/screens/main/map/MapFragment.kt#L115-L122) or by [clearing the tiles database](https://github.com/Helium314/SCEE/blob/maplibre/app/src/main/java/de/westnordost/streetcomplete/data/maptiles/MapTilesDownloader.kt#L33).

The cache size setting can only lead to removal of tiles that are not part of an `OfflineRegion`. These tiles either were downloaded by panning the map outside a downloaded region, or they were part of an `OfflineRegion` that has already been removed.

When removing the cache setting, the default limit of 50 MB is used (and to clarify: when having many tiles in `OfflineRegions`, it's possible to exceed this limit).

Helium314:
> Don't mind how it is currently, more important how it **should** be.

I don't know how it should be, I was just irritated by a road hiding the track.

westnordost:
"irritated" is a bit of a false friend in German -> English, by the way. You were confused, or were you annoyed?

Anyway, I see, so since anything beyond offline regions is just used for browsing (to find which part to download) it can be very small, or at least, the user should not have to worry about it. In this case, the setting can be removed.

Helium314:
Rather confused, probably even though it's an _over_lay, it looks more like a (colored) _background_ map.

westnordost:
I see, anyway, I will look into it now.

westnordost:
The PR that maybe solves https://github.com/maplibre/maplibre-native/issues/2410 has been merged and a new version was released. Would you test if you can still reproduce it now? I'd love to finally merge this to master.

(I tested whether it solves https://github.com/maplibre/maplibre-native/issues/2372 - it doesn't. But in an emulator, I was never able to reproduce that also e.g. the cycleway overlay would not display in certain tiles, so maybe (hopefully) it really *is* a different issue.)

westnordost:
Also, when I first started it with the new version, it crashed with. I dimly remember that this was mentioned before, but I thought that we solved it?

```
Process: de.westnordost.streetcomplete.debug, PID: 10770
java.lang.IllegalStateException: Already resumed
	at kotlin.coroutines.SafeContinuation.resumeWith(SafeContinuationJvm.kt:44)
	at de.westnordost.streetcomplete.screens.main.map.maplibre.OfflineManagerKt$awaitDownload$2$1.onStatusChanged(OfflineManager.kt:84)
	at org.maplibre.android.offline.OfflineRegion$setObserver$1.onStatusChanged$lambda$0(OfflineRegion.kt:249)
	at org.maplibre.android.offline.OfflineRegion$setObserver$1.$r8$lambda$ppSc6Vjf3zgbrFuf7K1etl42M0w(Unknown Source:0)
	at org.maplibre.android.offline.OfflineRegion$setObserver$1$$ExternalSyntheticLambda2.run(D8$$SyntheticClass:0)
	at android.os.Handler.handleCallback(Handler.java:958)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:205)
	at android.os.Looper.loop(Looper.java:294)
	at android.app.ActivityThread.main(ActivityThread.java:8177)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)
```

Helium314:
> but I thought that we solved it?

I also thought so. Maybe there are other ways to resume twice?

Helium314:
> Would you test if you can still reproduce it now?

Still happening, unfortunately.

westnordost:
Darn it!

westnordost:
@Helium314 As per suggestion by a maplibre developer, I attempted a workaround for the blocking issue maplibre-native#2410 in the branch `maplibre-workaround` - always remove all layers and source and re-add it whenever a new geojson is set. Would you test if this circumvents the issue?

Helium314:
When switching from address overlay to some other overlay, I can reliably trigger that bug.

Other issues:
No clusters are displayed when zooming out.
No addresses are shown in address overlay.
Places overlay doesn't show any places (but things overlay works).

**very interesting**: everything works after starting the app once while connected to the internet.
I'll still try to reproduce that bug by switching between overlays more often, but this hidden internet dependency could explain why it sometimes works for me and sometimes not.
Though I'm not sure why (whatever is needed) isn't always downloaded together with map data. Or maybe it has only a very short cache lifetime?

Also got a crash first time switching overlays (never again afterwards)
```
java.util.ConcurrentModificationException
    at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:757)
    at java.util.LinkedHashMap$LinkedValueIterator.next(LinkedHashMap.java:785)
    at de.westnordost.streetcomplete.screens.main.map.components.StyleableOverlayMapComponent.set(StyleableOverlayMapComponent.kt:458)
    at de.westnordost.streetcomplete.screens.main.map.StyleableOverlayManager.setStyledElements(StyleableOverlayManager.kt:167)
    at de.westnordost.streetcomplete.screens.main.map.StyleableOverlayManager.access$setStyledElements(StyleableOverlayManager.kt:33)
    at de.westnordost.streetcomplete.screens.main.map.StyleableOverlayManager$setStyledElements$1.invokeSuspend(Unknown Source:15)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:811)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:715)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:702)
    Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@bc50a8e, Dispatchers.Default]
```

Helium314:
Possibly the bug is related to this thing I had in the original PR:

> * MapLibre always does some downloads right after start. Possibly glyphs, considering that when you start SC without WiFi there is a message `This request was cancelled (https://api.jawg.io/glyphs/Roboto%20Regular%2cNoto%20Regular/0-255.pbf). This is expected for tiles that were being prefetched but are no longer needed for the map to render.`. It's not horrible, but downloading the same thing on every start seems like a waste of bandwidth and resources.
>   * maybe goes away when properly pre-loading tiles for offline use

westnordost:
So, for the record, the issue is still reproducible with `maplibre-workaround`, right? On the contrary, it triggers some other issues that may or may not be additional bugs either because the workaround is not perfect or because of other bugs in MapLibre?

Helium314:
> So, for the record, the issue is still reproducible with `maplibre-workaround`, right?

Yes

> On the contrary, it triggers some other issues that may or may not be additional bugs either because the workaround is not perfect or because of other bugs in MapLibre?

Here I'm not sure. I remember at least the clustering not working once or twice (but most of the testing is done on high zoom, so possibly I just didn't notice). Now that I noticed the connection to being offline, I can better re-check on the `maplibre` branch.

Helium314:
One more test (`maplibre` branch), I did everything here sequentially (did _not_ yet test what are the actually necessary parts to get into a broken state):
1. reset app
2. start when offline
3. wait some time (minutes)
4. go online
5. zoom to city
6. download data
7. go offline
8. clustering looks fine
9. overlays work fine (_all_ overlays, and switching between them)
10. zoom out, move, zoom in
11. switch to address overlay -> no addresses
    * that broken overlay bug triggers, e.g. when switching to sidewalk overlay
    * tile is still broken when switching overlay on broken zoom level
    * zoom in or out (more than 1 zoom level) and change overlay to make it look like it works
    * but changing overlay does not work properly from now on, the bug comes back again
    * **I was able to fix it by going online, zoom out and change to address overlay**

My first guess: there is some necessary data that is not downloaded (but then why did the address overlay work initially?)

westnordost:
Just to understand. You write "all overlays work fine", but then you say that when switching to Address overlay, it stops working. So you mean all overlays except the address overlay?

(I will also try to reproduce the issue given your recipe)

Helium314:
> Just to understand. You write "all overlays work fine", but then you say that when switching to Address overlay, it stops working. So you mean all overlays except the address overlay?

The address overlay is included here, it really did work initially. For whatever weird reason, after zooming out (iirc lit overlay was active), moving to a different (already downloaded) area, and zooming in, switching to address overlay did not show addresses any more.

westnordost:
Wow, I can reproduce this! So, this is not device-specific at all!

(And jesus, my phone turned into a heater... I hope it is just the debug mode and/or the issue that is triggering this)

westnordost:
For me, the address overlay did not show up at all. Switching back to other overlays produced the effect described in the issue. This seems to only happen on the first start of the app, i.e. when I remove the app from recents and start again, even if it is still offline, everything looks fine. Right?

Helium314:
> For me, the address overlay did not show up at all.

I just tried again, and for me the addresses showed up, but only in the tile where I first enabled an overlay. (and no zoom out was necessary)

westnordost:
Since it was suspected that the issue has something to do with MapLibre trying to re-load already loaded glyphs while being offline, and that this is triggered by JawgMaps not including a `control-cache` header to denote the lifetime of the glyphs, I asked @Joxit of JawgMaps if he could add that header to not only reduce the number of requests in general but also to help us either workaround the issue or at least find the source of it.

And he did that, with a lifetime of 12 hours ♥

Is the issue still reproducible with this change?

Helium314:
> Is the issue still reproducible with this change?

I managed to reproduce it by downloading data, going offline and switching to the address overlay. Looks like some of the resources necessary for address overlay are not downloaded during the initial download.
After going online once and enabling address overlay, the issue was not reproducible so far (not even when restarting the app).

westnordost:
Is MapLibre still trying to download more glyphs when offline? (It should appear in the log)
Because now, I would say, it shouldn't, because it should recognize that the glyphs should not be considered stale at least for 12 hours.

Helium314:
I was unable to reproduce it until I noticed the problem was actually switching to the places overlay, which also seems to require glyphs that are not pre-downloaded.

Is there a way to make sure that overlay texts only use glyphs that are already downloaded together with tiles? Or maybe MapLibre should use some fallback instead of (whatever causes this bug)?

westnordost:
Hmm... perhaps it is possible to have all the fonts locally already. As long as one doesn't go wild with different fonts and font weights, it is not that much.

Not sure how to refer to them in the style JSON though.

They can be downloaded here: https://github.com/kylebarron/openmaptiles-fonts/tree/master/fonts

(Otherwise, ) does MapLibre download them as files into the app cache directory, or again into some database? If the former, it should be possible to download them all beforehand manually (like we did with tangram tiles).

Helium314:
> **Edit**: Could try with `asset://...`, apparently it works for sprites, maybe it does work for glyphs, too.

This works nicely, I wasn't able to reproduce the bug so far when having glyphs in assets!

------
One more issue I noticed: Sometimes switching between overlays takes a second or so. When I select an element of the old overlay before the new map features are set, I can e.g. set the address of a waste bin. It should be better for most users due to faster phone and non-debug version, but still should not happen.
Perhaps we should clear the overlay before doing the switch, or give each feature some `overlay` property that is checked on selecting the feature.

westnordost:
> This works nicely, I wasn't able to reproduce the bug so far when having glyphs in assets!

Hm! 

🎉?

Is this the workaround we have been looking for?

(Is it possible to also specify the streetcomplete.json itself to sit in the assets folder? Currently for offline download, we have to have this remote duplicate of the streetcomplete.json)

---

Regarding the issue: Well, sounds like that would make sense. (The first, if it doesn't work, the second)

westnordost:
>> Is this the workaround we have been looking for all this time?

> Really looks like it is. So I'll put the fonts to assets.

But then I have to change the remove streetcomplete.json, too. I hope that works with the offline downloader

Helium314:
Btw the glyph folder names are a little weird, e.g. `Roboto Bold,Noto Bold`. MapLibre appears to require it due to `"text-font": ["Roboto Bold", "Noto Bold"]` in the json.
`Open Sans Regular,Arial Unicode MS Regular` is accessed for address and places overlays only. This appears to be some default. We could try switching to Roboto/Noto here, then the `Open Sans Regular,Arial Unicode MS Regular` might not be necessary any more.

westnordost:
Yeah, or just one font, so that we don't need to keep around font fallbacks. After all, we do keep the fonts ourselves, no need for fallbacks.

Helium314:
> We could try switching to Roboto/Noto here

Current state (Open Sans Regular) on the left, Roboto Regular on the right:
fonts
I think switching to Roboto is fine, also for increased consistency e.g. with street names (different style of _g_).

> Yeah, or just one font, so that we don't need to keep around font fallbacks. After all, we do keep the fonts ourselves, no need for fallbacks.

So I'll just use Roboto Bold and Roboto Regular and remove the Noto names.
Also will set overlay text font to Roboto Regular and remove Open Sans regular folder.

Helium314:
So I guess it's time for a little more testing now, and then... I'm not sure that https://github.com/maplibre/maplibre-native/issues/2206 will eventually get fixed. Maybe we should merge this regardless?

app/src/main/java/de/westnordost/streetcomplete/screens/main/map/components/StyleableOverlayMapComponent.kt
```diff
@@ -184,6 +184,7 @@ class StyleableOverlayMapComponent(
             .withProperties(
                 iconImage(get("icon")),
                 textField(get("label")),
+                textFont(arrayOf("Roboto Regular")),
```
westnordost, line 187: You had to write this because MapLibre uses some font by default?

Helium314: Yes, the default is `Open Sans Regular` and `Arial Unicode MS Regular`

westnordost:
Yes, let's do it.

westnordost:
And I'll document our findings in the respective tickets.

Helium314:
I did a "live test" and noticed a few minor visual glitches. Will open issues, or do PRs/commits depending on fixes I find. (but first I'll probably spend 2 days or so on merging this into SCEE)

westnordost:
One thing that just came to my mind with all this offline region gehampel, are we maybe massively producing overlapping offline regions by just downloading and hence producing massive amounts of (unused / doubly used) cached map tiles? E.g. just when opening the app, the app will already download two regions (small rectangle around the user, then big rectangle around the user).

Since offline regions seemed to have never been thought through completely (https://github.com/maplibre/maplibre-native/issues/2724), maybe there are issues with overlapping and managing (many) offline regions, too.

Helium314:
As far as I remember from looking through the offline database, tiles are downloaded only once. Tiles and regions are linked in some db table, and a single tile may belong to more than one offline region.

Thus I think overlapping offline regions are not an issue. The small-rectangle region will be useless in the db, but only occupy very little space (download url, some dates, which tiles belong to this region).

Coehill:
Thank you to the developers for all of your efforts on making this available on iOS! StreetComplete is an invaluable tool and we will all benefit from more people able to map with it.

PR #5694, opened 2024-06-17
Treat `cycleway=opposite*` as invalid

westnordost:
Can be merged as soon as https://wiki.openstreetmap.org/wiki/Proposal:Deprecate_cycleway%3Dopposite_family is approved.

**This causes:**
- roads with such tagging to be displayed as red in the overlay
- roads with such tagging as immediately due for (re)survey

When the user answers that a road has no cycleway for the contra-flow side in a oneway but it was tagged with any `cycleway=opposite*` before, there will still be a confirmation dialog seeking confirmation by the user that there is indeed no bicycle contra-flow allowed.


FloEdelmann (approved):
Looks good. Nice to see so much removed/simplified code :slightly_smiling_face:

westnordost:
Heh, you see, it is very little less code, most of it is just removing some test code. The test code is so large because it tests every single permutation.

PR #5706, opened 2024-06-24
WIP: test(KMP): use mockative instead of mockito

riQQ:
Preparation for the move to KMP.

~150 tests are still failing. Some are blocked by bugs / missing features in mockative, but maybe some or all of the failing tests can be changed to work around the issue or to just don't mocks.


westnordost:
I see you are using some plugin to "open" all classes in Kotlin so that they can be extended(? - or whatever magic it is doing) by the mock framework. Does this plugin have an effect on production code? If yes, could it be made so that it doesn't? (It seems ugly to me if production code is automatically modified for tests)

westnordost:
Also, I read somewhere that (at least) JavaScript does not support the <code>fun \`my quoted function name\`()</code> syntax and therefore this syntax cannot be used in multiplatform projects. I am not sure if this also applies if one does not target JS at all. But if it does, I guess all the nicely readable test functions would need to be renamed from <code>fun \`my cool function name\`()</code> to <code>fun my_lame_function()</code> (or worse, camel case, but 🤮)

westnordost:
Okay, sorry, I just glanced over it really quick so far.

westnordost:
(There are conflicts with master and the PR is marked as draft, that's why I didn't look at it in detail yet. In case you need feedback about something, let me know.)

riQQ:
> Also, I read somewhere that (at least) JavaScript does not support the `` fun `my quoted function name`() `` syntax and therefore this syntax cannot be used in multiplatform projects. I am not sure if this also applies if one does not target JS at all. But if it does, I guess all the nicely readable test functions would need to be renamed from `` fun `my cool function name`() `` to `fun my_lame_function_name()` (or worse, `camelCase`, but ...🤮)

I've seen this error at some point, I think when writing a test for a Mockative bug. So you're correct that we'll have to rename the methods when switching to the KMP project structure.

riQQ:
> (There are conflicts with master and the PR is marked as draft, that's why I didn't look at it in detail yet. In case you need feedback about something, let me know.)

Yes, that's why this PR is still a draft. I don't want to put in the work if Mockative doesn't fix / implement everything that is required for the tests here. Progress has slowed down there so I'm waiting, too.

riQQ:
Mockative is looking for a new maintainer, so this PR depends on Mockative getting a new maintainer.

See mockative/mockative#129.

TSampley:
What a coincidence! I just joined the OpenStreetMap site last week as I'm looking for non-corporate alternatives to Google maps! I just saw the bad news about Mockkative 😔 - I hope I can support this nifty effort!

westnordost:
Recently, a lot of tests are failing for me (on master), somethig something with mocks and mockito, even though I didn't even touch the mockito dependency version. So maybe with mockito this is already the same problem as with mockative, using K2 compiler. Do you (guys) also have this problem on master or is it just an issue with my setup?

westnordost:
hm, okay ¯\_(ツ)_/¯

riQQ:
[Mockative V3.0.0](https://github.com/mockative/mockative/releases/tag/v3.0.0) with support for Kotlin 2 has been released recently.

There is a [migration guide to V3](https://github.com/mockative/mockative/blob/main/wiki/MIGRATING-TO-3.md).

riQQ:
> (Say, do you know Clonk?)

No, why do you ask?

westnordost:
I thought I saw your name somewhere, but maybe it was someone else or I misremember

westnordost:
Will then need to decide whether to use [Mockative](https://github.com/mockative/mockative) or [Mokkery](https://github.com/lupuuss/Mokkery).

Issue #5678, opened 2024-06-06
Barrier=gate - What is the available opening


Nick-Tallguy:
**"What is the available opening through this gateway"**
   To be applied to **barrier=gate**
   tag to be added **maxwidth:physical=**

Checklist
-------------

Checklist for quest suggestions (see guidelines):

- [x]  🚧 To be added tag is established and has a useful purpose
**barrier=gate** is common with over 5,000,000 instances <https://taginfo.openstreetmap.org/tags/barrier=gate#overview>  
**maxwidth:physical=** has just over 2,800 instances <https://taginfo.openstreetmap.org/keys/maxwidth:physical#overview>. It's possible that mappers have used other tags such as **maxwidth=** ( *the legal restriction on size* ) with 50,000 uses <https://taginfo.openstreetmap.org/keys/maxwidth#overview> or **width=** with over 3,000,000 uses - not sure how many of these would refer to gates. But even with these figures there are many gates / gateways with no tag showing their available opening.
Without an indication of the gap available it's difficult to know if a particular item will fit through - this could be too narrow for a pram / pushchair to fit through or so wide that two buses can pass through together.   
   Adding a measurement of the available opening makes it possible to start to create navigation for such vehicles as electric wheelchairs, tricycles, hand powered cycles, larger electric vehicles for disabled users such as the Tramper mobility scooter <https://www.tramper.co.uk/products/tramper>   
   
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
  Once the tag **maxwidth:physical=** is applied to the **barrier=gate** node or way, this particular quest is complete and should not show for other surveyors.  

- [x] 🐿️ Easily answerable by any pedestrian from the outside but a survey is necessary  
  A similar screen to that which appears for users solving the road width quest would be suitable - I've tested the StreetMeasure app using a tape measure, and with care measurements of 0.5 metres can be obtained quite accurately.  

- [x] 💤 Not an overwhelming percentage of quests have the same answer (No spam)  
There's a lot of variations in gates! 

- [x] 🕓 Applies to a reasonable number of map data (Worth the effort)
Worldwide potential for over 5,000,000 items.   

Notes
---------

[...]

PR #5713, opened 2024-07-01
Barrier opening

Nick-Tallguy:
~~Draft  - ToDo~~  Now all complete with icon
--------

~~Requires a new icon - currently using icon from cycleway width~~  **New icon created and added**

https://github.com/streetcomplete/StreetComplete/issues/5678 refers

Have tested with emulator on Android Studio, and have apk installed on my phone (Google pixel 8A).  

[This changeset](https://www.openstreetmap.org/changeset/153382394)  by me on barrier = entrance.  

First attempt at anything to do with kotlin or Android, so please check carefully & please let me know if there is anything I can do to improve - would like to contribute more in the future.


app/src/main/res/values/strings.xml
```diff
@@ -1772,5 +1772,7 @@ Alternatively, you can leave a note (with a photo)."</string>
     <string name="label_log_message_contains">Message contains</string>
     <string name="title_logs_filters">Filter messages</string>
     <string name="quest_isAmenityIndoor_outside_covered">Not inside, but covered</string>
+    <string name="quest_barrier_opening_width_gate">What is the width of the opening here</string>
+    <string name="quest_barrier_opening_width_explanation">What is the maximum usable opening width here</string>
```
westnordost, line 1776: Question mark missing. Also, question and explanation sound redundant. Better put it into the question.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        // update width:barrier if it is set
+        if (key == "width" && tags.containsKey("maxwidth")) {
+            tags["maxwidth:physical"] = answer.width.toOsmValue()
+        }
+    }
+}
+
+val GATEWAYS = setOf("gate", "entrance", "sliding_gate", "swing_gate", "wicket_gate", "bollard", "block")
```
westnordost, line 79: Please remove `bollard` and `block`, both don't fit the question and will lead to confusion (What to measure, left or right of bollard? What if there are several bollards? The smallest distance or the biggest distance between the bollards? Up to where off the path should be measured - many paths don't have clear boundaries.)

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+
+    override fun isApplicableTo(element: Element) =
+        nodeFilter.matches(element)
+
+    override fun createForm() = AddGateWidthForm()
+
+    override fun applyAnswerTo(answer: WidthAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        val key = if (tags["barrier"] in GATEWAYS) "maxwidth:physical" else "width"
```
westnordost, line 62: I don't understand this logic. So, for which elements would `width` be tagged?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        tags[key] = answer.width.toOsmValue()
+
+        if (answer.isARMeasurement) {
+        tags["source:$key"] = "ARCore"
+        } else {
+        tags.remove("source:$key")
+        }
```
westnordost, line 70: indentation

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        // update width:barrier if it is set
+        if (key == "width" && tags.containsKey("maxwidth")) {
+            tags["maxwidth:physical"] = answer.width.toOsmValue()
+        }
```
westnordost, line 75: the code does something else than the comment claims.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,79 @@
...
+    private val excludedWaysFilter by lazy { """
+        ways with
+          highway and access ~ private|no
+    """.toElementFilterExpression() }
+
+    override val changesetComment = "Specify width of opening"
+    override val wikiLink = "Key:barrier"
+ // icon needed
+    override val icon = R.drawable.ic_quest_bicycleway_width
+    override val achievements = listOf(BICYCLIST, CAR, PEDESTRIAN, WHEELCHAIR)
```
westnordost, line 39: I'd suggest just wheelchair and bicyclist. Cars usually don't go through gates / gates on roads are broad enough for cars, of course. Gates that are not large enough for pedestrians wouldn't be gates, but uhh...

So anyway, please just use Wheelchair and maybe bicyclist.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    override fun onMeasured(length: Length) {
+        lengthInput.length = length
+        isARMeasurement = true
+    }
+
+    override fun onClickOk() {
+        val length = lengthInput.length!!
+        val newTags = element.tags + ("width" to length.toMeters().toString())
+        if (hasDubiousRoadWidth(newTags) != true) {
```
westnordost, line 67: hasDubiousRoadWidth seems unfit to check dubious values for gates on paths etc.

westnordost (changes requested):

app/src/main/res/values/strings.xml
Nick-Tallguy: Updated as per comment

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: Updated - bollard and block removed.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: **To be done** needs research by me - I copied text direct from another quest and need to research how it should be phrased.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: Indentation added, and also to previous line (Android Studio flagged previous line as a Lint issue).

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: **To be done** - needs research by me - copied from another quest and I need to research how it should be phrased.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: CAR & PEDESTRIAN removed from list.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
Nick-Tallguy: Should I remove this completely, or amend the values?  I wondered if it would be useful if an update included a measurement in the 100's of metres - missing decimal point.

westnordost: The first time I reviewed the code, I misread. I thought that as long as `hasDubiousRoadWidth` does not return false, it will always ask for confirmation.

But currently, it only displays that confirmation dialog if `hasDubiousRoadWidth(newTags)` is `true`, so that's fine. No change necessary, IMO

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: I've removed these lines of code completely - don't think they were needed at all.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: Lines of code amended - believe this is now correct, but if it's not I will need more help, or someone else amending this.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,72 @@
...
+
+        val key = if (tags["barrier"] in GATEWAYS) "maxwidth:physical" else 0
+            tags[key.toString()] = answer.width.toOsmValue()
+
+            if (answer.isARMeasurement) {
+            tags["source:$key"] = "ARCore"
+            } else {
+            tags.remove("source:$key")
+            }
+
```
westnordost, line 68: 1. indentation
2. `else 0`?! What?
3. Why `maxwidth:physical` and not `maxwidth`?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        savedInstanceState?.let { isARMeasurement = it.getBoolean(AR) }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val isBarrier = element.tags["barrier"] in GATEWAYS
```
westnordost, line 33: When is `isBarrier` false?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
```diff
@@ -0,0 +1,94 @@
...
+import de.westnordost.streetcomplete.databinding.QuestLengthBinding
+import de.westnordost.streetcomplete.osm.Length
+import de.westnordost.streetcomplete.osm.hasDubiousRoadWidth
+import de.westnordost.streetcomplete.quests.AbstractArMeasureQuestForm
+import de.westnordost.streetcomplete.quests.width.WidthAnswer
+import de.westnordost.streetcomplete.screens.measure.ArSupportChecker
+import de.westnordost.streetcomplete.view.controller.LengthInputViewController
+import org.koin.android.ext.android.inject
+
+class AddGateWidthForm : AbstractArMeasureQuestForm<WidthAnswer>() {
```
westnordost, line 17: Can you explain why `AddWidthForm` can not be used here?

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: Sorry, if I try to answer these remaining problems, I will just be wasting more of your time. I've tried following the provided guide on producing a quest, and have put together sections of 3 or 4 other guides, but have reached the point where I don't have the skills to go any further. 

Sorry if I've wasted your time.

westnordost:
Sorry to hear that. If you are unable to continue, I am sure someone else will be able to pick this up and bring this to completion as it seems to me that it is quite close to completion already.

westnordost:
Without having checked if there was a discussion about that beforehand, to me it appears like it makes more sense to **always** tag `maxwidth` on barrier openings rather than `maxwidth:physical` because the latter is usually only tagged if it differs from the signed maxwidth which is irrelevant here because the user is not asked for what is signed but is directly asked to provide the measured width.

mnalis:
The [wiki](https://wiki.openstreetmap.org/wiki/Key:maxwidth) seems pretty insistent that `maxwidth` exclusively _"specifies a legal restriction of the maximum permissible width of a vehicle"_, and has a separate _"common mistakes"_ section which warns about:

- Do not confuse the specification of the maximum permissible width of vehicles with the specification of the actual width via [width](https://wiki.openstreetmap.org/wiki/Key:width)=*.

- Do not confuse the specification of the legally restricted maximum width of vehicles with the specification of a physical width limit via [maxwidth:physical](https://wiki.openstreetmap.org/wiki/Key:maxwidth:physical)=*.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
mnalis: @Nick-Tallguy if you don't feel like working on it anymore, that is of course your choice and OK; I just wanted to say your effort is appreciated, and that **_multiple rounds of asking for improvements on PR are normal and expected_**, and not intended to be taken as a discouragement, but instead as opportunity to grow and produce quality code you'd later be proud of (and which would be easy for everyone else to understand, use, and improve). 

I know it can feel overwhelming and seemingly neverending stream of improvement requests,  but (having done it several times myself) I can assure you those eventually get resolved and results in great code and no more changes are requested. (With time, one gets better and can shave off few iterations, but even for people doing it for years it is rare  for anything but must basic changes not to get that _"changes requested"_ dialogs. )

So, if you're willing, I'd recommend trying some more for those finals pushes, as you seem close to the end of the tunnel. :heart: Remember: the harder the effort, the more valued the result will be too.  

And if you get stuck, don't be afraid to ask questions and clarifications. I'm not that good at Kotlin itself, but perhaps can help with OSM and some StreetComplete logic; and there are others lurking too.

Nick-Tallguy:
**maxwidth:physical or maxwidth**  
I have been using maxwidth:physical because of the guidance in the wiki at <https://wiki.openstreetmap.org/wiki/Key:maxwidth:physical> which also appears in <https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcycle_barrier#Using_dimension_attributes>.   
  I can think of many instances in the UK and France where the local authority imposes a 'maxwidth except for access' restriction on a narrow road to avoid larger vehicles clogging up the roads. However, if you are the driver of a larger vehicle which has to make a delivery there, the physical dimension of the gateway or entrance is important and may necessitate using another approach to your destination. 
  To me maxwidth:physical is the option to go for.

Nick-Tallguy:
@westnordost & @mnalis  
I feel that for this quest to be processed in a reasonable time it would be best if someone else finished it.

I like the fact that StreetComplete and SCEE are closely curated, and I intend to be involved, but I would rather not delay this quest any further by my lack of knowledge - I am studying, but have a way to go! At the moment my knowledge of kotlin is so poor that it will take a considerable time before I would be able to understand the code, and answer the queries.  

Regards & thank you

westnordost:
Ok, you are right about `maxwidth:physical`

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: @westnordost - I think I've updated this particular problem now

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
Nick-Tallguy: Sorry, I don't understand - it should be interpreted as a string, and only applies to those items that fit the filters?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
westnordost: Indentation is still wrong. (val key, the content of the ifs)

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
westnordost: But the form is only _opened_ for items that fit the filters. So unless I misunderstand something, `isBarrier` should always be true.

Basically, I don't see the reason why this line exists at all. But actually, I don't see the reason why this whole class exists at all. (See comment above.) Can the `AddBarrierOpening` quest not instead use the `AddWidthForm`?

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
Nick-Tallguy: AddWidthForm amended and used

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddGateWidthForm.kt
Nick-Tallguy: You're right, sorry - should be correct now.

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: think I've corrected this now - thanks for bearing with me!

app/src/main/java/de/westnordost/streetcomplete/quests/width/AddWidthForm.kt
westnordost, line 1: Please revert your changes of this file, they don't make sense.  (Or otherwise explain why you think they are necessary)

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
westnordost: Please look at line 64 and line 66

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,72 @@
...
+import de.westnordost.streetcomplete.quests.width.WidthAnswer
+import de.westnordost.streetcomplete.screens.measure.ArSupportChecker
+
+class AddBarrierOpening(
+        private val checkArSupport: ArSupportChecker
+    ) : OsmElementQuestType<WidthAnswer> {
+
+    private val nodeFilter by lazy { """
+       nodes with
+         barrier ~ ${GATEWAYS.joinToString("|")}
```
westnordost, line 22: Relating to my comment in `AddWidthForm`:

When GATEWAYS is not used in any other file, which it shouldn't, IMO, then you should write it here like this

barrier ~ gate|entrance| ...etc

and delete the GATEWAYS field.

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/quests/width/AddWidthForm.kt
Nick-Tallguy: all reverted - by way of explanation though - on line 33 the file reads   
" val isRoad = element.tags["highway"] in ALL_ROADS" -   

which I now realise was only because of the following line   
"val explanation = if (isRoad) getString(R.string.quest_road_width_explanation) else null"

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: Changes made

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -0,0 +1,70 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.WHEELCHAIR
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.width.AddWidthForm
+import de.westnordost.streetcomplete.quests.width.WidthAnswer
+import de.westnordost.streetcomplete.screens.measure.ArSupportChecker
+
+class AddBarrierOpening(
+        private val checkArSupport: ArSupportChecker
+    ) : OsmElementQuestType<WidthAnswer> {
```
westnordost, line 18:
```suggestion
    private val checkArSupport: ArSupportChecker
) : OsmElementQuestType<WidthAnswer> {
```

westnordost (approved):
Looks good to me.

Thank you for seeing it through!

PR #5719, opened 2024-07-03
Re-do UI of About screens and Settings screens in Compose

### LINKED ISSUE #561: New quest: Peak features

### LINKED ISSUE #5613: Logs view crash
How to Reproduce
1. Open log menu
2. Click Share

https://github.com/streetcomplete/StreetComplete/assets/25708359/c7f8f7ac-058c-4a33-ac74-4a55d7586903

It probably comes from a large number of logs, but I'm not sure because sometimes sending worked for a large number of logs.

Expected Behavior
The share menu is displayed

Versions affected
 StreetComplete 57.3

Bonus: crash when selecting filters (logs sended via email)

westnordost:
This is quite large.

### About screens

<br clear="both">

- Re-did About screen in compose
- Re-did Changelog screen in compose
- Re-did Credits screen in compose
- Re-did Privacy statement screen in compose
- Re-did Logs screen in compose (also fixes #5613)
- Screens that used a WebView before to display simple HTML for markup don't anymore (no official multiplatform WebView), instead, I wrote an own simple HTML parser + simple renderer for compose 🤯
- About screen no longer displays in a split-screen when in landscape mode. Didn't re-implement this behavior, I never was such a fan of it anyway. E.g. for the log-screen, it is better to show it at full-width always, and for the other sub-screens it actually also makes sense (credits, changelog, ...)

<hr>

### Settings screens

<br clear="both">

- Re-did Settings screen in compose
  - Changed how the settings are laid out a bit: The currently selected value is now always shown on the right, rather than below the name of the preference for settings selectable from a list or picker. This is also how it is usually done in iOS settings screens. This is more consistent and the design of settings screen was never part of any Material design guidelines anyway. The text below the setting name is now reserved for explanations of what this setting does
  - Subsequently, changed some wording / naming to be more brief and clear
  - Removes dependency to androidx.settings or how it was called
  - Removed the setting for map tile cache size (irrelevant for soon-to-be-merged MapLibre implementation and would have been more effort to implement, as there are no number pickers out-of-the-box in compose)
  - Removed the "auto" theme setting, which would change the theme based on time of day. Users that selected that option before will follow the system setting now. (As of June 2024, 95% of active installs from google play use an Android where AUTO setting is deprecated)
- Re-did Quest selection screen in compose
  - search interface looks different now
  - **Reordering quests with drag & drop is no longer supported** because this is a missing feature in compose so far
- Re-did managing quest presets in compose
  - re-did sharing quest presets with QR code diaog
  - Replaces dependency on google's all-purpose QRCode library zxing with qrose, a multiplatform compose-based QR code renderer
- Re-did showing quest forms in compose (debug feature)
- Settings screen no longer displays in a split-screen when in landscape mode. Didn't re-implement this behavior, I never was such a fan of it anyway. The UI looked so crammed that way.

### Preferences store
- I made it so that any access to the persistent key-value store (i.e. SharedPreferences on Android) goes through the new `Preferences` class which provides a type-safe interface to access, edit and listen to changes to that data

### Caveats
(all also noted as `TODO Compose` in the code)

- AlertDialogs are not scrollable in Compose. I worked around this by creating an own ScrollableAlertDialog, but this one always takes the maximum possible size for dialogs. E.g. the "What's New" dialog will fill basically the whole screen also if the changelog is really short
- scrollbars are not shown yet in Compose (it is in the roadmap, but in the backlog), hence the changelog, logs view, quest selection screen have no scrollbars
- also noted above: Compose offers no built-in possibility to easily implement drag and drop in lists, so reordering quest priorities has been disabled

### Notes
- on landscape mode, the QR code dialog and the logs filters dialog are too long to fit. This is not new, so I didn't redesign it. Possible solutions (for anyone wanting to perfect the UI): Make them each a new screen, or specifically for the logs screen, have the filters inline in the same screen, e.g. like the search input in the quest selection screen.

### TODOs for Multiplatform
- Compose is lacking (Android) toasts
- there are no date and time pickers in Compose (yet). Necessary for the logs filters and later also for the opening hours stuff


riQQ:
> * Re-did Logs screen in compose (also fixes #561)

From a cursory view, this seems to be unrelated to the linked issue. Am I'm missing something or did you link the wrong issue here?

app/src/test/java/de/westnordost/streetcomplete/data/visiblequests/QuestPresetControllerTest.kt
```diff
@@ -55,12 +56,12 @@ class QuestPresetControllerTest {
         ctrl.delete(55)
         verify(questPresetsDao).delete(55)
         verify(listener).onDeletedQuestPreset(55)
-        verify(selectedQuestPresetStore).set(0)
+        verify(prefs).selectedQuestPreset = 0L
     }
 
     @Test fun `change current preset`() {
```
FloEdelmann, line 62: It seems like this test now fails on the master branch (even with the [post-merge fix commit `6ba0371`](https://github.com/streetcomplete/StreetComplete/commit/6ba0371a3a9da23ee79b91b4f6606739127fc72d)): https://github.com/streetcomplete/StreetComplete/actions/runs/9829796567/job/27135248550

westnordost: Thanks, I will look into it

app/src/main/java/de/westnordost/streetcomplete/data/preferences/Preferences.kt
```diff
@@ -0,0 +1,246 @@
...
+
+    var resurveyIntervals: ResurveyIntervals
+        set(value) { prefs.putString(RESURVEY_INTERVALS, value.name) }
+        get() = prefs.getStringOrNull(RESURVEY_INTERVALS)?.let { ResurveyIntervals.valueOf(it) }
+            ?: DEFAULT_RESURVEY_INTERVALS
+
+    var showAllNotes: Boolean by prefs.boolean(SHOW_ALL_NOTES, false)
+
+    fun getLastPicked(key: String): List<String> =
+        prefs.getStringOrNull(LAST_PICKED_PREFIX + key)?.split(',') ?: listOf()
```
Helium314, line 40: Using comma as a separator is a bit dangerous. Cycleways are serialized as json and have two properties, so `{"cycleway":"SEPARATE","direction":"FORWARD"}` is serialized correctly, but broken on deserialization as it's now `{"cycleway":"SEPARATE"` and `"direction":"FORWARD"}`.

westnordost: Okay, but this is not a change. It has been that way also before. So how can this be an issue now and before not?

westnordost: Oh, nevermind, yes, that part didn't use the last picked values store class, only the same key

Helium314: I'm not sure, I only noticed that last picked selection is not shown in cycleway quest and overlay, and decided to have a look where things are going wrong. Since it did work before, I assumed thinks broke in this PR.

PR #5720, opened 2024-07-03
Do not ask AddParkingType on garage(s)

rusty-snake:
Do not ask AddParkingType on garage(s) as we do not provide a reasonable parking tag for garages.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/parking_type/AddParkingType.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/parking_type/AddParkingType.kt
@@ -12,6 +12,7 @@ class AddParkingType : OsmFilterQuestType<ParkingType>() {
         nodes, ways, relations with
           amenity = parking
           and (!parking or parking = yes)
+          and building !~ garage|garages
     """
     override val changesetComment = "Specify parking types"
     override val wikiLink = "Tag:amenity=parking"


rusty-snake:
Understand, let's catch up a little.

- This quests adds `parking=(surface|street_side|lane|underground|multi-storey)` to `amenity=parking`.
- Adding `parking=multi-storey` to a `building=parking` like https://wiki.openstreetmap.org/wiki/File:Gie%C3%9Fen-Uniklinik-Parkhaus.jpg seems to be correct. Even if it is incorrectly tagged as `building=garage`.
- Adding any of the avaiable `parking=` values of this quests to a `building=garage` (with `amenity=parking` and maybe `access=private|customers|...`) like https://wiki.openstreetmap.org/wiki/File:Two-Car_Garage.jpg seems incorrect.

Possible solutions:
- Do not ask on garage/s.
- Offer "This is a garage."

westnordost:
StreetComplete's job is to add missing data so that data consumers can consume this data. While the parking type data may be inferred by other tags, it is still useful to add it, otherwise this responsibility falls to data consumers. 

We only add this kind of refinement of the quest filter to quests that would otherwise be spammy. For example, we don't ask if a road has a sidewalk if it is unpaved. For parking garages I don't think this is the case. Private garages at detached homes are not tagged with `amenity=parking` and otherwise (garage) parking amenities are not _that_ numerous.

westnordost:
I reckon that maybe in the past I haven't been consistent when deciding on this point and/or leaning very much in the "don't bore users" direction.

In the end, it is always a trade-off between keeping the number of quests down to not bore users and to meet StreetComplete's mission in completing the OSM data, so of course there can be no clear guideline. But notice that in the current quest guidelines, there is no point about "don't ask about things that can be inferred by other tags", rather, the guidelines only speak about not boring the users.

rusty-snake:
> While the parking type data may be inferred by other tags, it is still useful to add it, otherwise this responsibility falls to data consumers.

Correct. However, if we ask for parking data of a object (garage(s) in this case) and don't offer a correct answer (`parking=multi-storey` is described with "a number of floors or levels") we will get incorrect data.
Which could be accepted with "is rare in reality".

> Private garages at detached homes are not tagged with `amenity=parking`

So your conclusion would be rather that

- `access=private`
- `amenity=parking`
- `building=garage(s)`

is likely wrong in the first place?

HolgerJeromin:
IMO a building=garage is always a tiny building for 1-10 cars. 
So I would call https://www.openstreetmap.org/way/363764265 a tagging mistake.
But there are many of them....

But the other way round (amenity=parking on a small building)

> Private garages at detached homes are not tagged with amenity=parking and otherwise (garage) parking amenities are not that numerous.

is probably really not very often.

matkoniecz:
>  So I would call https://www.openstreetmap.org/way/363764265 a tagging mistake.

I opened https://www.openstreetmap.org/note/4318881

westnordost:
> So your conclusion would be rather that
>
> access=private
> amenity=parking
> building=garage(s)
>
> is likely wrong in the first place?

For what?

rusty-snake:
For things like

- https://hoffmann-fertiggaragen.de/wp-content/uploads/2020/02/einzelgarage-als-hausanbau.jpg
- https://www.hausbaudirekt.de/wp-content/uploads/2020/06/Einfamilienhaus-mit-Garage-OEKOHAUS-HERB-Baufritz-1170x785.jpg
- https://www.garagen-riese.de/tl_files/sbx/images/web/garagen-carport-kombination-8.jpg

matkoniecz:
we have `access=private amenity=parking` but use of this for things like shown at https://github.com/streetcomplete/StreetComplete/pull/5720#issuecomment-2209085666 seems to be a tagging mistake to me

westnordost:
@rusty-snake See @mcliquid. I have never seen this tagged in the wild and I would somewhat treat this as a tagging error while for apartment complexes that have (shared, i.e. `private`) parking facilities using these tags would be fine IMO.  This falls in the same category as mapping stuff like benches, swings, sandboxes and what else one may typically find on private single-family-home property - it is off-limits.

mnalis:
I'd just add a comment that while something was built in a _format_ of garage building (i.e. `building=garage`) does not imply its **use** (while most popular use might indeed be private car parking, many people use them as a private motorcycle or bicycle parking instead, or as a shed, or as a small DIY workshop etc).

Issue #5687, opened 2024-06-14
Add `parcel_mail_in=returns_only` option to parcel locker drop off quest


rhhsm:
**Use case**
The new Parcel Locker drop off quest (#5639) shows options Yes or No. However, the wiki gives a third option `parcel_mail_in=returns_only` (Machine does not allow sending parcels but returning parcels is allowed.) For one of the 4 brands of parcel lockers in Bulgaria, this would be the correct choice, and choosing No would be wrong.

**Proposed Solution**
Add the `parcel_mail_in=returns_only` option too (so the options would be "Yes", "No", and "return parcel only")

**Remark**
I find this quest quite spammy. So far, the reply to "Can you pick up parcels here?" has always been "Yes". I wonder if there any parcel lockers where you can't pick up parcels, because that's the reason for their existence, isn't it? Wrt. parcel drop-off, the possibility is likely to be the same for all parcel lockers of a certain brand, so the answer is likely to be the same for each brand. It would then be much more efficient to add this information to the parcel locker brand preset, and exclude that brand from the SC quest.

westnordost:
@Arthur-GYT

Arthur-GYT:
I'll will take a look as soon as I can.

westnordost:
Regarding spamminess: Indeed, more than 99% of parcel lockers tagged with `parcel_pickup` do have `parcel_pickup=yes`, so indeed it seems to be a good idea to at least disable that quest by default.

Regarding brands: Hmm, really not sure if this would really always the same for the same brand. A situation where a brand have different types of lockers, some allowing e.g. mail-in and others not would be completely possible. Also, note that that quest just asks for the brand, and then just tags the `brand`, nothing else. It does not use any NSI presets.

rhhsm:
Did some more research in the meantime: only 0.42% (130) of parcel lockers have `parcel_pickup=no`. They are fairly evenly spread over the world, i.e. there is no particular country where most of them are located. None of them are in Bulgaria. In the Netherlands, there were two: one DHL which according to DHL website does allow parcel pickup, and one PostNL where the website doesn't mention pickup at all, possibly because it's obvious. Both the DHL and PostNL parcel locker presets in iD contain `parcel_mail_in=yes` and `parcel_pickup=yes` tags. I suspect that a large number of parcel lockers with  `parcel_pickup=no` are wrongly tagged.
[...]

PR #5730, opened 2024-07-09
Use correct answer label in AddParcelLockerMailInForm.kt

mpol:
There is a small mistake in the fix for #5687


mnalis (approved):

mnalis:
Nice catch, thanks!

westnordost:
Oh, how sloppy of me! Thank you!

PR #5731, opened 2024-07-09
Update AddTrackType quest to not include any ways tagged with bridge

### LINKED ISSUE #5566: Tracktype for bridges
Paths with `highway=track` may contain bridges, see here: https://www.openstreetmap.org/way/122304106

For these bridges, the tracktype quest (#487) „What’s the surface firmness of this track?“ is a bit misleading, since the icon and the images don't really fit to bridges.

I don't know what's the best way to fix this, some ideas:
* Hide quest for bridge=*
* Change quest wording / special quest for bridges
* Such bridges are incorrectly mapped and should actually not be `highway=track`

**Versions affected**
v57.1

kmpoppe:
Resolves #5566.
PR as per https://github.com/streetcomplete/StreetComplete/issues/5566#issuecomment-2214090173

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/tracktype/AddTracktype.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/tracktype/AddTracktype.kt
@@ -20,7 +20,7 @@ class AddTracktype : OsmFilterQuestType<Tracktype>() {
           or tracktype older today -8 years
         )
         and (access !~ private|no or (foot and foot !~ private|no))
-        and (!bridge)
+        and !bridge
     """
     // ~paved tracks are less likely to change the surface type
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/westnordost/streetcomplete/quests/tracktype/AddTracktype.kt
```diff
@@ -20,6 +20,7 @@ class AddTracktype : OsmFilterQuestType<Tracktype>() {
           or tracktype older today -8 years
         )
         and (access !~ private|no or (foot and foot !~ private|no))
+        and (!bridge)
```
westnordost, line 23: superfluous `(...)`

mnalis (approved):

PR #5733, opened 2024-07-10
skip roof and building levels quest for houseboats

HolgerJeromin:
Level is probably always `1`.
Roof shape is hard to decide either.

Greetings from the netherlands :)

https://taginfo.openstreetmap.org/tags/building=houseboat#combinations

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
@@ -17,6 +17,7 @@ class AddBuildingLevels : OsmFilterQuestType<BuildingLevelsAnswer>() {
                or !roof:levels and roof:shape and roof:shape != flat
            )
            and !building:min_level
+           and building !~ no|houseboat
            and !man_made
            and location != underground
            and ruins != yes
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/roof_shape/AddRoofShape.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/roof_shape/AddRoofShape.kt
@@ -21,7 +21,7 @@ class AddRoofShape(
           ((building:levels or roof:levels) or (building ~ ${BUILDINGS_WITH_LEVELS.joinToString("|")}))
           and !roof:shape and !3dr:type and !3dr:roof
           and building
-          and building !~ no|construction
+          and building !~ no|construction|houseboat
           and location != underground
           and ruins != yes
     """.toElementFilterExpression() }


app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
```diff
@@ -17,6 +17,7 @@ class AddBuildingLevels : OsmFilterQuestType<BuildingLevelsAnswer>() {
                or !roof:levels and roof:shape and roof:shape != flat
            )
            and !building:min_level
+           and building !~ no|houseboat
```
mnalis, line 20: Is this line needed? Few lines above it checks only for `building ~ ${BUILDINGS_WITH_LEVELS.joinToString("|")}` and it doesn't seem to include those two:
https://github.com/streetcomplete/StreetComplete/blob/05ef7ce4564e8eba42220ae02e38f87f0e7b6aa5/app/src/main/java/de/westnordost/streetcomplete/osm/BuildingLevels.kt#L3-L9

Are you being asked for building levels on houseboats?

kmpoppe: There are 17.5k `building=houseboat` worldwide and none of them have a `levels`-tag. I guess SC isn't currently asking for levels on houseboats.

westnordost:
Hmm, not sure, depends on the size of the houseboat, no?

westnordost:
In a quick DuckDuckGo image search for houseboat, the first 7/16 results showed a boat with more than one visible level. So, they definitely exist. Can't hurt to ask for the building levels of a houseboat.
For roof shape, it was 6/16 that had a non-flat roof.

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt
HolgerJeromin: Gna. You both are right. 
I was asked about roof shape and 
wanted to go the extra mile and fix the other building quest, too :-)

HolgerJeromin:
> For roof shape, it was 6/16 that had a non-flat roof

I had these in mind: 
https://www.mapillary.com/app/?pKey=243474601860404&lat=52.158796&lng=4.5050387&z=1.5&x=0.377273175545624&y=0.4429972381890228&zoom=0&focus=photo
But perhaps they are the same as other buildings (mixed shape).
Others in this canal are clear flat

westnordost:
Ok, well, they look rather like regular boats used as homes. But in any case, other house boats exists that do have several levels and different roof forms, so it makes sense to ask it.

Anyway, here are some house boats in Hamburg:

https://www.google.de/maps/@53.5463225,10.0200979,3a,42.6y,42.41h,84t/data=!3m7!1e1!3m5!1sXjtUAyQd6aczCM1jzFeWgw!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fpanoid%3DXjtUAyQd6aczCM1jzFeWgw%26cb_client%3Dmaps_sv.share%26w%3D900%26h%3D600%26yaw%3D42.406638630584176%26pitch%3D5.999792744316181%26thumbfov%3D90!7i16384!8i8192?coh=205410&entry=ttu

This is in a commercial district, so I am not sure if they are actually used as homes or just additional office space, event space, airbnb or whatever.

HolgerJeromin:
Fine with closing this PR: but do we want the other way round for level (asking level)?
mnalis found that this is not the case right now.

PR #5737, opened 2024-07-10
Add houseboats to BUILDINGS_WITH_LEVELS

HolgerJeromin:
See #5733 for discussion


westnordost:
Ah, right! So, it wasn't asked for houseboats before at all! Well. Thank you!

PR #5741, opened 2024-07-14
[WIP] Add Boat Rental Quest

### LINKED ISSUE #5426: Ask what sort of boats can be rented
### General
Affected tag(s) to be modified/added: [`*_rental=*`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dboat_rental)
Question asked: **What sort of boats can be rented here?**

### Checklist
Canoe rental is very different to motorboat rental, it is important to be able to distinguish these
Only just over 5000 `amenity=boat_rental` but many (>~75%?) do not have a `_rental` tag.

### Ideas for implementation

**Element selection:** 
`amenity = boat_rental` and `*_rental != *`

**Metadata needed:** 
Don't think so. Maybe different boats in different countries?

**Proposed UI:**
Picture/icon multi-select for kayak, canoe, stand-up paddleboard, sailing boat, dinghy, motorboat and pedalboat.

kmpoppe:
Resolves #5426.

Known issues - feedback required:
* Order in QuestModule
* Is this a RARE achievement (only 5.3k boat_rentals tagged today)
* Answer SVGs range from 0 to 100 in my view as to their fit (I love the floaty duckboat for "pedalboard" ;-)) Wouldn't CC0 images from wikicommons be the better idea? Lacking any kind of artistic skills, "fun icons", as asked for by westnordost in the comments of the Issue is nothing I can provide, any help would be appreciated!


mnalis:
- I'd rather use [rowboat_rental](https://wiki.openstreetmap.org/wiki/Key:rowboat_rental)=`yes/no` instead of `dinghy_rental=yes/no` to avoid ambiguity ([Dinghy](https://en.wikipedia.org/wiki/Dinghy) may use motor instead of oars); especially as both tags (while documented) are not really used currently (single-digit uses worldwide).
- I think you meant (both in description and tag name) `pedalboat` instead of `pedalboard` (I was looking at https://en.wikipedia.org/wiki/Pedalboard in confusion just now :smiley: )
- I don't think it is RARE (but I might be affected by having dozens of boat rental places in few miles around me ATM)

westnordost:
Hm, I am sad that you didn't take the pictures that I proposed. I see no reason whatsoever why one should prefer black-and-white icons over the more colorful icons.

app/src/main/res/values/strings.xml
```diff
@@ -799,6 +799,17 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_board_type_map_title">Is it just a map and only a map?</string>
     <string name="quest_board_type_map_description">If the board is about a particular topic, please specify it, regardless of whether it also includes a map. For example a public transport board may have a bus route map. If the topic is not listed among the available answers, please consider leaving a note instead.</string>
 
+    <string name="quest_boat_rental_title">Which type of boats are for rent here?</string>
+    <string name="quest_boat_rental_motorboat">Motorboat</string>
+    <string name="quest_boat_rental_jetski">Jetski</string>
+    <string name="quest_boat_rental_sailboat">Sailboat</string>
+    <string name="quest_boat_rental_pedalboat">Pedalboat</string>
+    <string name="quest_boat_rental_kayak">Kayak</string>
+    <string name="quest_boat_rental_canoe">Canoe</string>
```
westnordost, line 808: A kajak is also considered a type of canoe. To distinguish, I think the term [_Canadian canoe_](https://en.wikipedia.org/wiki/Canadian_(canoe)) is preferable.

app/src/main/res/values/strings.xml
```diff
@@ -799,6 +799,17 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
+    <string name="quest_boat_rental_houseboat">Houseboat</string>
+    <string name="quest_boat_rental_dinghy">Dinghy</string>
+    <string name="quest_boat_rental_paddleboard">Standup Paddleboard</string>
```
westnordost, line 811:
```suggestion
    <string name="quest_boat_rental_paddleboard">Standup Paddleboard (SUP)</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -799,6 +799,17 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
...
 
+    <string name="quest_boat_rental_title">Which type of boats are for rent here?</string>
+    <string name="quest_boat_rental_motorboat">Motorboat</string>
+    <string name="quest_boat_rental_jetski">Jetski</string>
+    <string name="quest_boat_rental_sailboat">Sailboat</string>
+    <string name="quest_boat_rental_pedalboat">Pedalboat</string>
+    <string name="quest_boat_rental_kayak">Kayak</string>
+    <string name="quest_boat_rental_canoe">Canoe</string>
+    <string name="quest_boat_rental_houseboat">Houseboat</string>
+    <string name="quest_boat_rental_dinghy">Dinghy</string>
```
westnordost, line 810: Is that the same as a rowboat? Maybe call it "Dinghy / Rowboat" then

kmpoppe:
You listed 3 or 4 icons that could be used without any changes, for all others there's work needed that I am not capable of doing artistically to the standards I hold myself to.

This is why this is a draft PR.

res/graphics/quest/boat_rental.svg
westnordost, line 1: A houseboat is not the most representative specimen of a boat rental. Rather a canadian or rowboat would be. 

Also, the background color should be green because it is a leisure amenity.

res/graphics/boat rental/sailboat.svg
westnordost, line 1: Most (i.e. all that I found) sailboat rentals do **not** rent sailing ships (with a lower deck, i.e. where you can sleep) but sailing boats. I think to show a smaller sailing boat would be more representative.

res/graphics/boat rental/motorboat.svg
westnordost, line 1: See comment about the sailing ship. Motor boats that you can rent are usually smaller ones. E.g. for fishing. No cabin.

res/graphics/boat rental/kayak.svg
westnordost, line 1: This is not a kayak, it is a (canadian) canoe. You can see that by the shape of the hull and by the paddle.

res/graphics/boat rental/pedalboat.svg
westnordost, line 1: Hmm, it might be funny, but I am not sure if this could lead to misunderstandings. Most rentals I know that do rent pedal boats only have 1-2 or none of such "special design" pedal boats.

res/graphics/boat rental/sailboat.svg
mnalis: Interesting, I guess that might depend on country... In Croatia, most sailboats are at seaside and are indeed sailing ships intended for richer clientele groups (with sleeping lower decks), and moreso are charged per _week_ (which would be inconvenient if they were without sleeping quarters). e.g. [here](https://www.croatia-yacht-charter.com/hr/jedrilice/) or [here](https://www.yacht-rent.hr/najam-plovila/jedrilica). You must also rent at least a skipper, unless you hold a valid license yourself (and there is usually an option for renting full crew complement and cook too).

app/src/main/res/values/strings.xml
mnalis: FWIW, I did know what SUP is, but TIL what it is acronym _of_ :smiley:

app/src/main/res/values/strings.xml
mnalis: I'd even go for `rawboat` alone. As noted above:

> I'd rather use [rowboat_rental](https://wiki.openstreetmap.org/wiki/Key:rowboat_rental)=`yes/no` instead of `dinghy_rental=yes/no` to avoid ambiguity ([Dinghy](https://en.wikipedia.org/wiki/Dinghy) may use motor instead of oars); especially as both tags (while documented) are not really used currently (single-digit uses worldwide).

IOW, rowboat is a subtype of dingy which uses oars (and not a motor) for moving.

res/graphics/boat rental/sailboat.svg
westnordost: Right, it makes sense that they exist. Yet, they are **very** different kind of rentals than the kind of sailing boats you take on a joyride for a few hours on some lake. It makes no sense to put these into the same tag. Hence, these should at least be two selection options

See also https://github.com/streetcomplete/StreetComplete/issues/5426#issuecomment-2227423403:

> a sailboat is not a sailboat. There should be some distinction between a sailboat you can rent for a few hours with a capacity of a few people and a sailboat that you treat basically like a camper van on water. Maybe a good line to draw would be whether the boats have a cabin or not, i.e. you can go below deck. On the other hand, of the rentals above, there was not a single one that would rent out big sailboats in which you could spend the night.

mnalis: True. There are few uses of [yacht_rental=yes](https://taginfo.openstreetmap.org/keys/yacht_rental#overview), but not yet documented on the wiki. I may document that value linking to [wikipedia Yacht article](https://en.wikipedia.org/wiki/Yacht) if people agree?

So perhaps `small sailboat` and `yacht (with sail and/or motor)` (note that those bigger sailing ships which have sleeping quarters, also usually have backup motors, and/or might even have their own dinghy with outboard motor of its own (in case of emergency or to avoid paying for mooring :smiley:).

app/src/main/res/values/strings.xml
kmpoppe: Resolved with 435b6c7b552a89b61281b81582df87a42acd75d5

eginhard:
Catamarans would be another common boat type.

westnordost:
Sounds like a hull type rather than a boat type? I mean, many/most pedal boats are actually catamarans. 

And this thing, whatever you'd call that



is also a catamaran.

But I guess the usual type of catamaran you get on boat rentals (that you would also call a catamaran) would be sailing boats, right? So, if this is common, maybe better have a value like `sailing_catamaran`.

westnordost:
There's also an activity called "tubing" where you'd rent the tube of a truck tire. Not sure how common it is (compared to rafting/kayaking), I've only seen it in Vang Vieng (Laos):

https://duckduckgo.com/?q=vang+vieng+tubing&t=ffab&iax=images&ia=images

kmpoppe:
Added help wanted, people being able to contribute good graphics are welcome to do so.

westnordost:
I'll look into icons (maybe next week). Is it otherwise done? E.g. I think certain values should be considered ambiguous and thus if that value is present, the quest should be asked again (such as dinghy).

kmpoppe:
I'll look through the comments when I get home and check how the answer could be made better!

westnordost:
I added some icons on master. Most are from https://www.svgrepo.com/collection/water-sports-flat-icons/ or mashups from these.

Missing are:
- pedalboat - probably should show from front, otherwise looks too similar to motoboat
- yacht - should be a smallish yacht with sails
- (raft - could be similar to SUP icon)

westnordost:
Are you still planning to work on this? It would be a shame if I prepared all those icons for nothing.

Issue #5616, opened 2024-05-05
Migrate to Jetpack Compose (master)


westnordost:
This is the master ticket for migrating from the Android view system (layout XMLs etc.) to [Jetpack Compose](https://developer.android.com/jetpack/compose) as UI framework.

The primary reason for this is #5421: To make possible an iOS version of this app that has a shared codebase for the UI through the use of [Compose Multiplatform](https://www.jetbrains.com/lp/compose-multiplatform/). The first step however is to migrate to Jetpack Compose and only later switch to Compose Multiplatform.

The first foray into Jetpack Compose has been made in PR #5447, I am also new to this but I tried to follow guidelines closely how things should be done, so feel free to look into that PR to learn how the UI code typically looks like before and after.

## Contributing

As this is a larger undertaking which can be implemented step by step - this ticket shall serve as a master ticket. Whenever you would like to contribute, post on which screen or view you would like to work on now here and later post a PR when you are done. 

Migration is generally done bottom-up, i.e. start with single views first and then work your way up to the screen-level, see [Migration Strategy](https://developer.android.com/jetpack/compose/migrate/strategy). For converting a whole screen, it is very advisable to do this only after the Fragment or Activity in question has been refactored to use ViewModels, see #5530.

westnordost:
**Current blockers and issues with Compose:**

- Appear-animation missing for achievements and links screen. [Lazy item animations not supported by Jetpack Compose yet](https://developer.android.com/jetpack/androidx/compose-roadmap).

- Fancy animate-from-icon animation is missing for achievement dialog. [Not supported by Jetpack Compose yet](https://developer.android.com/develop/ui/compose/animation/shared-elements)

- (Material2) DatePicker and TimePicker (dialogs) are not available in Jetpack Compose (yet) 

- scroll bars are not supported by Jetpack Compose yet. (They would be handy at least in the Log screen)

westnordost:
Re-do of About and Settings screen done, in review. Details here: https://github.com/streetcomplete/StreetComplete/pull/5719/files

westnordost:
User screens are also done, except for the quest statistics / country statistics screen. See #5595

westnordost:
I'll look into the two tutorial screens (start tutorial, overlay tutorial) coming week.

westnordost:
tutorials screen is done

westnordost:
Next up would be the buttons and controls on the main screen plus the menus and dialogs there.

[...]

PR #5746, opened 2024-07-15
Reimplemented Tutorials in Compose

westnordost:
Part of #5616


Helium314:
I remember discussing about choppy animations. On S4 mini it seems surprisingly fine, as long as you wait for a while (until tangram sprite sheet creation is done) before going through the tutorial.

PR #5786, opened 2024-08-04
Change wording in CONTRIBUTING_A_NEW_QUEST.md

danieldegroot2:
Ref https://github.com/streetcomplete/StreetComplete/commit/ebcfb20a0ed542343c996c619c3c96cfe8baf1f0

The words "how to answer" are used at the start of both the first and the second paragraph.
This seems like duplicate/ambiguous wording.

When, or for what type of instructions, is the info button not the right choice?
This only becomes clear once you read the examples.

There are two paragraphs:

'instructions on how to interpret the terms and interface'
'instructions on how to formulate the answer and when to use an answer'

This PR changes the wording used to clarify this difference.


westnordost:
Hm, I don't fancy that wording.

The previous wording was clearer, IMO. Adding several notes in braces doesn't help readability.

I find it important to mention that **instructions** for in what way to answer the quest should be written either in the question itself or as a label next to the input field, i.e. always visible.
Your wording suggestion (again) uses the "doubts about the..." wording, which is duplicating the other paragraphs and missing the point. **It is not about clearing doubts**, it's about **instructions on how to answer**.

Doubts can be cleared in the explanatory text shown when clicking on the info button. It is important to make a clear distinction here.

westnordost:
E.g. the wording could be changed like "Instructions on in what way the quest should be answered" or whatever in order to avoid duplicating the "how to answer" from the previous paragraph.

Issue #5776, opened 2024-08-01
tourism=camp_site: power_supply asks wrong question


vsessink:
**How to Reproduce**
Find a tourism=camp_site or tourism=caravan_site. Wait for the question quest_camp_power_supply_title

The question says: "Is there any way for customers to charge their phone, laptop etc. here?" It then changes the power_supply tag.

This is wrong. The power_supply tag has nothing to do with _phones or laptops_.

The OSM tag "power_supply" just means that there is electricity "available" on the camp site: _power_supply=yes - means, there is a power supply available (e.g. on a camp_site)_ 

The tag seems to _imply_ that this means power supplies _at the pitches_, because a blue 3-pin socket is being described - which is the typical (European?) outdoor power supply for camp sites. (`power_supply cee_17_blue Blue 3-pin socket with 230V is available` is what the example says but  that is deprecated, `power_supply=yes` and `socket=cee_17_blue` would be the preferred way to encode a blue 3-pin socket).

**Expected Behavior**
Question should conform to the tag it is asking about, in this case: power_supply. As the camp_site OSM tag says that power_supply means "There is power supply available for customers", better wording would be: "is there electricity available for customers to use".

This _still_ is ambiguous, but more aligned to the OSM tag it changes.

"Is there electricity available on the pitches" or "are there power sockets available on the pitches" is probably the original meaning of the power_socket tag.

**External links**
https://wiki.openstreetmap.org/wiki/Proposal:Extend_camp_site camp_site extension proposals (with power_supply examples)
https://wiki.openstreetmap.org/wiki/Key:power_supply original power_supply tag
https://wiki.openstreetmap.org/wiki/Key:socket:* list of socket types

**Versions affected**
v58.w

matkoniecz:
> power_supply tag has nothing to do with phones or laptops.

this is quite overstating things, as both are typically powered by electricity provided by power sockets

In which cases this question will result in wrong answers?

matkoniecz:
maybe exact text was discussed at https://github.com/streetcomplete/StreetComplete/pull/4213

vsessink:
Well, I'm actually at a camping place now. All pitches have a `cee_blue` socket, but my answer to "Is there any way for customers to charge their phone, laptop etc. here"  would definately have been "no", as there are no ways to charge anything here.

The only reason I answered "yes" was because I questioned the wording and took a search to see what the answer would do - which ended up being `power_supply=yes/no`.

[...]

PR #5796, opened 2024-08-14
try to improve Camping power supply quest wording

mnalis:
Fixes https://github.com/streetcomplete/StreetComplete/issues/5776

People seem to like the new texts, so let's go with it. Hopefully it will clear up more than it misleads (it should, provided that users that are unsure _do_ click on that info button).

Also, if people think texts could be further improved, now is the time to suggest improvements

(draft, needs testing)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -958,6 +958,7 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
     <string name="quest_camp_drinking_water_title">Is drinking water freely available to campers here?</string>
     <string name="quest_camp_shower_title">Are there showers here?</string>
     <string name="quest_camp_power_supply_title">Are there any power supply sockets available for guests?</string>
+    <string name="quest_camp_power_supply_hint">This can be sockets at the pitches, guest usable sockets in a service building, sockets for phone/laptop charging, etc.</string>
     <string name="quest_camp_type_title">Who may camp here?</string>
     <string name="quest_camp_type_tents_and_caravans">Both tents and motor homes, travel trailers, …</string>
     <string name="quest_camp_type_caravans_only">Only motor homes, travel trailers, …</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -957,7 +957,7 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
 
     <string name="quest_camp_drinking_water_title">Is drinking water freely available to campers here?</string>
     <string name="quest_camp_shower_title">Are there showers here?</string>
-    <string name="quest_camp_power_supply_title">Is there any way for customers to charge their phone, laptop etc. here?</string>
+    <string name="quest_camp_power_supply_title">Are there any power supply sockets available for guests?</string>
     <string name="quest_camp_type_title">Who may camp here?</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -957,7 +957,8 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
 
     <string name="quest_camp_drinking_water_title">Is drinking water freely available to campers here?</string>
     <string name="quest_camp_shower_title">Are there showers here?</string>
-    <string name="quest_camp_power_supply_title">Is there any way for customers to charge their phone, laptop etc. here?</string>
+    <string name="quest_camp_power_supply_title">Are there any power supply sockets available for guests?</string>
```
westnordost, line 960:
```suggestion
    <string name="quest_camp_power_supply_title">Are there any power sockets available for guests?</string>
```
_power supply socket_ sounds needlessly complicated. Wikipedia also calls them power sockets:

https://en.wikipedia.org/wiki/AC_power_plugs_and_sockets

app/src/main/res/values/strings.xml
```diff
@@ -957,7 +957,8 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
...
+    <string name="quest_camp_power_supply_hint">This can be sockets at the pitches, guest usable sockets in a service building, sockets for phone/laptop charging, etc.</string>
```
westnordost, line 961:
```suggestion
    <string name="quest_camp_power_supply_hint">This could be sockets directly at the campsite, sockets in a service building accessible by guests, a dedicated charging station for phones, laptops, etc.</string>
```

mnalis: Hmmmm, _"directly at the **campsite**"_ sounds quite confusing/**problematic** to me. I know SC nontranslated texts are in US English, but while _"campsite"_ is indeed more popular in US, it seems to be regional there too?

If we want to avoid _"camping pitch"_ (more popular in British English) we should IMO find some alternative way to say what we want (because _"campsite"_ in BE means what _"campground"_ does in US -- i.e. the whole fenced area, and not the individual pitch, which is highly confusing --  see https://en.wikipedia.org/wiki/Campsite for details)

(another related issue is that _"power socket"_ is AFAICT more British term, preferred US term apparently being _"electrical outlet"_).

Any native English speakers around? Help!

westnordost: Oh, I didn't know that. I actually meant the British English version. I.e. some sockets somewhere on the camping area. (And you meant sockets directly for each parcel?)

westnordost: Right, electrical outlet makes sense.

https://books.google.com/ngrams/graph?content=power+outlet%2C+electrical+outlet%2C+power+socket%2C+electrical+socket&year_start=1900&year_end=2019&corpus=en-US&smoothing=3

Power outlet is also fine. (Just asked in the #language channel on slack-us)

westnordost: People in the #language channel find "power outlet" to be more common. It's also shorter!

mnalis: OK I've tried to incorporate that; how does it look now? Let's suggest new edits if needed.

westnordost:
Ready for merge?

mnalis:
It's seems like a simple change, but I've only tested it compiles without errors

westnordost:
Can we go below 90 open issues before I release the alpha? :->

PR #5797, opened 2024-08-14
add support for highway=busway (fixes #5749)

### LINKED ISSUE #5749: Treat busways mostly like roads
**Use case**
Busways (`highway=busway`) aren't rendered and aren't considered like roads. There aren't quests on them (while there may be crossroads, bus stops, bike lanes...). They should be treated just like normal roads for most of the quests, except some like number of lanes, speed limit...

You can look at Rouen, France to see a busway in urban area (https://www.openstreetmap.org/#map=17/49.43990/1.09030&layers=P, do not change the layers, default OSM style doesn't render busways)
Rouen's busway is also allowed to bikes, a quest for busways asking if bicycles, taxis and/or deliveries are allowed could be added.

Here's a zoom onto a part of Rouen's main busway :
ResizedImage_2024-07-17_21-02-16_3667
*There is a busway in the blank, and a bus station below the building on which there's a clock quest*

**Proposed Solution**
- render busways on the map
- enable a subset of the main quests on such ways (most quests except some like number of lanes or speed limit)
- (eventually) add a quest asking if bicycles/taxis/deliveries are allowed

westnordost:
- `AddRoadName` OK. 62% of busways already have a name
- `AddLanes` OK. Some will have 2, some (the oneway ones) will have 1.
- `AddOneway` OK. 70% of busways are tagged with oneway=yes
- `AddMaxSpeed`, `AddMaxWeight`, `AddMaxHeight` OK. Signage for buses is the same as signage for car traffic in general
- `AddRoadSurface`, `SurfaceOverlay` OK
- `AddRoadSmoothness` OK.
- `AddRoadWidth` Hmm, OK. It is only asked under very specific circumstances anyway, i.e. `lanes < 2` or `lane_markings=no` and also` access != no` (like many others)
- `AddWayLit`, `WayLitOverlay` OK, could make sense for the big picture
- ~`AddSidewalk`~, ~`AddCycleway`~ While theoretically possible, it has zero usage. Oftentimes busways are layouted in the same way trams would be layouted, i.e. in the middle of the road, between the car lanes. Of course there is no sidewalk or cycleway. However, it **should** be asked in the case that the busway has invalid sidewalk or cycleway tagging
- `AddSidewalkSurface` OK. This is only asked for ways that DO have a sidewalk already tagged
- ~`AddProhibitedForPedestrians`~ No need to ask.
- ~`StreetParkingOverlay`~ Can't park there anyway
- `AddCrossing` OK. Pedestrian crossings with busways are possible.
- `AddBarrierOnRoad` OK. If a fence intersects with a busway, it is fair to ask about the type of opening in that barrier for the busway
- `AddKerbHeight`, `AddTactilePavingKerb` OK. (when footway crosses busway)
- `SidewalkOverlay`, `CyclewayOverlay` OK. Theoretically possible after all. But it should not be highlighted in red (i.e. as missing).

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Highway.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Highway.kt
@@ -1,10 +1,10 @@
 package de.westnordost.streetcomplete.osm
 
 val ALL_ROADS = setOf(
-    "motorway", "motorway_link", "trunk", "trunk_link", "primary", "primary_link",
-    "secondary", "secondary_link", "tertiary", "tertiary_link",
+    "motorway", "motorway_link", "trunk", "trunk_link",
+    "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link",
     "unclassified", "residential", "living_street", "pedestrian",
-    "service", "track", "road"
+    "service", "track", "busway", "road",
 )
 
 val ALL_PATHS = setOf(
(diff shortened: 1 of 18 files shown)


Aeris1One:
Agreed on the described implementation.
Didn't review the code.

PR #5799, opened 2024-08-15
Migrate main screen controls etc. to Compose

westnordost:
What is migrated:
- buttons on main screen
- main menu
- team mode dialog (it is now wizard-like)
- overlay selection dropdown
- edit history sidebar and undo dialog

To do / Known bugs:
- **pull out the general changes not related to the above and put them on master!**
- recording track doesn't work? Button does not appear
- also migrate upload button dialog ("you should login") now?
- the crosshair is misplaced in landscape mode
- (restore paint roller animation from before migration to compose?)
- also migrate location pointer pin now? Or rather turn it off because of https://github.com/maplibre/maplibre-native/issues/2255 ?


westnordost:
Since this will not go into v59.0 and thus it should not be merged to master before that version is released, I will just continue working in this branch and extend the scope of this.

Next on would be to also viewmodel & compos-ify the logic in MainActivity and remove MainFragment (put it all into MainActivity).

westnordost:
I had to stop myself from ever-expanding the scope of this. I reviewed the stuff myself and tested most of the things.

westnordost:
@Helium314 @FloEdelmann @matkoniecz I will soon merge this to master. If you would like to lend a hand in reviewing this before it is merged, you should do it soon.

westnordost:
I kind of don't like about this work that it - necessarily - leaves to many open ends, i.e. contact points between the ViewModel+Compose code and the (old) Android Fragments/Activity code. But it is currently not possible really to reduce this, as everything comes together at the MainActivity and while not all things the MainActivity needs to talk to are all re-done in ViewModel+Compose.

It is also particularly hard to migrate this properly because there is so much stuff happening specific to Android.
Certainly there may be some parts that could be migrated now, but the longer this PR sits around, the more problematic it gets to finally merge it.

FloEdelmann (commented):
I had a cursory look over the changes (without running the app or tests) and found nothing that caught my eye.

westnordost:
Thanks! I took it out for a test today and found two issues that I'll try to fix today/tomorrow:

1. Need darker shadow on the star counter, the number is hard to see
2. The crosshair position, the pin position and the actually placed position each differ when placing things in e.g. the things overlay

westnordost:
3. edit history sidebar: background color of already uploaded edits is ugly

westnordost:
4. when selecting "don't keep activities" in developer options and then tabbing out of StreetComplete, the following things don't seem to work at all anymore: compass, zoom in, zoom out and location button; tapping on overlays, tapping on any pins, display/highlight of selected edit in edit history sidebar

rugk:
The second animation/video with the quest icons sliding up from below the screen but in tilted mode looks kinda uhm… tbh unprofessional /childish.

I guess the actual reason for that is that it is tilted in perspective but it does not account for that in the animation. The animation is just a slide in effect from below, which would maybe work for a flat image, but not the 3d view there. IMHO this reminds me of some silly PowerPoint animations I once did. Not blaming here, just trying to make you aware of that. Especially maybe as the icons are flat and have no "depth", it looks very artificial and not "real" of you get what I mean? 
It may be especially surprising if/as it does not follow the tilt from the map aka the map in the video was in 2d top-down view.
Porposed solution would this just be to adhere for the tilt of the image itself aka I guess just zoom in while moving in or so or some real animation I dunno. Or maybe animate each icon/quest at once and not in "bulk". Like dropping that animation altogether and have the icons "pop in" with some random delay and possibly location (transparency fade effect maybe with zoom), so it looks like what you seem to convey here:"quests pop up everywhere, this is too much"?

PR #5800, opened 2024-08-15
add bollard / block to barrier_opening

### LINKED ISSUE #5783: Quest request - what is the available opening around this bollard or block?
### General
Question asked: **Is there an example tag on item XY?**

Adding **maxwidth:physical=** to instances of **barrier=bollard** and **barrier=block**  

### Checklist
- [x] 🚧 To be added tag is established and has a useful purpose.  
       Bollards are frequently placed to prevent access by vehicles - but without an indication as to the available gap around a bollard, or between bollards, you cannot assess what can pass - wheelchair, cycle, pedestrian, or car?

     [Wiki OpenStreetMap barrier=bollard](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbollard)  

     [TagInfo for barrier=bollard](https://taginfo.openstreetmap.org/tags/barrier=bollard#overview)

     Taginfo shows over 735,000 are mapped, mainly as nodes, but some are as ways. The wiki indicates that **maxwidth:physical** should be used, but searches show there are few instances : [Overpass Turbo search for large part of Europe](https://overpass-turbo.eu/s/1OR1)

       Once the tag **maxwidth:physical** with a value has been added, the quest should be shown as complete.  

       The easiest method of obtaining the measurement is using **StreetMeasure** from within the **StreetComplete** app, but the quest could also be completed by someone willing to use a tape measure or similar.  

       There does not appear to be a standard distance between bollards, or between a bollard and another physical feature.  

       735,000 items already tagged as barrier=bollard with very few having a maxwidth:physical tag as well.  

### Ideas for implementation

**Element selection:** 
       Applies to bollards / blocks on paths, cycleways, & highways. Any ways / bollards that are tagged with either access = private, no, customers or agricultural should be ignored.  

   I suggest that, if this request for a quest is agreed, **barrier=block** is dealt with at the same time.

**Metadata needed:** 
      I think this quest should be applied worldwide.  

**Proposed UI:**

   Almost identical to " **What is the width of the opening here?** "  pull request #5713 - could use the same icon, and an almost identical form. The main change is the wording for the question, which will need agreement, and hopefully comment from others will help with this. My initial thought is; 
   
   **"Think Wheelchair! - What is the width of the biggest gap you could pass through with a wheelchair whilst continuing on this path?"**   
   
   A clarifying comment could be something like;
   
   **"A wheelchair may not be able to negotiate;***
   
   +  **a high kerb,** 
   +  **anything with a steep sideways slope,**
   +  **boggy or  marshy ground,**
   +  **or very rough ground.**  

**When measuring the available gap choose the biggest gap allowing you to continue along the path, but avoiding these obstacles"**

Nick-Tallguy:
closes #5783 

Currently using the shorter wording from @mnalis - have also tested with longer version - I can change to whichever is preferred.  

Will update the wiki as suggested.

Testing on phone


app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -1,6 +1,7 @@
 package de.westnordost.streetcomplete.quests.barrier_opening
 
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.R.string.quest_barrier_opening_width_bollard
```
westnordost, line 4:
```suggestion
```

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
```diff
@@ -38,7 +39,15 @@ class AddBarrierOpening(
...
-    override fun getTitle(tags: Map<String, String>) = R.string.quest_barrier_opening_width_gate
+    override fun getTitle(tags: Map<String, String>) =
+        if (tags["barrier"] == "bollard")
+        {
+            quest_barrier_opening_width_bollard
+        } else if (tags["barrier"] == "block") {
+        quest_barrier_opening_width_bollard
+        } else {
+            R.string.quest_barrier_opening_width_gate
+        }
```
westnordost, line 50:
```suggestion
    override fun getTitle(tags: Map<String, String>) =
        if (tags["barrier"] == "bollard" || tags["barrier"] == "block") {
            R.string.quest_barrier_opening_width_bollard
        } else {
            R.string.quest_barrier_opening_width_gate
        }
```

westnordost (changes requested):

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt
Nick-Tallguy: Thank you

Nick-Tallguy:
wiki updated <https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dbollard#How_to_map>

PR #5808, opened 2024-08-18
Set map images as loaded only after actually loading

Helium314:
When the coroutine running `addOnce` is canceled, `addOnce` will not continue past `withContext`. This results in images being added to `images`, but not to the style, and thus the images never make it to the map.
In SCEE this problem shows more often for some reason, but I also (rarely) noticed missing quest pins in SC.

The simple solution is only adding ids to `images` after they are actually added to the map style.

Btw I think `addOnce` could also return early if `loadIds` is emtpy?


westnordost:
Nice find and fix, thank you!

> Btw I think addOnce could also return early if loadIds is emtpy?

Yes, looks like it.

PR #5818, opened 2024-08-19
callOnUpdated if an updated element does not have a local edit

### LINKED ISSUE #5745: Element is not updated after conflict on upload
In building mode: after the building has been selected, it does not change color. You have to make the entry again so that the building is displayed as an apartment building, for example. 

**How to Reproduce**
 Go into the building mode, select a building and change the building type to an apartment building. After the confirmation, the building type is not displayed.

**Expected Behavior**
The apartment building is displayed after the confirmation

Helium314:
Fixes #5745 as in the latest comments.
(the new test fails without the changes to `MapDataWithEditsSource`)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -84,9 +84,7 @@ class MapDataWithEditsSource internal constructor(
                  *  * The modified versions of all elements in updated are the same before and after
                  *    rebuildLocalChanges, except for the timestamp (expected to have few ms
                  *    difference) and version (never updated locally).
-                 * However, we do want to callOnUpdated if an updated element does not have an associated
-                 * modification, because in that case we cannot know whether the updated version is changed
-                 * compared to the previous version.
+                 *   * No new elements are being added
                  */
                 val deletedIsUnchanged = deletedElements.containsAll(deleted)
                 val hasUnknownElement = updated.any { it.key !in updatedElements }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
@@ -87,7 +87,7 @@ class MapDataWithEditsSource internal constructor(
                  *   * No new elements are being added
                  */
                 val deletedIsUnchanged = deletedElements.containsAll(deleted)
-                val hasUnknownElement = updated.any { it.key !in updatedElements }
+                val hasNewElements = updated.any { it.key !in updatedElements }
                 val elementsThatMightHaveChangedByKey = updated.mapNotNull { element ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -78,14 +78,18 @@ class MapDataWithEditsSource internal constructor(
...
-                 * by MapDataWithEditsSource.
+                 * by MapDataWithEditsSource
                  * This is the case if
                  *  * All keys in deleted are already in deletedElements.
                  *  * The modified versions of all elements in updated are the same before and after
                  *    rebuildLocalChanges, except for the timestamp (expected to have few ms
                  *    difference) and version (never updated locally).
+                 * However, we do want to callOnUpdated if an updated element does not have an associated
+                 * modification, because in that case we cannot know whether the updated version is changed
+                 * compared to the previous version.
```
westnordost, line 89:
```suggestion
                 *   * No new elements are being added
```

Well, this is `MapDataWithEditsSource`. It is not unexpected that there is map data without a modification. E.g. new elements could be added. (Only because we know the internals on how the whole system works together we can say that this usually doesn't happen except for this special case, but for `MapDataWithEditsSource` it is certainly not relevant how the whole system works, but only relevant what's on the interface)

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -78,14 +78,18 @@ class MapDataWithEditsSource internal constructor(
...
                  */
                 val deletedIsUnchanged = deletedElements.containsAll(deleted)
+                val hasUnknownElement = updated.any { it.key !in updatedElements }
```
westnordost, line 92:
```suggestion
                val hasNewElements = updated.any { it.key !in updatedElements }
```

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/MapDataWithEditsSource.kt
```diff
@@ -99,7 +103,7 @@ class MapDataWithEditsSource internal constructor(
 
                 /* nothingChanged can be false at this point when e.g. there are two edits on the
                    same element, and onUpdated is called after the first edit is uploaded. */
-                val nothingChanged = deletedIsUnchanged && elementsThatMightHaveChangedByKey.all {
+                val nothingChanged = deletedIsUnchanged && !hasUnknownElement && elementsThatMightHaveChangedByKey.all {
```
westnordost, line 106:
```suggestion
                val nothingChanged = deletedIsUnchanged && !hasNewElements && elementsThatMightHaveChangedByKey.all {
```

westnordost (commented):
Very nice! Did you test it (beyond the unit test)?

Helium314:
> Very nice! Did you test it (beyond the unit test)?

Not yet, will do a test a bit later.

Helium314:
Works as intended, tested in buildings overlay.

PR #5825, opened 2024-08-21
update screenshots used in the readme file

RubenKelevra:
Due to the switch of renderers in version 59.0 (Alpha 1) I think it's time to update the screenshots.

I tried to reproduce the original screenshots, used in the readme, as close as possible. But some quests and the notes were no longer available, so I replaced them with similar content.

Hope that's fine.


westnordost:
see https://github.com/streetcomplete/StreetComplete/discussions/5812#discussioncomment-10411118

RubenKelevra:
Yes of course. The default is laughably large for some reason.

westnordost:
Or maybe you just have big eyes

westnordost:
...or small eyes? 🤔

RubenKelevra:
> Or maybe you just have big eyes

I'm the Big Bad Wolf ;D

Issue #5852, opened 2024-08-29
"Uhh..." -> "It's a tree stump now" on tree leaf quest


mnalis:
Recently it has been suggested in https://github.com/Helium314/SCEE/issues/622 that two tree-related quests (_"Does this tree have needles or leaves?"_ and _"What is the genus or species of this tree?"_) the OtherAnswer of _**"This is just a tree stump"**_ be implemented (which is done there in https://github.com/Helium314/SCEE/pull/625)

It has been implemented for SCEE in https://github.com/Helium314/SCEE/pull/625, but the former of those quests also exists in StreetComplete,  so the question is would StreetComplete also like that answer implemented? (it is easy to cherry-pick the changes and submit PR here, if wanted)

**Use case**
Sometimes trees get cut down and when you pass them, you see just a tree stump left.

**Proposed Solution**
Would be nice to have an easy way to answer this in the tree leaf type quest (instead of leaving a note, or switching to other editor)

PR #5853, opened 2024-08-29
Implement "Uhh..." -> "It's a tree stump now" on tree leaf quest

mnalis:
Fixes: https://github.com/streetcomplete/StreetComplete/issues/5852

Tested that it [compiles](https://github.com/mnalis/StreetComplete/actions/runs/10616115988) and works as expected

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1132,6 +1132,7 @@ If any lanes are reserved for buses, please leave a note instead."</string>
     <string name="quest_leaf_type_broadleaved">Leaves</string>
     <string name="quest_leaf_type_mixed">Both are present</string>
     <string name="quest_leafType_tree_title">Does this tree have needles or leaves?</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1132,6 +1132,7 @@ If any lanes are reserved for buses, please leave a note instead."</string>
     <string name="quest_leaf_type_broadleaved">Leaves</string>
     <string name="quest_leaf_type_mixed">Both are present</string>
     <string name="quest_leafType_tree_title">Does this tree have needles or leaves?</string>
+    <string name="quest_leafType_tree_is_just_a_stump">"It's just a tree stump"</string>
```
FloEdelmann, line 1135: Please use a typographic apostrophe, like in line 1111.

mnalis: done, tnx... Where have gone good old days when 640KB and ASCII was good enough for everyone :smile:

westnordost (approved):

PR #5855, opened 2024-08-29
Compose statistics

### LINKED ISSUE #5595: Replace ballpit quest-view in user profile with something else

### LINKED ISSUE #5838: Allow screen rotating during (possible) password prompting
**Use case**
Password entering is currently unneccesary difficult

**Proposed Solution**
When the user is prompted to enter a password, e.g. when requesting the OAuth-authorisation (which looks like being delegated by SCEE (SC, too?) to a browser instance), enable screen rotating.
Screen rotating is apparently disabled when executing SCEE (SC, too?) for whatever reason, but consider enabling it during a password prompt.

**Why?**
Some virtual keyboards offer different keyboard layouts depending on whether screen is rotated in landscape or portrait mode. In my case, only in landscape mode I got an ample, more PC-like keyboard layout. Most passwords are memorised not in a direct char by char manner, but rather in shapes the fingers move around or as combination of basic chars and held modifier keys. Thus, to me it is almost impossible to enter a passwort using a virtual keyboard in portrait mode orientation.

Just to be clear: By "enabled screen rotating" I mean the feature that screen content and layout is following the orientation the device is held.

Note: I am not a SC user, but a SCEE one. Hence I first stated this issue there. Ref. https://github.com/Helium314/SCEE/issues/616

westnordost:
fixes #5595

fixes #5838

fixes keyboard obscures part of the login form.

https://github.com/user-attachments/assets/65dc0c7a-defc-4451-ab24-4acd69df75bd


westnordost:
Done. Anyone interested in testing before I dump it into the alpha4?

westnordost:
Not visible in the video because I haven't been mapping at all in the last 7 days: This is also displayed in the new statistics.

kmpoppe:
> Done. Anyone interested in testing before I dump it into the alpha4?

Dunno if I can build the master branch right now, could you throw me an APK with the change in it?

westnordost:
I'll just release the next alpha

kmpoppe:
Weird different coloured stripe at the end of one of the bars in my stats.

Also: I saw that you added statistics for changes from layers (I like that!), what do you think about expanding SC's OSM Wiki entry to include all the layers, what they do, how they select objects et Al, and the linking to that part of the wiki as a documentation link?

HolgerJeromin:
> Weird different coloured stripe at the end of one of the bars in my stats.

Same here on android 14. But with about 9 of 12 visible bars.

rugk:
Sad to see the colorful bubbles that move when you tilt your phone go… they were a very nice gimmick! :eyes:

PR #5868, opened 2024-09-02
New statistics page: Add wikilinks to overlays, that haven't already got one

kmpoppe:
Adding a `wikiLink` to `Things` and `Places` overlays.

Links point to the corresponding overlay description on SC's wiki page that I've added link anchors to prior.

~Can't test or show pictures as test in Android Studio emulator crashes constantly.~ Wireless Debugging to the rescue, there are images now:

| Link in statistics | Wiki |
| --- | --- |
|  |  |

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlay.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlay.kt
@@ -23,7 +23,7 @@ class PlacesOverlay(private val getFeature: (Element) -> Feature?) : Overlay {
     override val title = R.string.overlay_places
     override val icon = R.drawable.ic_quest_shop
     override val changesetComment = "Survey shops, places etc."
-    override val wikiLink = null
+    override val wikiLink = "StreetComplete#ol_places"
     override val achievements = listOf(EditTypeAchievement.CITIZEN)
     override val hidesQuestTypes = setOf(
         AddPlaceName::class.simpleName!!,
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlay.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlay.kt
@@ -23,7 +23,7 @@ class PlacesOverlay(private val getFeature: (Element) -> Feature?) : Overlay {
     override val title = R.string.overlay_places
     override val icon = R.drawable.ic_quest_shop
     override val changesetComment = "Survey shops, places etc."
-    override val wikiLink = "StreetComplete#ol_places"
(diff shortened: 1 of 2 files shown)


matkoniecz:
>  as test in Android Studio emulator crashes constantly.

See https://github.com/maplibre/maplibre-native/issues/2791

If you use different API level in your emulator or you are willing to run some debug build (if would be provided) maybe it may sense to comment there.

westnordost:
There's very little information about this overlay in the one-lined bullet-point and wouldn't consider an anchor to that stable. (Also, the anchor #ol_places doesn't seem to work.)

I'd recommend either changing this to e.g. https://wiki.openstreetmap.org/wiki/Map_features#Shop or create an own wiki page that explains a little more in detail what StreetComplete considers a place.

kmpoppe:
To create headings for all overlays that have a little more substance, would you mind if I list the colours and their values meanings for each overlay in the wiki?

westnordost:
I don't mind, do you think it is useful? For places / things it could be useful to know what should be covered by it and what not.

westnordost:
Alternatively, there could be a StreetComplete/Overlays wiki page analogous to the StreetComplete/Quests wiki page.

kmpoppe:
Yeah, that sounds like a feasible option. I had started on reworking the OL section but it always looked weird ;-)

I'm gonna implement that in the Wiki, once that's done I'll update the links in the PR.

kmpoppe:
@westnordost @mnalis @matkoniecz @FloEdelmann, please take a look at the new [Overlay Wiki page](https://wiki.openstreetmap.org/wiki/StreetComplete/Overlays). If I overlooked something please update the page :) Comments welcome!

westnordost:
Wow, that looks like a lot of effort! I like the legends for the colors!

Some notes:
- I'd also mention the red color for every overlay, even if it has the same meaning for every overlay (=not set or invalid)
- for some colors in some overlays, it would be clearer to name the category of things this is for rather than the concrete tags. E.g. in buildings overlay, blue is for free-standing single-family homes, green is for any kind of outbuilding, yellow for commercial etc.. For the places and things overlay, it is going to be quite a bit of effort to keep this information up-to-date. If you look into `Things.kt` or `Places.kt`, you will find a length comment what kind of things are include and which are not.
- that █  █  is supposed to be a dashed line is really hard to understand. I think it would be easier to understand if there was just a second column that just adds "(dashed)" or something.
- Regarding cycleway and sidewalk overlay, I think it would avoid misunderstandings if you referred to the tags like this `sidewalk:<side>` / `cycleway:<side>` or similar instead. 

Anyway, as far as this PR is concerned, it is done

kmpoppe:
Thanks for the comments, there's some free time coming up, I will look into incorporating your feedback :)

Issue #5872, opened 2024-09-04
Crossing Markings Quest - improve wording


RubenKelevra:
Currently the wording of the Crossing Markings Quest is pretty ambiguous, as it does not specify where the markings are supposed to be:

![Screenshot_2024-09-04-07-31-31-204-edit_de westnordost streetcomplete expert](https://github.com/user-attachments/assets/bcd8b95e-cfe8-46df-a81c-17c678a3ed66)

Here are some examples, which are technically "markings":

![Screenshot_2024-09-04-07-35-47-365-edit_com google android apps maps](https://github.com/user-attachments/assets/68332ff9-e812-4e59-8389-2ab2c3a6d969)
![Screenshot_2024-09-04-07-36-46-217-edit_com google android apps maps](https://github.com/user-attachments/assets/d85d5690-bcab-483c-8856-7424c5146991)
![Screenshot_2024-09-04-07-38-00-025-edit_com google android apps maps](https://github.com/user-attachments/assets/c110ac8f-6767-4cb4-a45a-7dc0e7850f05)
![Screenshot_2024-09-04-07-38-43-812-edit_com google android apps maps](https://github.com/user-attachments/assets/d0d33231-7ea3-4871-a1f6-4386a0fdab97)
![Screenshot_2024-09-04-07-40-25-988-edit_com google android apps maps](https://github.com/user-attachments/assets/81c3f3a2-b880-4c03-a802-37ad6108f677)

But all of them doesn't fit the OSM definition of markings.

So the question should be more along the lines of "Does this crosswalk have painted markings of any kind across the road?"

### Version

SC 58.x / 59.x

kmpoppe:
The source code currently says
```
        /* We only tag yes/no, however, in countries where depending on the kind of marking,
         * different traffic rules apply, it makes sense to ask which marking it is. But to know
         * which kinds exist per country needs research. (Whose results should be added to the
         * wiki page for crossing:markings first) */
```
I think this is part of the problem, there are currently no examples of "valid" `crossing:marking` values - i.e. pictures/graphics - in SC, which might lead to misunderstandings in what the correct answer to the quest is. If the diagrams from the wiki page were used to determine how the crossing is marked, all your examples would most likely be recognised as `no`.

[...]

PR #5874, opened 2024-09-05
Update wording of the crossing markings quest

RubenKelevra:
Fixes https://github.com/streetcomplete/StreetComplete/issues/5872

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1327,7 +1327,7 @@ If there are no signs along the whole street which apply for the highlighted sec
 
     <string name="quest_pedestrian_crossing_island">"Does this crossing have an island?"</string>
 
-    <string name="quest_pedestrian_crossing_markings">"Does this crossing have markings on the ground?"</string>
+    <string name="quest_pedestrian_crossing_markings">"Does this crossing have markings across the roadway?"</string>
 
     <string name="quest_amenityCover_title">Is this covered (protected from rain)?</string>
 
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -1358,7 +1358,7 @@
     <string name="quest_buildingType_tower_description">"أي نوع من الأبراج"</string>
     <string name="quest_buildingType_under_construction">"تحت الإنشاء"</string>
     <string name="overlay_buildings">"مباني"</string>
-    <string name="quest_pedestrian_crossing_markings">"هل توجد علامات على الأرض لمعبر المشاة هذا؟"</string>
+    <string name="quest_pedestrian_crossing_markings">"هل يحتوي هذا المعبر على علامات عبر الطريق؟"</string>
     <string name="overlay_places">"أماكن"</string>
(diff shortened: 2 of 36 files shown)


RubenKelevra:
Sorry, Kate threw in some new lines at the end of some of the files. Can I leave that like this or should I clean it up?

westnordost:
Whooat? Where do all those translations come from? You didn't go through the effort to try to translate them all, did you?

RubenKelevra:
I've battled with ChatGPT for an hour or so and checked with Google Translate if the translation is precise.

westnordost:
Damn, I am sorry for that. It is wasted effort. Translations will be overwritten on the next import from POEditor. When a source string changes, all translations will automatically be marked as "fuzzy" in POEditor, so translators can have a look at the string and adapt the translation.
While you could logon to POEditor and change all those strings there too, I think it is always better to leave it to people actually speaking the language. (I wouldn't mind if you did, but it seems a lot of effort.)

From this PR, I'd prefer if you removed the changes to the translations because they will be overwritten anyway, as I stated.

app/src/main/res/values/strings.xml
```diff
@@ -1327,7 +1327,7 @@ If there are no signs along the whole street which apply for the highlighted sec
 
     <string name="quest_pedestrian_crossing_island">"Does this crossing have an island?"</string>
 
-    <string name="quest_pedestrian_crossing_markings">"Does this crossing have markings on the ground?"</string>
+    <string name="quest_pedestrian_crossing_markings">"Does this crossing have markings that cross the roadway?"</string>
```
westnordost, line 1330: To my ears, this sounds weird. When I think of that something "crosses the street", I think of people or animals actually going from one side to the other. I don't think it can be used in the context of something that is laid out across (from one end to the other) the street.
Didn't you like my suggestion?

westnordost: E.g. bandits that want to rob oncoming vehicles will put a log **across** the road. But the log doesn't cross the road. It doesn't have legs.

RubenKelevra:
Yeah, it's best to have them checked by a native speaker, but they should be fine.

For German, for example, the translation before was actually wrong — I'm a native speaker.

In German, it stated "Fußgängerüberweg" which means pedestrian crossing. But while the quest is confusingly named "quest_pedestrian_crossing_markings" a crossing in OSM could very well be a bicycle crossing or a mixed crossing.

That's why the English string does not specify it - I assume, but the translator saw probably the naming of the variable and assumed "pedestrian crossing" are meant.

If possible, it would be nice if the translators could just look over it and either modify or accept it. I thought that's what going to happen anyway. :)

app/src/main/res/values/strings.xml
RubenKelevra: Well no, it should have been the same as in "app/src/main/res/values-en/strings.xml".

I think that's just written wrong from memory. My bad.

westnordost:
> If possible, it would be nice if the translators could just look over it and either modify or accept it. I thought that's what going to happen anyway. :)

Nope, the translations are only synced one-way, i.e. from POEditor to this repo, never the other way round.

Issue #4334, opened 2022-09-02
remove unneeded quotes in strings


mnalis:
**Use case**

in [app/src/main/res/values/strings.xml](https://github.com/streetcomplete/StreetComplete/blob/v46.0/app/src/main/res/values/strings.xml) some strings values are quoted (e.g. `"Paving stones"`) and some are not (e.g. `Unhewn cobblestone`). There seems to be no rule why it happens sometimes and not other times.

This is confusing (and asking for trouble when copy/pasting parts of them).

**Proposed Solution**

Preferably remove unneeded quotes (`"`) from **all** string values which have it. e.g.
>     <string name="quest_surface_value_sett">"Sett"</string>
should become:
>     <string name="quest_surface_value_sett">Sett</string>
 
(Alternatively, add the quotes to **all** string values. Just make it consistent...)

westnordost:
This is so very very very unimportant that even the super sensitive Android Linter does not report that as a `(weak) warning` and not even as an `info`. Might as well sort all attributes in layout xmls alphabetically or something.

I think when there are no quotes, you cannot use certain characters without like (maybe) `<` but not sure, you'd have to look it up.

PR #5882, opened 2024-09-08
Implements AddBoardName quest

mnalis:
Fixes: https://github.com/streetcomplete/StreetComplete/issues/5871

Tested that it builds in https://github.com/mnalis/StreetComplete/actions/runs/10757025499; and it seems to work.
(made from bus_stop_name quest)

Help wanted:
- [ ] needs an icon (currently `ic_quest_board_name.xml` is copy of regular `ic_quest_label.xml`), as I'm no artist :smiley:

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/board_name/AddBoardName.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/board_name/AddBoardName.kt
@@ -1,0 +1,42 @@
+package de.westnordost.streetcomplete.quests.board_name
+
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.AllCountriesExcept
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.OUTDOORS
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.applyTo
+
+class AddBoardName : OsmFilterQuestType<BoardNameAnswer>() {
+
+    override val elementFilter = """
+        nodes, ways, relations with
+        (
+          tourism=information and information=board
+          and board_type != notice
+          and !board:title
+        )
+        and !name and noname != yes and name:signed != no
+    """
+
+    override val changesetComment = "Determine information board names"
+    override val wikiLink = "Tag:information=board"
+    override val icon = R.drawable.ic_quest_board_name
+    override val achievements = listOf(OUTDOORS)
+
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -793,6 +793,8 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_air_pump_bicycle_repair_station_title">Is there a working bicycle pump here?</string>
     <string name="quest_air_pump_compressor_title">Is there an air compressor available here?</string>
 
+    <string name="quest_board_name_title">"What’s the title of this information board?"</string>
```
kmpoppe, line 796: There are no quotation marks anywhere else in the string, do these really need to be here?

mnalis: I have no idea whether `’` counts, and I had problems with some strings without quotes in the past, so I'm playing it safe since... 

And, it was copied from `<string name="quest_busStopName_title2">"What’s the name of this stop?"</string>` which was pretty similar and had quotes.

Anyway cleanup was suggested before in https://github.com/streetcomplete/StreetComplete/issues/4334, but was deemed _"very very very unimportant"_ so I won't worry too much about that one. Thanks for review though!

app/src/main/java/de/westnordost/streetcomplete/quests/board_name/AddBoardName.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val achievements = listOf(OUTDOORS)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_board_name_title
+
+    override fun createForm() = AddBoardNameForm()
+
+    override fun applyAnswerTo(answer: BoardNameAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        when (answer) {
+            is NoBoardName -> {
+                tags["name:signed"] = "no"
```
westnordost, line 35: I'd rather tag `noname=yes` because if there is no board-title on-site, it really has no name.

mnalis: Oh, right, it was actually suggested that way in issue, but I left default from `bus_stop_name` (which prefers `name:signed=no`. Updated to `noname=yes`

westnordost:
Will create an icon after merge.

westnordost (approved):

westnordost:
Hm, I think the icon is fine. I tried with something like this

board_type

(T for Title or Text), but it doesn't really look good. A mash-up of the info board + label would also be too busy.

paulklie:
> Hm, I think the icon is fine. I tried with something like this

How about replacing the orange background with a green one, since this is more of an "thing" than an place?
Happy to implement, just thought it could be good to make the icons different.

westnordost:
would make sense

PR #5889, opened 2024-09-10
Display object note above quest and overlay answers

### LINKED ISSUE #5794: Use `note` tag "to inform mappers about non-obvious information about an element"
### Use case

I was approached by someone who got frustrated that one of his favorite shops was deleted twice now by people using StreetComplete. The problem is caused by the shop looking like a disused/vacant place outside of opening hours, and the opening hours aren't even fixed: they open at 6 in the morning and close when they're sold out, usually just after lunch time.

Normally, the [`note` tag](https://wiki.openstreetmap.org/wiki/Key%3Anote) can be used to signal to other mappers that there is something special about the object, per the wiki:

> The tag `note=*` is used:
> 
> - to document a general note about an object and
> - as a prefix, e.g. in `note:access=*`, to document why something has a specific `access=*` value.
> 
> [...] a note is generally used on nodes/ways/relations to describe real-world things like "the busstop is actually here but the signpost was knocked down in september 2009"

StreetComplete does not show this extra information to mappers who are editing the object.

### Proposed Solution

The simplest workaround would be to hide places with notes from the quest system, but I can't say that I would be a fan of this.

The best solution seems to me to display notes to the user somehow, before or while answering the quest. To avoid cluttering a bunch of quest UIs, *before* is probably the better option. It could show:

> This object has the following note attached: [...]  
> <kbd>Continue</kbd> <kbd>Hide quest</kbd> <kbd>Close</kbd>

(Where the <kbd>close</kbd> button just goes back to the map view.)

A middle ground would be to check for the presence of the `note` tag, and give the user a link to click so they can decide if they want to continue or not. For example, after answering a quest:

> This place has a note, please review it here: [openstreetmap.org/way/4443697](https://www.openstreetmap.org/way/4443697). Do you still want to apply this change?  
> <kbd>Apply</kbd> <kbd>Cancel</kbd>

Or before:

> This place has a note, please review it here: [openstreetmap.org/way/4443697](https://www.openstreetmap.org/way/4443697). Do you still want to answer this quest?  
> <kbd>Continue</kbd> <kbd>Cancel</kbd>

Further features might include letting the user move the `note` to the `description` tag if it describes the object rather than containing information relevant for mappers ([example](https://www.openstreetmap.org/node/1305725519/history/1)), preventing others from being shown the prompt unnecessarily, but that can be a later step. I would prioritise having some warning (just that a note exists is already enough for me) sooner rather than later, so I can prevent myself from making such a mistake again

kmpoppe:
Resolves #5794.
Removed outdated screenshots ans behaviour, see https://github.com/streetcomplete/StreetComplete/pull/5889#issuecomment-2342710897
Marking this as a draft, looking for feedback.


westnordost:
This should be implemented generally for any quest type form and overlay form and should not use the `hint` thing but be displayed always. Otherwise, one would need to add the code you added to every single quest type where this might be important or turn out to be important later.

Regarding the wording, I think it should be made clear that this is a note from another user. Maybe something like 

Another mapper wrote: _"bla bla bla"_

kmpoppe:
Does the AbstractOsmQuestForm know the element it's displaying? I wasn't sure if that could be used as a hook for that.

mnalis:
> This should be implemented generally for any quest type form and overlay form and should not use the hint thing but be displayed always

I absolutely agree, it would be most useful that `note` should be displayed _always_ when present (and as some quests use hint field, it should not be overwritten).

lgommans:
I like this implementation, thank you for making the proposal Kai!

The font looks small, which should make the maximum note length at 255 characters not be a problem. My phone screen is a bit less tall than yours (still too tall, I can't reach notifications one-handed, but alas) so I checked the screenshots on there and the text is easily readable at arm's length.

Adding quotes and perhaps italics as mentioned above, to make it clear that it's user-supplied text, sound good to me. I was missing something there as well, but thought that having a vertical bar on the left side for a blockquote, like GitHub's quote areas, would take more vertical space than necessary.

kmpoppe:
How do we like this?
grafik

The note is now part of the `fragment_quest_answer.xml` and `AbstractQuestForm` which should make it visible whenever a quest should be answered. This does not include Overlays or fragments like creating a Note (Note API) or moving a node.

I would assume that people have seen the `note` before chosing any of these answers?

westnordost:
I like it! Or maybe "Another mapper noted:" is even better? Maybe put the user's text in quotation marks or like the GitHub quote areas as suggested above?

Also, seeing your screenshot, I guess it often happens that the note is completely unrelated to the question. How can we avoid confusing the user here?

UI-wise, I do wonder whether it would be better to display it in an extra bubble just like in @mnalis mockup in the issue ticket. Semantically, it would make sense - it is not part of the question but a note by a third party.

kmpoppe:
Even though the screenshots not showing, I changed the wording to __"Another person noted:"__, because I'm not too sure that the average SC user knows the term "mapper" and even considers themselfs as one.

This is a "normal" object:


This is an object with a relatively long note (183 characters):


And I have no idea what THIS is tbh, but these are 255 characters including line breaks:


All these look pretty good I think. The scrolling works to still show the quest pin.
Where else would we need the notes? Do they make sense in some or all quest overlays? I could think of the "Things" and "Places" maybe?

westnordost:
Yeah, I think notes should also be shown in the overlays (in all overlays).

app/src/main/res/layout/fragment_quest_answer.xml
```diff
@@ -120,6 +120,58 @@
 
             </LinearLayout>
 
+            <androidx.core.widget.NestedScrollView
```
westnordost, line 123: As far as I can see, this nested scroll view makes no sense here, it should be removed.

app/src/main/res/values/textStyles.xml
```diff
@@ -19,6 +19,18 @@
...
+        <item name="android:textColor">@color/hint_text</item>
+        <item name="android:textStyle">bold</item>
+    </style>
+
+    <style name="TextAppearance.TitleLarge.Italic">
+        <item name="android:textSize">16sp</item>
+        <item name="android:textColor">@color/hint_text</item>
+        <item name="android:textStyle">italic</item>
+    </style>
+
```
westnordost, line 33: (Keeping in mind migration to Jetpack Compose) I'd prefer the following:

- use TitleMedium for the "Another person noted"
- use Body for the hint text and just set the italic + color in the layout itself
- remove these new text styles

westnordost (changes requested):

mnalis:
What would be very useful (and I don't know if that is the case currently), is if the text from that `note=*` tag **could be copied** (to be pasted elsewhere).

Also, if there are http/https links, could they be **made clickable**?

For example, when showing/solving **OSM Note quest, both are possible**, and both have been proven very useful in practice! So perhaps code could be borrowed from there?

small_Screenshot_20240911_145034_StreetComplete

app/src/main/res/layout/fragment_quest_answer.xml
kmpoppe: Resolved by fa183c21d

app/src/main/res/values/textStyles.xml
kmpoppe: Resolved by fa183c21d

kmpoppe:
> What would be very useful (and I don't know if that is the case currently), is if the text from that `note=*` tag **could be copied** (to be pasted elsewhere).
> 
> Also, if there are http/https links, could they be **made clickable**?

Both done in fa183c21d

grafik

westnordost:
Awesome!

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
```diff
@@ -169,6 +170,13 @@ abstract class AbstractQuestForm :
         updateInfoButtonVisibility()
     }
 
+    protected fun setObjNote(text: CharSequence?) {
+        binding.titleNoteLabel.isGone = text == null
+        binding.titleNoteLabel.text = getString(R.string.note_for_object)
+        binding.noteLabel.isGone = text == null
+        binding.noteLabel.text = text
+        decideObjNoteVisible()
+    }
```
westnordost, line 179: This can be simplified. The title is static, it could be specified in the layout XML. Also, no need to set these to gone or not, as the container is already set to gone or not.

```suggestion
    protected fun setObjNote(text: CharSequence?) {
        binding.noteLabel.text = text
        binding.speechbubbleNoteContainer.isGone = text == null
    }
```

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
```diff
@@ -119,6 +119,10 @@ abstract class AbstractOsmQuestForm<T> : AbstractQuestForm(), IsShowingQuestDeta
 
         setTitle(getString(osmElementQuestType.getTitle(element.tags)))
         setTitleHintLabel(getNameAndLocationSpanned(element, resources, featureDictionary))
+        val objNote = element.tags["note"]
+        if (objNote != null) {
+            setObjNote(objNote)
+        }
```
westnordost, line 125:
```suggestion
        setObjNote(element.tags["note"])
```
No need for null check here, as null will be handled by setObjNote

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
```diff
@@ -181,6 +189,10 @@ abstract class AbstractQuestForm :
         updateInfoButtonVisibility()
     }
 
+    private fun decideObjNoteVisible() {
+        binding.speechbubbleNoteContainer.isGone = binding.titleNoteLabel.text == null
+    }
+
```
westnordost, line 195:
```suggestion
```

No need, see above.

kmpoppe: Resolved with afcb138

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
kmpoppe: Resolved with afcb138

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
kmpoppe: Resolved with afcb138, but had to use text.isEmpty() instead of text == null because of calling setObj without prior null check.

kmpoppe:
@lgommans As you brought this up, would you like to make a comment on the looks of it?

westnordost:
To me it seems that it would be more consistent to have the house number / location label above the note. You wrote you liked it (that?) thus far. Why do you think it is better that way?

kmpoppe:
grafik
Swapped in b32740f, you're totally right, looks more consistent that way around.
I'd really like some beta testing of that, do you think we could put this in a 59.0-beta2?

kmpoppe:
> I'd put this in the next major version

Sounds good to me :)

lgommans:
> @lgommans As you brought this up, would you like to make a comment on the looks of it?

I saw the email notifications come in today but only just now had a chance to catch up with the thread. Thanks for inviting my feedback :)

It looks great! Also the extra copying and link-parsing features, I did not even expect it to be so featureful! :rocket:

The only thing I was just wondering about while eyeing the screenshots is the phrasing of "Another person noted". With [notes](https://wiki.openstreetmap.org/wiki/Notes) being another OSM feature and also used in StreetComplete, that phrasing might be confusing for those who aren't used to looking at tags (also considering iD's habit of hiding those). One might wonder "who wrote this, why, how can I comment on it? Can I modify or remove it if it's outdated?"

Some options:

- The heading could be a clickable link, but that is probably UI overload for the user
- Tapping the bubble could show a pop up with an explanation, such as "*This text is in the `note` tag on [this object]<b></b>(https<b></b>://osm.org/node/321). This tag can be edited with an OSM editor such as [Vespucci]<b></b>(...) or [iD]<b></b>(...).*", but that's another UI to design and I really appreciate all the work everyone involved has already put into this and I don't want to make more when I'm not even sure it's the best solution  
- A wording like "**This object has a `note` attached:**" could be used instead
- Leave it as-is; it's not a significant problem anyway

What do you think?

westnordost:
None of the suggestions I think would make it less confusing, but more. "Noted" is used here in the sense of "remarked" or "commented".

(But overall I personally find it better to use "noted" because it hints at - for advanced users who know the tag - that this might indeed be the `note` field.)

lgommans:
Okay, then let's just leave it as it is :)

Thanks again for everyone's work on it!

PR #5898, opened 2024-09-13
Use platform CancellationSignal

Isira-Seneviratne:
The Jetpack [`CancellationSignal`](https://developer.android.com/reference/androidx/core/os/CancellationSignal) class has been deprecated as the Core library now requires at least API level 19.


westnordost:
Ok, thanks

PR #5900, opened 2024-09-15
Use System.currentTimeMillis()

Isira-Seneviratne:
Replace the `nowAsEpochMilliseconds()` method with the native [`System.currentTimeMillis()`](https://developer.android.com/reference/java/lang/System#currentTimeMillis()) method.


westnordost:
No, we want to get away from Java.

PR #5903, opened 2024-09-16
Check Date for opening_hours:signed

### LINKED ISSUE #4273: Opening hours check date is updated when marked as not signed

kmpoppe:
Resolves #4273, ref #4276.

Is this, what we were looking for?

If the answer is that `opening_hours:signed=no` (either from quest/opening_hours or quest/opening_hours_signed) we add `check_date:opening_hours:signed`.

If the answer is that `opening_hours:signed=yes` (either from quest/opening_hours or quest/opening_hours_signed), that means `opening_hours:signed` is removed and `opening_hours=*` are set, we remove `check_date:opening_hours:signed` and try to set `check_date:opening_hours` to the last edit timestamp so that the apply check for opening_hours works.

quest/opening_hours_signed checks for this info being older than a year OR the element being last edited over a year.

quest/opening_hours now either filters for `opening_hours:signed != no` OR (`opening_hours:signed = no` AND `check_date:opening_hours:signed` older than 1 year)

Testing with applicable objects, the behaviour seems to be programmatically correct, yet, I am unable to formulate the Tests correctly, I might need help in understading how these work. I have thrown away my changes in the tests and will work on them again later -> WIP PR.

Also, please provide feedback on the setting/removing/interpretation of the various check dates.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -168,6 +168,7 @@ class AddOpeningHours(
         if (answer is NoOpeningHoursSign) {
             tags["opening_hours:signed"] = "no"
             tags.updateCheckDateForKey("opening_hours:signed")
+            tags.removeCheckDatesForKey("opening_hours")
             // don't delete current opening hours: these may be the correct hours, they are just not visible anywhere on the door
         } else {
             val openingHoursString = when (answer) {
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -14,6 +14,7 @@ import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.
 import de.westnordost.streetcomplete.osm.Tags
 import de.westnordost.streetcomplete.osm.isPlaceOrDisusedPlace
 import de.westnordost.streetcomplete.osm.opening_hours.parser.isSupportedOpeningHours
+import de.westnordost.streetcomplete.osm.removeCheckDatesForKey
 import de.westnordost.streetcomplete.osm.updateCheckDateForKey
 import de.westnordost.streetcomplete.osm.updateWithCheckDate
 
@@ -116,7 +117,7 @@ class AddOpeningHours(
           or barrier
           or amenity ~ toilets|bicycle_rental
         )
-        and opening_hours:signed != no
+        and (opening_hours:signed != no or (opening_hours:signed = no and opening_hours:signed older today -1 years))
     """).toElementFilterExpression() }
 
     override val changesetComment = "Survey opening hours"
@@ -166,7 +167,7 @@ class AddOpeningHours(
     override fun applyAnswerTo(answer: OpeningHoursAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
         if (answer is NoOpeningHoursSign) {
             tags["opening_hours:signed"] = "no"
-            tags.updateCheckDateForKey("opening_hours")
+            tags.updateCheckDateForKey("opening_hours:signed")
             // don't delete current opening hours: these may be the correct hours, they are just not visible anywhere on the door
         } else {
             val openingHoursString = when (answer) {
@@ -178,6 +179,7 @@ class AddOpeningHours(
             tags.updateWithCheckDate("opening_hours", openingHoursString)
             if (tags["opening_hours:signed"] == "no") {
                 tags.remove("opening_hours:signed")
+                tags.removeCheckDatesForKey("opening_hours:signed")
             }
         }
         tags.remove("opening_hours:covid19")
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
@@ -11,6 +11,7 @@ import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.
 import de.westnordost.streetcomplete.osm.Tags
 import de.westnordost.streetcomplete.osm.getLastCheckDateKeys
 import de.westnordost.streetcomplete.osm.isPlaceOrDisusedPlace
+import de.westnordost.streetcomplete.osm.removeCheckDatesForKey
 import de.westnordost.streetcomplete.osm.setCheckDateForKey
 import de.westnordost.streetcomplete.osm.toCheckDate
 import de.westnordost.streetcomplete.osm.updateCheckDateForKey
@@ -27,7 +28,7 @@ class CheckOpeningHoursSigned(
         nodes, ways with
           opening_hours:signed = no
           and (
-            $hasOldOpeningHoursCheckDateFilter
+            $hasOldOpeningHoursSignedCheckDateFilter
             or older today -1 years
           )
           and access !~ private|no
@@ -37,8 +38,8 @@ class CheckOpeningHoursSigned(
           )
     """.toElementFilterExpression() }
 
-    private val hasOldOpeningHoursCheckDateFilter: String get() =
-        getLastCheckDateKeys("opening_hours").joinToString("\nor ") {
+    private val hasOldOpeningHoursSignedCheckDateFilter: String get() =
+        getLastCheckDateKeys("opening_hours:signed").joinToString("\nor ") {
             "$it < today -1 years"
         }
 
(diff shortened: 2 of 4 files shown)


westnordost:
Thank you! Will look over this PR soon.

The tests are nothing special, just normal (Java-like) unit tests. If you are confused about `OsmElementQuestType::answerAppliedTo`, that's just a convenience function for the tests, to get the changes that are made as a `Set` of `StringMapEntryChange`s for easy comparison with `assertEquals` in the tests.

kmpoppe:
Thanks for the info!

After throwing away my prior changes and re-doing them (thinking about them, being actually more awake than before) the tests I added/changed now (0d3626f and ea10d1f) work as expected.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -166,7 +167,7 @@ class AddOpeningHours(
     override fun applyAnswerTo(answer: OpeningHoursAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
         if (answer is NoOpeningHoursSign) {
             tags["opening_hours:signed"] = "no"
-            tags.updateCheckDateForKey("opening_hours")
+            tags.updateCheckDateForKey("opening_hours:signed")
```
westnordost, line 170: If we don't update or delete the opening hours check date here, the opening hours quest will be asked again right away, wouldn't it?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -116,7 +117,7 @@ class AddOpeningHours(
           or barrier
           or amenity ~ toilets|bicycle_rental
         )
-        and opening_hours:signed != no
+        and (opening_hours:signed != no or (opening_hours:signed = no and opening_hours:signed older today -1 years))
```
westnordost, line 120: Why? doesn't this lead to this quest being asked also if not signed (instead of the CheckOpeningHoursSigned)?

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
```diff
@@ -27,7 +28,7 @@ class CheckOpeningHoursSigned(
         nodes, ways with
           opening_hours:signed = no
           and (
-            $hasOldOpeningHoursCheckDateFilter
+            $hasOldOpeningHoursSignedCheckDateFilter
```
westnordost, line 31: How does this behave with current tagging, though? E.g.

```
shop=kiosk
opening_hours:signed=no
check_date:opening_hours=2022-02-02
```
?

Looks to me that for the tags above (as previously tagged with StreetComplete), this quest is currently not created. But it should, shouldn't it?

westnordost (changes requested):
It should also be documented by the tests how the quest(s) behave with current tagging, e.g. 

```
shop=kiosk
opening_hours:signed=no
check_date:opening_hours=2022-02-02
```

In other words, taking care of the "migration path" from the old tagging to new tagging.

kmpoppe:
Currently AFK from my programming machine for a week, I shall check your comments then.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
kmpoppe: If I read `filter by lazy` correctly, `and (!opening_hours or opening_hours older today -1 years)` is within a block that's been ANDed together with L120, that is only true when OH:signed != no OR when OH:signed = no AND OH:signed is older than a year, which will NOT be true, if we just set OH:signed = no.

The quest would therefore not be asked, even if we came to that point when OH was older than a year.

At least that's what I was thinking. I might need to add tests to confirm that it works as expected.

kmpoppe:
```
    @Test fun `isApplicableTo returns false if the opening hours check date is older than 1 year, while they are not signed and signed check date is less than 1 year old`() {
        assertFalse(questType.isApplicableTo(node(
            tags = mapOf(
                "shop" to "supermarket",
                "name" to "Supi",
                "opening_hours:signed" to "no",
                "check_date:opening_hours" to RelativeDate(-420f).date.toCheckDateString(),
                "check_date:opening_hours:signed" to RelativeDate(-180f).date.toCheckDateString()
            ),
            timestamp = RelativeDate(-365f).date.toCheckDateString().toCheckDate()?.toEpochMilli()
        )))
    }
```
Correctly asserts false

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
westnordost: Can you add test cases that test:

- `isApplicableTo returns false if the opening hours check date is older than 1 year, while they are not signed` (i.e. no signed check date is set)
- `isApplicableTo returns false if last edit date is older than 1 year, while they are not signed` (not any check date is set)

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSigned.kt
westnordost: (Would make sense to add a test case for this and see if it runs through or not)

kmpoppe:
will add tests as requested in a few.

app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
kmpoppe: Tests added

kmpoppe:
Tests added.

CheckOpeningSigned now checks for both check_dates (or older 1 year), that why for second test mentioned (in my eyes) correctly asserts true not false because it's older than a year and has no check_dates at all.

app/src/test/java/de/westnordost/streetcomplete/quests/opening_hours_signed/CheckOpeningHoursSignedTest.kt
```diff
@@ -67,6 +69,16 @@ class CheckOpeningHoursSignedTest {
...
+    @Test fun `is applicable if the opening hours not signed and only check date for OH exists`() {
+        assertTrue(questType.isApplicableTo(node(
+            tags = mapOf(
+                "name" to "rundumdieuhr kiosk",
+                "opening_hours:signed" to "no",
+                "check_date:opening_hours" to "2021-03-01"
+            ),
+            timestamp = "2000-11-11".toCheckDate()?.toEpochMilli()
+        )))
+    }
```
westnordost, line 81: This test doesn't test what you want it to test. You want it to test that it takes the older of the two of `check_date:opening_hours` and `check_date:opening_hours:signed`, but to test this, `timestamp` would need to be "now" (i.e. the default). Otherwise the quest would anyway be applicable because the last edit date is 24 years old.

```suggestion
    @Test fun `is applicable if the opening hours not signed and only check date for OH exists`() {
        assertTrue(questType.isApplicableTo(node(
            tags = mapOf(
                "name" to "rundumdieuhr kiosk",
                "opening_hours:signed" to "no",
                "check_date:opening_hours" to "2021-03-01"
            )
        )))
    }
```

westnordost (changes requested):
Tests for
- that `check_date:opening_hours` is removed when answering `NoOpeningHoursSign`
- that `check_date:opening_hours:signed` is removed when answering with anything else
are missing in `AddOpeningHoursTest`

When you add new behavior to tested code, always also test this new behavior (not only modify the tests so that they run through green again). If the tests are not kept up-to-date with the implementation, the tests don't make sense.

Also, the new tests you added are puzzling to me.

`AddOpeningHours` quest should **NEVER** be applicable while `opening_hours:signed=no`, because when it is not signed, it should never be asked. Instead, we have `CheckOpeningHoursSigned` which is asked in its place. When the user answered that it **is** signed **now**, only then `AddOpeningHours` should be asked.

Instead of this desired behavior, the tests seem to test that the quest is being asked also for places with unsigned opening hours as long as it was checked more than one year ago.

kmpoppe:
With the recent changes from @mnalis, this PR has become out of date (i.e. conflicted), so I'll close this and reopen a new one when I had time to reflect on the function again.

PR #5915, opened 2024-09-21
Add roofs to maxheight quest

### LINKED ISSUE #5912: Show max height quests for road under roof
**Use case**
I noticed that the max height quest doesn't show up for service roads below `building=roof`, despite very common:

Thats the usual combination used for gas stations, which usually have a height sign on the edge of the roof. 

**Proposed Solution**
It would be nice if that could be added, for as long as there's not `service=driveway` tag on the road, to avoid having all car ports show up as quest.

RubenKelevra:
- `building=roof` will show up for the max height quest
- `man_made=pipeline` will now also properly show up for this quest

Fixes #5912

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/max_height/AddMaxHeight.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/max_height/AddMaxHeight.kt
@@ -59,10 +59,13 @@ class AddMaxHeight : OsmElementQuestType<MaxHeightAnswer> {
 
     private val bridgeFilter by lazy { """
         ways with (
-            highway ~ ${(ALL_ROADS + ALL_PATHS).joinToString("|")}
-            or railway ~ rail|light_rail|subway|narrow_gauge|tram|disused|preserved|funicular|monorail
-          ) and (
-            bridge and bridge != no
+            (
+              highway ~ ${(ALL_ROADS + ALL_PATHS).joinToString("|")}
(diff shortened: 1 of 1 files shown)


westnordost:
Thanks, and thanks for your fix, I think you are right!

westnordost (approved):

RubenKelevra:
@westnordost would you mind testing this change? I think it's not working. I've tried it and asked someone else to try it, and they confirmed it's now showing up for them either.

### Case one:

Service road:

```
highway=service
lit=yes
smoothness=good
surface=paving_stones
``` 

Roof:

```
building=roof
layer=1 
```

Location: https://www.openstreetmap.org/#map=19/51.208604/7.231552&layers=N


### Case 2:

Service road:
```
highway=service
surface=asphalt
``` 

Roof:
``` 
building=roof
layer=1
source=Bing
``` 

Location: https://www.openstreetmap.org/#map=19/48.482845/12.000903&layers=N


Versions tested: v60-alpha1 and v59.3

## Screenshot

Screenshot_2024-11-16-19-47-59-009-edit_de westnordost streetcomplete

mnalis:
@RubenKelevra is some quest is not working properly, I'd suggest opening a new issue (and just referencing the PR); instead of commenting on old, merged PRs (or old closed issues, for that matter) - where such report is likely to be overlooked.

westnordost:
I found the issue. The `bridgeFilter` expression is really just for linear bridge-features, not for areas.

PR #5917, opened 2024-09-22
Patch ref:FR:SIRET, ref:FR:SIREN and ref:FR:NAF when changing place

waterced:
Currently, changing a place using the overlay removes most of old tags that no longer apply to the new place, such as opening_hours, brand, and many more. The same should apply to the french SIRET, SIRENE and NAF codes, as discussed [here](https://forum.openstreetmap.fr/t/attention-au-layer-lieux-sur-streetcomplete/27288) (in french).

https://en.wikipedia.org/wiki/SIRET_code
https://en.wikipedia.org/wiki/SIREN_code
https://wiki.openstreetmap.org/wiki/Key:ref:FR:NAF

This patch addresses the situation


HolgerJeromin:
Perhaps you could add the french specific entries in a new line.

matkoniecz:
note https://github.com/streetcomplete/StreetComplete/blob/7f0cdd3f75a862109a177f59ccc6b9c3bb5eb22a/app/src/main/java/de/westnordost/streetcomplete/osm/Place.kt#L262

this should be updated at https://github.com/mnalis/StreetComplete-taginfo-categorize/ - here it will be replaced and lost on the next update

matkoniecz:
@waterced Do you plan to make PR there?

And thanks for working on it! Just that it needs to be addressed in a bit different place.

waterced:
Hi @matkoniecz , thanks for the feedback, yes I'll make a new PR there!

waterced:
The PR there : https://github.com/mnalis/StreetComplete-taginfo-categorize/pull/27

matkoniecz:
thanks! I will close this one then

Issue #5101, opened 2023-06-28
Automatically split long ways because they are hard to survey


Zaczero:
**Use case**

I am sometimes hesitant to answer quests on VERY long ways. I suppose I am not the only person. It would be very convenient for such ways to be automatically split. I know there is a split feature but deciding where to split from the pedestrian perspective is difficult in such cases. Doing it automatically would be a welcome feature.

**Proposed Solution**

When loading a quest, the way is automatically split above some threshold. Only the nearest part is then shown to the user as part of a quest. The split way is submitted to OSM once user answers the quest.

There should be a different limit for footways and roads. The split should prefer an intersection location, if such is nearby the threshold distance.

westnordost:
I don't see how this could be done in a meaningful way automatically. And certainly, we don't want the app to be acting on its own, splitting ways "just in case" the user is going to answer it differently for the sections presented. The changes are made in the user's name after alll.

matkoniecz:
> but deciding where to split from the pedestrian perspective is difficult in such cases

Split into surveyed and not surveyed part

Zaczero:
If you both agree in this being a non-issue, maybe it's just me.

amandasaurus:
I have this problem often when hiking, and I would like an easier way to split long ways. I don't think it should be done automatically, but I think it should be easier & quicker than it is now.

(a) When I'm hiking in the forest, some of the OSM ways are large, and I'm also uncertain about adding a tag to the whole thing in one second. If I could split a way into ~ 100m (?) sized ways, and then add the surface many times. A downside of this is that it makes many small OSM ways, which may, or may not, be needed in the end.

(b) Sometimes I walk past a junction, and can see the first ~50m(?) of the track. and can see the surface, or smoothness. I'm not walking down that path, so I can't say for certain. I'd like to be able to split the way into the first X m (?) quicker & easier than currently.

I think (b) (“Quickily split off first X m”) is few downsides, and would be benefitial. (a) (“Quickly split into X m sized chunks”) has downsides, and I'm unsure if the outways the benefits

mnalis:
> (a) (“Quickly split into X m sized chunks”) has downsides, and I'm unsure if the outways the benefits

yes, that one is definitely too problematic IMHO, as it would _both_ create a ton of unneeded chunks _and_ would increase mapper load significantly.

> (b) (“Quickily split off first X m”) is few downsides, and would be benefitial. 

[...]

Issue #4588, opened 2022-10-31
"What’s the height limit below the bridge"-quest does something wrong


habi:
I'm not using StreetComplete, because I'm on iOS but I watch notes in my vicinity and in Switzerland.

There's a bunch of recent notes complaining that 'the height limit below the bridge' cannot be answered because there's no bridge.

Examples:

- https://www.openstreetmap.org/note/3422505
- https://www.openstreetmap.org/note/3422474
- https://www.openstreetmap.org/note/3421397
- https://www.openstreetmap.org/note/3399306

Some of the rather long ways actually do pass under a bridge, the notes in Bern show ways which have a `max_height` of 20 to 30 m :)

Could this quest be only shown closer to the bridges?

peternewman:
> Examples:
> 
>     * https://www.openstreetmap.org/note/3422505

The bridge here seems pretty close to me:
https://www.openstreetmap.org/way/48443730

> Could this quest be only shown closer to the bridges?

There's already an option and a note to split the way, at which point it will only actually ask for the bit under the bridge.

https://github.com/streetcomplete/StreetComplete/blob/c000b8e6539e553aed6c20eafceec2c78db5d4cd/app/src/main/res/values/strings.xml#L1045

matkoniecz:
I have solved all linked notes by splitting way (which could be also done in StreetComplete)

habi:
My observations based on the three notes I just went through: https://www.openstreetmap.org/changeset/128331354

In all these situations (at least in Bern I *know*, all the other notes it's probable) the `max_height` is larger than 4 meters.
In this situation the height is not signed in Switzerland (see Art. 21 of [our signage regulation](https://fedlex.data.admin.ch/filestore/fedlex.data.admin.ch/eli/cc/1979/1961_1961_1961/20210101/de/pdf-a/fedlex-data-admin-ch-eli-cc-1979-1961_1961_1961-20210101-de-pdf-a-11.pdf)), does the user get  suggested to use [`maxheight=default`](https://wiki.openstreetmap.org/wiki/Key:maxheight#Non-numerical_values) in this case?

matkoniecz:
> does the user get suggested to use [maxheight=default](https://wiki.openstreetmap.org/wiki/Key:maxheight#Non-numerical_values) in this case?

yes, if user answers correctly then it will be applied

Luzandro:
[...]

PR #5918, opened 2024-09-22
Remove "bridge" from height limit texts, as pipelines and roofs will now appear as well

RubenKelevra:
Additional fix for #5912

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1143,9 +1143,9 @@ If any lanes are reserved for buses, please leave a note instead."</string>
     <string name="quest_lit_title">"Is it lit here?"</string>
 
     <string name="quest_maxheight_title">"What’s the height limit here?"</string>
-    <string name="quest_maxheight_below_bridge_title">What’s the height limit below the bridge?</string>
+    <string name="quest_maxheight_below_bridge_title">What height limit is indicated here?</string>
     <string name="quest_maxheight_sign_title">"What height limit is indicated here?"</string>
-    <string name="quest_maxheight_sign_below_bridge_title">What height limit is indicated below the bridge?</string>
+    <string name="quest_maxheight_sign_below_bridge_title">What height limit is indicated here?</string>
     <string name="quest_maxheight_answer_noSign">There is no sign…</string>
     <string name="quest_maxheight_split_way_hint">If it doesn’t apply for the whole way, consider answering “%s”.</string>
     <string name="quest_maxheight_unusualInput_confirmation_description">This height looks implausible. Are you sure that it is correct?</string>


westnordost:
What? No, we already have a text for that, `quest_maxheight_sign_title`

westnordost:
Ah, but I see the issue... hmm...

westnordost:
It is now impossible to tell apart the situation "below a bridge", "below an overhead pipeline" and "below a building roof". So, I guess, the string should be removed for all. I hope this does not lead to confusion.

RubenKelevra:
I don't think so. I mean if you're standing on top of a bridge and it marks the road crossing blow on the map, I doubt anyone would think "there's no sign here on the bridge, let's put this in". :)

mnalis:
Although, I seem to recall discussions there was confusion even earlier, if the section of the road going under the bridge was long (i.e. the quest would pop up a kilometer before the bridge is even visible). I do not recall if a proper fix was devised; if not, making the text even more generic is likely to be even more confusing.

westnordost:
But there is no alternative, or is there?

mnalis:
> But there is no alternative, or is there?

For the original confusion, or the new one?

- for the original confusion (e.g. https://github.com/streetcomplete/StreetComplete/issues/4588), perhaps quest might be skipped if road segment going under the bridge is too long (maybe `> 1km` or something like that?) I don't know how complicated that would be, but it would reduce confusion (at the expense at not asking the quest in some situations)

- for this newly increased confusion, `getTitle()` can be changed programmatically, like e.g. in `air_pump/AddBicyclePump.kt`. That would require separate translations, though, but would reduce the confusion to (much smaller) original levels.

Also quest description (the one behind info button) could be improved to at least mention all the things that this quest may apply to, and the fact that they may be somewhat far away.

RubenKelevra:
> Although, I seem to recall discussions there was confusion even earlier, if the section of the road going under the bridge was long (i.e. the quest would pop up a kilometer before the bridge is even visible). I do not recall if a proper fix was devised; if not, making the text even more generic is likely to be even more confusing.

This highlights the main training issue I currently see in SC. Many beginners tend to walk to a quest marker, click on it, and then ignore the map, because the answer field opened up and draws their attention. Leading to them applying the information for the current situation the specific spot to the entire way. For longer ways, especially those over 200 meters, users who behave correctly still have to rely on memory, which becomes unreliable over such distances, leading to incorrect tagging.

My suggestion to address this could be implementing a prompt when a way exceeds 200 meters. The prompt would encourage them to split the way into smaller, manageable sections and only tag what they can visually confirm at the time of answering. This would prevent users from having to rely on memory, which often results in inaccuracies. By guiding users to divide the way into sections, they can tag each part with the appropriate information and avoid applying localized data across an entire way.

This approach could significantly improve data accuracy and reduce the common mistakes I’ve observed in OSM contributions from StreetComplete users.

The only draw back I see is that ways may be split up unnecessarily in some circumstances, but to be fair, that's better than having wrong information in the database IMHO.

RubenKelevra:
@westnordost thinking about it... It should be pretty easy to implement:

- If the way of a quest is longer than 200m, show a popup:

"The way on the map is pretty long! Please mark on it how far you can see from your current location" (with an "OK" button on the right and a "skip" button on the left) 

- Then transition to the regular "split way" function, but limit it to two cuts and require at least one cut.
- Afterwards reopen the quest for the middle section (on two cuts) or the small section (on one cut)

westnordost:
Nobody reads prompts of the "are you sure" type, though.

westnordost:
> for this newly increased confusion, getTitle() can be changed programmatically, like e.g. in air_pump/AddBicyclePump.kt. That would require separate translations, though, but would reduce the confusion to (much smaller) original levels.

No, it can't, because passed into `getTitle` are the tags of the object for which the quest is created. In this case, the road below the roof or bridge. There is nothing tagged on that section of road that would help in deciding whether that section of road is below a bridge, roof or overhead pipeline.

mnalis:
> There is nothing tagged on that section of road that would help in deciding whether that section of road is below a bridge, roof or overhead pipeline.

Ah, that is unfortunate. Splitting the quest into three separate similar-looking quests would work instead I guess, but seems quite inelegant solution :thinking: 

> If the way of a quest is longer than 200m, show a popup

While cutting would in the end reduce the confusion, I'm afraid that cure might be quite _worse_ then the original illness  **_if applied generally_** (spamminess, significant work multiplication, etc. See related https://github.com/streetcomplete/StreetComplete/issues/5101 for some of the issues). 

However, applied just for the (rather rare) `maxheight` quest on too long ways there, such force-splitting might work (especially if they're instructed to just cut on the sides of the bridge)

goldfndr:
Instead of getTitle, could getLevelLabel (or a caller or other function) be overridden with another (or optional) parameter that would pass the overhead object and supplement the italicized description?

mnalis:
I'd add _", overhead pipes"_ before _"etc..."_, but yes; it should help  (provided people read it and are not overly confused)

PR #5936, opened 2024-10-01
Clarify wording of the crossing quest

schra:
I was confused by the wording of the answer "No, but crossing is possible" to the question "Is there a crossing here?". I needed to check in the source code of StreetComplete to understand what the answer means.

I propose to change the wording of the answer. My proposed wording mentions informal trails and adds a picture of such an informal trail, which would have helped me.


app/src/main/res/drawable-hdpi/informal_crossing.jpg
schra, line 1: I see that there are multiple of these res/drawable-*/ folders. Do I just copy-paste this pic to the other folders as well? Or do I only keep this pic in one of the folders? Which?

schra, line 1: This is a cropped version of https://wiki.openstreetmap.org/wiki/File:West_Highland_Way_crossing_road_-_geograph.org.uk_-_3986498_(cropped).jpg

app/src/main/res/values-de/strings.xml
schra, line 1: I have never worked with these internationalization files. Do I just keep the other languages' xml files unchanged?

app/src/main/res/drawable-hdpi/informal_crossing.jpg
mnalis: you need differently sized ones. See https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#new-photos for more information (and helpful ruby script to generate them).

mnalis: credits  for images should be updated in `app/src/main/res/authors.txt`

app/src/main/res/values-de/strings.xml
mnalis: Generally you don't touch the translations in the git code, they will be pulled up automatically later; see https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app

But then, in cases when translation needs to be forced, that is usually accomplished by using different string id. Without that change, it is likely that many/most of the translations will remain unchanged.

schra: Thanks for your swift response :) I will follow up on this tomorrow

app/src/main/res/values/strings.xml
```diff
@@ -999,8 +999,9 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
 
     <string name="quest_crossing_title2">Is there a crossing here?</string>
     <string name="quest_crossing_yes">Yes (e.g. the curb is lowered here or there are markings)</string>
-    <string name="quest_crossing_no">No, but crossing is possible</string>
+    <string name="quest_crossing_no">No, but there is an informal crossing (e.g. informal trail)</string>
```
westnordost, line 1002: The wording seems redundant to me (two times "informal").

"Informal crossing" is not a common term and the "OSM meaning" of that term is, well, OSM specific. E.g. the crossing as shown in the title image of this youtube video (found in one of the first results of a duckduckgo search for "informal crossing") https://www.youtube.com/watch?v=ctNrjp_m4To is **NOT** an `crossing=informal`, but a normal crossing (with no markings).

So, I want to avoid this term.

"Informal trail" on the other hand is better known. But not all `crossing=informal` are with informal paths. So, the hint text should reflect that, i.e. not contain "usually looks like..."

So, I'd suggest:

quest_crossing_no="No, but crossing is possible (e.g. at informal trail)"
quest_crossing_hint="A crossing with an informal path:"

westnordost (changes requested):
Regarding the size of the hint picture, best use the same height as for other hint pictures.

E.g. the hint pictures for tactile paving have a height of:
mdpi - 160
hdpi - 240
xhdpi - 320
xxhdpi - 480

(Don't upscale pictures, blurry pictures are not accepted. But I see, the source photo has a resolution of 718x718px, so that should work.)

app/src/main/res/drawable-hdpi/informal_crossing.jpg
schra: Thanks for the hint to the script :) 3e2091c122d49fcdb3cc2588183f4c31f23bdfd4

schra: 3e2091c122d49fcdb3cc2588183f4c31f23bdfd4

app/src/main/res/values-de/strings.xml
schra: You are saying that I should not touch values-*/strings.xml but I can (should) touch values/strings.xml, right? 4a5d6033922cb9562d065bec3a5f177bf01331bd

app/src/main/res/values/strings.xml
schra: Good point. I'd also like to change the wording "but crossing is possible" a bit more towards what https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dinformal says.

What about changing "informal crossing" to "intended crossing" like this:

quest_crossing_no="No intended crossing, but pedestrians commonly cross here (e.g. informal trails)"
quest_crossing_hint="An informal trail crossing a road:"

app/src/main/res/values-de/strings.xml
westnordost: yes

westnordost (commented):
Uhh, what happened with that picture, why does it look all ugly now? Not only the colors/contrast, but also it looks super grainy now:

imagen

(zoomed in to 300%)

see what I mean? When resizing, use some interpolation, like e.g. cubic.

app/src/main/res/values/strings.xml
westnordost: I don't know... it's a lot of text and I don't really get what it is supposed to clarify. You know that the wiki article was mostly written by us, right?

I don't quite get what is the issue with the current wording, what exactly confused you?

app/src/main/res/values-de/strings.xml
schra: okay thanks :thumbsup:

schra:
Does https://github.com/streetcomplete/StreetComplete/pull/5936/commits/1d5d518b91b7c0153df5eff9fdc6d50921d9b135 fix this for you?

app/src/main/res/values/strings.xml
schra: I am new to the OSM world and the first time I used StreetComplete, there was a situation where I was unsure which answer to choose for "Is there a crossing here?". Here is the pic of the crossing I took:

Screenshot_20241002-232610

You can see that the sidewalk looks exactly the same as the road (there are no markings, no lowered curbs, no traffic signs). I wondered if this constitutes a crossing or not. Now after doing some research I think this is highway=crossing with crossing=unmarked. But as a newbie I didn't know that. Reading the answers in the app quiz I wasn't sure if I should pick "Yes (e.g. the curb is lowered here or there are markings)" or "No, but crossing is possible". I mean the curb is not lowered and there are no markings, but crossing also is possible.

Looking back I think two things would have helped me:

* An example of crossing=informal

* Being more explicit that crossings are also crossings if they do not have any markings, as long as they are intended by the municipality.

  That's why I am proposing to add the "intended crossing" or "informal crossing" wording somewhere.
  Maybe a better approach would be to add the picture from above with a text like "Crossings don't need any markings:"?

Does this make sense? Do you have any ideas on making this situation of a pedestrian crossing more clear? Or do you want to only incorporate the crossing=informal example?

westnordost: Ah, thank you for this explanation. Well, as far as I know, for situations like above - da scheiden sich die Geister - there is no consensus in the community as far as I am aware.

Some argue, there shouldn't be a `highway=crossing` at all, others argue that the "Road" in red in your example, even if just some sort of alley, could just as well be asphalted and lowered, in which case it is definitely a `highway=crossing`, so tagging it always is fair.

Reviewing what I remember of the discussion(s), I tend to agree with the latter group, although while it is not clear in the wiki documentation, it would be problematic to propagate one over the other. So, this is the reason for why the example in your picture is not really mentioned at all in StreetComplete.

It is also the reason why StreetComplete never asks this quest for intersections of footways with `highway=service` in order to avoid asking users at all for situations like pictured above. Isn't the above a driveway of sorts? So, I am puzzled why you were asked at all.
Anyway, I guess `highway=residential` roads that look like above also exist, although rarely.

I understand the issue better now and your suggestions make sense. Maybe using "intended" in wording indeed helps making it clearer:

> **Is there a crossing here?**
> 
> - Yes, intended as crossing
> - Not intended as crossing, still possible to cross (e.g. informal trails)
> - No, crossing is prohibited or impossible

I gather that including the examples in the "yes" answer may actually rather confuse than help, as no markings are necessary for something to be a crossing. And as mapping sidewalks as separate ways instead of properties of the roadway is becoming more and more popular, we have more and more cases where there are just "implicit" crossings on every intersection, even if it is deep in a residential area (with no markings, traffic signals etc). And sometimes, the curbs are just not lowered there.

schra: Thanks for your detailed answer :)

> Isn't the above a driveway of sorts? So, I am puzzled why you were asked at all.

I will go and check sometime this week. I don't remember.

app/src/main/res/values/strings.xml
```diff
@@ -998,9 +998,10 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
     <string name="quest_crossing_kerb_height_title">What’s the height of the curbs at this crossing?</string>
 
     <string name="quest_crossing_title2">Is there a crossing here?</string>
-    <string name="quest_crossing_yes">Yes (e.g. the curb is lowered here or there are markings)</string>
-    <string name="quest_crossing_no">No, but crossing is possible</string>
+    <string name="quest_crossing_yes2">Yes, intended as crossing</string>
+    <string name="quest_crossing_no2">Not intended as crossing, still possible to cross (e.g. informal trails)</string>
```
schra, line 1002: Is it correct to change the "name" field because I want to force the translations to also change their wording?

Is it correct to remove the old translation strings?

app/src/main/java/de/westnordost/streetcomplete/quests/crossing/AddCrossingForm.kt
```diff
@@ -12,11 +12,10 @@ class AddCrossingForm : AListQuestForm<CrossingAnswer>() {
     private val mapDataSource: MapDataWithEditsSource by inject()
 
     override val items = listOf(
-        TextItem(YES, R.string.quest_crossing_yes),
-        TextItem(NO, R.string.quest_crossing_no),
+        TextItem(YES, R.string.quest_crossing_yes2),
+        TextItem(INFORMAL, R.string.quest_crossing_no2),
```
schra, line 16: Can I simply bump the string id here or do I need to implement a fallback for translations that do not yet contain the new string? (Something like "if quest_crossing_yes2 does not exist, then use quest_crossing_yes")

app/src/main/res/values/strings.xml
westnordost: Hm well, if the wording changes substantially, then yes. But in this case, IMO, it doesn't change substantially, so the name should not be changed.

When the source translation changes, existing translations of this string are shown as "fuzzy" in POEditor (the crowd translation platform), it is also possible to filter showing only fuzzy strings.

schra: Ah, I see https://github.com/streetcomplete/StreetComplete/pull/5936/commits/e9b9b93489a2e653aec58d8dc01d842f3aa14648

westnordost (approved):

app/src/main/res/values/strings.xml
```diff
@@ -998,9 +998,10 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
     <string name="quest_crossing_kerb_height_title">What’s the height of the curbs at this crossing?</string>
 
     <string name="quest_crossing_title2">Is there a crossing here?</string>
-    <string name="quest_crossing_yes">Yes (e.g. the curb is lowered here or there are markings)</string>
-    <string name="quest_crossing_no">No, but crossing is possible</string>
+    <string name="quest_crossing_yes">Yes, intended as crossing</string>
```
FloEdelmann, line 1001: I think the examples were really helpful. Can we change this to "Yes, intended as crossing (e.g. with lowered curb or markings)"?

westnordost: I gave the reason for removing it in https://github.com/streetcomplete/StreetComplete/pull/5936#discussion_r1785984014 . Do you disagree?

FloEdelmann: Ah, I haven't seen the explanation before. So it should be fine.

PR #5937, opened 2024-10-01
Docs: suggest reducing image size for new quests

mnalis:
In SCEE discussion it has been found out that image sizes can be reduced significantly (e.g. from `21k` down to `6.7k` [without noticeable loss of quality](https://github.com/Helium314/SCEE/pull/644#issuecomment-2341964292)!)

As the app size increases noticeably with time[^1], it has even been [suggested](https://github.com/Helium314/SCEE/pull/644#issuecomment-2341964292) to check whether the size of already existing images can be reduced (without visual quality loss) too?

[^1]:  e.g. `59.1` is 70 MB, while not so very far ago `38.2` just was 47 MB

### DIFF
--- a/CONTRIBUTING_A_NEW_QUEST.md
+++ b/CONTRIBUTING_A_NEW_QUEST.md
@@ -466,6 +466,8 @@ The [rescaling script](https://github.com/matkoniecz/rescaling_for_android) may 
 
 Please make sure that the images do not take too much disk space. On StretComplete images, JPEG quality of 60% usually looks exactly the same as 80% (or 90%) while being noticeably smaller, and sometimes reducing it down to even 20% will not degrade quality visibly! Play with settings to see which is the smallest size which does not degrade image quality visibly.
 
+There are online tools like [squoosh](https://squoosh.app/) which allow for quick visual comparison if you prefer that.
+
 After adding a photo, remember to update [the credits file](app/src/main/res/authors.txt) (different to the one for icons).
 
 ## Resurvey
--- a/CONTRIBUTING_A_NEW_QUEST.md
+++ b/CONTRIBUTING_A_NEW_QUEST.md
@@ -336,7 +336,7 @@ Keep similar style to existing ones and the app in general. Note that the backgr
 
 Once the quest icon is ready:
 
-- when using Inkscape, save as "Optimized SVG" to remove unnecessary cruft or use another tool for that, like [svgo](https://github.com/svg/svgo)
+- when using Inkscape, save as "Optimized SVG" to remove unnecessary cruft or use another tool for that, like [svgo](https://github.com/svg/svgo) or online [SVGOMG](https://svgomg.net/)
 - Put SVG into [`res/graphics/quest`](res/graphics/quest) folder
   - SVG is a standard format editable in various software, unlike internal Android Studio XML that will be produced in the next steps.
 - Open Android Studio
--- a/CONTRIBUTING_A_NEW_QUEST.md
+++ b/CONTRIBUTING_A_NEW_QUEST.md
@@ -464,7 +464,7 @@ Each of these folders should hold the same image resized to a different resoluti
 
 The [rescaling script](https://github.com/matkoniecz/rescaling_for_android) may be useful, but you can also do this manually with Gimp or similar software.
 
-Please make sure that the images do not take too much disk space. Most useful way to do that is by lowering JPEG quality, which can make images noticeably smaller. Useful range is usually between 20% and 85%. Play with settings to see which is the smallest size which does not degrade image quality visibly.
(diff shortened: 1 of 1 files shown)


kmpoppe:
My [other project](https://github.com/2factorauth/twofactorauth) also asks for reducing SVG file sizes in the [Guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#Guidelines). 

I would recommend to also name SVG as a potential target for reducing filesize as they tend to contain clutter of some degree. Mentioning SVGOMG as a web tool is also fine. 

K

kmpoppe:
> it has even been [suggested](https://github.com/Helium314/SCEE/pull/644#issuecomment-2341964292) to check whether the size of already existing images can be reduced

Also, while JPG/PNG would do the heavy lifting on reduing app size when files get smaller, running all 631 SVGs in res/graphics through svgo would reduce the size from 1756.344 KiB to 1308.133 (~25%).

mnalis:
> I would recommend to also name SVG as a potential target for reducing filesize as they tend to contain clutter of some degree. 

I agree, but optimizing SVG is _already_ mentioned at the bottom of https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#adding-quest-icon section?  Do you think it needs improvement?

kmpoppe:
The new quest MD is quite a long read, I have glossed over it and didn't see that part - until you just told me.

I'd add SVGOMG as a web tool for completeness.

westnordost:
My experience is that JPEG quality below 80% _DOES_ become noticeable. I usually end up saving with 85% image quality for StreetComplete. 

So rather than mentioning some "60%" target, rather mention tools to compare original image with image reduced in quality (GIMP has that feature, too).

CONTRIBUTING_A_NEW_QUEST.md
```diff
@@ -464,6 +464,10 @@ Each of these folders should hold the same image resized to a different resoluti
 
 The [rescaling script](https://github.com/matkoniecz/rescaling_for_android) may be useful, but you can also do this manually with Gimp or similar software.
 
+Please make sure that the images do not take too much disk space. On StretComplete images, JPEG quality of 60% usually looks exactly the same as 80% (or 90%) while being noticeably smaller, and sometimes reducing it down to even 20% will not degrade quality visibly! Play with settings to see which is the smallest size which does not degrade image quality visibly.
```
matkoniecz, line 467: are you sure about 60%? I was not adding images in some time, but I remember noticing visible quality damage at 80%, to say nothing about 60%.

That is why https://github.com/matkoniecz/rescaling_for_android/blob/master/rescale.rb is using 85%

matkoniecz (changes requested):

CONTRIBUTING_A_NEW_QUEST.md
mnalis: Well, it depends on exact image, so for automated tools its hard to guess :shrug: 

See this [image comparison](https://github.com/Helium314/SCEE/pull/644#issuecomment-2341964292), where `20%` (!!!) image quality (on the right) if pretty comparable to much higher JPEG quality **even when viewed on desktop sizes** - yes, you can say they're not exactly the same if you have them one near the other, but if anything, that `20%` JPEG might even look slightly better.   

When reduced to size of mobile phone screen, however, even those tiny differences become invisible.

That is however probably very hard to do without human in the loop, which is why I've tried to emphasize playing with settings to get best results. That `rescale.rb` might get configurable JPEG quality as a parameter though, to ease the trying.

mnalis: anyway, I've removed specific percentage recommendations now, and emphasized testing.

matkoniecz: Oh, I guess that depends on topic of image. I was testing with gravel and leaves which are likely far more fragile than simple mechanical object on monocolour background.

mnalis:
Ok, reworded to emphasize playing with values and not recommend specific parameters, and also additionally mentioned GIMP and SVGOMG. Does it look better now?

westnordost:
> Useful range is usually between 20% and 85%

Uh, what? You mean 100% to 85%, right?

It is more important that the pictures look good, crisp and clear than that we can squeeze a few kilobytes out of these.

mnalis:
>>    Useful range is usually between 20% and 85%

> Uh, what? You mean 100% to 85%, right?

No, I've actually meant that. :sweat_smile: Have you looked at the example image comparison linked in original post?

Reducing below 20% almost always produces visible artifacts, and increasing over 85% almost always just increases size and not quality (esp. if we're talking mobile devices). Especially 100% is completely ridiculous IMHO...

> It is more important that the pictures look good, crisp and clear than that we can squeeze a few kilobytes out of these.

OK, I've removed any mention of value range  technical detail -- most people will use some slider and their eyes as a gauge anyway.

Issue #4810, opened 2023-02-13
Sidewalk overlay: Allow editing the kind of footway


westnordost:
The sidewalk overlay displays footways and paths, currently only for orientation and to see whether a sidewalk has been mapped separately. The [`footway`](https://wiki.openstreetmap.org/wiki/Key%3Afootway) key allows to define the kind of footway. It should be possible to set the kind of footway in the overlay.

The UI should look similar to the Lit way or separate cycleway UI. Selection options should be at least `sidewalk`, `crossing`, `traffic_island`, nothing (i.e. tag not set or unsupported), `access_aisle` and `link`. 

[`footway`](https://wiki.openstreetmap.org/wiki/Key%3Afootway) is used first and foremost to differentiate (other) foot paths from sidewalks & crossings. It is not possible to create a quest to have users define the kind of footway because there is no catch-all "other footways" `footway=*` value (maybe `footway=path` but it is not documented). Overlays however do not have this constraint, so it would be possible to specify the kind of footway within the sidewalk overlay, and it seems to be a good fit.

The information whether a footway is really a separately mapped sidewalk plus `sidewalk:both=separate` helps software to properly associate each sidewalk with a street. Also, knowing which footways are sidewalks, e.g. rendering software could opt to not show or show sidewalks differently from other footways etc.

Vinay-Khanagavi:
I’ve reviewed this issue and would love to work on it. My proposed solution is to enhance the sidewalk overlay UI to allow users to select the type of footway (sidewalk, crossing, traffic island, etc.), 
similar to the existing UI for lit ways and cycleways. This will help differentiate between different footway types and improve the overall functionality of the overlay.

mnalis:
> The UI should look similar to the Lit way or separate cycleway UI. Selection options should be at least sidewalk, crossing, traffic_island, nothing (i.e. tag not set or unsupported), access_aisle and link.

I'm wondering though does it fulfil _"💵 Valuable Surveyors: Surveyor time is valuable, they shouldn't be asked to complete data that does not require a survey and can more efficiently be done remotely."_ ?

It would seem to me that (at least for vast majority of the cases) such tagging would be easier to do with aerial imagery at regular desktop editor, and rarely requires on-ground survey? Or am I missing something?

[...]

PR #5941, opened 2024-10-02
Clarify OVERLAY_GUIDELINES.md

mnalis:
that not all **Quest** Guidelines apply equally to **Overlays**.

See https://github.com/streetcomplete/StreetComplete/issues/4810#issuecomment-2381565723 and below


westnordost:
Well, no, not really.

Things that are unanswerable can also not be put into an overlay unless "not defined" is not seen as "missing". We are walking the line here in #4810. And the spam rule of course also applies, that's why untagged `cycleway:<side>` is not shown as missing on e.g. motorways.
And in these guidelines I certainly don't want to describe this in detail. What's written is "in general", meaning, apply common sense.

A change that would be fine by me would be to append, ", when applicable".

Issue #5940, opened 2024-10-02
Do not automatically remove amenity details in Places overlay when extra details are added via NSI


mnalis:
**Use case**
Sometimes, some existing amenity/shop has lacking (e.g. no `brand`, `brand:wikidata` etc) details, or shortened (or slightly misspelled) name (e.g. `Raiffaisen bank` or `RBA` instead of correct `Raiffeisen bank`). 

Nice way to quickly remedy that is Places Overlay, where one just searches for correct brand in NSI, and all the ideal data (correct most popular name, `brand`, wikidata etc.) is filled in automatically!

That works **great** when existing amenity does not have much extra details, e.g. it only has `amenity=bank` (and maybe `name=*`).

However, if amenity had extra details (e.g. `amenity=bank` + `name=RBA` + `opening_hours` + `wheelchair` + `drive_through` + `bic`) when NSI for _"Raiffeisen bank"_ is selected, all those extra details (`opening_hours`, `wheelchair`, ...) are ***removed without question***, which makes using NSI for such improvements unusable (as it breaks more stuff than it fixes)

It seems to be because [shouldAlwaysReplaceShop](https://github.com/streetcomplete/StreetComplete/blob/85f2c9febdd5619d363c8065fb322a6bf995a080/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt#L277-L284) gets triggered.

(Inspired by https://github.com/Helium314/SCEE/discussions/670#discussioncomment-10818963)

**Proposed Solution**

I would suggest that in such cases when **selected NSI does not change the type** of the feature, user should be presented with [confirmReplaceShop()](https://github.com/Helium314/SCEE/blob/a5043084be37f44bbe5bd5529b4d2e52157c1f5a/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt#L297) dialog instead (that should happen automatically if `shouldAlwaysReplaceShop` is `false`), thus allowing them to choose to preserve all those pre-existing details like `opening_hours` etc.

Furher refinement to slightly reduce effort for mapper (but not strictly necessary) would be to set `shouldNotReplaceShop` instead if **both the feature type _and_ the name remain the same**, which would automatically preserve pre-existing data without even asking the user.

westnordost:
The issue that lead to the current behavior of replacing the feature even when the name was not changed: #5195

Anyway, your suggestion sounds reasonable.

So the check would go

[...]

PR #5944, opened 2024-10-02
(Ask to) preserve amenity details when chosen NSI preset has the same type

mnalis:
Fixes https://github.com/streetcomplete/StreetComplete/issues/5940

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
@@ -274,7 +274,7 @@ private suspend fun createEditAction(
 
     if (newFeature.isSuggestion) {
         // selecting NSI preset will always return empty newNames, even if NSI does set new name=* tag
-        hasChangedNames = listOf(LocalizedName("", newFeature.addTags.get("name").orEmpty())) != previousNames
+        hasChangedNames = parseLocalizedNames(newFeature.addTags) != previousNames
     }
 
     val shouldNotReplaceShop =
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
@@ -261,6 +261,7 @@ private suspend fun createEditAction(
     val hasAddedNames = previousNames.isEmpty() && newNames.isNotEmpty()
     val hasChangedNames = previousNames != newNames
     val hasChangedFeature = newFeature != previousFeature
+    val hasChangedFeatureType = previousFeature?.id?.substringBeforeLast("/") != newFeature.id.substringBeforeLast("/")
     val isFeatureWithName = newFeature.addTags.get("name") != null
     val wasFeatureWithName = previousFeature?.addTags?.get("name") != null
     val wasVacant = element != null && element.isDisusedPlace()
@@ -276,7 +277,7 @@ private suspend fun createEditAction(
         || element == null
     val shouldAlwaysReplaceShop =
         // the feature is or was a brand feature (i.e. overwrites the name)
-        isFeatureWithName || wasFeatureWithName
+        (isFeatureWithName || wasFeatureWithName) && hasChangedFeatureType
         // was vacant before but not anymore (-> cleans up any previous tags that may be
         // associated with the old place)
         || wasVacant && hasChangedFeature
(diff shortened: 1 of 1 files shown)


mnalis:
Should work now. If people want to try it, [here is debug .apk](https://github.com/mnalis/StreetComplete/actions/runs/11154983310)

app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
```diff
@@ -259,24 +259,34 @@ private suspend fun createEditAction(
     val tagChanges = StringMapChangesBuilder(element?.tags ?: emptyMap())
 
     val hasAddedNames = previousNames.isEmpty() && newNames.isNotEmpty()
-    val hasChangedNames = previousNames != newNames
+    var hasChangedNames = previousNames != newNames
     val hasChangedFeature = newFeature != previousFeature
+    val newFeatureType = if (newFeature.isSuggestion) newFeature.id.substringBeforeLast("/") else newFeature.id
+    val previousFeatureType = if (previousFeature?.isSuggestion == true) previousFeature?.id?.substringBeforeLast("/") else previousFeature?.id
+    val hasChangedFeatureType = previousFeatureType != newFeatureType
```
westnordost, line 266: could you put this into an own function? Like

```
/** return the id of the feature, without any brand stuff */
private val Feature.featureId get() = if (isSuggestion) id.substringBeforeLast("/") else id
```

app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
```diff
@@ -259,24 +259,34 @@ private suspend fun createEditAction(
...
     val isFeatureWithName = newFeature.addTags.get("name") != null
     val wasFeatureWithName = previousFeature?.addTags?.get("name") != null
     val wasVacant = element != null && element.isDisusedPlace()
     val isVacant = newFeature.id == "shop/vacant"
 
+    if (newFeature.isSuggestion) {
+        // selecting NSI preset will always return empty newNames, even if NSI does set new name=* tag
+        hasChangedNames = listOf(LocalizedName("", newFeature.addTags.get("name").orEmpty())) != previousNames
+    }
```
westnordost, line 275: You can call `parseLocalizedNames(newFeature.addTags)` to get a list of `LocalizedName`.

Also, what about `previousFeature`?

westnordost (changes requested):

app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
```diff
@@ -259,24 +262,34 @@ private suspend fun createEditAction(
     val tagChanges = StringMapChangesBuilder(element?.tags ?: emptyMap())
 
     val hasAddedNames = previousNames.isEmpty() && newNames.isNotEmpty()
-    val hasChangedNames = previousNames != newNames
+    var hasChangedNames = previousNames != newNames
     val hasChangedFeature = newFeature != previousFeature
+    val newFeatureType = newFeature.featureId
+    val previousFeatureType = previousFeature?.featureId
+    val hasChangedFeatureType = previousFeatureType != newFeatureType
```
westnordost, line 269:
```suggestion
    val hasChangedFeatureType = previousFeature?.featureId!= newFeature.featureId
```

app/src/main/java/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
mnalis: > Also, what about previousFeature?

This "hack" for `parseLocalizedNames(newFeature.addTags)` is needed, because `newNames` are empty if NSI is chosen (as opposed as to when regular type+name are entered in separate fields, when `newNames` is set), so we have to fetch the new name (which would be set by preset) **ourselves** if it is NSI.

On the other hand, such `parseLocalizedNames(previousFeature.addTags)` hack is not needed because `previousNames` seems to be supplied _regardless_ of whether previous value was type+name or NSI.

(I've put some `Log.w` debugs in that [debug apk](https://github.com/mnalis/StreetComplete/actions/runs/11154983310) while I was testing, and that is what they've shown)

If that is what you meant?

westnordost: That's what I meant. Alright

westnordost (approved):

PR #5946, opened 2024-10-03
Fix newline in strings.xml

schra:


schra:
Ah yeah I need to do that in the translation app. sorry

Issue #5716, opened 2024-07-02
Cycleway overlay marks both sides of roundabout as missing


FloEdelmann:
**How to Reproduce**
For https://osm.org/way/492794174 (OSM element version 11), the cycleway quest correctly recognizes that it's a oneway street ([implied](https://wiki.openstreetmap.org/wiki/Tag:junction%3Droundabout#How_to_map) by [`junction=roundabout`](https://wiki.openstreetmap.org/wiki/Tag:junction%253Droundabout) but not explicitly set with [`oneway=yes`](https://wiki.openstreetmap.org/wiki/Tag:oneway%3Dyes)). The cycleway overlay however does not recognize this and marks both sides as red:

<img alt="Screenshot_20240702-141300_Street­Complete" src="https://github.com/streetcomplete/StreetComplete/assets/202916/49a1ea6a-9a94-46dd-9c04-7cdb0a54236d" width="300"> <img alt="Screenshot_20240702-141318_Street­Complete" src="https://github.com/streetcomplete/StreetComplete/assets/202916/2a1e23a8-052d-4066-a88a-6c0e9a90bfc5" width="300">

**Expected Behavior**
The cycleway overlay should also only mark the outer side as red, and possibly also only allow setting the outer side.

**Versions affected**
StreetComplete v58.1, Android 13

FloEdelmann:
Tangentially relevant: There is still some special behavior left in the cycleway parser for `opposite` tags: https://github.com/streetcomplete/StreetComplete/blob/05ef7ce4564e8eba42220ae02e38f87f0e7b6aa5/app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt#L18
Maybe that can be removed/simplified now?

westnordost:
> Tangentially relevant: There is still some special behavior left in the cycleway parser for `opposite` tags: https://github.com/streetcomplete/StreetComplete/blob/05ef7ce4564e8eba42220ae02e38f87f0e7b6aa5/app/src/main/java/de/westnordost/streetcomplete/osm/cycleway/CyclewayParser.kt#L18
Maybe that can be removed/simplified now?

(Unfortunately) no, this is still necessary to parse correctly for which side a `cycleway=*` (without sides given) in oneways is meant for. `opposite` is deprecated, and it should be marked a such, but still it must be handled correctly when it is tagged.

westnordost:
Short answer: Will not fix.

Long answer:
[...]

PR #5961, opened 2024-10-14
Cycleway overlay: don't show contra-flow side of roundabouts etc. as missing when there is no data

### LINKED ISSUE #5959: Do not ask if there is an _opposite_ cycling lane for roundaboits
StreetComplete asks for cycling lanes in the two directions for one-way streets (which is good as some streets accepts cyclists in the two ways). However, I'm not sure if there is any roundabout that can be taken in the opposite way with a bike. 

**How to Reproduce**
Screenshot_20241014-080213_StreetComplete

**Expected Behavior**

**Versions affected**
- 59.2

westnordost:
fixes #5959 / #5716


FloEdelmann (approved):
From only reading the code, this seems to make sense :+1:

FloEdelmann:
@westnordost Slightly off-topic here, but related:

After you moved the `Direction` class to the `osm` package in https://github.com/streetcomplete/StreetComplete/commit/9be66204540017a861d13ff42cadb9fa52879ddc and https://github.com/streetcomplete/StreetComplete/commit/bd7905fcfc193f58baa17ccf7bdb581e2d4b7317 (probably right before this PR), the tests on the `master` branch are now failing because of unresolved imports: https://github.com/streetcomplete/StreetComplete/actions/runs/11327607582/job/31499031620

westnordost:
Yes, thanks, I corrected it.

PR #5962, opened 2024-10-14
be more aggressive in removing keys associated with surface when surface is added or changed

### LINKED ISSUE #5951: Solving some track related quests can leave inconsistent data behind
Hi there,

some weeks ago I came across a situation where solving some simple quests turn consistent tagging into an inconsistent tag combination.

Consider the following situation: An unpaved agricultural track, tagged consistently with

```
highway=track
surface=compacted
tracktype=grade2
```

The smoothness quest popped up. However, this track has obviously been asphalted recently, so I answered with _"Other surface"_ from the _Äh..._ menu.

Now the surface quest popped up, which I answered with _"Asphalt"_.

Now the smoothness popped up again, which I answered with _"Excellent"_.

No further quests popped up, leaving the tagging as

```
highway=track
surface=asphalt
smoothness=excellent
tracktype=grade2
```

**Which is a contradiction** - I cannot imagine any case where this combination could be valid - tracks with "excellent" asphalt will **_always_** be "grade1" ([see the wiki](https://wiki.openstreetmap.org/wiki/Key:smoothness#Smoothness)).

Is there anything that could be improved here?

**Ideas:**

```
IF (highway="track" // or "path" or similar
    AND surface was changed from anything to "asphalt" // or "concrete" or "paving stones" or similar
    AND tracktype != "grade1") {
        delete tracktype
}
```
(SC users with activated "tracktype" quest would be able to correct this immediately, otherwise at least the now wrong "tracktype" tag would be deleted.)

Or maybe even better:

```
IF (highway="track"
    AND surface="asphalt"
    AND smoothness was changed from anything to "excellent"
    AND tracktype exists) {
        set tracktype to "grade1"
}
```

(which changes the tracktype to the only correct value but does not silently create this tag; if the latter is no problem then the last AND clause can be of course omitted.)

Observed with SC v58.

The cycling cat

westnordost:
fixes #5951

Whenever `surface` is changed or added, keys potentially associated with `surface` are removed, e.g. `smoothness` or `tracktype`.

- Before, `tracktype` was only removed when it conflicted with the newly set `surface` value.
- Before, keys were only removed if the `surface` was **changed**, now they are also removed if the `surface` was **added** (e.g. road with `tracktype=grade3` is removed when `surface=dirt` is added)

Reasons for the change in behavior was stated in #5951, in a nutshell:

When answering the smoothness quest, one can answer that the displayed surface is incorrect. This removes the `surface` tag.
The surface quest will pop up and ask about the surface. Now, when specifying any surface, a previously set `tracktype` will not be removed because 1. `surface` is now "new" and 2. `tracktype` is only removed if it conflicts with the new `surface`, regardless of the `smoothness` value.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/surface/SurfaceCreator.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/surface/SurfaceCreator.kt
@@ -1,7 +1,6 @@
 package de.westnordost.streetcomplete.osm.surface
 
 import de.westnordost.streetcomplete.osm.Tags
-import de.westnordost.streetcomplete.osm.removeCheckDatesForKey
 import de.westnordost.streetcomplete.osm.updateWithCheckDate
 
 /** Apply the surface and note to the given [tags], with optional [prefix], e.g. "footway" for
@@ -16,15 +15,9 @@ fun SurfaceAndNote.applyTo(tags: Tags, prefix: String? = null, updateCheckDate: 
     val key = "${pre}surface"
     val previousOsmValue = tags[key]
 
-    val shouldRemoveTracktype = prefix == null && isSurfaceAndTracktypeConflicting(osmValue, tags["tracktype"])
-    if (shouldRemoveTracktype) {
-        tags.remove("tracktype")
-        tags.removeCheckDatesForKey("tracktype")
-    }
-
-    // remove smoothness (etc) tags if surface was changed
-    // or surface can be treated as outdated
-    if ((previousOsmValue != null && previousOsmValue != osmValue) || shouldRemoveTracktype) {
+    // remove smoothness, tracktype (etc), i.e. tags that are (potentially) associated with a given
+    // surface type, as they can potentially be incorrect now that the surface changed (see #5951)
+    if (previousOsmValue != osmValue) {
         getKeysAssociatedWithSurface(pre).forEach { tags.remove(it) }
     }
 
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/surface/SurfaceUtils.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/surface/SurfaceUtils.kt
@@ -109,6 +109,8 @@ fun getKeysAssociatedWithSurface(prefix: String = ""): Set<String> =
         "${prefix}smoothness",
         "${prefix}smoothness:date",
         "source:${prefix}smoothness",
+        "${prefix}tracktype", // it's at least *potentially* associated, see #5951
+        "source:${prefix}tracktype",
         "${prefix}paving_stones:shape",
         "${prefix}paving_stones:pattern",
         "${prefix}paving_stones:length",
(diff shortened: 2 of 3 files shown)


westnordost:
> road with tracktype=grade3 is removed when surface=dirt is added

I am getting second guesses about this solution now though...

westnordost:
So, this is not the way.

westnordost:
On the other hand, there is no 100% solution other than letting the user confirm always if the tracktype is also still the same whenever `surface` is specified or changed, which is out of the question.

Considering that wrong data is miles worse than no data, I think indeed it is better to delete **possibly** outdated/wrong data than only deleting it if it is definitely conflicting with just surveyed data. As @mnalis [stated earlier](https://github.com/streetcomplete/StreetComplete/issues/5951#issuecomment-2399219022):

> After all, when surface changes ... [the tracktype and smoothness] could've remained the same, but that is by accident, not by design, so should be re-verified anyway even if happened to remain the same on the ground

matkoniecz:
I think that in general it is thorny problem where any solution will leave someone unhappy.

I am relatively happy about my solution, but deleting all surface-related info also seems fine, even if I prefer my version.

For completeness: showing mapper what is tagged now and asking them to confirm whether this info applies is possible but would be overcomplicated, confusing and time-consuming to implement.

westnordost:
I know, that dialog is a bit orthogonal to the topic of what is tagged / what tags are removed, but I came to review the whole implementation because it just seems to be a bottomless pit.

westnordost:
How about this.

Replace all the "is conflicting", "is suspicious" `surface` <-> `tracktype` logic with the following simple logic:

1. Delete `tracktype` whenever `surface` **changes** from any _paved_ surface to a _non-paved_ surface or the other way round.
2. (to solve #5951: ) Delete `tracktype` alongside when user answers in the smoothness quest that the surface is different. (Or more generally, delete `tracktype` whenever `surface` is deleted for whatever reason)

matkoniecz:
> Delete tracktype whenever surface changes from any paved surface to a non-paved surface or the other way round.

not saying that it is a blocker, as no solution will be ideal: but note that `tracktype=grade5 surface=mud` changing to `surface=compacted` will happily keep `tracktype=grade5`

westnordost:
As will the current implementation. Anyway, there could be three categories which each trigger a deletion of `tracktype`: paved surfaces, natural surfaces, man-made unpaved surfaces

mnalis:
> Anyway, there could be three categories which each trigger a deletion of `tracktype`: **paved** surfaces, **natural** surfaces, **man-made unpaved** surfaces

Yeah, that might be good compromise. After all, most changes are likely to either be:

- intentional **man-made improvements**  (e.g. mud gets bunch of gravel on top of it, or compacted road get asphalted etc.), _or_
- **degradations** - usually **man-induced temporary degradation** like asphalt road becomes gravel until it is resurfaced (although that one is often better tagged with `highway=construction` tag if road is closed while works are ongoing), but might be due to longer-term neglect / natural decay too.

and deleting `tracktype` when `surface` jumps from one of those 3 categories to other would seem to cover vast majority of those cases I can think of.

> I know, that dialog is a bit orthogonal to the topic of what is tagged / what tags are removed, but I came to review the whole implementation because that topic just seems to be a never-ending story.

Well, yeah, but addressing that issue here would require more checks - e.g. what if it is not the **surface quest** but **tracktype quest** instead that **produces the incompatibility**? Should we remove the `surface` then? Or refuse the `tracktype` change altogether? Or pop up the same warning dialog as before?

PR #5969, opened 2024-10-16
Regard country-specific presets

### LINKED ISSUE #4916: No label / feature for post boxes

westnordost:
Would fix #4916

but I have serious doubts about this. It makes everything a lot more complicated, as indicated by the sheer amount of code I had to touch.

Also, what's missing is to adapt the tests, so a final PR would add a few hundred lines of code more.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Things.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Things.kt
@@ -70,7 +70,7 @@ private val IS_THING_EXPRESSION by lazy {
             "whirlpool",
 
             // public service
-            // "post_box", - blocked by https://github.com/streetcomplete/StreetComplete/issues/4916 - which is blocked by https://github.com/westnordost/osmfeatures/issues/23
+            "post_box",
 
             // facilities & others
             "baking_oven",
@@ -86,7 +86,7 @@ private val IS_THING_EXPRESSION by lazy {
             "give_box",
             "karaoke_box",
             "kitchen", // usually an amenity within campsites etc, i.e. like shower, toilets, ...
-            // "letter_box", - see "post_box", but also, it would be very spammy to comprehensively map this
+            // "letter_box", - it would be very spammy to comprehensively map this
             "library_dropoff",
             "locker",
             "lounger",
@@ -205,7 +205,7 @@ private val IS_THING_EXPRESSION by lazy {
             "planter",
             "snow_cannon",
             "stele",
-            // "street_cabinet", - blocked; see note at amenity=post_box, most are included by dedicated filter below
+            "street_cabinet",
             "surveillance",
             // "survey_point" - this can be very very small -> verifiability issue
             //                  danger that mapper deletes it because he can't find it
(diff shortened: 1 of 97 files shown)


app/src/main/java/de/westnordost/streetcomplete/osm/Things.kt
```diff
@@ -241,7 +241,6 @@ private val IS_THING_EXPRESSION by lazy {
         or attraction
         or boundary = marker
         or leisure = pitch and sport ~ chess|table_soccer|table_tennis|teqball
-        or man_made = street_cabinet and street_cabinet != postal_service
```
FloEdelmann, line 244: Nooo :persevere: 
Why remove street cabinets from the Things overlay?

matkoniecz:
What would be alternative? Close #4916 as WONTFIX?

Start using modified preset that patches out some country-specific restrictions and generates mailbox presets that is easier to use?

app/src/test/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetTest.kt
```diff
@@ -18,7 +18,7 @@ class AddAddressStreetTest {
         val addr = node(tags = mapOf("addr:housenumber" to "123"))
         val mapData = TestMapDataWithGeometry(listOf(addr))
         assertEquals(1, questType.getApplicableElements(mapData).toList().size)
-        assertNull(questType.isApplicableTo(addr))
+        assertNull(questType.isApplicableTo(addr,))
```
FloEdelmann, line 21: What happened here? :open_mouth: 
I think most formatting changes in the test files could be completely reversed.

app/src/main/java/de/westnordost/streetcomplete/osm/Things.kt
westnordost: because all street cabinets are added.

app/src/test/java/de/westnordost/streetcomplete/quests/address/AddAddressStreetTest.kt
westnordost: > Also, what's missing is to adapt the tests, so a final PR would add a few hundred lines of code more.

westnordost:
> What would be alternative? Close https://github.com/streetcomplete/StreetComplete/issues/4916 as WONTFIX?

> Start using modified preset that patches out some country-specific restrictions and generates mailbox presets that is easier to use?

The second. -> #5970

app/src/main/java/de/westnordost/streetcomplete/osm/Things.kt
FloEdelmann: Ah, I see now. Sorry for the confusion.

PR #5970, opened 2024-10-16
Patch out country-specific presets

### LINKED ISSUE #4916: No label / feature for post boxes

westnordost:
* Fixes #4916
* Replaces and closes #5969


westnordost:
Note for reviewers:

See https://raw.githubusercontent.com/openstreetmap/id-tagging-schema/main/dist/presets.json and search for "`locationSet`" to see what is removed/patched.

app/src/androidTest/java/de/westnordost/streetcomplete/util/NameAndLocationLabelTest.kt
```diff
@@ -45,8 +45,6 @@ class NameAndLocationLabelTest {
         )))
     }
 
-    // https://github.com/streetcomplete/StreetComplete/issues/2640
-    @Ignore("https://github.com/streetcomplete/StreetComplete/issues/4916")
```
FloEdelmann, line 49: I think the first comment should stay, to give some context what exactly is tested here (similar to the other test cases). But the `@Ignore(…)` can of course be deleted.

FloEdelmann (approved):
This looks much cleaner than #5969 :+1::slightly_smiling_face:

westnordost:
So, the following is patched **out**:
- maxspeed sign had a different icon in US, CA, LR
- `office=bail_bond_agent` was only available in US, PH
- `historic=memorial` + `memorial=stolperstein` was only available in Europe
- `amenity=post_box` had the field `drive_through` more prominently displayed in the US
- `amenity=townhall` + `townhall:type=barangay` was only available in PH
- `highway=toll_gantry` was not available in DE
- `network:type=node_network` vertices were only available in BE, DE, LU, NL
- `man_made=footwear_decontamination` was only available in AU, NZ (*come on! As if it wasn't possible to have footwear decontamination facilities elsewhere! This is one good example why this whole thing reeks of shoehornification!*)
- trunk, motorway and primary road links showed advisory maxspeed more prominently than maxspeed, for some reason, in US, CA
- "moped link roads" were only available in NL

Many of these are not relevant for StreetComplete, as preset fields are not used.

PR #5972, opened 2024-10-16
add acrylic surface

### LINKED ISSUE #5965: Add surface=acrylic option for sports pitches
### General
Affected tag(s) to be modified/added: [leisure=pitch](https://wiki.openstreetmap.org/wiki/Key:leisure), [surface=acrylic](https://wiki.openstreetmap.org/wiki/Tag:surface%3Dacrylic)
Question asked: The question is already asked as **What's the surface here?**

### Checklist

### Ideas for implementation

**Element selection:** Already implemented

**Metadata needed:** Not sure, common in the U.S.

**Proposed UI:** Add surface=acrylic to the list of options available when selecting the surface of a `leisure=pitch` element

**Notes**: Adding this as a "Quest suggestion", because it didn't seem like a feature or bug.

I'm a little vague on how to tag certain sports pitches, notably tennis courts and some basketball courts. I've inspected them while using StreetComplete, and I've left the quests unsolved for now. From what I can tell, the surfaces are acrylic, with concrete or asphalt underneath. It's hard to tell for sure without being an expert, but that's what visual inspection and the internet seem to think.

If there's another option that should be used instead like `surface=paved`, it would be good if StreetComplete could guide users towards that option. You can enter that, but it shows up as "Paved (generic)", which seems like it's a less-specific option, with a picture that is just bare concrete/asphalt. If that's the preferred option, maybe StreetComplete could change the input for `leisure=pitch` to show a different picture when answering `surface=*`?

Here's an example of a tennis court currently tagged as `surface=paved`, but should be tagged as `surface=acrylic` AFAICT:

https://www.openstreetmap.org/way/39153147

Here's an example of a tennis court tagged with `surface=acrylic`:

https://www.openstreetmap.org/way/362822900

[leisure=pitch taginfo](https://taginfo.openstreetmap.org/tags/leisure=pitch#combinations) doesn't show any results for combinations with `surface=acrylic` for some reason, even though it's more common than some of the tags shown

westnordost:
Fixes #5965


westnordost:
I also changed it so that the "rubber granules" option tags `surface=rubber` instead of `surface=tartan`. Do you think this is an improvement?

matkoniecz:
I am trying to figure out is this soft merged granule surface a rubber or not. Failed so far.

westnordost:
Well, it's certainly not *natural* rubber, but neither is the rubbery surface on playgrounds etc (`surface=rubber`) nor are "rubber" tyres.

https://en.wikipedia.org/wiki/Tartan_track by 3M specifically is polyurethane according to that article, but that doesn't mean that tracks made from rubbery granules from other manufacturers couldn't be made from other materials.

What they have in common is that they are, well bound rubber(y) granules.

The wiki states that `surface=rubber` may also be loose rubber granules, solid rubber (and bound rubber granules paving plates), but not sure if any of these is actually a thing. In any case, that tag is defined more widely.

knkski:
Thanks for handling this! This looks like it will help the clarity of the quest by a lot.

PR #5976, opened 2024-10-20
prohibited for pedestrians: clear sidewalk tags when user answers that there is a sidewalk

### LINKED ISSUE #5952: "Actually, there is a sidewalk, but it is displayed separately on the map." - applied answer may be wrong, sidewalk may be only on one side
`quest_accessible_for_pedestrians_separate_sidewalk` has

> Actually, there is a sidewalk, but it is displayed separately on the map.

wording.

If selected https://github.com/streetcomplete/StreetComplete/blob/9f6475f25b70e264e35f803c8cb19dba0b81fb1d/app/src/main/java/de/westnordost/streetcomplete/quests/foot/AddProhibitedForPedestrians.kt#L63 will apply `sidewalk:both=separate`

But sidewalk may be actually only on one side. Maybe adding `sidewalk=separate` would be better? Or just remove existing `sidewalk*` tags?

westnordost:
…t there is a sidewalk (fixes #5952)


matkoniecz (approved):
I have not run code, but from looking at it (including tests) it looks great and fixes problem I spotted

Issue #5991, opened 2024-10-30
Cycleway overlay: cycle track with walking allowed


westnordost:
### Discussed in https://github.com/streetcomplete/StreetComplete/discussions/5990

<div type='discussions-op-text'>

<sup>Originally posted by **olecky** October 30, 2024</sup>
When editing in the cycleways overlay, there doesn't seem to be an option to select cycle track (signed with [D4](https://commons.wikimedia.org/wiki/File:Vienna_Convention_road_sign_D4a.svg) or similar sign), respecting the fact that pedestrians are allowed to use it because of national or international legislation. In most cases, if there is no sidewalk, pedestrians are allowed to walk on the cycle track (article 20 paragraph 3 of the Convention on Road Traffic). Example:

![IMG_20200927_145527-resized](https://github.com/user-attachments/assets/2ae7e379-837b-4f15-b57c-4711ae1ccb50)

In the app, the closest I can choose are:

- EXCLUSIVE - which adds foot=no, banning pedestrians
- NON_SEGREGATED - which adds foot=designated, implying that there is sign designating the highway for pedestrians (so a combination of D4 and D5 rather than only D4)

Both of the option seem incorrect to me in the example shown on the picture. I'm not sure if it's necessary to add one more option to distinguish cycle tracks with walking allowed and not. From my point of view, removing the foot=no addition in EXCLUSIVE would solve the problem. Is it really useful for pedestrian routing in situations where there is parallel cycle track and footpath?

</div>

westnordost:
So, I think @olecky's suggestion makes sense - remove `foot` when retagging to `highway=cycleway`. 

@olecky: Care to implement this yourself and submit a PR, as you already looked at the source code?

olecky:
I did the edit I suggested, but I didn't test it, so feel free to reject the pull request.

westnordost:
Yeah, it doesn't work, see my comment in #5992.

Maybe it needs to be discussed whether the new behavior should be 
a) always remove `foot`
b) remove `foot` if it is not `no`

olecky:
I'm a bit afraid of removing foot, because occasionally there are also cases with:

a) signs explicitly forbidding pedestrians from the track (for example: https://www.openstreetmap.org/way/826279591) 
[...]

PR #5992, opened 2024-10-30
do not set foot=no when selecting a designated cycle track

olecky:
see: https://github.com/streetcomplete/StreetComplete/issues/5991


westnordost:
Cool! However, don't we also want to remove `foot` if `EXCLUSIVE` is selected? Or at least, it should be removed if it is not `foot=no`. (Not sure which would be the better option, maybe should be discussed)

Also, you probably need to adapt some tests in `SeparateCyclewayCreatorKtTest.kt`

westnordost:
Closing, as it needs more discussion in the linked ticket.

PR #6008, opened 2024-11-15
add bicycle repair station equipment quest (#5910)

westnordost:


app/src/main/res/values/strings.xml
```diff
@@ -785,6 +785,12 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_bicycle_rental_type_human">Staffed bicycle rental</string>
     <string name="quest_bicycle_rental_type_shop_with_rental">Bicycle store that also rents bicycles</string>
 
+    <string name="quest_bicycle_repair_station_services_title">What equipment is available here?</string>
+    <string name="quest_bicycle_repair_station_pump">Pump</string>
```
westnordost, line 789: Or "Functioning pump"?

Wiki is silent about this.

app/src/main/res/drawable/ic_quest_bicycle_repair_amenity.xml
FloEdelmann, line 1: Can you please add the SVG source for this drawable?

FloEdelmann (approved):
Apart from one small comment this looks fine to me :slightly_smiling_face:

app/src/main/res/drawable/ic_quest_bicycle_repair_amenity.xml
westnordost: it's just `ic_quest_bicycle_repair` with green background

FloEdelmann: Still, it'll break the [GenerateQuestListTask](https://github.com/streetcomplete/StreetComplete/blob/5250ec2c56859713a0383de2bc309802cc10443b/buildSrc/src/main/java/GenerateQuestListTask.kt#L136-L147) if it's not there :grimacing:

app/src/main/java/de/westnordost/streetcomplete/quests/bicycle_repair_station/BicycleRepairStationServiceItem.kt
```diff
@@ -0,0 +1,22 @@
...
+
+val BicycleRepairStationService.titleResId: Int get() = when (this) {
+    PUMP -> R.string.quest_bicycle_repair_station_pump
+    TOOLS -> R.string.quest_bicycle_repair_station_tools
+    STAND -> R.string.quest_bicycle_repair_station_stand
+    CHAIN_TOOL -> R.string.quest_bicycle_repair_station_chain_tool
+}
+
+val BicycleRepairStationService.iconResId: Int get() = when (this) {
+    PUMP -> R.drawable.bicycle_repair_station_pump
```
mnalis, line 18: Those pictures looked tiny to me, but upon closed inspection that was just because they are too crowded by that text over them.  

Could we have text below the icon instead of over it, like in recycling quests? That would make it much more usable (as well as more consistent) IMHO:

small_Screenshot_20241115_220206_StreetComplete Dev small_Screenshot_20241115_215715_StreetComplete Dev

Or, if the text is going to be drawn over pictures, then make them bigger - i.e. put it in 2x2 instead 1x4 grid  (i.e. closer in looks to how sidewalk overlay icons look in 2+1, or like `surface` quest which is Nx3)?

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
```diff
@@ -16,13 +16,12 @@ import de.westnordost.streetcomplete.util.ktx.toYesNo
 class AddAirCompressor : OsmFilterQuestType<Boolean>() {
 
     override val elementFilter = """
-       nodes, ways with
-       amenity = fuel
-       and (
-           !compressed_air
-           or compressed_air older today -6 years
-       )
-       and access !~ private|no
```
mnalis, line 25: Is there a reason why `and access !~ private|no` was deleted here? Admittedly there are not that many of them but they do [seem to exist](https://overpass-turbo.eu/s/1UdT), and leaving the check in should not be harmful?

mnalis (changes requested):
I've left a few suggestions. Other than those, it seems to work fine (debug `.apk` is [here](https://github.com/mnalis/StreetComplete/actions/runs/11862480805) if someone wants to quickly try it out and give feedback)

app/src/main/res/values/strings.xml
mnalis, line 789:
```suggestion
    <string name="quest_bicycle_repair_station_pump">Pump (operational)</string>
```

or "working pump", "functional pump" or similar wording. Rationale why that emphasis is needed in previous discussion: https://github.com/streetcomplete/StreetComplete/issues/3053#issuecomment-879326785

app/src/main/java/de/westnordost/streetcomplete/quests/bicycle_repair_station/BicycleRepairStationServiceItem.kt
matkoniecz: > as well as more consistent

AFAIK all photo selections work in this way

app/src/main/res/drawable/ic_quest_bicycle_repair_amenity.xml
```diff
@@ -0,0 +1,56 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="128dp"
+    android:height="128dp"
+    android:viewportWidth="128"
+    android:viewportHeight="128">
+  <path
+      android:pathData="m128,64c0,35.35 -28.65,64 -64,64s-64,-28.65 -64,-64 28.65,-64 64,-64 64,28.65 64,64"
+      android:fillColor="#9bbe55"/>
```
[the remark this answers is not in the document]
westnordost: You are right

app/src/main/java/de/westnordost/streetcomplete/quests/air_pump/AddAirCompressor.kt
westnordost: Because this is not about `amenity=air_compressor` but `amenity=fuel`.

Specifying a fuel station as private is a bit of a troll tag.

app/src/main/res/values/strings.xml
westnordost: I was thinking about this, too, but I decided against this because:
1. wiki gives no indication whether the tag is about a physical presence of a pump or that it is working
2. other than for obvious cases, e.g. where the tube has been torn off, it might not be easily possible for the mapper to notice when a pump doesn't work. If the pump is damaged enough that pieces are missing, I'd expect that mappers either answer "no" or leave a note also without any change in wording.

app/src/main/java/de/westnordost/streetcomplete/quests/bicycle_repair_station/BicycleRepairStationServiceItem.kt
matkoniecz: > quest definitely does not look like that; the text is below the pictures there, not over them.

that is not **photo** selection

westnordost:
Alright, thanks for the review, @mnalis and @FloEdelmann !

app/src/main/java/de/westnordost/streetcomplete/quests/bicycle_repair_station/BicycleRepairStationServiceItem.kt
mnalis: > that is not **photo** selection, these drawings are not photos

oh, ok, thanks for clarification, I did not notice that subtle distinction. My mind sees to conflate image/photo as any graphical representation  :sweat_smile: 

Anyway, I see it's changed to `itemsPerRow = 3` now, so hopefully that helps enough.

PR #6020, opened 2024-11-22
Cycleway overlay: Differentiate bicycle access on pedestrian roads

### LINKED ISSUE #6016: visibility of bicycle=yes at pedestrian zones
**Use case**
There is a pedestrian zone with allowed bicycle access (bicycle=yes, bicycle:signed=yes). In cycling mode of StreetComplete this way is not shown.
(https://www.openstreetmap.org/way/59038921)

**Proposed Solution**
Show pedestrian zone (highway=pedestrian, but also if tagged as area) in the same color like footways with bicycle=yes bicycle:signed=yes.

westnordost:
fixes #6016

**Changes**:
- pedestrian roads are now colored like footways: black by default, aquamarine when bicycles are allowed, cyan when also designated to bicycles
- added other-answer options (only displayed for pedestrian streets) to switch between these.
- Cannot select bicycle road for pedestrian streets
- current bicycles-on-pedestrian-road situation is displayed in form UI
- also for bicycle roads: the displayed sign is not modal anymore, i.e. can specify left-and-right-situation, too


westnordost:
Cool, thanks for testing it!

I don't understand your comment about "pavement mode". Could you add a screenshot or something?

> three dot menu entries: [...]

Oh! Weird, I thought I tested that. Will fix before merge.

> when adding protected bicycle lanes to both sides of a pedestrian road, colored lines appear e.g. for Weserpromenade, how it should be. But, it is only shown on one side for my [example way](https://www.openstreetmap.org/way/59038921). I think it's because of an overlaying area tagged as pedestrian zone.

Yes, overlaying things will overlay, ¯\_(ツ)_/¯. However, pedestrian **areas** shouldn't actually be highlighted at all in this overlay and thus also not appear in front of these things. I will have a look.

SebSeb11:
Pedestrian road could be made visible like foot paths:

westnordost:
Ah I see, well, makes sense, but I'll implement that separately.

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
```diff
@@ -95,56 +113,71 @@ class StreetCyclewayOverlayForm : AStreetSideSelectOverlayForm<CyclewayAndDirect
...
+                updateStreetSign()
+            })
+        }
+        if (bicycleInPedestrianStreet != BicycleInPedestrianStreet.NOT_SIGNED) {
+            result.add(AnswerItem(R.string.pedestrian_zone_no_sign) {
+                bicycleInPedestrianStreet = BicycleInPedestrianStreet.NOT_SIGNED
+                updateStreetSign()
+            })
+        }
+        return result
```
FloEdelmann, line 140: Would you consider this to be more readable?

```kt
        return listOfNotNull(
            AnswerItem(R.string.pedestrian_zone_designated) {
                bicycleInPedestrianStreet = BicycleInPedestrianStreet.DESIGNATED
                updateStreetSign()
            }.takeIf { bicycleInPedestrianStreet != BicycleInPedestrianStreet.DESIGNATED },

            AnswerItem(R.string.pedestrian_zone_allowed_sign) {
                bicycleInPedestrianStreet = BicycleInPedestrianStreet.ALLOWED
                updateStreetSign()
            }.takeIf { bicycleInPedestrianStreet != BicycleInPedestrianStreet.ALLOWED },

            AnswerItem(R.string.pedestrian_zone_no_sign) {
                bicycleInPedestrianStreet = BicycleInPedestrianStreet.NOT_SIGNED
                updateStreetSign()
            }.takeIf { bicycleInPedestrianStreet != BicycleInPedestrianStreet.NOT_SIGNED },
        )
```

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
```diff
@@ -80,13 +88,23 @@ class StreetCyclewayOverlayForm : AStreetSideSelectOverlayForm<CyclewayAndDirect
         }
     }
 
+    private fun onLoadInstanceState(state: Bundle) {
+        bicycleBoulevard = state.getString(BICYCLE_BOULEVARD)
+            ?.let { BicycleBoulevard.valueOf(it) }
+            ?: BicycleBoulevard.NO
+        bicycleInPedestrianStreet = state.getString(BICYCLE_IN_PEDESTRIAN_STREET)
+            ?.let { BicycleInPedestrianStreet.valueOf(it) }
```
FloEdelmann, line 96: Why does `bicycleBoulevard` get a fallback value and `bicycleInPedestrianStreet` doesn't?

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
```diff
@@ -193,18 +226,12 @@ class StreetCyclewayOverlayForm : AStreetSideSelectOverlayForm<CyclewayAndDirect
     }
 
     override fun onClickOk() {
-        if (bicycleBoulevard == BicycleBoulevard.YES) {
-            val tags = StringMapChangesBuilder(element!!.tags)
-            bicycleBoulevard.applyTo(tags, countryInfo.countryCode)
-            applyEdit(UpdateElementTagsAction(element!!, tags.create()))
+        // only tag the cycleway if that is what is currently displayed
```
FloEdelmann, line 229: Does this comment still make sense? I think now everything that changed is always tagged.

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
```diff
@@ -95,56 +113,71 @@ class StreetCyclewayOverlayForm : AStreetSideSelectOverlayForm<CyclewayAndDirect
...
+    private fun createSwitchBicycleBoulevardAnswer(): IAnswerItem? =
+        when (bicycleBoulevard) {
+            BicycleBoulevard.YES ->
+                AnswerItem2(getString(R.string.bicycle_boulevard_is_not_a, getString(R.string.bicycle_boulevard))) {
+                    bicycleBoulevard = BicycleBoulevard.NO
+                    updateStreetSign()
+                }
+            BicycleBoulevard.NO ->
+                // don't allow pedestrian roads to be tagged as bicycle roads
+                // (should rather be R.string.pedestrian_zone_designated
```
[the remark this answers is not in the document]
westnordost: clarified comment

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
westnordost: a little bit, I'll take it!

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
westnordost: because then it is null. (null means that it is not a pedestrian street, in that case)

app/src/main/java/de/westnordost/streetcomplete/overlays/cycleway/StreetCyclewayOverlayForm.kt
westnordost: true

westnordost:
Thanks for the review, @FloEdelmann !

FloEdelmann:
You're welcome :slightly_smiling_face: 

For reference, the suggestions were applied in https://github.com/streetcomplete/StreetComplete/commit/6dbea875337cc94d9387e51c07ba5a80c6b9a552.

PR #6022, opened 2024-11-23
Moving the Building Levels quest form to Jetpack Compose.

GaeaKat:


GaeaKat:
Spotted a couple flaws, fixing

westnordost (commented):
Hey, thank you for that contribution!

For being new to Compose, you are already doing a lot of things correctly, IMO.

Here's a small pre-review, without having looked at how it looks like in the app:

1. Some formatting could be improved. I think ktlint could help you there, @FloEdelmann is the expert regarding that. Things like space after `if`, space before `{`, space around operators like `!=`, space after `,` etc

2. For things like buttons (AddBuildingLevelsButton), try to use standard components. Doesn't using `Button` as the container for the row of text, image, text produce a satisfactory result? The issues with defining your own shapes resembling buttons are amongst other things 1. doesn't look right in dark mode, 2. no tap (ripple) feedback, 3. isn't changed automatically when app-wide theme changes are made (such as using Material 3 instead of Material 2)

3. avoid specifying the font size directly, use styles wherever possible, e.g. `style = MaterialTheme.typography.caption`  or `MaterialTheme.typography.body2` (See the `Typography` class for the available text styles in Material2). Same goes for other styling, e.g. regarding colors (for errors etc).

4. Are you sure about `wrapContentSize`/`wrapContentWidth`? I think it is superfluous / not doing what you think it does.

5. don't put compose code into fragments. So there's this levels+roof levels input field? That's one widget, put it into an own file! So there is this form consisting of the mentioned input field plus a row of previous-selection buttons? Put that into yet another file! There's that row of previous-selection buttons? Put that at least into its own function.

6. For the levels+roof levels input, you want to hoist the state, see https://developer.android.com/develop/ui/compose/state#state-hoisting . I.e. the function should take the levels as normal String parameters and instead have a function `onValuesChanged` or similar to report the changed value(s) to the parent. Basically, how `TextField` works, too.

GaeaKat:
Screenshot 2024-11-27 at 7 09 49 PM
My attempt at using Button lead to this, and it was more time and code removing the default things, then adding clickable to a row, especially as according to docs and testing, using the clickable modifier does the ripple feedback. though using the theme more so stuff like dark mode works is needed.

westnordost:
Well, you have several paddings there in nested composables (of which some are not necessary, by the way).
That the button is blue, is I think okay, that's the material style after all.

But if it really doesn't work out at all, you should use `Surface`, see e.g. `MapButton`.

GaeaKat:
Screenshot 2024-11-27 at 8 46 13 PM
This is what I end up with after I have trimmed the button down / using Button component.

GaeaKat:
Converting this back to draft while I work on the issues presented. thankyou for the points! :)

GaeaKat:
Moved out of draft, ran ktlint over the files I changed now, moved it to it's own file, and moved the row to it's own file. and moved MaterialTheme.color and MaterialTheme.Typography

westnordost:
So, is this ready to review/merge? Or do you want to review it yourself, first? Some small things like formatting etc. one might notice oneself.

GaeaKat:
Skittles be ready I think

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -23,96 +23,86 @@ class AddBuildingLevelsForm : AbstractOsmQuestForm<BuildingLevelsAnswer>() {
...
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-
-        if (savedInstanceState == null) {
-            binding.levelsInput.setText(element.tags["building:levels"])
-            binding.roofLevelsInput.setText(element.tags["roof:levels"])
+        binding.questBuildingLevelsBase.setContent {
+            regularLevels = rememberSaveable { mutableStateOf(if (element.tags["building:levels"] != null) element.tags["building:levels"] else "") }
+            roofLevels = rememberSaveable { mutableStateOf(if (element.tags["roof:levels"] != null) element.tags["roof:levels"] else "") }
```
FloEdelmann, line 46:
```suggestion
            regularLevels = rememberSaveable { mutableStateOf(element.tags["building:levels"] ?: "") }
            roofLevels = rememberSaveable { mutableStateOf(element.tags["roof:levels"] ?: "") }
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsForm.kt
```diff
@@ -23,96 +23,86 @@ class AddBuildingLevelsForm : AbstractOsmQuestForm<BuildingLevelsAnswer>() {
...
-        binding.levelsInput.setText(buildingLevelsAnswer.levels.toString())
-        binding.roofLevelsInput.setText(buildingLevelsAnswer.roofLevels?.toString() ?: "")
     }
 
     override fun onClickOk() {
-        val answer = BuildingLevelsAnswer(levels!!, roofLevels)
-        prefs.addLastPicked(this::class.simpleName!!, listOfNotNull(levels, roofLevels).joinToString("#"))
+        val answer = BuildingLevelsAnswer(
+            regularLevels.value?.toInt() ?: 0,
+            if (roofLevels.value != null) roofLevels.value?.toInt() else null
```
FloEdelmann, line 74:
```suggestion
            regularLevels.value?.toInt() ?: 0,
            roofLevels.value?.toInt() ?: null
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsComponent.kt
```diff
@@ -0,0 +1,135 @@
...
+                    .weight(1f)) {
+                    TextField(
+                        roofLevels ?: "",
+                        onValueChange = onRoofLevels,
+                        keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number, imeAction = ImeAction.Done),
+                        modifier = Modifier
+                            .padding(0.dp, 12.dp)
+                            .align(Alignment.CenterHorizontally)
+                            .conditional(regularLevels != null) { focusRequester(focusRequester) }
+                            .conditional(!(roofLevels != null && roofLevels.isDigitsOnly())) { border(2.dp, color = MaterialTheme.colors.error) },
```
FloEdelmann, line 96:
```suggestion
                            .conditional(roofLevels == null || !roofLevels.isDigitsOnly()) { border(2.dp, color = MaterialTheme.colors.error) },
```

app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevelsComponent.kt
```diff
@@ -0,0 +1,135 @@
...
+                        modifier = Modifier.padding(0.dp, 12.dp),
+                    )
+                    TextField(
+                        regularLevels ?: "",
+                        onValueChange = onRegularLevels,
+                        keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number, imeAction = ImeAction.Next),
+                        modifier = Modifier
+                            .padding(vertical = 9.dp)
+                            .conditional(regularLevels == null) { focusRequester(focusRequester) }
+                            .conditional(!(regularLevels != null && regularLevels.isDigitsOnly())) { border(2.dp, color = MaterialTheme.colors.error) },
```
FloEdelmann, line 68:
```suggestion
                            .conditional(regularLevels == null || !regularLevels.isDigitsOnly()) { border(2.dp, color = MaterialTheme.colors.error) },
```

GaeaKat:
Thankyou for the reviewed changes @FloEdelmann  you are right that looks much better!

westnordost (commented):
Overall, it looks pretty good! You did a lot of things in Compose right, right away.

There are a many small things, so I rather worked on these myself then to complain about (and explain) all these. If you like to learn, have a look at what I changed.

Some general things specially for StreetComplete:
- max line width is 100 chars
- previews should be private
- I usually add a short documentation comment for every public composable
- formatting should be consistent: if it fits in one line, fine, if not, expand not like this

```kotlin
Row(modifier = Modifier
    .height(52.dp), verticalAlignment = Alignment.CenterVertically) {
```

  but like this

```kotlin
Row(
    modifier = Modifier.height(52.dp),
    verticalAlignment = Alignment.CenterVertically
) {
```

Some general things for Compose:
- gotta check if it works well in right-to-left layouts.
imagen
- for paddings between items in a Row, Column, LazyRow etc., use the *Arrangement parameter (e.g. `horizontalArrangement = Arrangement.spacedBy(16.dp)`)
- public composables should always have a `Modifier` parameter, for consistency. This parameter should always be passed to the immediate child composable
- The items function in LazyRow and LazyColumn has an overload that allows you to directly pass a List<T>, so that's a little more convenient than with indices
- check out https://developer.android.com/develop/ui/compose/layouts/intrinsic-measurements for e.g. expanding the height of an element to the maximum height within a row. Very handy. The article explains why fillMaxHeight() is often not what one wants

westnordost:
I didn't review the `AddBuildingLevelsForm` yet. I do think something needs to be changed where the compose world meets the Android world, but to say something definitive, I need to look closer myself, which i will do later.

But the outline is:
- values for the text fields should be TextFieldValues
- for the form, it shouldn't matter whether the input is changed by typing or by clicking on the buttons. (I.e. one callback is enough)
- the trick with errors is to not even allow in the first place for non-numbers to be entered

burrscurr:
> It would be great if there was some indicator whether a list of "favourite" buttons like the BuildingLevelsButton can be scrolled horizontally.
> 
> How is this usually indicated? How is it usually done one iOS?

I had a look at some of the iOS apps on my phone and found two occurrences of a horizontally scrollable list of items. First was in the github app, second one in Overcast ([which uses SwiftUI](https://marco.org/2024/07/16/overcast-rewrite)). In both cases, the list items are simply clipped at the right edge of the screen. If the gap between list items is smaller than the usual padding to the screen edge, there should always be a visual indication that there are more items available when scrolling.

Here is a screenshot of Overcast, both the recent episodes at the top and the playlists section below can be scrolled horizontally.


I think this approach is both simple and functional (I'm no UI designer though).

mnalis:
> I think this approach is both simple and functional (I'm no UI designer though).

But it won't really work unless the item are made / positioned in such a way that the last visible one ***never*** fits on the screen if there are more entries after it -- i.e. for it to be intuitive, part of the rightmost item must _always_ be obscured / falling off the screen (on every screen size), or almost nobody will ever even try horizontal scrolling there if it looks like there is just a normal fixed row...

For example, in this EveryDoor example, in House address like (:house:), it is obvious that there is a scrollable list, but if that credit card line (:credit_card:) were just a few pixels shorter, it would've been completely undiscoverable (and yes, for some time I didn't know I could scroll for more answers). 

small_Screenshot_20241215_191802_Every Door

---


I think maybe fadeout of rightmost answer (perhaps with arrow pointing to the right to make it explicit) would be more intuitive.

westnordost:
Okay, so the result of that little excursion is that also on iOS, there is no standard for indicating that a view is scrollable. The only thing that could be done is to flash a scrollbar (or always display it). That *may* be an option, but not currently, because scroll bars are not supported yet by Compose.

Anyway, @GaeaKat , would you like to look into the stuff I outlined yourself, or would you rather move on and leave it to me?

GaeaKat:
I am sorry I must admit I think I misread things! I thought your comment was about stuff you added in the rework you did.  I take it you mean https://github.com/streetcomplete/StreetComplete/pull/6022?new_mergebox=true#issuecomment-2540276702 here?

westnordost:
I meant https://github.com/streetcomplete/StreetComplete/pull/6022#issuecomment-2540276702 just now

GaeaKat:
Taking a look at doing those changes now!

westnordost:
I will look at completing this soon. If you are or currently or planning to do so soon yourself, shout out!

This PR can then serve as a template how a conversion of the UI should look like.

GaeaKat:
> I will look at completing this soon. If you are or currently or planning to do so soon yourself, shout out!
> 
> This PR can then serve as a template how a conversion of the UI should look like.

I honestly think I have misunderstood the issue, I tried to convert to TextFieldValue but I get runtime errors stating that a TextFieldValue can't be stored in the state.

java.lang.IllegalArgumentException: MutableState containing TextFieldValue(text='', selection=TextRange(0, 0), composition=null) cannot be saved using the current SaveableStateRegistry. The default implementation only supports types which can be stored inside the Bundle. Please consider implementing a custom Saver for this class and pass it as a stateSaver parameter to rememberSaveable().

westnordost:
Okay, sorry that it took so long.

After looking at it from all the sides, I understand your last comment and also have to admit that I was wrong: I assumed that always `TextFieldValue` should be used because only then, the cursor position and selection will work properly when editing the text field. However, it turns out that this is false, and it works just as well when passing a `String`. Looking into the source code, at `BasicTextField.kt` line 612, it becomes clear as to why: If a `String` is passed, the text field will manage a `TextFieldValue` internally. Albeit, not with `rememberSavable`, so on configuration change (device rotation etc), the cursor position and selection is lost, but nothing else.

Regarding `rememberSavable` + `TextFieldValue`, that could apparently be done with 
```kotlin
var value by rememberSaveable(stateSaver = TextFieldValue.Saver) { mutableStateOf(TextFieldValue()) }
```
but since this part is currently on the edge of Android view code / Compose code, I think this cannot be used. So anyway, current state is fine.

I also made a comment earlier that the `BuildingLevelsForm` could just handle `Int?`s and reject any non-integer input instead of showing it as an error (i.e. red outline), but in the end, the way it is now is fine, too, because it also handles gracefully a certain edge case when dealing with invalid incoming OSM data.

So, pretty much the two only changes that matters that I made bumbling around in the source code of your PR are:
- simplify code and fix a bug where the form would accept any garbage input for the roof levels if roof levels are optional (https://github.com/streetcomplete/StreetComplete/pull/6022/commits/fd841e31b84f34d8b72761c7725bd34b5a0ffeae)
- replace explicit styling for invalid input with using the `isError` parameter of the text field (https://github.com/streetcomplete/StreetComplete/pull/6022/commits/ad5dbe38371d37c76d3b10e0ed6a6e4b479def2b , then https://github.com/streetcomplete/StreetComplete/pull/6022/commits/217449055daaf6a6c4074f4d167f7ac544b1c3a8)

PR #6061, opened 2024-12-29
handle 410 error code for now gone elements

### LINKED ISSUE #6060: Error when uploading for deleted element
**How to Reproduce**
1. download data for an area
2. delete an element in that area using a different editor and save changes to OSM
3. delete the same element in SC (probably could also edit tags)
4. see "Upload error" dialog, with the error being
`de.westnordost.streetcomplete.data.ApiClientException: Client request(POST https://api.openstreetmap.org/api/0.6/changeset/xxx/upload) invalid: 410 Gone. Text: "The node with the id yyy has already been deleted"`

**Expected Behavior**
The error should be treated like other modifications of that element (throw a `ConflictException`, which will be handled by discarding the edit)

**Versions affected**
SC 60.0 (F-Droid)

matkoniecz:
fixes #6060

note: if wiki documentation is to be believed, HttpStatusCode.NotFound should never ever be triggered by SC

should we handle it this way?

Note: this code is untested

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiClient.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiClient.kt
@@ -64,7 +64,9 @@ class MapDataApiClient(
                 HttpStatusCode.Conflict,
                 // an element referred to by another element does not exist (anymore) or was redacted
                 HttpStatusCode.PreconditionFailed,
-                // some elements do not exist (anymore)
+                // some elements do not exist anymore as it was deleted
+                HttpStatusCode.Gone,
+                // some elements do not exist and never existed
                 HttpStatusCode.NotFound -> {
                     throw ConflictException(e.message, e)
                 }


matkoniecz:
@Helium314 if you would be interested in testing/review...

kmpoppe:
Also gonna try testing that later. Wonder if I can make the dev version attach to the testing servers 🤔

matkoniecz:
api url is defined somewhere

from quick grepping I found

`app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt:private const val OSM_API_URL = "https://api.openstreetmap.org/api/0.6/"`

which would need to be modified

kmpoppe:
Yep, that's it, thanks 👍

kmpoppe:
Well, no DEV API, because the OAuth2 Authorization didn't work (Client not registered), but anyway...

https://www.openstreetmap.org/node/311705840 was loaded in SC (does this still exist?), then removed on osm.org, and then the quest was answered.

SC successfully created [CS 160764906](https://osm.org/changeset/160764906), but didn't upload any changes into the CS and did NOT close it. Also it quietly ate the change and the exception, but put this friendly message in the log:

```
2024-12-29T19:16:39.584: I [ChangesetManager] Created changeset #160764906
2024-12-29T19:16:40.258: D [ElementEditsUploader] Dropped a UpdateElementTagsAction: Element deleted
```

Is this the expected behaviour when GONE was caught?

matkoniecz:
> Is this the expected behaviour when GONE was caught?

yes, in cases of unrecoverable edit conflicts SC throws away edit

this should happen very rarely and recovering from that would require very elaborate interface and investigation not doable by newbies

> SC successfully created [CS 160764906](https://osm.org/changeset/160764906), but didn't upload any changes into the CS and did NOT close it.

this is also expected, to upload edit (and discover potential edit conflicts) CS needs to be created
and it is now ready for potential further edits

> Well, no DEV API, because the OAuth2 Authorization didn't work (Client not registered), but anyway...

I guess you also need to register oauth client

kmpoppe:
> yes, in cases of unrecoverable edit conflicts SC throws away edit
[...]
> this is also expected, to upload edit (and discover potential edit conflicts) CS needs to be created and it is now ready for potential further edits

Then I guess the that PR works as intended and could be marked RFR.
 
> I guess you also need to register oauth client

`OAUTH2_CLIENT_ID` is a static value in UserModule.kt and must have been registered by @westnordost - registering the Application in the DEV environment would probably yield a different client id. So, maybe we need a whole different "api dev mode" to be worked into the sources, but that's a whole new can of worms :)

westnordost:
Ah, you tested the case and in that case, SC throws away the 'delete' edit silently?

kmpoppe:
> Ah, you tested the case and in that case, SC throws away the 'delete' edit silently?

Yep, the "change counter" shortly added 1 and then reverted back to the original value (3187) and did not throw any message. I would have expected a message box telling me the information that was put into the log, but I don't know if there's an option to show/hide that messages in the app options?

westnordost:
That's fine then, thanks for testing!

matkoniecz:
> OAUTH2_CLIENT_ID is a static value in UserModule.kt and must have been registered by @westnordost - registering the Application in the DEV environment would probably yield a different client id. So, maybe we need a whole different "api dev mode" to be worked into the sources, but that's a whole new can of worms :)

I meant that someone may register new editing app at dev server specifically by themselves and change oauth fields

I guess there could be also an official setup for dev server

kmpoppe:
> I meant that someone may register new editing app at dev server specifically by themselves and change oauth fields

Yeah, classic brain f**t ... Created an app in my dev settings, used the correct redirect url and got it to work ... Thanks for the nudge ;-)

PR #6066, opened 2024-12-31
update KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED

mnalis:
with some new keys from https://github.com/mnalis/StreetComplete-taginfo-categorize update


kmpoppe (changes requested):
https://github.com/mnalis/StreetComplete-taginfo-categorize/issues/30#issuecomment-2566140900

Issue #6057, opened 2024-12-28
Tag `source:opening_hours` should be deleted when updating a new place in the Places overlay


jdejaegh:
When a shop changed, the Places overlay does not delete the `source:opening_hours` tag from the object.  
The `source:opening_hours` is a documented tag: https://wiki.openstreetmap.org/wiki/Key:source:opening_hours

**How to Reproduce**
1. Find a shop that has at least those tags: `name`, `opening_hours`, `source:opening_hours`
2. Using the Places overlay, change the name of the shop and confirm that it is a new place
3. Observe that `name` is changed, `opening_hours` is deleted and `source:opening_hours` remains

**Expected Behavior**
The tag `source:opening_hours` should be deleted.

**Versions affected**
StreetComplete v60.0
Android 13

matkoniecz:
Can you link an effected object?

matkoniecz:
https://github.com/mnalis/StreetComplete-taginfo-categorize would need to be updated to fix it, if it happens

jdejaegh:
> Can you link an effected object?

Sure, here is one: https://www.openstreetmap.org/node/1210562035/history/13

Version 13 was created by StreetComplete (without deleting the tag `source:opening_hours`).  I manually deleted the tag in version 14.

westnordost:
Hmm, after looking at the PR, I am not sure if removing `source:<key>` when `<key>` is deleted is the wisest thing to do automatically. Because:

1. it opens a can of worms. There are quite a few more keys that are considered _meta keys_. Some more so, some less, some rather only for a certain group of features, etc.. If `source:` is treated specially, for consistency, so should others (`note:`?), but we will have the recurring discussion whether that _meta key_ can always be safely deleted when the key is deleted.

2. Even for `source` which is somewhat seen as a meta key, there are outliers, most prominently, `source:maxspeed`, which can hold valuable data also without an accompanying `maxspeed` value. (`source:maxspeed` is used 2.4M times, `source:opening_hours` is used 4k times) `source:maxspeed` should not be deleted when `maxspeed` is deleted.

So, in summary, I think it is too risky to treat meta keys in a special way.

I think it is best to treat such keys like any other. @mnalis, which `source:`... keys are generally in (enough) use to be included in _StreetComplete-taginfo-categorize_?

mnalis:
[...]

PR #6067, opened 2024-12-31
remove not only $key, but also source:$key and check_date:$key

mnalis:
when trimming keys in `KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED`, remove not only every `$key` that matches, but also every `source:$key` and `check_date:$key` that matches too. 

Rationale being that if we are removing no longer relevant key, we should remove its metadata too (e.g. from where it initially came, and when it was last verified).

- Fixes https://github.com/streetcomplete/StreetComplete/issues/6057
- Fixes https://github.com/mnalis/StreetComplete-taginfo-categorize/issues/30

Tested with [debug apk ](https://github.com/mnalis/StreetComplete/actions/runs/12554714035) on e.g. https://www.openstreetmap.org/node/832042729 which has `source:opening_hours` and which regular SC does not remove...

(it also adds `--info` to `gradlew test` invocation in GitHub action, to report reason in case of test failure - which I needed and think is useful all around when we run tests)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Place.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Place.kt
@@ -249,7 +249,7 @@ fun StringMapChangesBuilder.replacePlace(tags: Map<String, String>) {
     removeCheckDates()
 
     for (key in keys) {
-        if (KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED.any { it.matches(key) }) {
+        if (KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED.any { it.matches(key) || it.matches("source:" + key) }) {
             remove(key)
         }
     }
--- a/app/src/main/java/de/westnordost/streetcomplete/osm/Place.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/osm/Place.kt
@@ -248,9 +248,9 @@ val POPULAR_PLACE_FEATURE_IDS = listOf(
 fun StringMapChangesBuilder.replacePlace(tags: Map<String, String>) {
     removeCheckDates()
 
-    private val REMOVABLE_KEYS = KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED +
-                                 KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED.map { "source:$it" } +
-                                 KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED.map { "check_date:$it" }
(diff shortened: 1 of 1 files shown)


matkoniecz (changes requested):
Can you add test or two?

I would test that:

- lone source:opening_hours is removed
- when source:opening_hours and opening_hours are there both are removed

(I can also add them if you do not want, let me know in such case)

westnordost:
Hmm, what about Things, then?

I was thinking that maybe it would be a better solution if this is done for **any** edit? (I.e. if `key` is edited, `source:key` is deleted?)

mnalis:
> I was thinking that maybe it would be a better solution if this is done for any edit?

Agreed! Here is ([debug apk](https://github.com/mnalis/StreetComplete/actions/runs/12575449504) that does it that way (in `fun remove()`) if others want to try it out. I've quickly checked manually on https://www.openstreetmap.org/node/832042729 and it seems to work.

> Can you add test or two?

Good idea, will try

mnalis:
Current revision of this PR seems to be [successful](https://github.com/mnalis/StreetComplete/actions/runs/12605393625) in additionally removing `source:$key` when `$key` is removed (And added test seems to confirm it), so in its current state it would fix specific problem reported in #6057. 

---

However, I think it would be good idea to also remove `check_date:$key` and other related date keys, as those are also obsolete when `$key` is removed.

However, trying to do that breaks various tests (see `FIXME`s in this PR) as it seems to be removing `check_date:$key` in cases when instead a modification was done (so check date should be _updated_ instead of removed). 

I've been trying to look into it, but it is at boundary of my abilities. I'll give it another try after several days (going AFK now), so:
- if anyone in the meantime cares to gives me a hint or two where to look, it would be greatly appreciated. Or,
- if anyone more skilled want to take it from here instead, I'm good with that too. Or, 
- otherwise, this PR could also be merged as-is, as it fixes `source:$key` issue successfully, and fixing other related meta-keys (`check_date:$key`) and friends) can be left as an exercise for another time.

westnordost:
Did you read https://github.com/streetcomplete/StreetComplete/issues/6057#issuecomment-2571377199 ?

mnalis:
> Did you read https://github.com/streetcomplete/StreetComplete/issues/6057#issuecomment-2571377199 ?

***Now*** I did :open_mouth: :smiley: 

I've thus abandoned the idea of generic use of `tags.remove()` removing meta-prefixes like `source:`

However, I've rolled back this PR to only remove `source:x` and `check_date:x` if `x` is listed in `KEYS_THAT_SHOULD_BE_REMOVED_WHEN_PLACE_IS_REPLACED` and user has selected that that that shop/amenity no longer exists. Would that work?

matkoniecz:
I think so? While doing it in general would be potentially problematic, tagging space for shops should be smaller.

And here this tag removal on shop retagging is anyway a heuristic - quite smart and where a lot of effort went into it, but still not 100% accurate either.

mnalis:
(ref: https://github.com/streetcomplete/StreetComplete/commit/65bba70b714a0273313c3051fe89c7f01fd5ec27)

PR #6069, opened 2024-12-31
Allow developers to use Test-API

kmpoppe:
When developing SC where multiple changes to objects might be neccessary, it might be helpful to not do this sort of updates to an object in the Live Database but rather the Development/Testing one.

This PR creates a boolean application variable which will switch the API the software is working against.

`OAUTH2_CLIENT_ID` is an application registered to my dev-user, this could be changed to anything in @westnordost 's domain.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
@@ -84,6 +84,9 @@ object ApplicationConstants {
     During development it might be better to work against the Test-API, rather than the
     Live-API. Developers are reminded that they need a separate login for the Test-API and
     can register/logon via https://master.apis.dev.openstreetmap.org/
+    Note that test actions not applied to the database do not require test API (test edits that are reverted
+    locally before an upload etc) and that test API has a separate database that is mostly empty
+    (test data needs to be created there).
      */
     const val USE_TEST_API = false
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
@@ -79,4 +79,11 @@ object ApplicationConstants {
            locally for performance reasons */
         SplitWayAction::class
     )
+
+    /*
+    During development it might be better to work against the Test-API, rather than the
+    Live-API. Developers are reminded that they need a separate login for the Test-API and
+    can register/logon via https://master.apis.dev.openstreetmap.org/
+     */
+    const val USE_TEST_API = false
 }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
@@ -16,13 +16,13 @@ import org.koin.androidx.workmanager.dsl.worker
 import org.koin.core.qualifier.named
 import org.koin.dsl.module
 
-private var OSM_API_URL = "https://api.openstreetmap.org/api/0.6/"
+private const val OSM_API_URL_LIVE = "https://api.openstreetmap.org/api/0.6/"
+private const val OSM_API_URL_TEST = "https://master.apis.dev.openstreetmap.org/api/0.6/"
+
+val OSM_API_URL =
+    if (USE_TEST_API) OSM_API_URL_TEST else OSM_API_URL_LIVE
 
 val osmApiModule = module {
-
-    if (USE_TEST_API) {
-        OSM_API_URL = "https://master.apis.dev.openstreetmap.org/api/0.6/"
-    }
 
     factory { Cleaner(get(), get(), get(), get(), get(), get()) }
     factory { CacheTrimmer(get(), get()) }
--- a/app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
@@ -1,5 +1,6 @@
 package de.westnordost.streetcomplete.data
 
+import de.westnordost.streetcomplete.ApplicationConstants.USE_TEST_API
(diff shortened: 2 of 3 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/OsmApiModule.kt
westnordost, line 1: better have two constants and decide on which URL to use based on USE_TEST_API in the `module` function

app/src/main/java/de/westnordost/streetcomplete/data/user/UserModule.kt
westnordost, line 1: I'd very much prefer if those stayed constants. Like suggested solution above: private const string values, then the public values `val X = if (test api) test_api_endpoint else dev_api_endpoint`

westnordost (changes requested):

mnalis:
I'd much prefer if there was on option under `Settings` / `Debug` when doing debug build -- e.g. just like there is currently `Show quest form` (only available under debug builds), then also add a (non-persistent) boolean tooggle there `Use dev OSM servers` which chooses between them.

---

Because, if one has to change the code and recompile to reach that functionality, then we might as well just add commented-out-by-default extra values below, e.g.:

```
const val OAUTH2_TOKEN_URL = "https://www.openstreetmap.org/oauth2/token"
const val OAUTH2_AUTHORIZATION_URL = "https://www.openstreetmap.org/oauth2/authorize"
const val OAUTH2_CLIENT_ID = "Yyk4PmTopczrr3BWZYvLK_M-KBloCQwXgPGEzqUYTc8"

// for dev API use those instead:
// const val OAUTH2_TOKEN_URL = "https://master.apis.dev.openstreetmap.org/oauth2/token"
// const val OAUTH2_AUTHORIZATION_URL = "https://master.apis.dev.openstreetmap.org/oauth2/authorize"
// const val OAUTH2_CLIENT_ID = "ObZ7yPf4lfs4XJ3NWysI3ukJMN0SHey1oPnNQnLmvw8"
```

westnordost:
@mnalis Using the test api barely makes sense for StreetComplete, only in very specific cases, because the test api is not populated with actual data. But StreetComplete needs some actual data to work with. So, to me it is fine to make it a bit easier to test with the test API, but this is not a common use case, not even for developers of the app.

mnalis:
> but this is not a common use case, not even for developers of the app.

Well, I've needed it quite a few times (I resorted in just adding fake data in regular API and later reverting it :shrug:, as alternatives were too much work). I think it it would not be hard to add the extra toggle for that here:

https://github.com/streetcomplete/StreetComplete/blob/df50579c307bbe08b186495fd24add96cff0d97c/app/src/main/java/de/westnordost/streetcomplete/screens/settings/SettingsScreen.kt#L175-L181

>  Having a setting for that in the debug app might create wrong expectations.

Like what? I mean, not that many people are building developer debug versions I guess, and most of them I would think are aware of what OSM dev API is / why they'd want to use it (instead of them doing debug build and then going willy-nilly enabling random debug options they have no idea what they do and then complaining that it their map doesn't look like they want).

---

I fully understand if @kmpoppe prefers not to do it because it is too much work (and I agree that existing state of PR is improvement compared to current master), but if they (or someone else)  are fine with doing that work, I  think that having it as a Debug build  would be much **more useful and would result in less fake data landing in regular in main OSM database**

kmpoppe:
I'll take a look into both of your comments, there's a new year's party happening right now ;-)

mnalis:
On Tue, Dec 31, 2024 at 11:55:39AM -0800, Kai Michael Poppe wrote:
> I'll take a look into both of your comments, there's a new year's party happening right now ;-)

Happy new year everyone then :)

app/src/main/java/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -79,4 +79,11 @@ object ApplicationConstants {
            locally for performance reasons */
         SplitWayAction::class
     )
+
+    /*
+    During development it might be better to work against the Test-API, rather than the
+    Live-API. Developers are reminded that they need a separate login for the Test-API and
+    can register/logon via https://master.apis.dev.openstreetmap.org/
+     */
```
matkoniecz, line 87:
```suggestion
    Note that test actions not applied to the database do not require test API (test edits that are reverted
    locally before an upload etc) and that test API has a separate database that is mostly empty
    (test data needs to be created there).
     */
```

app/src/main/java/de/westnordost/streetcomplete/data/user/UserModule.kt
```diff
@@ -36,6 +37,12 @@ val OAUTH2_REQUIRED_SCOPES = listOf(
 
 val userModule = module {
 
+    if (USE_TEST_API) {
+        OAUTH2_TOKEN_URL = "https://master.apis.dev.openstreetmap.org/oauth2/token"
+        OAUTH2_AUTHORIZATION_URL = "https://master.apis.dev.openstreetmap.org/oauth2/authorize"
+        OAUTH2_CLIENT_ID = "ObZ7yPf4lfs4XJ3NWysI3ukJMN0SHey1oPnNQnLmvw8"
```
matkoniecz, line 43: @westnordost do you want own client_id for test API or is this fine?

kmpoppe:
@matkoniecz before merging, let me see if making an option available in the settings looks any good.

app/src/main/java/de/westnordost/streetcomplete/data/user/UserModule.kt
westnordost: fine for me to not own it.

westnordost (approved):
Looks good!

kmpoppe:
@westnordost thanks :)
@mnails Even though it would've been a nice idea to use the SettingsProvider, accessing prefs in the OsmApiModule and UserModule was a hassle and changing the setting on the fly only worked to a degree (The OAuth token didn't get reset even though I called userLoginController.logout), so I'll leave it for now. Maybe inspiration strikes at some time :)

PR #6074, opened 2025-01-02
update dependencies and android target version

westnordost:
most of the changes are because when targeting Android SDK level 35, all screens are edge-to-edge by default. so one needs to specify the insets manually.

It's quite complex to test this, as it may appear different for any combination of:
- portrait or landscape mode (navigation is on the left or right side, but only if it is button navigation)
- with display cutout (needs additional padding, can be located anywhere depending on display rotation)
- right-to-left layout (expected location of the display cutout within the layout differs)
- using gesture navigation vs using button-navigation (background of the navigation changes)
- dark mode or light mode (color of the status bar icons changes)

E.g. when rotated to landscape mode, the display cutout will be on the left (or right) side, so there also needs to be a padding there.

But I think I tested it all. All screens are affected.

The visible changes are:
- top bar has no background anymore (seems to be standard now in Android)
- nav bar uses the default color now (background color almost-opaque when using navigation buttons, otherwise transparent)
- main screen regards necessary padding for the display cutout (😕 which may look odd, sometimes, but we don't want that a button is below the display cutout)
- you see some content below the navigation bar on scrollable screens (e.g. settings screen)
- all the other screens other than the main screen also add a padding for the display cutout

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -107,7 +107,7 @@ repositories {
 }
 
 dependencies {
-    val mockitoVersion = "3.12.4"
+    val mockitoVersion = "5.14.2"
 
     coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:2.1.4")
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -24,7 +24,7 @@ android {
         }
     }
 
-    compileSdk = 34
+    compileSdk = 35
     testOptions {
         unitTests {
             isReturnDefaultValues = true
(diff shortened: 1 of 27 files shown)


app/build.gradle.kts
```diff
@@ -109,7 +109,7 @@ repositories {
 dependencies {
     val mockitoVersion = "3.12.4"
```
kmpoppe, line 110:
```suggestion
    val mockitoVersion = "5.14.2"
```

Update mockito to the newest version

PR #6075, opened 2025-01-02
OsmNoteQuestController: surveyme override and regex test update

### LINKED ISSUE #6052: #surveyme in last Note comment should show the Note (even if there are no preceding comments)
Back in https://github.com/streetcomplete/StreetComplete/issues/2641 (via https://github.com/streetcomplete/StreetComplete/commit/b3ad39b2130dabdf0db2132dabeb3eaaa2ad48c4), it was made that `#surveyme` tag in the last comment would always (unless manually hidden by the user) result in showing the OSM Note, even if that comment was created by the same user which is using the app.

In SC 60.0 it works *mostly*,  i.e. if that comment is any ***but*** the first one.

I don't think that is intentional (I see no logic for that, nor do the comments seem to indicate the reason for such strange behaviour). 
If the user's **last comment** contains a marker indicating the survey is required and we want to display the note in that case (as we do), then it **_should not matter how many comments preceded that last one_** (0 or more).

**How to Reproduce**
1) add a comment to existing note with just text `test`; and note gets hidden - OK
2) add a comment to existing note with text `#surveyme test`; and note remains displayed - OK
3) create a new note with text `#surveyme test` -- note gets hidden - BAD

**Expected Behavior**
- in step (3), note should remain displayed just as in does in (2), as the last message indicates the survey is required.

**Versions affected**
SC 60.0, Android 14

kmpoppe:
Resolves #6052.

* Restructured exclusion of notes that get overridden by #surveyme
* Added RegexOption.DOT_MATCHES_ALL to function NoteComment.containsSurveyRequiredMarker: #surveyme wasn't found at the end of the text added by the user, because SC add "\n\nvia StreetComplete <x.y>" to the text

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
@@ -256,7 +256,7 @@ private fun Note.containsSurveyRequiredMarker(): Boolean =
     comments.any { it.containsSurveyRequiredMarker() }
 
 private fun NoteComment.containsSurveyRequiredMarker(): Boolean =
-    text?.matches(".*#surveyme.*".toRegex(setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.IGNORE_CASE))) == true
+    text?.contains("#surveyme", ignoreCase = true) == true
 
 private fun Note.probablyCreatedByUserInThisApp(userId: Long): Boolean {
     val firstComment = comments.first()
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
@@ -196,15 +196,21 @@ private fun Note.shouldShowAsQuest(
     // don't show notes hidden by user
     if (id in blockedNoteIds) return false
 
-    // don't show notes where user replied last unless he wrote a survey required marker
-    if (comments.last().isReplyFromUser(userId)
+    /*
+        We usually don't show notes where either the user is the last responder, or the
+        note was created with the app and has no replies.
+        If the most recent comment contains "#surveyme", we want to show the note regardless.
+        See https://github.com/streetcomplete/StreetComplete/issues/6052#issuecomment-2567163451
+     */
+    if (
+        (
+            comments.last().isReplyFromUser(userId) ||
+            (probablyCreatedByUserInThisApp(userId) && !hasReplies)
+        )
         && !comments.last().containsSurveyRequiredMarker()
     ) {
         return false
     }
-
-    // newly created notes by user should not be shown if it was both created in this app and has no replies yet
-    if (probablyCreatedByUserInThisApp(userId) && !hasReplies) return false
 
     /*
         many notes are created to report problems on the map that cannot be resolved
@@ -250,7 +256,7 @@ private fun Note.containsSurveyRequiredMarker(): Boolean =
     comments.any { it.containsSurveyRequiredMarker() }
 
 private fun NoteComment.containsSurveyRequiredMarker(): Boolean =
-    text?.matches(".*#surveyme.*".toRegex()) == true
+    text?.matches(".*#surveyme.*".toRegex(RegexOption.DOT_MATCHES_ALL)) == true
 
 private fun Note.probablyCreatedByUserInThisApp(userId: Long): Boolean {
     val firstComment = comments.first()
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
```diff
@@ -250,7 +256,7 @@ private fun Note.containsSurveyRequiredMarker(): Boolean =
     comments.any { it.containsSurveyRequiredMarker() }
 
 private fun NoteComment.containsSurveyRequiredMarker(): Boolean =
-    text?.matches(".*#surveyme.*".toRegex()) == true
+    text?.matches(".*#surveyme.*".toRegex(RegexOption.DOT_MATCHES_ALL)) == true
```
FloEdelmann, line 259: Can we simplify this like this? I guess the regex is not really needed at all.

```suggestion
    text?.contains("#surveyme") == true
```

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
```diff
@@ -196,16 +196,22 @@ private fun Note.shouldShowAsQuest(
     // don't show notes hidden by user
     if (id in blockedNoteIds) return false
 
-    // don't show notes where user replied last unless he wrote a survey required marker
-    if (comments.last().isReplyFromUser(userId)
+    /*
+        We usually don't show notes where either the user is the last responder, or the
+        note was created with the app and has no replies.
+        If the most recent comment contains "#surveyme", we want to show the note regardless.
```
FloEdelmann, line 202: It probably makes sense to add a test case covering this.

kmpoppe:
@westnordost as Florian rightfully asked, I tried adding some tests for the changes I made and got a ton of errors hinting to the mockito-version being out of date. I changed those to the most recent (mockito-core and mockitor-android 5.14.2, mockito-inline is now merged into -core), which makes all tests work again for me.

Is there any reason for not using the newest mockito packages?

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
kmpoppe: done 3fa23fae5d8fd9a1a2865c0700cc5f5c3c8d9477

app/src/main/java/de/westnordost/streetcomplete/data/osmnotes/notequests/OsmNoteQuestController.kt
kmpoppe: I've changed the RegexOptions to also IGNORE_CASE in 3fa23fae5d8fd9a1a2865c0700cc5f5c3c8d9477. Granted, this wasn't what you asked for, but lowerCasing the text would've meant using a locale and I wanted to prevent breaking any internationalization by doing that. Would that agreeable for you?

FloEdelmann: Yes, with the `IGNORE_CASE` option, using a regex does make sense again.

kmpoppe: :-)

Helium314: Well, it could also be `text?.contains("#surveyme", ignoreCase = true)`

kmpoppe: Well why on earth are there sooo many paths to Rome *smh* ... updated in42b7eb5a2897875d8f74bdf2c601ff361de93ae6

westnordost:
No, I think there wasn't a reason for the old version. Never bothered to upgrade because it was working fine and I didn't need any new features.

kmpoppe:
Put the new version as a suggestion in your #6074.
@FloEdelmann can you "tests the tests" with the new code?

ivanbranco:
Slightly off-topic, but I'm curious. @kmpoppe, I used this comment as reference until now: https://github.com/kmpoppe/noteSolver/issues/26#issuecomment-1509729671
there you suggest #survey_me, the LatAm community instead uses #surveyneeded ([source](https://wiki-openstreetmap-org.translate.goog/wiki/ES:LatAm/Proyectos/Resoluci%C3%B3n_de_notas/hashtags?_x_tr_sl=es&_x_tr_tl=en&_x_tr_hl=es&_x_tr_pto=wapp&_x_tr_hist=true#Lista_de_hashtags_de_estados_para_notas)) but StreetComplete supports #surveyme. What set of tags should I use to be more "efficient" (more tools supporting it ecc.)? Do we have a table I can reference to?

PR #6080, opened 2025-01-04
Add ISO-3166 KR to enabledInCountries

### LINKED ISSUE #6076: Enable bus stop ref quest in South Korea
**Use case**

User can easily identify the bus stop ref number in South Korea.
Example 1: https://commons.wikimedia.org/wiki/File:202403_Seoul_Bus_Stop_Seobinggo_Station.jpg
Example 2: https://commons.wikimedia.org/wiki/File:Changdong_Station_East_Bus_Stop.JPG
In the case of Seoul, it consists of five-digit reference numbers(XX-XXX), and there are differences in the digit of reference numbers by region.

kmpoppe:
Resolves #6076.


kmpoppe:
I see the pictures in #6076 to be suitable to believe that the ref quest will make sense in all of South Korea, so I'll merge this as per #3450.

Issue #5272, opened 2023-09-25
fire_hydrant number for Germany


Elsensee:
I've seen that the `AddFireHydrantRef` quest is only enabled for Switzerland and France. I would request for it to be enabled for Germany as well.

**Use case**
I started tagging missing fire hydrants and found out that some have "ref" numbers tagged and some don't. After looking at the sign, I saw that there is a corresponding number to this. It's number 6 on the sign: https://wiki.openstreetmap.org/wiki/File:Hydrantenschild_mit_Legende.svg

I've only seen very few without any such number given on the sign.

**Proposed Solution**
Enable the quest for Germany as well. Probably even add a guiding image for entering it like with the diameter.

qugebert:
Some reference (the issue, where this quest was discussed)
https://github.com/streetcomplete/StreetComplete/issues/3059
And in the pull request @mcliquid explained why it is not enabled in germany
https://github.com/streetcomplete/StreetComplete/pull/4440

And at least in Bavaria, i've never seen any signed reference number for fire hydrants.

Elsensee:
I did find the first issue but it was closed and I don't like to discuss closed issues that much.

The second, I did not find. It's interesting that it seems to vary by state? In Hamburg they appear on every sign.
Wikipedia also mentions that they might not exist in every area: https://de.wikipedia.org/wiki/Hinweisschilder_zu_Stra%C3%9Feneinbauten#Hinweisschilder_f%C3%BCr_Hydranten (The images below also show that some signs have a number, some don't)

Googling "Hydrantennummer" yields some results, which confirms that this definitely is a regional thing.

Doing a quick overpass query shows that even in southern germany some fire_hydrant's have a numeric ref, while others have a text like "Im Garten", etc.

Sooo... I don't know?

westnordost:
Jo, steht auch im verlinkten Wikipediaartikel:

> Bauteilnummer zur eindeutigen örtlichen Zuordnung in einem Versorgungsgebiet (Stadt oder Gemeinde). Je nach Betreiber nicht angegeben.

PR #6082, opened 2025-01-04
Enable fire hydrant refs quest in Germany

tobiasgradl:
Reference codes for fire hydrants seem to be widely used all around Germany. I can confirm that for where I live. For all other federal states of Germany, I found many examples of hydrants with non-trivial reference codes in  [Overpass Turbo](https://overpass-turbo.eu/) with the following query:

```
node
  [emergency=fire_hydrant]
  [ref~"^[A-Z0-9]{3,}$"]
  ({{bbox}});
out;
```


HolgerJeromin:
This was rejected in #5272

rusty-snake:
> For all other federal states of Germany, I found many examples of hydrants with non-trivial reference codes in Overpass Turbo with the following query:

Well, that's a positive search, but what's the percentage of all hydrants?

Also it does not answer whether they are signed on ground or used, with permission, from an database/map of the relevant authority.

matkoniecz:
> Well, that's a positive search, but what's the percentage of all hydrants?

if someone would do such research and it would be convincing that it such confusing hydrant signs are extremely rare then maybe it would be worth creating issue

but doing some spot checks like this is normal, triaging issues takes enormous amount of time anyway

Issue #5417, opened 2023-12-20
Implement `Database` interface on iOS


westnordost:
Requires #5412.

Implement the [`Database.kt`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+Database.kt&type=code) interface for iOS using `libsqlite3`, analogous to android-specific implementation in [`AndroidDatabase`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+AndroidDatabase.kt&type=code)

Currently, the code that creates the database file if it does not exist yet and opens the database in a thread-safe way is done [`StreetCompleteSQLiteOpenHelper`](https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+StreetCompleteSQLiteOpenHelper.kt&type=code) (that inherits functionality from the Android class `SQLiteOpenHelper`. This functionality either needs to be reimplemented in platform-independent code or be reimplemented on iOS side.

westnordost:
Note: https://developer.android.com/kotlin/multiplatform/sqlite is a multiplatform sqlite library, currently in alpha, which may remove the necessity to implement `Database` on iOS separately.

PR #6083, opened 2025-01-05
Run Android instrumentation tests on GitHub Actions

FloEdelmann:
Just like the unit tests are run for every push onto the `master` branch (see https://github.com/streetcomplete/StreetComplete/actions/workflows/test.yml), this PR adds a new GitHub Actions workflow that runs the Android instrumentation tests using https://github.com/ReactiveCircus/android-emulator-runner.

Example workflow run: https://github.com/streetcomplete/StreetComplete/actions/runs/12621672783/job/35168756488

To be discussed:

1. ~These tests are quite resource-heavy, so running them for every push might not be justified. Should I change it to a cron job that runs once every day or every week, maybe only if there are changes to the last run? That might need some more investigation though.~ Actually, I think this is not worth the effort; the Android instrumentation tests take around the same time as the unit tests, and pushes to `master` are not _that_ often, so it should be fine.
2. Currently, they run for API level 34 (Android 14). The app's target level was recently bumped to 35 (see [#6074](https://github.com/streetcomplete/StreetComplete/pull/6074)), however while getting the tests to run I found a bug that was only occurring on API level 34 and below (see https://github.com/streetcomplete/StreetComplete/commit/6bbf0fbd3168571d90c49d4da059d0764fbf8ed3 and https://github.com/streetcomplete/StreetComplete/commit/96f2758ebe12aa76b54e5ed7c0d5e1799dabcd5d). So should we maybe run the tests on the lowest *and* highest supported API level (using [matrix strategies](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow))?


.github/workflows/android-test.yml
```diff
@@ -0,0 +1,64 @@
+name: Gradle Android tests
+
+on:
+  push:
+    branches:
+      - "master"
+      - "actions-instrumentation-tests"
```
FloEdelmann, line 7: Should be removed before merging:

```suggestion
```

kmpoppe:
Q: I seem to be lacking some information, which tests exactly should run here? That aren't the ~2,200 tests from `app/src/test/java/de/westnordost/streetcomplete/` but `app/srv/AndroidTest/...` right?

westnordost:
Tests in androidTests are actually only a few database test. They are of course important, but changes on the DAO that access the database are rare and I expect that whenever I change something there, I will also run these tests at least once.

I would like to get rid of (most of) these tests by moving these to the normal tests. Currently, the tests use the Android API to access the SQLite database. To have these tests work independent of Android, we'd need to use some SQLite Java (or multiplatform??) library just in the tests. In the unit tests, we'd need to implement the `Database` interface, using that SQLite Java library. Then, all these database tests could run in Java without issue, without needing the Android instrumentation.

The Android instrumentation tests may still be necessary for testing a few things, for example if the `AndroidDatabase` is implemented correctly, or for testing any other pure-Android platform code that may remain.

In general, running Android instrumentation tests on the lowest and highest supported Android version sounds reasonable.

westnordost:
Hmm, this sounds good: https://developer.android.com/kotlin/multiplatform/sqlite

Would even solve https://github.com/streetcomplete/StreetComplete/issues/5417

FloEdelmann:
@kmpoppe It's the tests from `app/src/androidTest`, not the ones from `app/src/test` (those already run in the existing [`test` workflow](https://github.com/streetcomplete/StreetComplete/actions/workflows/test.yml)).

> Tests in androidTests are actually only a few database tests.

Yes, most of them are, but there is also the `CountryInfosTest` and the `NameAndLocationLabelTest`, both of which test quite user-facing features. It would of course be good to migrate them to unit tests instead, but that might be hard because they use the Android resource system.

> I expect that whenever I change something there, I will also run these tests at least once

As noted in the PR description, while getting the tests to run I found a bug that was only occurring on API level 34 and below (see https://github.com/streetcomplete/StreetComplete/commit/6bbf0fbd3168571d90c49d4da059d0764fbf8ed3 and https://github.com/streetcomplete/StreetComplete/commit/96f2758ebe12aa76b54e5ed7c0d5e1799dabcd5d). That leads me to say that it still makes sense to automate the Android instrumentation tests, so that they are run in different API versions and also changes that seem unrelated are checked.

> I would like to get rid of (most of) these tests by moving these to the normal tests. […] Then, all these database tests could run in Java without issue, without needing the Android instrumentation.

Agreed! But until then (or for the other tests), I guess this PR still improves the situation.

> running Android instrumentation tests on the lowest and highest supported Android version sounds reasonable

Done that in [`559982b`](https://github.com/streetcomplete/StreetComplete/pull/6083/commits/559982b785ce2bc10ea50fee05fbcffb321164eb). These are the resulting job runs:

- https://github.com/streetcomplete/StreetComplete/actions/runs/13090395525/job/36526352923?pr=6083
- https://github.com/streetcomplete/StreetComplete/actions/runs/13090395525/job/36526352992?pr=6083

.github/workflows/android-test.yml
```diff
@@ -0,0 +1,67 @@
...
```
FloEdelmann, line 7: Should be removed before merging:

```suggestion
```

westnordost:
Regarding the review request: You have my approval, if that is what you are seeking, but other than that I fear cannot say anything more that I already said, you have the expertise regarding the setup of github action workflows.

Btw regarding CountryInfos tests etc: That could/should be migrated once we use Kotlin Multiplatform Resources.

PR #6093, opened 2025-01-08
Update AddPostboxRoyalCypherTest.kt

kmpoppe:
questType.isApplicableTo doesn't check the location against whether the quest is enabled at that location.


FloEdelmann:
Done: https://github.com/streetcomplete/StreetComplete/commit/d13ff395726ca027974a01dfed7460724b52a775

Issue #5934, opened 2024-09-29
Map Maintenance: Handling of `fixme=name` and suspicious combinations of `name=*+noname=yes`


kmpoppe:
### Discussed in https://github.com/streetcomplete/StreetComplete/discussions/5928

<div type='discussions-op-text'>

<sup>Originally posted by **kmpoppe** September 26, 2024</sup>
# Situation

There are currently ~160k objects tagged with `fixme=name`.

Understandably, SC currently ignores `fixme` tags completely - IMHO most values of that key would be impossible to survey without deeper investigation.

Yet, `fixme=name` seems like a very straight-forward request: "I (the last mapper) was unable to determine what the name of this feature is, please help out".

# Proposal

I therefore would suggest for SC to work the following way:

a) remove `fixme=name` (if present) when the user answers quests/road_name or quests/place_name with either a real name or `noname=yes` (because otherwise, things like [this](https://www.openstreetmap.org/way/208400787/history/4) (`noname=yes` set, `fixme=name` persited) or [this](https://www.openstreetmap.org/node/1840355375/history/5) (`name` set, `fixme=name` persisted) happen)

b) the aforementioned quests should force a survey of the name when a `fixme=name` is older than one year to get rid of it, because all viable answer paths from there (`name=*`, `noname=yes` or changing the type of road that doesn't get asked for a name again) lead to a situation where the `fixme` is no longer relevant.

c) `FIXME=name`, `Fixme=name`, `fixme=name?`, `Fixme=Name`, ... could be handled the same way.

# SCEE

I do not think that this option should only be something relevant for SCEE, because the handling is purely under to hood and there's no expert knowledge necessary to answer those already existing quests.</div>

# AE

An automated edit of things that are "clear cut" as proposed by @matkoniecz is [planned](https://wiki.openstreetmap.org/wiki/User:Kmpoppe/Automated_Edits/Fix_Name_Inconsistencies), SC's tagging cleanup is nevertheless needed for remaining instances of fixme occurances.

westnordost:
Why not always ask for the name if `fixme=name` is set?

A note regarding implementation: What if the name is already (partly) set? I am thinking about multilingual names. The previous name(s) should also be shown in the form, if they aren't, yet.

kmpoppe:
> Why not always ask for the name if `fixme=name` is set?

Makes sense

[...]

PR #6097, opened 2025-01-12
Add Bus Stop Name Suggestions, generalize Name Suggestions

### LINKED ISSUE #5187: [FR] give a hint for a bus-stop name
**Use case**
a bus stop needs a name while the one on the other side of the street already has the same name.

**Proposed Solution**
Give the option of the same name as nearby bus-stops. Similar to the streets names.

kmpoppe:
Resolves #5187 (Blocks #5934)

Finds Bus Stop Names around 250 m of the point the quest is asked for.
UI changed from a left-aligned simple text-input to a centered suggestion-enabled input just like road names.

Quest asked for [this node](https://www.openstreetmap.org/node/12249009334), suggestions from https://www.openstreetmap.org/node/123368775 and https://www.openstreetmap.org/node/418728678

Screenshot_20250112_211335

Thoughts? I'd like to get this implemented before I put #5934 to work.


mnalis:
FYI I've built a debug APK at https://github.com/mnalis/StreetComplete/actions/runs/12737099313 (if somebody else wants to test but can't compile their own)...

You @kmpoppe could click on `Actions` / `Build debug APK` in your [GitHub fork](https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#alternative-to-android-studio) (or otherwise build / upload debug APK version so other people can help test it without having to build their own)

---

Back to the PR, I've noticed the colors are broken in dark mode (white box, too dark dropdownlist triangle) - PR version on the right:

small_Screenshot_20250112_223047_SCEE Dev small_Screenshot_20250112_223013_StreetComplete Dev

Other than that, it seems to work as advertised: on that specific example node, it offers (and applies when selected) `name=B72/Kirchweg` tag


It would also be interesting how it handles if more than one language (e.g. currently it says only `de`) is offered (do you know of such location)?

mnalis:
And also in Light theme,  it seems that the flashing cursor / text / spacing / area for input text got bigger, and text is now centered?

small_Screenshot_20250112_224908_SCEE Dev small_Screenshot_20250112_224840_StreetComplete Dev

mnalis:
> It would also be interesting how it handles if more than one language (e.g. currently it says only de) is offered (do you know of such location)?

https://overpass-turbo.eu/s/1X3K might be helpful in finding them. My quick test seems to indicate it works fine there too (e.g. https://www.openstreetmap.org/node/3839537126):


https://github.com/user-attachments/assets/d1453eea-55d8-4163-82aa-423a83d6cc22

kmpoppe:
> And also in Light theme, it seems that the flashing cursor / text / spacing / area for input text got bigger, and text is now centered?\n\n

Yes, the `quest_localizedname_row` layout is a copy of ...roadname... which looks like that. I think it looks good, it might just look weird when comparing them side by side?

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/BusStopNameSuggestionsSource.kt
```diff
@@ -0,0 +1,81 @@
...
+                    (tags["highway"].equals("bus_stop") && !tags["public_transport"].equals("stop_position"))
+                    || (tags["public_transport"].equals("platform") && tags["bus"].equals("yes"))
+                    || tags["railway"].equals("halt")
+                    || tags["railway"].equals("station")
+                    || tags["railway"].equals("tram_stop")
+                )
+            )
+        )
+    }
+
```
FloEdelmann, line 80:
```suggestion
    private fun busStopFilter(tags: Map<String, String>) =
        tags.containsKey("name")
        && (
            (tags["highway"] == "bus_stop" && tags["public_transport"] != "stop_position")
            || (tags["public_transport"] == "platform" && tags["bus"] == "yes")
            || tags["railway"] == "halt"
            || tags["railway"] == "station"
            || tags["railway"] == "tram_stop"
        )
```

mnalis:
> Removed the forced colours, lo and behold: Dark and Light views now work as expected

Hooray!

> >    And also in Light theme, it seems that the flashing cursor / text / spacing / area for input text got bigger, and text is now centered?\n\n

> Yes, the quest_localizedname_row layout is a copy of ...roadname... which looks like that. I think it looks good, it might just look weird when comparing them side by side?

It was also a loss of general decoration around text that made the empty margins look excessive. With your newest debug APK, it got that underline, so it looks better already.

small_Screenshot_20250113_134235_StreetComplete Dev small_Screenshot_20250113_134042_StreetComplete Dev

I would probably look even nicer (and waste less screen real estate) if that empty space between `hr` and `en` lines was smaller, but that is just a minor cosmetic detail. (big spacing look OK on that "blue table" as it simulates how the sign might look in reality, but looks strange for regular fields - like I've enabled some fat-finger accessibility feature or something)

But not a showstopper for me (unless it is easy to fix)

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/BusStopNameSuggestionsSource.kt
westnordost, line 1: I see lots of duplication in this file and references to road names. Some stuff, like the things with the added buffer, are hardly relevant for bus stop names, as bus stop platforms are not that as long as roads, are they? (Might as well just look a the distance of the center points of all the platforms.) Also, why input a list of points? Doesn't make much sense for bus stop platforms.

So, this class can be simplified A LOT ....

... OR, the RoadNameSuggestionsSource is generalized (more) to a `NameSuggestionsSource` that accepts a filter function and looks at min distances to *any* geometry returned by the filter, returns their names.

Either way would be fine for me, but I slightly prefer the generalization because this will lead to less code in total, I believe.

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/AddBusStopNameForm.kt
```diff
@@ -15,11 +19,25 @@ class AddBusStopNameForm : AAddLocalizedNameForm<BusStopNameAnswer>() {
...
+    override fun getLocalizedNameSuggestions(): List<List<LocalizedName>> {
+        val polyline = when (val geom = geometry) {
+            is ElementPolylinesGeometry -> geom.polylines.first()
+            is ElementPolygonsGeometry -> geom.polygons.first()
+            is ElementPointGeometry -> listOf(geom.center)
+        }
+        return busStopNameSuggestionsSource.getNames(
+            listOf(polyline.first(), polyline.last()),
+            MAX_DIST_FOR_BUS_STOP_NAME_SUGGESTION
+        )
```
westnordost, line 39: why?

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/BusStopNameSuggestionsSource.kt
```diff
@@ -0,0 +1,75 @@
...
+    private fun busStopFilter(tags: Map<String, String>) =
+        tags.containsKey("name")
+        && (
+            (tags["highway"] == "bus_stop" && tags["public_transport"] != "stop_position")
+            || (tags["public_transport"] == "platform" && tags["bus"] == "yes")
+            || tags["railway"] == "halt"
+            || tags["railway"] == "station"
+            || tags["railway"] == "tram_stop"
+        )
```
westnordost, line 74: why not use the element filter syntax?

app/src/main/res/layout/quest_localizedname_row.xml
westnordost, line 1: What's wrong with `R.layout.row_localizedname`, i.e. the default?

westnordost (changes requested):

app/src/main/res/layout/quest_localizedname_row.xml
kmpoppe: I was somehow under the impression I had to use a `quest_`-layout to make this work and it looked like it wasn't working with the standard before - but I might have done something completely wrong here.

Will remove the special layout.

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/BusStopNameSuggestionsSource.kt
kmpoppe: Reworked in [86bd839](https://github.com/streetcomplete/StreetComplete/pull/6097/commits/86bd839b8d5aca7e0909fabdd73b9f8b34d9c6a6)

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/AddBusStopNameForm.kt
kmpoppe: removed in [86bd839](https://github.com/streetcomplete/StreetComplete/pull/6097/commits/86bd839b8d5aca7e0909fabdd73b9f8b34d9c6a6)

kmpoppe:
Reworked for a central NameSuggestionSource, @mnalis could you please find another example to test with more than one name?

westnordost:
> kmpoppe requested a review from westnordost

Could you first review the code yourself? Glancing over it, I am sure you will still find something.

westnordost:
Also, generally, I'd prefer if the bus stop name and road name quests would use the name suggestions source directly rather than having a class in-between (unless there is a good reason for it - at a glance, I don't see a good reason).

kmpoppe:
> Also, generally, I'd prefer if the bus stop name and road name quests would use the name suggestions source directly rather than having a class in-between (unless there is a good reason for it - at a glance, I don't see a good reason).

I was debating that with myself and thought that specific classes would make more sense but alas, if you don't, I'll remove the specific classes and just work with NameSuggestion

mnalis:
> Reworked for a central NameSuggestionSource, @mnalis could you please find another example to test with more than one name?

Sure, e.g.  https://www.openstreetmap.org/node/9207779395 or https://www.openstreetmap.org/node/8208818583 

(Found by some manual looking at https://overpass-turbo.eu/s/1XnZ)

westnordost (approved):
There were a few minor things, but since I had to dig into the code anyway and you certainly want to get on with the fixme name stuff to which this PR is merely the prerequisite, I just did those myself.

Other than small stuff like adding some explanatory comments and doc comments, the more important things I changed were:
- added `fun ElementGeometry.distance(point: LatLon): Double` instead of having to do this manually in the `getNames` function
- no linking to some static val of `AddRoadNameForm` from some other class
- properly typed `filter` parameter in `getNames`

kmpoppe:
Thanks for the edits @westnordost, the `ElementGeometry.distance(points)` function wouldn't have occured to me.

app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_name/AddBusStopNameForm.kt
```diff
@@ -20,6 +25,27 @@ class AddBusStopNameForm : AAddLocalizedNameForm<BusStopNameAnswer>() {
...
 
+    // this filter needs to be kept somewhat in sync with the filter in AddBusStopName
+    private val busStopsWithNamesFilter = """
+        nodes, ways, relations with
+        (
+          public_transport = platform and bus = yes
+          or (highway = bus_stop and public_transport != stop_position)
+          or railway = halt
+          or railway = station
+          or railway = tram_stop
```
FloEdelmann, line 36:
```suggestion
          public_transport = platform and bus = yes
          or highway = bus_stop and public_transport != stop_position
          or railway ~= halt|station|tram_stop
```

(same in `AddBusStopName`)

westnordost: Fair enough. It's just `~`, though.

PR #6100, opened 2025-01-15
Allow individual names for brand features; don't show "has no name" option for features with fixed name

### LINKED ISSUE #5985: individual name for places of a brand
**How to Reproduce**
Try to add a place in the places overlay of a brand with an individual name for example:
```
shop=supermarket
brand=EDEKA
brand:wikidata=Q701755
name=EDEKA Schreiber
```
At the moment you cannot set an individual name for a place of a brand in the places overlay.

---

Or try to add the brand tag in the places overlay to a place that already has an individual name:
```
shop=supermarket
name=EDEKA Schreiber
``` 
The name tag is currently overwritten and the following tagging is obtained:
```
shop=supermarket
brand=EDEKA
brand:wikidata=Q701755
name=EDEKA
```
**Expected Behavior**

1. For places of a brand there should still be field to enter an individual name or at least an option to open such a field. 
2. The name tag of a place shouldn't be overridden when a brand is set. It should also be displayed and editable.

**Versions affected**

59.2

**Issues that may have a certain relationship**
#4688
#5313
#5940

**Wiki**
https://wiki.openstreetmap.org/wiki/Key:brand#Examples

### LINKED ISSUE #6086: don't show the "There is no name sign" option in the places overlay when there is no input field for the name
Right now there is all ways the "There is no name sign" option in the places overlay.
- Also, if the user is not supposed to enter a name.
- Also, if there is already a name that is not displayed.

I think the "There is no name sign" option should only be shown when also the input field for the name is shown.

westnordost:
respect preserveTags property in osmfeatures (fixes #6086, fixes #5985)


mnalis:
Um, what problems  is it  supposed to fix exactly? I've tried compiling that branch at https://github.com/streetcomplete/StreetComplete/pull/6100/commits/bbf233dc5579d1c8c724f1137edcb1a6528b5ff6 (at https://github.com/mnalis/StreetComplete/actions/runs/12810652342) and:

- opening a `club=yes` still does not show the name nor allows one to enter it? (i.e. https://github.com/streetcomplete/StreetComplete/issues/6086#issuecomment-2581210599)
  e.g. https://www.openstreetmap.org/way/278910702 or https://www.openstreetmap.org/way/1143181142:
  

  https://github.com/user-attachments/assets/8dcc8e49-f8d1-4244-b6d7-1dfffc541b48

  OK, _"There is no name sign"_ option is gone, but is that the preferred solution? I'd expect to be able to enter a name for the generic `club=yes` instead?

mnalis:
- for https://github.com/streetcomplete/StreetComplete/issues/5985, it still seems to overwrite the name and does not allow to user to override it? (seemingly regardless of `preserveTags: ^name` which [EDEKA aktiv markt](https://nsi.guide/index.html?t=brands&k=shop&v=supermarket&tt=edeka#edekaaktivmarkt-0a839e) has, and  [EDEKA xpress](https://nsi.guide/index.html?t=brands&k=shop&v=supermarket&tt=edeka#edekaxpress-0a839e) doesn't)


https://github.com/user-attachments/assets/67cd3799-38c5-43ee-bd44-a9ebe50f65e5


Or am I misunderstanding what is supposed to happen here? :man_shrugging:

westnordost:
For it to work, a new version of the nsi-id-presets.json must be published which includes the `preserveTags` property in the data.

westnordost:
I just updated from the current NSI that includes the `preserveTags`. If you like, you could test again, @mnalis 
(But I already tested the implementation, so I don't think it is required.)

PR #6109, opened 2025-01-26
Encode git revision in debug .apk

mnalis:
When using GitHub actions to build debug `.apk`, encode actual git revision built into its `versionName`. 

It helps using About window to identify exact git revision being tested, among other things. Especially helpful when testing several different branches at the same time.

tested (and example debug `.apk`) at https://github.com/mnalis/StreetComplete/actions/runs/12996593370

### DIFF
--- a/.github/workflows/build-debug-apk.yml
+++ b/.github/workflows/build-debug-apk.yml
@@ -21,18 +21,21 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
 
+      - name: Set env
+        run: echo "COMMIT_SHA=$(git log -n 1 --format='%h')" >> $GITHUB_ENV
+
       - name: Encode git revision
-        run: sed -i -e "s/versionName = \"\([^ ]*\).*\"/versionName = \"\1 (git $(git log -n 1 --format='%h'))\"/"  app/build.gradle.kts
+        run: sed -i -e "s/versionName = \"\([^ ]*\).*\"/versionName = \"\1 (git $COMMIT_SHA)\"/"  app/build.gradle.kts
 
       - name: Build with Gradle
         run: ./gradlew assembleDebug
 
       - name: Rename APK
-        run: mv app/build/outputs/apk/debug/app-debug.apk app/build/outputs/apk/debug/StreetComplete-debug-$(git log -n 1 --format='%h').apk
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
Can/should we change archive name `debug-apk.zip` too? Line 36

mnalis:
> Can/should we change archive name `debug-apk.zip` too? Line 36

@HolgerJeromin Might be nice, but it does not seem easy (and  I can't tell if it is possible at all[^1]). 

At least the the `.apk` inside that `debug-apk.zip` is already (re-)named in detail,  so it is not a huge deal IMHO (one just needs to keep raw `.apk` files instead of their parent `.zip` files)

[^1]: Someone would need to check `actions/upload-artifact@v4` code to see if it is possible and how. Quick test (https://github.com/mnalis/StreetComplete/actions/runs/12977633973) seems to indicate that its `name` argument is a fixed string only, i.e. does not expand shell commands like `run` lines do, so the approach above cannot be used.

FloEdelmann:
It should work with [expressions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions). Also, to deduplicate the repeated hardly-readable git command, you could add another step that [writes it to an env variable](https://stackoverflow.com/a/57969570/451391).

So stitched together:

```yaml
      - name: Set env
        run: echo "COMMIT_SHA=$(git log -n 1 --format='%h')" >> $GITHUB_ENV ##### set it here

      - name: Encode git revision
        run: sed -i -e "s/versionName = \"\([^ ]*\).*\"/versionName = \"\1 (git $COMMIT_SHA)\"/"  app/build.gradle.kts ##### use it here

      - name: Build with Gradle
        run: ./gradlew assembleDebug

      - name: Rename APK
        run: mv app/build/outputs/apk/debug/app-debug.apk app/build/outputs/apk/debug/StreetComplete-debug-$COMMIT_SHA.apk ##### use it here

      - name: Archive APK
        uses: actions/upload-artifact@v4
        with:
          name: debug-apk-${{ env.COMMIT_SHA }} ##### use it here with expression syntax
          path: app/build/outputs/apk/debug/*.apk
          retention-days: 30
```

mnalis:
> It should work with [expressions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions).

Thank @FloEdelmann, it is indeed nicer and works with `.zip` filename too. Implemented and tested (https://github.com/mnalis/StreetComplete/actions/runs/12996593370).

PR #6111, opened 2025-01-27
synchronize more on quest visibility updates 

### LINKED ISSUE #5545: House number quest is asked again immediately after solving it
I hope what I saw today is not another case of #4258, but I was asked to solve a house number quest several times again after solving it during a quest update download. See screen shot (note it's asking for the house number of the apartment building with house number 4, i.e. the number is already known). Restarting the app made the quest go away. Maybe #5473 was not a complete solution?

Edit: I noticed in the history of the object here https://www.openstreetmap.org/way/1135323961/history that each time I answered the house number quest again, a new edit was made to the object. The house number quest was preceded with an "Is this building still under construction?" quest, which I answered with Yes. That answer apparently resulted in the activation of the house number quest, but not in an edit to the object (it's still under construction on the map).

Screenshot_20240321-161338

**Versions affected**
Android 10, StreetComplete 57.1

westnordost:
fixes #5545

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/osm/osmquests/OsmQuestController.kt
@@ -85,6 +85,7 @@ class OsmQuestController internal constructor(
             }
 
             val obsoleteQuestKeys: List<OsmQuestKey>
+            val visibleQuests: Collection<OsmQuest>
             synchronized(this) {
                 val previousQuests = db.getAllForElements(updated.map { it.key })
                 // quests that refer to elements that have been deleted shall be deleted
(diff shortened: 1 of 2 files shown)


Helium314:
Looks good!
Though now I'm pondering whether it would also be a problem if `onUpdated` overtakes `onReplacedForBbox` while quests for the bbox are being created. Anyway, this is not what happened in the log provided in https://github.com/streetcomplete/StreetComplete/issues/5545#issuecomment-2504715815

Helium314:
`OsmQuestController.mapDataSourceListener`
If we assume persisting elements is done, and `onReplacedForBBox` is called, then could it be bad if `onUpdated` if called while `onReplacedForBBox` is in `createQuestsForBBox`?
But maybe we should just work on problems that actually occurred, and think about this only if the quest-asked-again issue comes up again.

westnordost:
I think this is covered by the fact that the MapDataWithElementsSource will replay any updates done while replacing the bbox.

PR #6114, opened 2025-01-29
Refactor visible quests

westnordost:
Pull out the managing of `NoteQuestsHiddenDao` and `OsmQuestsHiddenDao` from `OsmNoteQuestController` and `OsmQuestController`, respectively and instead have it both managed by the new `QuestsHiddenController`, which also manages caching of hidden quests for fast access by the `VisibleQuestsSource`. This makes the interfaces a little cleaner, IMO.

[See /res/documentation/overview_data_flow.svg](https://github.com/streetcomplete/StreetComplete/pull/6114/files#diff-6dfed29937ad9dda100398bade0a5d6fc794589b7c070c9bf33f2df9fa0d6e75)


app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
```diff
@@ -91,46 +114,53 @@ class VisibleQuestsSource(
...
         visibleQuestTypeSource.isVisible(questType) &&
         selectedOverlaySource.selectedOverlay?.let { questType.name !in it.hidesQuestTypes } ?: true
 
-    private fun isVisible(quest: Quest): Boolean =
-        isVisibleInTeamMode(quest) && isVisible(quest.type)
+    private fun isVisible(questKey: QuestKey): Boolean =
+        questsHiddenSource.get(questKey) == null
 
     private fun isVisibleInTeamMode(quest: Quest): Boolean =
         teamModeQuestFilter.isVisible(quest)
```
FloEdelmann, line 166: Previously, we checked `teamModeQuestFilter.isEnabled` before calling this function. Is that faster and should it be added back, or does it not matter?

```suggestion
        teamModeQuestFilter.isEnabled && teamModeQuestFilter.isVisible(quest)
```

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
```diff
@@ -47,7 +47,6 @@ import de.westnordost.streetcomplete.view.confirmIsSurvey
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
-import kotlinx.serialization.encodeToString
```
FloEdelmann, line 50: Why is this import removed when nothing about `encodeToString` in this file changed? Does it still work correctly?

app/src/main/java/de/westnordost/streetcomplete/screens/settings/SettingsViewModel.kt
```diff
@@ -31,7 +28,7 @@ import kotlinx.coroutines.flow.StateFlow
 abstract class SettingsViewModel : ViewModel() {
     abstract val selectableLanguageCodes: StateFlow<List<String>?>
     abstract val selectedQuestPresetName: StateFlow<String?>
-    abstract val hiddenQuestCount: StateFlow<Long>
+    abstract val hiddenQuestCount: StateFlow<Int>
```
FloEdelmann, line 31: What if I want to hide more than 2,147,483,647 quests now? :open_mouth: :wink:

res/documentation/overview_data_flow.svg
FloEdelmann, line 1: The SVG previously had a white background, which makes it easier to read in apps with a dark background or in viewers (like GitHub) that display transparency as a checkerboard background.

FloEdelmann (commented):
I only had a cursory look over the code and data flow diagram, and only have a few nitpicks/questions. I haven't checked out the code locally or tried to run it though.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/SettingsViewModel.kt
Helium314: Don't worry, if you hide 10 quests per second we still have a few years to fix it

app/src/main/java/de/westnordost/streetcomplete/data/quest/VisibleQuestsSource.kt
westnordost: It doesn't matter, the TeamModeQuestFilter checks itself it is enabled in the first line of the `isVisible` function.

app/src/main/java/de/westnordost/streetcomplete/quests/AbstractOsmQuestForm.kt
westnordost: it was an unused import

Issue #6113, opened 2025-01-28
Quest "Does this place have a baby changing table?" almost never asks for cafés and restaurants because toilets=yes is a precondition. Proposal to change this precondition


agent-redd:
**Use case**
When I enabled the quest "Does this place have a baby changing table?"  to survey diaper changing tables, I recognized the quest appears seldomly and almost never for cafés and restaurants, which often have changing tables.

Currently there is a precondition toilets=yes for the shops and restaurants:
```kotlin
nodes, ways with
        (
          (
            (amenity ~ restaurant|cafe|fuel|fast_food or shop ~ mall|department_store)
            and toilets = yes
          )
          or amenity = toilets
        )
        and !diaper and !changing_table
```
from https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt

On the other side, the quest "Is there a restroom here?" does not ask for restaurants or cafés, as they are not considered public enough:
```kotlin
    // only for malls, big stores and rest areas because users should not need to go inside a non-public
    // place to solve the quest. (Considering malls and department stores public enough)
    override val elementFilter = """
        nodes, ways with
        (
          shop ~ mall|department_store
          or highway ~ services|rest_area
          or tourism ~ camp_site|caravan_site
        )
        and !toilets
```
from https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/quests/toilet_availability/AddToiletAvailability.kt

Therefore the precondition toilets=yes can never be fullfilled by StreetComplete itself.

**Proposed Solution**
I am open for discussion. My simple proposed solution is removing the precondition, as this quest is deactivated by default anyway:
```kotlin
nodes, ways with
        (
          // ask for POIs like toilets, restaurants, which often have diaper changing tables
          amenity ~ toilets|restaurant|cafe|fuel|fast_food 
          or shop ~ mall|department_store)
        )
        and !diaper and !changing_table
```

mnalis:
> Therefore the precondition toilets=yes can never be fullfilled by StreetComplete itself.

Yeah... Although, you could use Things Overlay to place `amenity=toilets` node, but that is likely overkill.

>  I am open for discussion. My simple proposed solution is removing the precondition, as this quest is deactivated by default anyway:

[...]

PR #6115, opened 2025-01-31
Proposal to change elementFilter for AddBabyChangingTable.kt

agent-redd:
Hello everybody,
this is my first Pull Request, please give me Feedback if something is wrong.

I created an Issue #6113 earlier to discuss the changes. I am still waiting for more feedback on it. 
This Pull Request should resolve the problem, that the precondition toilets=yes for the BabyChangingTable Quest could not be fulfilled by StreetComplete itself. 

Edit: see comments, the filter was now changed several times

Fixes https://github.com/streetcomplete/StreetComplete/issues/6113

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
@@ -14,7 +14,7 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
         nodes, ways with
         (
           amenity ~ restaurant|cafe|toilets
-          or (amenity = fuel or (shop and shop !~ no|vacant)) and toilets = yes
+          or amenity = fuel or shop ~ mall|department_store and toilets = yes
           or amenity = fast_food and ~toilets|indoor_seating|outdoor_seating = yes
         )
         and !diaper and !changing_table
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
@@ -14,7 +14,7 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
         nodes, ways with
         (
           amenity ~ restaurant|cafe|toilets
-          or amenity = fuel or shop ~ mall|department_store and toilets = yes
+          or (amenity = fuel or shop ~ mall|department_store) and toilets = yes
           or amenity = fast_food and (toilets = yes or indoor_seating = yes or outdoor_seating = yes)
         )
         and !diaper and !changing_table
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
@@ -13,11 +13,14 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
     override val elementFilter = """
         nodes, ways with
         (
-          (
-            (amenity ~ restaurant|cafe|fuel|fast_food or shop ~ mall|department_store)
-            and toilets = yes
-          )
-          or amenity = toilets
+          // ask for all restaurants and toilets
+          amenity ~ restaurant|toilets
+          or
+          // ask for fuel stations or shops which have toilets
+          ((amenity ~ fuel or (shop and shop !~ no|vacant)) and toilets=yes)
+          or
+          //ask for cafés and fast_food which have toilets or seating
+          (amenity ~ cafe|fast_food and (toilets=yes or indoor_seating=yes or outdoor_seating=yes))
         )
         and !diaper and !changing_table
     """
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
@@ -13,9 +13,9 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
     override val elementFilter = """
         nodes, ways with
         (
-          amenity ~ restaurant|toilets
+          amenity ~ restaurant|cafe|toilets
           or (amenity = fuel or (shop and shop !~ no|vacant)) and toilets = yes
-          or amenity ~ cafe|fast_food and ~toilets|indoor_seating|outdoor_seating = yes
+          or amenity = fast_food and ~toilets|indoor_seating|outdoor_seating = yes
         )
         and !diaper and !changing_table
     """
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
@@ -14,7 +14,7 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
         nodes, ways with
         (
           amenity ~ restaurant|toilets
-          or (amenity ~ fuel or (shop and shop !~ no|vacant)) and toilets = yes
+          or (amenity = fuel or (shop and shop !~ no|vacant)) and toilets = yes
           or amenity ~ cafe|fast_food and ~toilets|indoor_seating|outdoor_seating = yes
         )
         and !diaper and !changing_table
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
@@ -13,14 +13,9 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
     override val elementFilter = """
         nodes, ways with
         (
-          // ask for all restaurants and toilets
           amenity ~ restaurant|toilets
-          or
-          // ask for fuel stations or shops which have toilets
-          ((amenity ~ fuel or (shop and shop !~ no|vacant)) and toilets=yes)
-          or
-          //ask for cafés and fast_food which have toilets or seating
-          (amenity ~ cafe|fast_food and (toilets=yes or indoor_seating=yes or outdoor_seating=yes))
+          or (amenity ~ fuel or (shop and shop !~ no|vacant)) and toilets = yes
+          or amenity ~ cafe|fast_food and ~toilets|indoor_seating|outdoor_seating = yes
         )
         and !diaper and !changing_table
     """


app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
```diff
@@ -13,11 +13,14 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
...
-          )
-          or amenity = toilets
+          // ask for all restaurants and toilets
+          amenity ~ restaurant|toilets
+          or
+          // ask for fuel stations or shops which have toilets
+          ((amenity ~ fuel or (shop and shop !~ no|vacant)) and toilets=yes)
+          or
+          //ask for cafés and fast_food which have toilets or seating
+          (amenity ~ cafe|fast_food and (toilets=yes or indoor_seating=yes or outdoor_seating=yes))
```
FloEdelmann, line 23: Comments are not allowed in the element filter syntax. Also, some parentheses can be avoided because `and` has higher precedence than `or`. And the last condition can be simplified by using a regex in the key. In general, see the cheatsheet in [ElementFilterExpression.kt](https://github.com/streetcomplete/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/ElementFilterExpression.kt). All these changes don't change your intended logic:

```suggestion
          amenity ~ restaurant|toilets
          or (amenity ~ fuel or (shop and shop !~ no|vacant)) and toilets = yes
          or amenity ~ cafe|fast_food and ~toilets|indoor_seating|outdoor_seating = yes
```

agent-redd: Thanks for your quick feedback! Is it okay to leave the comments away? Or shall I put the comments above the filter? I even saw, that sometimes the Github Issue URL is put in as comment..

FloEdelmann: Adding comments that explain what the element filter does is IMO not necessary, since the element filter syntax is quite readable. Adding the issue URL as a comment is sometimes helpful to understand the reasoning and/or link to the discussion that lead to a specific piece of code. That could be helpful here, but IMO it's not necessary here either, since `git blame` can always lead you to the commit that last changed a line of code, and thus lead you to the PR.

mnalis (commented):
Few points (some  mentioned in the issue but this implementation is slightly different):
- `amenity=cafe` is defined as _"generally informal place with sit-down facilities selling beverages and light meals and/or snacks"_, so there should be not much point in checking if it has seating places (i.e. it would mostly be asked on all cafes; so it might be clearer if it joined `restaurant`  as the amenity where quest is unconditionally asked on, as the result would be mostly the same).
- the PR would of course ask the quest on much more places (which is kind of whole point I guess). Whether that would be too [spammy](https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md) or not might be a consideration.

(also I've edited original post with "magic text" so issue would close automatically if this PR is merged)

mnalis:
> Warning: This code is not yet linted or tested, yet! (I currently do not have an Android Studio setup yet.)

On that subject - if you prefer, you can test if your PR compiles (and how does resulting `.apk` work) by few simple clicks in GitHub actions in your fork (without installing Android Studio) - see https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#alternative-to-android-studio for quick instructions. 

Such debug `.apk`s can be installed alongside your regular StreetComplete, without affecting it.

agent-redd:
>     * `amenity=cafe` is defined as _"generally informal place with sit-down facilities selling beverages and light meals and/or snacks"_, so there should be not much point in checking if it has seating places (i.e. it would mostly be asked on all cafes; so it might be clearer if it joined `restaurant`  as the amenity where quest is unconditionally asked on, as the result would be mostly the same).

Thanks. I moved the cafe into first line without preconditions to join the restaurants.

>     * the PR would of course ask the quest on much more places (which is kind of whole point I guess). Whether that would still fullfill [quest guidelines](https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md) (_"💤 No spam"_ and _"🕵️ Publicly accessible on foot"_) might be a consideration.

Thanks for that feedback. I reverted the shops to mall and department_store to be less spammy. My biggest intention is to ask for restaruants (also fast_food) and cafés.

Regarding the "🕵️ Publicly accessible on foot". This quest will probably never fulfill this requirement and that is the reason why this quest is deactivated by default. Perhaps there needs to be bigger discussion how indoor survey quests are handled in StreetComplete. This would also concern other deactivated quests like wheelchair_access, toilets_fee, air_conditioning, internet_access, accepts_card and accepts_cash.

agent-redd:
Thanks to @mnalis for the hint of Github Actions to build a debug APK. 

I built this change with Github Actions to test the changes:
GitHub Run: https://github.com/agent-redd/StreetComplete/actions/runs/13088445391
Debug APK: https://github.com/agent-redd/StreetComplete/actions/runs/13088445391/artifacts/2521143573

I updated also the initial comment of this Pull Request to match the changes

mnalis:
> Regarding the "🕵️ Publicly accessible on foot". This quest will probably never fulfill this requirement and that is the reason why this quest is deactivated by default. 

Yeah, the guidelines have _some_ wiggle room; and for deactivated-by-default quests it is often because one has to enter the place. As you note, this one is already deactivated by default, so that looks fine to me.

> https://github.com/agent-redd/StreetComplete/actions/runs/13088445391

So, have you installed that debug `.apk` and tried how it works? Does the selection of the quests work as you expected? Does it look too spammy or still too restricted to you? Does it select strange things, or fails to select things you think it should select?

Because unless I'm wrong that compiled 8e6fea3 is just vanilla StreetComplete `master`, and not a branch with your changes (`patch-1`). I.e. you've probably missed this part of the instructions:

> make sure you select the correct branch with your changes when clicking `Run workflow`,

agent-redd:
> So, have you installed that debug `.apk` and tried how it works? Does the selection of the quests work as you expected? Does it look too spammy or still too restricted to you? Does it select strange things, or fails to select things you think it should select?
> 
> Because unless I'm wrong that compiled [8e6fea3](https://github.com/streetcomplete/StreetComplete/commit/8e6fea3b123270b40d7939fefbfc234af570f147) is just vanilla StreetComplete `master`, and not a branch with your changes (`patch-1`). I.e. you've probably missed this part of the instructions:
> 
> > make sure you select the correct branch with your changes when clicking `Run workflow`,

thanks for pointing that out. I ran the workflow from branch `patch-1`, but then the workflow ran again from `master` unintendenly.

Here is the correct link from first workflow from branch `patch-1`:

https://github.com/agent-redd/StreetComplete/actions/runs/13088443571

I will now download again and test it thoroughly.

agent-redd:
So unfortunately with the build from my branch the download of data fails. I assume a problem in the filter syntax:

Screenshot_2025-02-01-16-22-33-846_de westnordost streetcomplete debug

agent-redd:
I dont understand the `~toilet` yet in line 18:

```or amenity = fast_food and ~toilets|indoor_seating|outdoor_seating = yes```

Could this be a syntax error?

**Comment by FloEdelmann** [comment]
`–toilets|indoor_seating|outdoor_seating ~ yes` would mean any tag whose key matches the regex `toilets|indoor_seating|outdoor_seating` and whose value matches the regex `yes` (see the [cheatsheet](https://github.com/streetcomplete/StreetComplete/blob/8e6fea3b123270b40d7939fefbfc234af570f147/app/src/main/java/de/westnordost/streetcomplete/data/elementfilter/ElementFilterExpression.kt#L43)).

But putting an `=` after a regex key might actually be a syntax error, at least it is not documented. So you might have to change:

```diff
- –toilets|indoor_seating|outdoor_seating = yes
+ –toilets|indoor_seating|outdoor_seating ~ yes
```

mnalis:
> But putting an = after a regex key might actually be a syntax error

Probably. Or just expand it, as not much is gained by that key regex and simple value... e.g.

```diff
- or amenity = fast_food and ~toilets|indoor_seating|outdoor_seating = yes
+ or amenity = fast_food and (toilets = yes or indoor_seating = yes or outdoor_seating = yes)
```

agent-redd:
> Also, when things don't work, sometimes looking at `About` / `Show logs` might be helpful (not always though).

Indeed, it shows some hintful message:

```
2025-02-01T16:25:34.172: E [Download] Unable to download de.westnordost.streetcomplete.data.elementfilter.ParseException: At position 237: Unexpected operator '=': The key prefix operator '~' must be used together with the binary operator '~' or '!~'
```

agent-redd:
> > But putting an = after a regex key might actually be a syntax error
> 
> Probably. Or just expand it, as not much is gained by that key regex and simple value... e.g.
> 
> ```diff
> - or amenity = fast_food and ~toilets|indoor_seating|outdoor_seating = yes
> + or amenity = fast_food and (toilets = yes or indoor_seating = yes or outdoor_seating = yes)
> ```

Thanks. I took over your suggestion in a testing branch, rebuilt a dev version and testet my use cases:

1. all restaurants and cafes are now asked
2. fast_food is not usked until i activate the seating quest and add indoor/outdoor/both seating

I assume this Pull Request is now technically ready to merge.

app/src/main/java/de/westnordost/streetcomplete/quests/baby_changing_table/AddBabyChangingTable.kt
```diff
@@ -13,11 +13,9 @@ class AddBabyChangingTable : OsmFilterQuestType<Boolean>() {
     override val elementFilter = """
         nodes, ways with
         (
-          (
-            (amenity ~ restaurant|cafe|fuel|fast_food or shop ~ mall|department_store)
-            and toilets = yes
-          )
-          or amenity = toilets
+          amenity ~ restaurant|cafe|toilets
+          or amenity = fuel or shop ~ mall|department_store and toilets = yes
```
mnalis, line 17:
```suggestion
          or (amenity = fuel or shop ~ mall|department_store) and toilets = yes
```

If the intention is to ask only on fuel stations that have toilets, then there are missing parenthesis (as suggested here).

If the intention is instead to ask on all fuel stations regardless of presence of toilets, then it should move to that regexp above (with restaurants) to be more readable.

agent-redd: Thanks for pointing that our! It was not my intention to ask for all fuel stations, therefore i applied you code proposal.

My main intention is about restaurants, cafés and fast_food.

I am rebuilting a dev version right now and will test this again.

agent-redd:
I applied the proposal from @mnalis to revert the unintended change regarding fuel stations.

I then rebuilt a dev version with Github Actions: https://github.com/agent-redd/StreetComplete/actions/runs/13097378940

And then tested that its only asking for fuel stations which have toilets=yes.

Thanks everybody for the helpful feedback! I hope that this Pull Request is now ready to merge.

westnordost:
Without having followed the massive discussion here, I think @agent-redd 's proposal makes sense. The toilets quest is not asked for every restaurant etc. because the assumption is that many/most *do* have toilets, so asking for it specifically would be spammy. The baby changing table quest is already disabled by default (because the mapper would need to go inside and check/ask), I don't think it is necessary to require `toilets=yes` being set, as that is what can already be assumed.

Regarding the PR in its current state:
1. What if if specifically `toilets=no` is set? Should it still be asked? I'd say that "no"?
2. The general assumption that there is a toilet is I'd say even more true for malls and department stores, so IMO it's fine to ask for baby changing tables always, like for restaurant etc.
3. TBH I tend towards also asking for fast food.

I understand why in the current state of the PR, 2 and 3 were made the way they are now: Because the toilets quests **is** asked for malls and department stores plus there's also a (deactivated by default) seating quest. But IMO that's not necessary. It is enough to check that they are **not** tagged like `toilets=no` (, `indoor_seating=no`, ...)

agent-redd:
Thank you @westnordost for your helpful fedback. Your suggestion to alwys ask unless toilets=no would simplify the filter and be more intuitiv. The seating precondition could be eliminated again.

I will rework and test my Pull Request and give an update then.

agent-redd:
I now simplified the Pull Request to the logic as @westnordost suggested. Built a dev apk with Guthub Actions and tested it against some `amenity=fast_food` without `changing_table=*`  POIs

B) `toilets=no`: https://www.openstreetmap.org/node/43804961
==> Quest does not appear
B) `toilets=yes`:https://www.openstreetmap.org/node/2866502289#map=19/48.789136/9.195583&layers=N
==> Quest appears
C) without `toilet=*`: https://www.openstreetmap.org/node/12165822426
==> Quest appears

Here is the dev build if anybody wants to do further tests: 
https://github.com/agent-redd/StreetComplete/actions/runs/13102320869

I guess if this Pull Request gets merged the commit history should be crunched to hide the back and forth. Not sure if Github can do this, or if I should prepare this by creating a new cleaner branch?

PR #6119, opened 2025-02-03
Cleanup `Database` interface

FloEdelmann:
In preparation for #5417, which I started looking into.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
@@ -65,7 +65,6 @@ class AndroidDatabase(private val db: SQLiteDatabase) : Database {
         having: String?,
         orderBy: String?,
         limit: Int?,
-        distinct: Boolean,
         transform: (CursorPosition) -> T
     ): List<T> {
         val strArgs = args?.primitivesArrayToStringArray()
--- a/app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
@@ -16,7 +16,6 @@ import androidx.core.database.getDoubleOrNull
 import androidx.core.database.getFloatOrNull
 import androidx.core.database.getIntOrNull
 import androidx.core.database.getLongOrNull
-import androidx.core.database.getShortOrNull
 import androidx.core.database.getStringOrNull
 import androidx.core.database.sqlite.transaction
 import de.westnordost.streetcomplete.data.ConflictAlgorithm.ABORT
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
```diff
@@ -65,12 +64,11 @@ class AndroidDatabase(private val db: SQLiteDatabase) : Database {
         groupBy: String?,
         having: String?,
         orderBy: String?,
-        limit: String?,
-        distinct: Boolean,
+        limit: Int?,
```
westnordost, line 67: why no distinct?

(yeah, I noticed that the variable is not used, but that just looks like a bug)

westnordost (commented):
Oh, you are? That's cool!

By the way, there is _probably_ little reason to keep the `Database` interface when the database dealing stuff has been migrated to that multiplatform-enabled thing. (We'll see)

app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
FloEdelmann: > the variable is not used

Yes, that's the reason. And since it is also never called with `distinct = true`, I figured that the parameter could go.

app/src/main/java/de/westnordost/streetcomplete/data/Database.kt
```diff
@@ -89,14 +96,12 @@ interface CursorPosition {
     /* It would be really nice if the interface would be just
        operator fun <T> get(columnName: String): T
        if T is one of the below types. But this is not possible right now in Kotlin AFAIK */
-    fun getShort(columnName: String): Short
```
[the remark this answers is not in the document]
FloEdelmann: First, the [multiplatform SQLite library](https://developer.android.com/kotlin/multiplatform/sqlite) indeed does not support shorts. And second, it is never used in the current code, so I figured it's not needed.

FloEdelmann:
> there is _probably_ little reason to keep the `Database` interface

But we will need to keep the `AndroidDatabase` class (renamed to something better of course) as a wrapper around the multiplatform SQLite library, because that library is quite bare-bones. Is it better to then refer to that class everywhere? Or should we keep the interface and let the class implement it, just like it is now?

app/src/main/java/de/westnordost/streetcomplete/data/AndroidDatabase.kt
westnordost: well, hmm, I'd prefer if it would be passed, instead.

FloEdelmann: Done in [`7aa8f04`](https://github.com/streetcomplete/StreetComplete/pull/6119/commits/7aa8f04ed27cb23e9b75f55dd9076ef01645d4f1).

westnordost:
You are probably right about this. The interface of [that library](https://developer.android.com/kotlin/multiplatform/sqlite) is pretty bare-bones and thus it would be less convenient (and more error-prone) to use it directly.

westnordost (approved):

Issue #3652, opened 2022-01-19
Allow paths/steps to be converted to escalators/travelators


peternewman:
**Use case**
Similar to #3288 steps which are actually escalators (or travelators) are sometimes not tagged as such.

**Proposed Solution**
Similar to #3288 add an other-answer to add conveying:
https://wiki.openstreetmap.org/wiki/Key%3Aconveying

I appreciate you could leave a note, but then it needs processing; also you can answer all the questions about it anyway, so it's a bit odd to go down the can't say route to add more info, and a bit vague if you just drop a note in the general area, then you've got to specify which way it is too, especially because (at least in London), there are often steps nearby too.

westnordost:
Escalators outside of buildings are rare enough that I think "can't say" is enough.

peternewman:
Ah, I hadn't realised about that filtering; I've opened #3654 to fix the data issue that actually caused this.

PR #6135, opened 2025-02-20
split out escalators from AddStepsIncline quest to allow different quest wording

andrewharvey:
At present the AddStepsIncline quest applies to both steps and escalators, however the quest wording only mentions steps. This can lead to confusion.

For example https://www.openstreetmap.org/way/1241003118 was tagged `highway=steps` + `conveying=yes` to indicate it is an escalator, however the user was then asked "Which direction leads upwards for these steps?" to which they answered via the note https://www.openstreetmap.org/note/4351347 "Not steps, escalator".

By creating a new quest we can present the wording "Which direction leads upwards for these escalators?" to reduce confusion, and also help catch mistaggings of `conveying=yes/no`.

#3652 requested the ability to convert between the two, but that's a different issue and this change can be applied even without giving an opportunity to answer "these are escalators" / "these are steps".


Helium314:
> By creating a new quest we can present the wording "Which direction leads upwards for these escalators?"

Why not just edit `getTitle(tags: Map<String, String>)` in the normal incline quest?

andrewharvey:
1. Because I couldn't find that being done in any of other quests code I looked at and wanted to try and retain the code style.
2. I don't actually know how to do it like that

FloEdelmann:
An example can seen in

https://github.com/streetcomplete/StreetComplete/blob/6fd88ebc71284e5e398a4012e3ad56eb4949fedc/app/src/main/java/de/westnordost/streetcomplete/quests/general_fee/AddGeneralFee.kt#L28-L33

<details>
<summary>Other examples (for reference)</summary>

in no particular order:

* https://github.com/streetcomplete/StreetComplete/blob/6fd88ebc71284e5e398a4012e3ad56eb4949fedc/app/src/main/java/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeight.kt#L58-L69
* https://github.com/streetcomplete/StreetComplete/blob/6fd88ebc71284e5e398a4012e3ad56eb4949fedc/app/src/main/java/de/westnordost/streetcomplete/quests/barrier_opening/AddBarrierOpening.kt#L48-L53
* https://github.com/streetcomplete/StreetComplete/blob/6fd88ebc71284e5e398a4012e3ad56eb4949fedc/app/src/main/java/de/westnordost/streetcomplete/quests/building_levels/AddBuildingLevels.kt#L33-L36
* https://github.com/streetcomplete/StreetComplete/blob/6fd88ebc71284e5e398a4012e3ad56eb4949fedc/app/src/main/java/de/westnordost/streetcomplete/quests/construction/MarkCompletedHighwayConstruction.kt#L25-L41
* https://github.com/streetcomplete/StreetComplete/blob/6fd88ebc71284e5e398a4012e3ad56eb4949fedc/app/src/main/java/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt#L132-L141

</details>

andrewharvey:
Thanks @FloEdelmann I obviously didn't check enough other quests! I can try to make the adjustment to use that style, assuming that the idea behind this PR is considered accetable?

andrewharvey:
> Yeah, I guess the wording change makes sense. However, maybe we can just rename "Which direction leads upwards for these steps?" to "Which direction leads upwards here?"? Does the name label below the question already mention "Stairs" or "Escalator"?

I prefer to have "Steps/Stairs" and "Escalators" included in the quest title as it helps validate the data and catch any potentially mistagings. It also helps the quest mapper, user (quester? not sure what you call us solvers?) as you are more clear about exactly what the "direction leads up" is referring to.

andrewharvey:
I've refactored to use a single quest but just varying the title. I haven't tested this yet.

westnordost (changes requested):
(see above)

andrewharvey:
> Actually, when I introduced the name label, I did that to reduce the text. Because the less text the better (as long as it remains clear). So, in general I concur with @FloEdelmann , it would be better if the wording was changed to "Which direction leads upwards here?". This then would also cover other things with an incline, such as ramps.

Ah that makes sense, I can see you do this already for other quests like "What is the surface here?".

goldfndr:
Might specify "this escalator" instead of "these escalators" for the string, as more than one escalator seems unlikely to be tagged as a single object. \[If they are, they're possibly perceived as a single mechanism (interconnected motors), although that doesn't seem safe (consider [LOTO](https://en.wikipedia.org/wiki/Lockout%E2%80%93tagout)).\]

westnordost:
Alright, so the mentioning of the escalator in the label is implemented in PR #6140. It was even much more work than I expected 😅

Would you make the change I requested?

andrewharvey:
Sorry I'm confused, if the label is fixed to display "escalator" then we don't need this PR, we can just change `quest_steps_incline_title` from "Which direction leads upwards for these steps?" to "Which direction leads upwards here?"

Changeset comment can probably stay the same.

andrewharvey:
ok done, much simpler now.

PR #6140, opened 2025-02-23
support presets with "*" values

### LINKED ISSUE #4607: icons of vacant shops are sometimes wrong
### How to Reproduce
Search for vacant shops that have additional tags, such as `building=house` or `indoor=room`
The icons of the house or the room are displayed on the map, but in the "editing window" is the correct icon for vacant shops.

grafik
grafik
Locations to reproduce: E. g.: [here](https://www.openstreetmap.org/way/201705739) or [here](https://www.openstreetmap.org/?mlat=53.11138&mlon=8.74746#map=18/53.11138/8.74746
)

### Expected Behavior
Show the correct icon on the map

### Versions affected
v49

### LINKED ISSUE #5728: Ignore conveying way or steps in the surface overlay.
**Use case**
In the surface overlay, an escalator was displayed red. When selecting in, I was told by the StreetComplete interface that it was steps, and I was asked what its pavement was.
I left a note as it felt odd: I was assuming that it was mapped as steps and not as an elevator.
But when later checking the tags of the way, it was actually mapped as an escalator:
```
conveying=forward
highway=steps
incline=down
```

**Proposed Solution**
I would suggest to ignore the ways with `conveying=*` in the surface overlay, or otherwise rephrase a bit the associated quest to explicitly name it an escalator and not steps.

### LINKED ISSUE #6085: don't show club=yes and unknown club values as "Other kind of place"
Right now club=yes and unknown club values are shown as "Other kind of place". I think SC should show "Club" instead. Also the input field for the name should be shown for club=yes and unknown club values.

westnordost:
using de.westnordost:osmfeatures:7.0

Might
- fix #5728 
- fix #4607
- fix #6085
- unblock #6135

Helping with testing this would be appreciated. This is a change into some really (over-*)complex parts of StreetComplete, so quite a potential for issues.

---

\* _but necessarily so_


HolgerJeromin:
If you create a debug build some will test it for sure :-)

westnordost:
After reading through the PR, at least these things should be tested:

- [ ] dialog that appears when selecting "it doesn't exist anymore" functions correctly for shops, amenities, ...: when selecting a generic feature (like "club"), when selecting that it's vacant
- [ ] "shop still vacant?" quest functions correctly (with some real data)
- [ ] "which shop type is this" quest functions correctly (with some real data, testing both answers)
- [ ] things overlay: *disused* POIs are displayed/handled correctly
- [ ] things overlay: adding POIs works correctly, including when selecting a generic feature (requires adding `traffic_sign=*` to the list of "Things" temporarily)
- [ ] places overlay: *disused* POIs are displayed/handled correctly
- [ ] places overlay: marking a POI as vacant will change tags to `disused:*=*`, marking a newly created POI as vacant will add `shop=vacant` instead

westnordost:
Here's a debug APK:

https://github.com/streetcomplete/StreetComplete/actions/runs/13485459314

(Don't ask me why it is packed in a zip. Is this normal, @FloEdelmann ?)

westnordost:
For me, the same way shows as "auf Ebene -1; 0: Rolltreppe"

HolgerJeromin:
> For me, the same way shows as "auf Ebene -1; 0: Rolltreppe"

My about screen shows v58.2. I think my upgrade of the apk did not succeeded... Sorry for the noise

westnordost:
Alright, I tested everything listed. The boxes unchecked will not be done (though, the second checkbox would be solved with https://github.com/openstreetmap/id-tagging-schema/pull/1457 ).

Would anyone review and/or test it?

mnalis:
> Would anyone review and/or test it?

I've been playing a little with provided debug APK, and I've noticed vacant icon problem (not the uncheched _"vacant shop icon should be the same on the map and in the selection screen(s)"_ problem though, but maybe related):

- disused shops (e.g. `disused:shop=clothes`) have nice understantable icon (crossed-over-shop), but disused amenities (e.g. `disused:amenity=cafe`) have generic (i.e. "working") shopping-bag icon. e.g.
small_Screenshot_20250227_223041_StreetComplete Dev

I would've expected that both `disused:*` icons look the same (or if different, that both have some kind of crossed-over icon indicating they are disused).

mnalis:
I've also noticed that in shop overlay I can click on that `disused:amenity=cafe` (whose icon does not look like disused, nor like cafe, as shown above), and in three-dots menu I can still choose `It's vacant`, which seems to add ***additional*** `shop=vacant` tag to it. 

While one perhaps shouldn't chose that option in normal usage (well, I wouldn't, but I was misled by the wrong icon) it seems like undesirable behaviour to me :shrug:

westnordost:
That's - both - odd, will have a look, thanks!

Issue #3031, opened 2021-07-08
"Does it offer gluten free food?" Quest


peternewman:
### General
Affected tag(s) to be modified/added:
https://wiki.openstreetmap.org/wiki/Key:diet:gluten_free
 
It's becoming increasingly popular as a diet, as well as being vital for survival for Coeliacs.

### Checklist
Checklist for quest suggestions (see [guidelines](https://github.com/streetcomplete/StreetComplete/wiki/Adding-new-Quests-to-StreetComplete)):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary
- [x] 💤 Not an overwhelming percentage of quests have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of map data (Worth the effort)

### Ideas for implementation
This can be implemented exactly the same way as the existing dietary quests

HolgerJeromin:
Quite hard to survey.
I remember a small sign in the paper menu in very few restaurants...

westnordost:
See holger. It is extremely rare that restaurants put up signs for this.  People with gluten intolerance will know themselves what to avoid.

Also, you skipped on looking at if it is applicable to the new quest guidelines,  this alone is already a reason for rejection.

rhhsm:
With the increasing number of diet quests, could it be an idea to combine them into one, with a list of check boxes that can be ticked for each diet available? There's a list here that may inspire https://www.lufthansa.com/bg/en/special-meals

peternewman:
> See holger. It is extremely rare that restaurants put up signs for this. People with gluten intolerance will know themselves what to avoid.

In my experience its often got an equivalent level of signage to vegan (and to some extent vegetarian), i.e. just small notes against some items on the menu. Aside from specifically that diet only restaurants.

I think they often need to know about the risk of cross-contamination, so some vegetables say may be gluten free by default, but if they've used the same pan to cook them in, a Coeliac would still need to avoid them, so only places with signed GF meals can probably be guaranteed safe.

> Also, you skipped on looking at if it is applicable to the new quest suggestions guidelines, this alone is already a reason for rejection.

Sorry, I just copied #2990 and changed the tag name.

[...]

PR #6143, opened 2025-02-25
implement Gluten-free quest

mnalis:
Fixes https://github.com/streetcomplete/StreetComplete/issues/3031

TODO:
- [x] add proper icon for `ic_quest_glutenfree.xml`
- [x] test if it works propely


agent-redd:
The logic and text looks good to me. If you can create anf share a dev built, I would like to test it.

mnalis:
Thanks, dev build is available now at https://github.com/mnalis/StreetComplete/actions/runs/13529866385

mnalis:
I've also added a proper icon, but feel free to replace it without hard feelings if it is not up to standards; I'm no great artist

res/graphics/quest/gluten.svg
FloEdelmann, line 1: Please rename this file to `glutenfree.svg`, so that it matches the XML version (to not confuse the [quest list generator](https://github.com/streetcomplete/StreetComplete/actions/workflows/generate-quest-list.yml)).

westnordost:
Icon, hmm...  it's weird to display gluten for a quest that asks whether something is gluten-free, no? What do you think about this kind of 🚫 icon on top of gluten?
In any case, the style of the icon (with those fat black strokes) doesn't fit the flat-style of the other icons. But well, don't worry, I will create a good one after/before this is merged.

res/graphics/quest/gluten.svg
mnalis: renamed, thanks

app/src/main/res/values/strings.xml
```diff
@@ -1059,6 +1059,9 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
 
     <string name="quest_cycleway_width_title">"How wide is the bicycle path here?"</string>
 
+    <string name="quest_dietType_glutenfree_name_title2">Are gluten-free products offered here?</string>
+    <string name="quest_dietType_explanation_glutenfree">Preparing gluten-free products requires knowledge and procedures to avoid contamination of ingredients. Gluten-free food would be explicitely labeled as such.</string>
```
westnordost, line 1063: Are you sure about this? I thought in the linked ticket, we came to the conclusion that "gluten-contamination" is not a thing as gluten intolerance is an intolerance rather than an allergy? Maybe I misremember.

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddGlutenFree.kt
```diff
@@ -0,0 +1,50 @@
...
+    override val changesetComment = "Specify whether places are gluten-free"
+    override val wikiLink = "Key:diet:gluten_free"
+    override val icon = R.drawable.ic_quest_glutenfree
+    override val isReplacePlaceEnabled = true
+    override val achievements = listOf(CITIZEN)
+    override val defaultDisabledMessage = R.string.default_disabled_msg_go_inside
+
+    override val hint = R.string.quest_dietType_explanation_glutenfree
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_dietType_glutenfree_name_title2
```
westnordost, line 37: rename to `R.string.quest_dietType_glutenfree_name_title`?

mnalis: OK, renamed (I don't know why, but all the other diet quests had `title2` in name, so I just kept it like that)

app/src/main/res/values/strings.xml
mnalis: I'm not an expert on the subject :shrug:  

I was going by the description on the [wiki](https://wiki.openstreetmap.org/wiki/Key:diet:gluten_free). I've tried to modified it now to be more lax, does that sound better?

agent-redd: I was the editor who updated the wiki few days ago and spent quiet some investigation time into this topic. 

For people with Coeliac disease even the slighest amount of gluten can lead to damage in the gut. Which then for example can lead to a shortage of calcium which the body then takes from it's own bowns.

Coeliac disease is not an intolerance. It is an autoimmune disorder, where the body starts to fight against itself. 

Maybe we can enhance the text to sensitize the users even more. Example:

* Preparing gluten-free products requires knowledge and procedures to avoid contamination of ingredients. Gluten-free food would be explicitely labeled as such.  **For People with coeliac disease is recommended to not eat the slightest amount of gluten because it triggers harmful immune responses that damage their intestines.**

westnordost (approved):

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddGlutenFree.kt
```diff
@@ -0,0 +1,50 @@
...
+    override val elementFilter = """
+        nodes, ways with
+        (
+          amenity ~ restaurant|cafe|fast_food|food_court and food != no
+          or amenity ~ pub|nightclub|biergarten|bar and food = yes
+          or shop ~ supermarket|convenience
+        )
+        and (
+          !diet:gluten_free
+          or diet:gluten_free != only and diet:gluten_free older today -4 years
```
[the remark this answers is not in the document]
westnordost: A shop that *only* sells gluten free products, i.e. is a specialized "gluten free" shop will certainly never change that except if there is a new shop (under a new name etc.). This statement is present for all the other diet quests for the same reason. I don't see why gluten-free-ness would be an exception.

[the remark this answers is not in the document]
mnalis: I agree, if specialized glutrenfree-only shop suddenly stops being gluten-free, it is very probably because it is now a different shop, and we have a procedure for that under `Uh...` / `It does not exist` / `It is now a(n) xxxx`.

agent-redd (approved):

ScharadaValcta:
Can someone merge ?

app/src/main/java/de/westnordost/streetcomplete/quests/diet_type/AddGlutenFree.kt
```diff
@@ -0,0 +1,50 @@
...
+
+
+class AddGlutenFree : OsmFilterQuestType<DietAvailabilityAnswer>() {
+
+    override val elementFilter = """
+        nodes, ways with
+        (
+          amenity ~ restaurant|cafe|fast_food|food_court and food != no
+          or amenity ~ pub|nightclub|biergarten|bar and food = yes
+          or shop ~ supermarket|convenience|deli
```
FloEdelmann, line 21: if #6200 gets merged, we should probably also apply the same changes here:

```suggestion
          or shop ~ supermarket|convenience|deli
          or tourism ~ alpine_hut and food != no
```

ghost:
I should take inspiration from this pull for my keto quest.

matkoniecz:
better wait a moment for comment on the issue created about it - adding quest for something with less than 100 tags is dubious

Issue #6068, opened 2024-12-31
OutOfMemoryError when downloading


Helium314:
When downloading large areas with high element density, the app may crash with an OOM error.
I got a bunch of error reports for SCEE since release of 59.0, and reproduced it also in SC. Never seen it with older versions.

**How to Reproduce**
1. go to a well mapped city, e.g. https://www.openstreetmap.org/#map=15/25.00673/121.31786
2. zoom out as much as possible, but to a level where you can still download data
  note that it the area actually doesn't need to be close to the 12 km² limit, I got reports for 7.4 km² and even one coming from auto-downloading a 3.5 km² area
3. manually download data
4. see crash

**Expected Behavior**
Download completes, like it did on older versions.
I think it's related to the download, so probably not connected to MapLibre, as the stacktrace contains `MapDataApiClient.getMap`. #5686 was tested for performance, but looks like increased memory use was not a consideration.

<details><summary>Stacktrace for 60.0</summary>

```
java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Failed to allocate a 38910912 byte allocation with 25165824 free bytes and 29MB until OOM, target footprint 195379296, growth limit 201326592
    at androidx.work.impl.utils.futures.AbstractFuture.getDoneValue(SourceFile:515)
    at androidx.work.impl.utils.futures.AbstractFuture.get(SourceFile:474)
    at androidx.work.impl.WorkerWrapper$2.run(SourceFile:316)
    at androidx.work.impl.utils.SerialExecutorImpl$Task.run(SourceFile:96)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
    at java.lang.Thread.run(Thread.java:919)
Caused by: java.lang.OutOfMemoryError: Failed to allocate a 38910912 byte allocation with 25165824 free bytes and 29MB until OOM, target footprint 195379296, growth limit 201326592
    at java.lang.StringFactory.newStringFromChars(StringFactory.java:260)
    at java.lang.StringBuilder.toString(StringBuilder.java:413)
    at io.ktor.utils.io.charsets.EncodingKt.decode(SourceFile:102)
    at io.ktor.utils.io.core.StringsKt.readText(SourceFile:246)
    at io.ktor.utils.io.core.StringsKt.readText$default(SourceFile:245)
[...]

PR #6144, opened 2025-02-25
Use kotlinx-io for access to files

westnordost:
...almost everywhere.

The following should be tested before merge:
- [x] deletes tangram cache correctly (or well, at least no error is thrown)
- [x] parses country infors correctly
- [x] downloads, caches and displays user avatars shown in the note discussion & profile screen correctly
- [x] uploading notes with photos works correctly
- [x] statistics are downloaded correctly
- [x] compass works as before


westnordost:
Part two - using `Sink` and `Source` to decode/serialize OSM XML data on download/upload is still blocked by https://github.com/pdvrieze/xmlutil/issues/217

(I.e. that part is already multiplatform, but could be more performant / less memory-hungry, see also #6068)

westnordost:
Also, reading from a file and streaming it into a HTTP request body is blocked by

https://youtrack.jetbrains.com/issue/KTOR-8241/Support-conversion-from-kotlinx-io-Source-to-ByteReadChannel

FloEdelmann:
The unit tests fail with these changes: https://github.com/streetcomplete/StreetComplete/actions/runs/13634924981/job/38111192985

Seems like these tests need to be adapted:

- AvatarsDownloaderTest
- PhotoServiceApiClientTest

PR #6150, opened 2025-03-03
stream parse XML

westnordost:
When downloading (OSM data), no longer download it into a string, then parse it. Rather, parse it as the bytes come in from the http connection. This reduces memory footprint and also makes the process much faster. (The bigger the data, the bigger the performance gain.)

This is currently draft, because this has been implemented using a not-yet-released version of xmlutil, see https://github.com/pdvrieze/xmlutil/issues/217#issuecomment-2684638631 ff.


westnordost:
new xmlutil version is released.

PR #6152, opened 2025-03-03
handle HTTP 413 response from OSM API

### LINKED ISSUE #217: Offline mode
I live in an area where there is spotty or no cellular network access, coincidentally this is also an area that great need for this app.

Being able to use this app while offline would be invaluable.

### LINKED ISSUE #6151: Upload error on too large bounding box
I am getting some crash reports.

```
de.westnordost.streetcomplete.data[https://de.westnordost.streetcomplete.data/].ApiClientException: Client request(POST https://api.openstreetmap.org/api/0.6/changeset/163137077/upload) invalid: 413 Payload Too Large. Text: "Changeset bounding box size limit exceeded."
at de.westnordost.streetcomplete.data[https://de.westnordost.streetcomplete.data/].osm.mapdata.MapDataApiClient.uploadChanges(SourceFile:42)
at de.westnordost.streetcomplete.data[https://de.westnordost.streetcomplete.data/].osm.mapdata.MapDataApiClient$uploadChanges$1.invokeSuspend(Unknown Source:16)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(SourceFile:33)
at kotlinx.coroutines.DispatchedTask.run[https://kotlinx.coroutines.dispatchedtask.run/](SourceFile:100)
at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run[https://worker.run/](SourceFile:113)
at kotlinx.coroutines.scheduling.TaskImpl.run[https://kotlinx.coroutines.scheduling.taskimpl.run/](SourceFile:89)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(SourceFile:586)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(SourceFile:820)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(SourceFile:717)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run[https://worker.run/](SourceFile:704)
Caused by: io.ktor.client.plugins.ClientRequestException: Client request(POST https://api.openstreetmap.org/api/0.6/changeset/163137077/upload) invalid: 413 Payload Too Large. Text: "Changeset bounding box size limit exceeded."
at io.ktor.client.plugins.DefaultResponseValidationKt$addDefaultResponseValidation$1$1.invokeSuspend(SourceFile:52)
... 8 more
```

Some time ago, the OpenStreetMap API introduced some API limits. 

One of these limits was a maximum bounding box size of changesets, relative to how new the user is.

I understand why it was done. A vandal moved nodes of (renamed to something vulgar) a couple of (hundreds of) ways over half the world to achieve an effect like criss-crossing barrier-tape on the rendered map with said vulgar writing on it. A side-effect was that the rendering pipeline of map renderers was completely overloaded, as suddenly millions and millions of tiles needed to be re-rendered.

The side-effect of this measure is, even though nodes of ways cannot be moved at all in StreetComplete, that this app is affected, because of how the limit was implemented. It is rather annoying that this limit wasn't implemented as a limit for the maximum distance a node (of a way) can be moved, but as a maximum bounding box size.

Well, anyway, we have to handle a 413 HTTP error code on upload now and as a response, create a new changeset, then retry the upload.

westnordost:
fixes #6151

Need to add tests (but not tonight). Is the code itself comprehensible, @Helium314 , @FloEdelmann , @matkoniecz ?


westnordost:
(Converting to draft because I still need to add tests)

mnalis:
> fixes https://github.com/streetcomplete/StreetComplete/issues/217

Ummm, maybe it fixes:
- https://github.com/streetcomplete/StreetComplete/issues/6151

instead of (already closed years ago):
- https://github.com/streetcomplete/StreetComplete/issues/217 

?

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
```diff
@@ -22,31 +24,50 @@ class ElementEditUploader(
...
+                return uploadChanges(edit, localChanges, false)
             }
-        } else {
-            try {
-                uploadChanges(edit, localChanges, false)
-            } catch (e: ConflictException) {
-                // either changeset was closed, or element modified, or local element was cleaned from db
+            // either changeset was closed, element modified on remote, element not available in DB
+            catch (e: ConflictException) {
+                // continue execution outside of this if-block (-> try with remote data)
```
FloEdelmann, line 42: Maybe this bit can be made more readable by extracting the remote upload handling into a separate function and calling that here and after the `if` block?

Or maybe even extract the local upload handling into a separate function, too?

westnordost:
Uh, yes, don't know why I wrote 217.

app/src/main/java/de/westnordost/streetcomplete/data/osm/edits/upload/ElementEditUploader.kt
westnordost: So, I've rewritten it. Take another look! :-)

FloEdelmann: Yeah, that's a little easier to read. Thanks!

FloEdelmann (approved):

PR #6158, opened 2025-03-09
ignore private features for bus-stop related quests

k-yle:
similar to [many other quests](https://grep.app/search?f.repo=streetcomplete/StreetComplete&q=access+!~+private|no), it would be good if the bus-stop quests ignored private features. 

For example, bus stops within a gated business-park, factory, or school, all of which are inaccessible to the public.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/bus_stop_bench/AddBenchStatusOnBusStop.kt
@@ -18,6 +18,7 @@ class AddBenchStatusOnBusStop : OsmFilterQuestType<Boolean>() {
           or (highway = bus_stop and public_transport != stop_position)
         )
         and physically_present != no and naptan:BusStopType != HAR
+        and access !~ no|private
         and (!bench or bench older today -4 years)
     """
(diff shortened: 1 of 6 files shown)


westnordost:
A private bus stop??

mnalis:
> A private bus stop??

Sounds funny, but we've had a few: it's company busses which only takes workers to the factory operated by the company. While _most_ stops are regular ones to pick-up passengers (and are used by public busses too), the last one or two are inside factory grounds, and thus "access=private". 

**However, I don't see a point in ignoring those**, and would in fact prefer the quests continue to be asked on them too. 

Because, you won't be _near_ those private bus stops to even see the quest (much less solve them) unless you have already reached those private grounds (by the said private bus) - and the fact that you reached them implies that you probably have permission to access them, so why not improve their data if you can and want to (and it is quite useful data for those allowed to use those stops, esp. stop name/ref). 

And as users without permissions won't be reaching those private bus stops, they won't be spammed by the quest either, so no need to hide the quest on their behalf.

westnordost:
Depends, the bus stop may be not far from the entrance to the company grounds. (We neither ask for properties of private roads)

k-yle:
The problem we have is that these private stops might be very close to where the user is standing, but on the other side of a fence/wall (inside a school/factory/compound), thereby making them unsurveyable

PR #6160, opened 2025-03-09
Fix the PR number for the bus stop names improvement

peternewman:


FloEdelmann:
Thanks for noticing! I also fixed in the release description: https://github.com/streetcomplete/StreetComplete/releases/tag/v60.2

PR #6167, opened 2025-03-16
Baby Changing Tables: Ask on more POIs

agent-redd:
I would like to enhance the quest to more POIs. I added POIs where its likely that a baby changing table is available from my experience.
This Quest is disabled by default as indoor survey is needed. Therefore I think it is okay to add more places to this quest.  

Debug APK for this PR built here:  https://github.com/agent-redd/StreetComplete/actions/runs/13877790232

Tested succesfully on a few bakeries, community_centres and a library. (without upload)

I will be open and thankful for any review and feedback! Thank you in advance.


HolgerJeromin:
Perhaps filter bakeries with `indoor_seating=yes`...

agent-redd:
> Perhaps filter bakeries with `indoor_seating=yes`...

agent-redd:
> Perhaps filter bakeries with `indoor_seating=yes`...

yes, good idea to ask on less bakeries. I will do that

westnordost:
Sounds good. 

But `shop=chemist` ? Do some chains offer that?

agent-redd:
> Sounds good.
> 
> But `shop=chemist` ? Do some chains offer that?

yes in Germany all chemist chains offer baby changing tables if the shop is big enough. My experience here is about DM, Rossmann but not Müller. There are DMs and Rossmanns with and without baby changing tables. For Müller I was not able to find any shop with changing table.

For other countries I don't have any experience. Also my overpass searches did hardly find any changing_table=* tagging for shop=chemist outside of Germany. If we release this quest I can follow up with OsmCha and look if there is diverse yes/no tagging in other countries as well. 

As always for this quest.. the survey is not very easy. Most of the times you need to look in the hidden corners or the toilets of the places.

westnordost (approved):

PR #6173, opened 2025-03-17
Prevent Surface Quest from immediately asking again when choosing paved or unpaved

### LINKED ISSUE #6172: SC asking for surface type twice
When defining the surface of a path for the first time, it seems like check_date:surface is not added. This results in SC asking for the surface twice in a row. The second answer will then add the check_date. Not sure if this is limited to surfaces only. Happened with surface=paved. 

**How to Reproduce**
Add surface to path that has no surface yet. Look at the path after adding the surface type. 

**Expected Behavior**
SC only asking once. 

**Versions affected**
60.2 from F-Droid

agent-redd:
This PR is created to fix #6172 

Adds time constraint for surface = paved|unpaved  precondition.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
@@ -25,7 +25,7 @@ class AddCyclewayPartSurface : OsmFilterQuestType<Surface>() {
           !cycleway:surface
           or cycleway:surface ~ ${INVALID_SURFACES.joinToString("|")}
           or (
-            cycleway:surface ~ paved|unpaved and cycleway:surface older today -6 years
+            cycleway:surface ~ paved|unpaved and cycleway:surface older today -6 months
             and !cycleway:surface:note
             and !check_date:cycleway:surface
           )
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
@@ -25,7 +25,7 @@ class AddCyclewayPartSurface : OsmFilterQuestType<Surface>() {
           !cycleway:surface
           or cycleway:surface ~ ${INVALID_SURFACES.joinToString("|")}
           or (
-            cycleway:surface ~ paved|unpaved
+            cycleway:surface ~ paved|unpaved and surface older today -6 years
             and !cycleway:surface:note
             and !check_date:cycleway:surface
           )
--- a/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/quests/surface/AddCyclewayPartSurface.kt
@@ -25,7 +25,7 @@ class AddCyclewayPartSurface : OsmFilterQuestType<Surface>() {
           !cycleway:surface
           or cycleway:surface ~ ${INVALID_SURFACES.joinToString("|")}
           or (
-            cycleway:surface ~ paved|unpaved and surface older today -6 years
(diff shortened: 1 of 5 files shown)


agent-redd:
I used 6 years for all surface quests. We could adjust them individually if needed.

agent-redd:
Debug APK is available here https://github.com/agent-redd/StreetComplete/actions/runs/13909019962

I succesfully tested for highway=service and highway=footway. Still need to find untagged objects the other quest types. Need to dig into Overpass until I find objects to test.

agent-redd:
Update: I now tested all 5 modified subquests for issue fix and regression:
- [AddCyclewayPartSurface.kt](https://github.com/streetcomplete/StreetComplete/pull/6173/files#diff-92cc359fcd8a1d5b22f25c800721b751316fce4079ef63c07a9f1183b715a4dc) on a segregated highway=cycleway
- [AddFootwayPartSurface.kt](https://github.com/streetcomplete/StreetComplete/pull/6173/files#diff-1de7ea522c76b117ab0d4848aaa77ff3ef5671b5bbf10cc57dad0a809064dfbc) on a segregagted highway=cycleway
- [AddPathSurface.kt](https://github.com/streetcomplete/StreetComplete/pull/6173/files#diff-1ff1ea42d145e9fd37a9749a5cf8a920c694dcb6ccaa641ddd1fd8c337d94caf) on a non-segregated highway=path
- [AddPitchSurface.kt](https://github.com/streetcomplete/StreetComplete/pull/6173/files#diff-1b4f219a5fd3cd2266deee663b726cd3e3d0272a2b03e06a96789b08fe861b42)  on a leisure=pitch
- [AddRoadSurface.kt](https://github.com/streetcomplete/StreetComplete/pull/6173/files#diff-1c3d0e860c36bf02643af40770a15a520c8853d9a5a13d94f412bfe646c8a4df) on a highway=service

(tested without uploading the modifications on the debug APK)

agent-redd:
Previously on CyclewayPartSurface and FootWayPartSurfe I used `surface` instead of `cylceway:surface` and `footway:surface`. It is strange that the subquests still worked as expected and not asked twice for it. Maybe `surface older today -6 years` gets `FALSE` when the `surface=*` does not even exist?

Nevertheless I fixed it now, rebuilt a debug APK https://github.com/agent-redd/StreetComplete/actions/runs/13910833668 and succesfully tested both subquests again.

westnordost:
Hmm... I think it would be much more useful if StreetComplete would continue to prompt its users **once** for `paved` (etc.) ways to specify a (more detailed) surface if possible, as these generic values are much less useful than precise values.

It is not by chance that these surface types are mentioned in the filter explicitly, there is a [mountain](https://taghistory.raifer.tech/#***/surface/paved&***/surface/unpaved) of imprecise values tagged, which could be enhanced. That's the job of this app, after all, to enhance the precision of OpenStreetMap.

So, what to do? I think the best solution here would be to not change the filter, but instead, when the user tags `paved` or another generic value, **always** also tag the `check_date:surface` in order to prevent StreetComplete to immediately ask again.
This should solve the issue while retaining the functionality to help overcome that mountain of imprecise values.

agent-redd:
It would still prompt for paved/unpaved after the time constraint which I currently set to 6 years.

Maybe we could just set this down to 1 year? As @HolgerJeromin  mentined in #6172 `check_date:surface` may be considered as spam by some users.

agent-redd:
I updated the PR title to make it more clear

agent-redd:
Thanks for all the feedback. Even when this PR is not the favorite solution to solve #6172, I changed the time constraint from 6 years to 6 months, as StreetComplete is seen as a on-site survey tool to quickly get rid of generic paved/unpaved values to more specific values like asphalt, paving stones, grass, sand, etc. 

Setting the time to 6 months, we can make sure that the paved/unpaved values get often requested until they finally have a more specific value.

agent-redd:
closing my PR, since it is not reflecting the proposed solutions in #6172

PR #6183, opened 2025-03-23
Change wording of `charging_station` quest

kadiwa4:
Not every [`charging_station`](charging_station_wording) is for cars. For example, [here](charging_station_wording) someone encountered a charging station for bike batteries and was asked how many cars could be parked there.


mnalis:
Makes sense to me to make it more generic (and force translations). There apparently exist also truck-only chargers

westnordost:
This is actually deliberate. Data users usually assume `amenity=charging_station` without extra tags to be a charging station for cars if not stated otherwise.
If it is in fact not a charging station for cars, users should leave a note, so that `bicycle=yes` and/or `motorcar=no` etc. can be added.

mnalis:
(somewhat related discussion: https://community.openstreetmap.org/t/bicycle-chargers-discussion-in-advance-of-potential-proposal/127195)

PR #6200, opened 2025-04-08
add tourism=alpine_hut in vegetarian and vegan quest

tiptoptom:


HolgerJeromin:
Fine with me, but we have (nearly) no such entries in openstreetmap right now:
https://taginfo.openstreetmap.org/tags/tourism%3Dalpine_hut#combinations

PR #6202, opened 2025-04-08
Migrate to multiplatform SQLite library

### LINKED ISSUE #5417: Implement `Database` interface on iOS

FloEdelmann:
Attempt to fix #5417.

As mentioned in #6119, the [multiplatform Jetpack SQLite library](https://developer.android.com/kotlin/multiplatform/sqlite) is quite bare-bones, so a lot of stuff that was previously handled by the [Android SQLite library](https://developer.android.com/training/data-storage/sqlite?hl=en) now has to be implemented on our side.


app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,278 @@
...
+        databaseConnection.execSQL("BEGIN IMMEDIATE TRANSACTION")
+        try {
+            val result = block()
+            databaseConnection.execSQL("END TRANSACTION")
+            return result
+        } catch (t: Throwable) {
+            databaseConnection.execSQL("ROLLBACK TRANSACTION")
+            throw t
+        }
+    }
```
FloEdelmann, line 136: This doesn't work correctly yet: I consistently get an error when downloading data:

```
Unable to download
android.database.SQLException: Error code: 5, message: cannot commit transaction - SQL statements in progress
	at androidx.sqlite.driver.bundled.BundledSQLiteStatementKt.nativeStep(Native Method)
	at androidx.sqlite.driver.bundled.BundledSQLiteStatementKt.access$nativeStep(BundledSQLiteStatement.jvmAndroid.kt:1)
	at androidx.sqlite.driver.bundled.BundledSQLiteStatement.step(BundledSQLiteStatement.jvmAndroid.kt:100)
	at androidx.sqlite.SQLite.execSQL(SQLite.kt:56)
	at de.westnordost.streetcomplete.data.StreetCompleteDatabase.transaction(StreetCompleteDatabase.kt:130)
	at de.westnordost.streetcomplete.data.osm.mapdata.WayDao.getAll(WayDao.kt:67)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController$getWays$1.invoke(MapDataController.kt:193)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController$getWays$1.invoke(MapDataController.kt:193)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataCache.getWays$lambda$26(MapDataCache.kt:265)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataCache.$r8$lambda$ZWQirpDjfvRnK1EiAAaEtzZT8dc(Unknown Source:0)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataCache$$ExternalSyntheticLambda8.invoke(D8$$SyntheticClass:0)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataCache.getElements(MapDataCache.kt:235)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataCache.getWays(MapDataCache.kt:265)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController.getWays(MapDataController.kt:193)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController.completeMapData(MapDataController.kt:151)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataController.putAllForBBox(MapDataController.kt:65)
	at de.westnordost.streetcomplete.data.osm.mapdata.MapDataDownloader$download$2.invokeSuspend(MapDataDownloader.kt:29)
```

The [Android SQLite library](https://developer.android.com/training/data-storage/sqlite?hl=en) seems to have done much more than this function currently does: https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/core/java/android/database/sqlite/SQLiteSession.java

But I have no idea how much of that is needed. If it's everything, then it's a lot, and I think we should then wait until the [multiplatform Jetpack SQLite library](https://developer.android.com/kotlin/multiplatform/sqlite) has some convenience functions around transactions.

mnalis: I'm no expert on Android side, but that SQLite error _"cannot commit transaction - SQL statements in progress"_ sounds like there are **SQL operations running in some other thread too**? That should ideally be avoided IMHO - as the Android SQLite library docs you linked seem to indicate (starting from _"Session objects are not thread-safe"_ and below).

westnordost:
Wow, awesome! You did it!

Unfortunately I won't be able to review this until sometime next week probably.

One comment from scrolling through it vertically: The API of the old wrapper interface was basically a copy of the Android one, because that's what was exposed on Android.
The new interface doesn't *need* to copy that as long as it remains well readable.

Don't know regarding transactions. Did you check the syntax for it in the sqlite docs?

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,278 @@
...
+                require(values.size == columnNames.size)
+                statement.bindAll(values)
+                val rowId = statement.executeInsert()
+                result.add(rowId)
+                statement.clearBindings()
+                statement.reset()
+            }
+            statement.close()
+        }
+        return result
```
westnordost, line 105: By the way, this can potentially be optimized (quite a bit?).

SQLite allows a syntax like this:
```SQL
INSERT INTO my_table (column_a, column_b, column_c)
VALUES (1,2,"a"), (2,4,"b"), (3,1,"c"), (4,5,"d")
```
to insert 4 values. I.e. no need for a transaction, no need for repeating the first line x times.

To be honest, I thought I've already long done this, but apparently not.

https://sqlite.org/lang_insert.html

westnordost: (But before blindly doing a premature optimization, it might be worth it to confirm that this would actually be faster in some unit test or something. Maybe binding like tens of thousands of parameters to a prepared statement with tens of thousands placeholders comes with implications or limitations in itself. Also, it needn't be part of this PR.)

westnordost:
**Any news here?** Or rather, do you intend to follow this through?

I imagine some `transactions` are not really necessary, we only need transactions strictly for data integrity. Otherwise, when piecing together data, a consistent result could maybe be guaranteed on the `*Controller` layer.
Generally, though, I believe the integrity of tables that use a foreign key to link to a key in another table should be guaranteed on the database layer because I am not sure if queries that involve _foreign keys_ to non-existing rows in another table (e.g. UNION on something or whatever) would not already go haywire (i.e. cause an exception).

In case the above is not viable, any stuff that needs transactions to guarantee database integrity should go into one DAO, so for strongly interconnected tables, we have one DAO that actually manages multiple tables. Plus, of course, any queries involving transactions, i.e. multiple queries, the queries need to be rewritten probably with `Database.rawQuery` (or `Da.exec`).

FloEdelmann:
I _finally_ got around to looking into this PR, and I think I figured it out :slightly_smiling_face: 

Executing `BEGIN TRANSACTION;` on its own is fine and works as expected (a transaction is started), but multiple concurrent database accesses can bring the transactions out of sync. The old Android database layer had concurrent access linearization, I now rebuilt something similar with `ReentrantLock` (to allow nested transactions).

I rebased and force-pushed to make reviewing easier; the first commit is basically the one you already reviewed, just based upon the current master branch (i.e. DB instantiation in `AndroidModule` instead of `DbModule`). The other commits are individual fixes.

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,278 @@
...
+    override fun getBlobOrNull(columnName: String): ByteArray? = statement.getBlobOrNull(index(columnName))
+    override fun getStringOrNull(columnName: String): String? = statement.getTextOrNull(index(columnName))
+
+    private val columnNames by lazy { statement.getColumnNames() }
+
+    private fun index(columnName: String): Int {
+        val index = columnNames.indexOf(columnName)
+        require(index != -1) { "Column $columnName not found" }
+        return index
+    }
```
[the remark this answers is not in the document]
FloEdelmann: Done in [`b68488e` (this PR)](https://github.com/streetcomplete/StreetComplete/pull/6202/commits/b68488e4c2511dd5cd573f5f87d99056a0147b84)

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,278 @@
...
+    having: String?,
+    orderBy: String?,
+    limit: Int?,
+): SQLiteStatement {
+    require (having.isNullOrBlank() || !groupBy.isNullOrBlank()) {
+        "`having` clauses are only permitted when using a `groupBy clause"
+    }
+
+    val columnNames = if (columns.isNullOrEmpty()) "*" else columns.joinToString(", ")
+    var sql = "SELECT"
```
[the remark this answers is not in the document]
FloEdelmann: Done in [`10fc41a` (this PR)](https://github.com/streetcomplete/StreetComplete/pull/6202/commits/10fc41aabe23ab34de3c7480ede2b38bc5fa94ff)

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,278 @@
...
+    ): Int =
+        databaseConnection.prepareUpdate(table, values, where, conflictAlgorithm).use { statement ->
+            statement.bindAll(args)
+            statement.toSequence { }.count()
+        }
+
+    override fun delete(table: String, where: String?, args: Array<Any>?): Int =
+        databaseConnection.prepareDelete(table, where).use { statement ->
+            statement.bindAll(args)
+            statement.toSequence { }.count()
```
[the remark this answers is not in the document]
FloEdelmann: Fixed in [`bfa723c` (this PR)](https://github.com/streetcomplete/StreetComplete/pull/6202/commits/bfa723cca5ffec949e467f3fee96f63fe8f274c0). We do need the number of updated rows (or at least if any rows have been updated) in `NoteEditsDao`'s `markSynced` function. `RETURNING 1` works fine for that.

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,278 @@
...
+    override fun update(
+        table: String,
+        values: Collection<Pair<String, Any?>>,
+        where: String?,
+        args: Array<Any>?,
+        conflictAlgorithm: ConflictAlgorithm?,
+    ): Int =
+        databaseConnection.prepareUpdate(table, values, where, conflictAlgorithm).use { statement ->
+            statement.bindAll(args)
+            statement.toSequence { }.count()
```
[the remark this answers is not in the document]
FloEdelmann: See above, fixed in [`bfa723c` (this PR)](https://github.com/streetcomplete/StreetComplete/pull/6202/commits/bfa723cca5ffec949e467f3fee96f63fe8f274c0)

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
FloEdelmann: See https://github.com/streetcomplete/StreetComplete/pull/6202#issuecomment-4071196999: Fixed in [`57bd664` (this PR)](https://github.com/streetcomplete/StreetComplete/pull/6202/commits/57bd664c1d7180b0d373e581fb9f09b64606002e) and [`b3a74f3` (this PR)](https://github.com/streetcomplete/StreetComplete/pull/6202/commits/b3a74f3ad6963402c018b63537d801a40c28f65d).

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
```diff
@@ -0,0 +1,301 @@
...
+import de.westnordost.streetcomplete.data.ConflictAlgorithm.ROLLBACK
+import kotlinx.atomicfu.locks.ReentrantLock
+import kotlinx.atomicfu.locks.withLock
+
+class StreetCompleteDatabase(private val databaseConnection: SQLiteConnection) : Database {
+    private val lock = ReentrantLock()
+    private var transactionDepth = 0
+
+    init {
+        databaseConnection.execSQL("PRAGMA journal_mode=WAL")
```
westnordost: A comment as to why this is necessary is necessary. Related page from the documentation:

https://www.sqlite.org/isolation.html

I read a bit about this.

Apparently, all intermediate changes made within one database transaction are actually visible for all readers on the same database connection. We use just one database connection. So, basically the very thing that (we hoped) a transaction should do, guarantee data consistency at all times (e.g. there are no elements without geometry) is actually not guaranteed currently, ***unless*** this is prevented by Android database driver in one way or another.

A claim made in some blog article that intermediate changes made within one transaction would not be visible for all readers on the same database connection by turning WAL mode on isn't supported by the documentation. (To the contrary, it even says that when an UPDATE runs in parallel to a SELECT, that the data returned by the SELECT may contain old data, new data or both, i.e. it is undefined.)

Do you have more info on this matter?

(This is kind of important. Because if transactions already don't do what we hoped they do, and it did not lead to obvious problems, maybe we don't need them at all, then. Or, at least, in places where we need to make guarantees about data consistency, e.g. every element has a geometry, guarantee it programmatically *after* the query is done)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/DatabaseImpl.kt
```diff
@@ -0,0 +1,301 @@
...
+    ): List<T> = lock.withLock {
+        databaseConnection.prepareQuery(distinct, table, columns, where, groupBy, having, orderBy, limit).use { statement ->
+            statement.bindAll(args)
+            statement.toSequence(transform).toList()
+        }
+    }
+
+    override fun insert(
+        table: String,
+        values: Collection<Pair<String, Any?>>,
```
westnordost, line 68: (If we have to pick apart keys and values anyway, might just as well make the interface consistent to `insertMany` and have `columns` and `values` as separate parameters. Same for `update`)

westnordost (commented):
Oh?

Hm. My mind was already set on that it would be necessary to either check if we need transactions spanning multiple queries at all, or to put any transactions we might need into **one** sql statement.

Let me try to wrap my head around this.

So if `BEGIN TRANSACTION;` executes just fine and opens a transaction and then, in code, submit more queries to sqlite, like that,

```pseudocode
db.startTransaction()
db.query(…)
db.query(…)
db.endTransaction()
```

some other part of the program code might also call `db.query()` from a different thread and thus cause that query to be lumped in together with the transaction that was started by a completely unrelated DAO (as the database runs concurrent to the program code, of course). Worse yet, it is completely possible that another DAO explicitly starts *another* transaction, and latest then this should result in an error. (This error you already encountered, IIRC.) That resulting behavior is what you describe as "transactions out of sync", right?

The obvious solution for me would be to do everything for one transaction in one SQL statement
```pseudocode
db.query("""
BEGIN TRANSACTION;
…
…
COMMIT;
""")
```

And, I would have expected, that the Android SQLite database layer would have done that under the hood. But, you checked this, and actually it does something else? Does it truly linearize all access to the database? That sounds... uh, that is has horrible consequences for performance: *Any* write operation will block *any* read operation? (Or already did in Android's SQLite database layer?)

Does the preparing of the statements also have to be in the synchronized section? (Or, for that matter, everything except `statement.execute*`)

---

Notes/thoughts to myself (can refactor later):

- rename `Database` interface to `DatabaseConnection`
 
- rename `StreetCompleteDatabase` to `SQLiteDatabaseConnection` (or similar) and upgrade + create logic should move out of this class (a `SQLiteDatabaseConnection` should be the *result* of a database (connection) initialization operation, e.g. `DatabaseInitializer.initialize(sqliteConnection) : DatabaseConnection`

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
mnalis: > We use just one database connection

Note that while I do not have Android experience specifically, generally with SQL databases if you use different (i.e. parallel) contexts / threads / etc. you should use **different** (multiple) database handles for each of them if you intend for their transactions to be separate.

ref: https://sqlite.org/isolation.html

westnordost: Actually, for proper encapsulation, we'd need to use one SQLite connection for _every_ operation. E.g. getting OSM data by bounding box must be done with one SQLite connection, writing all those OSM elements and their geometries  just downloaded to the database must be done with another SQLite connection. Otherwise, the getting of the bbox data may include partly updated and partly not-updated data.

However, there are a few question marks here:

- The SQLite connections would need to be managed by the *Controller classes, though, because one operation might include updating several tables at once, see e.g. [MapDataController](https://github.com/streetcomplete/StreetComplete/blob/ff841d25ac998f8eeddef8b4f2b94e4db144c1c2/app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataController.kt#L120-L124) and these should be in the same transaction / same connection

- What would be the performance overhead for creating and closing a new sqlite connection for each operation? Is it greater than serializing all access to the database?

- Using the standard [Android SQLite API](https://developer.android.com/training/data-storage/sqlite), which is of course also used under the hood at least when using androidx.sqlite:sqlite-bundled, is it actually *possible* to use more than one SQLite connection? The documentation sounds like [SQLiteOpenHelper#getWritableDatabase()](https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper#getWritableDatabase()) will always return the same connection

westnordost: And this is why the next step is to find out what is actually the status quo. I.e. what exactly does the current Android SQLite API do (because it works and doesn't create obvious problems). We then have the option to mirror that behavior, or do something else, in the hope of improving performance through parallelization.

westnordost:
Dug around in `android.database.sqlite` source code a bit.

Each call to SQLite (`transaction`, `query`, `update`, ...) in `SQLiteDatabase` is done via `SQLiteSession`, of which there is one for each thread. Each `SQLiteSession` has a `SQLiteConnectionPool` which lets it wait for and (eventually) hands out `SQLiteConnection`s to the session. There is definitely some mutex logic involved, although it is quite complex and I didn't understand it when reading it roughly.

Anyway, what we know now is that in `android.database.sqlite` there is not just one SQLite connection, but many, so I'd say that my previous surmise that either access to the SQLite database is fully serialized or intermediate data within one transaction is fully visible to any other parallel operations (as is expected for operations that use the same SQLite connection) is very likely wrong.
Hence, serializing all access to the DB as done in this PR now would likely lead to distinct performance issues.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
westnordost: So, summary on status quo:

Access to the database on the same connection is serialized. Different threads will use different connections (probably; it is a lot of source code to read through).

westnordost:
After some further research I come to the following intermediate conclusions:

**TLDR**: Maybe do nothing for now, hoping a solution will fall from the sky. The approach of this PR seems to be the most prudent one, even if it may have performance implications.

### Room

Using Room would require to rewrite all DAOs from scratch. Also, if this was done, it should be done after the migration to Compose + ViewModels is complete, because methods on Room DAOs all are `suspend`ing and (can) return `Flow`s which is a different approach from what we use now (blocking `*Dao`s internal to `*Controller`s which implement observer pattern).

So, this would further encompass a refactor of the `*Controller` classes and then, as well, all users of that. (However, that last part would remove a lot of boiler plate code, actually).

In summary, a lot of work, likely to break eggs due to making omelet, and would need to be done at the end. Personally, I don't like all that compiler plugin magic (i.e. code generation behind the curtains), but hey, we use Compose, lol.

### Simple approach

SQLite only allows one SQLiteConnection per thread ([see SQLite documentation on _Multi-thread_](https://www.sqlite.org/threadsafe.html)). So, the approach in this PR is correct. (Except that the connection is apparently never closed.)

Instead of using just one SQLiteConnection for everything and serialize everything, we could use one SQLiteConnection _per thread_ so that we can off-load long-running SQLite operations such as saving all the downloaded OSM data into a separate connection and thus don't block other queries (e.g. reads) in the meantime. Java's `ThreadLocal<T>` doesn't exist on KMP however, there is a replacement, but it is more unwieldy to use (IMO). Also, since Kotlin coroutines are not threads, two subsequent calls within the same `suspend` function _could_ actually run in different threads, meaning they would be executed on different connections, which could mess with expectations, especially in context of transactions.

Also, question is if this would actually improve things. The biggest bottleneck is when the OSM data is being downloaded: All the OSM data is saved to DB, and in that time, users can not solve any quests (the UI appears as if frozen) because we already deliberately don't allow parallel updating of OSM data from two different sources (download vs user solved a quest) on the `*Controller` layer.

app/src/main/java/de/westnordost/streetcomplete/data/StreetCompleteDatabase.kt
westnordost: So I read a bit up about this. The binding of thousands of parameters

```
INSERT INTO my_table (column_a, column_b, column_c)
VALUES (?, ?, ?), (?, ?, ?), (?, ?, ?), (?, ?, ?) // , ....
```

 is exactly the reason why this wouldn't work for the general case. There is a maximum allowed variable count.

westnordost (approved):

westnordost:
> Except that the connection is apparently never closed

By the way, this is fine. Android closes any open file handles on exit of an application automatically. iOS too. There is not even a callback to the app so that it *could* clean up when the app is terminated.

FloEdelmann:
Thanks for finishing this up!

westnordost:
(Just assigned you to have a nice overview for attribution in the https://github.com/orgs/streetcomplete/projects/1/views/1 board)

PR #6205, opened 2025-04-11
Move quest filtering logic to ViewModel

irakliy01:
A small refactoring on quest filtering to improve performance. Now filtering is done in the ViewModel. Also introduced `ResourceProvider`, useful for accessing resources from non-UI layers/components.


mnalis:
> Now filtering is done in the ViewModel [...]

Thanks for the effort! But, I noticed however there was this comment in the code:

> // the filtering is not done in the view model because it involves accessing localized
> // resources, which we consider UI (framework) specific data and view models should be
> // free of that.
> // NOTE: This is very slow though, it involves getting the string resource, lowercasing it

Perhaps I'm misunderstanding it, but it seems to be saying that filtering in view model is unwanted (even if it brings performance improvements) for design decisions? Or does this PR address that concern somehow?

irakliy01:
To avoid the ViewModel directly accessing resources , I introduced the ResourceProvider abstraction (interface). The `QuestSelectionViewModel` now depends on the ResourceProvider interface to fetch the necessary quest titles. It doesn't know about context or specific android resource mechanisms. The `DefaultResourceProvider` implementation encapsulates the logic for actually retrieving the strings, including handling the correct application locale to ensure the ViewModel receives the properly localized titles.
So, the key idea is that while the filtering operation happens in the ViewModel for performance, the access to localized resources needed for that operation is handled through an abstraction (`ResourceProvider`), maintaining the separation of concerns and keeping the ViewModel decoupled from direct android framework dependencies like context for resource loading. This felt like a standard way to approach it, but let me know if you have concerns or if it doesn't quite align with your vision for the app's architecture.

app/src/main/java/de/westnordost/streetcomplete/util/ResourceProvider.kt
westnordost, line 1: Note / FYI:

Sooner or later, we want to migrate to [Compose Resources](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-multiplatform-resources.html) ([`org.jetbrains.compose.components`](https://central.sonatype.com/artifact/org.jetbrains.compose.components/components-resources/overview)), I hope this class won't make it more difficult to migrate.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
```diff
@@ -46,6 +59,11 @@ class QuestSelectionViewModelImpl(
     prefs: Preferences,
 ) : QuestSelectionViewModel() {
 
+    private val _searchText = MutableStateFlow(TextFieldValue())
+    override val searchText: StateFlow<TextFieldValue> = _searchText
```
westnordost, line 63:
```suggestion
    override val searchText = MutableStateFlow(TextFieldValue())
```
(etc.)

...would be more consistent with the rest of the codebase. I.e. the implementation uses MutableStateFlow while the interface just exposes a StateFlow. So, I don't need to effectively duplicate all StateFlow fields (one private MutableStateFlow, one public StateFlow).

app/src/main/java/de/westnordost/streetcomplete/data/quest/QuestTypeRegistry.kt
```diff
@@ -9,4 +9,4 @@ import de.westnordost.streetcomplete.data.ObjectTypeRegistry
  * It is also used to define a (display) order of the quest types and to assign an ordinal to each
  * quest type for serialization.
  */
-class QuestTypeRegistry(ordinalsAndEntries: List<Pair<Int, QuestType>>) : ObjectTypeRegistry<QuestType>(ordinalsAndEntries)
+class QuestTypeRegistry(val ordinalsAndEntries: List<Pair<Int, QuestType>>) : ObjectTypeRegistry<QuestType>(ordinalsAndEntries)
```
westnordost, line 12: No need to expose this this parameter. `ObjectTypeRegistry<T>` already inherits from `AbstractList<T>`, so you can access the contents directly.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
```diff
@@ -107,6 +131,13 @@ class QuestSelectionViewModelImpl(
         }
     }
 
+    private fun loadQuestTitles() {
```
westnordost, line 134: Can this be made to update when the language changed (i.e. either the system language changed or the app setting changed)? If not (with reasonable effort), maybe add a short comment that explains this caveat.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
```diff
@@ -86,6 +104,11 @@ class QuestSelectionViewModelImpl(
 
     override val quests = MutableStateFlow<List<QuestSelection>>(emptyList())
 
+    override val filteredQuests: StateFlow<List<QuestSelection>> =
+        combine(quests, searchText, questTitles) { quests, searchText, titles ->
+            filterQuests(quests, searchText.text, titles)
+        }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), emptyList())
```
westnordost, line 110: What is this `SharingStarted.WhileSubscribed(5000)` for? 🤔 Honest question, I don't understand.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionList.kt
```diff
@@ -88,7 +88,22 @@ fun QuestSelectionList(
                 bottom = contentPadding.calculateBottomPadding()
             ),
         ) {
-            itemsIndexed(reorderableItems, key = { _, it -> it.questType.name }) { index, item ->
+            itemsIndexed(
+                reorderableItems,
+                key = { _, it -> it.questType.name },
+                contentType = { _, _ -> "QuestSelectionItem" },
```
westnordost, line 94: this is not necessary, as far as I know, because all items have the same content type (`null` if not specified), it makes no difference here.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionList.kt
```diff
@@ -88,7 +88,22 @@ fun QuestSelectionList(
...
+
+                val rememberedOnToggleSelection = remember(item.questType) {
+                    { isSelected: Boolean ->
+                        if (isSelected && item.questType.defaultDisabledMessage != 0) {
+                            showEnableQuestDialog = item.questType
+                        } else {
+                            onSelectQuest(item.questType, isSelected)
+                        }
+                    }
+                }
```
westnordost, line 105: Since the introduction of [strong skipping](https://developer.android.com/develop/ui/compose/performance/stability/strongskipping#lambda-memoization), it is not necessary anymore to wrap lambdas in a `remember` in order to avoid the lambda being re-created on every recomposition (and thus causing the composable to which the lambda is passed to also recompose).

It is done automatically.

westnordost (commented):
I think yours is a pretty swift solution! I didn't even think of that! In general, you've made it so that there are no direct dependencies of the view model to the UI parts. I don't quite like that `TextFieldValue` is passed to the ViewModel, but it looks like it is just a pure data class, so I guess that's okay. (Couldn't it just be a String, though?)

Say, how did you spot this in the code? Were you familiarizing yourself with the codebase, or did you notice slowdowns yourself and started investigating as to the cause of this?

Note that `ShowQuestFormsScreen` the exact same quest filtering as in the quest selection setting screen. It would make sense if those two places would work consistently (But this needn't be done in this same PR).

irakliy01:
> I don't quite like that `TextFieldValue` is passed to the ViewModel, but it looks like it is just a pure data class, so I guess that's okay. (Couldn't it just be a String, though?)

Yeah, I agree that a String would be better here than TextFieldValue. I'll fix that in the next commit. 

> Say, how did you spot this in the code? Were you familiarizing yourself with the codebase, or did you notice slowdowns yourself and started investigating as to the cause of this?

I wanted to add a country-specific presets feature (before I found `enabledInCountries` property), since in my country, some quests are not relevant. And that's when I stumbled upon this code and decided to refactor it. Refactor was motivated by fixing that potential performance bottleneck and improving the architecture by introducing the `ResourceProvider`.

app/src/main/java/de/westnordost/streetcomplete/util/ResourceProvider.kt
irakliy01: While I haven't worked directly with KMM before, I checked the [documentation on accessing resources outside composables](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-multiplatform-resources-usage.html#strings) you mentioned. It looks like they provide ways to do that, which suggests the core functionality needed by the ViewModel should still be possible.

My hope is that introducing the `ResourceProvider` interface might actually help with the migration. Since the ViewModel depends on the interface, the `DefaultResourceProvider` (which uses Android Context) could potentially be swapped out later with a new implementation that uses the Compose Resources API, without needing major changes in the ViewModel itself.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
irakliy01: Here's how the `ResourceProvider` in this PR handles language changes:

1. It uses an internal cache for the locale-specific context to improve performance for repeated calls.
2. It listens for changes to the app's language preference (`prefs.onLanguageChanged`).
3. When the language preference changes, this listener invalidates the internal cache.
4. The next time `getString` is called after the cache invalidation, it will recreate the context using the new locale from the preference and fetch the correct string.

So, it does update based on the app's language setting changing, but it requires the preference update and a subsequent call to `getString` to see the effect, rather than instantly pushing an update.

The fully seamless, automatic update behavior (where the application context itself might inherently reflect the correct locale immediately after any change, system or in-app) is what the feature proposed in #6211 aims to implement more comprehensively. This PR's `ResourceProvider` is designed to be compatible with and benefit from that future change.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
irakliy01: I'm just used to it since not all projects have their ViewModel scopes aligned with ui lifecycle. Because of some mistake viewModel won't cleanup in time and flow hangs. WhileSubscribed is safer solution (and also [recommended by Google](https://developer.android.com/topic/architecture/recommendations#viewmodel)) since it stops flow after the last subscriber unsubscribes plus grace period (5 sec for config changes). But in this project everything works fine with Lazily since it uses androidx compose navigation, so I'll replace it in the next commit.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
irakliy01: > either the system language changed

Oh, I've skipped that line. Will do something about it

irakliy01:
I can wait for #6210 to update deprecated `configuration.locale` to `configuration.locales`

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
westnordost: Well, I am not desperate. I don't cast :-)

westnordost: Also, "desperate" is not the word I'd use for devs that break the interface by casting (instead of doing it properly). ;-)

app/src/main/java/de/westnordost/streetcomplete/screens/settings/debug/ShowQuestFormsScreen.kt
```diff
@@ -54,18 +57,23 @@ fun ShowQuestFormsScreen(
...
-    var searchText by remember { mutableStateOf(TextFieldValue()) }
+    val searchText by viewModel.searchText.collectAsStateWithLifecycle()
+
+    val filteredQuests by viewModel.filteredQuests.collectAsStateWithLifecycle()
+
+    val configuration = LocalConfiguration.current
+
+    LaunchedEffect(configuration.locale.toString()) {
+        viewModel.onConfigurationChanged()
+    }
```
westnordost, line 68: That's a clever solution. From a data flow point of view, however, it's upside down. It's not the job of the view to notify the viewmodel that some (non-ui) data changed. Rather, the flow should be that if the language changes, the view model updates its `filteredQuests` (because the `questTitles` were updated) and thus causes the UI to recompose.

But I reckon that maybe (currently) it is not possible to get notified of when the language changed otherwise. In that case, however, I'd slightly prefer to not update the quest titles at all (and add a short comment that explains that it is not done) rather than a non-optimal solution. After all, this case doesn't really happen as it would require the user to change the system language while he is in that screen (and he is searching for something). And even if it happens, exiting the screen and reentering would fix it (right?).

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionTopAppBar.kt
```diff
@@ -79,7 +81,8 @@ import de.westnordost.streetcomplete.ui.common.dialogs.ConfirmationDialog
                 colors = TextFieldDefaults.textFieldColors(
                     textColor = MaterialTheme.colors.onSurface,
                     backgroundColor = MaterialTheme.colors.surface
-                )
+                ),
+                keyboardOptions = KeyboardOptions(hintLocales = LocaleList.current),
```
westnordost, line 85: Why is this necessary? Isn't that the default? After all, if it was not the default, then basically it would make sense to add this line into every single input field, no?

westnordost (commented):
> But I'll have to introduce my own version of TextFieldValue data class for cursor location to work.

I think the cursor will be fine. Only on configuration change, the cursor location wouldn't be remembered. Otherwise it should work just as fine with a String. FWIW, `TextFieldValue` is available in common, i.e. no need to create an own class to make it work in multiplatform.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionTopAppBar.kt
irakliy01: It changes input language to the language of the app. I thought it would be convenient to the user since it's the only language in which he can perform search in these screens.

irakliy01: https://github.com/user-attachments/assets/4e68c0d8-25cb-4f1f-bdd3-4acfcfd5e4aa

westnordost: And without the hint, keyboards would generally ignore the languages set (programmatically) in the LocaleList but use the system language(s) instead?

It would make sense to me if the currently active Locale of the app is handed to the keyboard software, but I don't know if Android actually does this.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
irakliy01: I meant in big projects sometimes managers push devs to make quick fixes. In those cases developers forget about clean code and start making mess in a project. Backing fields, Api/Impl and other patterns/techniques are meant to fight with that.

app/src/main/java/de/westnordost/streetcomplete/screens/settings/debug/ShowQuestFormsScreen.kt
westnordost: > Even in KMM you'll have to notify ViewModel from the UI layer about those events happening in Android/iOS. There isn't a clean, built-in mechanism for a ViewModel to directly observe these system-level events (or at least I'm not aware of it).

Well, there must be, otherwise the UI would also not be notified of it via Compose. But true, maybe it is not exposed on a public API.

> Basically, all UI events are passed to ViewModel from the UI layer

Of course, but changing of the system language is not a UI event. It's as much of a UI event as for example internet connection state having changed or location status having changed.

> So, there are two options

You know, I am not hellbent on it, so if _you_ have a strong preference, put it as you like. As for myself, I'd slightly prefer to remove the `LaunchedEffect` (and add a short comment in the viewmodel that questtitles are not updated because there is no listenable callback from the system for when the language changes) because I deem it not worth it to break the ideal data flow to cover an edge case like this. (Broken window, etc.)

app/src/main/java/de/westnordost/streetcomplete/screens/settings/quest_selection/QuestSelectionViewModel.kt
westnordost: Yeah, I know what you mean. ...And it is only a matter of time until such botch comes back haunting the developers, oftentimes not even hitting the dabblers but those that come after... 😬 Hrr, memories, that gives me the creeps. 

(So, maybe the pattern you use is generally preferable, as a guard against such practices. But well, being kind of the BDFL in this project, I won't allow botch in the first place, so such safeguard is less necessary. Readability and consistency is higher on my list.)

app/src/main/java/de/westnordost/streetcomplete/screens/settings/debug/ShowQuestFormsScreen.kt
irakliy01: System language is stored in configuration within context. When config change occurs app receives new context which is stored in activity. So you have to somehow send the data from activity/fragment/composable to the place that will process it.
I don't have any preferences)
I removed this code and added a comment

westnordost (approved):
Perfect.

By the way, `collectAsStateWithLifecycle()` is Android-only. But I can also change that later when I actually start migrating to Compose Multiplatform.

westnordost:
> I wanted to add a country-specific presets feature (before I found enabledInCountries property), since in my country, some quests are not relevant. 

Anyway, for quests that are not only not relevant but even useless (because the thing is always X or never Y, e.g. if there are no tactile pavings on crosswalks *at all*, if there are no cycleways _at all_, etc.), you are invited to create a PR to disable such quests for Georgia.

Issue #6206, opened 2025-04-14
Clarify count for bicycle parking of type wheelbenders


mnalis:
**Use case**
SC currently asks for count of bicycle parking spaces on all parking types. 

While that answer is exactly known on  most bicycle parking types (e.g. stands, lockers, handlebar holders, two-tier racks, ...) it is **ambiguous and approximate mental calculation on "wall loops"** (wheelbenders), which many (esp. non-cyclists) are likely to get quite wrong, as e.g. in this (unfortunately all too common in Croatia) picture example:

![Image](https://github.com/user-attachments/assets/f6ec6fbb-0887-4cf3-9ec4-73aa4ee615e3)

(Prompted by recent confusion reported in Croatian OSM telegram group; also previously  in https://github.com/streetcomplete/StreetComplete/discussions/3704)

**Proposed Solution**
- extend the explanation in info button how to handle that situation (currently it only talks about bicycle stands: _"Note that you can usually park one bicycle on each side of a stand."_ (this option should be less controversial and still help _some_ users)

- alternatively, do not ask for count on `bicycle_parking=wall_loops` as it fails at least _"🐿️ Easy answer: Users are out and about and impatient. A quick, straightforward and clear answer must be possible"_ (which is how I've worked around the problems for myself in SCEE ages ago and how I personally prefer it, but not everybody seems to agree that it is ideal solution and prefer approximation instead). 
  We [currently skip the quest](https://github.com/streetcomplete/StreetComplete/blob/cf5c89c54b760789bae9e2a7d622d2ef7653807c/app/src/main/java/de/westnordost/streetcomplete/quests/bike_parking_capacity/AddBikeParkingCapacity.kt#L19) for `floor` and `informal` which are ambiguous too.

westnordost:
Don't be stressed out. The value of `capacity` for bicycle stands lies in its statistical value, i.e. when looked at in a heatmap or when compared to the population density by area. In short, if you over- or under-guess the number of bicycles that can be parked in the above picture wrong, this is not severe at all.

I'd guess 3-4

rusty-snake:
> While that answer is exactly known on most bicycle parking types (e.g. stands, …) it is ambiguous … "wall loops" …

[...]

PR #6209, opened 2025-04-15
Update "Bicycle type" Quest hint to explain what to do with wheelbenders

mnalis:
Fixes https://github.com/streetcomplete/StreetComplete/issues/6206


westnordost:
Ok, whatever, fair enough.

PR #6212, opened 2025-04-18
use other icon for street name

mnalis:
As accepted in https://github.com/Helium314/SCEE/issues/771#issuecomment-2809027643

tested and debug apk at https://github.com/mnalis/StreetComplete/actions/runs/14536267152


westnordost:
The background of `ic_quest_housenumber_street` is grey. It should be yellow for quests that are about streets.

Issue #6215, opened 2025-04-20
Parking access quest: too general, doesn't respect access tags


wolfbert:
A surface parking lot private to a residential apartment block is tagged as 
`amenity=parking parking=surface vehicle=private` with the service highway as `highway=service service=parking_aisle vehicle=private`. This is as signed (vehicle=no except tenants), and you can walk through to the apartments. 

SC offers the parking space for tagging and tags it as `access=private` (option 3 only persons with permit), which is incorrect (apart from not retagging the highway, which is then out of sync).

**How to Reproduce**
See above.

**Expected Behavior**
Exclude the parking area from the quest.

**Versions affected**
v60.3 on Android 14

matkoniecz:
>  which is incorrect

not really

`access=` on parking is understood to refer for parking

if going through parking would be incorrect then it would be applied to `highway=` lines

though skipping such parkings may make sense anyway

mnalis:
> though skipping such parkings may make sense anyway

Yes, see https://github.com/streetcomplete/StreetComplete/issues/4538#issuecomment-1288187859

wolfbert:
> access= on parking is understood to refer for parking

I wasn't aware of that, so may change that. In Austria, there are 155 cases like this, not a lot, but should be changed only if you know what you are doing (SC users don't, they can neither see which tags already exist nor which will be set).

@mnalis Thanks for the quick fix.

matkoniecz:
> SC users don't, they can neither see which tags already exist nor which will be set

note: if you want you an see used tagging (start undo without confirming yes)

though responsibility for correctly tagging what user answered is on SC (and therefore StreetComplete developers)

PR #6216, opened 2025-04-20
parking - skip (few more) detailed vehicles types

mnalis:
Fixes https://github.com/streetcomplete/StreetComplete/issues/6215

We are already skipping the quest for majority of more detailed access tags in https://github.com/streetcomplete/StreetComplete/pull/4547 (as accepted in https://github.com/streetcomplete/StreetComplete/issues/4538#issuecomment-1288187859), but few were missing or incorrect.

This followup PR fixes that


mnalis:
Compiles OK with debug `.apk` [available here](https://github.com/mnalis/StreetComplete/actions/runs/14560309579); and  tested on [way #33771841](https://www.openstreetmap.org/way/33771841)

PR #6217, opened 2025-04-20
update icon for street name quest (#2)

mnalis:
But with yellow background, as requested in https://github.com/streetcomplete/StreetComplete/pull/6212#issuecomment-2816019808 (originally https://github.com/Helium314/SCEE/issues/771#issuecomment-2809027643)

Tested and debug .apk [here](https://github.com/mnalis/StreetComplete/actions/runs/14561734274). Screenshot:

small_Screenshot_20250420_193522_StreetComplete Dev


westnordost:
You need to update the source, too. The SVG

mnalis:
> And I very much prefer if you don't overwrite the old SVG file

OK, done.

westnordost:
Can you rename the xml accordingly? IIRC if it is not named the same, it's going to be a problem for some script of @FloEdelmann , IIRC

westnordost (approved):

PR #6229, opened 2025-04-27
Demote baby changing table and wheelchair access toilets

peternewman:
Now they are asked for things like restaurants and cafes too, but you generally have to go inside them to answer (and likely use the toilets), compared to say dietary quests where a menu is often posted outside, or sometimes vegetarian/vegan/halal/GF may be an advertised selling point too.

Also below shop payment quests as everyone who uses an amenity will pay (hopefully!) but only some people will use the toilet.


westnordost:
LGTM

PR #6230, opened 2025-04-27
Improve quest form opening performance

### LINKED ISSUE #5813: Jitter when opening the quest «What is width of opening here?»
**How to Reproduce**
Find the quest "What is width of opening here?" on the map and open it

When zooming in, the animation is ragged. When opening other point quests, the map zooms in smoothly

https://github.com/user-attachments/assets/da266257-c8df-44f5-afe6-6e1e021f37df

https://www.openstreetmap.org/node/3359945255

**Versions affected**
v59.0-alpha1
Android 14 OneUI 6.1 
Samsung A52

p.s. Congratulations on the long-awaited migration to MapLibre!

irakliy01:
Addresses UI freezes (~500ms+) that occurred when opening quest forms.

This was caused by synchronous loading during view creation for:

- `FeatureDictionary` data (used for the title hint label)
- `CountryInfo` data

Changes:

- Loads the title hint label asynchronously in `AbstractOsmQuestForm` (note: the hint is briefly hidden on the UI while loading).
- Pre-fetches `CountryInfo` asynchronously in `MainActivity` and passes it as a fragment argument.

Partially fixes #5813. This PR lowers the main thread work for form view creation to ~6-8ms based on logs. However, some minor jitter is still visible during the map zoom animation, potentially related to the animation itself, layout complexity or quest form opening logic.

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/data/meta/CountryInfo.kt
@@ -84,6 +84,7 @@ data class IncompleteCountryInfo(
     val workweekDays: Int? = null,
 )
 
+@Serializable
 data class CountryInfo(private val infos: List<IncompleteCountryInfo>) {
     val countryCode get() = infos.first().countryCode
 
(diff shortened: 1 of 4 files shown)


matkoniecz:
> Doesn't this just move the perceived unresponsiveness from the initialization of the quest form to just before opening the quest form? I.e., when the user taps on a quest, there will still be a delay.

would work be happening while opening quest form is animated? That may hide part or all of delay

westnordost:
No. Before it is opened.

irakliy01:
`countryInfo` and `titleHintLabel` are now computed on background thread. This leads to better ui performance since it's not busy calculating those data
You can run a profiler before changes and after changes. You'll see, there won't be janky frames during quest form opening.

> I find puzzling that by the time the user opens a quest form, the country boundaries / country infos are still not loaded yet. They _should_ be loaded already on application startup in the background, and be done in 1 second or so. That's hardly enough time for the user to open a quest.

I can do country info loading on app startup if needed.

Issue #6240, opened 2025-05-02
reverse positions of download/upload in 61.0-beta1


mnalis:
**Use case**
v61.0-beta1 moves upload button into main menu. However, it takes the space (directly below big square icons) which was previously used by (often used) download button, which breaks muscle memory.

**Proposed Solution**

It would be nicer if download button would retain its previous place just below  the the big square icons, and (new) upload button was below it. (IOW, if download and upload buttons changed places.)

![Image](https://github.com/user-attachments/assets/a8e6643b-abfa-4a94-b882-7c7e9c129ad3)

westnordost:
makes sense

HolgerJeromin:
UP-load above DOWN-load is imo better.

westnordost:
Okay... I don't really mind either way.

mnalis:
> UP-load above DOWN-load is imo better.

Well, even ISPs list bandwidth as DL/UL (showing more often requested thing first)... as does e.g. JOSM first show download and then upload button

PR #6241, opened 2025-05-02
reverse order of upload and download buttons

mnalis:
(to preserve the muscle memory)
Fixes https://github.com/streetcomplete/StreetComplete/issues/6240


westnordost:
hint: the change you have to make is in MainMenuDialog.kt

mnalis:
compiled and tested [here](https://github.com/mnalis/StreetComplete/actions/runs/14802592604):

small_Screenshot_20250502_222141_StreetComplete Dev

westnordost:
Would you revert your first commit?

mnalis:
> Would you revert your first commit?

Sure, done.

Issue #2294, opened 2020-11-18
Clarify the tracktype quest


cyclingcat:
Hi there,

the tracktype quest (= the one with the green tractor icon) shows five short descriptions with associated pictures the user is requested to choose from. However, these descriptions sometimes do not match the ones in the German/English OSM wiki pages, thus causing some confusion. In English language, the wording **"Mostly solid"** is even used in both contexts with different tracktypes!

See this comparison (SC string first, OSM wiki label second):

**[In English:](https://wiki.openstreetmap.org/wiki/Key:tracktype#Values)**
_tracktype=grade1_: Solid / Solid [ok]
_tracktype=grade2_: **Mostly solid** / Solid but unpaved [hm...]
_tracktype=grade3:_ Mixture / **Mostly solid** [huh?]
_tracktype=grade4:_ Mostly soft / Mostly soft [ok]
_tracktype=grade5:_ Soft / Soft [ok]

**[In German:](https://wiki.openstreetmap.org/wiki/DE:Key:tracktype#Schl.C3.BCsselwerte)**
Here it's even more confusing, because the German wording in the wiki is a little inconsistent as it mixes more precise (albeit uncommon) descriptions ("Wassergebundene Decke") with very vague ones ("Einfacher Weg"/"Unbefestigter Weg"):

_tracktype=grade1_: Befestigt / Wasserfester Belag [ok]
_tracktype=grade2_: Größtenteils befestigt / Wassergebundene Decke [hm...]
_tracktype=grade3_: Gemischt / Befestigter oder ausgebesserter Weg [hm...]
_tracktype=grade4_: Größtenteils unbefestigt / Einfacher Weg [hm...]
_tracktype=grade5_: Unbefestigt / Unbefestigter Weg [ok]

[...]

PR #6242, opened 2025-05-02
Import new tracktype_grade2 pictures from the wiki to show dirt

Caerbannog:
This is similar to issue #2294 which was closed by 9dd1ddf.

## Problem
The tracktype quest asks to select between grade1 and grade5.
The picture for grade2 (Mostly solid) is not clear in my opinion. When displayed on a medium screen, I had trouble to see the distinction with grade1 (Paved). The surface color and texture looked similar to asphalt/rocks.

## Solution
This PR imports the grade2 reference picture from the wiki:
https://wiki.openstreetmap.org/wiki/Key:tracktype#Values

## Result

| Before | After |
| ----- | ----- |
| image | image |


westnordost:
I agree with both of you. Can you find a picture that uses the same perspective but clearly shows some compacted surface?

Caerbannog:
What about one of these three pictures?

https://commons.wikimedia.org/wiki/File:Track_in_Schwinkendorf_2021-09-07_21.jpg
https://commons.wikimedia.org/wiki/File:Dirt_track_leading_into_Safari.jpg
https://commons.wikimedia.org/wiki/File:Dirt_track_-_geograph.org.uk_-_7486405.jpg

Here is a mockup with the first file (color adjusted):
image

matkoniecz:
argh, sorry to complain but this one looks quite sandy to me on tiny thumbnail (note: I was similarly complaining when I was selecting pictures on my own)

curiously, it is not a problem when I look at https://commons.wikimedia.org/wiki/File:Track_in_Schwinkendorf_2021-09-07_21.jpg

and maybe it will not be a problem given that description is there?

westnordost:
How about this one? It's the same track, only at a slightly different position. Due to the shadows cast by the trees, the surface doesn't look so yellow / sandish-ly, while the grass makes it clear that it is not paved:

https://commons.wikimedia.org/wiki/File:Track_in_Schwinkendorf_2021-09-07_20.jpg

By the way, this, also the same track at a slightly different position, maybe use to replace the current grade5 pic?:

https://commons.wikimedia.org/wiki/File:Track_in_Schwinkendorf_2021-09-07_04.jpg

Caerbannog:
Great.
Here is a commit with the two links you posted.

Here is a screenshot in the Android simulator:
image

I have other things to say about the pictures (too small! hidden by text! who cares about the sky! grade1 season is confusing! irrelevant water puddles on grade3!). But at least it's a step forward.

westnordost:
Yes, about the text, I think this should be changed to text below the picture, not on top of it. For all quests. But yeah, that's a different ticket, and kind of currently blocked by a WIP PR.

Regarding "too small", I thought about being able to show a larger version of the picture by tapping on it (another time after selecting it), but that would be a major work, as that should then work for all of the pictures. It would also blow up the app size a lot, making it maybe worth it to think about not bundling those resources in the app but downloading+caching them after app install or something. Anyway, that's a bit more blue sky, and also anyway blocked by a WIP PR / near-future refactorings.

PR #6244, opened 2025-05-03
Do not ask for mtb:scale if bicycles are explicitly forbidden (or impossible)

mnalis:
compiled debug apk is [here](https://github.com/mnalis/StreetComplete/actions/runs/14812284660)

### DIFF
--- a/app/src/main/java/de/westnordost/streetcomplete/overlays/mtb_scale/MtbScaleOverlay.kt
+++ b/app/src/main/java/de/westnordost/streetcomplete/overlays/mtb_scale/MtbScaleOverlay.kt
@@ -32,6 +32,8 @@ class MtbScaleOverlay : Overlay {
                 or foot ~ yes|permissive|designated
                 or bicycle ~ yes|permissive|designated
               )
+              and bicycle != no
+              and mtb != no
               and (
                 surface ~ ${UNPAVED_SURFACES.joinToString("|")}|wood
                 or (highway = track and tracktype and tracktype != grade1)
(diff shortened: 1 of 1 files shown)


mnalis:
tested on e.g. on [way 268394040](https://www.openstreetmap.org/way/268394040), [way 123441110](https://www.openstreetmap.org/way/123441110), [way 124658932](https://www.openstreetmap.org/way/124658932)

westnordost:
Added mtb != no in a separate commit. I think the other values go too far.

What may be considered impassable by most vehicles can be a high mtb:scale. Also, bicycle=no may also be set not only because there is a sign but because it is impassable for bicycles. Again, that may just be a high mtb:scale grade.

mnalis:
OK... just for info:

- wiki for [smoothness](https://wiki.openstreetmap.org/wiki/Key:smoothness) says for `impassable`: _"Ways, often [highway](https://wiki.openstreetmap.org/wiki/Key:highway)=[path](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dpath), that are not passable by any vehicles, **not even mountain bikes** or specialised off-road vehicles, but are recognisable in the landscape and may still be useful for hikers"_

- [bicycle=no](https://wiki.openstreetmap.org/wiki/Key:bicycle) should be as wiki says _"specifies legal restriction for cyclists"_, buy I do get it it might be abused for unrelated bicycle issues. Issue is that many hiker-only paths are properly tagged only as "bicycle=no" to indicate legal restrictions, and I worry people will add `mtb:scale` on them, which will then in turn invite MTB-ers there. I'll guess alternative is to retag them all as `mtb=no` in addition of `bicycle=no` :shrug:

PR #6249, opened 2025-05-05
feat: moved AListQuestForm to jetpack compose

GaeaKat:
Moved the base AListQuestForm to compose, 

Before:
image

After:
image


GaeaKat:
If needed I can cut this commit down a lot by bringing the files out of the sub-folder, but as it had become a pair of files, rather then one, I had thought they belonged in one.

app/src/main/res/layout/quest_generic_radio_list.xml
westnordost, line 1: I notice this layout XML can (and should) be eliminated:

I.e. rename this to `compose_view.xml` or similar, and use that layout XML for all quest forms whose UI has been migrated.

app/src/main/java/de/westnordost/streetcomplete/quests/list_quests/ListQuestForm.kt
```diff
@@ -0,0 +1,61 @@
...
+                    .align(Alignment.CenterVertically)
+                    .padding(horizontal = 16.dp))
+            }
+        }
+    }
+}
+
+@Composable
+@Preview(showBackground = true)
+@Preview(showBackground = true, locale = "ar")
```
westnordost, line 48: I had added this specifically in the building form to test right-to-left layout cause there was some complexity involved. I think here, there's nothing special to test.

app/src/main/java/de/westnordost/streetcomplete/quests/list_quests/ListQuestForm.kt
```diff
@@ -0,0 +1,61 @@
...
+@Composable
+fun <T> ListQuestForm(options: List<TextItem<T>>, onSelectionChange: (TextItem<T>) -> Unit, currentOption: TextItem<T>?, modifier: Modifier = Modifier) {
+    Column(
+        modifier = modifier
+    ) {
+        options.forEach {
+            Row(modifier = Modifier
+                .padding(vertical = 8.dp)
+                .selectable(
+                    selected = (it == currentOption),
```
westnordost, line 33: I wonder, is the change animated or is it changed immediately on clicking a different item?

westnordost (commented):
You're fast! 🚀

---

Regarding formatting, I prefer
- one line for each parameter if it exceeds 100 characters length (e.g. the `ListQuestForm` function)
- if more than one parameter is passed to a composable, one line with parameter name for each parameter, e.g. 
  ```kotlin
  Text("my text")
  ```
  is fine, but if a parameter is added, I prefer if it is formatted like this
  ```kotlin
  Text(
    text = "my text",
    modifier = Modifier.padding(8.dp)
  )
  ```
- if a lambda / block ranges over several lines, use named parameters. E.g.
  ```kotlin
  filter { it.isCool() }
  ```
  is fine, but on several lines, it's better like
  ```kotlin
  filter { person ->
    val coolness = person.calculateCoolness()
    val ambientCoolness = getAmbientCoolness()
    coolness - ambientCoolness > 42
  }
  ```

---

I am not sure about the move to `list_quests` package. Currently, the `quests` package is ordered by type of quest, not by form functionality. I.e. `list_quests` sounds like there
are maybe quests that ask for something about lists (bus stop itinerary lists?).

The `AListQuestForm` is currently in the root because it can be used by all the other quests. And there are not that many files in the root that it becomes cluttered, IMO, but I am open to ideas how to better structure it.

westnordost:
> If needed I can cut this commit down a lot by bringing the files out of the sub-folder, but as it had become a pair of files, rather then one, I had thought they belonged in one.

Actually, at least for this "ListQuestForm", the composable is now so generic that there's no direct relation to "quests" (which is a good thing). So, it might just as well be renamed "TextItemSelect" or similar and moved to the `ui.common` package. 🤷

westnordost:
> Screenshot

It's interesting that the Jetpack compose radio button does not look exactly like the radio button from Android views. but anyway, it would be nice if horizontal padding is chosen as such so that the buttons align with the content in the title.

image

westnordost:
> Does have a dependency on Textitem though, Should I bring that with it rather then keeping it in AListQuestForm ?

`TextItem` is also generic.

westnordost:
>> Screenshot

> It's interesting that the Jetpack compose radio button does not look exactly like the radio button from Android views. but anyway, it would be nice if horizontal padding is chosen as such so that the buttons align with the content in the title.

Nevermind. (Deleted the comment.) Let's leave it like that for now. I don't even know right now if the other forms align with the title like that too or not (probably not). So, it would be inconsistent to enforce a certain padding here just for the list.

GaeaKat:
Renaming it to TextItemRadioGroup I think, as it is basically what it is acting as

app/src/main/java/de/westnordost/streetcomplete/quests/list_quests/ListQuestForm.kt
GaeaKat: ScreenRecording2025-05-05at1 57 13PM-ezgif com-video-to-gif-converter

westnordost: ah cool, so that is handled internally in the radio button. I notice a little thing here but will comment in the code

app/src/main/java/de/westnordost/streetcomplete/quests/list_quests/ListQuestForm.kt
```diff
@@ -0,0 +1,61 @@
...
+
+@Composable
+fun <T> ListQuestForm(options: List<TextItem<T>>, onSelectionChange: (TextItem<T>) -> Unit, currentOption: TextItem<T>?, modifier: Modifier = Modifier) {
+    Column(
+        modifier = modifier
+    ) {
+        options.forEach {
+            Row(modifier = Modifier
+                .padding(vertical = 8.dp)
+                .selectable(
```
westnordost, line 32: It would be better if the vertical padding was still part of the selectable area. I think this can be achieved by putting `.padding(...)` after `.selectable(...`

GaeaKat: Done that and it worked

GaeaKat:
I think these are those changes done

westnordost (approved):
LGTM 👍

PR #6253, opened 2025-05-06
Work on Image List Compose component

GaeaKat:
This is work to move the Image List quests to a compose component.


GaeaKat:
Looks like I will need to remake this after the other pr is merged, but this acts as a good placeholder for now to remind me to keep working

GaeaKat:
image
This is the progress so far

westnordost:
> Looks like I will need to remake this after the other pr is merged, but this acts as a good placeholder for now to remind me to keep working

How come?

GaeaKat:
> > Looks like I will need to remake this after the other pr is merged, but this acts as a good placeholder for now to remind me to keep working
> 
> How come?
> 
> You can simply specify the merge target not to be master but the branch you did the other work in.

Yeah, true

GaeaKat:
https://github.com/user-attachments/assets/95ad1eee-8f26-4b3d-b066-c09a153d2b91

Progress on this so far!

GaeaKat:
ScreenRecording2025-05-07at12 36 51AM-ezgif com-video-to-gif-converter
Progress on this so far

GaeaKat:
ScreenRecording2025-05-07at2 06 18AM-ezgif com-video-to-gif-converter
And the radio like behaviour also!

GaeaKat:
@westnordost Would you prefer I make a ui.common.Colors class, or use the colour resource? I keep seeing both recommended so I figure it is a matter of your preference here.

GaeaKat:
Screenshot 2025-05-07 at 5 22 20 PM
Need to slightly rewrite as I am having this issue, looks like some part isn't getting refreshed

westnordost:
I wonder how you managed to implement that the items can have different layouts. It's not ready yet, but I'll have a quick look at it!

GaeaKat:
> I wonder how you managed to implement that the items can have different layouts. It's not ready yet, but I'll have a quick look at it!

haven't uploaded that part yet, but basically

```
override val itemContent = @androidx.compose.runtime.Composable { item: ImageListItem<BoatRental>, index: Int, onClick: () -> Unit, role: Role ->
            SelectableIconItem(
                item = item.item,
                isSelected = item.checked,
                onClick = onClick,
                role = role
            ) }
            
  with default being
  protected open val itemContent = @androidx.compose.runtime.Composable { item: ImageListItem<I>, index: Int, onClick: () -> Unit, role: Role ->
            SelectableImageItem(
                item = item.item,
                isSelected = item.checked,
                onClick = onClick,
                role = role
            )
    }
```

app/src/main/java/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
```diff
@@ -0,0 +1,48 @@
...
+import de.westnordost.streetcomplete.quests.bicycle_repair_station.asItem
+
+@Composable
+fun <T> ImageList(imageItems: List<ImageListItem<T>>, onClick: (index: Int, item: ImageListItem<T>) -> Unit, modifier: Modifier = Modifier, itemsPerRow: Int = 4) {
+    LazyVerticalGrid(columns = GridCells.Fixed(itemsPerRow),
+        horizontalArrangement = Arrangement.spacedBy(4.dp), verticalArrangement = Arrangement.spacedBy(4.dp),
+        modifier = modifier.height(400.dp)
+        ) {
+        itemsIndexed(imageItems) { index, item ->
+            SelectableImageItem(
```
westnordost: Ah, of course, regarding my earlier comment about different layouts for the items:

Simply 

```kotlin
  fun ImageList(
    ...,
    itemContent: @Composable (item: Item<T>) -> Unit = { /* reference to default composable here */} )`
```

app/src/main/java/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
westnordost, line 1: By the way, I have been thinking to reduce the number of different cell layouts for the image selection lists, to not only make the appearance more consistent, but specifically remove the "text with shadow on top of photograph" layout, because it is a bit ugly shadow-ed text and also because especially if the text is long, it obscures the image.

(That layout is one of the first things I made for the app, back in 2017. Back then, the space on smartphones was much more constrained (because they were smaller). I remember back then, I targeted as minimum display dimensions 3.7 inch display diagonal. So, naturally, I tried to compress the information on as little space as possible. That is how the text-on-photo came to pass)

app/src/main/java/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
```diff
@@ -0,0 +1,48 @@
...
+                imageResId = item.item.drawableId!!,
+                title = stringResource(item.item.titleId!!),
```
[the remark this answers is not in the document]
GaeaKat: Definatly, I still plan to do a clean up once I know I have everything working well

app/src/main/java/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
GaeaKat: That is v similar to what I did in the end.

app/src/main/java/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
westnordost: And as a "new default", I think the "icon cell with text below" makes the most sense, for extra usability points, add somewhat rounded corners to the selection frame for single-selection (to mimic radio buttons) and somewhat square corners for multi-selection (to mimic checkboxes).

Conceivable would also be to add a checkmark (with animation) when selected. I should do some research what modern apps (that have picture/file-selection) show nowadays. In 2017 it was pretty much all over the place. Maybe nowadays everyone settled on a more material design-y look.

Anyway, just brainstorming a little. No need to make this part of this PR.

GaeaKat:
ScreenRecording2025-05-07at9 24 45PM-ezgif com-video-to-gif-converter
Finally got this working. Recycling Materials one is going to be a BIG pain currently, might need a few changes to how I did things

westnordost:
Hmm, I understand the issue.

 Functionally, IIRC, it is like an image select with groups:

Certain items have subitems of which one must be selected.

So, showing a popup dialog to select one could be the default behavior for the image select for such grouped items, while the grouped image select (building types) will have the expand/contracr-behavior

westnordost:
What's the issue with the lazy grid? It is better for performance, and also I'd like to keep the number of dependencies low. (Every dependency is a liability.)

GaeaKat:
> What's the issue with the lazy grid? It is better for performance, and also I'd like to keep the number of dependencies low. (Every dependency is a liability.)

It doesn't like  (read, gives errors about being an infinite height) being in a scrollable without there being a set size, and then when you have it in a set size, it gets iffy about scrolling within that thanks to the Scrollable inside the old UI section.

westnordost:
Right... I remember reading about issues with nested scrolling with compose. So, maybe indeed the fixed Grid must serve as an (hopefully) intermediate solution. In that case, best keep the code in a state where the fixed grid can be exchanged easily with the LazyGrid later. Also, best add a `// TODO Compose: should use LazyGrid as soon as possible` to e.g. the dependency in the gradle.kts

Here's also some documentation on View<->Compose nested scrolling interop, but maybe it is helpful:
https://developer.android.com/develop/ui/compose/touch-input/pointer-input/scroll#nested_scrolling_interop

GaeaKat:
Will do so, to be honest thats why I used the dependency, I tried making my own but it ended up being too specialised, and not similar enough to the LazyViewGrid's use, this dependency wasn't much of a change from it to a Lazy one

westnordost:
I merged that other PR. I also merged a massive PR that sorts all the project files into different directories (for KMP). Will you continue this in this PR or do you prefer to remake this in a new one? I am referring to [your second comment].(https://github.com/streetcomplete/StreetComplete/pull/6253#issuecomment-2854541179)

westnordost:
cool! No hassle, I only asked because I wanted to make sure whether it makes sense if I merged the current master into this PR. Good that I asked, so I saved some time :-)

GaeaKat:
Perfect, it will mostly be the same as this, but while checking over, I noticed some discrepencies from some files where I did some quests before I figured out using key() was needed so some had it some didn't, so if one change was like that, chances are more will be, so redoing to make sure that that doesn't happen

PR #6285, opened 2025-05-23
Port length form to Jetpack Compose

ltoenning:
This ports the 

- [ ] Max Physical Height
- [ ] Width 

quests to Jetpack Compose.


westnordost (commented):
Wow, nice! Since you marked this as draft so far, I am not going to comment on it in detail, yet. Just a few things from reading vertically:

1. same as in this PR: https://github.com/streetcomplete/StreetComplete/pull/6282/files/8074146b61358cd111dc43636e555aa889938096#diff-f900becb28b48fe1591b767e9f32292f1595207f632d1a6dafccbac02231b3af

2. `dimensionResource` is Android, let's use a `MaterialTheme.typography` constant. (I am trying to consolidate the number of different text styles.)  
  
    Maybe MaterialTheme.typography.h3? Adding a new constant in Typography.kt and using that would also be fine by me. (Should be the same as in the linked PR) (Try not to do any custom styling but rely on the Material typography)

3. Composables specific to the "main" widget in a file should be `private` (e.g. the `MeasureButton`

4. Regarding MeasureButton, have a look at ActionIcons. I think it would make sense to define the `Icon` there. Don't fear adding a new string for a fitting content description, if there's no eligible string yet

ltoenning:
This PR is ready for review now. By the way: These are my first steps with Kotlin, Android and Compose. Sorry in advance, if the changes contain obvious issues 🙈  

Open questions from my side:
- The ``LengthForm.kt`` includes a check function to validate that a string only contains a decimal digit. It is derived from the (Android-specific?) filter from ``InputValidators.kt``. Should it be moved to a different file?
- To use ``Length?`` as a state, I was required to add a compose-specific ``Saver``. Should this be placed directly next to the Length classes or into a different file?

westnordost (changes requested):
So, this PR is pretty impressive regarding that you never touched neither Android, Kotlin nor Compose before? Wow 😯 - but you did some reading up on Compose, right? Cause, you do a lot of things right that would otherwise not be intuitive.

---

You reimplemented `LengthInputViewController` in compose. So, that class can be deleted now, right? Ah no, there's still one usage left: `AddMaxHeightForm`

So, this should probably get the same treatment as `AddMaxPhysicalHeightForm`.

Ah, right, I get it. The layout in `AdMaxHeightForm` is different, as it includes the actual sign as a background graphic. (Or actually, not even a backgrould graphic, but a background XML layout). Additionally, the spinner to select the unit is not directly next to the input field. Plus, the text appearance and size of the input field may actually differ depending on the sign displayed. This is interesting. I have to think about this a bit how to solve this, but in general the best path to follow is to try to create generic widgets that can be puzzled together with maximum flexibility. E.g. the `explanation` parameter is really not necessary, instead, wherever the length input is used and an additional text should be displayed, well, that `Text` can just be added as a regular composable. Avoid "Eierlegende Wollmilchsau" composables.
(One could argue that the `BuildingLevels` composable is somewhat of that, but this layout is also used nowhere else, so there is no need for a "generic" composable)

(Anyway, this turns out to be more work than originally thought. It's okay if you would rather leave all that rest to me.)

ltoenning:
Yes, I read about Compose and its concepts first. I also have some experience with other declarative UI frameworks (Qt QML) so the switch wasn't too hard.
I will try to look into your comments next week!

app/src/main/java/de/westnordost/streetcomplete/osm/Length.kt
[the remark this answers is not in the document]
ltoenning: > Did this stem from a particular requirement because something wouldn't work correctly, or did you just follow best practices / tutorials?

Without it, the app crashes when closing the quest because of the missing saver for the custom data type.
I also thought about replacing the custom data type with a primitive data type. But this will require multiple variables: meters, feet and inch. And it will complicate the handling on which value is the currently used one.

ltoenning:
I've just set the state back to draft. This now also includes changes to port the MaxHeight quest to compose. But this is not fully ready yet.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/MaxHeightForm.kt
```diff
@@ -0,0 +1,79 @@
...
+        modifier = modifier.fillMaxWidth(),
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalArrangement = Arrangement.Center
+    )
+    {
+        Box {
+            Image(
+                painter = painterResource(R.drawable.background_maxheight_sign),
+                contentDescription = "Maximum Height Sign"
+            )
```
westnordost, line 47: Haha, sorry, I think it's not going to be that easy 😜

So, you might wonder why in the old MaxHeightForm, I set a different content **layout** depending on the country:

```kotlin
    override val contentLayoutResId get() = when (countryInfo.countryCode) {
        "AU", "NZ", "US", "CA" -> R.layout.quest_maxheight_mutcd
        "FI", "IS", "SE" -> R.layout.quest_maxheight_fi
        else -> R.layout.quest_maxheight
    }
```

and not just a background graphic. The reason is that this widget mimics the appearance and layout of the street sign (I know, this is kind of "Spielkram" and not necessary from a data-input point of view, but this kind of stuff makes the app unique and more "fun" to use, so I'd not like to loose that.):

image
image

So, the arrangement of the elements, size, padding etc differs. (' " vs FT IN)

Now, I think for the most part, it might actually work just with a background image and some parameters passed on to the `LengthInput`. But, what will definitely work more cleanly, I think, is to make the signs itself a `Composable` that accepts content (like a `Box`). E.g. something like

```kotlin
@Composable fun MaxHeightSign(
  //...
  content: @Composable () -> Unit
) {
  Box(modifier) {
    // ...layout the sign (background)
    Box(...) {
      // and the content within it
      content()
    }
  }
}
```

(Using `CompositionLocalProvider`s, one can set the default text style, color etc. for the child composables. This means, that the generic `LengthInput` composable should probably not even define an own font size / text style, but just use whatever is the current default)

ltoenning: > So, you might wonder why in the old MaxHeightForm, I set a different content layout depending on the country:

I planned to use different composables, based on the country (sign) here:

https://github.com/ltoenning/StreetComplete/blob/4fc55e0698a1fe4a7ef53c17570b6cf7453705d1/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxHeightForm.kt#L35-L46

I just haven't implement the FI and MUTCD yet ;)

app/src/main/java/de/westnordost/streetcomplete/quests/LengthForm.kt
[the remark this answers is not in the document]
westnordost: Not sure about `Button` vs `IconButton`, but well, if a specific composable exists for that in material design, then I'd say it indeed makes sense to use that specific composable for an icon button.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/MaxHeightForm.kt
westnordost: Well, right, better use the approach mentioned above, so that you don't end up duplicating composables. In case you didn't plan that all along.

E.g. there could be a 

- `MaxHeightSignDefault` that displays a normal max height sign and has a `content` parameter for the actual content of the sign
- MaxHeightSignNordic` same but for Nordic countries (i.e. different background graphic)
- MaxHeightSignMutcd` same for US etc.

and then a `MaxHeightSign` composable with a `when` like in my previous comment.

The `MaxHeightForm` then just has this code (mockup code):

```kotlin
Row {
  MaxHeightSign(
    countryCode = countryInfo.countryCode,
    ...
  ) {
    LengthInput(
      unit = ...,
      ...
    )
  }
  LengthUnitSelector(
    selectableUnits = ...,
    onSelected = { ... },
    ...
  )
}

```

ltoenning: If ``MaxHeightForm`` includes the same ``LengthInput`` for all countries, how would you trigger the different display for the US sign (with "FT" and "IN")? Maybe the ``LengthInput`` should be part of the ``when`` in the MaxHeightSign composable and then be injected into the specific Sign composable?
Would you prefer to extend the ``LengthInput`` do display "FT"/"IN" instead of ' and '' or does this again creates the Eiermilchlegende Wollmilchsau again? ;)

I am currently also struggling the the ``LengthInput`` to make it flexible to use with different text styles. The ``TextField`` with material design seems to include much (fixed) padding which creates a large box for small texts. Maybe it would be better to switch to ``BasicTextField`` for more control? But then the question: How to make the size of the box match around the size of the textstyle/font...

westnordost: The `LengthInput` would need an additional parameter that is passed to it, e.g. 

```kotlin
enum class FootInchAppearance { PRIME, UPPERCASE_ABBREVIATION }
```

( Yeah, didn't know that it is called that way before https://en.wikipedia.org/wiki/Prime_(symbol) )

Or, well, since you also gotta pass to it which unit has to be used, this could be one parameter as it doesn't make sense to pass appearance for ft/in if it should display meters. 🤔

So, something like...

```kotlin
sealed interface LengthUnit {
  data class FootInch(val appearance: FootInchAppearance) : LengthUnit
  data object Meter : LengthUnit
}
```

Well, at least that's my idea about it.

---

> TextField with (fixed) padding

Oh, yes? Hmm, that could be. All I remember is that `TextField` comes with some functionality one usually expects from text fields, IIRC, so usually you always want to use that, and not `BasicTextField`. If you don't set any text styling or padding yourself, and have the (in this case) max height sign background image scale to accomodate whatever is the resulting size of the whole input thingy, does it look odd?
After all, I think it would be fine if the (new) length input field results in being a little larger than the previous version, if it conforms to accessibility guidelines enforced by Material (e.g. minimum tap size of elements)

> How to make the size of the box match around the size of the textstyle/font...

The size of the `Box` automatically grows to accommodate its content. Do you mean how to ensure that the background sign graphic also scales with it? I.e. takes as much space as the input fields inside the box? Such things can be done with https://developer.android.com/develop/ui/compose/layouts/intrinsic-measurements

ltoenning: I've just added the different composables for the the default, nordic and mutcd. They are considerable larger than the old inputs as the ``TextField`` is larger. But this probably helps for better accessibility. The round sign still contains a hardcoded size to keep it (a little) larger than the text input, so that the input fits into the sign. Is this acceptable?

The FilledIconButton also requires Material3. Is it okay to add this as a dependency?

westnordost:
> In the settings screen, at the very bottom, you can look at the quest form for every quest.

After you merge from master, you now need to add the line 

```
app.streetcomplete.debug=true
```

to your `~/.gradle/gradle.properties` (on Linux, Mac) `C:\Users\<your username>\.gradle\gradle.properties` (on Windows, IIRC) to have this functionality.

ltoenning:
> Wow, this PR got really large. Let me know when I should review this, or parts of this. Don't forget to test your changes before requesting a review (including in dark mode).

This is the current state. It still uses the fixed size for the height sign. I tried different options but couldn't get it properly scaled without the fixed scale, especially with the large default padding of the TextFields. Any help is much appreciated! Otherwise, should be ready for review now.

## Max Height

height_ft_darkmode
height_ft_us_darkmode
height_m_darkmode

## Max Width
width_ft_darkmode
width_m_darkmode
width_m_lightmode

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/util/InputValidatorsUtil.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/** Only accept inputs that are numbers with the max number of digits before and after the decimal point */
+fun onlyDecimalDigits(string: String, beforeDecimalPoint: Int, afterDecimalPoint: Int): Boolean {
+    if (!string.all { it.isDigit() || it == '.' || it == ',' }) return false;
+    val texts = string.split(',', '.')
+    if (texts.size > 2 || texts.isEmpty()) return false
+    val before = texts[0]
+    val after = if (texts.size > 1) texts[1] else ""
+    return string.toDoubleOrNull() != null && after.length <= afterDecimalPoint && before.length <= beforeDecimalPoint
+}
```
westnordost: On e.g. German locale/keyboards, I'd expect people to input e.g. "5,3".

You added `string.toDoubleOrNull() != null`  to the check, which then would return false for such numbers. 
We already check if the string only consists of digits + at most one of "." or ",", so that should be enough.

Also, this function checks more than "only decimal digits", it also checks that the string isn't something like `5.000,40`. See also the comment mentioning `anyDecimalToDouble`. Maybe if there was (such) a function that would convert a string to a Double regardless of locale, this function, or a `*OrNull`-variant of it could be used here to not duplicate code.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/LengthUnitSelector.kt
```diff
@@ -0,0 +1,43 @@
...
+        onClick = {
+            selectedUnitIndex.intValue =
+                (selectedUnitIndex.intValue + 1) % selectableUnits.size
+            onUnitChanged(selectableUnits[selectedUnitIndex.intValue])
+        },
+        enabled = selectableUnits.size > 1,
+        modifier = modifier
+    ) {
+        Text(selectableUnits[selectedUnitIndex.intValue].toString())
+    }
```
westnordost: Ah, was it just a button before? I think it would make sense to have it (now) as an "[exposed dropdown menu](https://m3.material.io/components/menus/overview)" (aka "[Pull-down button](https://developer.apple.com/design/human-interface-guidelines/pull-down-buttons)" on iOS). Especially the ⯆ is important to indicate that this is a unit selection. Would you agree?

The composable to be used for that is [ExposedDropdownMenuBox](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary?hl=en#ExposedDropdownMenuBox(kotlin.Boolean,kotlin.Function1,androidx.compose.ui.Modifier,kotlin.Function1)) (et al).

(#6309 also needs a unit selector. Maybe it makes sense to make it a template function (i.e. with `<T>`)?)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/MaxHeightSign.kt
```diff
@@ -0,0 +1,77 @@
...
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.meta.LengthUnit
+import de.westnordost.streetcomplete.osm.Length
+import de.westnordost.streetcomplete.ui.common.FootInchAppearance
+import de.westnordost.streetcomplete.ui.common.LengthInput
+import de.westnordost.streetcomplete.ui.common.RectangularSign
+import de.westnordost.streetcomplete.ui.theme.TrafficYellow
+
+@Composable
+fun MaxHeightSign(
```
westnordost, line 21:
```suggestion
fun MaxHeightSignInput(
```
(cause it's an input field, to distinguish properly from the other `*Sign` composables. Change file name too)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/MaxHeightSign.kt
```diff
@@ -0,0 +1,77 @@
...
+            ) {
+                LengthInput(
+                    selectedUnit = selectedUnit,
+                    currentLength = null,
+                    syncLength = false,
+                    onLengthChanged = onLengthChanged,
+                    maxFeetDigits = maxFeetDigits,
+                    maxMeterDigits = maxMeterDigits,
+                    footInchAppearance = FootInchAppearance.PRIME
+                )
```
westnordost: missing is something like (pseudo-code):
```
Row {
  LengthInput(...)
  if (selectedUnit == meters) {
    Text("m")
  }
}
```
because that's what you find on the sign

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/LengthInput.kt
```diff
@@ -0,0 +1,243 @@
...
+                if (validInchInput(input)) {
+                    val value = input.toInt();
+                    if (feet.isNotEmpty()) {
+                        onLengthChanged(Length.FeetAndInches(feet.toInt(), value))
+                    }
+                    inches = value.toString()
+                }
+            },
+            textStyle = LocalTextStyle.current.copy(textAlign = TextAlign.Center),
+            modifier = Modifier.width(75.dp),
```
westnordost: In Android XML layouting, there was a `ems` property to define the width of a text field according to actual font size used (which may differ depending on the accessibility settings of the phone). This doesn't seem to exist in compose.
I also tried my luck to un-hardcode the size of the round maxheight sign, but didn't succeed.

Let's look into getting rid of these hardcoded values later when the other issues are solved. I have a few ideas.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/LengthInput.kt
```diff
@@ -0,0 +1,243 @@
...
+            if (input.isEmpty()) {
+                onLengthChanged(null)
+                lengthMeterText = input
+            } else if (onlyDecimalDigits(
+                    input,
+                    maxMeterDigits.first,
+                    maxMeterDigits.second
+                )
+            ) {
+                onLengthChanged(Length.Meters(input.toDouble()))
```
westnordost: So this is going to fail for `"3,5"`, AFAIK. We should test what happens on e.g. a German locale and one inputs a decimal with a comma. If it fails, we need some generic solution that we can apply to all the other text fields that allow decimal input.
If there exists nothing in the Kotlin standard library itself, we might want to define an own extension function on `String` - `anyDecimalToDouble` or something. 🤷

(See also https://en.wikipedia.org/wiki/Decimal_separator )

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/LengthForm.kt
```diff
@@ -0,0 +1,103 @@
...
+                LengthInput(
+                    selectedUnit = selectedUnit.value,
+                    currentLength = currentLength,
+                    syncLength = syncLength,
+                    onLengthChanged = onLengthChanged,
+                    maxFeetDigits = maxFeetDigits,
+                    maxMeterDigits = maxMeterDigits,
+                    footInchAppearance = FootInchAppearance.PRIME
+                )
+            }
```
westnordost:
```suggestion
            LengthInput(
                selectedUnit = selectedUnit.value,
                currentLength = currentLength,
                syncLength = syncLength,
                onLengthChanged = onLengthChanged,
                maxFeetDigits = maxFeetDigits,
                maxMeterDigits = maxMeterDigits,
                footInchAppearance = FootInchAppearance.PRIME
            )
```

I don't think this is needed here, right?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/LengthInput.kt
westnordost, line 1: The thing with the syncLength, I feel this has been done at the wrong point. I.e. that logic should be handled not at the deepest point in the composable hierarchy, but at the point where there's both the camera button and the length input field.

Anyway, I am kind of unsure what should even be achieved with that. So, let's take a step back: What _do_ we want to achieve?

1. we want the AR measurement to update the length input.
2. we want to keep track of where the last update to the length input originated from - from the AR measurement or from user input. (we tag `source:width=ARCore` in the former case)

Both should be no problem to do on the upper level of the hierarchy (i.e. `LengthForm`), or did I miss something?

westnordost (changes requested):
All in all, a very clean implementation, I am impressed! I added some comments for some issues and solved some other minor things myself (see the previous commits).

Here are some general comments:

- it would be nice to have short doc comments for all public composables

- all public composables should have a `modifier: Modifier = Modifier` parameter. I already changes for that, I think I got them all

- I am considering using `OutlinedTextField` instead of `TextField` for all text input fields, because that material design text field looks so Android-specific (and also ugly, IMO), while an outline text field is universally understood. I am not entirely sure though if outlined text fields would look fine on the signs?

- For consistent paddings, I try to limit paddings to only 4, 8, 16 or 32 (i.e. 2^x)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/LengthForm.kt
```diff
@@ -0,0 +1,110 @@
...
+
+
+            LengthUnitSelector(
+                selectableUnits = selectableUnits,
+                selectedUnit = selectedUnit.value,
+                onUnitChanged = { selectedUnit.value = it },
+                modifier = Modifier.padding(5.dp)
+            )
```
westnordost:
```suggestion
            if (selectableUnits.size > 1) {
                LengthUnitSelector(
                    selectableUnits = selectableUnits,
                    selectedUnit = selectedUnit.value,
                    onUnitChanged = { selectedUnit.value = it },
                    modifier = Modifier.padding(5.dp)
                )
            }
```

If there's only one selectable unit, we don't want to show the selector at all.

westnordost:
I'll look into the locale-neutral decimal parsing

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/util/InputValidatorsUtil.kt
westnordost: Alright, I added locale-aware formatting and parsing [in this commit on master](https://github.com/streetcomplete/StreetComplete/commit/39b770f13cee0ca15cc85134ee92bf917b8db41a).

You can use it as such:
```kotlin
// assuming current locale is de-DE
val formatter = NumberFormatter()
formatter.format(1000.123, maxFractionDigits = 2, useGrouping = true) // returns "1.000,12"
formatter.format(1000.123, maxFractionDigits = 1, useGrouping = false) // returns "1000,1"
formatter.format(0, minFractionDigits = 2) // returns "0,00"
formatter.parse("1.000,12", allowGrouping = true) // returns 1000.12
formatter.parse("1000,123") // returns 1000.123
formatter.decimalSeparator // returns ','
formatter.groupingSeparator // returns '.'
```

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/LengthInput.kt
westnordost: **TL;DR:** Let's keep the hardcoded sizes for the sign layout(s) and also the hardcoded sizes in `dp` for the text fields, **but** make those text fields automatically reduce their font size so that any text entered always fits inside. I added the `AutoFitTextField` and `OutlinedAutoFitTextField` composable to `master`, they do have the same interface as `TextField` etc. but reduce their font size automatically if it exceeds its set width.

---

Alright, I spent the whole day trying this and that to make the max height sign while keeping the 1:1 aspect ratio by:
- programming a `Modifier` that would allow to set the width of a text field to whatever space X number of digits take (depending on the current font size)
- programming a `Modifier` that would set the height equal to the measured width so to make the sign that's stretched to the width of the sign content stretch the same into the y-direction

However,  it. just. doesn't. work.. I could post the current state of the code of those modifiers, but, well, they don't work (completely), so this is not going to be helpful.

Instead, let's do it the other way round. See the TL;DR at the top. This is, by the way, also exactly the way it works in the current UI.

westnordost:
If it is okay by you, I'll do bring this PR to completion. Thanks for your work so far!

ltoenning:
> If it is okay by you, I'll do bring this PR to completion. Thanks for your work so far!

Sorry for being inactive the last days. I was quite busy. That's fine for me. Thanks!

westnordost:
You won't believe how complicated and thus time-consuming it turned out to be, which is weird, because it shouldn't be! But IMO that's on the Compose Material / Foundation team at Google. It is outlandish that there isn't already a "decimal input" composable or something that makes creating it trivial.

Also, as `TextField` and `OutlinedTextField` are completely separate composables, it is really awkward to create generic composables based on text fields as one has to implement it twice for each such composable...? (Once for the outlined version and once for the filled version?)

And then, `TextField`s have some arbitrary default min width in fixed pixels while the text size is given in scalable pixels (i.e. font size can be changed in the system setting), fine, but one can
- neither specify that the text field should be as long to fit a certain text or certain number of characters in whatever the current font size is
- nor tell the text field to expand its size to fit its content
- nor tell the text field to shrink the font size so that the content fits the bounds

This is a really outlandish combination, ludicrous.

Never mind my failed attempts at sizing the maxheight sign background in 1:1 aspect ratio to whatever the size of content would be. I _thought_ that should be easy, apparently it is not. But maybe that's on me 🤷.

Anyway, I ended up also doing the following:

- introduced `TextField2` which is a generic `TextField` where the style (outlined or filled) can be set with parameter (so that it is not that cumbersome to create specialized `TextField`s independent of styling - such as `DecimalInput`). 

- refactored the autofit font size functionality to be a parameter of `TextField2` (so that those composables that are based on it can also use that functionality)

- introduced `DecimalInput`, a generic composable to input decimal numbers (locale-aware), like meters

- refactored the `NumberFormatter` for better usability by `DecimalInput`

- introduced `Selector`, a generic simple selector of different elements

- introduced `RectangularSign`, a box with the appearance of a rectangular traffic sign. Currently unused.

- introduced `largeInput` text style

And this is how it looks:

### Physical width



_in metric country with device that has AR capability_

--- 



_in country where both measurement systems are used with device that has no AR capability_

### Height sign






---

Now, I rummaged around in that code a lot, so maybe you are interested to review *my* code now?

~Anyway, regarding the UI, I am not sure if I like that selector field, it's so... unwieldy. Maybe would look better as a button(?) - that's the standard styling for a `<select>` in HTML, too.~ (_done that_)

Also, I am considering dropping this play with font sizes in different forms and using the standard font size everywhere, even if there would be space for a larger font size here and there. I thought it looks cool if the input (on a sign) is big and fat, but then again, maybe the UI will look a bit inconsistent.

westnordost:
I made the unit selector look like a button after all and moved it to the left of the input field(s) in both the maxheight and physical height forms, for consistency.

From my point of view, this'd be ready to merge.

westnordost:
(I lifted all those additions out of this PR)

westnordost:
Alright, I interpret your 🎉-reaction as "nice, go ahead, no I don't need or want to review it" 😉. You can still review the PR after merge, too, after all.

ltoenning:
> Alright, I interpret your 🎉-reaction as "nice, go ahead, no I don't need or want to review it" 😉. You can still review the PR after merge, too, after all.

Yes. Thanks for finishing the MR!

PR #6290, opened 2025-05-26
Surface synonyms, remove concrete:plates as selectable surface

### LINKED ISSUE #6265: Discourage tagging surface=concrete:plates, especially on footways
Between November 2024 and March 2025, a [discussion on the OSM community forum](https://community.openstreetmap.org/t/surface-concrete-vs-concrete-plates/122298) resulted in general consensus there that `surface=concrete:plates` should be used only for precast plates, and not for concrete cast in place. The OSM wiki pages for [surface=concrete](https://wiki.openstreetmap.org/wiki/Tag:surface%3Dconcrete) and [surface=concrete:plates](https://wiki.openstreetmap.org/wiki/Tag%3Asurface%3Dconcrete%3Aplates) were updated accordingly. (I pinged previous contributors to concrete:plates discussion in StreetComplete, @matkoniecz and @mnalis, in the community forum thread.)

Currently, about 40% of `surface=concrete:plates` worldwide is tagged on footways, half of that on sidewalks ([Taginfo](https://taginfo.openstreetmap.org/tags/surface=concrete%3Aplates#combinations)). In Canada, 92% of `surface=concrete:plates` is tagged on footways, over two-thirds of that on sidewalks ([Taginfo Canada](https://taginfo.geofabrik.de/north-america:canada/tags/surface=concrete%3Aplates#combinations)). From local knowledge, it would be very rare for Canadian sidewalks to be built from precast reusable concrete plates - most of them are cast-in-place concrete and some are asphalt.

As seen in https://www.openstreetmap.org/way/1115854620/history/2, StreetComplete users are tagging `concrete:plates` on ordinary cast-in-place sidewalks as recently as April 2025, using `StreetComplete_ee 60.1`.

We can and will attempt to reach out to users to let them know of the difference, but I was wondering if it would be a good idea to remove `concrete:plates` from preset answers for the surface quest in StreetComplete, at least for footways, or at least in some geographies.

Is there a region of the world that commonly uses prefab concrete plates for sidewalks? Or, is there anything else you can think of to make this error less likely? Maybe change the label to mention "prefabricated" concrete?

Previously discussed in StreetComplete in 2022: https://github.com/streetcomplete/StreetComplete/issues/4344

westnordost:
fixes #6265

Turned out to be more involved.

Because previously, we didn't have the notion of aliases for surfaces (e.g. `surface=earth` == `surface=dirt`) like we had for building types.

This would lead to that when one would acknowledge the shown surface in a re-survey or overlay, let's say paving stones, and it was previously tagged as `surface=bricks` (considered an alias/subtype of this), then it would be re-tagged as `surface=paving_stones`. But we don't want that, we want it to stay `bricks` in this case.

So, I made this work like the aliases for `BuildingType`, and also made it consistent how both of these are implemented.


westnordost:
@matkoniecz a review by you would be highly appreciated, as you've been the person most involved with `surface` stuff

aikencz:
I strongly disagree with removing of `concrete:plates`. I used this surface type relatively often (and correctly!). With the new version, it brings additional work to me. I have to add a note in StreetComplete and then set the correct surface in iD or JOSM.

westnordost:
Well, see here for my final opinion on it and why I agreed to remove it in the end:
https://github.com/streetcomplete/StreetComplete/issues/6265#issuecomment-2912022626

PR #6292, opened 2025-05-27
KMP setup

westnordost:
Setup project as KMP and move many files into the common source directory already


FloEdelmann:
@westnordost some unit tests are failing since this pull request: https://github.com/streetcomplete/StreetComplete/actions/runs/15283298338/job/42987341515

Do they need to be adjusted or did something break during the KMP setup?

matkoniecz:
> Maybe you should do a full rebuild / clear caches, but IIRC everything worked fine for me just checking it out.

yes, it worked perfectly fine

PR #6294, opened 2025-05-27
Convert Image List to Compose

GaeaKat:
Redo of #6253 With less clutter


GaeaKat:
Wooooo
Screenshot 2025-05-29 at 1 03 10 AM

GaeaKat:
Tasks to do before this is ready:

- [x] Make Recycling List work with it's categories
- [x] Make smoothness form show properly with the emotes
- [x] Make Add Steps None exlusive
- [x] Re-add option to move favourites to the top

westnordost:
[ ] you can probably delete quite a few now-unused XML layout resources.

GaeaKat:
> * [ ]  you can probably delete quite a few now-unused XML layout resources.

Every one I look at is still being used elsewhere, mostly by AImageSelectOverlayForm Items,

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/AImageListQuestComposeForm.kt
```diff
@@ -0,0 +1,155 @@
...
+        itemsByString = items.associateBy { it.value.toString() }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        currentItems.value = if ( items.size > itemsPerRow && moveFavoritesToFront) {
+            val favourites = prefs.getLastPicked(this::class.simpleName!!)
+                .map { itemsByString[it] }
+                .takeFavorites(n = itemsPerRow, history = 50)
+            Log.d("Temp", "Favourites: " + favourites.map { it.value.toString() })
```
GaeaKat, line 81: I need to remove this temp log line

GaeaKat:
There we go, I have no doubt I have a lot to do , but this is the point I need to make sure is being done in a way that fits in the app!

GaeaKat:
Need to make this work in dark mode

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt
```diff
@@ -41,6 +41,9 @@ val GrassGreen = Color(0xff80b158)
 val GrassGray = Color(0xff888888)
 val LeafGreen = Color(0xff006a00)
 
+val SelectionColor = Color(0x66ff6600)
+val SelectionFrameColor = Color(0xbbff6600)
+
```
westnordost: Actually, this ought to be the `MaterialTheme.colors.secondary` (which is slightly different in light and dark mode), but with transparency. 
So you should write something like `MaterialTheme.colors.secondary.copy(alpha = ...)` to achieve this.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
```diff
@@ -0,0 +1,163 @@
...
+        when (item.image) {
+            is ResImage -> {
+                val resDrawable = ContextCompat.getDrawable(context, (item.image as ResImage).resId)
+                resDrawable?.toBitmap(resDrawable.intrinsicWidth, resDrawable.intrinsicHeight)?.asImageBitmap()
+            }
+
+            is DrawableImage -> (item.image as DrawableImage).drawable.toBitmap().asImageBitmap()
+            else -> null
+        }
+    }
```
westnordost: Oh wow, what is all this complexity? It's so ugly! Oh, I get it, because of `DrawableImage`.

So, we should
- convert all places that use `DisplayItem`s to compose (`AGroupedImageSelectOverlayForm`, `AImageSelectOverlayForm`, `AGroupedImageListQuestForm`)
- change `DisplayItem` to use `Painter` instead

But that may some more effort, maybe until then, add a TODO:

```
    // TODO compose: after converting all classes that use DisplayItem to compose, simplify this
```

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    val title = remember(item.title) {
+        when (item.title) {
+            is ResText -> context.getString((item.title as ResText).resId)
+            is CharSequenceText -> (item.title as CharSequenceText).text.toString()
+            null -> ""
+        }
+    }
+
+    var imageWidth by remember { mutableStateOf(0) }
```
westnordost: What's that for?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/RadioImageList.kt
```diff
@@ -0,0 +1,65 @@
...
+@Composable
+fun <T> RadioImageList(items: List<DisplayItem<T>>,
+    onSelect: (List<DisplayItem<T>>) -> Unit,
+    modifier: Modifier = Modifier,
+    itemsPerRow: Int = 4,
+    itemContent: @Composable (item: ImageListItem<T>, index: Int, onClick: () -> Unit, role: Role) -> Unit) {
+    var listItems by remember { mutableStateOf(items.map { ImageListItem(it, false) }) }
+    LaunchedEffect(items) {
+        listItems = items.map { ImageListItem(it, false) }
+    }
```
westnordost: Why would you need this, you already have this in the remember. (Also, the remember should be on the `items`, i.e. `var listItems by remember(items) { ... `

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/RadioImageList.kt
```diff
@@ -0,0 +1,65 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.tooling.preview.Preview
+import de.westnordost.streetcomplete.quests.bicycle_repair_station.BicycleRepairStationService
+import de.westnordost.streetcomplete.quests.bicycle_repair_station.asItem
+import de.westnordost.streetcomplete.view.image_select.DisplayItem
+
+@Composable
+fun <T> RadioImageList(items: List<DisplayItem<T>>,
+    onSelect: (List<DisplayItem<T>>) -> Unit,
```
westnordost: Why does it return a list, if only one item can be selected?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
```diff
@@ -0,0 +1,131 @@
...
+import de.westnordost.streetcomplete.quests.bicycle_repair_station.asItem
+import kotlin.math.max
+
+@Composable
+fun <T> ImageList(
+    imageItems: List<ImageListItem<T>>,
+    onClick: (index: Int, item: ImageListItem<T>) -> Unit,
+    modifier: Modifier = Modifier,
+    itemsPerRow: Int = 4,
+    itemRole: Role = Role.Checkbox,
```
westnordost: Hmm, this is a parameter, yet, the composable doesn't seem to do anything with this information. Just passes it back to the caller on click. (But the caller already knows each item's role, otherwise, it wouldn't wouldn't be able to pass it here).

westnordost (changes requested):
A first review, I've mostly looked at the composables so far.

Would you rename the composable `*Item`s to `*Cell`, `*Row` etc.? This is consistent with how the items are named in e.g. the user screen and that naming doesn't collide with the data classes (`DisplayItem`, `ImageListItem`).

Analogous, the `*List`s should be renamed to (Lazy, if applicable) `*Grid` or `*Column`.

Also, short doc comments for the public composables would be nice. (I.e. what are they)

I have some problems grokking ImageList. The general structure looks good

It is a bit unfortunate if  you put much effort/complexity into a custom layout, when this will hopefully just be an intermediate solution until this is all compose.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
```diff
@@ -0,0 +1,131 @@
...
+    itemContent: @Composable (item: ImageListItem<T>, index: Int, onClick: () -> Unit, role: Role) -> Unit
```
westnordost: Regarding this interface:

What is `onClick` and `index` for? Is the composable just expected to pass it through? In that case, should just wrap it in a `Box` in line 31 and register the `Modifier.clickable {}` there. So, it could look like this:
`itemContent: @Composable (item: DisplayItem<T>, selected: Boolean) -> Unit`

GaeaKat:
> A first review, I've mostly looked at the composables so far.
> 
> Would you rename the composable `*Item`s to `*Cell`, `*Row` etc.? This is consistent with how the items are named in e.g. the user screen and that naming doesn't collide with the data classes (`DisplayItem`, `ImageListItem`).
> 
> Analogous, the `*List`s should be renamed to (Lazy, if applicable) `*Grid` or `*Column`.
> 
> Also, short doc comments for the public composables would be nice. (I.e. what are they)
> 
> I have some problems grokking ImageList. The general structure looks good
> 
> It is a bit unfortunate if you put much effort/complexity into a custom layout, when this will hopefully just be an intermediate solution until this is all compose.

Honestly, it was good 'practice' to figure out how compose does custom layout.

I will start work on the other changes

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt
GaeaKat: Will do so

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
GaeaKat: Oh, I think it might be left-over from an earlier attempt, will remove, don't know why the ide didn't ping me it was unused

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/RadioImageList.kt
GaeaKat: It is one of the things where if I remove it, The orientation sensitive items no longer work

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt
westnordost: The frame should actually not have any alpha at all, you can just take the `secondary` as-is

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/RadioImageList.kt
GaeaKat: This is a bit legacy code, I am honestly thinking of removing this, but it was to make it match the other list's output

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
westnordost: Well, it's kind of used, but I don't understand its purpose :-)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
GaeaKat: I can remove this, but it makes the 2 special cases a lot more difficult, Those being https://github.com/streetcomplete/StreetComplete/pull/6294/files#diff-9299f4406f52c91166f6a6c5524b942151b659826eb9f31b9e4ae7b30ac1567cR30-R53 and https://github.com/streetcomplete/StreetComplete/pull/6294/files#diff-c10b39917ab89f663266b5d79800b35ffb8dcbdc435af7d45fa8b3bafc4ddfdeR26-R35

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/RadioImageList.kt
GaeaKat: Removed RadioImageList.kat and MultiImageList.kt

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt
GaeaKat: Which one? the Selection Color or the Selection Frame Color?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/RadioImageList.kt
GaeaKat: Seems not using an extra composable also did that trick though

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
GaeaKat: I was having trouble with having the orange selection box going between the image and the text, and sticking the right size, partially because of the text box sizing. them all being in the box meant the text had no idea of the image size without this code.

GaeaKat: Looking at it, you have removed the box that does that orange cover?

westnordost: uhh... I don't think so. At least that wasn't my intention.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/ImageList.kt
westnordost: I'd say, arguably, this logic doesn't belong there, it belongs outside of the `ImageList`.

I.e.

```kotlin

ImageList(
  items = items,
  onSelected = { newItems ->
    if (plastic in newItems) showADialog(...)
    else items = newItems
  }  
)
```

Then again... I mean, this is how it usually should be done. But in this case, maybe that wouldn't work, as we actually want to respond to the item just having been clicked, not respond to the state... hm. Well, I guess the ramp part can be *very well* done with the approach I mentioned. But the recycling thing is more difficult.

GaeaKat: Only thing I can think of is pass a secondary argument of clickeditem that most implementations of onSelected ignore?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt
GaeaKat: https://github.com/streetcomplete/StreetComplete/blob/a16543cb49cc9090a1b0d2e91e83826fad5d6125/app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/common/image_select/SelectableImageItem.kt#L118-L122

This box was there as it needed to be above image but below the text, to give the orange overlay.

GaeaKat: I have however figured out how to remove that box alltogether, and remove the imageWidth!

GaeaKat:
```kotlin
Image(
                    bitmap = bitmap,
                    contentDescription = title,
                    modifier = Modifier
                        .align(Alignment.Center)
                        .onSizeChanged {
                            imageWidth = it.width
                        },
                    colorFilter = ColorFilter.tint(SelectionColor.copy(alpha = animatedAlpha),
                        BlendMode.Color)
                )
                ```

**Comment by GaeaKat** [comment]
Ok changed a couple, waiting for feedback about selectedItem before I can fix the rest!

File: app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt (hunk 2 above)
Comment (westnordost) [reply app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt]: Both. They are the same color. But the frame should not have any alpha, the selection has some.-

Comment (GaeaKat) [reply app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt]: They aren't the same colour though, 66 ff 66 and bb ff 66 . unless it is alpha then red green blue?

Comment (GaeaKat) [reply app/src/androidMain/kotlin/de/westnordost/streetcomplete/ui/theme/Color.kt]: ... yeah thats alpha, sorry! ok will change

**Comment by GaeaKat** [comment]
Screenshot 2025-06-07 at 7 25 57 PM
I currently have this issue ...

**Comment by westnordost** [comment]
oh, i caused that, didnt i? will fix it tmrw

**Comment by GaeaKat** [comment]
> oh, i caused that, didnt i? will fix it tmrw 

It's not only the weird rounding, using the MaterialTheme styles is messing with the cells text too unfortunatly

**Comment by westnordost** [comment]
I changed the text style deliberately, though

**Comment by westnordost** [comment]
(wrong button)

**Comment by GaeaKat** [comment]
Unfortunatly the text style seems to have lead to the text getting rather badly word-wrapped

**Comment by westnordost** [comment]
Sorry, I hope I will have time today to look at it. Badly word-wrapped is... I guess a general problem though. After all, other, more kompliziert languages exist which in general have longer wording. I wonder if a general solution can be found for this general problem. I'd hope that compose would be able to somewhat intelligently add hyphens at ~correct locations depending on the locale :-/

**Comment by westnordost** [comment]
Regarding some of the other stuff, it would be great if it was possible to put the custom layout stuff into a generic `Grid` composable (that has a somewhat similar interface to `LazyGrid`) so that we can switch it out later without touching `ImageList` much.
(Can write a bit more about my train of thought later when I come back.)

**Comment by GaeaKat** [comment]
> Regarding some of the other stuff, it would be great if it was possible to put the custom layout stuff into a generic `Grid` composable (that has a somewhat similar interface to `LazyGrid`) so that we can switch it out later without touching `ImageList` much. (Can write a bit more about my train of thought later when I come back.)

Honestly LazyGrid has a bigger interface then I thought when I looked into it, with there not being like a itemsPerRow but a columns argument that can take a GridCells.adaptive or GridCells.fixed etc.

File: app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/tourism_information/AddInformationForm.kt [hunk 9]
```diff
@@ -1,18 +1,33 @@
...
 
 class AddInformationForm : AImageListQuestForm<TourismInformation, TourismInformation>() {
 
     override val itemsPerRow = 2
 
     override val items = TourismInformation.entries.map { it.asItem() }
 
+    override val itemContent =
+        @Composable { item: ImageListItem<TourismInformation>, index: Int, onClick: () -> Unit, role: Role ->
+            key(item.item) {
```
westnordost, line 19: what does this `key(item.item)` do? Can't this be outside of this function, i.e. in `AImageListQuestForm.kt` line 118? (Or `ImageGrid.kt` line 32? I have no idea what it does, but it is certainly some boilerplate)

westnordost:
Well, we don't need all those parameters. The issue right now is, that `ImageGrid` and the code around it (in `AImageListQuestForm`) is overcomplex.

Most of this complexity is directly due to you having had to fit the composable into the current structure with the fragments, the Android resources and all:
I.e. you couldn't use `LazyGrid` because of the nested scrolling issue.
Then, there is all this wiring to `AImageListQuestForm` and you kind of needed to implement the API that `AImageListQuestForm` is offering even though this might not even be necessary (e.g. that `maxSelectableItems` can be anything else than 1 or infinite). And this is why it implements all the things: multi-selection, but also single-selection, but also allows for custom logic. Plus, of course, since it was possible with `AImageListQuestForm` and e.g. the recycling quest also inherited from `AImageListQuestForm`, you implemented a way to overload this behavior. The emojis in the smoothness quest required you to add a whole new layout that handles that because the `DisplayItem<I>` data structure is still Android-view-based.

In the short/medium term, we are of course replacing all of this `AImageListQuestForm`, `AbstractOsmQuestForm` etc., so, we shouldn't actually put so much effort into making it work well with the current forms.
And of course, we'll have to untangle this, then.

So, separating the layout part from the `ImageGrid` part is something that we could do already now.

Maybe, at least for the more complex widgets, we should just not incorporate it into the current structure. There's no need to push the new compose-based widgets into production ASAP, especially if it complicates things. Instead, we can make it work independently from that, on a greenfield, with a nice API from the get-go and make use of these widgets only later.

Regarding hyphens, should be solvable with this:  https://developer.android.com/reference/kotlin/androidx/compose/ui/text/style/Hyphens

(I am just contemplating whether we should just apply `Hypen.Auto` to all our text styles in `Typography.kt`? Anyway, I will take care of this outside of this PR)

---

So, uhh, given the complexity of this and the many lines of code you had to change, I am a bit worried about bug regressions. Unnecessary regressions, after all, because as explained, there's no need to push this into production ASAP. What's important is to have the widgets to rebuild the UI in compose available and ready.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/tourism_information/AddInformationForm.kt
GaeaKat: it basically just tells compose, to recompose the display when that object changes, it is needed in the ones that can change on orientation or they don't change , I thought it was good practice to keep them elsewhere too

GaeaKat:
Ok, so maybe the image list is kind of one to do once the rest is done you thinking if I understand that right?

westnordost:
Not quite, I was thinking that widgets could be first and foremost developed on a greenfield, without necessarily integrating it into the current forms so that these widgets don't (necessarily) grow into swiss-army-knife composables just to suffice the current APIs.

On a greenfield, we could just use LazyGrid, we could solve the emoji thing by having a data class like
```kotlin
data class DisplayItem(
  val title: CharSequence, // can be AnnotatedString or String
  val icon: Painter, // can be a resource or something dynamically created 
)
```
etc.

So, the image list can be done now, but maybe should only be deployed, i.e. wired to the parent composables when there are parent composables and not the current structure. But this is all just from the top of my head, I am unsure myself.

westnordost:
Next step, in any case, would be to also migrate (the innards of) the grouped image select and the image select(s) in the overlay forms. Because then, indeed, we could and should change the `DisplayItem` data class to be more to our needs, we could and should migrate all the image and string resources from these to multiplatform.

GaeaKat:
it kinda sounds like what is best, is if this one is closed, and just used as inspiration / guideline for the greenfield one with both imagelist and image select ?

westnordost:
for some reason I wasn't able to push to this branch, so I pushed to origin instead.

#6488

Issue #5481, opened 2024-02-10
OSMOSE OpenData merge quests


XioNoX:
Hi,

In addition to running consistency checks on the OSM database, OSMOSE handles 3rd party OpenData sources for integration.

For example, implemented in https://github.com/osm-fr/osmose-backend/pull/2143
Here is a map of all the bicycle parking that are present in my city's OpenData but are not in OSM, as well as the ones that are present, but could be improved (eg. missing tags, like `capacity`)
https://osmose.openstreetmap.fr/en/map/#source=448196&zoom=14&lat=48.39667&lon=-4.47183&item=xxxx&level=3 
The same data as a table : https://osmose.openstreetmap.fr/en/issues/open?source=448196
Note that this url filters on `source=448196` but only filtering on `item=8150` would show the data for all sources and thus all cities.

Additionally APIs are available, (and support bbox filtering), for example:
https://osmose.openstreetmap.fr/api/0.3/issues?item=8150&bbox=-4.493634,48.382699,-4.483377,48.390152 for all the bicycle parking related "issues" in an area.
A given "issue" : https://osmose.openstreetmap.fr/api/0.3/issue/86fe9024-0549-7f06-63e6-ecbd824896cb
matching : https://osmose.openstreetmap.fr/en/issue/86fe9024-0549-7f06-63e6-ecbd824896cb
Note that there are already some translations in.
Full API doc: http://osmose.openstreetmap.fr/api/docs

Following up on this use-case, **it would be extremely convenient to offer to StreetComplete users a "yes/no" quest such as "Is there a bicycle parking at this location" ?** A `yes` would add the node and mark it as "resolved" on OSMOSE, a `no` would mark it as "false positive".

Multiple scenarios on which extra tags to add to the node:
* Either we don't want to trust the OpenData further, and StreetComplete only adds `amenity=bicycle_parking` and leaves it to the following quests to add the capacity or if it's covered
* We fully trust the OpenData, and automatically adds the tags suggested by OSMOSE
* We display more info to the user (eg. "is there a 10 spots bicycle parking here?")

Of course the 3rd option makes it more complex to generically support OSMOSE mergers while keeping it user-friendly.
This also raises the question on how to handle the cases where the node is already in OSM but is missing some tags (or they're invalid). We can filter those with `class=4` https://osmose.openstreetmap.fr/fr/issues/open?source=448196&class=4
* Easiest 1st step is to ignore them
[...]

PR #6302, opened 2025-06-04
AllThePlaces quests

matkoniecz:
It is an attempt to add new quest type. It would take data from https://github.com/alltheplaces/alltheplaces that were processed with https://codeberg.org/matkoniecz/list_how_openstreetmap_can_be_improved_with_alltheplaces_data

you can see processed data at https://matkoniecz.codeberg.page/improving_openstreetmap_using_alltheplaces_dataset/

-----------------------------------------

AllThePlaces POI quest would show to users as "here supposedly POI exists here - can you confirm its existence and specify an exact location?"

Also, it enables improvements to opening hours quest - to detect when OSM data and AllThePlaces data mismatches and to produce quest, before POI age crosses current threshold.
One more possible feature is to even prefill opening hours quest (if no data was collected yet) by AllThePlaces opening hours data (this one is not implemented yet at all).

Relevant discussions include https://community.openstreetmap.org/t/improving-openstreetmap-shop-coverage-with-alltheplaces/119979 and https://community.openstreetmap.org/t/what-you-think-about-importing-opening-hours-data-from-alltheplaces/120608 and https://osmfoundation.org/wiki/Licensing_Working_Group/Minutes/2023-08-14#Ticket#2023081110000064_%E2%80%94_First_party_websites_as_sources and https://github.com/alltheplaces/alltheplaces/issues/5133 and https://github.com/alltheplaces/alltheplaces/issues/8790

Would potentially open way for #5481 (but review of licensing is blocker for now anyway there) and similar ones, but I have no specific plans to work on either for now.

Review of specific spiders like done at https://community.openstreetmap.org/t/what-you-think-about-importing-opening-hours-data-from-alltheplaces/120608/72 could be helpful if anyone is interested. (it is stored in https://codeberg.org/matkoniecz/list_how_openstreetmap_can_be_improved_with_alltheplaces_data/src/branch/master/reviewed_data_sources.py )

If you see bogus data at https://matkoniecz.codeberg.page/improving_openstreetmap_using_alltheplaces_dataset/ reporting it also would be helpful. Probably in https://community.openstreetmap.org/t/improving-openstreetmap-shop-coverage-with-alltheplaces/119979 or https://codeberg.org/matkoniecz/list_how_openstreetmap_can_be_improved_with_alltheplaces_data/issues (@westnordost probably prefers to not get here notifications about all bugs in ATP or my code). Also, if someone is interested in their area but spiders are all disabled now due to lacking review: let me know, I can review at least some.

https://github.com/Helium314/SCEE/pull/523 was another attempt to do this, https://github.com/alltheplaces/alltheplaces/issues/6787 was discussing making such ATP-OSM comparison a part of the ATP project. Both seem dead as of now.

------------------------------------------------------------

I started building infrastructure for handling data from new API in SC, mostly copying how note handling is done (all this AtpDao, AtpTable, AtpApiClient)

I am now loooking through OsmQuestController/OsmNoteQuestController and I initially started building AtpQuestController.

So far all code is general, littered with TODOs and there is still ATP-specific  code to be actually written. And current code is nonfunctional, quest is not even yet shown on the map. AI is not called, API response is for now hardcoded.

But maybe someone can take a look whether general layout makes sense? 

If there is some serious general structure issue then maybe pointing it now would be nice. But I am also entirely fine with people waiting until proper code exists.

-----------------------------------------

Disclaimer: my work on ATP-OSM matching to enable this quest and quest itself is sponsored by European Union via NGI0 via NLnet foundation

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -258,7 +258,7 @@ fun questTypeRegistry(
         quest always gets a new sequential ordinal.
      */
     // TODO LATER: move this quests in appropriate location as far as priority goes
-    176 to CreatePoiBasedOnAtp,
+    200 to CreatePoiBasedOnAtp, // reduce id if needed TODO
 
     /* always first: notes - they mark a mistake in the data so potentially every quest for that
     element is based on wrong data while the note is not resolved */


app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpDao.kt
```diff
@@ -0,0 +1,116 @@
...
+        LAST_SYNC to nowAsEpochMilliseconds()
+    )
+
+    private fun CursorPosition.toAtpEntry(): AtpEntry {
+        val tagsInOsmText = getStringOrNull(OSM_TAGS)
+        val tagsInOsm = if(tagsInOsmText == null) {
+            null
+        } else {
+            Json.decodeFromString(tagsInOsmText)
+        }
```
westnordost: That's what `let`  is for

```suggestion
        val tagsInOsmText = getStringOrNull(OSM_TAGS)?.let { Json.decodeFromString(it) }
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpDao.kt
westnordost, line 1: Tests are missing.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/DatabaseInitializer.kt
```diff
@@ -254,6 +258,9 @@ object DatabaseInitializer {
             db.deleteQuest("AddParcelLockerPickup")
             db.deleteQuest("AddShoulder")
         }
+        if (oldVersion <= 19 && newVersion > 19) {
+            db.exec(AtpTable.CREATE)
```
westnordost, line 266: You forgot to also create the index

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpEntry.kt
```diff
@@ -0,0 +1,22 @@
...
+
+@Serializable
+data class AtpEntry(
+    // should we have separate Dao, Downloader, Controller etc for cases
+    // where it represents a new OSM Element and cases where it represents
+    // mismatching opening hours data?
+    val position: LatLon,
+    val id: Long,
+    val osmElementMatchId: Long?,
+    val osmElementMatchType: String?, // surely there is a better way to do this rather than keeping object id and object type in separate variables
```
westnordost: no, that's fine for the tables. `EditElementsTable` also has it like that.

For the data structure, there is `ElementKey`.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursAtp.kt
```diff
@@ -0,0 +1,206 @@
...
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+
+class AddOpeningHoursAtp(
+    private val getFeature: (Element) -> Feature?,
+    private val atpOpeningHoursDao: (Element) -> Boolean
+) : OsmElementQuestType<OpeningHoursAnswer> {
+
+    /* See also AddWheelchairAccessBusiness and AddPlaceName, which has a similar list and is/should
+       be ordered in the same way for better overview */
+    private val filter by lazy { ("""
```
westnordost: Did you copy-pasta that from the opening hours quest? That's of course not acceptable, if it is 100% shared, it should be defined in a shared location.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -240,6 +251,8 @@ fun questTypeRegistry(
         even if the quest's order is changed or new quests are added somewhere in the middle. Each new
         quest always gets a new sequential ordinal.
      */
+    1181 to AddOpeningHoursAtp(getFeature, atpOpeningHoursDao),
+    1182 to CreatePoiBasedOnAtp(),
```
westnordost: Please read the comment right above. You should also place this at an appropriate location

westnordost (commented):
Okay, let's focus on that *add All the places-place quest* first. Maybe you should just delete the opening hours one for now, to keep the size of the PR managable. It seems to be a stub right now anyway. The PR will be large in any case.

So, you want to introduce a new quest type, one that is not generated from OSM data. For this:

- `AtpDownloader` hooks into `Downloader`

- `AtpDownloader` tells `AtpController` to persist the data

- The responsibility of `AtpController` is to wrap `AtpDao` and provide the ability to listen to updates of the data. 

- `AtpDao`'s responsibility is to load/save the ATP data from the database

- Now, you need a `AtpQuestController`. It listens to updates from `AtpController` and `MapDataWithEditsSource` and after whatever logic you need to produce quests from these two data sources, in turn creates quests of type `AtpQuestType`.

- `VisibleQuestsSource` will listen to updates from `AtpQuestController` to include those quests for display

- `AtpQuestType` inherits from `QuestType` and `ElementEditType`, but is not simply a `OsmElementQuestType` which makes this all so complex. It cannot be a `OsmElementQuestType` because it doesn't work *on* OSM elements (it works on the *absence* of OSM elements. It will be an `object`, just like the `NoteQuestType`. There is just *one* AtpQuest, after all. 

- You need a `AtpQuestForm` in which whatever data needs to be recorded is recorded. On quest completion, if I understood this correctly, it will create a new OSM element. So, the UI will then create a new `ElementEdit` at the `ElementEditsController`.

- The addition of a new `ElementEdit` to `ElementEditsController` will then automatically notify `MapDataWithEditsSource` of the update, thus triggering an update to `AtpQuestController` too. Since the `AtpQuestController` checks if there is a (similarily named) POI already a the location of the Atp POI, it will tell its listeners (i.e. `VisibleQuestsSource`) that this one quest (the user just solved) has been removed. That is how it works.

Now, regarding `AtpQuestController`, you might have noticed that `OsmNoteQuestController` and `OsmQuestController` look quite different. The former has no backing DAO in which note quests are persisted. This is because the quest creation logic for `OsmNoteQuestController` is relatively simple, so it basically just transforms the incoming data (notes) to note quests.
`OsmQuestController` persist the quests instead because the logic is much more involved (all this quest filter stuff) and thus also has to manage adding/removing quests based on the incoming data.
I can't say which of the two approaches would be more eligible for Atp, but I'd suggest you try the notes approach and see if the performance will be adequate.

---

That opening hours quest by the way doesn't need all this. It would (probably) be a normal `OsmElementQuestType` which just additional taps into the Atp data in the same vein as the old now-deleted improveOsm oneway quest did.
So, it would still need the downloader, the controller and the dao.

matkoniecz:
> That opening hours quest by the way doesn't need all this. It would (probably) be a normal OsmElementQuestType which just additional taps into the Atp data in the same vein as the old now-deleted improveOsm oneway quest did.

I am not sure. From my understanding of https://github.com/streetcomplete/StreetComplete/pull/5725/files code it simply calls API during quest creation in

```
-    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
-        val bbox = mapData.boundingBox ?: return emptyList()
-
-        val trafficDirectionMap = try {
-            runBlocking(Dispatchers.IO) { trafficFlowSegmentsApi.get(bbox) }
-        } catch (e: Exception) {
-            Log.e("AddSuspectedOneway", "Unable to download traffic metadata", e)
-            return emptyList()
-        }
```
without caching or reusing this data. So if someone runs few downloads in the same area they need to download data each time.

Is it actually fine? At least for start?

I though about something more elaborate that will not call API if it already got data for given area already and it is not too old. But maybe I am overengineering and few extra API calls are fine? And such optimisation can be added later?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursAtp.kt
matkoniecz: OK, there should be TODO for it somewhere (added now). But for now I am still unsure should it be a separate quest or added to main opening hours quest.

So have not tried to deduplicate content yet.

westnordost:
Well, yes, you are correct. This architecture came from a time when the app was not really working 100% offline. So nowadays, for the opening hours quest, the data should be downloaded independently (not called in `getApplicableElements` but done in the `Downloader`) and then persisted. That opening hours quest would then use the persisted data. 

It is still easier than the "add atp place" quest, because it is still a OsmElementQuestType

matkoniecz:
So, there was progress!
Now a lot of TODOs got processed (and many got added, it is far from being complete).
Quests can be actually answered now.
Though I am still not calling API but hardcoding data in app, so yes, it is far from being complete.

I have some open, not solved questions:
(1) should I make two opening hours quests, one as is and second with ATP-powered detection of outdated opening hours? Or to extend current one, so it will appear also when ATP claims opening hours to be different? 

Two quests: 
- would allow user to disable ATP if it gives bad hints in their area (on the other hand I would fix/tweak it if we get reports, with preference to removing false positives)
- make easier for users to notice when ATP is giving bad hints
- makes easier to notice quests in changesets to see when something is wonky and it will make easier for me to monitor edits for bad data once it is released

But two quests for the same thing may introduces confusion for users, and users may prefer to not care how rules and detection work at all.

In either case it would be probably nice to prefill suggested opening hours by ATP data if currently OSM has nothing.

(2) how can I properly handle element highlighting configuration? I have something that works ( https://github.com/streetcomplete/StreetComplete/pull/6302/commits/d0fb475a7e3d42005312e0a7cc802fbda6fd7f73 ) but I am not exactly happy about code there.
For now I cannot decide between leaving this as is or adding interface where both Element and AtpEntry would implement and would give access to location, tags, comparing with element to check is it the same. Or is there another and better way to do this?

(3) is it OK to start with this ATP-adding quest disabled by default? To test, initially, on people opting in and aware what is going on.
I spend a lot of effort already on ensuring it will not suggest garbage but I assume that people will find more issues once it goes live.
And enable it by default once initial testers stop finding issues?

(4) Currently in many places it is tightly bound to specifically ATP (and partially it is a proper way to do things, I think). In general it is without out-of-the-box support for pulling other data sources in a similar fashion.
I think it is fine, at least for now. Especially as there is no other data source that is known to be usable and impossible to integrate into ATP.
This is complex enough without guessing what would be needed to support API of unknown structure, that may be never ever used.
But in some places there is structure for more general handling. Is it fine as is, or should it be removed? Or should it be supported more widely and ensure that next one can be easily added?

(5) I know it was mentioned before - but maybe using Atp instead of AllThePlaces is fine? Names of variables and functions get a bit unwieldy with full form.
I will migrate it if it would be confirmed that it should be done (or if I will get no answer).

matkoniecz:
BTW, @westnordost - thanks so much for hints in https://github.com/streetcomplete/StreetComplete/pull/6302#pullrequestreview-2896794560 ! I was getting a stuck/confused and this allowed me to progress easily.

matkoniecz:
Also, I am planning to disable note making for this quest. If ATP entries are insane, wonky, broken or otherwise silly, then OSM note is not place to report it.

I thought about adding option "email maintainer of this" that would allow user to create email send to me, but not sure is it OK to do it.

westnordost:
Disabling note-making is fine, as the note-making is only necessary as a mechanism to inhibit a non-answerable quest being shown to the next user. If ATP quests are always answerable, then, notes are not necessary. "Email maintainer of this" indeed sounds bad, instead, there might as well be "ATP notes", i.e. a mechanism with which a user can send free text to a TBD API of yours.

>  (1) should I make two opening hours quests, one as is and second with ATP-powered detection of outdated opening hours? Or to extend current one, so it will appear also when ATP claims opening hours to be different?

I'd prefer a separate PR for the opening hours quest. The PR is already huge enough (i.e. time consuming to review) as it is. In general, it sounds like ATP could trigger an opening hours quest to be re-asked early, but I am missing information for making an informed decision. After all, what should the opening hours quest then show in that case for the "is this still correct"? - The current OH info, or the (new) OH info as it comes from ATP?
Also, depends very much on the capability of your TBD API - If it triggers the general OH quest early (i.e. now), how to tell ATP "no, it's okay, keep current opening hours" without that quest immediately re-appearing?

> (2) how can I properly handle element highlighting configuration? I have something that works ( https://github.com/streetcomplete/StreetComplete/commit/d0fb475a7e3d42005312e0a7cc802fbda6fd7f73 ) but I am not exactly happy about code there.

Interfaces.

Create an interface like e.g. this
```kotlin
interface HighlightsElements {
  fun getHighlightedElements(
    element: Element,
    mapDataWithEditsSource: MapDataWithEditsSource
  ): Sequence<Element> = emptySequence()
}
```

and implement it both for `OsmQuest` and `AtpQuest`.

> (3) is it OK to start with this ATP-adding quest disabled by default? To test, initially, on people opting in and aware what is going on.

Sure.

> (4) Currently in many places it is tightly bound to specifically ATP (and partially it is a proper way to do things, I think). In general it is without out-of-the-box support for pulling other data sources in a similar fashion.
> I think it is fine, at least for now. Especially as there is no other data source that is known to be usable and impossible to integrate into ATP.
> This is complex enough without guessing what would be needed to support API of unknown structure, that may be never ever used.
> But in some places there is structure for more general handling. Is it fine as is, or should it be removed? Or should it be supported more widely and ensure that next one can be easily added?

Well, generally, I agree. It doesn't make sense to introduce additional complexity simply on the prospect that maybe later something like ATP but not quite ATP is added in the future.
Structures can be generalized *then*, when this would actually happen. (I guess you are hinting at Osmose quests.) Optimizing for a potential extension in the future usually doesn't go well, because one usually can not anticipate how a future feature will look like exactly.
Anyway, that doesn't meant that things (functions, classes, ...) that are already general because they do something general now should be remade to *not* be general anymore. Now I haven't looked at your code yet. If you give me a concrete example, I can probably give you a concrete answer.

> (5) I know it was mentioned before - but maybe using Atp instead of AllThePlaces is fine? Names of variables and functions get a bit unwieldy with full form.

It's fine. Don't forget (short but useful) doc comments for public classes.

matkoniecz:
> I'd prefer a separate PR for the opening hours quest. The PR is already huge enough (i.e. time consuming to review) as it is. In general, it sounds like ATP could trigger an opening hours quest to be re-asked early, but I am missing information for making an informed decision. After all, what should the opening hours quest then show in that case for the "is this still correct"? - The current OH info, or the (new) OH info as it comes from ATP?

OK, I will offload it into separate one (though it actually adds minimal amount of code compared to other things)
Will take a bit of detangling.

> Also, depends very much on the capability of your TBD API - If it triggers the general OH quest early (i.e. now), how to tell ATP "no, it's okay, keep current opening hours" without that quest immediately re-appearing?

1) it will update check date and therefore OSM data will stop matching expected tags (implemented already)

```
                // if tags changed, then report may be not applicable anymore
                it.tagsInOSM?.get("brand") == element.tags["brand"] &&
                it.tagsInOSM?.get("name") == element.tags["name"] &&
                it.tagsInOSM?.get("opening_hours") == element.tags["opening_hours"] &&
                it.tagsInOSM?.get("check_date:opening_hours") == element.tags["check_date:opening_hours"] &&
                it.tagsInOSM?.get("opening_hours:signed") == element.tags["opening_hours:signed"]
```

2) it will report back to my API that it was nonsense and API will stop showing it also to others (one of ten million TODOs)

> After all, what should the opening hours quest then show in that case for the "is this still correct"? - The current OH info, or the (new) OH info as it comes from ATP?

currently OSM OH data is shown, because I was unsure and not changing things was easier to do

though in cases where OSM has no OH data and ATP has, showing it may be nice

matkoniecz:
> > (2) how can I properly handle element highlighting configuration? I have something that works ( https://github.com/streetcomplete/StreetComplete/commit/d0fb475a7e3d42005312e0a7cc802fbda6fd7f73 ) but I am not exactly happy about code there.

> Interfaces.
> 
> Create an interface like e.g. this
> ```kotlin
> interface HighlightsElements {
>   fun getHighlightedElements(
>     element: Element,
>     mapDataWithEditsSource: MapDataWithEditsSource
>   ): Sequence<Element> = emptySequence()
> }
> ```
> 
> and implement it both for `OsmQuest` and `AtpQuest`.

that solves only part of the problem

because Element cannot be passed to `getHighlightedElements` in Atp (as there is no element to be operated on and even empty one cannot be created) but changing all `getHighlightedElements` function signatures to `Element?` seems really ugly (and may cause nulls to slip through in future)

westnordost:
Why would it be ugly to change to `Element?`? No, looks like the correct way to do it. That doesn't mean that the interface changes for those quest types that inherit from `OsmElementQuestType`s. It's only a few functions in MainActivity or so.

matkoniecz:
A bit other way around: `element: Element,` is problematic as MainActivity cannot provide an element for quest that operates on lack of element. It cannot also pass null.

Maybe creating and passing `val dummyNode = Node(-1, LatLon(0.0,0.0))` would be fine?

westnordost:
No, it wouldn't be fine.

Err, I think you deleted some comments because you've figured it out, hopefully?

Anyway, as to your last comment... uh... I hope you'll figure it out, too. You pass an `element` if it is an `OsmQuest`, and you pass no element if it is not. That does neither sound problematic nor complicated to do. (I am kind of afraid to review the code if you are struggling with this 😬)

(After short review of https://github.com/streetcomplete/StreetComplete/pull/6302/commits/84ced5d712b5e2ec43036cea2626fe996f4f3ed2 : 

Okay, maybe there was a misunderstanding.

**I wrote that the interface I suggested to be implemented by `OsmQuest` and `AtpQuest`, not `OsmQuestType`!**)

matkoniecz:
I will look at it again. Somehow this part unusually confuses me.
I hope that it is not representative of my coding ability.

> Err, I think you deleted some comments because you've figured it out, hopefully?

Yeah, I noticed that I posted something that makes no real sense. I guess you were checking email notification?

matkoniecz:
tl;dr I am still confused, somehow. I will now go to tend unfinished/failing tests. If code mocks me anyway, I prefer to deal with `mock()` instead.

Feel free to ignore this below but if someone wants to unconfuse me, feel free to read on.

--------------------------------------------------------------------------

status: this time I have reread it and tried coding it, I remain confused. Sorry. I still hope that it is not indicating that I somehow lobotomized myself.

The problem with

```
interface HighlightsElements {
  fun getHighlightedElements(
    element: Element,
    mapDataWithEditsSource: MapDataWithEditsSource
  ): Sequence<Element> = emptySequence()
}
```

interface is `element: Element` - it cannot be called by `MainActivity` for new ATP quest as it has no `Element` to pass

it also cannot pass null, as changing signature to `Element?` it would change interface of all normal quests displaying nearby objects and use of `!!`
Or would require equivalent of `getHighlightedElementsGeneric` (this hack is shown below)

As far as I see the same problem is present also when interface is implemented by `OsmQuest` and `AtpQuest`, not `OsmQuestType` - `getHighlightedElements` would be still called by `MainActivity` and still there would be problem how to pass null.



>  You pass an element if it is an OsmQuest, and you pass no element if it is not. That does neither sound problematic nor complicated to do

that is not problematic, even https://github.com/streetcomplete/StreetComplete/commit/d0fb475a7e3d42005312e0a7cc802fbda6fd7f73 is doing that in

```
    private fun showHighlightedElements(quest: OsmQuest, element: Element) {
        return showHighlightedElementsShared(
            quest,
            element.tags,
            element,
            quest.type.highlightedElementsRadius
        )
    }

    private fun showHighlightedElementsAroundAtpEntryQuest(
        quest: CreateElementQuest,
        atpEntry: AtpEntry,
    ) {
        val tags = atpEntry.tagsInATP
        showHighlightedElementsShared(quest, tags, null, quest.type.highlightedElementsRadius)
    }

    private fun showHighlightedElementsShared(
        quest: Quest,
        tags: Map<String, String>,
        element: Element?,
        highlightedElementsRadius: Double,
    ) {
```
`showHighlightedElementsShared` calls `getHighlightedElementsGeneric` and there problems really begin:

use of `getHighlightedElementsGeneric` with something that is between ugly kludge and horrific hack


```
    override fun getHighlightedElementsGeneric(element: Element?, getMapData: () -> MapDataWithGeometry): Sequence<Element> {
        return getHighlightedElements(element!!, getMapData)
    }

    fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) : Sequence<Element> = emptySequence()
```

so I hope there is a better way to do this

westnordost:
It's quite a deep point in the code, so I can understand when one is confused.

Yeah, change that interface to accept `Element?`.

`OsmQuest` implements it as such
```kt
override fun getHighlightedElements(
    element: Element?,
    mapDataWithEditsSource: MapDataWithEditsSource
): Sequence<Element> {
    if (element == null) return emptySequence()

    return type.getHighlightedElements(element) {
        val bbox = geometry.bounds.enlargedBy(type.highlightedElementsRadius)
        mapDataWithEditsSource.getMapDataWithGeometry(bbox)
    }
}
```

(Now this is still a little suboptimal as the geometries of `getMapDataWithGeometry` are actually thrown away and `MainActivity` has to get the geometry one-by-one itself. However, since we are caching that data in memory, it should have no performance repercussions. Long-term, proper solution would be to put element+geometry into a data structure together, e.g. `data class ElementWithGeometry(val element: Element, val geometry: ElementGeometry)`)

matkoniecz:
Now I removed last of utterly horrific hacks!

There are still many scattered TODOs and some ugly stuff but general flow of quests should be now working.

"API" is for now covering just [Kraków, Poland](https://www.openstreetmap.org/#map=14/50.06565/19.96018) (see also map below) - but quest download, instantiation, solving, hiding, unhiding works as expected.

The biggest blockers are
- bidirectional API that accepts also responses "this does not exist" (not part of this PR or repo)
- global coverage of API (not part of this PR or repo)
- StreetComplete quest able to send back to API that on selecting options that the object exists or that it is not there at all (currently it just hides quest)
- whatever is mentioned in remaining TODOs

And I think that current state if the last one (just hiding quest is fine_. Though I want to add it before it gets to users.

(note that individual spiders needs to be reviewed, so in general areas have only few entries, some none despite ATP having data - I plan to work on it after MVP running - for what is expected to be used now see https://matkoniecz.codeberg.page/improving_openstreetmap_using_alltheplaces_dataset/ )

screen-2025-06-21-14-46-53

westnordost:
I'm relieved that this PR shrunk again to 2.6k lines of code 😅

Still, huge, so let me know about any any completely done parts, so that I can already review those in parallel to you developing the other parts.

matkoniecz:
> I'm relieved that this PR shrunk again to 2.6k lines of code 😅

yes, before I implemented actual API I was hardcoding sample of locations in the source code

matkoniecz:
And for review - main data flow with quest download, instantiation, solving, hiding, unhiding works should be almost finished. It also works as expected in my initial testing. Though I am still working on automated tests and few TODOs, so I am fine with review waiting a bit.

The biggest blockers are
- bidirectional API that accepts also responses "this does not exist" (not part of this PR or repo)
- whatever is mentioned in remaining TODOs
- StreetComplete quest able to send back to API that on selecting options that the object exists or that it is not there at all (currently it just hides quest)
- 
And I think that current state if the last one (just hiding quest) is fine. Though I want to add it before it gets to users.

Global coverage of API (not part of this PR or repo) got implemented since last status update. Note that individual spiders need to be still reviewed. So some areas have only few entries, some none despite ATP having useful data. You can look at https://matkoniecz.codeberg.page/improving_openstreetmap_using_alltheplaces_dataset/ to see what would be expected to be available right now.

app/src/androidInstrumentedTest/kotlin/de/westnordost/streetcomplete/data/atp/AtpDaoTest.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test fun deleteButNothingIsThere() {
+        assertFalse(dao.delete(1))
+    }
+
+    @Test fun delete() {
+        val entry = createAtpEntry()
+        dao.put(entry)
+        assertTrue(dao.delete(entry.id))
+        assertNull(dao.get(entry.id))
```
westnordost, line 51:
```suggestion
        assertNull(dao.get(entry.id))
        assertFalse(dao.delete(entry.id))
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpDao.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    fun getAllWithMatchingOsmElement(match: ElementKey): List<AtpEntry> {
+        return db.query(NAME,
+            where = "$OSM_ELEMENT_MATCH_ID = ? AND $OSM_ELEMENT_MATCH_TYPE = ?",
+            args = arrayOf(
+                match.id,
+                match.type.toString(),
+            )
+        ) { it.toAtpEntry() }
+    }
```
westnordost: Is this in a hot path? (I.e. how often is this called?)

(Currently, it doesn't seem to be called *at all*?)

But well, if it is called often, then you should add an index on these two to `AtpTable.kt` (and subsequently in `DatabaseInitializer.kt`)

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/data/quest/VisibleQuestsSourceTest.kt
westnordost, line 1: TODO add test that confirms that `VisibleQuestsSource` is using the `AtpQuestSource` correctly

app/src/androidMain/kotlin/de/westnordost/streetcomplete/data/quest/atp/CreatePoiBasedOnAtp.kt
```diff
@@ -0,0 +1,22 @@
...
+import de.westnordost.osmfeatures.Feature
+import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.osm.mapdata.Element
+import de.westnordost.streetcomplete.data.osm.mapdata.MapDataWithGeometry
+import de.westnordost.streetcomplete.data.quest.AndroidQuest
+import de.westnordost.streetcomplete.data.quest.OsmCreateElementQuestType
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CITIZEN
+import de.westnordost.streetcomplete.osm.isPlaceOrDisusedPlace
+
+class CreatePoiBasedOnAtp() : OsmCreateElementQuestType<CreatePoiBasedOnAtpAnswer>, AndroidQuest {
```
westnordost:
```suggestion
object CreatePoiBasedOnAtp : OsmCreateElementQuestType<CreatePoiBasedOnAtpAnswer>, AndroidQuest {
```

Make this an `object`, just like `OsmNoteQuestType`

app/src/androidMain/res/values/strings.xml
```diff
@@ -723,6 +723,10 @@ Before uploading your changes, the app checks with a &lt;a href=\"https://www.we
     <string name="quest_atm_cashin_title">"Can you deposit cash at this ATM?"</string>
     <string name="quest_atm_operator_title">"Which bank operates this ATM?"</string>
 
+    <string name="quest_atp_add_missing_poi_title">This place may be here - where exactly it is located? Tap on its exact location</string>
+    <string name="quest_atp_add_missing_poi_mapped_already">"This place is mapped already"</string>
+    <string name="quest_atp_add_missing_poi_does_not_exist">"This place does not exist"</string>
```
westnordost: Instructions on how to answer the quest should not be in the question but in the quest form body.

Also, IMO better: "It's already displayed on the map" and "No such place here". 🤷

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpModule.kt
```diff
@@ -0,0 +1,17 @@
...
+    factory { AtpDownloader(get(), get()) }
+    // TODO: connect to actual bidirectional API
+    val OSM_ATP_COMPARISON_API_BASE_URL = "https://matkoniecz.codeberg.page/improving_openstreetmap_using_alltheplaces_dataset/read_only_api_prototype/"
+    factory { AtpApiClient(get(), OSM_ATP_COMPARISON_API_BASE_URL, get()) }
+    factory { AtpApiParser() }
+
+    single {
+        AtpController(get()).apply {
+        }
+    }
```
westnordost:
```suggestion
    single { AtpController(get()) }
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpTable.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    const val CREATE = """
+        CREATE TABLE $NAME (
+            ${Columns.ID} int PRIMARY KEY,
+            ${Columns.LATITUDE} double NOT NULL,
+            ${Columns.LONGITUDE} double NOT NULL,
+            ${Columns.OSM_ELEMENT_MATCH_ID} int,
+            ${Columns.OSM_ELEMENT_MATCH_TYPE} varchar(15),
+            ${Columns.ATP_TAGS} text NOT NULL,
+            ${Columns.OSM_TAGS} text NOT NULL,
```
westnordost: sure that this should be `NOT NULL`? Don't you put a `null` there if the tags are empty? (There's a TODO in `AtpDao` uttering that you are confused about why there are strings with `"null"` in the table. That's the reason.)

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/data/visiblequests/QuestsHiddenControllerTest.kt
```diff
@@ -39,27 +45,51 @@ class QuestsHiddenControllerTest {
...
+        assertEquals<Pair<QuestKey, Long>>(
+            Pair(h1.key, 250L),
+            ctrl.getAllNewerThan(123L)[2]
+        )
+        //TODO above works, below fails, why? Oh and below fails while claiming no difference
+        // assertEquals is called on listOf in few places so it does not appear due to comparing
+        // specific list instances instead of actual content
         assertEquals(
             listOf(
+                AtpQuestKey(h6.allThePlacesEntryId) to 100000,
```
westnordost:
```suggestion
        assertEquals(
            listOf(
                AtpQuestKey(h6.allThePlacesEntryId) to 100000L,
```

I am pretty sure because `100000` is an `Integer`, so the resulting `List` will be of type `List<Pair<QuestKey, Any>>`. Lists can only be equal if the type is the same.

app/src/androidInstrumentedTest/kotlin/de/westnordost/streetcomplete/data/atp/atpquests/AtpQuestsHiddenDaoTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    @BeforeTest fun createDao() {
+        dao = AtpQuestsHiddenDao(database)
+    }
+
+    @Test fun addGetDelete() {
+        assertFalse(dao.delete(123L))
+        dao.add(123L)
+        assertNotNull(dao.getTimestamp(123L))
+        assertTrue(dao.delete(123L))
+        assertNull(dao.getTimestamp(123L))
```
westnordost, line 28:
```suggestion
        assertTrue(dao.delete(123L))
        assertFalse(dao.delete(123L))
        assertNull(dao.getTimestamp(123L))
```

westnordost (commented):
Preliminary incomplete review. Looked at 28/59 files (though, mostly the small ones TBH), avoided files with TODOs but got to go now.

While scrolling, I noticed a AtpEditType or something. Why is this necessary? Shouldn't the Atp quest also result in a change of the OSM data?

matkoniecz:
> While scrolling, I noticed a AtpEditType or something. Why is this necessary? Shouldn't the Atp quest also result in a change of the OSM data?

currently "already mapped" and "no such POI" result in hiding quest

they are supposed to result in API call to record it in my API - this is not yet implemented

to support offline editing these need to be able to be hold somewhere until internet is available - and I plan to use AtpEditType for that

app/src/androidMain/res/values/strings.xml
matkoniecz: > Also, IMO better: "It's already displayed on the map" and "No such place here". 🤷 

changed

matkoniecz: > Instructions on how to answer the quest should not be in the question but in the quest form body.

BTW, I was thinking about migrating to cross selector like in things overlay and places overlay (not sure have you tried using or looked at code for now)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpDao.kt
matkoniecz: > (Currently, it doesn't seem to be called at all?)

IIRC it is needed by OH quest, I guess I should make commit reverting it and commit in OH segment restoring it and adding indexes

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpTable.kt
matkoniecz: thanks, that was it

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/data/quest/VisibleQuestsSourceTest.kt
matkoniecz: TODO added

matkoniecz:
> Will continue review of the non-TODO-containing files once these small things I found are resolved.

that new review is now processed

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/data/quest/VisibleQuestsSourceTest.kt
```diff
@@ -141,10 +160,12 @@ class VisibleQuestsSourceTest {
     }
 
     @Test fun `getAll does not return those that are invisible because of an overlay`() {
+        // TODO how ATP fits into this test? Should ATP be suppressed by some overlay? Maybe places, but I feel it is actually useful there
```
westnordost: ok, no, doesn't need to be supressed

```suggestion
```

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/data/quest/VisibleQuestsSourceTest.kt
```diff
@@ -191,6 +212,24 @@ class VisibleQuestsSourceTest {
         verifyNoInteractions(listener)
     }
 
+    @Test fun `at quests added or removed triggers listener`() {
```
westnordost:
```suggestion
    @Test fun `atp quests added or removed triggers listener`() {
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -73,6 +73,11 @@ object ApplicationConstants {
     //  e.g. AddRecyclingContainerMaterials, AddCycleway
     const val QUEST_FILTER_PADDING = 20.0 // m
 
+    /** ATP quest gets a larger padding
+     *  In some cases ATP entries are near OSM objects, but offset is larger than QUEST_FILTER_PADDING
+     *  Note that download still uses QUEST_FILTER_PADDING, that should be fine overall
+     */
+    const val ATP_QUEST_FILTER_PADDING = 50.0 // m
```
westnordost, line 80: Why should it be fine overall?

Let's say there is a supermarket already on OSM 21m to the East of an ATP quest that wants to add a supermarket. But the ATP quest is at the edge of the downloaded area (or well, 20m away from it) -> the ATP quest will still be shown.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpApiClient.kt
```diff
@@ -0,0 +1,52 @@
...
+     * @throws ConnectionException if a temporary network connection problem occurs
+     *
+     * @return the incoming atp entries
+     */
+    suspend fun getAllAtpEntries(bounds: BoundingBox): List<AtpEntry> = wrapApiClientExceptions {
+        val gathered = mutableListOf<AtpEntry>()
+
+        for (longitudeAnchor in floor(bounds.min.longitude).toInt()..floor(bounds.max.longitude).toInt()) {
+            for (latitudeAnchor in floor(bounds.min.latitude).toInt()..floor(bounds.max.latitude).toInt()) {
+                val url = baseUrl + "lat_${latitudeAnchor}/lon_${longitudeAnchor}_gathered.geojson"
```
westnordost: Uff, what, this is not really an API. It is just a bunch of files. Is this the final state of the API? Why not do it properly? It seems to me you need a proper server (and with a server I mean at least a couple of PHP scripts and a MySql database) to report false-positives.

Maybe check https://github.com/streetcomplete/oneway-data-api - the original data source also only supplied a huge CSV file IIRC correctly. I hade a PHP script that would be executed to import that CSV regularly and than a simple script that returned the data in a certain bbox.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpApiClient.kt
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Retrieve all atp entries in the given area
+     *
+     * @param bounds the area within where ATP entries should be queried.
+     *
+     * @throws ConnectionException if a temporary network connection problem occurs
+     *
+     * @return the incoming atp entries
+     */
+    suspend fun getAllAtpEntries(bounds: BoundingBox): List<AtpEntry> = wrapApiClientExceptions {
```
westnordost, line 35:
```suggestion
    suspend fun getAllAtpEntries(bounds: BoundingBox): List<AtpEntry> = wrapApiClientExceptions {
        if (bounds.crosses180thMeridian) {
            throw IllegalArgumentException("Bounding box crosses 180th meridian")
        }
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpApiClient.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val baseUrl: String,
+    //private val userAccessTokenSource: UserAccessTokenSource,
+    private val atpApiParser: AtpApiParser
+) {
+    /**
+     * Retrieve all atp entries in the given area
+     *
+     * @param bounds the area within where ATP entries should be queried.
+     *
+     * @throws ConnectionException if a temporary network connection problem occurs
```
westnordost, line 30:
```suggestion
     * @throws ConnectionException if a temporary network connection problem occurs
     * @throws IllegalArgumentException if the bounds cross the 180th meridian.
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpDownloader.kt
westnordost, line 1: I'd tend to say that if any exception happens here, it should just log it but not cancel the whole download. (Since the ATP stuff is out of my control,) if the ATP stuff is not reachable on network, it should not lead to the app being unusable cause no download goes through.

See `MapTilesDownloaderAndroid.kt`

app/src/androidMain/kotlin/de/westnordost/streetcomplete/screens/main/map/EditHistoryPinsManager.kt
```diff
@@ -133,6 +137,11 @@ private fun Edit.toProperties(): List<Pair<String, String>> = when (this) {
         MARKER_ELEMENT_ID to elementId.toString(),
         MARKER_QUEST_TYPE to questType.name
     )
+    is AtpQuestHidden -> listOf(
+        MARKER_EDIT_TYPE to EDIT_TYPE_HIDE_ATP_QUEST,
+        ATP_ENTRY_ID to atpEntry.id.toString()
+        // TODO add more context here? maybe?
```
westnordost:
```suggestion
        MARKER_EDIT_TYPE to EDIT_TYPE_HIDE_ATP_QUEST,
        ATP_ENTRY_ID to atpEntry.id.toString()
```

no, why what context? You only need as much context to create the `AtpQuestKey`, nothing more.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/atp/AtpApiParser.kt
westnordost, line 1: Why not use kotlinx.serialization.json `@Serializable`? I.e. specify the expected data structure, pass that to `Json`.
(But anyway, if this is not the final state of the API, makes no sense to optimize that)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/screens/main/map/EditHistoryPinsManager.kt
matkoniecz: I meant passing more info to the history view so undoing hiding shows more identifying info

matkoniecz:
@westnordost are you interested in making icon?

I though about radar as an idea (scan view, see https://ar.inspiredpencil.com/pictures-2023/radar-icon-flat ) - or simply a plus icon, but have not tried yet anything specific.
I can try doing it but few times you made icons for new quest so maybe this time it will also happen.

westnordost:
Hm, I don't have an idea for an icon. Yeah, maybe radar is fine

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
matkoniecz: my thinking here is that it is heuristic anyway - and any value between 20m and 80m is probably justifiable, with some preference to larger distance to avoid false positives

but sometimes you just get insane outliers (say, shop misplaced by huge distance and geolocation check failed to throw it away for one reason or another) - so some false claims are not wholly avoidable anyway

And always fetching data within 50m would require downloading far more data and then throwing it away, for marginal improvement

so checking within 50m seems justifiable if we have data anyway, but 20m is fine if data is not there

---------------------------------------------

Does it make sense? Or is it an insane rationalisation to avoid dealing with boring and annoying technical challenge?

matkoniecz:
@westnordost 

Is it expected that `app/src/androidInstrumentedTest/kotlin/de/westnordost/streetcomplete/data/atp/AtpDaoTest.kt` will have no access to helper function defined in `app/src/commonTest/kotlin/de/westnordost/streetcomplete/testutils/TestDataShortcuts.kt` ?

Due to being in a different subfolder of `src` ?

(I am irritated by `fun createAtpEntry`  and `fun atpEntry` - tried to deduplicate but failed so far)

westnordost:
yes, I think thats normal. they are separate gradle modules.

PR #6303, opened 2025-06-04
Use kotlinx.atomicfu

### LINKED ISSUE #5552: Multiplatform synchronization (Use kotlinx-atomicfu)
[`synchronized`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/synchronized.html) is a JVM-only API in Kotlin and as far as we know, there are no plans to change that.

`synchronized` is used in StreetComplete in many places in the data layer. The data layer is architecture-wise a blocking API (i.e. no coroutines, no suspending functions) and hence we can't use [`Mutex`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) for threadsafe locks. (Let's not preclude that we refactor the data layer one day to be coroutine-based although it would be a considerable effort, but for now:)

We need a multiplatform replacement for `synchronized`. The most obvious choice is the official `kotlinx-atomicfu`, which is however still marked as experimental, especially the locks.
So, while there are other things to do in preparation of #5412, it might be better to hold off towards (at least) merging a PR that implements this (because according to one maintainer of `kotlinx-atomicfu`, they are working towards stabilization)

westnordost:
... and move to commonMain those things. At very few places, I noticed that locking was done consistently (on all write-access), so I added that.

fixes #5552


westnordost:
Now, about 30% of source files have moved to commonMain. 

(Moving anything closely related to quests or overlays is blocked until the very end, because `QuestType` references Android-stuff in `createForm`. Quest-related stuff make up about 2/3 of the remaining unmoved files)

FloEdelmann:
@westnordost I think you broke the `NoteEditsUploaderTest`: https://github.com/streetcomplete/StreetComplete/actions/runs/15450318756/job/43490529418 now errors with `Unresolved reference 'NAME'`.

PR #6307, opened 2025-06-05
Fix angle normalization with non-zero startAt parameter

RubenKelevra:
The previous implementation incorrectly normalized angles when using a non-zero startAt parameter. It would first normalize to [0, 360), then attempt to adjust for startAt, which failed for many cases, especially with negative input values.

The new implementation correctly handles all cases by:
1. Shifting the coordinate system by -startAt
2. Normalizing to [0, 360)
3. Shifting back by +startAt

This ensures the result is always in the range [startAt, startAt+360).

Example of the bug: normalizeDegrees(-720.0, 200.0) would return 0 instead of the correct value 360.

### DIFF
--- a/app/src/commonTest/kotlin/de/westnordost/streetcomplete/util/math/AngleMathTest.kt
+++ b/app/src/commonTest/kotlin/de/westnordost/streetcomplete/util/math/AngleMathTest.kt
@@ -1,0 +1,47 @@
+package de.westnordost.streetcomplete.util.math
+
+import kotlin.math.PI
+import kotlin.test.Test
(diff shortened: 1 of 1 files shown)


riQQ:
This is the perfect change to add accompanying unit tests.

westnordost:
I agree with @riQQ

RubenKelevra:
Was about to add some :)

westnordost:
ah, cool!

RubenKelevra:
@westnordost ping.

westnordost (approved):
Thank you!

PR #6309, opened 2025-06-06
Max Weight to Compose format

GaeaKat:
Working on Max Weight form


westnordost:
Wow, you are relentless! :+1: 

Max weight... that has some similarities with the max height form - with the unit select, the input inside that sign graphic (for some countries) and all. Would you mind reviewing https://github.com/streetcomplete/StreetComplete/pull/6285 to make sure the implementation of both is consistent? (It's draft, not sure if he wants feedback so far, didn't yet ask)

GaeaKat:
> Max weight... that has some similarities with the max height form - with the unit select, the input inside that sign graphic (for some countries) and all. Would you mind reviewing [#](https://github.com/streetcomplete/StreetComplete/pull/6285)

I will take a look and try to match more ! I must admit I already had somewhat of a skeleton here with a bit of a tree structure after looking at how they handled the one layer of signs

the form asks for a GeneralMaxWeightSign and that returns depending on sign type
MaxWeightSign - MaxWeightGeneralSign, MaxWeightFinnishSign, or MaxWeightUSSign depending on country 
MaxTandemAxleLoadSign
MaxMassSign
MaxLoadSign

GaeaKat:
And yeah, I must admit, I want Street complete on iOS, and the best way for that is to help! That and I love the idea behind the app so helping it even if it never gets on iOS is ALSO good!

westnordost:
Well, TBH I only have had a cursory look at #6285, important basically is my feedback to @ltoenning about the structure so to not duplicate composable code for the different signs. I.e I don't know if he implemented it that way, yet:

If all those sign layouts have both an input field and a unit selector dropdown (but depending on the sign it may be in- or outside of the sign layout), that means that there should be one composable for (length) input and one composable for the unit selector, separate from each other. The e.g. MaxWeightGeneralSign composable then glues those components together

GaeaKat:
That makes sense yeah. I will spend some time looking at #6285 and try and make sure I go along with it!

westnordost:
So, for #6285 I added a lot of utilities which should make this task in particular also very much easier:
- a generic `DecimalInput` composable, a text field for decimal numbers
- a generic `SelectButton` composable, to select e.g. units
- a generic `RectangularSign` composable, for backgrounds that should look like a rectangular traffic sign

I'm in a bit in a flow now, so I could try to take over this PR and steamroll it or would you like to continue with it?

GaeaKat:
it might be best if there is new things to help with it, I am learning more about view models etc etc to help with later points

westnordost:
Okay, so I will take this over, maybe as a procastination while doing the localized names in compose because the scope of it is currently exploding... 😅

ViewModels, yeah, I think I wrote something about that in either the iOS master ticket or the ViewModel one. Not sure if it is best to work on these already, I think it would be easier to work on these in the next step, i.e. when the view is migrated to compose, then as a next step replace the logic in the Fragments with logic in the ViewModels (in the Fragments).

Otherwise, on the Jetpack Compose front, I think we ran out of smallish things to do, so the rest is somewhat more complex. The note discussion form might be fun, though, although it probably has the same issue as the image select stuff that `LazyColumn` will have problems with the nested scrolling. 

If you are interested in some busywork (I mean, work with little brain-work but "doing" a lot), there's something neat that I planned to do soon also as a sort of procastination / diversion from the more complicated jetpack compose work: move all the pure compose code to `commonMain`, i.e. the stuff in/for settings, about, profile screen and parts of the main screen. This involves (however) to move or copy the `strings.xml`s into the common compose resources and referencing those strings (`Res.strings.…`) from that code instead of Android's `R.strings.…` from these. Maybe more. Sometimes one just needs this kind of mostly brainless busywork as a diversion, IMO, but if you'd like to do this, I could also deviate to doing some OSM mapping with JOSM instead ;-)

westnordost:
I need a break from that mammoth compose migration I merged today, so I will do this next.

westnordost:
Damn, I noticed that the select button opens a dialog/popup in which one should select the signs, which is an image picker dialog.... I guess I first have to bring #6294 to completion.

Also, there is no easy way to draw a composable into an ImageBitmap to be used as Painter, so I have to create icons for all the layouts. (I actually did today).

PR #6311, opened 2025-06-06
fix: correct Vector3d division and subtraction operations

RubenKelevra:
- Fix div operator to divide each component (x/v, y/v, z/v) instead of (x/v, x/v, x/v)
- Fix minus operator to use o.z for z component instead of o.y
- Add Vector3dTest to verify corrected operations


westnordost (approved):
oh man, what a stupid mistake! Thank you again! How on earth do you find these things?

PR #6312, opened 2025-06-06
Add dirt surface support to smoothness quest

### LINKED ISSUE #5499: Include more unpaved highways in smoothness quest
**Use case**
At the moment, most unpaved highways such as tracks are not included in the smoothness quest because their surface is often `ground`, `dirt`, `earth`, etc. I think that for car users, these ways are especially interesting to be tagged with smoothness because they are often worse than `bad` and it's at this level of smoothness where it really matters which kind of car you have for a road to be passable or not.

**Proposed Solution**
I think ways with `ground`, `dirt`, `earth` and even `unpaved` surfaces can be included in the smoothness quest by using the images now used for `surface=gravel`. I think the differences between these surfaces are not very clear from the wiki anyway, so the same images can be used for all of them. The images now used for `very_bad`, `horrible` and `very_horrible` are mine and are of tracks that I tagged with `surface=ground/dirt`.

RubenKelevra:
- Add 'dirt' to supported surfaces list for road/path smoothness quest
- Implement dirt-specific smoothness descriptions
- Add synthetic surface images for good/intermediate/bad/very_bad/horrible/very_horrible levels for dirt

Fixes #5499


GaeaKat:
ooh smoothness quest, my compose image list pr at #6294 does touch on theese files a bit also!

matkoniecz:
media source and license should be recorded in https://github.com/streetcomplete/StreetComplete/blob/af4b08c72016fd0d694f00acdf1aa72439b5f8df/app/src/androidMain/res/authors.txt

RubenKelevra:
Ah thanks, forgot that one. :)

westnordost:
Thank you! Those pictures though... oof. They look so un-real.

Actually, if I remember correctly, we use the same pictures for all non-paved surfaces starting from a certain horrible-level. So, to include dirt, we actually only need photos of the "better" surfaces.

Especially for dirt surfaces, it makes sense to make photos of typical situations, like you would find on forest paths etc. (e.g. roots). And I really much prefer real photos. Aren't you keen on mountain biking? It should be possible to find plenty of examples in a typical hiking/biking spot.

RubenKelevra:
> Actually, if I remember correctly, we use the same pictures for all non-paved surfaces starting from a certain horrible-level. So, to include dirt, we actually only need photos of the "better" surfaces.

Yeah I know, but the pictures were so "rocky" that I thought it would make more sense to generate them specifically for dirt as well.

> Thank you! Those pictures though... oof. They look so un-real.

Well, they look somewhat unrealistic, but I don't think it's "that" bad. I mean, compared what we got a year before...

I think they work fine, until we get better ones captured by someone. But it's quite an annoying blocker at the moment, as we can't capture smoothness for most tracks and paths, as `ground`, `unpaved` and `dirt` are ~25 % of all surface tags. While `ground` and `unpaved` is pretty non-descriptive, I think we can assume that most of them are "dirt".

> Especially for dirt surfaces, it makes sense to make photos of typical situations, like you would find on forest paths etc. (e.g. roots). And I really much prefer real photos. Aren't you keen on mountain biking? It should be possible to find plenty of examples in a typical hiking/biking spot.

Well, I tried to keep "forests" out of the equation here, to avoid my bias of living in Germany, as SC is an international project, but if that helps with believability, I can certainly generate more believable images without this constraint.

But sadly I cannot supply images for "dirt" with a lot of variation in surface quality, as we got pretty well maintained paths and tracks here in my corner of Germany. Maybe if I'm on holiday somewhere else eventually. The difficulties here are more "persistent mud", rocky paths and extreme elevation changes.

westnordost:
or maybe look at wikimedia commons somewhere, there should be *some* pictures of the run of the mill "querfeldein-trampfelpfad"

RubenKelevra:
> or maybe look at wikimedia commons somewhere, there should be _some_ pictures of the run of the mill "querfeldein-trampfelpfad"

I actually did that last year already for an hour or so... it's really hard to find pictured of really nasty paths, because people seem more inclined to picture beautiful things, I guess.

In addition, `good` is actually also very difficult to find, as usage destroy "dirt" paths, but also creates them. `good` paths are usually `compacted` not `dirt`. (see below in the list for a potential candidate)

How about we mix this up: We use pictured for path qualities we *do* have pictured from and use synthetic ones for the rest?

This gives us at least the chance to ship it, and replacing pictures in the future is pretty simple.

So I've looked through commons/osm wiki and found the following candidates:

#### `good` 

Here's a candidate, but I'm not sure if you would argue, that it's compacted:

Farm_Track_in_North_Dale_-_geograph org uk_-_1804240

_[source](https://commons.wikimedia.org/wiki/File:Farm_Track_in_North_Dale_-_geograph.org.uk_-_1804240.jpg)_

vs synthetic:

surface_dirt_good

#### `intermediate`

Track_to_the_Outgang_-_geograph org uk_-_6954701

_[source](https://commons.wikimedia.org/wiki/File:Track_to_the_Outgang_-_geograph.org.uk_-_6954701.jpg)_

vs synthetic:

surface_dirt_intermediate

#### `bad` 
Farm_Track_through_Farm_Buildings_-_geograph org uk_-_6644612

_[source](https://commons.wikimedia.org/wiki/File:Farm_Track_through_Farm_Buildings_-_geograph.org.uk_-_6644612.jpg)_

vs synthetic: 

surface_dirt_bad

#### `very_bad`

Eroded_byway_-_geograph org uk_-_7471811

_[source](https://commons.wikimedia.org/wiki/File:Eroded_byway_-_geograph.org.uk_-_7471811.jpg)_

vs synthetic: 

surface_dirt_very_bad



#### `horrible` 

Track_alongside_field_-_geograph org uk_-_1100535

_[source](https://commons.wikimedia.org/wiki/File:Track_alongside_field_-_geograph.org.uk_-_1100535.jpg)_

vs synthetic:

surface_dirt_horrible


#### `very_horrible` 

surface_dirt_very_horrible

matkoniecz:
I want to remind about existence of https://community.openstreetmap.org/t/how-to-map-key-smoothness/109860 where there are ongoing attempts to gather images for various surfaces

RubenKelevra:
@matkoniecz I've scrolled through it. Not much about `dirt` in there except your image for very_horrible ([File:Forest road damaged by truck traffic.jpg](https://wiki.openstreetmap.org/wiki/File:Forest_road_damaged_by_truck_traffic.jpg)). 

Btw: I think your categorization may be a bit too high for that. A heavy-duty [off road vehicles](https://en.wikipedia.org/wiki/Off-road_vehicle) can easily pass that. I know that because that are not tractor tire marks. :)

matkoniecz:
Do you remember where you found that very_bad one? Seems quite fitting. Can you post it in that thread? 

And yes, these images are eye - gougingly artificial

RubenKelevra:
> Do you remember where you found that very_bad one? Seems quite fitting. Can you post it in that thread?

The sources are all linked below each image. 
> 
> And yes, these images are eye - gougingly artificial

I don't really mind that look, as long as it conveys what it needs to convey I don't have hard feelings about it. It's not like that it's pretending to be art.

Anyway, how about the pictures I've posted?

westnordost:
I agree with @matkoniecz on the AI pictures.

The `bad` and `very_bad` pictures look good, IMO. The `horrible` one is a good example, I think, but the perspective is noticeably different from the other pictures and the picture size/quality might be too small.
In any case, I'd propose to _first_ add those pictures to the smoothness gallery.

westnordost:
Anyway, I'd close this PR, but not without having added those nice pictures of bad to horrible dirt to the wiki gallery: https://wiki.openstreetmap.org/wiki/Key:smoothness/Gallery

PR #6314, opened 2025-06-06
test: fix range for 'too low' test case in CircularSectionTest

RubenKelevra:
Fix tooLow test case for CircularSection() to [-2, -1], to test intersection logic correctly with range below the base range [0, 10].


westnordost:
Thanks!

westnordost:
How do you find these things?

PR #6317, opened 2025-06-06
osm: implement non-integer range handling in Level.Range.getSelectableLevels()

RubenKelevra:
Function was only yielding values for integer ranges. Added else case to yield start and end values for non-integer ranges, aligning implementation with existing code documentation.


westnordost:
Oh wow, so much documentation and simply wrong, thank you :-)

PR #6318, opened 2025-06-06
util: handle duplicate elements in Collection.containsExactlyInAnyOrder()

RubenKelevra:
containsAll() only checked for unique element presence, not frequencies. Changed to use count maps for duplicate element comparison.

The original implementation would incorrectly return true for collections 
like [1, 2, 2, 3] and [1, 2, 3, 3] because:
- Both have the same size (4)
- Both contain the elements 1, 2, and 3

The new implementation uses groupingBy().eachCount() to create frequency maps ({1=1, 2=2, 3=1} vs {1=1, 2=1, 3=2}), which correctly identifies these collections as different due to varying element frequencies.

This ensures the function behaves as its name suggests - checking for exact content equality regardless of order.

### DIFF
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/ktx/Collections.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/ktx/Collections.kt
@@ -74,5 +74,10 @@ inline fun <T, R : Comparable<R>> Iterable<T>.indexOfMaxBy(selector: (T) -> R): 
 }
 
 // TODO clean: for tests, rather use assertEquals with sets for better errors on failure
-fun <T> Collection<T>.containsExactlyInAnyOrder(other: Collection<T>): Boolean =
-    other.size == size && containsAll(other)
+fun <T> Collection<T>.containsExactlyInAnyOrder(other: Collection<T>): Boolean {
+    if (size != other.size) return false
+    if (this === other) return true
+
+    fun Collection<T>.toCountMap() = groupingBy { it }.eachCount()
+    return toCountMap() == other.toCountMap()
+}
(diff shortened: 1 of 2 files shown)


westnordost:
Right, thank you!

That reminds me, I wanted to get rid of this function. IIRC it is only used in tests, for things like

```kotlin
assertTrue(
  actualResult.containsExactlyInAnyOrder(expectedResult)
)
```

while I'd rather use this

```kotlin
assertEquals(
  expectedResult.toSet(),
  actualResult.toSet()
)
```

I use this at a few places already.

Why is it better? Because if it fails, I get a better feedback error message from the unit test.

And yeah, I realize now that it has the same issue as the current implementation of that function... hmm... 🤔

So maybe we just need a new function

```
assertEqualsInAnyOrder(
  expectedResult,
  actualResult
)
```
...which uses your implementation but will print a useful error when the test fails.

RubenKelevra:
> That reminds me, I wanted to get rid of this function. IIRC it is only used in tests, for things like

Yeah I was curious if that's even an issue, so I searched for uses before I changed it. But there were uses in the codebase as well in `app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/edits/create/CreateNodeFromVertexAction.kt`.

That's also the reason I've chosen this more per formant variant which should run at O(n) space, O(n) time so it works for large data lists as well.

RubenKelevra:
Added some tests

RubenKelevra:
I'll add the function which is described in the TODO/in your comment, I just can't complete it today anymore.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/ktx/Collections.kt
```diff
@@ -74,5 +74,10 @@ inline fun <T, R : Comparable<R>> Iterable<T>.indexOfMaxBy(selector: (T) -> R):
...
 // TODO clean: for tests, rather use assertEquals with sets for better errors on failure
-fun <T> Collection<T>.containsExactlyInAnyOrder(other: Collection<T>): Boolean =
-    other.size == size && containsAll(other)
+fun <T> Collection<T>.containsExactlyInAnyOrder(other: Collection<T>): Boolean {
+    if (size != other.size) return false
+    if (this === other) return true
+
+    fun Collection<T>.toCountMap() = groupingBy { it }.eachCount()
+    return toCountMap() == other.toCountMap()
+}
```
westnordost:
```suggestion
fun <T> Collection<T>.containsExactlyInAnyOrder(other: Collection<T>): Boolean {
    if (size != other.size) return false
    if (this === other) return true
    return toCountMap() == other.toCountMap()
}

private fun <T>  Collection<T>.toCountMap() = groupingBy { it }.eachCount()
```

I dislike function-local functions

PR #6320, opened 2025-06-07
util: handle empty lists in List.firstAndLast()

RubenKelevra:
Return empty list if given empty list, to avoid potential NoSuchElementException from first() call.

---

Not sure if that could be triggered by the codebase atm, but to avoid potential crashes this makes more sense I think.


westnordost:
Well, I'd disagree here. `first()` also throws an exception if there is no element. That's why `firstOrNull()` exists.
But this method is called `firstAndLast()`, and not `firstAndLastOrNull()` or similar.

westnordost:
(Doesn't mean that in some cases it may not be useful to have a `firstAndLastOrNull()` that returns null if size < 2, but this is not what this PR was about. And it would need some research into how/where this function is used)

PR #6322, opened 2025-06-07
ui: ensure updateOkButtonEnablement() runs after buttons exist in ShopGoneDialog

RubenKelevra:
updateOkButtonEnablement() was called during init when getButton(BUTTON_POSITIVE) may return null. Move to show() after super.show() to ensure button state update actually executes, addressing potential timing issue.

---

UI is not my can of worms, but I think I found a bug here. 

During the init() `updateOkButtonEnablement()` is called, but since the button does not exist yet `getButton()` will return NULL here, which is masked by `?.`.

This can lead to the button not being disabled, despite the user not yet having selected an element from the list.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/shop_type/ShopGoneDialog.kt
@@ -83,6 +83,7 @@ class ShopGoneDialog(
 
     override fun show() {
         super.show()
+        updateOkButtonEnablement()
         // to override the default OK=dismiss() behavior
         getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener {
             when (selectedRadioButtonId) {


westnordost:
What? Where? I don't see any `?.` there. Also, the shop gone dialog doesn't have any radio button checked initially, so why should one want to update the ok button enablement which is disabled at the beginning?

RubenKelevra:
So `init()` in the last line runs `updateOkButtonEnablement()`, but at the time the button does not exist yet.

So in `updateOkButtonEnablement()` the `getButton()`'s return of NULL will be ignored by the `?.` in line 111 (patched).

So the button won't be deactivated if `selectedRadioButtonId == 0` as it's supposed to be.

That's why I think we need to call it again, after the button exists.

westnordost:
In line 67 the button is created.

RubenKelevra:
I'm sorry, but I don't think so. Line 67 just stores, that you want a button, but you can only access it after it is created by `AlertDialog.show()`....

> Gets one of the buttons used in the dialog. Returns null if the specified button does not exist or the dialog has not yet been fully created (for example, via [Dialog.show()](https://developer.android.com/reference/android/app/Dialog#show()) or [Dialog.create()](https://developer.android.com/reference/android/app/Dialog#create())).

https://developer.android.com/reference/android/app/AlertDialog#getButton(int)

westnordost:
then, why does it work currently?

RubenKelevra:
Looking at the code and the documentation, I have no idea. As I said, I'm not good at UI stuff in general. But I still wanted to make the suggestion, as it doesn't hurt anything to send the button twice a signal to be "disabled".

westnordost:
Well, I guess we can merge it. You did test it to ensure that there are no regressions?

I/we are in the progress of recreating the UI in compose, so I don't really want to spend time on hunting things that _could_  be bugs in the old UI implementation.

RubenKelevra:
> I/we are in the progress of recreating the UI in compose, so I don't really want to spend time on hunting things that _could_ be bugs in the old UI implementation.

Ah, alright, makes sense.  

Yeah I read your post about how to help and looked already at the source code of maplibre-compose.

> Well, I guess we can merge it. You did test it to ensure that there are no regressions?

I remember building it, but I think it did not test it. I'll do this now to be sure and report back :)

westnordost:
> Yeah I read your post about how to help and looked already at the source code of maplibre-compose.

Migrating to maplibre-compose is the next big thing on my list. The author of that library already added the one missing feature that blocked integration of that library into StreetComplete 🎉. Migrating to that lib is a big item on the list, so if you are looking to contribute some stuff, better start with something small :-) - e.g. some quest form.

RubenKelevra:
> > Yeah I read your post about how to help and looked already at the source code of maplibre-compose.
> 
> Migrating to maplibre-compose is the next big thing on my list. The author of that library already added the one missing feature that blocked integration of that library into StreetComplete 🎉. Migrating to that lib is a big item on the list, so if you are looking to contribute some stuff, better start with something small :-) - e.g. some quest form.

Alright, I'll have a look if I got time to spare :)

---

Regarding this issue. Tested on Android 12 on my phone: I've just cherry-picked this onto `v61.1 ` and everything works regarding this quest.

PR #6326, opened 2025-06-07
build: extract helper methods in DownloadBrandLogosTask to reduce complexity

RubenKelevra:
Split run() method into downloadLogo() and computeSmallImageUrl() to improve readability and maintainability.

Fixes two detekt complains:

```
buildSrc/src/main/java/DownloadBrandLogosTask.kt:20:21: The function run appears to be too complex based on Cyclomatic Complexity (complexity: 19). Defined complexity threshold for methods is set to '15' [CyclomaticComplexMethod]

buildSrc/src/main/java/DownloadBrandLogosTask.kt:20:21: Function run is nested too deeply. [NestedBlockDepth]
```

### DIFF
--- a/buildSrc/src/main/java/DownloadBrandLogosTask.kt
+++ b/buildSrc/src/main/java/DownloadBrandLogosTask.kt
@@ -26,51 +26,46 @@ open class DownloadBrandLogosTask : DefaultTask() {
         /* NSI uses (atm) a slightly different format than the normal presets: The presets are in
            a sub-object called "presets" */
         val presets = nsiPresetsJson.obj("presets")!!
-        for (entry in presets.entries) {
-            val id = entry.key
-            val presetJson = entry.value as JsonObject
-            val imageURL = presetJson["imageURL"] as String?
-            if (imageURL != null) {
-                val smallImageUrl = when {
-                    imageURL.contains("graph.facebook.com") && imageURL.endsWith("large") -> {
-                        // small is about 50x50px. large would be 200x200px
-                        imageURL.replaceRange(imageURL.length - 5, imageURL.length, "small")
-                    }
-                    imageURL.contains("commons.wikimedia.org") && imageURL.endsWith("width=100") -> {
-                        imageURL.replaceRange(imageURL.length - 3, imageURL.length, "50")
-                    }
-                    imageURL.contains("pbs.twimg.com/profile_images") -> {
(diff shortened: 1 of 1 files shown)


westnordost (commented):
Did you test it?

RubenKelevra:
Well, I didn't really expect that you're working on the weekend. So the two tasks are still running for validation. Got _bamboo piping_ - how we say in Germany - so this will take a while.

But I noticed that we skip all svgs... that would be easily fixable, as Wikipedia does render SVGs, we would just need to convert the URL from the original SVG file to the rendered PNG ones. Should I add this, or is there no plan to ever use this again?

matkoniecz:
> Well, I didn't really expect that you're working on the weekend. 

At least I am appearing on this issue tracker quite randomly, especially as basically all of my issue tracker activity here is an unpaid hobby :)

I would expect that PRs not yet finished and not yet ready for full review and merge would be created as drafts (I did it with say #6302)

RubenKelevra:
Sure, let's make it a draft. :)

RubenKelevra:
... and I think retries would also help. The current version is failing hard on a single timeout :/

``` 

> Task :app:downloadBrandLogos FAILED

[Incubating] Problems report is available at: file:///home/ruben/Projects/streetcomplete_StreetComplete/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:downloadBrandLogos'.
> java.net.ConnectException: Connection timed out

```

westnordost:
Maybe the URLs are not current anymore. Or maybe in the meantime there is some sort of (LLM-training-induced) quota on getting these logos.

In any case, this class is basically dead code. We don't actually download the brand logos for StreetComplete, because I noticed that they are just too big to include. The code just sits there because it also doesn't bothers anyone.

So... anyway, let's just not spend time on that.

RubenKelevra:
Sad, I liked the idea to get small icons, say if you're asked about the opening hours, next to the name of the shop.

I mean, instead of saving all of them inside the app, they could also be fetched on the fly and stored it in the cache of the app instead?

matkoniecz:
And if it is dead and use is not planned then maybe it can be deleted?

westnordost:
Well, I guess it would be possible to download them directly when being displayed and then cached for some time. That's how iD does it. It would be unusual though for StreetComplete, because otherwise all functionality is available offline.

PR #6332, opened 2025-06-07
avoid integer math in StreetParkingDrawable

RubenKelevra:
With the current images "ic_car*" this is not causing any harm, however if the image is turned 90° the integer math can lead to 0 as result for height/width. Leading to Infinity as results.

To avoid issues after exchanging/adding the images in the future, implicit floating point operations are required.


westnordost (approved):

westnordost:
thanks

PR #6337, opened 2025-06-08
Re-do survey checker

westnordost:
- move to common
- add test for RecentLocations class
- put survey-checking into singleton class (instead of having RecentLocations as the singleton)


Helium314:
Looks mostly good to me, but clearing all locations closer than `minDistanceMeters` to a newly added location seems to break the purpose of `RecentLocations`.
If the user is walking, any location within ~30 seconds will be within the 40 m `minDistanceMeters` of the previous location. So if there is more than one location update every 30 seconds, each update will very likely trigger removal of the previous location. `RecentLocations` would only work when moving very fast, having bad GPS, or having network-based locations.

westnordost:
Oh, right, of course! Dang it, thanks for the review! Will correct this...

Helium314:
I guess that's why I did the distance check when getting the location sequence. But of course that's much less efficient.

PR #6339, opened 2025-06-10
remove incorrect comment

matkoniecz:
accoding to my testing it is used in regular operations in overlay editing, at least in Things, Places and Buildings overlays

BTW, maybe I got blind but I cannot find ShowQuestFormsScreen anymore in debug builds

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/overlays/AbstractOverlayForm.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/overlays/AbstractOverlayForm.kt
@@ -109,7 +109,6 @@ abstract class AbstractOverlayForm :
             return localizedContext.resources
         }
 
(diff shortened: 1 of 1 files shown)


westnordost:
No, that comment is quite correct. Maybe it is misunderstandable. What I mean with this is that usually they'd use elementEditsController directly, it's just for ShowQuestFormsScreen that it doesn't.

Anyway, 

> BTW, maybe I got blind but I cannot find ShowQuestFormsScreen anymore in debug builds

I had to change something in the build config. To enable them again, add the line 
```
app.streetcomplete.debug=true
```

to your `~/.gradle/gradle.properties`

matkoniecz:
what about `// used to enable testing via ShowQuestFormsScreen! Found no better way to do this` ?

westnordost:
ok

matkoniecz:
done in https://github.com/streetcomplete/StreetComplete/commit/80f7b28cf623e668fa747cd41015c362bfd17ccb

Issue #6310, opened 2025-06-06
Multiplatform geolocation and permissions


westnordost:
We need to implement access to the device's geolocation and asking user permission to do that also on iOS.

This could be implemented separately on iOS, wrapped behind a common `expect` interface. But it looks like this has already been done, so it looks like we could use 

- compass-geolocation
- compass-permissions

from  https://github.com/jordond/compass to replace current Android platform specific code.

westnordost:
Unfortunately, Compass depends on Google Play Services on Android, see https://github.com/jordond/compass/issues/385 . This means, we can not use it. Since the license allows it, we could copy&paste the parts from Compass we need and roll our own Android implementation, leaning on what we already have. It's not that much, after all.

Alternatively, we can hope that some API will be available in maplibre-compose (or another package under the maplibre umbrella) which we could use to 1. take care of asking for location permissions and 2. provides location updates. The hope is not unfounded, as some work regarding this has been going on, but I am not sure whether the scope of this API will be sufficient for our requirements.
In any case, already now it includes some Android and iOS code that registers for location updates (without the use of Google Play Services), so even if it may not be eligible in the end, we could copy&paste some stuff from there.

Right now, I'd say, we wait.

westnordost:
So, to document what are our requirements:

- Manage asking for precise location permissions (without necessarily immediately starting to track location). Including asking again when it was denied once.
- Knowing the current location state, i.e. permission denied, no location available, location services turned off by user etc.
- register to get location updates and unregister from it, as a callback or as a flow

Issue #6072, opened 2025-01-01
Migrate to Maplibre-Compose


westnordost:
[MapLibre Compose](https://github.com/sargunv/maplibre-compose/) is a library to directly use MapLibre in Compose Multiplatform. It is currently not stable yet, but it is in very active in development, so it may be soon ready for use.

Currently, we use MapLibre for Android.

As far as I've followed development, the following things are still missing from the library that we require, which is why it is currently marked as blocked:
- downloading tiles for offline use

Contribution to the linked library is much appreciated. Ecosystem: MapLibre, Android, iOS, JavaScript, Compose Multiplatform, Kotlin.

westnordost:
I think I will expedite this. I broke the compose preview tooling with #6292, rather than looking into how to properly configure it now, might as well replace the dependencies to use compose multiplatform.

westnordost:
So, I did this in #6295 but I still need to move all those composables over to `src/common` which might reveal more (hidden) dependencies. Anyway, this can happen step by step as the quest forms are migrated, so I am keeping this in progress but unassign myself.

pwbriggs:
FYI, Maplibre-Compose looks stuck—I don't see any activity for the last few months.

sargunv:
> FYI, Maplibre-Compose looks stuck—I don't see any activity for the last few months.

There's been some small activity, mainly bug fixes and getting past an upgrade blocker on the iOS side. But most of my time right now is going towards https://github.com/maplibre/maplibre-native-ffi, which I intend to be a dependency of MapLibre Compose, initially for the desktop target, and eventually for Android and iOS too, to eliminate a lot of the complexity and hacks currently in place to expose a common api over three differing SDKs over MapLibre Native.

PR #6352, opened 2025-06-17
Migrate to MapLibre compose

westnordost:
Implements #6072.

This will happen in three phases. **We are in phase 1.** It looks as if I added a lot of stuff, but about 1700 lines added is just moving the map resources around. In reality it is about 3600 lines added, 521 lines removed. 

### 1st phase

- Re-create the map with all its StreetComplete layers with maplibre-compose. `MainMap` composable.
- For testing, we can add a new screen in (e.g. the settings screen) in which the new map can be viewed. `ShowMapScreen`.
- The map on the main screen remains based on maplibre-native Android in this phase.
- There's a `MainMapViewModel` from which `MainMap` sources data and data updates.
- Data not sourced from that view model (e.g. because it is dependent on UI state which better is not part of the view model) is passed directly to the map composable
- Testing whether the map looks and behaves fine can happen here
- Completing this will require maplibre-compose support all features we need it to have. Currently missing is https://github.com/maplibre/maplibre-compose/issues/468 . We can already build towards a TBD API, but we can only actually test it once it is available

### 2nd phase

- Actually replace the map in the main view with the compose-based one.
- This includes recreating all functionality of `MapFragment` and `MainMapFragment` in compose / viewmodels
- Also, all the wiring and communication between the `MainScreen` and the then-`MainMap` need to be rewired
- this will ripple into some more simplifications in communication, hopefully. Maybe some viewmodels can be merged into one
- also, some things now in viewmodels should rather be moved into composable code as UI state better resides in compose while viewmodels should rather only handle access to and updates from persistent data
- managing offline map downloads latest need to happen in phase 2
- completing this will require #6310 to be done, or at least, the location updates etc. be wrapped behind an interface that'd be not implemented on iOS yet

### 3rd phase

- Remove everything that was used for the maplibre-native Android map


westnordost:
Current blockers:
- Not possible / very awkward to refer to images from composeResources dynamically. ETA for a better interface: 1-2 months 
- not possible to refer to local (font) resources in the base style. ETA for an interface that makes it possible: 0-1 months.

riQQ:
Ref
- maplibre/maplibre-compose#468

mnalis:
> - Not possible / very awkward to refer to images from composeResources dynamically. ETA for a better interface: 1-2 months
> - not possible to refer to local (font) resources in the base style. ETA for an interface that makes it possible: 0-1 months.

Out of curiosity, what is the plan on this issue / when it might be expected approximately / [NET](https://en.wiktionary.org/wiki/NET#Preposition) [^1]  ?  
[Asking for a friend](https://github.com/Helium314/SCEE/pull/864#issuecomment-3548232447) :smiley_cat: 

[^1]: Hopefully my intention won't be misinterpreted; I'm just curious, definitely not trying to force people to work on it faster or anything, but any way I try to write this it seems to come out as rude, demanding and impatient :sweat_smile:

westnordost:
Hm well, the ETA for the first point is basically unknown now. As you see in the linked issue, there is some discussion which I haven't been following.

Regarding point two, not sure, but I think this is possible now.

Helium314:
Ok, so doesn't look like it's going to arrive anytime soon, which was a major reason for me pausing.

PR #6359, opened 2025-06-18
Expanded AddPitchLit to outdoor fitness station

paulklie:
Of the 120k [leisure](https://wiki.openstreetmap.org/wiki/Key:leisure)=[fitness_station](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfitness_station)  238 have lit=yes while 163 have lit=no.
This PR expands the quest for pitches to also ask if outdoor fitness stations are lit.
The [leisure](https://wiki.openstreetmap.org/wiki/Key:leisure)=[fitness_station](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfitness_station) tag is also often used for the individual machines, however since we only query for ways, these nodes will not be shown and only the underlying area.


westnordost:
Makes sense

PR #6361, opened 2025-06-19
Expand AddBusStopName to also ask for ferry terminal names

paulklie:
50.89% of Ferry terminals ([amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[ferry_terminal](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dferry_terminal)) already have names. In my experience they generally all have.


matkoniecz:
> In my experience they generally all have.

it is not a blocker, but in Poland inland ones on rivers tend to not have names, see https://overpass-turbo.eu/s/26kM (few have tagged names but often invalid ones)

matkoniecz:
BTW, is it not asked by AddPlaceName by any chance? See https://www.openstreetmap.org/node/3542167760/history (though there it was done by SCEE, SC)

yep, see https://github.com/streetcomplete/StreetComplete/blob/master/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt#L59

so is this PR a good idea? have you tested before/after?

westnordost:
See @matkoniecz , already asked for in different quest.

paulklie:
Oops sorry, good call, I had disabled the AddPlaceName quest for a different branch. Thanks!

PR #6365, opened 2025-06-21
expanded quests to Wilderness Hut (#6347)

### LINKED ISSUE #6347: Expand quests to tourism=wilderness_hut
There are currently around 22k wilderness huts mapped. I believe many quests can be extended to them, similar to alpine_hut:

**AddPlaceName:**
73.3% of wilderness huts already have one mapped. In my experience, all such huts have names.

**AddGeneralFee:**
1 544 have fee=no while 1 217 have fee=yes. I feel this will be highly dependent on region, in my region they are mostly free.

<s>**toilets_availability**
1202 have toilets=yes, 351 have toilet=no.</s>
Already implemented

**AddInternetAccess**
72 have internet_access=wlan, 200 have internet_access = no

**wheelchair_access**
137 have wheelchair=no, 13 have wheelchair=yes. 

I feel like some of these might be a bit "spammy", however, since wilderness huts are generally quite sparse, the user is unlikely to answer any of them more than 2-3 times a day.
Happy to develop these if approved.

paulklie:
resolves #6347

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -80,7 +80,7 @@ class AddPlaceName(
                 // name only
                 "wilderness_hut"
 
-                // and tourism = information, see abovegit
+                // and tourism = information, see above
             ),
             "leisure" to arrayOf(
                 // common
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -25,6 +25,7 @@ class AddPlaceName(
           or craft
           or amenity = recycling and recycling_type = centre
           or tourism = information and information = office
+          or tourism = wilderness_hut
           or """ +
 
         // The common list is shared by the opening hours quest and the wheelchair quest.
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -25,7 +25,6 @@ class AddPlaceName(
           or craft
           or amenity = recycling and recycling_type = centre
           or tourism = information and information = office
-          or tourism = wilderness_hut
           or """ +
 
         // The common list is shared by the opening hours quest and the wheelchair quest.
@@ -76,9 +75,12 @@ class AddPlaceName(
 
                 // name & wheelchair
                 "attraction",
-                "hotel", "guest_house", "motel", "hostel", "alpine_hut", "apartment", "resort", "camp_site", "caravan_site", "chalet" // accommodations
+                "hotel", "guest_house", "motel", "hostel", "alpine_hut", "apartment", "resort", "camp_site", "caravan_site", "chalet", // accommodations
 
-                // and tourism = information, see above
+                // name only
+                "wilderness_hut"
+
+                // and tourism = information, see abovegit
             ),
             "leisure" to arrayOf(
                 // common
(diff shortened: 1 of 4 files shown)


matkoniecz:
Have you tested this changes?

I have no real opinion on whether this changes make sense.

paulklie:
> Have you tested this changes?

Yep!

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
```diff
@@ -23,6 +23,7 @@ class AddWheelchairAccessBusiness : OsmFilterQuestType<WheelchairAccess>(), Andr
             or amenity = recycling and recycling_type = centre
             or amenity = social_facility and social_facility ~ food_bank|clothing_bank|soup_kitchen|dairy_kitchen
             or tourism = information and information = office
+            or tourism = wilderness_hut
```
matkoniecz: this one seems most dubious of proposed - I would expect wilderness hut to be in wilderness such as mountains and therefore not accessible with wheelchair anyway

but I have no great experience with those and maybe sizeable number is in places where you can get wheelchair to, but cannot get one inside (or only with major difficulty)

but I am not familiar with such places

matkoniecz: if it is added both to name & wheelchair - then it should be added in their shared section, see `// name & wheelchair` comment in line 65

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
```diff
@@ -25,6 +25,7 @@ class AddPlaceName(
           or craft
           or amenity = recycling and recycling_type = centre
           or tourism = information and information = office
+          or tourism = wilderness_hut
```
matkoniecz: if it is added both to name & wheelchair - then it should be added in their shared section, see `// name & wheelchair` comment in line 78

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
matkoniecz: if only name is asked then it also should be moved to this section

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
matkoniecz: if only name is asked then it also should be moved to this section

this query part is rather for ones where sadly more complex filtering is needed

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/wheelchair_access/AddWheelchairAccessBusiness.kt
paulklie: Hmm, I do agree. I have yet to vist a wheelchair accessible hut, however there are currently 13 marked with wheelchair=yes, and 137 with wheelchair = no. Though these small numbers are likely not representative.

However I think since this is a disabled by default quest, and these huts are quite sparse I feel like this would not be to spammy.

matkoniecz: I am not going to block altogether asking about wheelchair status, but I am also not going to merge with it.

I guess that either you can wait for someone else with merge ability to review one way or another or drop wheelchair part.

For what it is worth I asked on osm IRC and got similar opinion, that `wheelchair` is dubious here

> I feel like this would not be to spammy.

for me bigger problem is that it would be highly confusing

paulklie: Ok, I reverted the wheelchair change

matkoniecz:
I guess you can make a new PR that would add only wheelchair

(and yes, I should have rebased and squashed it, not only rebased, sorry - at least it ended with three commits, not three hundreds of them)

matkoniecz:
And thanks for contribution!

PR #6367, opened 2025-06-22
Slightly refactory MaxHeightPhysical Quest

### LINKED ISSUE #6362: Clarify wording for MaxPhysicalHeight Quest when asked for below a bridge
**Use case**
As per parent issue, the only thing giving users the hint that SC is asking for the physical height below a brdige is the little ruler to the right of the icon, the quest text reads exactly the same as when the maximum height was asked for reading a sign.

Together with the fact, that the physical quest is asked even when `maxheight:signed=no` (which is a valid answer for the MaxHeight quest), this leads to the confusing scenarion of users being asked the quest twice.

**Proposed Solution**
Update wording to "What is the physical height measured below this bridge?" or similar.

### LINKED ISSUE #6363: Revisit road types AR-measurements are asked for
**Use case**
The physical height quest (AR-measure) is currently asked for the `ALL_ROADS` constant, which includes motorways, primary and secondary roads as well as their links.
Even though users are asked to excercise caution when answering quests, insisting on asking AR-measuremeants on highspeed roads doesn't sound like a worthwhile task.
If users get annoyed by being asked for a measurement at a place where it is unsafe to do so, they might deactivate the quest, even though it would be fine to use in other "safe" situations.

**Proposed Solution**
Remove ways from AR-measurement quests, where it is unsafe and/or illegal to stop without sufficient reasons.

kmpoppe:
Resolves #6362, Resolves #6363

This PR is WIP for multiple reasons that I'd like your opinions on:

* What y'all think about HighwayUtils change? I believe that splitting `ALL_ROADS` into multiple vals could be useful later when adding and subtracting sets for various reasons. And it looks sleek ;-)
* Is the quest wording what you had envisioned ("...measured height...")? For me, that makes it crystal-clear what is asked for.
* I haven't added tests. Why: Because there's a "if `isApplicableTo` for nodes and ways is different then fail" condition, which works fine for `motorroad=*`, `motorway=*` and `expressway=*`, but testing with `highway=motorway` fails, even though I had thought it should work because nodes aren't filtered by `highway` anyway. I might be missing something, but I fail to see what :-/

Feedback appreciated!

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeight.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeight.kt
@@ -9,6 +9,7 @@ import de.westnordost.streetcomplete.data.osm.osmquests.OsmElementQuestType
 import de.westnordost.streetcomplete.data.quest.AndroidQuest
 import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement
 import de.westnordost.streetcomplete.osm.ALL_ROADS
+import de.westnordost.streetcomplete.osm.MOTORWAYS
 import de.westnordost.streetcomplete.osm.Tags
 import de.westnordost.streetcomplete.screens.measure.ArSupportChecker
 
@@ -37,7 +38,7 @@ class AddMaxPhysicalHeight(
 
     private val wayFilter by lazy { """
         ways with
-        highway ~ ${ALL_ROADS.joinToString("|")}
+        highway ~ ${(ALL_ROADS - MOTORWAYS).joinToString("|")}
         and (
           maxheight = below_default
           or source:maxheight ~ ".*estimat.*"
@@ -47,7 +48,13 @@ class AddMaxPhysicalHeight(
         and !maxheight:physical
         and access !~ private|no
         and vehicle !~ private|no
+        and motorroad !~ yes
+        and motorway !~ yes
+        and expressway !~ yes
     """.toElementFilterExpression() }
+    // explicitly removed motorway-style ways (highway=motorway*, motorroad/motorway=yes,
+    // expressway=yes) from ever being matched as measuring heights on these ways can be
+    // extremely difficult and/or dangerous and/or illegal
 
     override val changesetComment = "Specify maximum physical heights"
     override val wikiLink = "Key:maxheight"
@@ -64,8 +71,8 @@ class AddMaxPhysicalHeight(
             && tags["man_made"] != "pipeline"
         // only the "below the bridge" situation may need some context
         return when {
-            isBelowBridge -> R.string.quest_maxheight_below_bridge_title
-            else          -> R.string.quest_maxheight_title
+            isBelowBridge -> R.string.quest_maxheight_physical_below_bridge_title
+            else          -> R.string.quest_maxheight_physical_title
         }
     }
 
--- a/app/src/androidMain/res/values/strings.xml
+++ b/app/src/androidMain/res/values/strings.xml
@@ -1191,6 +1191,8 @@ If any lanes are reserved for buses, please leave a note instead."</string>
 
     <string name="quest_maxheight_title">"What’s the height limit here?"</string>
     <string name="quest_maxheight_below_bridge_title">What’s the height limit below the bridge?</string>
+    <string name="quest_maxheight_physical_title">"What’s the measured height here?"</string>
+    <string name="quest_maxheight_physical_below_bridge_title">What’s the measured height below the bridge?</string>
     <string name="quest_maxheight_sign_title">"What height limit is indicated here?"</string>
     <string name="quest_maxheight_answer_noSign">There is no sign…</string>
     <string name="quest_maxheight_split_way_hint">This question is also asked for roads that run under bridges, roofs, overhead pipes etc.\nAlso, if it doesn’t apply for the whole highlighted section, consider answering “%s”.</string>
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/HighwayUtils.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/HighwayUtils.kt
@@ -1,11 +1,17 @@
 package de.westnordost.streetcomplete.osm
 
-val ALL_ROADS = setOf(
-    "motorway", "motorway_link", "trunk", "trunk_link",
-    "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link",
-    "unclassified", "residential", "living_street", "pedestrian",
-    "service", "track", "busway", "road",
-)
+val MOTORWAYS = setOf("motorway", "motorway_link")
+val TRUNKS    = setOf("trunk", "trunk_link")
+val PRIMARY   = setOf("primary", "primary_link")
+val SECONDARY = setOf("secondary", "secondary_link")
+val TERTIARY  = setOf("tertiary", "tertiary_link")
+
+val CLASSIFIED_ROADS = PRIMARY + SECONDARY + TERTIARY
+val PEDESTRIAN_ROADS = setOf("residential", "living_street", "pedestrian")
+val OTHER_ROADS      = setOf("unclassified", "service", "track", "busway", "road")
+
+val ALL_ROADS = MOTORWAYS + TRUNKS + CLASSIFIED_ROADS +
+                PEDESTRIAN_ROADS + OTHER_ROADS
 
 val ALL_PATHS = setOf(
     "footway", "cycleway", "path", "bridleway", "steps"


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeight.kt
```diff
@@ -47,7 +48,13 @@ class AddMaxPhysicalHeight(
         and !maxheight:physical
         and access !~ private|no
         and vehicle !~ private|no
+        and motorroad !~ yes
+        and motorway !~ yes
+        and expressway !~ yes
     """.toElementFilterExpression() }
+    // explicitly removed motorway-style ways (highway=motorway*, motorroad/motorway=yes,
+    // expressway=yes) from ever being matched as measuring heights on these ways can be
```
matkoniecz:
```suggestion
    // expressway=yes) from ever being matched as measuring heights on these ways likely is
```

can be on any road

matkoniecz:
> I haven't added tests. Why: Because there's a "if isApplicableTo for nodes and ways is different then fail" condition, which works fine for motorroad=*, motorway=* and expressway=*, but testing with highway=motorway fails, even though I had thought it should work because nodes aren't filtered by highway anyway. I might be missing something, but I fail to see what :-/

can you share this broken test?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeight.kt
kmpoppe: The comment is broken in your suggestion lines, isn't it? I'd have commited it otherwise :)

matkoniecz: should be fixed now

kmpoppe:
> can you share this broken test?

These tests

```
    @Test fun `not applicable when on motorroad`() {
        assertFalse(isApplicableTo(mapOf("maxheight:signed" to "no", "motorroad" to "yes")))
    }
    @Test fun `not applicable when on motorway`() {
        assertFalse(isApplicableTo(mapOf("maxheight:signed" to "no", "motorway" to "yes")))
    }
    @Test fun `not applicable when on expressway`() {
        assertFalse(isApplicableTo(mapOf("maxheight:signed" to "no", "expressway" to "yes")))
    }
```

run successfully, ONLY when 

`private val nodeFilter by lazy` gets amended to _also_ include

```
        and motorroad !~ yes
        and motorway !~ yes
        and expressway !~ yes
```

just like the wayFilter.

When those are missing, the tests fail with `Result of isApplicableTo is not the same for nodes and ways` due to

https://github.com/streetcomplete/StreetComplete/blob/c83e0f963f1a6e160760db4ec3c4850b1eebc56b/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeightTest.kt#L60-L62

Regardsless,

```
    @Test fun `not applicable when on motorway_link`() {
        assertFalse(isApplicableTo(mapOf("maxheight:signed" to "no", "highway" to "motorway_link")))
    }
```

still fails, even though the nodeFilter is amended as stated above. It won't make sense excluding `highway=motorway*` to the node filter, because those most likely aren't on nodes, I presume?

app/src/androidMain/res/values/strings.xml
```diff
@@ -1191,6 +1191,8 @@ If any lanes are reserved for buses, please leave a note instead."</string>
 
     <string name="quest_maxheight_title">"What’s the height limit here?"</string>
     <string name="quest_maxheight_below_bridge_title">What’s the height limit below the bridge?</string>
+    <string name="quest_maxheight_physical_title">"What’s the measured height here?"</string>
+    <string name="quest_maxheight_physical_below_bridge_title">What’s the measured height below the bridge?</string>
```
mnalis: Should as well remove now unused old strings?
The _"measured height"_ works for me...

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/HighwayUtils.kt
```diff
@@ -1,11 +1,17 @@
...
-)
+val MOTORWAYS = setOf("motorway", "motorway_link")
+val TRUNKS    = setOf("trunk", "trunk_link")
+val PRIMARY   = setOf("primary", "primary_link")
+val SECONDARY = setOf("secondary", "secondary_link")
+val TERTIARY  = setOf("tertiary", "tertiary_link")
+
+val CLASSIFIED_ROADS = PRIMARY + SECONDARY + TERTIARY
+val PEDESTRIAN_ROADS = setOf("residential", "living_street", "pedestrian")
+val OTHER_ROADS      = setOf("unclassified", "service", "track", "busway", "road")
```
mnalis: I like the `(ALL_ROADS - MOTORWAYS)` construct, it is quite readable and performance isn't the issue on such small sets. I didn't even know Kotlin can do that (not that I have any reasonable knowledge of Kotlin :smile: )

But I dislike term `CLASSIFIED_ROADS`  - we have enough misunderstanding about `highway=unclassified` as it is. And road national road classification might include e.g. residential roads, it's just that don't have a `ref` but a `name` instead. And motorways have a ref, so are classified too? So I'd avoid term relating to _"classified"_ in any case...

I also dislike the term `PEDESTRIAN_ROADS`... For me it means `highway=pedestrian` (and _maybe_ `living_street` -- in any case something designed ***primarily*** for use by pedestrians), but certainly not `residential`. Why would `highway=residential` be "pedestrian" but for example `track` or `service` not (they usually get much less traffic then `residential`, so are more pedestrian-friendly if anything)?

So I'd change the names / content of `CLASSIFIED_ROADS` and `PEDESTRIAN_ROADS`  to something more reasonable and non-ambiguous:

- Perhaps ` val MAJOR_ROADS` for `PRIMARY + SECONDARY + TERTIARY` would be better than `CLASSIFIED_ROADS`? But then it should include `MOTORWAYS` and `TRUNKS` too

- Not sure about `PEDESTRIAN_ROADS`, maybe just move `residential` from it to `OTHER_ROADS`?  or create separate `MINOR_ROADS`?

mnalis:
As for the failing tests, it seems a little bit over my head, sorry

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/HighwayUtils.kt
kmpoppe: > I didn't even know Kotlin can do that

We use `ALL_ROADS + ALL_PATHS` a few times, so I figured, `-` would work too 🙂 

> But I dislike term `CLASSIFIED_ROADS`

What do you think about `IHCE_ROADS` ([wiki](https://wiki.openstreetmap.org/wiki/International_highway_classification_equivalence)), albeit not including `motorway` and `trunk` - which I would very much like to keep seperate from "non-motorways".

> or create separate `MINOR_ROADS`?

I like those ;-)

matkoniecz: `IHCE_ROADS` is just incomprehensible to me when looking at code

kmpoppe: Updated naming in 9436302cb567015b3666151b256604df4e34a67a

app/src/androidMain/res/values/strings.xml
westnordost: Yes, but in this case, @kmpoppe , I would prefer if you just changed the strings to include the new wording (i.e. previous translations will just be marked as "fuzzy" rather than deleted) instead of adding new strings.

westnordost (commented):
> What y'all think about HighwayUtils change? 

It's okay, it's good

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeight.kt
```diff
@@ -47,7 +48,13 @@ class AddMaxPhysicalHeight(
         and !maxheight:physical
         and access !~ private|no
         and vehicle !~ private|no
+        and motorroad !~ yes
+        and motorway !~ yes
+        and expressway !~ yes
```
westnordost:
```suggestion
        and motorroad != yes
        and motorway != yes
        and expressway != yes
```

kmpoppe:
@westnordost any ideas on how to solve the test node/way equality stuff?

app/src/androidMain/res/values/strings.xml
kmpoppe: Resolved in 526530ee748f664ec7a8d493de21742b10a500cf

westnordost:
I don't know, what do you want to test? If for nodes and ways the filter should be different, then you can split it up.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/HighwayUtils.kt
westnordost, line 1: Probably should rename the file to `RoadUtils` now (since you use the word "highway" for meaning motorway/trunks.

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeightTest.kt
```diff
@@ -57,9 +72,9 @@ class AddMaxPhysicalHeightTest {
...
             tags = mapOf("highway" to "service") + tags
         ))
-        if (nodes != ways) {
-            fail("Result of isApplicableTo is not the same for nodes and ways")
-        }
-        return nodes
+        //if (nodes != ways) {
+        //    fail("Result of isApplicableTo is not the same for nodes and ways")
+        //}
+        return nodes && ways
```
westnordost: Uhh, what?

---

Okay. Let's clear the confusion.

Do this (or similar):

```kt
    private fun isApplicableToHeightRestrictor(tags: Map<String, String>): Boolean =
        nodes = questType.isApplicableTo(node(
            tags = mapOf("barrier" to "height_restrictor") + tags
        ))

    private fun isApplicableToRoad(tags: Map<String, String>): Boolean =
        nodes = questType.isApplicableTo(way(
            tags = mapOf("highway" to "service") + tags
        ))
```

Then, for every test that should test both for nodes and for ways, use both functions in order, e.g.

```kt
    @Test fun `applicable if maxheight is only estimated`() {
        val tags = mapOf("maxheight" to "3","source:maxheight" to "estimation")
        assertTrue(isApplicableToHeightRestrictor(tags))
        assertTrue(isApplicableToRoad(tags))
    }
```

westnordost (changes requested):

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/HighwayUtils.kt
kmpoppe: done with c87b0aca42ac1118c94d0da423a9bc6da7effa91

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/max_height/AddMaxPhysicalHeightTest.kt
kmpoppe: less-foggy-brain-me-commit: 2178940b1a8988724ebf51ccb0010335b2fd9075

westnordost (approved):

PR #6368, opened 2025-06-22
Added quests to leisure=sauna

paulklie:
Currently there are around 14k non private saunas.
This PR expands the AddOpeningHours, AddWheelchairAcessBuisness and AddPlaceName Quests to them.

182 of saunas have wheelchair = no while 92 have yes.
56% have names.

I have tested and verified its functionality.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
@@ -73,7 +73,7 @@ class AddOpeningHours(
             "leisure" to arrayOf(
                 // common
                 "fitness_centre", "golf_course", "water_park", "miniature_golf", "bowling_alley",
-                "amusement_arcade", "adult_gaming_centre", "tanning_salon",
+                "amusement_arcade", "adult_gaming_centre", "tanning_salon", "sauna",
 
                 // not sports_centre, dance etc because these are often sports clubs which have no
                 // walk-in opening hours but training times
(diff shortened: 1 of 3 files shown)


burrscurr:
Thank you for your suggestion.

I'm no maintainer, but I think at least one aspect of the proposed changes needs more discussion: Wheelchair access. [Generally, quests should be easy to answer and the relevant places should be accessible by foot](https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md#users). I don't think this is (sufficiently) the case here. In many cases, getting to the sauna area can be difficult and may require buying tickets. This is different from most shops, where the most common wheelchair issue is getting *into* the shop at the first place. In a sauna, there may be lots of potential practical issues that could limit wheelchair friendliness (physical access to the sauna area, turnstiles etc., design of changing areas, design of showers, the saunas themselves, ...). I'm somewhat sceptical this can be surveyed competently by an average StreetComplete user.

paulklie:
> Thank you for your suggestion.
> 
> I'm no maintainer, but I think at least one aspect of the proposed changes needs more discussion: Wheelchair access. [Generally, quests should be easy to answer and the relevant places should be accessible by foot](https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md#users). I don't think this is (sufficiently) the case here. In many cases, getting to the sauna area can be difficult and may require buying tickets. This is different from most shops, where the most common wheelchair issue is getting _into_ the shop at the first place. In a sauna, there may be lots of potential practical issues that could limit wheelchair friendliness (physical access to the sauna area, turnstiles etc., design of changing areas, design of showers, the saunas themselves, ...). I'm somewhat sceptical this can be surveyed competently by an average StreetComplete user.

Thank you for your comment.
From my understanding this is to be expected from the wheelchair quest, and this is why it is disabled by default. Already implemented places like cinemas are also very hard to verify without a ticket.

westnordost:
@burrscurr The quest is already disabled by default for this reason. E.g. it is also asked for zoos, aquariums, etc.

I guess, it can be answered by asking at the ticket counter at the entrance.

PR #6375, opened 2025-06-26
#6373 Ask for ferry relationships and ferry ways not part of a ferry relationship

### LINKED ISSUE #6373: Ferry access quests misshandle ways in relations
Ferry routes can be ways or relations.
Ways are generally used to mark single point to point ferrys that only stop in two places and where each side is used only by one line.
In other cases routes are a relation made up out of multiple ways.
For the ferry access quests we ask for access for both cases: 
https://github.com/streetcomplete/StreetComplete/blob/6eaf52d3eb7027b8d292f907b2b0b02366120523/app/src/main/java/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessPedestrian.kt#L14
However this can lead to issues when a ferry route is made up of multiple ways:

1. Since we ask for both the relation and the way we massively increase the workload, especially for very complex areas. This is very spammy for the user since they are being asked if the ferry suddenly stopped accepting pedestrians / motor_viecles midway in the journey.
2. Asking for ways that are shared by multiple ferry routes can lead to bad data: The user might take route 1 on a way that is shared with route 2, if their access differs the user is probably unaware of this and will answer only for ferry 1.

I also find that the query for relations tends to not work at times? Does the entire relation need to be loaded? It seems very unreliable on my end.

I instead of filtering for any way or relation with route=ferry, we filter for:
relation with route=ferry and not route_master
way with route=ferry and not part of a relation with route=ferry.

paulklie:
resolve #6373
I don't have much experience in kotlin, or this repo so am unsure if I conforms to coding guidelines, so some feedback would be good. But it seems to work with my testing. 
Since we might be adding many ferry quests I think it might make sence to move some duplicate code to a single place, but not sure where that should be.


app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiParser.kt
```diff
@@ -64,7 +64,7 @@ private fun XmlReader.parseMapData(ignoreRelationTypes: Set<String?>): MutableMa
         END_ELEMENT -> when (localName) {
             "node" -> result.add(Node(id!!, position!!, tags.orEmpty(), version!!, timestamp!!))
             "way" -> result.add(Way(id!!, nodes, tags.orEmpty(), version!!, timestamp!!))
-            "relation" -> if (tags.orEmpty()["type"] !in ignoreRelationTypes) {
+            "relation" -> if (tags.orEmpty()["type"] !in ignoreRelationTypes || tags.orEmpty()["route"] == "ferry") {
```
matkoniecz: probably `if (tags.orEmpty()["type"] !in ignoreRelationTypes || tags.orEmpty()["route"] == "ferry")` should be rebuilt into `ignorableRelation(tags)` or similar - that will check both `type` and `route` tags

this (or something similar) should be done to avoid making `ignoreRelationTypes` misleading
as now `ignoreRelationTypes` includes `route` which is no longer entirely ignored

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt
```diff
@@ -26,4 +31,44 @@ class AddFerryAccessMotorVehicle : OsmFilterQuestType<Boolean>(), AndroidQuest {
...
+            return null
+        }
+
+        return true
+    }
+
+    private fun isApplicableTo(element: Element, mapData: MapDataWithGeometry): Boolean {
+        val tags = element.tags
+
+        // Filter out ferries that are part of a ferry route relation
```
matkoniecz: maybe it is possible to extract this to shared `de.westnordost.streetcomplete.quests.ferry` file? 

Like was done in say https://github.com/streetcomplete/StreetComplete/blob/master/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/construction/MarkCompletedConstructionUtils.kt ?

(note: I have lower confidence here, maybe wait if someone else will protest)

matkoniecz:
Big thanks for implementing it!

> But it seems to work with my testing.

it may be worth adding also an automated tests

see say https://github.com/streetcomplete/StreetComplete/blob/2c4303539600c176c8adcb7d1b831edad128496d/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/building_entrance/AddEntranceTest.kt or https://github.com/streetcomplete/StreetComplete/blob/2c4303539600c176c8adcb7d1b831edad128496d/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/address/AddHousenumberTest.kt as a base

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiParser.kt
paulklie: I have no idea how this could be achieved other than by adding another includeRelationTypes with only a single element, which seems quite unnecessary

paulklie:
> Big thanks for implementing it!
> 
> > But it seems to work with my testing.
> 
> it may be worth adding also an automated tests
> 
> see say https://github.com/streetcomplete/StreetComplete/blob/2c4303539600c176c8adcb7d1b831edad128496d/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/building_entrance/AddEntranceTest.kt or https://github.com/streetcomplete/StreetComplete/blob/2c4303539600c176c8adcb7d1b831edad128496d/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/address/AddHousenumberTest.kt as a base

Added tests.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiParser.kt
matkoniecz: I thought about replacing current `ignoreRelationTypes()` that takes no arguments by `ignorableRelation(tags)` that would take tag list as an argument, and remove current `ignoreRelationTypes` function

(it would replace it - I would not add an extra new function)

paulklie: I am sorry I dont understand what you mean, I am quite new to kotlin so I dont quite understand: how can ignoreRelationTypes() (which is a variable right?) take arguments?

matkoniecz: I thought about creating a function that would replace `ignoreRelationTypes` set (which is not a function, you are right, sorry for a confusion)

paulklie: Ah no worries, can you think of any other way this could be improved?

matkoniecz: I will try to take look at this, but may take few days as I will be away on multiple family events over next days

feel free to ping me on Tuesday if I will do nothing

(I would try introducing `ignorableRelation(tags)` function)

paulklie:
For testing, I can recommend this place:

https://www.openstreetmap.org/?mlat=47.72750&mlon=13.43909#map=16/47.72750/13.43909

It has all combinations: a simple ferry route that is just a way, and a way that is part of a relation right next to eachother.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt
```diff
@@ -26,4 +31,24 @@ class AddFerryAccessMotorVehicle : OsmFilterQuestType<Boolean>(), AndroidQuest {
...
+            isApplicableTo(element, mapData)
+        }
+    }
+
+    override fun isApplicableTo(element: Element): Boolean? {
+        // Check if it's a relation. If not its a way and is required not to be part of a relation.
+        val tags = element.tags
+        if (element.type == ElementType.RELATION) {
+            return true
+        }
```
westnordost: So, any relation is applicable? Cool! ;-) 

(I think you are missing a TODO here)

westnordost (changes requested):
@matkoniecz is correct, this is the wrong point at which to make the ignored relation types more dynamic/flexible because which relations are ignored is already relevant for the **`Uploader`**.

The correct point would be to replace `val IGNORED_RELATION_TYPES: Set<String>` in `ApplicationConstants.kt` with a function like 
```
fun isRelationIgnored(tags: Map<String, String>): Boolean
```

This function is then passed as a parameter to all places where currently `IGNORED_RELATION_TYPES` is passed to and those places need to be refactored to accept a function. (For example, maybe there are other solutions, too)

This part, by the way, should ideally be a separate PR, enabling this PR.

---

Regarding  `FerryUtils.kt`, it looks fine. More consistent with the rest of the code and probably more performant however would be a more functional style, e.g. something like
```kt
val ferryRouteWays: Collection<Way> = ...
val ferryRouteRelations: Collection<Relation> = ...
val waysInFerryRouteRelations: Set<Way> = ...
return ferryRouteRelations + (ferryRouteWays - waysInFerryRouteRelations)
```

This is pretty compact and straightforward, little duplicated code, so maybe `FerryUtils` is not even necessary anymore.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt
paulklie: Oops, good catch. I fixed it and added a testcase for this.

paulklie:
> Regarding `FerryUtils.kt`, it looks fine. More consistent with the rest of the code and probably more performant however would be a more functional style, e.g. something like
> (More performant because currently you have to iterate through all relations for every single ferry route way)

I am afraid this exceeds my limited kotlin knowledge, and I dont know how todo this.
> 
> This is pretty compact and straightforward, little duplicated code, so maybe `FerryUtils` is not even necessary anymore.

Hmm I think we can also reuse it for other ferry quests (MaxWeight, Fee) as well as further ones that might be useful (Opening Hours, Maxweight?) to.

westnordost:
Well, for the record, I don't like it, but it seems to be common tagging practice, so it unfortunately makes sense for StreetComplete to take this into account.

In my last commit, I implemented what I noted in my last comment. I kept your utils file and changed the tests to test exactly only that helper function.

paulklie:
> In my last commit, I implemented what I noted in my last comment. I kept your utils file and changed the tests to test exactly only that helper function.

Ok great thanks!

PR #6379, opened 2025-06-30
Update RoadUtils and val usages

kmpoppe:
This draft-PR removes multiple manual lists for `highway` and replaces them with vals defined in `RoadUtils` in lazyFilters.

Reasoning is obvious, should one ever want to change multiple lists this makes it easier to do so, even more so as when val names are descriptive.

Quest-specific lists have been transferred to `RoadUtils` as well.

Feedback is sought for specifics (combinations of sets) or general use of this approach.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/overlays/cycleway/CyclewayOverlay.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/overlays/cycleway/CyclewayOverlay.kt
@@ -12,7 +12,11 @@ import de.westnordost.streetcomplete.data.overlays.OverlayColor
 import de.westnordost.streetcomplete.data.overlays.Overlay
 import de.westnordost.streetcomplete.data.overlays.OverlayStyle
 import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement
+import de.westnordost.streetcomplete.osm.ALL_LINKS
 import de.westnordost.streetcomplete.osm.ALL_ROADS
+import de.westnordost.streetcomplete.osm.HIGHWAYS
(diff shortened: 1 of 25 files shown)


westnordost:
> Feedback is sought for specifics (combinations of sets) or general use of this approach.

In general, this might be an improvement. But please don't over-engineer it!

Directive  number one for a refactor is code readability, i.e. it should be easy to understand the code, without browsing through different files to find the meaning of something.

If the resulting code is longer and less expressive, that's a bad sign. E.g.

```diff
- highway ~ cycleway|path|footway
+ highway ~ ${PATH_FOR_FOOT_AND_CYCLE.joinToString("|")}
```

I find questionable. It is both longer and more complicated to read (with all those letters shouting at you plus this ${}"|"()-stuff) and actually less expressive as it is not entirely clear if steps (with ramps) and (forest) tracks would be included. Also, and that's also important, it is less flexible. These kind of expressions are often slightly refined in PRs by contributors to include this or to include that only under certain circumstances (e.g. steps only if there is also a ramp), and by referring to some app-wide constant, we are taking away this flexibility and open the door for hacky workarounds, like... (for example)

```
highway ~ ${(PATH_FOR_FOOT_AND_CYCLE - "cycleway").joinToString("|")}
or highway = cycleway and foot = designated
```

and stuff like that.

Scrolling through it, furthermore `TAGGING_NOT_ASSUMED` would be another example where the readability is actually downgraded, because its meaning is not clear from reading it. While `MAJOR_ROADS` for example is mostly clear (does it include trunks or not, though?), `TAGGING_NOT_ASSUMED` is super unclear: tagging of *what* not assumed *where*? 


So, having scrolled through it, I am hard pressed to see any example that is a clear upgrade. I think the ones that are clearly useful, we already have in place (`ALL_ROADS.joinToString("|")`).

PR #6383, opened 2025-07-02
added cave to AddWheelchairAccessBuisness and AddPlaceName

### LINKED ISSUE #6381: Extend quests to natural=cave_entrance
I suggest asking the following quests to [natural](https://wiki.openstreetmap.org/wiki/Key:natural)=[cave_entrance](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dcave_entrance):

**AddGeneralFee**
I think we can limit this to cave entrances with access=yes. Of these 2 597, 129 have fee=yes and 1605 have fee=no. 

**AddOpeningHours**:
I would limit this to caves with fee=yes. Currently has 650 uses.

**AddWheelchairAccessBusiness**: I would similarly limit this to caves with fee=yes, since in my experience those without a fee are almost always not accessible. Data is a bit more limited here with 86 such caves marked _yes_ and 7 _no_.

I was also thinking about suggesting adding the name quest, however since almost 72% of caves already have a name I believe that many of the remaining ones are not named.

paulklie:
closes #6381

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/place_name/AddPlaceName.kt
@@ -25,6 +25,7 @@ class AddPlaceName(
           or craft
           or amenity = recycling and recycling_type = centre
           or tourism = information and information = office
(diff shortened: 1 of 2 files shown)


westnordost:
LGTM

PR #6386, opened 2025-07-04
added pastry shop to AddVegan quest

paulklie:
Many pastrys are not vegan due to containing eggs.

However all pastry shops I have been to had vegetarian options, so the vegetarian quest makes less sense to me.


westnordost:
Yeah, makes sense

PR #6387, opened 2025-07-05
implemented isRelationIgnored, which does not ignore ferry routes

paulklie:
As suggested in #6375 this function replaces the simple relationship filter and allows ferry routes to be processed while all other routes are still ignored.

### DIFF
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiClient.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiClient.kt
@@ -33,7 +33,7 @@ class MapDataApiClient(
      *
      * @param changesetId id of the changeset to upload changes into
      * @param changes changes to upload.
-     * @param ignoreRelationTypes omit any updates to relations of the given types from the result.
+     * @param ignoreRelation omit any updates to relations of the given types from the result.
      *                            Such relations can still be referred to as relation members,
      *                            though, the relations themselves are just not included
      *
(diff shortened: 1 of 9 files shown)


westnordost:
By the way, thanks for all your recent contributions so far! It's really quite a bunch!

paulklie:
> By the way, thanks for all your recent contributions so far! It's really quite a bunch!

Thanks! Really love this tool, so thought it was time to give back!

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiClient.kt
```diff
@@ -88,7 +88,7 @@ class MapDataApiClient(
      *
      * @param bounds rectangle in which to query map data. May not cross the 180th meridian. This is
      * usually limited at 0.25 square degrees. Check the server capabilities.
-     * @param ignoreRelationTypes omit any relations of the given types from the result.
+     * @param ignoreRelation omit any relations of the given types from the result.
      *                            Such relations can still be referred to as relation members,
      *                            though, the relations themselves are just not included
```
westnordost:
```suggestion
     * @param ignoreRelation omit any relations for which the given function returns true.
     *                       Such relations can still be referred to as relation members,
     *                       though, the relations themselves are just not included
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataApiClient.kt
```diff
@@ -33,7 +33,7 @@ class MapDataApiClient(
      *
      * @param changesetId id of the changeset to upload changes into
      * @param changes changes to upload.
-     * @param ignoreRelationTypes omit any updates to relations of the given types from the result.
+     * @param ignoreRelation omit any updates to relations of the given types from the result.
      *                            Such relations can still be referred to as relation members,
      *                            though, the relations themselves are just not included
```
westnordost:
```suggestion
```suggestion
     * @param ignoreRelation omit any relations for which the given function returns true.
     *                       Such relations can still be referred to as relation members,
     *                       though, the relations themselves are just not included
```
```

westnordost (approved):

PR #6388, opened 2025-07-05
fixed charging station filter by adding 'motor_vehicle' and 'designated'

### LINKED ISSUE #6385: charging capacity quest mishandles non car chargers.
The wiki specifies that bicycle= designated or bicycle= yes can be used for bicycle charging stations: https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Vehicles

However the charging station capacity filter only checks for bicycle and scooter != yes and does not consider access tags with designated.
The filter also does not consider motor_vehicle = no:
https://github.com/streetcomplete/StreetComplete/blob/6eaf52d3eb7027b8d292f907b2b0b02366120523/app/src/main/java/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt#L19

For example this user is incorrectly asked about the "car capacity" of this node:
https://www.openstreetmap.org/node/11067171659#map=17/46.508343/11.758386

I see three possible solutions to this issue:

1. Correct the filter to: `bicycle != yes|designated and scooter != yes|designated and motorcar != no and motor_vehicle != no`.
2. Modify the question to "How many vehicle can be charged here at the same time? And remove any vehicle specific filters. The issue here is that this could lead to issues at charging stations where different vehicles can be charged (eg. combined car / bike stations with different capacities for each).
3. We could add specific icon/string based on which tags are used:

- `bicycle = yes|designated and and (motorcar = no or motor_vehicle = no)`: How many bicycles can be charged here?
- `hgv = yes|designated and and (motorcar = no)`: How many lorries can be charged here?
- `bicycle != yes|designated and scooter != yes|designated and motorcar != no and motor_vehicle != no`: How many cars can be charged here?

paulklie:
resolves #6385

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
@@ -17,7 +17,10 @@ class AddChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
         nodes, ways with
           amenity = charging_station
           and !capacity
-          and bicycle != yes and scooter != yes and motorcar != no
+          and bicycle != yes|designated
+          and scooter != yes|designated
+          and motorcar != no
+          and motor_vehicle != no
           and access !~ private|no
     """
     override val changesetComment = "Specify charging stations capacities"
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
@@ -17,8 +17,8 @@ class AddChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
         nodes, ways with
           amenity = charging_station
           and !capacity
-          and bicycle != yes|designated
-          and scooter != yes|designated
+          and bicycle !~ yes|designated
+          and scooter !~ yes|designated
           and motorcar != no
           and motor_vehicle != no
           and access !~ private|no


paulklie:
I was also thinking about renaming the quest to "Add**Car**ChargingStationCapacity" to specifiy that it is only for cars. What do you think?

mnalis:
> I was also thinking about renaming the quest to "AddCarChargingStationCapacity" to specifiy that it is only for cars. What do you think?

As that is what quest text says, it makes sense to me:

https://github.com/streetcomplete/StreetComplete/blob/c19a76a43d46479b91eb39ff3dcb9cb8fca61a8a/app/src/androidMain/res/values/strings.xml#L1025


But that reminds me of another problem; what about e-trucks? Those are also covered by `motor_vehicle`... I'm not familiar if they use different charging connectors than e-cars or same? But probably either:

- the filter should be changed to exclude trucks too (if connectors are not compatible), **_or_** 
- the text should be change to include trucks (if connectors are compatible)

matkoniecz:
Have you tested it? Based on source code it looks fine

westnordost:
Don't rename it, please, quests are saved in the database by name, each rename would require to write an upgrade-step for the database. Not worth it.q

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
```diff
@@ -17,7 +17,10 @@ class AddChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
         nodes, ways with
           amenity = charging_station
           and !capacity
-          and bicycle != yes and scooter != yes and motorcar != no
+          and bicycle != yes|designated
+          and scooter != yes|designated
```
westnordost:
```suggestion
          and bicycle !~ yes|designated
          and scooter !~ yes|designated
```

westnordost (changes requested):
> Have you tested it? Based on source code it looks fine

He didn't, because I can see that based on the source code, it doesn't look fine :-P

westnordost (approved):

PR #6394, opened 2025-07-07
added toilets:handwashing quest

### LINKED ISSUE #6389: Quest: Can you wash your hands here?
### General
Affected tag(s) to be modified/added: [toilets:handwashing](https://wiki.openstreetmap.org/wiki/Key:toilets:handwashing)
Question asked: **Can you wash your hands here?**

### Checklist
- [x] 🚧 **To be added tag is established and has a useful purpose**
Around 40k uses
It might need to be disabled by default, similar to the changing table quest.
38.62% have `yes`, while 61.32% have `no`
430k toilet nodes currently do not have this tag.

### Ideas for implementation

**Element selection:** 
`amenity=toilets and toilets:disposal and toilets:disposal != flush`
The vast majority of flush toilets have handwashing, so it seems like this would avoid alot of spam.

**Proposed UI:**
A simple yes/no quest, similarly to the changing table quest.

paulklie:
fixes #6389 .
I have tested just needs an icon

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
@@ -490,7 +490,7 @@ fun questTypeRegistry(
     // shop and others, but have to go inside
     120 to AddBabyChangingTable(), // used by OsmAnd in the object description, have to go inside
     121 to AddWheelchairAccessToiletsPart(), // have to go inside
-    168 to AddHandwashing(), // have to go inside
+    176 to AddHandwashing(), // have to go inside
 
     133 to AddFuelSelfService(),
     156 to CheckShopExistence(getFeature), // after opening hours and similar so they will be preferred if enabled


mnalis:
(Note: I've updated the PR text to use GitHub [magic word](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue), so the issue will be closed _automatically_ by GitHub if/when PR is merged)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -489,6 +490,7 @@ fun questTypeRegistry(
     // shop and others, but have to go inside
     120 to AddBabyChangingTable(), // used by OsmAnd in the object description, have to go inside
     121 to AddWheelchairAccessToiletsPart(), // have to go inside
+    139 to AddHandwashing(), // have to go inside
```
westnordost: quest ordinals may not be reassigned. New quests must always use the highest ordinal + 1. Should the comment at the top be more clear about this?

paulklie: I am sorry but I dont understand, what do you mean with reassigning? 139 was not used yet and was the smallest unused right?

HolgerJeromin: I think SC had 139 in the past.
But the mass of quests and the sorted order makes this very strange indeed.
What if the largest number is removed? Then noone will know that is exists in the past.
`168` seems the largest unused number...

matkoniecz: https://github.com/streetcomplete/StreetComplete/commit/2819a8d440fdf33fc1bff178679ae1fcfd739998 - 139 was used in past

westnordost: >  But the mass of quests and the sorted order makes this very strange indeed.

Yeah, I am sorry about that. Every quest needs an ordinal assigned, for the QR code / sharing presets feature. I didn't have an idea how to make this less awkward.

HolgerJeromin: Keeping the deleted numbers in a comment line would be a bit better

paulklie:
For the icon I was thinking a simple sink like https://www.svgrepo.com/svg/535644/sink.

paulklie:
handwashing
what do you think about this one?

westnordost:
Difficult to recognize as a sink, and bland (only white). I'd think that a (skin-colored) hand with a drop of water from the top would work.
But don't worry about an icon too much, historically I've always supplied an icon if a quest author didn't come up with one himself (that I liked).

westnordost (approved):

westnordost:
In retrospect, your icon was fine, I didn't manage to create something really better. But anyway, mine is probably fine too 🤷

paulklie:
ooh, yours looks nice, great job!

Issue #5870, opened 2024-09-02
AddBicyclePump, for bicycle_repair_station. Adding only pump


ArathonV:
**Use case**
A standalone bike pump is usually tagged as 
`service:bicycle:pump=yes
service:bicycle:tools=no `
the latter being assumed to be yes by default. However this currently seems impossible to tag in the app, as the pump can be added, and using the AddBicyclePump the pump tag can be set, the tools tag can not be set to no, thus incorrectly implying that this pump also provides tools.

**Proposed Solution**

- Add a Bike pump as an addable "thing"
- Add an "only" option to AddBicyclePump when asking about pumps that sets service:bicycle:tools=no

matkoniecz:
> the latter being assumed to be yes by default.

why you think so?

> using the AddBicyclePump the pump tag can be set, the tools tag can not be set to no, thus incorrectly implying that this pump also provides tools.

well, in such case the same was also implied earlier

> Add a Bike pump as an addable "thing"

Iff it makes sense to have it also such way in iD then it can be added https://github.com/openstreetmap/id-tagging-schema

Or is it maybe already possible in iD?

(StreetComplete is using iD presets)

ArathonV:
> why you think so?

the wiki says so, see [service:bicycle:tools](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:tools)

matkoniecz:
Hmm, in the worst case SC does not make things more clear - if 

- `amenity=bicycle_repair_station` was tagged without `service:bicycle:tools=no`
- `amenity=bicycle_repair_station` without `service:bicycle:tools=no` means that it has tools
- it has no tools

Then SC mapper adding `service:bicycle:pump=yes` still does not make any things worse.

matkoniecz:
If new quest asking about `service:bicycle:tools=` makes sense (not sure!) then it seems that it was not proposed yet: https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete+service%3Abicycle%3Atools&type=issues

(see https://github.com/streetcomplete/StreetComplete/issues/new?assignees=&labels=&projects=&template=quest-suggestion.md - note that I have not checked is it fulfilling this requirements, I suspect that it fails "no spam" as vast majority of this stands has tools)

ArathonV:
> Then SC mapper adding `service:bicycle:pump=yes` still does not make any things worse.

[...]

PR #6397, opened 2025-07-08
check for existence of bicycle repair stations

paulklie:
Feels like a small enough map feature to fit into the existence quest. 4 years seemed like the most fitting category. bike repair stations that have not been updated for 4 years are currently quite rare though, mostly due to the new bike repair quest. However I found this node that can serve for testing: https://www.openstreetmap.org/node/6145704153#map=19/54.892393/9.550448&layers=C

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/existence/CheckExistence.kt
@@ -44,6 +44,7 @@ class CheckExistence(
             or amenity = grit_bin and seasonal = no
             or amenity = vending_machine and vending ~ parking_tickets|public_transport_tickets
             or amenity = ticket_validator
+            or amenity = bicycle_repair_station
             or tourism = information and information ~ board|terminal|map
             or advertising ~ column|board|poster_box
             or (highway = emergency_access_point or emergency = access_point) and ref


mnalis:
Or perhaps `AddBicycleRepairStationServices` quest _itself_ could be reasked every 4 years? 

That would seem more useful for me, as repair stand might still be there, but pump stolen/broken for example (which is unfortunately quite common here)...

paulklie:
I think such a reasking quest would be significantly more complicated:

- What do we do with tools that were previously tagged yes and are damaged / gone? Simply set to `no`? Add `destroyed:` or  `was:`?
- Do we ask this quest if not all tools are set? eg. what if only [service:bicycle:pump](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:pump) but not [service:bicycle:tools](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:tools) are set?

matkoniecz:
> What do we do with tools that were previously tagged yes and are damaged / gone? Simply set to no? Add destroyed: or was:?

`no`, the same as with newly asked

> Do we ask this quest if not all tools are set? eg. what if only [service:bicycle:pump](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:pump) but not [service:bicycle:tools](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:tools) are set?

I would ask as long as unsupported values are not present (maybe allow them if set to `no`?)

paulklie:
> I would ask as long as unsupported values are not present (maybe allow them if set to `no`?)

Hmm, I think we need a new ui design then, that shows the user not just yes/no of each tool as currently, but can also display a value as "undefined" so the user can understand where additional information is needed.

matkoniecz:
undefined is the same as set to `no` - right?

paulklie:
> undefined is the same as set to `no` - right?

Not for [service:bicycle:tools](https://wiki.openstreetmap.org/wiki/Key:service:bicycle:tools) which is implied to be no. Though I think this fact makes such a re asking quest even more appealing, since many bike repair stations that have only pumps (very common in my area) are not tagged with tools=no. See my original issue on the topic: https://github.com/streetcomplete/StreetComplete/issues/5870#issuecomment-2352615247

westnordost:
I agree with the others, it would make sense to re-ask the bicycle tools quest, but this does not exclude that repair stations' existence could not also be be asked.

PR #6401, opened 2025-07-16
Add squeezer stiles to AddBarrierOpening

arrival-spring:
It seems to make sense to ask about the width of squeezer stiles

I haven't tested this change


mnalis:
Makes sense for some of them, but as [wiki](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dstile#Squeeze_stile) says _"They are sometimes V- or Y-shaped"_, which might confuse users _which_ width needs to be entered in such cases?

arrival-spring:
True, although the same could be said about a block or a bollard where there could be different sized gaps

westnordost:
I agree with @mnalis and @matkoniecz, might be difficult to answer. Additionally, I don't know any stile type that would be wheelchair accessible, and that is the primary **use** of this data.
Squeezer stiles certainly aren't, and whether fat hikers will fit through depends on a couple of other factors anyway.

Issue #5312, opened 2023-10-15
Japan (and probably others) - Ability to enter English language shop names


peternewman:
**Use case**
Allow languages which might not exist on street signs to be entered for shop names.

See for example:
https://wiki.openstreetmap.org/wiki/JA:Naming_sample#amenity=cafe
And
https://nsi.guide/index.html?t=brands&k=amenity&v=cafe#starbucks-0cf217

There are lots of places which only seem to display an English as opposed to a Latin transliteration (although I'm not 100% sure how you'd identify which if they didn't include a word like Cafe.

I've erroneously added a few as ja-Latn as I wasn't any the wiser, but it doesn't seem possible to name them correctly via SC.

**Proposed Solution**
I suspect this probably needs an additional file on top of https://github.com/streetcomplete/countrymetadata/blob/master/data/additionalStreetsignLanguages.yml which allows languages which might not exist on street signs to be entered for shop names.

peternewman:
As per https://wiki.openstreetmap.org/wiki/Multilingual_names#Japan this should probably apply for things that aren't shops too (e.g. train stations, paths etc), although I'm not sure there are many of those we can currently name in SC apart from roads (which it probably doesn't apply to).

rhhsm:
I think what you want to say is that when adding names to Japanese PoI, in addition to `name` and `name:ja-Latn`, SC should offer the possibility to add `name:en` (and `name:ja-Hira`, probably?)? That's a matter of adding it to the list at https://github.com/streetcomplete/countrymetadata/blob/master/data/additionalStreetsignLanguages.yml isn't it? That should not be difficult, I suppose: a few other countries already have multiple additional name fields.
I remember Japanese stations (Tokyo subway, JR stations, ...) usually have a name sign with the name in Japanese Kanji (`name`), hiragana (`name:ja-Hira`) and transliteration to latin (`name:ja-Latn`) so SC should allow at least those 3 to be added.
![Tokyo_Subway_Station_sign](https://github.com/streetcomplete/StreetComplete/assets/69100427/b7a9a725-51d1-46b7-90ac-9ebb7bc71eb5)

peternewman:
[...]

PR #6403, opened 2025-07-22
Reimplement Localized names + address in Compose

### LINKED ISSUE #5479: abcd letters as separate keys for mapping house numbers
**Use case**
When mapping house numbers that have an "alphabetic extension" such as: 45b, it is not convenient to use the full keyboard to look up the latin letter you need.
**Proposed Solution**
It would be helpful to have the first 3-4 letters of the alphabet highlighted as separate keys. This would solve the great majority of cases.

westnordost:
also fixes #5479


westnordost:
should also look into #5312 while I am at it

westnordost:
### AddHousenumberForm

https://github.com/user-attachments/assets/bb7cb93e-9020-4a92-a874-629196f320b0

westnordost:
### AddAddressStreetForm

https://github.com/user-attachments/assets/4a4c2f78-93bf-4eb9-9a16-9f0e2d5dc255

westnordost:
### AddRoadNameForm, AddPlaceNameForm, AddBusStopNameForm

https://github.com/user-attachments/assets/b39609f0-53ba-429e-87d8-b50a7d90ace1

westnordost:
### AddressOverlayForm

https://github.com/user-attachments/assets/0d933b34-54da-4562-868b-8988e620f3a4

westnordost:
### AddEntranceReference

https://github.com/user-attachments/assets/3268fba7-de0e-4730-afbf-2fd1d4d3bd77

westnordost:
### Places overlay

 (only the name part, this PR shall not grow into infinity)

https://github.com/user-attachments/assets/4c860e9b-2cf3-4a0e-a982-e960f7445cec

westnordost:
As follow-up: Should revisit the auto-sizing of TextField2, TextField2 is basically not able to apply `Modifier.fillMaxWidth` correctly currently.

Ready for review.

westnordost:
Also as follow-up: drop auto-expand-abbreviations support. See #6416. (Should still contain hint that names should be expanded and should probably still warn user about not expanded abbreviations)

PR #6420, opened 2025-07-30
Align highlighted elements for board type quests

burrscurr:
## Summary

I noticed that the set of highlighted elements for the two board type quests (name and type) was different, which could be confusing when the quests trigger on the same (or nearby) elements. This PR aligns both by expanding the set of highlighted elements for the board type quest. This is a follow-up on #6418.

## Screenshot

https://www.openstreetmap.org/node/9199014037#map=19/50.705403/10.695204

board type highlight

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/board_type/AddBoardType.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/board_type/AddBoardType.kt
@@ -30,7 +30,7 @@ class AddBoardType : OsmFilterQuestType<BoardTypeAnswer>(), AndroidQuest {
     override fun getTitle(tags: Map<String, String>) = R.string.quest_board_type_title
 
     override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
-        getMapData().filter("nodes with tourism = information and information = board")
+        getMapData().filter("nodes with tourism = information")
 
     override fun createForm() = AddBoardTypeForm()
 


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/board_type/AddBoardType.kt
```diff
@@ -30,7 +30,7 @@ class AddBoardType : OsmFilterQuestType<BoardTypeAnswer>(), AndroidQuest {
     override fun getTitle(tags: Map<String, String>) = R.string.quest_board_type_title
 
     override fun getHighlightedElements(element: Element, getMapData: () -> MapDataWithGeometry) =
-        getMapData().filter("nodes with tourism = information and information = board")
+        getMapData().filter("nodes with tourism = information")
```
burrscurr, line 33: Interestingly, the board type quest only triggers on nodes (see elementFilter above), while the board title quest also includes ways and relations. I mirrored this for the highlighting (don't include ways+rels here in the highlighting filter as was done for the board title quest).

Should this be aligned as well? Its probably of low consequence, since `information=board`s are [almost exclusively](https://taginfo.openstreetmap.org/tags/information=board) nodes (about 1 in 400 is a way).

matkoniecz: Aligning them makes sense, my instinct would be to treat way ones as tagging mistakes or nanomapping reaching parody levels

but investigating them to check is this guess true would make sense

matkoniecz:
I see it was tested and seems to make sense to me

I looked at file history including earlier name and https://github.com/streetcomplete/StreetComplete/pull/5882 and it does not seem discussed/changed so far

matkoniecz:
And thanks for a PR!

PR #6428, opened 2025-08-08
added ferry routes to AddMaxWeight

### LINKED ISSUE #6354: Expand AddMaxWeight to Ferries
Currently the AddMaxWeight Quest only asks the maxweight of bridges and around 0.79% of bridges have such a max weight tagged.

Expanding this to ferries (route=ferry) seems rather easy. I would suggest keeping the other limitations on access and motor_vehicle, since we dont want to ask about pedestrian only ferries: https://github.com/streetcomplete/StreetComplete/blob/6eaf52d3eb7027b8d292f907b2b0b02366120523/app/src/main/java/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeight.kt#L26

Currently around 0.79% of bridges have a max weight tag, while 0.787% of ferries have a max weight tag. While around 1.4% of bridges are maxweight:signed=no and no ferries have this tag. This leads me to believe that ferries are more likely to have their maxweight signed, which also matches my experience in my area.

paulklie:
fixes #6354 
Copies the ferry way filter from #6387

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeight.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeight.kt
@@ -20,6 +20,8 @@ import de.westnordost.streetcomplete.quests.max_weight.MaxWeightSign.MAX_WEIGHT
 
 class AddMaxWeight : OsmElementQuestType<MaxWeightAnswer>, AndroidQuest {
 
+    // We ask for the maximum weight of bridges and ferries.
+    // The general filter is used for both:
     private val generalFilter by lazy {
         """ways, relations with
          !maxweight and maxweight:signed != no
@@ -82,10 +84,12 @@ class AddMaxWeight : OsmElementQuestType<MaxWeightAnswer>, AndroidQuest {
         if (!generalFilter.matches(element)) return false
         if (element is Way) {
             if (highwayFilter.matches(element)) return true
+            // This way may be part of a ferry relation, in which case we will not ask
             if (ferryFilter.matches(element)) return null
             return false
         }
         if (element is Relation) {
+            // Highway relations are not matched
             return ferryFilter.matches(element)
         }
         return false


westnordost:
Do some strings need to be updated? Also, doesn't the quest show some traffic signs? Can those signs actually be expected to be present there?

paulklie:
> Do some strings need to be updated?

Seems to be fine to me. "Whats the weight limit here?" fits well.
> Can those signs actually be expected to be present there?

I checked ferries in a few different country's, these are the first ones I found:
Germany: 
image
Austria: 
image
France:
image
Spain: 
image
USA:
image
Canada:
image

It seems to depend, but that is already the case right? We are asking with some very euro-centric sign designs, and the signs in other countires differ strongly (see the examples in https://wiki.openstreetmap.org/wiki/Key:maxweight).

westnordost:
Thanks you for the research! So, that looks good to me.

paulklie:
I think it might be good to add a text title to each of the signs in the sign selection dialogue anyway so that users in countries that do not use these signs can still select the correct one

westnordost:
The signs should be localized, i.e. they should look different in the USA

paulklie:
> The signs should be localized, i.e. they should look different in the USA

oh wow, never noticed, good point.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/max_weight/AddMaxWeight.kt
```diff
@@ -49,6 +62,23 @@ class AddMaxWeight : OsmFilterQuestType<MaxWeightAnswer>(), AndroidQuest {
             }
         }
     }
+
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
+        // the quest shall not be asked for ways tagged with route=ferry that are part of a relation
+        // also tagged with route=ferry because that makes the former not actually a "real" ferry
+        // route (╯°□°）╯︵ ┻━┻. Tagging mistake or not, it is very common tagging (#6373)
```
westnordost:
```suggestion
    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
        // copied from AddFerryAccessMotorVehicle - see comment there why this filtering is necessary
```

westnordost (changes requested):
Please separate the filter for roads from the filter for ferry routes. (You can still have a private `notAnyMaxWeightTagged: String` field which contains the common filter)

The reason is that otherwise, `isApplicableTo` would always return `null` for any road, not only for ferry routes. This is relevant because after editing an element,  `isApplicableTo` of all quests is called for that edited element - to see which quest(s) apply to it now. If **any** quest returns `null` (= cannot be determined without additional data) here, all OSM elements in a certain radius around the element need to be pulled from the database and `getApplicableElements` is run for these again.

I.e. it has a somewhat large performance cost to return `null` in `isApplicableTo` (only mitigated through that the elements are actually cached, so it is not a I/O-operation anymore), so it should be avoided when possible.

PR #6430, opened 2025-08-08
added quests to observation tower

### LINKED ISSUE #6427: Expands quests to observation towers
I believe the following quests could be extended to the 26k [observation](https://wiki.openstreetmap.org/wiki/Tag:tower:type%3Dobservation) towers:

**Things overlay:**
Currently [leisure](https://wiki.openstreetmap.org/wiki/Key:leisure)=[bird_hide](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbird_hide) can be added via the things overlay. An observation tower feels quite similar to a bird observation place, so I think they should be addable to.

**AddStepCount:**
Currently only 189 observation towers  have a step_count set, but it is suggested by the wiki. I have never seen an observation tower that does not have steps. Useful for people with limited physical capabilities,

**AddGeneralFee:**
In my experience the majority of observation towers are free, the current stats show fee=yes 209, no=450. Additionally users are unlikely to encounter multiple observation towers since they are sparse I doubt this would be spamy.

**AddWheelchairAccessOutside:**
Am not quite sure if it would better fit into the AddWheelchairAccessBuisness quest. Currently 88 towers have wheelchair=yes, while 831 have wheelchair=no.

**AddOpeningHours:**
The vast majority of free observation towers are open 24/7, thus I would limit this quest to fee=yes towers.

paulklie:
fixes #6427

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/step_count/AddStepCount.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/step_count/AddStepCount.kt
@@ -11,15 +11,14 @@ class AddStepCount : OsmFilterQuestType<Int>(), AndroidQuest {
 
     override val elementFilter = """
         nodes, ways with
-        ((
+        (
+          (
             highway = steps
             and (!indoor or indoor = no)
             and (!conveying or conveying = no)
+          )
+          or tower:type = observation
         )
-        or
-        (
-            tower:type = observation
-        ))
         and access !~ private|no
         and !step_count
     """
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/step_count/AddStepCount.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/step_count/AddStepCount.kt
@@ -10,11 +10,18 @@ import de.westnordost.streetcomplete.osm.Tags
 class AddStepCount : OsmFilterQuestType<Int>(), AndroidQuest {
 
     override val elementFilter = """
-        ways with highway = steps
-         and (!indoor or indoor = no)
-         and access !~ private|no
-         and (!conveying or conveying = no)
-         and !step_count
+        nodes, ways with
+        (
+            highway = steps
+            and (!indoor or indoor = no)
+            and (!conveying or conveying = no)
+        )
+        or
+        (
+            tower:type = observation
+        )
+        and access !~ private|no
+        and !step_count
     """
     override val changesetComment = "Specify step counts"
     override val wikiLink = "Key:step_count"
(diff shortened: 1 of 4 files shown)


paulklie:
The strings seem to generally fit, am only unsure with the wheelchair quest, what do you think: image

westnordost:
I think this is okay. The hint text for WheelchairAccessBusiness is certainly less well fitting ;-)

paulklie:
> Hmm, are there `nodes` with `highway=steps`? Seems wrong to me, I think it would be more clear if there were two filters.

Sure, though there are less than 3k

westnordost:
Oh, that's weird. What is this supposed to be? Or is it a tagging error? The wiki doesn't mention it at all.

paulklie:
> Oh, that's weird. What is this supposed to be? Or is it a tagging error? The wiki doesn't mention it at all.

There has been some discussion on this topic (https://wiki.openstreetmap.org/wiki/Talk:Tag:highway%3Dsteps#nodes), and I think I agree that tagging ladders and single steps like this could be reasonable. Though I guess we could also ask the step question for these nodes right?

westnordost:
hm yeah, I guess we could

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/step_count/AddStepCount.kt
```diff
@@ -10,11 +10,18 @@ import de.westnordost.streetcomplete.osm.Tags
...
+            highway = steps
+            and (!indoor or indoor = no)
+            and (!conveying or conveying = no)
+        )
+        or
+        (
+            tower:type = observation
+        )
+        and access !~ private|no
+        and !step_count
```
westnordost: `and` has a higher operator precedence than `or`. This expression will return true for private steps (and more importantly: steps that already have `step_count` tagged)

westnordost (changes requested):

paulklie:
Due to the query of the Opening Times quest, users will not be asked for the opening times of towers without a name. I don't think this is to big of an issue.
 I don't think we should add observation towers to the name quest eater, since many towers don't have names (almost 70%).

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/step_count/AddStepCount.kt
```diff
@@ -10,11 +10,18 @@ import de.westnordost.streetcomplete.osm.Tags
...
+            highway = steps
+            and (!indoor or indoor = no)
+            and (!conveying or conveying = no)
+        )
+        or
+        (
+            tower:type = observation
+        ))
+        and access !~ private|no
+        and !step_count
```
westnordost:
```suggestion
        nodes, ways with
        (
          (
            highway = steps
            and (!indoor or indoor = no)
            and (!conveying or conveying = no)
          )
          or tower:type = observation
        )
        and access !~ private|no
        and !step_count
```

clearer indentation

westnordost (approved):

matkoniecz:
See https://community.openstreetmap.org/t/reporting-streetcomplete-issues/135672/3 - it looks like that there are tags that should cause this quest to be excluded for towers.

paulklie:
> See https://community.openstreetmap.org/t/reporting-streetcomplete-issues/135672/3 - it looks like that there are tags that should cause this quest to be excluded for towers.

Thanks for the pointer, as mentioned in the thread I will submit a patch.

PR #6432, opened 2025-08-08
Ask fee of bicycle_wash

paulklie:
Of the only 303 [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity?uselang=en) =[bicycle_wash](https://wiki.openstreetmap.org/wiki/Tag:amenity=bicycle%20wash?uselang=en), 77 have fee=yes while 70 have fee=no.
Its not common, but very usefull to know if it has a fee.


westnordost:
Thanks, makes sense!

PR #6433, opened 2025-08-09
AddLevelThing quest

### LINKED ISSUE #6431: Quest: On which level number is this thing located?
The level quest currently only asks for places. However this information could be very usefull for some "things" to.  This quest could use the same filtering logic (searching for nodes inside of train stations, malls and airports).

I would suggest asking this from everything in the _IsAmenityIndoor_ quest with indoor=yes: https://github.com/streetcomplete/StreetComplete/blob/6eaf52d3eb7027b8d292f907b2b0b02366120523/app/src/main/java/de/westnordost/streetcomplete/quests/amenity_indoor/AddIsAmenityIndoor.kt#L25

Additionally, we should probably also ask for toilets, showers, bins, artworks.

Alternatively we could just ask for everything defined in Things.kt however this would include a bunch of stuff that is unlikely to be inside a building (beehive, whirlpool, bbq etc.)

Otherwise the implementation would be rather identical, and only require a new icon (with a green background) from my understanding right?

PS: this was also suggested in the original issue: https://github.com/streetcomplete/StreetComplete/issues/1487#issuecomment-1093802765

paulklie:
Fixes #6431


westnordost:
It would definitely be necessary to highlight other things, too. E.g. there might be a bench at level 1 and 2 at basically the same location as the one one is asked for. When one switches through the levels, those on 1 and 2 must be visible, otherwise the user will add the wrong level.

westnordost:
But it is anyway a good idea to also highlight shops, because this is necessary for orientation as the map itself is not helpful for indoor things.

westnordost:
Regarding the icon, could just be the same icon as for the place level quest, but with an amenity-green background.

paulklie:
> Regarding the icon, could just be the same icon as for the place level quest, but with an amenity-green background.

Aye, thats what I added.

paulklie:
> It would definitely be necessary to highlight other things, too. E.g. there might be a bench at level 1 and 2 at basically the same location as the one one is asked for

The quest currently does not ask about benches though

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
```diff
@@ -0,0 +1,128 @@
...
+          (
+            emergency ~ defibrillator|first_aid_kit
+            or amenity ~ atm|drinking_water|fountain|payment_validator|food_sharing|payment_terminal|give_box|telephone|toilets|parcel_locker|luggage_locker|locker|post_box|lounge|photo_booth|security_booth|smoking_area|public_bookcase|ticket_validator|shower|vending_machine
+            or man_made = water_tap
+            or historic = memorial
+            or tourism ~ artwork|viewpoint
+            or tourism = information and information !~ office|visitor_centre
+          )
+          and !level
+          and !location or location = outdoor
```
westnordost: huh? Non indoor? Do you mean `location != outdoor`? Also, should look `indoor`, `level:ref`, see the indoor quest.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
```diff
@@ -0,0 +1,128 @@
...
+          and (name or brand or noname = yes or name:signed = no)
```
westnordost:
```suggestion
```
POIs with name, brand etc.?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
```diff
@@ -0,0 +1,128 @@
...
+     * outline */
+    private val filter by lazy { """
+        nodes with
+          (
+            emergency ~ defibrillator|first_aid_kit
+            or amenity ~ atm|drinking_water|fountain|payment_validator|food_sharing|payment_terminal|give_box|telephone|toilets|parcel_locker|luggage_locker|locker|post_box|lounge|photo_booth|security_booth|smoking_area|public_bookcase|ticket_validator|shower|vending_machine
+            or man_made = water_tap
+            or historic = memorial
+            or tourism ~ artwork|viewpoint
+            or tourism = information and information !~ office|visitor_centre
```
westnordost: What speaks against using .isThing() rather than maintain a list manually?

westnordost:
Hmm, I wonder if this could even be put into the `AddLevel` quest.

The question text would need to be adapted. And things should maybe only be highlighted if the element that is asked for is also a thing.

paulklie:
> Hmm, I wonder if this could even be put into the `AddLevel` quest.
> 

I see a few issues with this:

1. Icon colour: Things normaly use green and places yellow, which one do we use for the merged quest?
2. Harder to disable: In busy malls with alot of micromapping, a user might like to disable the "things" quest since this can cause alot of workload. However if it is only a single quest they cannot chose to not disable the (likely to be less intense) places quest.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
paulklie: Mostly because it is hard to differentiate which bench, fire_extinguisher etc. is intended as discussed here: https://github.com/streetcomplete/StreetComplete/issues/6431#issuecomment-3172070619

I tried to form a list of the most useful "things" that are unlikely to have duplicates on multiple levels, and also commonly exist in malls, train stations etc.

westnordost: Well, no, I disagree with @mnalis here - to which level the e.g. fire extinguisher belongs to, or rather, *not* belongs to, can be ascertained by switching through the levels in the quest form UI. If several levels have a fire extinguisher at the same location (e.g. 2 and 4), and only one fire extinguisher is mapped, then inputting either 2 or 4 would be fine and correct. If there's one mapped with `level=2` and another without any level, then when entering "2", that other fire extinguisher will already turn up on the map. (Worst case, then, would be that the user deletes the fire extinguisher number two, wrongly assuming that it is duplicate. And that worst case is not that bad, because a thing-POI in a multi-level mall without level information is somewhat useless anyway.)

westnordost:
Right, I agree, especially on point 2.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
paulklie: Good point.

However I think we should also consider the workload:
Malls and train stations are already some of the most quest-dense places in my experience, with alot of places requiring regular opening-hour updates etc.. Adding a bunch of additional quests asking the level of each planter, bench and fire extinguisher will further increase this huge workload.

Maybe we should instead limit it to the most "important" things, where people are most likely to require level information?

westnordost: Well, a thing-POI in a multi-level mall without level information is somewhat useless. So it makes a lot of sense to define the level of such POIs (or delete them).

It will only add a lot of workload if whoever mapped them in the first place did not specify a level. (E.g. someone using StreetComplete - but then it would be really handy to have a quest like this one that immediately asks for the level)

mnalis: > It will only add a lot of workload if whoever mapped them in the first place did not specify a level. (E.g. someone using StreetComplete

Regarding that, perhaps we can make Things overlay also ask for level (at least inside malls)? Kind of like Places Overlay asks for name when adding e.g. shop... (that should not eliminate a need for this Quest, but would make it less crowded)

westnordost: That came into my mind, too, but really, when that quest pops up right after in that case, that's just as fine.

It would also be somewhat problematic to implement. The level input and controls should only be displayed, after all, if the POI is actually inside an e.g. mall. Since the user can move the crosshair for creating a new POI, one would need to check this on every (pixel) moved. The check for this is also not exactly trivial, i.e. it involves getting the surrounding data from database/cache and doing some point in polygon checks on it.

I mean, it could certainly be implemented, but I don't think it would be worth the effort (and the extra effort to avoid performance issues) when there is already a quest for that.

paulklie: > Well, a thing-POI in a multi-level mall without level information is somewhat useless. 

Tbh, it feels like alot of these things are useless anyway: Fire extinguishers, planters etc. who is using a map to find these? If I would implement this quest in the way you would suggest I think I would also personally disable it.

paulklie: Maybe we could use the age filter here, by asking the level for:

- The most "usefull" things (toilets, water taps, atms etc.)
- Any other thing that has been added recently, maybe in the last hour.

This way we still ask for things the user has just added with less spam.

westnordost: Yeah, a lot of things people map I find really hard to warrant a use case for. It's the tragedy of OpenStreetMap. Someone somewhere wrote one time something along the lines of "on the gravestone of OSM there will be the text 'more was not enough'". The more we map, the bigger the maintenance effort becomes, and that not only for the people who choose to map that stuff, but for all contributors as it adds to the overall complexity.

So what should be mapped and what should not be mapped, there'd need to be a some balance, a relevance-criteria like in wikipedia. But OSM doesn't have that, famously.

At least for this app I am requiring in the quest guidelines that the recorded piece of information must have a _🤷 Useful purpose_ . In this case however, the piece of information is added is useful, orthogonal of the individual usefulness of the thing it is added to.

Are you sure it would be very spammy? I don't expect micro-mapped malls etc. in which all those little things have been added having omitted the `level` tag for those. Maybe just a few, which were omitted in error.

mnalis: > This way we still ask for things the user has just added with less spam.

Hm, maybe it would help, but that feels unexpected and confusing (i.e. AFAIK there is no other quest which works that way)

> Tbh, it feels like alot of these things are useless anyway: Fire extinguishers, planters etc.

OTOH, if some of those are considered truly useless, then they probably should be removed from Things overlay completely :man_shrugging: 

Usefulness is in the eye of the beholder :smiley_cat: - I for one would prefer to map level for fire extinguishers then for say (probably equally spammy) `vending_machine`  or (less spammy but also likely even less useful) `fountain`. But yes, probably very few would like to map level for `man_made=planter` 

But I do somewhat  agree with the idea of having only more useful (and rarer?) Things having a quest asking for a level, but mostly to avoid spamminess and difficulty of solving (esp. due to issues with precise positioning in a mall due to bad GPS signal)

mnalis: > But yes, probably very few would like to map level for `man_made=planter`

Then again, that alleged spamminess problem might not even exist...? 

For example, AFAICT there are exactly [0 planters inside malls in Croatia](https://overpass-turbo.eu/s/2aag) and several neighboring countries.

Also others are maybe not really problematic in practice (e.g. there are [17k `fire_extinguisher`](https://taginfo.openstreetmap.org/tags/emergency%3Dfire_extinguisher) worldwide, and about [75k malls](https://taginfo.openstreetmap.org/tags/shop=mall) -- so even if all of them were mapped inside malls it would _on average_ be much less than one fire extinguisher per mall.

Perhaps @paulklie you could publish a debug `.apk` version  which asks for all Things, and then interested parties can check how spammy it looks in their area?

westnordost: Anyway, your suggestion is clever and would work when the quest would otherwise be somewhat spammy. On the other hand, we *do* allow to add unimportant stuff like planters in the overlay, so it would be somewhat inconsistent to then not include these things in a quest that asks for a property that they have.

On the other hand, we for example don't ask for the surface of driveways in the surface quest because it is spammy and the information value is low..., hmmm.... well, the (small) difference here is that the information value of the `surface` tag on a driveway itself is low because for its primary purpose routing, it is usually always the start or end of a calculated route, so it makes little difference. For `level`, the information itself is not worth less for more unimportant POIs than for important POIs.

paulklie: I have looked around a bit and it seems like the level of micromapping in malls/train stations is indeed much less than I expected. I have modified the filter to ask for all Things and the workload does look quite OK in my local malls, though I think we might have some issues if someone decides, for example, to map every tree inside of the Jewel Changi Airport^

paulklie:
What do you think of also asking for things inside of [shop](https://wiki.openstreetmap.org/wiki/Key:shop)=[department_store](https://wiki.openstreetmap.org/wiki/Tag:shop%3Ddepartment_store)?
These are mostly multi-level in my experience, and often have toilets or similar amenity's.

paulklie:
> It would definitely be necessary to highlight other things, too. E.g. there might be a bench at level 1 and 2 at basically the same location as the one one is asked for. When one switches through the levels, those on 1 and 2 must be visible, otherwise the user will add the wrong level.

Done

westnordost:
> What do you think of also asking for things inside of [shop](https://wiki.openstreetmap.org/wiki/Key:shop)=[department_store](https://wiki.openstreetmap.org/wiki/Tag:shop%3Ddepartment_store)?
These are mostly multi-level in my experience, and often have toilets or similar amenity's.

Other than toilets, I've no idea what one would want to map in a department store. I don't expect department stores to be indoor-mapped for each level.

paulklie:
> > What do you think of also asking for things inside of [shop](https://wiki.openstreetmap.org/wiki/Key:shop)=[department_store](https://wiki.openstreetmap.org/wiki/Tag:shop%3Ddepartment_store)?
> > These are mostly multi-level in my experience, and often have toilets or similar amenity's.
> 
> Other than toilets, I've no idea what one would want to map in a department store. I don't expect department stores to be indoor-mapped for each level.

Hmm,I think aeds, first aid kits, phones, artwork, atms could all also be in department stores

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelForm.kt
```diff
@@ -28,7 +29,7 @@ import org.koin.android.ext.android.inject
 import kotlin.math.ceil
 import kotlin.math.floor
 
-class AddLevelForm : AbstractOsmQuestForm<String>() {
+class AddLevelForm(private val includeThings: Boolean) : AbstractOsmQuestForm<String>() {
```
westnordost: Unfortunately, Android Fragments must have zero constructor parameters. For other forms, we've used inheritance instead, see e.g. `AAddLocalizedNameForm`

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AAddLevelForm.kt
```diff
@@ -0,0 +1,116 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.koin.android.ext.android.inject
+import kotlin.math.ceil
+import kotlin.math.floor
+
+abstract class AAddLevelForm : AbstractOsmQuestForm<String>() {
+
+    abstract suspend fun filter(mapData: MapDataWithGeometry): List<Element>
```
westnordost: `suspend`? I think it does not suspend, does it?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
```diff
@@ -0,0 +1,112 @@
...
+    private val thingsWithLevelFilter by lazy { """
+        nodes, ways, relations with level
+    """.toElementFilterExpression() }
+
+    override val changesetComment = "Determine on which level things are in a building"
+    override val wikiLink = "Key:level"
+    override val icon = R.drawable.ic_quest_level_thing
+    /* disabled because in a mall with multiple levels, if there are nodes with no level defined,
+     * it really makes no sense to tag something as vacant if the level is not known. Instead, if
+     * the user cannot find the thing on any level in the mall, delete the element completely. */
```
westnordost:
```suggestion
```
irrelevant for things

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
```diff
@@ -0,0 +1,112 @@
...
+                if (element.tags.containsKey("level")) {
+                    if (level != null) {
+                        if (level != element.tags["level"]) return@filter true
+                    } else {
+                        level = element.tags["level"]
+                    }
+                }
+            }
+            return@filter false
+        }
```
westnordost: It would make sense to extract this into a helper function in LevelUtils.kt or similar, named e.g. `getMultiLevelMallGeometries(mapData: MapDataWithGeometry): List<ElementPolygonsGeometry>` to avoid duplication.

Or something. Maybe there is another function that could be extracted to remove some duplication.

westnordost (commented):
Good job! I found a few smallish things.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AAddLevelForm.kt
westnordost: (and I wonder how could it compile? `getApplicableElements` does not suspend, so calling a suspendable function in a non-suspendable scope shouldn't even compile)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/AddLevelThing.kt
paulklie: I added a parent abstract class. You think its a good solution?

westnordost: > You think its a good solution?

No, sorry 😬🫣

This would be the first quest type that did that. I'd like to avoid inheritance here. (And also for the forms, this I will do in the migration to compose)

paulklie: better?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/level/LevelUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+        nodes, ways, relations with level
+    """.toElementFilterExpression() }
+
+/* only nodes because ways/relations are not likely to be floating around freely in a mall
+ * outline */
+val filter by lazy { """
+        nodes with
+          !level
+          and (name or brand or noname = yes or name:signed = no)
+    """.toElementFilterExpression() }
```
westnordost: well, uh, that would mean that only for things with a name, a quest would be created.

Also, a project-wide public `val` named `filter` is going to be quite confusing.

Anyway, rather than more forth and back, I will just do the remaining review suggestions myself.

matkoniecz:
it seems that svg file was not committed:

```
Execution failed for task ':generateQuestList'.
> Could not find the SVG for icon 'level_thing' (quest AddLevelThing).
```

paulklie:
> it seems that svg file was not committed:
> 
> ```
> Execution failed for task ':generateQuestList'.
> > Could not find the SVG for icon 'level_thing' (quest AddLevelThing).
> ```

I never created a .svg, did not know it was needed. I only modified the .xml from the level quest to change the colour.
Will have a look at creating one though now.

westnordost:
It's not needed for the app but for that task that helps some people keep the documentation in the OpenStreetMap wiki up-to-date with the app. Also, since the Android drawable resources are somewhat of a proprietary format, it makes sense to keep the vector resources around in a format that can actually be easily edited.

PR #6451, opened 2025-08-17
replace OrganicMaps with CoMaps

westnordost:
organic maps is no longer fully open source


westnordost:
Currently only based on [this comment](https://gitlab.com/fdroid/fdroiddata/-/issues/3605#note_2687424703). OM owners are somewhat mute on the topic, other than accusing CoMaps contributors of "stealing".

So, it might very well be that they will publish that latest code... but at least the current version currently available on Google Play and F-Droid isn't fully open source anymore. And actually, according to [this comment](https://codeberg.org/comaps/comaps/issues/1247#issuecomment-6134581), it may have been like that for years. (In a nutshell: New code for generation of map data which is required for the app to work correctly was not added into public repository).
The guideline for included links is that the software should be free and open source, without ads etc.. Organic Maps still somewhat fulfills this guideline, but the CoMaps fork is, well, fully open source, so that's better to recommend from a FOSS/OSM point of view, even if it may (currently) lag behind in features.

The drama about OM is still unfolding, maybe we should wait with merging this, but then again, the drama has been going on since last year, it seems unlikely that OM team will have a change of heart now.

matkoniecz:
> In a nutshell: New code for generation of map data which is required for the app to work correctly was not added into public repository

to be 100% clear: this code is not part of the app itself (it technically remains open source as it is not part of the app itself), but without it you cannot generate map files needed for app to work

> other than accusing CoMaps contributors of "stealing".

they did it in quite unclear way and refused to elaborate despite requests, but clearly claimed that there is proprietary non-public code

which confirms that project is not fully open-source

Also, they now have quite confusing NOTICE file, and refused to fix it: https://github.com/organicmaps/organicmaps/pull/10987#issuecomment-3160830533

app/src/androidMain/kotlin/de/westnordost/streetcomplete/data/user/achievements/AchievementDefinitionsModule.kt
```diff
@@ -292,11 +292,11 @@ val links = listOf(
...
-        "organic_maps",
-        "https://organicmaps.app/",
-        "Organic Maps",
+        "co_maps",
+        "https://comaps.app/",
+        "CoMaps",
         LinkCategory.MAPS,
-        Res.drawable.link_organic_maps,
+        Res.drawable.link_co_maps,
         R.string.link_organic_maps_description
```
matkoniecz: I guess that we keep this now a bit confusing name (`link_organic_maps_description`) rather than migrate translations manually?

(maybe we should have list of cases like this and at some point rename them?)

westnordost: yeah, its fine

matkoniecz:
I would be happy to merge it and take responsibility for it (OK, I would still need to check whether icon works - I do not remember whether I did it already)

@westnordost would you prefer to wait more with this change?

westnordost:
I'll merge it just before the next release. I'm still undecided whether to keep OM alongside CoMaps or not.

matkoniecz:
> I'm still undecided whether to keep OM alongside CoMaps or not.

it is nearly total duplicate right now and it seems unlikely (for now) that they will diverge much

I would not include both

mnalis:
> I'm still undecided whether to keep OM alongside CoMaps or not.

I don't think SC should keep promoting OM. As @matkoniecz said CoMaps is almost the same, and given OM reluctance to clarify [whether the map generating code is FOSS](https://github.com/organicmaps/organicmaps/issues/11249) or to [explain/correct their accusations of "stealing" FOSS code](https://community.openstreetmap.org/t/organic-maps-open-letter/128851/64) I think OrganicMaps should be just _replaced_ with CoMaps...

matkoniecz:
also that, if they claim that CoMaps is breaking license/similar then they really could have answered followup questions

if CoMaps is not doing that and is doing fork (like Organic Maps did to MAPS.ME), then claiming otherwise is really bad

I waited quite a long time and at that point I am going to assume that accusations from the Open Letter are all accurate

westnordost:
Ref https://github.com/organicmaps/organicmaps/issues/11249

While the maintainer _claims_ it is still open source (well, at least the _app_, which remains true, as far as I know), he also threatens in the same breath the open source fork that takes code from OM with legal action and explicitly reserves the right to withhold parts of the in-production codebase for basically any amount of time, possibly required to run the software properly. 

So, while I wouldn't remove an app from suggestions based on public behavior of the owner, he tries very hard to introduce legal uncertainty into the republishability of its code, which I find unacceptable for an open source project.

---

(I was considering to leave both as suggestions, but decided to just replace OM with CoMaps.)

PR #6462, opened 2025-08-22
disabled its vacant option for vacant places

### LINKED ISSUE #6461: "Its vacant... " on vacant place
Currently, in the places overlay, you can select the "It's vacant..." option for all shops, including ones already marked as vacant. I was about to create a PR to fix this minor oversight, but thought of a better solution: how about replacing it with a "It's **not** vacant" button, that simply triggers the store replacement dialogue in this specific situation?

paulklie:
fixes #6461

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/overlays/places/PlacesOverlayForm.kt
@@ -61,7 +61,7 @@ class PlacesOverlayForm : AbstractOverlayForm() {
     private lateinit var vacantShopFeature: Feature
 
     override val otherAnswers get() = listOfNotNull(
-        AnswerItem(R.string.quest_shop_gone_vacant_answer) { setVacant() }.takeIf { originalFeature != vacantShopFeature },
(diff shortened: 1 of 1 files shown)


westnordost (commented):
I'd slightly prefer to have a function `createVacantAnswer` which does the if check, just for consistency-sake. (Or alternatively, use the `takeIf` also for the other one)

westnordost:
Also, a place can be `shop=vacant` but also e.g. `disused:shop=*` (, `disused:amenity=*`) etc.
How is this handled? Should it be handled?

paulklie:
> I'd slightly prefer to have a function `createVacantAnswer` which does the if check, just for consistency-sake. (Or alternatively, use the `takeIf` also for the other one)

good point, done

paulklie:
> Also, a place can be `shop=vacant` but also e.g. `disused:shop=*` (, `disused:amenity=*`) etc. How is this handled? Should it be handled?

The button does not appear in these cases eather, since the place is still correctly identified as vacant

westnordost:
Thank you!

peternewman:
That's a bit of a shame as I often use this to tidy up an erroneously left name (e.g. a bank that's not just vacant still having it's brand name showing).

PR #6463, opened 2025-08-22
added exersize station equipments to things

paulklie:
This allows these 10 exersize station equipments to be added via the "things" menu: https://github.com/openstreetmap/id-tagging-schema/tree/ff66003e33074b241b2ecaba240b5a128c1dce26/data/presets/leisure/fitness_station

see the corresponding wiki here: https://wiki.openstreetmap.org/wiki/Key:fitness_station
These things mostly use the "fitness station" icon, only the exercise rings, horizontal bar, balance beam and parallel bars have unique icons.


westnordost:
Great!

PR #6464, opened 2025-08-22
Added toilet types to "Things"

paulklie:
This acts as a alternative fix to #6380
This way the user can better specify which type of toilet they wish to add: portable, flush or pitlatrine: https://github.com/openstreetmap/id-tagging-schema/tree/ff66003e33074b241b2ecaba240b5a128c1dce26/data/presets/amenity/toilets
The existing "toilets" stays available so if a user is unsure they can simply fall back to this option.

### DIFF
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
@@ -248,6 +248,7 @@ private val IS_THING_EXPRESSION by lazy {
         or amenity = recycling and recycling_type = container
         or attraction
         or boundary = marker
+        or fitness_station
         or leisure = pitch and sport ~ chess|table_soccer|table_tennis|teqball
         or playground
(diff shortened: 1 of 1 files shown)


westnordost:
I thought #6394 is an alternative to #6380?

Anyway, `amenity=toilets` is already included, so any presets that have `amenity=toilets` plus any additional tags will also be searchable, if they exist in the iD presets. So, this PR does effectively nothing.

paulklie:
> I thought #6394 is an alternative to #6380?

That was additional. I was hoping to make the handwashing more usefull with the disposal quest, since we only ask for non flush toilets.
> 
> Anyway, `amenity=toilets` is already included, so any presets that have `amenity=toilets` plus any additional tags will also be searchable, if they exist in the iD presets. So, this PR does effectively nothing.

Ok wtf, I was super sure it was not possible to add other toilets, but your right. Don't know what happened there, sorry

westnordost:
`¯\_(ツ)_/¯` :-)

PR #6470, opened 2025-08-30
Handwashing is easier to answer than baby changing and wheelchair toilets

peternewman:
As you only need to use your toilet, not potentially find/look at someone else's

Untested, but trivial


paulklie (approved):
good idea

PR #6474, opened 2025-09-02
Check for Shelter existence

paulklie:


westnordost:
Every six years seems to be a bit too often, IMO, since a shelter is a somewhat larger and more immobile structure than a recycling bin or container. I think it's better in the 10 years category.

paulklie:
> Every six years seems to be a bit too often, IMO, since a shelter is a somewhat larger and more immobile structure than a recycling bin or container. I think it's better in the 10 years category.

Hmm, I thought it might be a bit more akin to toilets / showers, but I changed it anyway.

PR #6475, opened 2025-09-02
Ask fee of binoculars

paulklie:
Currently 334 binoculars have fee=no while 475 have fee=yes. This seems to be very well suited for the generall fee quest


matkoniecz:
Is [added exersize station equipments to things](https://github.com/streetcomplete/StreetComplete/pull/6475/commits/365e3965ab354768897306d4d04e08a7028a9dcb) commit supposed to be here?

westnordost:
Makes sense!

paulklie:
> Is [added exersize station equipments to things](https://github.com/streetcomplete/StreetComplete/pull/6475/commits/365e3965ab354768897306d4d04e08a7028a9dcb) commit supposed to be here?

accidentally forked from the wrong branch, but does not matter since that commit was merged into master anyway.

PR #6477, opened 2025-09-03
migrate max weight form to compose

westnordost:
supersedes #6309.

Draft because image selection must be available in compose before this can be completed.

This actually gets rid of a lot of code, but I had to add vector icons for the different signs which are about 600 lines added.


westnordost:
Merged into #6490

PR #6478, opened 2025-09-03
WIP: use strings in compose, move compose stuff to common

westnordost:


HolgerJeromin:
Can we have the file in the new folder and add a preBuild step which copies the files to the old location?

westnordost:
Why?

matkoniecz:
To modify strings in one place, not two, while they need to be duplicated?

westnordost:
they are not modified in this repo, they are imported from poeditor

PR #6489, opened 2025-09-08
OpeningHours: preselect off days

paulklie:
Preselect unselected days when adding "off days" in the AddOpeningHours dialogue

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/adapter/OpeningHoursAdapter.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/adapter/OpeningHoursAdapter.kt
@@ -159,7 +159,7 @@ class OpeningHoursAdapter(private val context: Context) :
     }
 
     fun addNewOffDays() {
-        openSetWeekdaysDialog(null) { weekdays -> addOffDays(weekdays) }
+        openSetWeekdaysDialog(getOffDaysSuggestion()) { weekdays -> addOffDays(weekdays) }
     }
 
     private fun addOffDays(weekdays: Weekdays) {
@@ -315,6 +315,23 @@ class OpeningHoursAdapter(private val context: Context) :
             return Weekdays(result)
         }
         return Weekdays()
+    }
+
+    private fun getOffDaysSuggestion(): Weekdays {
+        // Returns all days that are not selected
+        val mentionedWorkdays = BooleanArray(Weekdays.OSM_ABBR_WEEKDAYS.size)
+        for (row in rows) {
+            if (row is OpeningWeekdaysRow) {
+                row.weekdays.selection.forEachIndexed { index, b ->
+                    if (b) mentionedWorkdays[index] = true
+                }
+            }
+        }
+        val unmentionedWorkdays = BooleanArray(Weekdays.OSM_ABBR_WEEKDAYS.size)
+        mentionedWorkdays.forEachIndexed { index, b ->
+            unmentionedWorkdays[index] = !b
+        }
+        return Weekdays(unmentionedWorkdays)
     }
 
     private fun openSetWeekdaysDialog(weekdays: Weekdays?, callback: (Weekdays) -> Unit) {


westnordost:
This improvement comes at a time where soon I will have to re-write the whole UI for opening hours in compose anyway. So, better I will review it just before I make myself acquainted with that code again in order to rewrite it. (In general, of course, I would prefer not to add enhancements on things that I'll have to rewrite soon anyway, because this or that may fall under the table in a rewrite because I choose to do things differently, but anyway. You'd not be interested in re-writing the opening hours form in Compose, are you? :-P)

Regarding the PR at hand: So, is "public holidays" also pre-selected? Because oftentimes, I see on opening hours plates that Monday is explicitly stated as an off day, but there is no mention about public holidays at all. I usually map and I recommend to map exactly what is seen on the sign. Pre-selecting "public holidays" may come across as a... suggestion... because it is one. (On the other hand, deliberately *not* pre-selecting "public holidays" but e.g. "Monday" would also be a suggestion... so not sure... if it possible to stay neutral?)

paulklie:
> You'd not be interested in re-writing the opening hours form in Compose, are you? :-P)

I'm afraid that that is probably far outside my expertise, I am more of a c++ programmer and have not really ever done kotlin (before the last few patches I submitted). I generally try to avoid ui programming^^

> Regarding the PR at hand: So, is "public holidays" also pre-selected? Because oftentimes, I see on opening hours plates that Monday is explicitly stated as an off day, but there is no mention about public holidays at all. I usually map and I recommend to map exactly what is seen on the sign. Pre-selecting "public holidays" may come across as a... suggestion... because it is one. (On the other hand, deliberately _not_ pre-selecting "public holidays" but e.g. "Monday" would also be a suggestion... so not sure... if it possible to stay neutral?)

Hmm, I agree. I have seen very few signs that mention public holidays. How about:

- If atleast one day of the week has no specified opening hours, we preselect it / them and do not select PH
- If all days of the week have been specified, we preselect PH

westnordost:
> I'm afraid that that is probably far outside my expertise, I am more of a c++ programmer and have not really ever done kotlin (before the last few patches I submitted). I generally try to avoid ui programming^^

Cool! Yeah, I understand. UI programming is quite a different piece of meat, especially Compose with its reactive / declarative approach requires a different way of thinking than normally. By the way, if you are proficient in C++ and want to help StreetComplete, there *is* something where good C++ knowledge is required... but I fear it is also somewhat related to UI: MapLibre (the map renderer we use). In particular, I want to migrate from maplibre-native on Android (C++ core, JNI bindings to Android-specific stuff) to maplibre-compose (C++ core, in the end bindings for the UI framework compose and then everything in between - JNI / bindings to Objective-C on iOS etc.). The latter is currently incubating, its main author is trying to ascertain what he has to do to dovetail the C++ core to make it available in JVM / Objective C / ...
I also contributed a bit to maplibre-compose to help out, but mainly documentation (on the Kotlin part).

Also, in MapLibre core, there is a bug having to do with OpenGL that makes the app crash on certain (old) mobile GPUs and as it so happens the maintainer of SCEE  (fork of SCEE for "experts") is using such an old device and is losing kind of the motivation to continue his work when the app doesn't work on his device anymore. Dunno, that's very specific, probably OpenGL weirdness on certain GPUs is also out of your expertise...

Anyway, 

> Hmm, I agree. I have seen very few signs that mention public holidays. How about:
> * If atleast one day of the week has no specified opening hours, we preselect it / them and do not select PH
> * If all days of the week have been specified, we preselect PH

(The case that all days + PH have been selected need to be covered, too) Indeed, that would be a solution, but... well, my experience is that *if* PH is mentioned, it is mentioned together with Sunday (or whatever day is the "no work day" in a country). In that case, it would be somewhat unhelpful to only pre-select only Sunday.
And usually, there are only 1-2 days that are explicitly mentioned as off days, so I really don't know how useful this preselection would be.

westnordost:
So, all in all, are you fine with this being closed? I don't really see this as a clear improvement.

paulklie:
> So, all in all, are you fine with this being closed? I don't really see this as a clear improvement.

Aye, makes sence. Was really more of a random thought I had one evening

PR #6490, opened 2025-09-10
New image select (WIP)

westnordost:
Also migrate to using compose multiplatform. Will most probably supersede the old image picker to compose PR and then also likely include
- any place where images are picked (i.e. also grouped image picker, image pickers in dialogs)
- street side select ui

Posting as draft PR because I enjoy the GitHub interface for reviewing code committed so far.


westnordost:
# Work completed so far

### Max weight form

in US

https://github.com/user-attachments/assets/4f2fa70a-9d2b-4923-8aac-aa12c098c7b0

in Germany

https://github.com/user-attachments/assets/838a1e9a-84f2-4756-b742-c611b60a7283

### Normal Image picker overlays

Lit

https://github.com/user-attachments/assets/91e73456-b717-4d44-87a2-67919f86cacc

Cycleway

https://github.com/user-attachments/assets/6d2737ef-0159-494b-bc95-ef5e70446f5d

Parking overlay (only adding the narrowing traffic calmings)

https://github.com/user-attachments/assets/f12620c0-f3c0-4ca6-97a0-b717c8086935

MTB

https://github.com/user-attachments/assets/b59743ed-2e93-4856-b233-18362ad82f1f

### Building overlay

https://github.com/user-attachments/assets/c8d4f79d-82db-401c-bf2e-cc164a6bedb7

### Surface overlay

https://github.com/user-attachments/assets/14e48ab9-2246-42ea-828e-ec83c42adf76

### image picker quests

for example...

Sport (multi-selection)

https://github.com/user-attachments/assets/5edea07c-5d20-4a54-adde-c1d890e17837

Cycleway segregated (different item layout)

https://github.com/user-attachments/assets/77531379-3955-45ab-82ee-b7acba20ea7f

Surface (showing the new selection look)

https://github.com/user-attachments/assets/fd7c72fb-d972-4d71-bda2-7e472ad45857

Railway barrier

railway_barrier

Smoothness (slightly new layout)

smoothness_quest

Recycling (multi-select + sub-menu 😮‍💨)

https://github.com/user-attachments/assets/2fd8b25d-3829-4881-a1d4-49dc59b60b4a

westnordost:
I am considering splitting up the lanes quest into two quests for a more standardized UI:

1. Ask whether there are road markings
2. Only if there are road markings, ask about lanes

For the case of the standard residential road without lane markings, this is actually one tap less, for the typical case of `lanes=2` markings, this is only one tap more, given there is a "apply last answer" button

westnordost:
I think the reason why it became a two-step quest at all that tags two things (lane markings + lanes) was because the `lane_markings` stuff was added later after user feedback that it is hard to specify the number of lanes for roads without markings. So, we come up with that `lane_markings` solution. 

But the more StreetComplete-y way is to have a dedicated quest for that. It would even be a simple yes/no quest, very easy.

westnordost:
Not much new, UI-wise. There is a little button that always adds 1 lane for each side (because that's the value one want to set in the vast majority of cases. No need to remember last choice, IMO)

https://github.com/user-attachments/assets/ada1cbe7-7c71-4245-a225-4b2be523ebca

paulklie:
I feel like the "Does this road have lane markings" Quest could use some more information, to inform the user that we are only interested in markings between lanes, and not just the lines around the road. Example https://wiki.openstreetmap.org/wiki/File:Reelsen1.JPG

paulklie:
image
image
Some clipping

paulklie:
"What sport is played here?" crashes when submitting due to a cast error (Set cannot be cast to List).

westnordost:
> I feel like the "Does this road have lane markings" Quest could use some more information, to inform the user that we are only interested in markings between lanes, and not just the lines around the road. Example https://wiki.openstreetmap.org/wiki/File:Reelsen1.JPG

Any ideas? After all, it already asks for **lane** markings, rather than **road** markings.

westnordost:
> Some clipping 

In the end, I decided to not show text in that UI at all anymore, just, if applicable, that floating icon. The reason is that I never managed to make long text always display well. Before, the text was displayed along the orientation of the road - https://github.com/streetcomplete/StreetComplete/blob/fe05308/metadata/en/images/phoneScreenshots/screenshot2.png - but this also lead to that long(er) text was cut-off, **especially** when the road is vertical.
When I changed it so that the text was always upright, like the floating icon, long text would still cut off, **especially** when the road is horizontal.
Plus, if the item had a floating icon **plus** a text, something will always be cut off.

So, I'd rather remove all the text. It's not that bad, too, because the user already can read the text in the selection dialog.

paulklie:
> Any ideas? After all, it already asks for **lane** markings, rather than **road** markings.

Hmm, I did not know the difference before looking them up, and I am unsure if that distinction exists in every language, so it could be lost in translation.

Maybe "Does this road have markings between driving lanes?"

westnordost:
Well, I would consider a separated turn lane that is just one lane, or any lane in a segregated dual carriageway with one lane in each direction also as `road_markings=yes`. Just `lanes=1` in that case.

westnordost:
Looks like it is finally done now. Now I gotta review the code and test

paulklie:
Minor nitpick:
When answering "plastic" in the recycling quest, and selecting and then deselecting an option, this option "remains", It might not be quite clear to a user who selected the wrong type of plastic, how to change it (click on the selected "wrong" plastic again).

I would suggest, when deselecting any plastic option to reset it back to "plastic".
This probably already was the case, but just thought it could be improved here.

westnordost:
https://github.com/user-attachments/assets/ba1ef347-762d-4a19-839c-81009bb7ee00


https://github.com/user-attachments/assets/b4349c3a-284c-42cb-ba8c-51e521c628f4


https://github.com/user-attachments/assets/4348e9d5-455e-45af-9e7d-484654fa9cca


https://github.com/user-attachments/assets/2ea2b189-a15d-45fb-811b-93873a1f175a


https://github.com/user-attachments/assets/02c15244-8c08-44a3-b3ec-085798b078fe


https://github.com/user-attachments/assets/4c6b173d-ff23-4c18-a6fc-3dcd89fb6932

westnordost:
> I would suggest, when deselecting any plastic option to reset it back to "plastic".
> This probably already was the case, but just thought it could be improved here.

After a short look, this seems not easily possible to do.

westnordost:
I clicked through the 2000+ trivial or easy to review files, now remain about 60 that I should read a bit more attentive.

westnordost:
**Remaining to review: (13 files)**
- ~street parking~
- cycleways
- ~lanes~

**Remaining to test:**
- cycleways: could it be that in America or UK, when answering "cycle lane", the quest will just pop up again? Maybe it is also currently a bug?
- cycleways: test re-survey, i.e. where first the current situation is asked and only when pressing "no" the form becomes editable

westnordost:
Alright, done. I don't suppose anyone would want to review this giant. @paulklie did you want to test more, or did you test what you wanted to test?

paulklie:
> Alright, done. I don't suppose anyone would want to review this giant. @paulklie did you want to test more, or did you test what you wanted to test?

Went through and tested all new features in sweden 2 days ago, and it all worked. Dont think anything substantial changed since then right? So I would say LGTM

Issue #6465, opened 2025-08-23
Opening hour quest filter quirks


paulklie:
**Current situation**
Currently we mostly ask for opening_hours only for places with a name (or brand or noname=yes). The reason given in https://github.com/streetcomplete/StreetComplete/issues/224#issuecomment-301276191 (lacking translation for tags) in my understanding, is now resolved. Additionally GPS accuracy has improved in the last 8 years, especially due to the inclusion of the Galileo navigation system in most most modern smart phones.

https://github.com/streetcomplete/StreetComplete/blob/9574a0d89d3fcf7d4ed6f7fce097b2684017c002/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt#L127

A few elements (toilets and barriers), are excluded from this filter (mostly during resurvey) while other are not, for example: charging_station, parking garages, showers and observation towers (the latter two being my fault) are not updated, since they rarely have names. Removing this filter would especially help in these cases.
https://github.com/streetcomplete/StreetComplete/blob/9574a0d89d3fcf7d4ed6f7fce097b2684017c002/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt#L128-L129

We also have a list of places that we we only resurvey, since they rarely have opening times, such as barriers and parks: 
https://github.com/streetcomplete/StreetComplete/blob/9574a0d89d3fcf7d4ed6f7fce097b2684017c002/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt#L117-L121
This makes scene, since we don't want to overwhelm the user with unanswerable quests. However many other places that might rarely have opening times, such as bus/train stations, airports, (botantical-)gardens etc. are never resurveyed. Keeping a list of every possible tag that could have opening_hours set seems like a huge effort, and will undoubtedly create oversights. I understand that we don't want to ask for everything, but why not update / resurvey everything?

Additionally, for all elements, we resurvey the opening hours after 1 year: https://github.com/streetcomplete/StreetComplete/blob/9574a0d89d3fcf7d4ed6f7fce097b2684017c002/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt#L114
[...]

PR #6492, opened 2025-09-11
Opening hours resurvey

paulklie:
Based on the discussion in https://github.com/streetcomplete/StreetComplete/issues/6465.
It changes two things:

1. It removes the name/brand requirement for the elements that only get resurveyed. In effect this only applies to `leisure = park` and `amenity=charging_station` since the other elements are already excluded from the name filter. This does not affect any of the other places that we add opening hours for.
2. I also added these elements to the resurvey filter:
```
leisure ~ garden|beach_resort|sports_centre|disc_golf_course|nature_reserve|playground|fitness_station
or tourism = attraction
or amenity ~ place_of_worship|parking|research_institute|shower|grave_yard|kitchen|marketplace
or railway = station
or aeroway = terminal
or man_made = observatory
```
Thus we only ask for them if they already have (potentially out of date) opening_hours tagged.
These all seem like places that may have opening hours but commonly but also might not have them, so surveying them for elements where opening_hours is not set seems spammy.
All of these places seem to me like places that are unlikely to be confused, so a name is not required to correctly identify them because they are either: large enough so that they can be clearly identified even with inaccurate GPS or rare enough so that they are very unlikely to be close to each other.
3. I have removed a duplicate filter: We filter for name / brand twice, once in the filter and once with the `hasName` function. The former allowed `noname = yes` while the latter does not. This gets rid of code duplication and inconsistency. Instead we only use the filter.
4. I added additional tests to verify the functionality.


westnordost:
Did you change the indentation? This makes it hard to see the changes in a review.

paulklie:
> Did you change the indentation? This makes it hard to see the changes in a review.

Hard to avoid, I had to change the indentation level for most of the filter to adapt to the new logic.
But this view makes it clearer: https://github.com/streetcomplete/StreetComplete/pull/6492/files?diff=split&w=1

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt
```diff
@@ -207,12 +207,42 @@ class AddOpeningHoursTest {
         )))
     }
 
-    @Test fun `isApplicableTo returns true for known places`() {
+    @Test fun `isApplicableTo returns false for unnamed places`() {
+        assertFalse(questType.isApplicableTo(node(
+            tags = mapOf("shop" to "sports")
+        )))
```
westnordost, line 219:
```suggestion
        assertFalse(questType.isApplicableTo(node(
            tags = mapOf("shop" to "sports")
        )))
        
        assertFalse(questType.isApplicableTo(node(
            tags = mapOf("barrier" to "unknown_type_of_barrier")
        )))
```

I believe this test would fail if modified like this. The previous check

```kotlin
    private fun hasName(element: Element) = hasProperName(element.tags) || hasFeatureName(element)

    private fun hasProperName(tags: Map<String, String>): Boolean =
        tags.containsKey("name") || tags.containsKey("brand")

    private fun hasFeatureName(element: Element) = getFeature(element)?.name != null
```

ensured that the situation that for a thing that cannot be named one way or another (either through `name` or it being a named feature in iD), the opening hours quest would never be asked.

I think it makes sense to require it for all elements still. Arguably, this would make sense for any quest, of course. But, at least other quests that have wildcards like `shop and shop !~ no|vacant` in their filter also have `and !name and !brand and noname != yes and name:signed != no` or similar.

westnordost: (Although, I am not sure if actually `barrier=anything` wouldn't match with an iD preset after all and hence it could be named. Then, it would be much less of an issue, it would only be an issue of theoretical nature. E.g. if a thing is added to the filter that doesn't at all exist in the iD preset yet.)

paulklie: I have run the test as suggested and it does not fail. Though I have no idea why.

westnordost: Well, probably it is as I surmised, `barrier=anything` will match the `barrier=*` preset  or something. Well then, since it is a somewhat theoretical issue, it is somewhat of a non-issue then.

westnordost (changes requested):
I believe that now the `getFeature` field is unused for this class, right? Then, it should be removed.

```kotlin
class AddOpeningHours(
    private val getFeature: (Element) -> Feature?
) : OsmElementQuestType<OpeningHoursAnswer>, AndroidQuest {
```

paulklie:
> I believe that now the `getFeature` field is unused for this class, right? Then, it should be removed.
> 
> ```kotlin
> class AddOpeningHours(
>     private val getFeature: (Element) -> Feature?
> ) : OsmElementQuestType<OpeningHoursAnswer>, AndroidQuest {
> ```

done!

app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHoursTest.kt
```diff
@@ -219,7 +219,7 @@ class AddOpeningHoursTest {
         )))
 
         assertFalse(questType.isApplicableTo(node(
-            tags = mapOf("barrier" to "unknown_type_of_barrier")
+            tags = mapOf("barrier" to "block")
```
westnordost: What's this, though? The point of this test is to test that it also works for an unknown type of barrier. Block is certainly known in iD presets.

paulklie: Ahh, my bad, was testing the test and forgot.
Made the test a bit clearer.

westnordost: Oh, no, I am confuse.

**It should be applicable if it can be named somehow.** and has previously set old opening hours.

`barrier=block` should be able to be named somehow, because it is a known iD preset (that assertion should return true)

For `barrier=unknown_type_of_barrier` it should probably be the same.

But we actually gave up on testing this, because we removed the `has any kind of name` check under the assumption that the iD presets will probably call `barrier=unknown_type_of_barrier` "barrier" or similar (you did test this, right?)

---

So, I guess what should be tested is that
1. some things don't require a `name` (etc) for asking about opening hours
2. BUT others (e.g. shops) do
3. for some things, opening hours can be re-asked only if opening hours have been specified before
4. BUT will not be asked if **no** opening hours have been specified before

westnordost:
Right, thanks! LGTM

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/opening_hours/AddOpeningHours.kt
```diff
@@ -18,123 +18,132 @@ import de.westnordost.streetcomplete.osm.opening_hours.parser.isSupportedOpening
...
-              leisure = park
-              or barrier
-              or amenity ~ toilets|bicycle_rental|charging_station
+            or (
+                opening_hours older today -1 years
+                and (
+                    leisure ~ park|garden|beach_resort|sports_centre|disc_golf_course|nature_reserve|playground|fitness_station
+                    or barrier
+                    or tourism = attraction
+                    or amenity ~ toilets|bicycle_rental|charging_station|place_of_worship|parking|research_institute|shower|grave_yard|kitchen|marketplace
```
matkoniecz, line 120: is there any specific reason why amenity=grave_yard got added and landuse=cemetery is not present?

I am also thinking about extracting this set from megastring into another dictionary, though I admit that primary reason would be to make it parsable in my little tool
maybe that is a really bad reason for changing format here

paulklie: Not really, I just did not even consider cemetery (did not know there was a distinction). Extracting sounds good though

PR #6504, opened 2025-09-16
Test fix

paulklie:
The unit tests are failing on my end, however the automated tests seem to work.
The reason seems to be that 0 cannot be casted to the expected StringResource. Maybe there is a better way todo this, but I simply passed the first string resource (it does not seem to matter anyway, since the title is not used in the tests).


westnordost:
Oh, my bad! Thank you for the fix!

PR #6514, opened 2025-09-22
Adding quests suggested in #6457

HugoTHOLLON:
Adding the following quests suggested in #6457:
1. Where is this first aid kit located?
2. In what direction can you ride this? (for an aerial way)
3. How many bikes can be charged here at the same time?
4. How many scooters can be charged here at the same time?
5. Do you have to pay to park your motorcycle here?
6. Does this aerialway transport bikes?

For now, the quests 2, 3, 4 and 6 icons are placeholders and need to be replaced.

These are our first quests on this projects so they might need a lot of review.


mnalis:
Hi @HugoTHOLLON,

I see that several PRs have been recently opened in StreetComplete repository (e.g. this one, but probably also related team #6512, #6513), and then closed.  Closing a PR means you've given up on them.

Was it intentional (i.e. they need more work / testing before you'd suggest them for inclusion in StreetComplete) or was it a mistake? Can we help?

HugoTHOLLON:
Hello @mnalis yes it was intentional, the first 2 PR were mistakes as my team wanted to merge them into 1 and I rushed this one and forgot to make some of the tests and remove some of the file changes.  
We're just begining with PR and weren't organized. We don't need help for now don't worry.

paulklie:
I would also suggest creating one pull request for each quest. This would make it easier to review, and would make it possible to merge some of them while still working on others.

Issue #6390, opened 2025-07-05
BusStopRef in sweden


paulklie:
Currently the AddBusStopRef quest does not ask for swedish stop references.
I can confirm that in the following regions the vast majority or all bus stops have such a ref (a single letter):

- SE-O ([Västra Götaland](https://en.wikipedia.org/wiki/V%C3%A4stra_G%C3%B6taland_County)
- SE-N ([Halland](https://en.wikipedia.org/wiki/Halland_County))
- ([Skåne](https://en.wikipedia.org/wiki/Sk%C3%A5ne_County))

While from my expierience in these regions only very large bus stops will have such refs:

- SE-F ([Jönköping](https://en.wikipedia.org/wiki/J%C3%B6nk%C3%B6ping_County))
- SE-BD ([Norrbotten](https://en.wikipedia.org/wiki/Norrbotten_County))

Maybe some people from other regions (läns) could share their experience?

mnalis:
>  I can confirm that in the following regions the vast majority or all bus stops have such a ref (a single letter):

Yeah I was wondering about that... they all seems to have tagged `A` or `B` (or very rarely `C`) for `ref=*` and practically nothing else. Are we sure those should actually be [ref=* on bus stops](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_stop#How_to_map) ? Is no bus line going more than 2-3 stops ?

To me, `ref` means _"unique identifier of bus stop"_; e.g. if I read `ref=3278` on a bus stop sign, I expect that I could look that number on my mobile app and know exactly where I am. I don't see how it could work if it is only `ref=A` or `ref=B`. 

Is it more like a _direction of bus line_  (e.g. when it goes in one direction, it only stops on `A` stop, and when it goes back in another directions it only stops on `B` stop) than a (per-network) unique reference number?

In Croatia we tag those _"bus stops for different directions"_ by putting them in different relations which are part of superrelation for bus line.  See e.g. https://www.openstreetmap.org/relation/11742416 with its two sub-relations for an example.

> Maybe some people from other regions (läns) could share their experience?

Provided above is not an issue, here is a quick [overpass](https://overpass-turbo.eu/s/27lN) which shows stops having `ref` in green, and those missing it in orange. (Those having `gtfs:stop_id` also get purple outline). Perhaps it might help in determining the predominant situation in various parts of SE in case we lack users with on-the-ground experience there.

paulklie:
[...]

PR #6515, opened 2025-09-22
Disable bus stop name sweden

paulklie:
As previously discussed (https://github.com/streetcomplete/StreetComplete/issues/6390), in most of sweden a different tagging sheme is used, where the "name" field of bus platforms is usually "A","B" etc. The bus stop name is applied to the stop position instead.
This PR disables the quest in sweden entirely, to ensure we don't add any unwanted data.
Inspired by: https://community.openstreetmap.org/t/streetcomplete-i-sverige/135708/2

I have tested this.


westnordost:
Huff, I really dislike country-specific tagging schemata. Makes it more complicated for everyone.

But anyway, could you add the link to the discussion in which this was decided in a comment in the format like it is done in `AddHousenumber`?

westnordost:
Thank you!

PR #6516, opened 2025-09-22
Expand moped acccess quest to sweden.

paulklie:
See https://github.com/streetcomplete/StreetComplete/discussions/6482


westnordost:
Can you actually include the link to the discussion in a comment in the code, for documentation purposes?

westnordost:
Thanks!

PR #6519, opened 2025-09-25
removed market from AddAcceptsCash quest

paulklie:
Market places often have multiple stands. Asking if the entire market accepts cash makes little sence to me as it may differ by stand.


westnordost:
Hmm, doesn't this also apply to food courts?

paulklie:
> Hmm, doesn't this also apply to food courts?

good point, added

Issue #6350, opened 2025-06-16
Expand fee quest to zoo,disc_golf_course,aquarium and public_bath


paulklie:
The AddGeneralFee Quest can be easily expanded to these nodes:

| **tag**                  | **fee=yes** | **fee=no** | **total** |
|--------------------------|-------------|------------|-----------|
| tourism=zoo              | 515         | 858        | 8775      |
| leisure=disc_golf_course | 60          | 659        | 2304      |
| tourism=aquarium         | 15          | 80         | 1551      |
| amenity=public_bath      | 1 104       | 556        | 12165     |

All of these nodes are quite sparse, with most cities having only one or two, thus even though some might be quite similar, the user is unlikely to encounter them repeatedly.

matkoniecz:
`tourism=zoo`  `amenity=public_bath` - is there a real chance that free ones are not already tagged as free?

paulklie:
> `tourism=zoo` `amenity=public_bath` - is there a real chance that free ones are not already tagged as free?

Hmm, a "wildgehege" is a name for an outdoor zoo in germany, and are commonly free. Of the 181 places tagged with that name 174 do not have a fee tagged: https://overpass-turbo.eu/s/268M . 

Not sure about the bath, but I feel like the result here will be similar

westnordost:
Yeah, but these are then tagged as `tourism=zoo` `zoo=enclosure`.

mnalis:
> Yeah, but these are then tagged as `tourism=zoo` `zoo=enclosure`.

Yes, but that would still match `tourism=zoo` element selection, no? Same is with e.g. `zoo=petting_zoo`, which is in my experience (and current tagging) equally likely to be free of paid.

But even if we're to just look at the existing numbers - i.e. if **all** of currently untagged zoos were `fee=yes` (which is at least somewhat unlikely), it would still have about **~10%** which are `fee=no`, which would make the quest suitable IMHO; especially as it could never feel spammy to the mapper (because they are so rarely encountered).

mnalis:
Closing manually as it was implemented by merged PR https://github.com/streetcomplete/StreetComplete/pull/6357 (but missed the `Fixed #6350` magic word so didn't auto-close)

PR #6520, opened 2025-09-25
AddAcceptsCash: limit further tourism values to fee=yes.

paulklie:
Since we now ask for fee of zoo and aquarium(https://github.com/streetcomplete/StreetComplete/issues/6350), we can limit this quest for places that have such a fee=yes.
Additionally I added wilderness_hut since we now also ask for its fee.


paulklie:
On second thought: I have never seen a wilderness hut where you can pay in cash, mostly it is reserved beforehand, so I undid that change.

westnordost:
Nice, good find!

PR #6521, opened 2025-09-25
added further parking to AddWheelchairAccess

paulklie:
parking=underground or rooftop are similar to multi-story: often require an elevator.


matkoniecz (approved):
only looked at code

westnordost:
Yeah, I agree that this makes sense, thanks!

PR #6522, opened 2025-09-25
AddInternetAccess to airport terminal

paulklie:
Around 1.9% of terminals already have internet access tagged.
Tested.


westnordost:
Agree, I think that makes sense.

PR #6531, opened 2025-09-30
Charging station quest

HugoTHOLLON:
Adding the following quests suggested in #6457:

1. How many bikes can be charged here at the same time?
2. How many scooters can be charged here at the same time?

For now, the icon of the quest 2 form is a bike and need to be replaced with a scooter. The icons of both quests are the same as the original charging_station_capacity quest, they could be replaced.

This PR include a french translation of the quests.


matkoniecz:
very quick comment

> This PR include a french translation of the quests.

this is not needed - translations are pulled from translation service and override previous contents, see https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app

> Before each [release](https://github.com/streetcomplete/StreetComplete/blob/master/res/documentation/creating%20new%20release.md), translations are pulled in from POEditor. Please, use POEditor for translating. Manual changes submitted as Pull Requests will not be merged as they do not help the project.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -4,7 +4,7 @@ import de.westnordost.streetcomplete.data.osm.edits.split_way.SplitWayAction
 import kotlin.time.Duration.Companion.minutes
 
 object ApplicationConstants {
-    const val NAME = "StreetComplete"
+    const val NAME = "StreetComplete_SAES5"
```
paulklie: I believe you changed this because @mnalis asked you here https://github.com/HugoTHOLLON/StreetCompleteSAE_S5/issues/40 right?
I think mnalis did this because they misunderstood, and thought you were creating a forked client of Street Complete. However since your are trying to push back to Street Complete you will need to revert this name change for this branch.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacity.kt
```diff
@@ -0,0 +1,42 @@
...
+import de.westnordost.streetcomplete.data.quest.AndroidQuest
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+
+class AddScooterChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
+
+    override val elementFilter = """
+        nodes, ways with
+          amenity = charging_station
```
paulklie, line 18: I think we should also filter out any charging stations with motor_vehicle=yes/designated since some charging stations can be used for both scooters / bikes and cars.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddBikeChargingStationCapacity.kt
```diff
@@ -0,0 +1,42 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.updateWithCheckDate
+
+class AddBikeChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
+
+    override val elementFilter = """
+        nodes, ways with
+          amenity = charging_station
+          and bicycle ~ yes|designated
```
paulklie, line 19: Same here

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -408,6 +410,10 @@ fun questTypeRegistry(
 
     86 to AddClothingBinOperator(),
 
+    179 to AddBikeChargingStationCapacity(),
+
```
paulklie: I think you can remove these empty lines?

paulklie:
Nice work so far!

paulklie:
I also think the quests are better placed together with the car one in quests/charging_station_capacity instead of a new directory.

app/src/androidMain/res/values/strings.xml
```diff
@@ -777,6 +777,8 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
     <string name="quest_barrier_bicycle_installation_openable">Openable</string>
     <string name="quest_barrier_bicycle_installation_removable">Removable</string>
 
+    <string name="quest_bicycle_charging_station_capacity_title">How many bicycle can be charged here at the same time?</string>
```
westnordost:
```suggestion
    <string name="quest_bicycle_charging_station_capacity_title">How many bicycles can be charged here at the same time?</string>
```

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacity.kt
westnordost: Hm well, I guess the situation with these `vehicle` tags is a bit difficult and/or undocumented.

Is a `amenity = charging_station and scooter ~ yes|designated` supposed to be a scooter-only charging station? Or are motor vehicles generally allowed unless specified otherwise? The wiki makes it sound like it is supposed to be the latter. 
On the other hand, unless specified explicitly in the wiki (and even then), we have to assume that we'll find both in the wild.

I am thinking that maybe it would be safer to **always** tag `capacity:scooter` for clarity because `amenity=charging_station` is . (Same for bicycles)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/scooter_charging_station_capacity/AddScooterChargingStationCapacity.kt
```diff
@@ -0,0 +1,42 @@
...
+          and (
+           !capacity
+           or capacity older today -8 years
+         )
+    """
+    override val changesetComment = "Specify scooter charging stations capacities"
+    override val wikiLink = "Tag:amenity=charging_station"
+    override val icon = R.drawable.ic_quest_car_charger_capacity // using the car charger icon because the logo also works for scooters
+    override val isDeleteElementEnabled = true
+    override val achievements = listOf(BICYCLIST) // no scooter achievement
```
westnordost:
```suggestion
    override val achievements = listOf(CAR)
```
Well, I think this fits better, then. (Need to change the import, too)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/bike_charging_station_capacity/AddBikeChargingStationCapacity.kt
```diff
@@ -0,0 +1,42 @@
...
+          and bicycle ~ yes|designated
+          and access !~ private|no
+          and (
+           !capacity
+           or capacity older today -8 years
+         )
+    """
+    override val changesetComment = "Specify bicycle charging stations capacities"
+    override val wikiLink = "Tag:amenity=charging_station"
+    override val icon = R.drawable.ic_quest_car_charger_capacity // using the car charger icon because the logo also works for bikes
```
westnordost: Yes, but bicycle quests generally have a violet background. You can just copy that icon and change the background color.

westnordost:
> I also think the quests are better placed together with the car one in quests/charging_station_capacity instead of a new directory.

I don't mind either way. Don't see a reason for that.

paulklie:
> > I also think the quests are better placed together with the car one in quests/charging_station_capacity instead of a new directory.
> 
> I don't mind either way. Don't see a reason for that.

If https://github.com/streetcomplete/StreetComplete/pull/6536  gets merged there will be shared code that all quests might use in that directory

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacity.kt
paulklie: Could we not reuse the solution in https://github.com/streetcomplete/StreetComplete/pull/6536?
Tag `capacity:scooter` if the charging station supports other modes of transport and `capacity` if not?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
matkoniecz:
```suggestion
    const val NAME = "StreetComplete"
```

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacityForm.kt
```diff
@@ -0,0 +1,9 @@
+package de.westnordost.streetcomplete.quests.charging_station_capacity
+
+import de.westnordost.streetcomplete.quests.AAddCountInput
+import de.westnordost.streetcomplete.resources.Res
+import de.westnordost.streetcomplete.resources.count_bicycle
+
+class AddScooterChargingStationCapacityForm : AAddCountInput() {
+    override val icon = Res.drawable.count_bicycle // TODO add a scooter icon (res/drawable/ic_smoothness_scooter.xml) could work but it needs to be copied in commomMain, right know it can be found in androidMain.
```
westnordost, line 8: Uhh, wait, `ic_smoothness_scooter` is 🛴

Isn't this quest about 🛵 ?

I checked the [wiki on `scooter`](https://wiki.openstreetmap.org/wiki/Key:scooter) and it says that "scooter" is not defined / ambiguous. At least 🛴 is tagged as `small_electric_vehicle`, so maybe that means that `scooter` is now better defined.

app/src/androidMain/res/drawable/ic_quest_bicycle_charger_capacity.xml
westnordost, line 1: Could you supply the same icon as SVG in /res/graphics/quest/bicycle_charger_capacity.svg ?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacity.kt
westnordost: The problem is that the list of "other modes of transport" is somewhat open ended.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacityForm.kt
HugoTHOLLON: I thought the quest was about 🛵 too (because  🛴 is called differently in France) but an electric 🛴 is way more common than an electric 🛵.

So should I keep it like this for now ?

HugoTHOLLON:
@westnordost I'm don't really see the problem about the "other modes of transport" behing open ended. A charging station with multiple vehicule types (e.g. bicycle, car and scooter) might have informations about the bicycle and car's capacity but not the scooter's.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacityForm.kt
paulklie: Ok my bad, I think I confused those two. As you can see here https://github.com/streetcomplete/StreetComplete/pull/6536#discussion_r2398752702 `motorcycle` is the third most used for charging station, while `scooter` is very rare.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacity.kt
HugoTHOLLON: My message is the duplicate of one I made below.

@westnordost I'm don't really see the problem about the "other modes of transport" behing open ended. A charging station with multiple vehicule types (e.g. bicycle, car and scooter) might have informations about the bicycle and car's capacity but not the scooter's.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddScooterChargingStationCapacityForm.kt
paulklie: My bad, I think I also confused `motorcycle` and `scooter`.
Maybe we could just switch this to motorcycle charging capacity?
It is used more often anyway: https://github.com/streetcomplete/StreetComplete/pull/6536#discussion_r2398752702

Interestingly, 95% of the scooter tag usage is on charging stations.

HugoTHOLLON: Could we just do both `motorcycle` and `scooter` ? Like adding a `AddMotorcycleChargingStationCapacity` ?  
It could be intresting to keep the scooter quest because as you said this tag is used mainly on the charging stations.

westnordost: > Interestingly, 95% of the scooter tag usage is on charging stations.

I see, well, then let's stick to `scooter`. It seems to be a somewhat established tagging scheme for charging stations...
**Unless** the charging stations that are tagged with `scooter` are **actually** meant for `small_electric_vehicle` 🛴, i.e. mistaggings! This should be researched first on a test basis.

HugoTHOLLON:
So... I'm not sure what to do now @westnordost @paulklie. 

Is my pull request correct ? 

Do I need to resolve the 2 remaining conflicts even if no modifications were made ?

paulklie:
> Is my pull request correct ?

I think we need to resolve the discussion in https://github.com/streetcomplete/StreetComplete/pull/6536 first, and decide if asking for the `capacity` of bike / scooter charging stations even makes sense.

Pardon me, I should have considered this before adding it to https://github.com/streetcomplete/StreetComplete/discussions/6457

westnordost:
Superseded by #6595

PR #6532, opened 2025-09-30
Motorcycle parking fee quest

ElPoraz:

Do you have to pay to park your motorcycle here?


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_fee/AddMotorcycleParkingFee.kt
```diff
@@ -0,0 +1,36 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CAR
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.fee.applyTo
+import de.westnordost.streetcomplete.osm.maxstay.applyTo
+
+class AddMotorcycleParkingFee : OsmFilterQuestType<ParkingFeeAnswer>(), AndroidQuest {
+
+    // element selection logic by @DerDings in #2507
+    override val elementFilter = """
+        nodes, ways, relations with amenity = motorcycle_parking
```
paulklie, line 15: My first thought was that we would need to refine this filter a bit further, since most [surface](https://wiki.openstreetmap.org/wiki/Tag:parking%3Dsurface) is free, so asking this for every surface parking lot could become spammy.
However, the statistics are not that bad: 1940 with fee=no and 391 with fee=yes: https://overpass-turbo.eu/?template=key-value&key=parking&value=surface
So maybe this is ok?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -4,7 +4,7 @@ import de.westnordost.streetcomplete.data.osm.edits.split_way.SplitWayAction
 import kotlin.time.Duration.Companion.minutes
 
 object ApplicationConstants {
-    const val NAME = "StreetComplete"
+    const val NAME = "StreetComplete_SAES5"
```
paulklie: see https://github.com/streetcomplete/StreetComplete/pull/6531#discussion_r2392359031

paulklie:
Good job with the icon btw! Maybe the coin could be moved to the left a bit so the "motorcycle" icon is a bit clearer?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
ElPoraz: Yes, that makes sense. I just corrected it.

paulklie:
Also: please include the quest icon in a .svg format in https://github.com/streetcomplete/StreetComplete/tree/master/res/graphics/quest

ElPoraz:
> Good job with the icon btw! Maybe the coin could be moved to the left a bit so the "motorcycle" icon is a bit clearer?

I just used the same icon as the one from the bicycle quest, but it's possible that it hides the motorcycle. Here are two other possible versions.

Coin on the left side:

image

Coin on the same place but the motocycle is reversed as the bicycle icon:

image

paulklie:
I think @westnordost might be best to judge this, but I feel like the former is clearer.

ElPoraz:
> Also: please include the quest icon in a .svg format in https://github.com/streetcomplete/StreetComplete/tree/master/res/graphics/quest

Sure, i wait for @westnordost opinion then i'll add it.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_fee/AddMotorcycleParkingFee.kt
westnordost: I think it's fine

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_fee/AddMotorcycleParkingFee.kt
```diff
@@ -0,0 +1,36 @@
...
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.AndroidQuest
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CAR
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.osm.fee.applyTo
+import de.westnordost.streetcomplete.osm.maxstay.applyTo
+
+class AddMotorcycleParkingFee : OsmFilterQuestType<ParkingFeeAnswer>(), AndroidQuest {
+
+    // element selection logic by @DerDings in #2507
```
westnordost:
```suggestion
```
that was for a different quest

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_fee/AddMotorcycleParkingFee.kt
```diff
@@ -0,0 +1,36 @@
...
+            !fee and !fee:conditional
+            or fee older today -8 years
+        )
+    """
+    override val changesetComment = "Specify motorcycle parking fees"
+    override val wikiLink = "Tag:amenity=motorcycle_parking"
+    override val icon = R.drawable.ic_quest_motorcycle_parking_fee
+    override val achievements = listOf(CAR)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_motorcycle_parking_fee_title
```
westnordost: I think we can use simply `quest_parking_fee_title` here instead of adding a new string, because

1. the type of object is mentioned below the question
2. the icon already shows a motorcycle
3. it's shorter
4. it's less effort for translators

westnordost (changes requested):
Can you supply the quest icon also in SVG in /res/graphics/quest/quest_motorcycle_parking_fee.xml

westnordost:
I find your first icon better because the motorcycle is visible better

ElPoraz:
> Can you supply the quest icon also in SVG in /res/graphics/quest/motorcycle_parking_fee.svg ?

Yes, it's done.

paulklie:
@ElPoraz can you add your icon to https://github.com/streetcomplete/StreetComplete/blob/master/res/graphics/authors.txt? Maybe refer to the other icons you used as a base.

westnordost:
This is not necessary. It's too small of a thing to claim copyright over, IMO. (But I don't mind if you do 🤷)

PR #6533, opened 2025-09-30
ask housenumber for buildings inside areas that contain an address no…

### LINKED ISSUE #6500: House Number Quests showing up only sometimes
There's a house that doesn't have a house number although it should have one, but the house number quest doesn't show up on most days. Sometimes, it does show up though. And when it does, I'm sure there's another house number quest or two showing up at the same time in the area that I'm normally not seeing either, though I haven't memorised where they are. It's not an issue of other quests overlapping with these quests, as there aren't any open ones in the vicinity. Just to be sure, I've put the house number quest on the top of my priority list, though that didn't change the behaviour. I don't have any quests hidden currently.

**How to Reproduce**

Not sure what makes it show up sometimes and what doesn't – I'm not changing anything on my end. The address is Blumenthalstraße in Köln (Germany), between house numbers 2 and 6, there's one house (presumably number 4, haven't surveyed it yet) without a number. The house number overlay doesn't show a house number. Cartro and OSMAnd don't show a house number for it either. 

No house number, marked as Mehrfamilienhaus:
Image

Yet no quest: 

Image

I don't have a screenshot of when the quests sometimes does occur, but it was the quest that alerted me to this house, and I saw it at least twice now.

I also just noticed that when I the cache ("Zwischenspeicher löschen") and let StreetComplete re-fetch the data,  for a second the House Number quest shows up before it vanishes. And this way I was able to locate anther vanishing house number quest, a backyard building between Weissenburgstraße 50 and Balthasarstraße 81.

**Expected Behavior**
If there's a quest there of a type I have enabled, it should be shown all the time if there are no quests overlapping it.  

**Versions affected**
61.3

westnordost:
…de but that are associated with another area (fixes #6500)

### DIFF
--- a/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/address/AddHousenumberTest.kt
+++ b/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/address/AddHousenumberTest.kt
@@ -115,6 +115,7 @@ class AddHousenumberTest {
         assertEquals(0, questType.getApplicableElements(mapData).toList().size)
     }
 
+    // https://github.com/streetcomplete/StreetComplete/issues/6500#issuecomment-3353073272
     @Test fun `does create quest for building that is inside an area with an address on its outline that is however associated to another building`() {
         val houseWithAddressNode = Pair(
             Way(1L, listOf(1, 2, 3, 4, 1), mapOf("building" to "detached"), 1),


app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/address/AddHousenumberTest.kt
```diff
@@ -115,6 +115,66 @@ class AddHousenumberTest {
         assertEquals(0, questType.getApplicableElements(mapData).toList().size)
     }
 
+    @Test fun `does create quest for building that is inside an area with an address on its outline that is however associated to another building`() {
```
FloEdelmann, line 119:
```suggestion
    // https://github.com/streetcomplete/StreetComplete/issues/6500#issuecomment-3353073272
    @Test fun `does create quest for building that is inside an area with an address on its outline that is however associated to another building`() {
```

Issue #903, opened 2018-02-22
additional details on charging_station - charging ports


RubenKelevra:
Charging stations offer different types of charging connections to accommodate different charging needs and vehicle types.

The more common ones in Europe are: Type2, Schuko, CHAdeMO and CCS (which is also called type2 combo). In USA there is also Type1 and Type1 CCS (which is also called type1 combo). Additionally Tesla got a widespread charging network with Type2 plugs in Europe and plugs which are pretty similar to Type1 in USA.

I know this might be very confusing if somebody is new to this type of stuff, but it would be great if we could add those details to the charging_station nodes via SC.

### General
Affected tag(s) to be modified/added: [socket:<type>=n]
Question asked: **How many plugs are avaible from which type at this charging station (add total numbers)?**

### Checklist
Checklist for quest suggestions (see guidelines):
- [x] 🚧 To be added tag is established and has a useful purpose
- [x] 🤔 Any answer the user can give must have an equivalent tagging (Quest should not reappear to other users when solved by one)
- [x] 🐿️ Easily answerable by everyone from the outside but a survey is necessary 
- [x] 💤 Not an overwhelming percentage of elements have the same answer (No spam)
- [x] 🕓 Applies to a reasonable number of elements (Worth the effort)

### Ideas for implementation

We need a picture plus name per plug type, like for street surfaces but overlayed by a number and plus/minus buttons.

When the element is selected, all charging plugs should be at an invisble -1 value, if the user presses plus once, the number 0 appears and we'll add this number to the tags.

If the user want to remove the zero again, he should be able to do so by pressing minus again.

**Element selection:** 

If amenity=charging station is present but none which matches to this filter: socket:*=*

We should group the plug types into groups:

DC-Fastcharging:
socket:type1_combo
socket:type2_combo
socket:chademo
socket:tesla_supercharger

AC charging:
socket:type1
socket:type2
socket:type3

General purpose plugs:

Inside Europe:
socket:schuko
socket:cee_blue
socket:cee_red_16a
socket:cee_red_32a
socket:cee_red_64a
socket:cee_red_125a

Inside USA/Canada:
socket:nema_5_15
socket:nema_5_20
socket:nema_14_30
socket:nema_14_50

Inside the UK only:
socket:bs1363

Inside Switzerland:
socket:sev1011_t13
socket:sev1011_t15
socket:sev1011_t23
socket:sev1011_t25

Inside Australia, New Zealand:
socket:as3112

[...]

PR #6536, opened 2025-09-30
AddChargingStationCapacity: answer specific capacity

paulklie:
A charging station can serve cars, hgv, bikes etc. Most only serve a single one, however some serve multiple modes of transport, for example bikes and cars.
For these the `capacity` tag is ill suited, since it is not specific enough. Currently we simply ignore such elements.
Instead we should use `capacity:motorcar` to clearly specify which form of transport has which capacity.
This PR adds a function that checks if the charging station supports any other vehicles. If yes it uses the more specific `capacity:motorcar` instead.

This was briefly suggested in https://github.com/streetcomplete/StreetComplete/issues/6385
I have tested.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/ChargingStationAccessCount.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/ChargingStationAccessCount.kt
@@ -5,7 +5,12 @@ package de.westnordost.streetcomplete.quests.charging_station_capacity
  */
 fun canBeAccessedByOtherVehicle(tags: Map<String, String>, tag: String): Boolean {
     val accessTags = listOf(
-        "motorcar", "hgv", "bus", "bicycle", "goods", "scooter", "motorcycle",
+        // All access tags with at least one usage.
+        "motorcar", "hgv", "bus", "bicycle", "goods", "scooter", "motorcycle", "electric_bicycle",
+        "cargo_bike", "kick_scooter", "caravan", "moped", "speed_pedelec", "mofa",
+        "small_electric_vehicle", "motorhome", "goods", "nev", "psv", "bus", "taxi", "disabled",
+        "wheelchair", "boat", "ship", "train"
+
     ).filter { it != tag }
     return accessTags.any { accessTag ->
         tags[accessTag].let { it == "designated" || it == "yes" }
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
@@ -22,14 +22,6 @@ class AddChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
           (
             motorcar ~ yes|designated
             or motor_vehicle ~ yes|designated
-            or
-            (
-              !motorcar
-              and !motor_vehicle
-              and bicycle !~ yes|designated
-              and hgv !~ yes|designated
-              and scooter !~ yes|designated
-            )
           )
           and access !~ private|no
     """
--- a/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/ChargingStationCapacityTest.kt
+++ b/app/src/androidUnitTest/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/ChargingStationCapacityTest.kt
@@ -22,12 +22,6 @@ class ChargingStationCapacityTest {
         ))))
     }
 
-    @Test fun `applicable to charging station without vehicle`() {
-        assertTrue(questType.isApplicableTo(node(tags = mapOf(
-            "amenity" to "charging_station"
-        ))))
-    }
-
     @Test fun `not applicable to charging station with only other vehicle set`() {
         assertFalse(questType.isApplicableTo(node(tags = mapOf(
             "amenity" to "charging_station",


westnordost:
Hmm, might it not to be expected that a charging station tagged with `bicycle=designated` (and nothing about `motorcar`) is a e-bicycle loading station?

paulklie:
> Hmm, might it not to be expected that a charging station tagged with `bicycle=designated` (and nothing about `motorcar`) is a e-bicycle loading station?

sadly `motorcar=yes` seems to always be implied. osm is very motonormative in that way :(
Maybe we could add a "Uh..." field that sets motorcar=no?

paulklie:
Ok I have randomly analysed 10 charging stations without `motor_vehicle` or `motorcar` set, and 9 of those were obviously only for bikes, so I think you might be on to something. I will see if I can come up with a better solution.

Though I wonder if it is necessary for less than 3500 nodes.

paulklie:
After looking through the tag usage a bit closer, if you add up all the possible combination of bicycle/scooter/hgv.... = yes/designated without motorcar information, you have just over 4000 elements. Compared to the 168k charging stations this is less than 2.4%.
I see three potential solutions:
1. Add an "Uh..." option "This charging station cannot charge cars". This would allow us to clear up some ambiguity/incoherent tagging in the database.
2. Add an additional quest with higher priority that asks if cars can be charged. This would also benefit the remaining 67% of charging stations without the motorcar tag explicitly set, and could be used as a blueprint for future quests ("Can bikes be charged here" etc.)
3. I can modify the filter of this quest to exclude such ambiguous places:

```
motorcar = yes|designated
or motor_vehicle = yes|designated
or
(
  !motorcar
  and !bicycle=yes|designated
  and !hgv=yes|designated
  and !scooter=yes|designated
etc...
)
```
However this excludes some dual use charging stations, and does not help improve the data ambiguity.

I believe the combination of the second and third suggestions is ideal. Since it offers the best experience for users adding new bike charging stations via the "things" menu, and helps reduces the motonormativity on open street map.

paulklie:
I have for now implemented suggestion 3. we can think of implementing suggestion 2 later.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/ChargingStationAccessCount.kt
westnordost, line 1: What about moped, mofa, small_electric_vehicle etc. etc.?

I am not sure if we can have a complete list here. After all, there is also minibus, coach, tricycle (auto-rickshaw / tuk tuk), cargo_bike etc. and anything that people will come up later. I.e. I have the impression that this hierarchy of vehicle types is somewhat open-ended.

https://wiki.openstreetmap.org/wiki/Key:access#Land-based_transportation

paulklie: I included anything with more than 1k uses (on charging_station), thinking that people can add notes otherwise.

westnordost: That doesn't work. The user has no way of knowing that he is supposed to leave a note. E.g. if there is a charging station with

```
amenity=charging_station
bicycle=yes
microcar=yes
```

and the app asks about bicycle charging ports, he can answer that correctly, but the app will tag `capacity` instead of `capacity:bicycle`.

paulklie: Most of them are unused for charging_station. However I will refine the list and add any vehicle with atleast one instance of yes/designated.

| Tag | yes/designated | total |
|-----|----------------|-------|
| bicycle | 9343 | 21486 |
| electric_bicycle | 18 | 19 |
| mtb | 0 | 0 |
| cargo_bike | 1 | 1 |
| kick_scooter | 1 | 1 |
| caravan | 9 | 9 |
| motorcycle | 8425 | 9430 |
| moped | 115 | 158 |
| speed_pedelec | 8 | 8 |
| mofa | 10 | 12 |
| small_electric_vehicle | 117 | 159 |
| motorhome | 22 | 22 |
| tourist_bus | 1 | 1 |
| goods | 2 | 2 |
| hgv | 156 | 5800 |
| agricultural | 0 | 0 |
| hgv_articulated | 0 | 0 |
| bdouble | 0 | 0 |
| auto_rickshaw | 0 | 0 |
| nev | 1 | 1 |
| golf_cart | 0 | 1 |
| microcar | 0 | 0 |
| atv | 0 | 0 |
| ohv | 0 | 0 |
| snowmobile | 0 | 0 |
| psv | 12 | 14 |
| bus | 323 | 4876 |
| taxi | 26 | 27 |
| minibus | 0 | 0 |
| share_taxi | 0 | 0 |
| disabled | 9 | 9 |
| school_bus | 0 | 0 |
| wheelchair | 1451 | 1879 |
| boat | 15 | 15 |
| ship | 5 | 6 |
| train | 1 | 1 |

westnordost: Well, that's somewhat beside the point I am making. There is no guarantee that the values unused *now* will be unused forever.

paulklie: Hmm, I don't see anyway to avoid this entirely.
It also feels very... improbable to me, as it requires all of:

1. A new vehicle type we did not consider
2. That can use the same charging station as a motorcar
3. Where the capacity differs by vehicle
4. With neither of these capacities already tagged (consider that 72% of charging stations already have `capacity` tagged)

matkoniecz: Or someone using novel tagging schema not existing (or not noticed) right now

paulklie: that would still require 2-4 to cause issues though

eisams:
Doesn't capacity follow the connector?

E.g., if you have a charging station with a type2 connector, it is implicit that a car can charge there (and any other vehicle smaller than a car, as long as you have an adapter and can plug in)

Are these bike charging stations just places with schuko outlets that you can plug in? I guess you can then plug in anything that's the same size, e.g., a scooter

paulklie:
> Doesn't capacity follow the connector?
> 
> E.g., if you have a charging station with a type2 connector, it is implicit that a car can charge there (and any other vehicle smaller than a car, as long as you have an adapter and can plug in)

In many cases yes, but there are exceptions: A few days ago I saw a type2 connector used to charge a ferry.
And the capacity is also limited by the availability of parking: you could have two connectors but only one parking for motorcars and one for micro car. Or One is reserved for hgv etc... 

> Are these bike charging stations just places with schuko outlets that you can plug in? I guess you can then plug in anything that's the same size, e.g., a scooter

I have also seen "lockers" inside of which you can charge your bike battery.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/ChargingStationAccessCount.kt
mnalis: >  It also feels very... improbable to me, as it requires all of:

I tend to agree... It does however imply someone should be looking at new values and modify quest as appropriate for all future. It is not a big task, but it is a maintenance effort...

> Hmm, I don't see anyway to avoid this entirely.

Well, we _could_ never tag generic `capacity` and instead **_only_** tag specific capacity for tags which we **explicitly** recognize -- e.g. if there is `bicycle=yes|designated` then ask for and tag `capacity:bicycle=*`, and if there is `motorcar=yes|designated` then ask for and tag `capacity:motorcar=*` etc. That way we would be 100% percent sure to only ask correct questions and only tag correct answers. For things we don't (yet) recognize, quest simply wouldn't be asked (resulting in little less contributions, but no wrong answers, no matter what the future brings).

As that solution would however skip all those `amenity=charging_station` which do **not** specify what vehicles they can charge (and many of which _might_ or _might not_ be implying `motorcar=designated`), we would need separate quest which would ask multianswer quest _**"What vehicles can be charged here?"**_ on all such charging stations, and add appropriate `bicycle=designated`, `motorcar=designated` etc. (which when answered would trigger appropriate capacity quests afterwards)

westnordost:
Hmm, regarding this whole thing, and being in danger to having made several 180° turns, I am now kind of the opinion that if there is no explicit `motor_vehicle=no` or `motorcar=no`, it should be assumed to be a charging station for motorcars **and** that `capacity` always refers to motorcars, despite preliminary research by @paulklie that indicated that an explicit `motor_vehicle=no` is often missing for bicycle-only charging stations.

Because:

- the wiki says so. `amenity=charging_station` is primarily motor vehicle / motorcar focused, in fact, how to tag charging stations for bicycles (etc) is not even mentioned and it is written that `motorcar=yes` is implied if not tagged otherwise

- `capacity` is mostly only meaningful for cars, because a capacity requires parking lots associated to the charging station. 

- For the "capacity" of bicycle-only charging stations, I do think that it makes sense to rather specify the number of sockets, because bicycles don't really need much space, i.e. that's usually not the limiting factor. **Also**, `capacity:bicycle` doesn't make that much sense because as e-bikes are typically charged with the normal domestic socket, it would then be capacity for **any** small electric vehicles such as 🛴 or maybe even 🛵. `capacity:bicycle` would be kind of wrong there.

- if there is a charging station with two associated parking lots and a schuko (type F) for 🛴, 🛵, 🚲, 🛹... then it would be quite problematic to tag `capacity=3 + (capacity:motor_vehicle=2 +) capacity:bicycle=1` because this omits the other things that can be charged **and** forces any data consumer to inspect any `capacity:*` and/or `<vehicle>=yes` tag to make sense of the data, i.e. makes this data super difficult to deal with (including SC, as we are seeing).

So, the notion that `capacity` should be the the sum of all capacities for all vehicle types is useless, that's useless data. `capacity` should always be for motorcars. For capacities for 🛴, 🛵, 🚲, 🛹... in particular, I think it makes much more sense to ask for number of domestic household plugs (to charge these things). I am aware that this unfortunately will make #6531 in its current form obsolete and I am not sure whether if a quest that asks for the number of sockets would be feasible.

For the car `AddChargingStationCapacity` quest, this means that we should probably provide an other-answer option that allows the user to answer "it's not a charging station for cars" that tags (at least) `motorcar=no`.

mnalis:
**TL;DR:** perhaps best ask what types of vehicles can be charged there, and then only tag only specific `capacity:*=*` (avoiding ambiguous `capacity=*`). See below the line break for exact suggestion.

> * the wiki says so. `amenity=charging_station` is primarily motor vehicle / motorcar focused, in fact, how to tag charging stations for bicycles (etc) is not even mentioned and it is written that `motorcar=yes` is implied if not tagged otherwise

Uh, where does it say so? The [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station) I'm seeing in the very first paragraph **explicitly** states _"An EV charging station is a place designated for the use by EVs, such as electric motorcars, trucks and **bicycles**, for the charging of their battery"_? (emphasis mine)

While `motorcar=yes` is indeed implied in absence of other tags; _"how to tag charging stations for bicycles (etc) is not even mentioned"_ claim seems to be incorrect, as it is explicitly explained both under `vehicle` row in the table, i.e.:

> (Charging Stations imply [motorcar](https://wiki.openstreetmap.org/wiki/Key:motorcar)=[yes](https://wiki.openstreetmap.org/wiki/Tag:motorcar%3Dyes), if not please specify with [motorcar](https://wiki.openstreetmap.org/wiki/Key:motorcar)=[no](https://wiki.openstreetmap.org/wiki/Tag:motorcar%3Dno) and reference [vehicle](https://wiki.openstreetmap.org/wiki/Key:vehicle) values for appropriate access. 

... and how to tag bicycles is also explicitly mentioned  both in ["Examples" section](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Examples) and ["Examples of bicycle charging stations" section](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Examples_of_bicycle_charging_stations). In fact, it has double the number of examples for bicycle than for motorcars.

> So, the notion that `capacity` should be the the sum of all capacities for all vehicle types is useless, that's useless data

That I agree with. 

> For the "capacity" of bicycle-only charging stations, I do think that it makes sense to rather specify the number of sockets

That I would agree with too -- number of specific sockets is way more important than capacity, for _both_ bicycles _and_ motorcars. I.e. it matters little that there are, say, 4 parking spaces for charging motorcars, if _none_ of the available sockets are compatible with your car. 

That _"socket types&number quest"_ however is a topic for https://github.com/streetcomplete/StreetComplete/issues/5164 and linked issues (like https://github.com/streetcomplete/StreetComplete/issues/903) [^1]  and is only tangentially related to `capacity:*=*` quest(s).

---

My suggestion however would be:

- If a `amenity=charging_station` has **no** `motorcar=*`, `hgv=*`, `bicycle=*` etc., then have a separate (multiselect image) quest asking _"What types of  vehicles can be charged here?"_ which adds that information. That would be add very useful information on its own, and is additionally very beneficial for implementing capacity quest properly.

- only after that vehicle types are present, explicitly ask for **specific vehicle capacity** for which we think it might make sense (e.g. all or only some of: _"How many e-bikes can be charged here at the same time?"_, _"How many electric cars can be charged here at the same time?"_, _"How many electric trucks can be charged here at the same time?"_ etc) and tag that **specific capacity** key only (e.g. `capacity:motorcar=4`, `capacity:hgv=1`, `capacity:bicycle=4` etc.) and ***avoid*** tagging ambiguous `capacity=*` (using it would can only bring confusion which is unneeded when we can be perfectly clear)

[^1]: implementing socket-types quest(s)  is however IMHO  significantly more complex than this capacity-related quest

westnordost:
Damn, and I see that there are also proprietary (bike) charging sockets, i.e. it's not always the domestic household one...

paulklie:
> Uh, where does it say so? The [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station) I'm seeing in the very first paragraph **explicitly** states _"An EV charging station is a place designated for the use by EVs, such as electric motorcars, trucks and **bicycles**, for the charging of their battery"_? (emphasis mine)

Since I wrote that sentence, I would like to mention that thee whole wiki situation is a bit dicy, It used to have more information on other vehicle sources: https://wiki.openstreetmap.org/w/index.php?title=Tag:amenity%3Dcharging_station&oldid=2871897
This was changed by a user and I was able to restore some information: https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dcharging_station#Vehicle_Information

> In fact, it has double the number of examples for bicycle than for motorcars.

Another relic of me reverting some changes.

paulklie:
> * For the "capacity" of bicycle-only charging stations, I do think that it makes sense to rather specify the number of sockets, because bicycles don't really need much space, i.e. that's usually not the limiting factor. **Also**, `capacity:bicycle` doesn't make that much sense because as e-bikes are typically charged with the normal domestic socket, it would then be capacity for **any** small electric vehicles such as 🛴 or maybe even 🛵. `capacity:bicycle` would be kind of wrong there.

Some things to consider:

1. Not all bike charging stations allow for easy counting. Lockable ones such as the third example here: https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Examples_of_bicycle_charging_stations make counting capacity easy, but it might be hard to find out how many sockets are in each box, especially in countries with multiple types of sockets. I guess this could be somewhat mitigated by filtering out `lockable=yes` stations. Such boxes might also impose limits on battery sizes: bike batteries might fit while scooter ones do not.
2. We would probably also need to filter out battery swapping stations for scooters / motorcycles which for example are big in taiwan right now, since the limiting factor here is the parking space.
3. We currently do not have a socket quest, and since sockets are different in every country such a quest would probably require some form of country data. I do agree that such a socket quest would be better, but since we are unlikely to research/implement the sockets in every single country it would be good to have a "fallback" even once we implement such a quest.
4. Some of the sockets at the charging stations might be reserved for bikes or vice versa. A user would not be able to know this from the sockets alone.

paulklie:
> * `capacity` is mostly only meaningful for cars, because a capacity requires parking lots associated to the charging station.

I think its also very meaningful for hgv. Since hgv require larger parking spots, and some larger spots might be reserved exclusively for hgv. For example this charging station https://www.openstreetmap.org/way/1430680846#map=19/52.358672/7.262173 allows  for only 4 hgv, but more motorcars.

Thus even if the bike/motorcycle capacity tags make less sense, I still think adding more specific motorcar capacity tags via this changeset makes sense.

westnordost:
Alright, I've read this thread again and agree with @mnalis. Rather than trying to snake our way around the ambiguities (in the `amenity=charging_station` wiki page), we should help clear those up. I.e.

- have a quest first that asks what vehicles can be charged here, even if it means in 99% cases that it is "car only"
- then, after that, have quests that ask for car capacity / bike capacity separately, each tagging `capacity:<vehicle type>`

There is still a slight difficulty with capacities for bicycles, as capacity for bicycles might very much be actually the capacity for anything charged with those small electric vehicles plugs, i.e. e-bicycles, e-scooters etc. (and actually **not** bicycles, but e-bicycles, lol). But it looks like `bicycle` seems to have established itself as meaning more or less that in the context of charging stations.

westnordost:
So, consequently, I am going to close this PR as this PR presents a different solution.

westnordost:
But I'll implement the solution myself

westnordost:
Hmpf, after basically having implemented it all, including graphics for the different vehicle types (bicycle, bus, car, truck, motorcycle) I am full of doubt again.

The problem is that for buses and trucks, the user may not be able to answer (correctly). He might think, okay, why not, there is no sign that would disallow this. He might also think, okay, why, there is no sign that allows this (maybe they require "industrial" chargers). So okay, we can leave these two out, so that users would need to add a note in such cases.

But **motorcycle** (i.e. including motor scooters) is the problem. It really depends on the motorcycle, I've read. Lower-end models (i.e. mopeds and similar, stuff limited to city-traffic) might only charge via household plugs (e.g. schuko), just like bicycles, e-scooters etc. (with adapter). Bigger ones / proper motorcycles will likely use the same plugs as cars (e.g. type2). So this means, that it is unlikely that motorcycles will ever be explicitly mentioned *unless* a charging station is specifically designed for motorcycles, like some special charging stations for proprietary charging systems such as a kind of "battery changing station" I've seen a video of once. (But then, it is also not any `motorcycle=yes`.)
So, here, too, users are bound to answer this wrong: Motorcycles are not mentioned, so don't select them -> results in `no` being tagged.

westnordost:
So, maybe this is a solution:

Change nothing much, but add a quest "Can bicycles be charged here, too?" to `amenity=charging_station` without `bicycle` tagged.
Answer options are Yes, No, and Only. Only will also tag `motorcar=no`.

paulklie:
> Change nothing much, but add a quest "Can bicycles be charged here, too?" to `amenity=charging_station` without `bicycle` tagged.
> Answer options are Yes, No, and Only. Only will also tag `motorcar=no`.

I fear that some users might answer "yes" for any charging station with a shuko plug, which I guess it not really incorrect, but confusing for data users.

Maybe this could further be limited by `socket:` data? For example only asking for charging stations with shuko or similar sockets, to make it less spammy?

westnordost:
> I fear that some users might answer "yes" for any charging station with a shuko plug, which I guess it not really incorrect, but confusing for data users.

It's not incorrect, no, not at all.

PR #6537, opened 2025-10-01
fix double percentage symbol

paulklie:
"English is translated 100%%" to "English is translated 100%"
I guess we would need to update in the translations to though.

### DIFF
--- a/app/src/commonMain/composeResources/values/strings.xml
+++ b/app/src/commonMain/composeResources/values/strings.xml
@@ -340,7 +340,7 @@ If you are overwhelmed by the number of quests, you can always fine-tune which q
 You can find donation information on our website or project home, the app store policies do not allow us to provide the link within the app.</string>
 
     <string name="about_title_translate">Translate this app</string>
-    <string name="about_description_translate">%1$s is translated %2$d%%</string>
+    <string name="about_description_translate">%1$s is translated %2$d%</string>
 
     <string name="about_title_privacy_statement">Privacy</string>
 


HolgerJeromin:
Is this a new bug?
Because in my v61.3 installed app this shows fine.
Could be looking strange only because of keyword replacement.

mcliquid:
Same here, 61.3 and it's already good:

mnalis:
Not an Kotlin programmer by any means, but if I had to guess, I'd guess that a literal sequence `%%` is a way to escape single `%` sign (i.e. because things like `%1` expand to other stuff). E.g. like `%%` works in `printf(3)` in C etc.

paulklie:
Aye, it works in 61.3 but not in the current master. it seems to be broken since https://github.com/streetcomplete/StreetComplete/commit/86abf056c7ae5c2972b2aaaaad7813e1dbf2d54e

image

FloEdelmann:
Or more specifically [`c364020` (#6478)](https://github.com/streetcomplete/StreetComplete/pull/6478/commits/c3640204b36e27281c3b7a6024fee92845e1e829), which interestingly has the commit description "fix X is translated %%" :thinking:

westnordost:
> Or more specifically [c364020 (#6478)](https://github.com/streetcomplete/StreetComplete/pull/6478/commits/c3640204b36e27281c3b7a6024fee92845e1e829), which interestingly has the commit description "fix X is translated %%" 🤔

but that commit did exactly the same thing as this PR does now

FloEdelmann:
Oh indeed, I got confused. So it somehow got lost during the merge probably?

FloEdelmann:
Ah, [`c364020` (#6478)](https://github.com/streetcomplete/StreetComplete/pull/6478/commits/c3640204b36e27281c3b7a6024fee92845e1e829) fixed it in "app/src/**androidMain**/res/values/strings.xml", this PR fixes it in "app/src/**commonMain**/composeResources/values/strings.xml".

westnordost:
ah, but there is a build script for that.

Turns out it wasn't included in the "updateStreetCompleteData" job that I run on every update. Did this now.

PR #6538, opened 2025-10-02
Firewood quest

NicodeH:
Issue : #6517 - Is firewood provided here?

Adding firewood quest

PR #6539, opened 2025-10-02
First aid kit quest

NicodeH:
Discussion : https://github.com/streetcomplete/StreetComplete/discussions/6457
Adding first aid kit quest


app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -4,7 +4,7 @@ import de.westnordost.streetcomplete.data.osm.edits.split_way.SplitWayAction
 import kotlin.time.Duration.Companion.minutes
 
 object ApplicationConstants {
-    const val NAME = "StreetComplete"
+    const val NAME = "StreetComplete_SAES5"
```
HolgerJeromin: Should be reverted

HolgerJeromin:
This has parts of #6538 and many more commits which do not belong here

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -300,7 +301,7 @@ fun questTypeRegistry(
 
     // motorcycle parking
     30 to AddMotorcycleParkingCover(),
-    31 to AddMotorcycleParkingCapacity(), // counting + number input required but usually well visible
+    31 to AddMotorcycleParkingCapacity(), // counting + number input required but usually well visible*
```
paulklie: seems like you accidentally added a  * here?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/first_aid_kit/AddFirstAidKitLocation.kt
```diff
@@ -0,0 +1,38 @@
...
+import de.westnordost.streetcomplete.data.quest.AndroidQuest
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.LIFESAVER
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddFirstAidKitLocation : OsmFilterQuestType<String>(), AndroidQuest {
+
+    override val elementFilter = """
+        nodes with
+        emergency = first_aid_kit
+        and !description and !first_aid_kit:description
```
paulklie: why [description](https://wiki.openstreetmap.org/wiki/Key:description) instead of [location](https://wiki.openstreetmap.org/wiki/Key:location)?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/first_aid_kit/AddFirstAidKitLocation.kt
```diff
@@ -0,0 +1,38 @@
...
+        and access !~ private|no
+    """
+    override val changesetComment = "Specify first aid kit location"
+    override val wikiLink = "Tag:emergency=first_aid_kit"
+    //override val icon = R.drawable.ic_quest_defibrillator // Temporary icon
```
paulklie: comment can be removed right?

res/graphics/quest/first_aid_kit.svg
paulklie, line 1: Hmm, I think it defiantly needs a shadow like the other icons.

app/src/androidMain/res/values/strings.xml
```diff
@@ -1169,6 +1169,9 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
     <string name="quest_defibrillator_location">Where is this defibrillator located?</string>
     <string name="quest_defibrillator_location_description">Please provide a concise description of its position (e.g. “in the porter’s lounge”).</string>
 
+    <string name="quest_first_aid_kit_location">Where is the first aid kit located?</string>
+    <string name="quest_first_aid_kit_location_description">Please provide a concise description of its position (e.g. “in the porter’s lounge”).</string>
```
paulklie: this is just a duplicate of quest_defibrillator_location_description. I think you can simply reuse that string to avoid extra translations.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/first_aid_kit/AddFirstAidKitLocationForm.kt
paulklie, line 1: Considering this file is identical to AddDefibrillatorLocationForm, I think it would be better so simply reuse that form here.

paulklie (changes requested):
Nice work so far!

res/graphics/quest/first_aid_kit.svg
matkoniecz: and is shadow oriented like for other icons?

paulklie:
Also if your icon is approved you should add it and your name to https://github.com/streetcomplete/StreetComplete/blob/master/res/graphics/authors.txt

westnordost:
Hmm, has it been discussed whether it would be better to make the `AddDefibrillatorLocation` quest generic enough that it can be asked for other things, too, as it is otherwise a complete duplicate of that quest? Otherwise, we might quickly end up with an array of quests like these, which each need
- an own icon
- an own title
- about 40 lines of code

To generalize the `AddDefibrillatorLocation` quest instead, we'd need to
- find a more generic icon
- more generic title
- change logic in `getHighlightedElements()` to only highlight the exact same feature
- rename quest and form, rename quest also in database via `DatabaseInitializer.kt` once, add alias for the old name in `EditTypeAliasesModule.kt` so that the old quest also appears correctly in statistics

but only once, of course.

---

On the other hand, [having had a look at taginfo](https://taginfo.openstreetmap.org/search?q=%3Alocation), the `*:location` key is **currently** not used (in that way) for any other map feature than the defibrillator. Even `first_aid_kit:location` is basically unused now. Was this tag discussed in the community? It is also not mentioned in the wiki.
Anyhow, my point is that there may be more things for which it is currently not tagged with a description where they are. I don't know, are there? Can anyone come up with more things that could use a precise location description?

---

Independent of the above, does the `AddDefibrillatorLocation` (and then subsequently the `AddFirstAidKit`) even appear at all with the current filtering?

Both filter out any objects with `location` already tagged. Doesn't the quest that asks whether the amenity is indoor add exactly this tag (`location=indoor/outdoor`)?

paulklie:
> Both filter out any objects with `location` already tagged. Doesn't the quest that asks whether the amenity is indoor add exactly this tag (`location=indoor/outdoor`)?

No, it adds `indoor=yes/no` :https://github.com/streetcomplete/StreetComplete/blob/aa9d7224c5d40028583217c2e64de196afb36324/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/amenity_indoor/AddIsAmenityIndoor.kt#L109-L111

paulklie:
> On the other hand, [having had a look at taginfo](https://taginfo.openstreetmap.org/search?q=%3Alocation), the `*:location` key is **currently** not used (in that way) for any other map feature than the defibrillator. 
Anyhow, my point is that there may be more things for which it is currently not tagged with a description where they are. I don't know, are there? Can anyone come up with more things that could use a precise location description?

I see quite a bit of usage for:
- [firehydrants](https://taginfo.openstreetmap.org/keys/fire_hydrant%3Alocation) 
- [changing tables](https://taginfo.openstreetmap.org/keys/changing_table%3Alocation).
- [Post boxes](https://taginfo.openstreetmap.org/keys/post_box%3Alocation)

Other than being useful, I feel like asking the location of an object should be done only if the following apply:

1. The object should not be to common. The free text location quest is one of the most time consuming quests in the app, asking for to many objects wastes users time. for example the over 2 million fire hydrants would quickly overload this app. 
2. The object should be easily accessible. Changing tables, for example, would require surveying every toilet. In countries with gendered toilets this can be illegal. Also this might require paying a fee.
3. The object should be in a complex, ideally indoor, environment: It does not make scene to ask for objects placed outdoors where their location will be obvious from its coordinates. since a description adds little additional information. For example, many post boxes are outdoors and especially in rural settings asking for them will quickly become very tedious.

I cannot think of any other objects that fulfill these requirements, other than the AED and first aid kit.

> Even `first_aid_kit:location` is basically unused now. Was this tag discussed in the community? It is also not mentioned in the wiki. 

I have not found any such discussion, however I would argue that since an first aid kit is very similar to a defibrillator, both in purpose and placement, that this tag makes a lot of sense here.

westnordost:
For fire hydrants, changing tables and post boxes, the key seems to be used in a different way, however. Look at the values. It's not a free-text description but some pre-defined values for e.g. changing tables, for post boxes it is some address and for fire hydrants it's mostly just the road... which doesn't make much sense.

app/src/androidMain/res/values/strings.xml
```diff
@@ -1169,6 +1169,8 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
     <string name="quest_defibrillator_location">Where is this defibrillator located?</string>
     <string name="quest_defibrillator_location_description">Please provide a concise description of its position (e.g. “in the porter’s lounge”).</string>
 
+    <string name="quest_first_aid_kit_location">Where is the first aid kit located?</string>
```
westnordost:
```suggestion
    <string name="quest_first_aid_kit_location">Where is this first aid kit located?</string>
```

westnordost:
I made the icon more similar to the defibrillator icon and better visible when displayed at a small size.

PR #6544, opened 2025-10-03
Aerialway bicycle quest

RaphaLLamothe:

- Does this aerialway transport bikes?

For now, the icon of the quest is SVG handmade with an aerialway. There are yes/no svg for answers and a blue background with a sun if it's only during summer.


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,53 @@
...
+    override val achievements = listOf(RARE, BICYCLIST)
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_aerialway_bicycle_title
+
+    override fun createForm() = AddAerialwayBicycleAccessForm()
+
+    override fun applyAnswerTo(answer: AerialwayBicycleAccessAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        tags["aerialway:bicycle"] = when (answer) {
+            YES -> "yes"
+            SUMMER -> "-1"
```
paulklie: Where find this"-1" value?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt
```diff
@@ -4,7 +4,7 @@ import de.westnordost.streetcomplete.data.osm.edits.split_way.SplitWayAction
 import kotlin.time.Duration.Companion.minutes
 
 object ApplicationConstants {
-    const val NAME = "StreetComplete"
+    const val NAME = "StreetComplete_SAES5"
```
paulklie: see https://github.com/streetcomplete/StreetComplete/pull/6531#discussion_r2392359031

paulklie (changes requested):

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
paulklie: https://wiki.openstreetmap.org/wiki/Key:aerialway:bicycle says "summer"

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,53 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.RARE
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.YES
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.SUMMER
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO
+
+class AddAerialwayBicycleAccess : OsmElementQuestType<AerialwayBicycleAccessAnswer>, AndroidQuest {
+
+    private val filter by lazy {
+        "ways, relations with aerialway and !aerialway:bicycle and !bicycle".toElementFilterExpression()
```
paulklie: All aerial ways?
I think this would need some more filtering, maybe some research which aerial ways commonly accept bicycles.

paulklie:
Might want to add an option "there is no sign" that tags aerialway:bicycle:signed=no. What do the others think?

app/src/androidMain/res/drawable/ic_quest_aerialway_bicycle.xml
paulklie, line 1: Can you include the original .svg in https://github.com/streetcomplete/StreetComplete/tree/master/res/graphics/quest?

The other shadows are all "below", but yours is more to the right, could you unify?
Also I think a purple background would be more appropriate, since this is a bike quest.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
paulklie: | aerialway type | with aerialway:bicycle=yes | without aerialway:bicycle | other aerialway:bicycle value set | total |
|----------------|-----------------------------|-----------------------------|-------------------|-------|
| cable_car | 22 | 1065 | 36 | 1123 |
| gondola | 49 | 1752 | 154 | 1955 |
| chair_lift | 19 | 8720 | 186 | 8925 |
| mixed_lift | 1 | 97 | 5 | 103 |
| drag_lift | 3 | 4223 | 3 | 4229 |
| t-bar | 1 | 2876 | 4 | 2881 |
| j-bar | 0 | 215 | 0 | 215 |
| platter | 0 | 4432 | 20 | 4452 |
| rope_tow | 0 | 1744 | 2 | 1746 |
| magic_carpet | 0 | 3130 | 4 | 3134 |
| zip_line | 0 | 2566 | 1 | 2567 |
| goods | 0 | 2709 | 0 | 2709 |

It seems like cable_car, gondula and chair_lift are reasonable candidates.

app/src/androidMain/res/drawable/ic_quest_aerialway_bicycle.xml
RaphaLLamothe: Hello, I reviewed all the comments and adjusted my quest based on all your feedback. I added all the SVGs, fixed the incorrect values, and also updated the XML for the quest.

paulklie: Thank you, can you hit "resolve" on everything you fixed?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override fun getTitle(tags: Map<String, String>) = R.string.quest_aerialway_bicycle_title
+
+    override fun createForm() = AddAerialwayBicycleAccessForm()
+
+    override fun applyAnswerTo(answer: AerialwayBicycleAccessAnswer, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        tags["aerialway:bicycle"] = when (answer) {
+            YES -> "yes"
+            SUMMER -> "summer"
+            NO_SIGN -> ":signed=no"
```
paulklie: Did you test this? To me it seems like it would tag it as `aerialway:bicycle=:signed=no`
You can test your work by disabling autoupload and check what tags are added with the "revert" functionality.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,55 @@
...
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.YES
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.SUMMER
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO_SIGN
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO
+
+class AddAerialwayBicycleAccess : OsmElementQuestType<AerialwayBicycleAccessAnswer>, AndroidQuest {
+
+    private val filter by lazy {
+        "ways, relations with aerialway~cable_car|gondola|chair_lift and !aerialway:bicycle and !bicycle".toElementFilterExpression()
```
paulklie: Add some line breaks and indentation here, similar to other quest filters.

app/src/androidMain/res/values/strings.xml
```diff
@@ -723,6 +723,12 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
     <string name="quest_address_street_street_name_label">Street name:</string>
     <string name="quest_address_street_hint2">Tap street on map to select it</string>
 
+    <string name="quest_aerialway_bicycle_title">Does this aerialway transport bicycles?</string>
+    <string name="quest_aerialway_bicycle_yes">Yes</string>
```
paulklie: There should be some other "Yes"/"No" strings here that you can reuse to avoid unneeded translations.

app/src/androidMain/res/values/strings.xml
```diff
@@ -723,6 +723,12 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
...
+    <string name="quest_aerialway_bicycle_summer">During summer</string>
```
paulklie: Suggestion, maybe "Only during summer" to be clearer?

app/src/androidMain/res/values/strings.xml
```diff
@@ -723,6 +723,12 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
...
+    <string name="quest_aerialway_bicycle_no_sign">No information</string>
```
paulklie: I think here you can also reuse the "it is not signed" that we have somewhere.

paulklie:
I also believe that AImageListQuestForm is probably not the best form type here since I don't think the images add much information.
I believe a simple list like https://github.com/streetcomplete/StreetComplete/blob/master/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/camping/AddCampType.kt might be better suited, but maybe some discussion here would be good.

app/src/androidMain/res/values/strings.xml
mcliquid: or something like "Yes, all year round"?

RaphaLLamothe:
> I also believe that AImageListQuestForm is probably not the best form type here since I don't think the images add much information. I believe a simple list like https://github.com/streetcomplete/StreetComplete/blob/master/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/camping/AddCampType.kt might be better suited, but maybe some discussion here would be good.

I have an issue with this situation. As I added a string for the summer answer, I don't know how to add properly my string in the Res file (the commonMain strings.xml). I want to do it because I am trying to follow the camping quest, such as you recommended me. Thank you in advance for your response.

westnordost:
You always have to add your strings to the `strings.xml` in `androidMain/res/values/`, that's the "source" translation file. To copy over the strings to `commonMain`, you can execute the gradle task `app:copyDefaultStringsToEnStrings`.

Sorry about that (duplication), we are currently migrating to another UI framework. That the strings are now duplicated into two directories is a temporary solution, as the new UI code uses the strings from commonMain and the old UI code uses the strings from androidMain.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,56 @@
...
+import de.westnordost.streetcomplete.data.quest.AndroidQuest
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.BICYCLIST
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.RARE
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.YES
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.SUMMER
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO_SIGN
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO
+
+class AddAerialwayBicycleAccess : OsmElementQuestType<AerialwayBicycleAccessAnswer>, AndroidQuest {
```
westnordost: Take a look a t `OsmFilterQuestType`.

I believe you can simply inherit from that class. This would remove the boilerplate code of `getApplicableElements` and `isApplicableTo` as they are unneeded. In fact, the implementation is a bit wrong anyway.

res/graphics/quest/quest_aerialway_bike.svg
westnordost, line 1: Quite nice!

I have a few suggestions to improve it, though. But it is not necessary to improve it for this PR to be merged. (I may improve it myself, then).

- Grey on pink is not that well visible. I'd suggest to make the lift white (or another color).
- Also, the line of the aerialway car could span the whole icon, it doesn't need to be cut off before it reaches the outline of the circle. 
- The icon would look more clean / less busy IMO if the bicycle was *in* the lift, i.e. a bicycle in a box that hangs off a line. I am not entirely sure if this would still be recognizable well as a lift without the "window". Something to try out.
- personal opinion: I like the yellow for the bicycle better, it mixes better with the pink background
- the shadow is inconsistent. IIRC the shadow of all icons is always exactly 4px under the icon itself (check other icons)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,39 @@
...
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.SUMMER
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO_SIGN
+import de.westnordost.streetcomplete.quests.aerialway.AerialwayBicycleAccessAnswer.NO
+
+class AddAerialwayBicycleAccess : OsmFilterQuestType<AerialwayBicycleAccessAnswer>(), AndroidQuest {
+
+    override val elementFilter = """ways, relations with
+        aerialway ~ cable_car|gondola|chair_lift
+        and !aerialway:bicycle and !bicycle
+    """
```
westnordost:
```suggestion
    override val elementFilter = """
        ways with
          aerialway ~ cable_car|gondola|chair_lift
          and !aerialway:bicycle and !bicycle
    """
```

formatting + no relations

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AerialwayBicycleAccessForm.kt
```diff
@@ -0,0 +1,18 @@
...
+    override val buttonPanelAnswers = listOf(
+        AnswerItem(R.string.quest_generic_hasFeature_no) { applyAnswer(AerialwayBicycleAccessAnswer.NO) },
+        AnswerItem(R.string.quest_generic_hasFeature_yes) { applyAnswer(AerialwayBicycleAccessAnswer.YES) }
+    )
+
+    override val otherAnswers = listOf(
+        AnswerItem(R.string.quest_aerialway_bicycle_summer) { applyAnswer(AerialwayBicycleAccessAnswer.SUMMER) },
+        AnswerItem(R.string.quest_hairdresser_not_signed) { applyAnswer(AerialwayBicycleAccessAnswer.NO_SIGN) }
+    )
+}
```
westnordost, line 18: "summer" is in other-answers, why?

According to taginfo, this is the most-used value: https://taginfo.openstreetmap.org/keys/aerialway%3Abicycle#values

app/src/androidMain/res/drawable/ic_no_answer.xml
westnordost, line 1: What are these files for?

app/src/androidMain/res/drawable/ic_no_sign.xml
westnordost, line 1: What are these files for?

app/src/androidMain/res/drawable/ic_yes_answer.xml
westnordost, line 1: What are these files for?

app/src/androidMain/res/drawable/ic_bicycle_during_summer.xml
westnordost, line 1: What are these files for?

westnordost (changes requested):

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AerialwayBicycleAccessForm.kt
RaphaLLamothe: To be honest, I didn't understand precisely how the AbstractOsmQuestForm worked, I figured that the buttonPanelAnswers where only "yes" and "no", and that the other answers were in the second list. I will change it really soon

app/src/androidMain/res/drawable/ic_bicycle_during_summer.xml
RaphaLLamothe: These files were for the first version of the quest with the images for all answers. When I updated my work, I should have deleted these files. I will create a new commit to correct everything.

res/graphics/quest/quest_aerialway_bike.svg
RaphaLLamothe: I am thankful for all these suggestions. I am actually working with InkScape, and it doesn't allow me to insert a good shadow properly. As the aerialway is from SVG Repo, I am not allowed (except if I am wrong) to edit the "item", and thus, with the added white lines, the shadows are superimposed.

paulklie: what lisence is the original from svg repo under?

RaphaLLamothe: The license is PD License, which means Public Domain License if I am not wrong

paulklie:
You forgot to include the .svg in your commit ;)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AerialwayBicycleAccessForm.kt
westnordost: You can add any number of answer items to buttonPanelAnswers.

The reason why you see this done very little, or only with short answers, such as for the vegetarian quest (yes, no, only), is space. I think it would look very bad if the button texts were longer, as that button bar would then span several lines, maybe even the buttons themselves would span several lines, depending on how long the translated texts turn out to be.

That is why, for quests that offer to select one of several texts, we use a different UI. See for example

https://github.com/streetcomplete/StreetComplete/blob/c9e02f828dbd3c13ff4c79b7d33742299e916939/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/board_type/AddBoardTypeForm.kt

As you see, it is *very* easy to use, just a few lines of code. (See `AListQuestForm` for how it works)

app/src/androidMain/res/drawable/quest_aerialway_bicycle.xml
westnordost, line 1: unused file(?)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AerialwayBicycleAccessForm.kt
westnordost: Anyway, not sure if we should/need to use the "radio button" UI for this. After all, the button-bar UI is a bit more sleek/quick to use.

The answer option for only during summer could maybe also be shortened to **"Only in summer"** (-during +in) in English, maybe it would also be short (enough) in other languages to not look odd? (You can test it)

- "Nur im Sommer"
- "solo en verano"
- "juste en été" (?)

westnordost:
> You forgot to include the .svg in your commit ;)

Speaking of which, even if the license is public domain, it is better to add attribution (with link to source and license information) in any case. This is done in the authors.txt in the /res/graphics/ directory.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AerialwayBicycleAccessForm.kt
matkoniecz: In Polish it could be "tylko latem" (a bit uncommon case of Polish text being shorter than English)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
```diff
@@ -0,0 +1,40 @@
...
+
+class AddAerialwayBicycleAccess : OsmFilterQuestType<AerialwayBicycleAccessAnswer>(), AndroidQuest {
+
+    override val elementFilter = """
+        ways with
+          aerialway ~ cable_car|gondola|chair_lift
+          and !aerialway:bicycle and !bicycle
+    """
+
+    override val changesetComment = "Specify aerialway access for bicycles"
```
matkoniecz: maybe 

```suggestion
    override val changesetComment = "Specify whether bicycles can be taken in aerialway"
```

would be more clear? Disclaimer: I am not a native speaker.

And I guess that "aerialway" osmism can be used instead of English "aerial lift"?

paulklie: I think "on" is normally the correct Preposition for lifts.

matkoniecz: entirely possible

app/src/androidMain/res/drawable/quest_aerialway_bicycle.xml
RaphaLLamothe: It had to replace the previous quest image, but something didn't work correctly

app/src/androidMain/res/drawable/ic_quest_aerialway_bicycle.xml
paulklie, line 1: Note that the guideline for naming quests icons has been changed in https://github.com/streetcomplete/StreetComplete/pull/6490/files

westnordost:
I remade the quest icon to my liking. I wouldn't say that it is clearly better than yours, but there were open issues with it anyway (missing shadow, you didn't update the SVG) anyway, so I took this as an opportunity as a shortcut to further PR forth-and-back.

(I removed the attribution to maki icons because I actually re-did the cabin - it's just a few lines anyway...)

PR #6545, opened 2025-10-03
Kind of lock quest 

RaphaLLamothe:

- What kind of lock is this?

For now, the icon of the quest is SVG handmade with a representation of a side-view lock. The answers icon are an open lock for self-service locks, a closed lock for manual ones, and an other with a toothed wheel for automated locks.


paulklie:
See https://github.com/streetcomplete/StreetComplete/issues/6540.
I dont think this is a great solution for ship locks.

Issue #6422, opened 2025-08-01
Add power=pole to the Things selection


vfosnar:
add `power=pole` to the Things selection

matkoniecz:
not sure how useful is to map power poles without mapping power lines

vfosnar:
I would find it useful. It's pretty easy to deduce where the power lines are from aerial imagery, mostly from casted shadows, but figuring where the poles are is basically impossible in some countries.

Edit: also obviously having a line of power poles makes deducing power lines much easier too, tho less useful at intersection and larger cities

westnordost:
I think there were omitted because they are usually vertices in a way. But I guess it is fine to map them also as free-standing nodes (intermittently).

However, isn't this then inconsistent to not also include power towers? But power towers, power portals, transformer towers etc. would definitely not be a "thing" anymore, but some structure, which are not included in the overlay.

vfosnar:
> But power towers, power portals, transformer towers etc. would definitely not be a "thing" anymore, but some structure, which are not included in the overlay.

I guess, but these are not as common so I can create a note and solve it from a pc later. Adding a note for each pole would be awful.

PR #6546, opened 2025-10-03
added things

paulklie:
See https://github.com/streetcomplete/StreetComplete/discussions/6473
I have tested this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Housenumbers: Ask for some buildings that were previously excluded (#6533)
 
 ### Overlay improvements
+
 - Things: Added Exersize Equipment, Seamarks, Disc Golf, Tomb, Grave, Pillory and Aerialway Pylon as things
 - Places: don't show "it's vacant" option for already vacant places (#6462), by @paulklie
 - Addresses: Allow omitting the street name (#6528)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 
 ### Overlay improvements
 
-- Things: Added Exersize Equipment, Seamarks, Disc Golf, Tomb, Grave and Pillory as things
+- Things: Added exercise equipment, moorings, buoys, disc golf, graves and pillory as things
 - Places: don't show "it's vacant" option for already vacant places (#6462), by @paulklie
 - Addresses: Allow omitting the street name (#6528)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@
 - Housenumbers: Ask for some buildings that were previously excluded (#6533)
 
 ### Overlay improvements
-- Places: don't show "it's vacant" option for already vacant places (#6462), by @paulklie
+- Things: Added Exersize Equipment, Seamarks, Disc Golf, Tomb, Grave, Pillory and Aerialway Pylon as things
+- Places: Don't show "it's vacant" option for already vacant places (#6462), by @paulklie
 - Addresses: Allow omitting the street name (#6528)
 
 ## v61.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 
 ### Overlay improvements
 
-- Things: Added Exersize Equipment, Seamarks, Disc Golf, Tomb, Grave, Pillory and Aerialway Pylon as things
+- Things: Added Exersize Equipment, Seamarks, Disc Golf, Tomb, Grave and Pillory as things
 - Places: don't show "it's vacant" option for already vacant places (#6462), by @paulklie
 - Addresses: Allow omitting the street name (#6528)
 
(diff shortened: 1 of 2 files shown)


app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
```diff
@@ -42,6 +42,9 @@ private val IS_THING_EXPRESSION by lazy {
             "windsock",
             "gate",
         ),
+        "aerialway" to listOf(
+            "pylon"
```
matkoniecz: this should be joined to aerialway line, right? not freefloating?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
```diff
@@ -256,6 +263,7 @@ private val IS_THING_EXPRESSION by lazy {
           or trolleybus = yes
           or tram = yes
         )
+        or seamark:type
```
matkoniecz: this sounds quite open-ended

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
paulklie: Aye, see https://github.com/streetcomplete/StreetComplete/issues/6422 for a similar discussion on power poles

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
paulklie: Currently includes 6 objects:
https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/seamark
You can see what they look like here:
https://wiki.openstreetmap.org/wiki/Seamarks/Buoys
I feel like they all fit quite well

westnordost: It is not guaranteed that iD will forever only support that limited amount of presets.

Blanket-adding `seamark:type` is not okay. Each seamark type to be added must be reviewed separately. E.g. mooring buoys are okay and so are bollards, about dolphins I am already not sure.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
westnordost: That's definitely a structure:

https://wiki.openstreetmap.org/wiki/File:Pilone_cabinovia_Arabba_Porta_Vescovo.JPG

Please remove it.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
```diff
@@ -168,6 +171,8 @@ private val IS_THING_EXPRESSION by lazy {
             // "ship" - too big, more like a structure/building
             "stone",
             "tank",
+            "pillory",
+            "tomb",
```
westnordost: A tomb seems to be rather a structure.

westnordost (changes requested):
.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
paulklie: Will do, but I think the structure definition is a bit harder to follow here. Is this a structure? https://mnd.com/wp-content/uploads/2020/12/mndgroup-gallery-leisure-geilo-3-1600x1800.jpg.webp

westnordost: Well, that's borderline, the structure itself is not much larger than what you may find on a playground. However, the point is that `aerialway=pylon` **may** be a bigger structure.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/Things.kt
```diff
@@ -229,6 +230,12 @@ private val IS_THING_EXPRESSION by lazy {
             "tree",
             "tree_stump",
         ),
+        "seamark:type" to listOf(
+            "buoy_isolated_danger",
+            "buoy_lateral",
+            "mooring",
+            "buoy_safe_water",
+        ),
```
westnordost, line 238: Any kind of [mooring](https://wiki.openstreetmap.org/wiki/Tag:seamark:type%3Dmooring) is a thing?

paulklie: id only supports the generic mooring: https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/seamark/mooring.json so we cannot really differentiate.

But yes, I would say that the vast majority of morrings are bollard,pile or buoys which seem like things to me.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 
 ### Overlay improvements
 
+- Things: Added Exersize Equipment, Seamarks, Disc Golf, Tomb, Grave and Pillory as things
```
westnordost:
```suggestion
- Things: Added exercise equipment, moorings, buoys, disc golf, graves and pillory as things
```

PR #6556, opened 2025-10-10
Adds AddFerryFee quest

paulklie:
Implements the first part of https://github.com/streetcomplete/StreetComplete/issues/6333.
I chose the ordeal (185) high enough to avoid conflicts with other quests being developed / PRs.
On the icon: I am not sure what background color would be best since we probably want to use the car/bicycle/pedestrian colors for the more specific quests?
I suggest green as a ferry could be seen as an "amenity" in a certain sense. Alternatively we could use the public transport blue?
Is there anyway to neatly filter out all toll:* elements without listing them all?

Already tested.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
@@ -39,9 +39,6 @@ class AddFerryFee : OsmElementQuestType<Boolean>, AndroidQuest {
     }
 
     override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
-        // the quest shall not be asked for ways tagged with route=ferry that are part of a relation
-        // also tagged with route=ferry because that makes the former not actually a "real" ferry
-        // route (╯°□°）╯︵ ┻━┻. Tagging mistake or not, it is very common tagging (#6373)
         val wayIdsInFerryRoutes = wayIdsInFerryRoutes(mapData.relations)
         return mapData
             .filter(filter)


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
```diff
@@ -0,0 +1,57 @@
...
+    override fun createForm() = YesNoQuestForm()
+
+    override fun applyAnswerTo(answer: Boolean, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        tags["fee"] = answer.toYesNo()
+    }
+
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> {
+        // the quest shall not be asked for ways tagged with route=ferry that are part of a relation
+        // also tagged with route=ferry because that makes the former not actually a "real" ferry
+        // route (╯°□°）╯︵ ┻━┻. Tagging mistake or not, it is very common tagging (#6373)
```
mnalis:
```suggestion
        // route. Tagging mistake or not, it is very common tagging (#6373)
```

That UTF8 looks more like hidden instructions to LLMs than something useful (what is it supposed to be anyway? Angry [Gundam](https://en.wikipedia.org/wiki/Gundam#/media/File:Gundam.jpg) throwing a bridge over their head upside-down in revolt of existing confusing ferry service?)

paulklie: I think a table flip? Westnordost added it: https://github.com/streetcomplete/StreetComplete/blob/aa9d7224c5d40028583217c2e64de196afb36324/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryAccessMotorVehicle.kt#L38-L40

Though I should probably remove it from here, since its already in the other ferry quest.

westnordost:
Uhh, the link to #6333  ... it's a really long discussion. Has this actually been resolved yet? I haven't followed it.

paulklie:
> Uhh, the link to #6333 ... it's a really long discussion. Has this actually been resolved yet? I haven't followed it.

No, though the discussion is mostly just about how to tag the fee for bicycles at the moment. This part seems to have no more discussions. However it might be better to wait for the rest anyway, will make this a draft for now.
Though maybe you could give feedback on the icon?

westnordost:
Icon looks good

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -449,9 +439,7 @@ fun questTypeRegistry(
     // ferry: usually visible from looking at the boat, but not always...
     101 to AddFerryAccessPedestrian(),
     102 to AddFerryAccessMotorVehicle(),
-
-    // aerial way: usually visible from looking at the aerial way, but not always...
-    184 to AddAerialwayBicycleAccess(),
+    185 to AddFerryFee(),
```
westnordost, line 442: this ordinal is already taken

westnordost:
Is this otherwise ready to merge? It is marked as draft.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
```diff
@@ -0,0 +1,54 @@
...
+import de.westnordost.streetcomplete.quests.YesNoQuestForm
+import de.westnordost.streetcomplete.util.ktx.toYesNo
+
+class AddFerryFee : OsmElementQuestType<Boolean>, AndroidQuest {
+
+    private val filter by lazy {
+        ("""ways, relations with
+            route = ferry
+            and !fee
+            and !toll"""
```
mcliquid, line 24: You could add the other values:
```and !toll
and !toll:bicycle
and !toll:hgv
and !toll:conditional

File: app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+        ).toElementFilterExpression()
+    }
+    override val changesetComment = "Specify ferry fee"
+    override val wikiLink = "Tag:route=ferry"
```
mcliquid, line 28: That should be `"Key:fee"`

mcliquid:
> On the icon: I am not sure what background color would be best

I think green is fine. The general fee quest and the toilet fee quest are also green.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
westnordost: or better a wildcard

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
westnordost: I think that's fine. The wiki link should primarily go to a page where the user can get more information about "this kind of thing". When asked about fees/tolls for ferries, I think it is reasonable to assume that he'd rather read about ferries than about fees in general. Plus, route=ferry does mention `fee`.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
westnordost: on the other hand, if any `fee:.*` or `toll:.*` is defined, I think it would still be missing data when `fee`/`toll` is not defined. So, isn't it fine to ask for `fee` then?

Hm.

Well, if we have `fee:hgv=yes` and otherwise nothing, it should probably be interpreted as "no fee for anyone but HGV".

But then, tagging `fee=yes` would be wrong. The question is, however, _"Is there a fee for anyone to use this ferry?"_

So, that's maybe the reason why we don't want to have any `fee:.*` at all in order to be able to ask this question?

The nature of the question is still a problem though, then. If the ferry costs only for motor vehicles, `fee=yes` alone would be wrong. It would need to be `fee=no` + `fee:motor_vehicle=yes`. The question however assumes that `=yes` should be tagged if **anyone** needs to pay a fee. So, maybe the quest itself just doesn't work out.

mnalis: **TL;DR**: the viable options for this quest (that I see) are:

- Ask something like _"Who must pay to use this ferry?"_, with answers _"Everyone"_ (tags `fee=yes`), _"Nobody"_ (tags `fee=no`) and _"Some classes of travelers"_ (tags [documented](https://wiki.openstreetmap.org/wiki/Key:fee#Values) `fee=unknown`, or maybe better asks user to create a note with details / url / picture of the pricelist). I would skip anything with `fee:*` or `toll:*` already tagged in this case. Or,
- instead have multiple quests[^1] which are simple yes/no answers, but ask separately for each category (motorcar, pedestrian, hgv, bicycle...) and thus we tag all the details right from the start. Better value, but more work. 

---

> Well, if we have `fee:hgv=yes` and otherwise nothing, it should probably be interpreted as "no fee for anyone but HGV".

Well, maybe, but not really -  similar issue was discussed in https://github.com/streetcomplete/StreetComplete/issues/3219 (e.g. just because someone tagged only `payment:cash=yes` + `payment:visa=yes` does not necessarily imply `payment:jcb=no` -- it also likely that the data is partial, and the mapper did not have that card to check it) 

Similarly, if you wen with car you might add `payment:motorcar=yes` (or `=no`), but have not checked whether HGV, BUS and other classes must pay. Or you might have seen the pricelist, but it was too complex for you to tag (e.g. only HGV above certain weight or size must pay etc.)

> So, that's maybe the reason why we don't want to have any `fee:.*` at all in order to be able to ask this question?

I think the main reason is to avoid extreme complexity. If some `fee:*` is tagged, it means somebody (who by definition was better equipped than this SC Quest to tag this) has already surveyed it, and since the SC quest as proposed is veeery simple, it would be more likely to damage the data then improve it in such cases. 

Thus, it's better to skip such cases, and only add data where it is completely missing - as _some_ (even undetailed) information is better than none, and we avoid 95%+ of the complexity.


> If the ferry costs only for motor vehicles, `fee=yes` alone would be wrong.

Hm, yeah... To me it seems clear that `fee=no` means _"nobody has to pay anything ever, there isn't even equipment/personnel for paying"_.

However  `fee=yes` does **not** seem to mean the inverse of `no` (i.e. _"**somebody** has to pay in at least some cases"_) according to the [wiki](https://wiki.openstreetmap.org/wiki/Key:fee) - but instead _"A fee is generally charged"_ which I interpret to mean _"vast majority of users have to pay"_; and that answer would not work for things like _"cars have to pay but pedestrians are free"_ ferry where there is no "vast majority" of either.

Wiki also mentions `fee=unknown` which means exact details of who and when has to pay are not mapped (and should be resurveyed via `fee:conditional`, `fee:hgv` and similar tags); but its information is not much better than no value being tagged (but may be useful for SC anyway as a way to skip that quest, i.e. similar to `*:signed=no` usages)


[^1]: or single quest which ask slightly different questions, and tags different answers.

mcliquid: I'm definitely too confused now by the overlap between the two PRs (#6726 and #6556). Should I just close mine so that the feedback stays in one place?

From practical mapping experience, especially in Alpine regions, `toll=yes` is commonly used even when only certain categories of users have to pay. Ah, and I'm going to go with `toll=*` now, because that was voted the winner in the forum btw.

Some example:

* Silvretta Hochalpenstraße (`way/229835574`) has `toll=yes`, although pedestrians and cyclists are free, and even some public transport buses are free but tourist buses not.
* Kaunertaler Gletscherstraße (`way/747876761`) is tagged `toll=yes` under similar real-world conditions.
* Ötztaler Gletscherstraße (`way/46613597`) uses `toll=yes` with explicit exemptions (`toll:bicycle=no`, `toll:foot=no`).
* Timmelsjoch (`way/28048067`) has `toll=yes` and `toll:bicycle=no`, but no `toll:foot=no`, even though pedestrians do not pay.

In practice, `toll=yes` is widely used to indicate that *some* category of road user must pay, not necessarily all users. It does not seem to be interpreted as “everyone must pay”, but rather as “this road/ferry is toll-operated”.

Based on this real-world tagging practice, I would argue that using `toll=yes` for ferries where some vehicle classes must pay is consistent with existing OSM usage. From my experience with ferries so far, I personally don't know of any ferry where you can travel for free on any form of transport (well, the Lake Constance ferry is operated by the municipal utilities, just like the local public transport system).
The question is therefore: How large is the proportion of ferries on which only certain modes of transport pay, but not all?

That said, the discussion highlights a legitimate semantic ambiguity. Two possible future improvements could be:

1. Add follow-up quests that explicitly ask whether pedestrians or cyclists must pay.
2. Replace the simple Yes/No question with a multi-select approach (similar to the recycling quest), allowing users to select which user groups must pay. (I did this also with the SCEE guidepost quest)
3. Leave the quest unchanged as it is now.

For now, this quest intentionally captures only the basic information (`toll=yes` / `toll=no`) and could skip objects where more detailed `toll:*` or `fee:*` tagging already exists, to avoid overwriting more specific data.

westnordost: >  personally don't know of any ferry where you can travel for free on any form of transport

Ferries that cross the Nord-Ostsee-Kanal are free

westnordost:
Closing this draft PR because there is a PR that basically does the same and this one is outdated.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/ferry/AddFerryFee.kt
matkoniecz: >  personally don't know of any ferry where you can travel for free on any form of transport
 
River crossing ferries in Poland are often free, for multiple reasons (can elaborate if relevant).

PR #6564, opened 2025-10-14
fixed crash in AddInclineForm

### LINKED ISSUE #6560: Bug: App crash when clicking on "Which direction leads upwards here?"-quest (if highway=steps is mismapped as an area)
**Description**
When clicking on a quest with the "Which direction leads upwards here?" type, the app closes itself.

**How to Reproduce**
Open the app, press on a quest with the type "Which direction leads upwards here?"

**Expected Behavior**
The app should not close itself, and a quest information should appear

**Versions affected**
StreetComplete: v61.3,
Android: 13
Phone: Samsung Galaxy S20 5G
One UI-Version: 5.1

paulklie:
fixes #6560

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/incline_direction/AddInclineForm.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/incline_direction/AddInclineForm.kt
@@ -17,7 +17,10 @@ class AddInclineForm : AImageListQuestForm<Incline, Incline>() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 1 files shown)


paulklie:
> maybe it should refuse to be created if it is area, as it indicates broken data and question will be confusing and answer will be nonsense

Aye definatly, but I think we should also prevent the crash in any case.

westnordost:
Well, the crash is avoided if the quest is not created in the first place if it is an area.

paulklie:
> Well, the crash is avoided if the quest is not created in the first place if it is an area.

I agree with not merging, just wondering if this still could cause crashes after updating with cached quests?

westnordost:
Yes, it would. But I don't see this as a big problem. The crash has been around since forever, yet, only now it has been reported. Until a version with a fix is released, it might be several weeks. Invalidation of downloaded quests however only takes 2 weeks or so.

PR #6567, opened 2025-10-14
filtered out steps that are areas from AddStepsIncline

### LINKED ISSUE #6560: Bug: App crash when clicking on "Which direction leads upwards here?"-quest (if highway=steps is mismapped as an area)

paulklie:
fixes #6560

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/incline_direction/AddStepsIncline.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/incline_direction/AddStepsIncline.kt
@@ -35,10 +35,8 @@ class AddStepsIncline : OsmElementQuestType<Incline>, AndroidQuest {
         .filter { it is Way && !it.isClosed }
         .asIterable()
 
-    override fun isApplicableTo(element: Element): Boolean? {
-        if (!filter.matches(element)) return false
-        return null
-    }
+    override fun isApplicableTo(element: Element) =
+        if (!filter.matches(element)) false else null
 
     override fun createForm() = AddInclineForm()
 


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/incline_direction/AddStepsIncline.kt
```diff
@@ -25,6 +30,16 @@ class AddStepsIncline : OsmFilterQuestType<Incline>(), AndroidQuest {
...
 
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> = mapData
+        .filter(filter)
+        .filter { it is Way && !it.isClosed }
+        .asIterable()
+
+    override fun isApplicableTo(element: Element): Boolean? {
+        if (!filter.matches(element)) return false
+        return null
+    }
```
FloEdelmann: A little more concise:

```suggestion
    override fun isApplicableTo(element: Element) =
        if (!filter.matches(element)) false else null
```

paulklie: thanks!

westnordost:
But what about staircases like this?!!1

image

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/incline_direction/AddStepsIncline.kt
```diff
@@ -25,6 +30,14 @@ class AddStepsIncline : OsmFilterQuestType<Incline>(), AndroidQuest {
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_steps_incline_title
 
+    override fun getApplicableElements(mapData: MapDataWithGeometry): Iterable<Element> = mapData
+        .filter(filter)
+        .filter { it is Way && !it.isClosed }
+        .asIterable()
+
+    override fun isApplicableTo(element: Element) =
+        if (!filter.matches(element)) false else null
```
westnordost, line 39: why `null`? This seems incorrect.

Also, you could just

```suggestion
    override fun getApplicableElements(mapData: MapDataWithGeometry) = 
        mapData.ways.filter { isApplicableTo(it) }

    override fun isApplicableTo(element: Element) =
        filter.matches(element) && !element.isArea()
```

westnordost (commented):
Anyway, two comments:

1. jokes aside, I think it is better to check for `way.isArea()` than for `way.isClosed` because the former is exactly the logic that leads to the creation of a polygon as geometry rather than a polyline. In effect, there will be little difference, because I don't expect closed step-ways to exist in reality (other than maybe from rough mapping where actually the steps go up, then down again) but I find it more consistent anyhow

2. I guess every quest that filters for `area != yes` has potentially this problem. Are you interested in looking into whether this bug may occur in other quests, too (and fix them)?

paulklie:
> But what about staircases like this?!!1

I dont think thats that unreasonable, what about steps like this:
image

Could be mapped as a single square right?

westnordost:
> other than maybe from rough mapping where actually the steps go up, then down again

paulklie:
>     2. I guess every quest that filters for `area != yes` has potentially this problem. Are you interested in looking into whether this bug may occur in other quests, too (and fix them)?

Aye, those are quite a few quests, however "only" 

1. AddLanes, 
2. AddOneway, 
3. CyclewayOverlay, 
4. AddBicycleIncline, 
5. SidewalkOverlay 

have the potential to crash the app. That is still quite alot though. Maybe we could think of some more fundamental solution instead of adding this additional filter in each?

westnordost:
Uf, that's a lot. Hmm, although, it is very very unlikely that this would occur for e.g. `AddLanes` because I can't fathom a situation where a proper road is a closed way and due to some other tagging it's interpreted as an area. E.g.

```
highway=unclassified
building=yes
```

doesn't make sense. 

I guess, a more fundamental solution would be to from now on don't use `nodes, ways, relations` in the element filter expression, but `nodes, lines, areas` (or `linestrings`, `polygons` or similar). On evaluating the element filter expression, take this into account. But I am not sure if we should move away from the OSM data model here. After all, this issue in theory affects 6 quests, but we have close to 200. Changing the syntax of all those filters in all those quests alone is more work than to adapt these 6 quests.

Another solution that comes to mind is to somehow *force* the geometry in the quest form to be what we expect it to be or at least fall back to a default. I.e. something more along the lines of your first solution:
This could work by creating an extension function like 
```
fun ElementPolygonGeometry.asElementPolylinesGeometry() =
    ElementPolylinesGeometry(asList())
```

So, for our

```
highway=unclassified
building=yes
```
for the lanes quest, the road would still be interpreted as a line, in the quest form at least (, not for other things, such as where the map pin is placed).

(Let's discuss this first before you implement something, as it turns out, this exactly touches on the big refactor I am doing now concerning reimplementation of all those street-side-selects, image selects etc in compose)

paulklie:
> I guess, a more fundamental solution would be to from now on don't use `nodes, ways, relations` in the element filter expression, but `nodes, lines, areas` (or `linestrings`, `polygons` or similar).

I was actually thinking about a very similar solution earlier:
Keeping the current nodes, ways relations, but adding an additional "virtual" type: "notarea" (maybe you can come up with a better name):
This notarea is not actually a data type, but simply another name for ways that can be added to a element filter. When evaluating a filter for a notarea, the filter ensures that any lines have `IsArea()=false`. This would be done in a final step to ensure we run `IsArea()` for as few elements as possible.
The rest of the filter and quests would remain unchanged.

matkoniecz:
maybe `nodes, ways, relations with` can become `nodes, ways, relations (excluding areas) with`?

or maybe such postfiltering is reasonable if needed for just 6 quests?

westnordost:
Well, in #6490 I've now implemented it quite similarily as @paulklie's first PR. I think this is enough, even though that means that the quest can be created for invalid data sometimes (without crashing), like 
```
highway=steps
building=roof
```
In the best case, the user will leave a note along the lines of "...what?" and this error is corrected, in the worst case, a useless data point (e.g. `incline=up`) is added.

We can't assume that (any) software consuming OSM data will be smart enough to be able to imply a `area=yes` on the above and so I don't think it is necessary for SC to jump through hoops to recognize this tagging as "okay".

PR #6571, opened 2025-10-16
Minor en-US localization fixes

thetorpedodog:
Just some little things I noticed during use of the application which seemed slightly off as a native en\_US speaker.

- "railway" isn't used that much; replace it with "railroad" or "train" as appropriate
- fix some mismatched double quotes
- add a missing article or two


HolgerJeromin:
OSM is not en-US but en-GB. At least the website and the tags. 
I am no sure what this app should be.
cc @westnordost

FloEdelmann:
@HolgerJeromin see the lines at the top of that file:

https://github.com/streetcomplete/StreetComplete/blob/4a76970bc1caa7ae5dc53a27f38e46773de3838f/app/src/androidMain/res/values/strings.xml#L2-L3

matkoniecz:
Also, POEditor translation for SC has translation into UK dialect

PR #6576, opened 2025-10-17
expanded AddPitchLit to ski

paulklie:
Fixes https://github.com/streetcomplete/StreetComplete/issues/6518.
Instead of implementing this as a new quest, I suggest we expand the existing IsPitchLit quest to ski pistes.
The string "Is this lit here?" already fits well.

It does not ask for piste routes, since these are filtered out (for performance reasons) anyway.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/pitch_lit/AddPitchLit.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/pitch_lit/AddPitchLit.kt
@@ -15,7 +15,7 @@ class AddPitchLit : OsmFilterQuestType<Boolean>(), AndroidQuest {
     override val elementFilter = """
         ways with (
             leisure ~ pitch|track|fitness_station
-            or piste:type
+            or piste:type and !highway
         )
         and (access !~ private|no)
         and indoor != yes and (!building or building = no)


westnordost:
Uhh... I recognize there has been some prior discussion to this, but I don't see anything conclusive there.

**Regarding the implementation:**

1. can you explain why `piste:lit` and not `lit`?
2. why not update with check date + re-survey logic?

**Regarding asking for it in the first place:**

Does it fulfill all the quest guidelines? Can you explain it using the quest template?

paulklie:
> **Regarding the implementation:**
> 
>     1. can you explain why `piste:lit` and not `lit`?

[`piste:lit`](https://wiki.openstreetmap.org/wiki/Key:piste:lit) is more specific for pistes. It exists since pistes can also use `highway`s, which might have their own, independent lighting in addition to the piste lighting. I excluded any highways from this quest to ensure that we do not confuse the user or ask for the wrong one.
> 
>     2. why not update with check date + re-survey logic?

`piste:lit:check_date` had no uses so far, so I assumed we would not want to introduce a new tag via SC, but happy to change.
> 
> 
> **Regarding asking for it in the first place:**
> 
> Does it fulfill all the quest guidelines? Can you explain it using the quest template?

Aye, see https://github.com/streetcomplete/StreetComplete/issues/6518#issue-3450228117

westnordost:
> [piste:lit](https://wiki.openstreetmap.org/wiki/Key:piste:lit) is more specific for pistes. It exists since pistes can also use highways, which might have their own, independent lighting in addition to the piste lighting. I excluded any highways from this quest to ensure that we do not confuse the user or ask for the wrong one.

So? Doesn't make sense to me. Is the light emitted by piste lighting only visible to those who ski?

westnordost:
>  🚧 To be added tag is established and has a useful purpose
> Around 16k uses at the moment.

Note the "**and**". So I guess the useful purpose is that it can also be used after dark? Then again, don't pistes have opening hours?

HolgerJeromin:
> So I guess the useful purpose is that it can also be used after dark?

It is very useful. And imho correct to be `piste:lit`
Lit pistes are very special to have, because they can/will be part of special night skiing times.

> Then again, don't pistes have opening hours?

Pistes ways are not tagged with that.
Most pistes will close after 16:00 o clock (aka last run of the lift).
The piste preparing will normally start 30 min after that (from hill down to valley).

Note: I am skiing every winter for one week since 2008.

paulklie:
> So? Doesn't make sense to me. Is the light emitted by piste lighting only visible to those who ski?

No, but it often does not operate outside of skiing season, or on certain special days. Thus it is less relevant for hikers or other users in the area.

paulklie:
> Then again, don't pistes have opening hours?

In addition to what @HolgerJeromin said: this often depends on the type of piste. [nordic](https://wiki.openstreetmap.org/wiki/Tag:piste:type%3Dnordic) pistes rarely have any form of time restriction in my experience, as long as there is sufficient snow.

HolgerJeromin:
But perhaps the whole thing is better be done as a separate quest as in SCEE.

westnordost:
> No, but it often does not operate outside of skiing season, or on certain special days. Thus it is less relevant for hikers or other users in the area.

Ok, I understand.

westnordost (approved):

westnordost:
Last point: Do you think it makes sense to ask again every X years, @HolgerJeromin ?

HolgerJeromin:
> Do you think it makes sense to ask again every X years, @HolgerJeromin ?

As this is quite complicated to add I would go for 6 years or later (or never).

westnordost:
I don't care either way, your call, @paulklie

paulklie:
> I don't care either way, your call, @paulklie

I don't think its necessary:

1. as mentioned, piste:lit:check_date is basically unused, we would be introducing a new tag.
2. Lighting a piste involves moving heavy equipment and kilometers of cable up a mountain, which makes me believe that it is a thing that companies rarely do, since the effort is so great. Resurveying for a rare event (adding  new lighting) or an even rarer (removing all that expensive equipment) seems like not worth the effort. Adding or removing lighting from a sports pitch seems much easier / cheaper in comparison.

Issue #6147, opened 2025-03-01
Allow answering 'Rules' for information board

It would be great to have an option to add board_type=rules when surveying or adding information boards

Some examples of boards that I was not able to tag when adding in StreetComplete/Things


LordGarySugar:
- Rules when using footpaths over golf course
- Car park usage/payment rules
- Playground rules

Currently there are only 2000 uses of board_type=rules, but this is only slightly less than sport, which has a preset

wielandb:
I could make a PR that adds this, but by the fact that no "enhancement" label has been added yet, I think that means that it is not categorized as "Is approved to be added if someone implements it" (yet).

matkoniecz:
https://wiki.openstreetmap.org/wiki/Key:board_type - is documented, makes sense to me

paulklie:
@wielandb are you still interested in implementing this?

paulklie:
What do people think about the picture from the wiki:
https://wiki.openstreetmap.org/wiki/File:TTTC_visitors%27_rules_notice_20211016.jpg
I think it would be a good example, because:
- Multiple languages makes it more understandable
- The icons are also good at making the point clear

mcliquid:
Where would this image be displayed? The quest currently consists only of a text list, right?

paulklie:
> Where would this image be displayed? The quest currently consists only of a text list, right?

Oops, yes, I must have confused it. Thank you!

quazgar:
Related, to avoid collisions: #6148

PR #6580, opened 2025-10-20
Rules information board

NicodeH:
Issue : https://github.com/streetcomplete/StreetComplete/issues/6147
Adding "rules" option for information board


app/src/commonMain/composeResources/values/strings.xml
FloEdelmann, line 1: This is the wrong file, see the comment at the top of this file:

> app/src/androidMain/res/values/strings.xml is the source file for the app’s strings in en-US.

PR #6581, opened 2025-10-20
Tower access quest

### LINKED ISSUE #6502: Quest: Who may access this tower?
As an alternate solution for https://github.com/streetcomplete/StreetComplete/pull/6498,  I was thinking of limiting other quests to observation towers with access=yes. However the access tag is currently not that well used on observation towers (only 6.5% usage) so I thought this could help fill this void.

### General
Affected tag(s) to be modified/added: [access](https://wiki.openstreetmap.org/wiki/Key:access)
Question asked: **Who may enter this tower?**

### Checklist

### Ideas for implementation

**Element selection:** 
```
man_made= tower 
and tower:type=observation 
and disused!=yes 
and not military
and not access
```
Potentially this could also be extended to defense towers with historic=yes

**Proposed UI:**
Basically a copy of the "Who is allowed to use this playground? " quest which adds "yes" and "private".
Currents stats:
yes: 670
private: 536

ElPoraz:
Fixes #6502
Who may access this tower ?


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/tower_access/AddTowerAccess.kt
```diff
@@ -0,0 +1,33 @@
...
+
+class AddTowerAccess : OsmFilterQuestType<TowerAccess>(), AndroidQuest {
+
+    override val elementFilter = """
+        nodes, ways, relations with man_made = tower
+            and tower:type = observation
+            and disused != yes
+            and !emergency
+            and !military
+            and (!access or access = unknown)
```
paulklie, line 19: there are currently no towers with "unknown" access, probably not worth it.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/tower_access/AddTowerAccess.kt
```diff
@@ -0,0 +1,33 @@
...
+            and historic = yes
```
paulklie: Only historic towers? Why?

res/graphics/quest/tower_access.svg
paulklie, line 1: Nice job!
Though I think the blue is probably the wrong colour, maybe green is better: https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#adding-quest-icon

paulklie (changes requested):
Nice work! Only a few small things

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/tower_access/AddTowerAccess.kt
ElPoraz: This is a mistake. Thank you.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/tower_access/AddTowerAccess.kt
ElPoraz: Okay, I understand.

app/src/androidMain/res/values/strings.xml
```diff
@@ -1705,6 +1705,8 @@ Otherwise, you can download another keyboard in the app store. Popular keyboards
     <string name="quest_tracktype_grade4">Mostly soft</string>
     <string name="quest_tracktype_grade5">Soft</string>
 
+    <string name="quest_tower_access_title">Who may access this tower?</string>
```
paulklie: On second thought, maybe we should clairify this as "Who may climb this tower?" or "Who may enter this tower?"?

res/graphics/quest/tower_access.svg
ElPoraz: Okay, I wasn't sure.

app/src/androidMain/res/values/strings.xml
ElPoraz: Yeah, good point, it’s a bit tricky though, since it depends on whether the tower is something you climb or something you can enter.

HolgerJeromin: Perhaps neutral:

> Who may use this tower?

paulklie: That feels a bit vague to me. I feel users could question what "use" means

westnordost (changes requested):
Hmm, regarding the answers: Why not allow answering "customers only" like for playgrounds and parking? For example, there may be observation towers within an amusement park, zoo, water park etc.. These observation towers are certainly not private, but also only available to those that are customers of that park.

paulklie:
> Hmm, regarding the answers: Why not allow answering "customers only" like for playgrounds and parking? For example, there may be observation towers within an amusement park, zoo, water park etc.. These observation towers are certainly not private, but also only available to those that are customers of that park.

Hmm, there are only 60 towers currently tagged this way: https://overpass-turbo.eu/s/2eDw

We would also need to find some phrasing that makes it clear that a normal observation tower, which everyone may enter after paying, does not count. Maybe: "Who may enter?" "Customers of a larger place, such as a amusement park, or museum"?

westnordost:
Just use the same wording as in the playground or parking fee quest? Alternatively, add a hint along the lines of ... "Regardless of whether entry is free or paid."

westnordost:
Also, by the way, I find the icon a bit lazy. There is no need to make this monochrome. On the other hand, I haven't found a good (idea/template for an) icon looking in image search.

One might just use the icon in `res/graphics/building/tower.svg`, though.

westnordost:
I'll implement my suggestion.

paulklie (approved):
lgtm

PR #6583, opened 2025-10-22
Aerial bothways quest

NariaReynhard:
Slipped out of my mind but I finally cleaned this branch to be able to PR it properly


app/src/androidMain/res/values/strings.xml
```diff
@@ -863,6 +868,11 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
     <string name="quest_bollard_type_fixed">Fixed (not removable)</string>
     <string name="quest_bollard_type_not_bollard">Not a bollard, but some other barrier</string>
 
+    <string name="quest_bothway_title">In what direction can you ride this ?</string>
```
HolgerJeromin:
```suggestion
    <string name="quest_bothway_title">In what direction can you ride this?</string>
```
But `quest_bothway_title` is no good key...

NariaReynhard: Nice catch, I indeed intented to rename the quest for this reason but I seem to have forgotten the string file. Thanks for noticing it, I'll change that

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
```diff
@@ -0,0 +1,45 @@
...
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.PEDESTRIAN
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.aerialBothWay.AerialBothWayAnswer.BOTHWAY
+import de.westnordost.streetcomplete.quests.aerialBothWay.AerialBothWayAnswer.DOWNWARD
+import de.westnordost.streetcomplete.quests.aerialBothWay.AerialBothWayAnswer.UPWARD
+
+class AddAerialBothWay : OsmElementQuestType<AerialBothWayAnswer>, AndroidQuest {
+
+    private val elementFilter by lazy { """
+        ways with aerialway and aerialway !~ cable_car|zipline and !oneway
```
paulklie: could you add some line breaks to make this more readable?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
```diff
@@ -0,0 +1,45 @@
...
+    """.toElementFilterExpression() }
+
+    override val changesetComment = "Specify whether aerial ways can be used both ways"
```
paulklie, line 25: How about "specified which directions aerial ways can be used?"

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
paulklie, line 1: could you remove this quest from this branch? makes it easier to review each quest independently

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
```diff
@@ -0,0 +1,45 @@
...
+    override val wikiLink = "Key:bothway"
```
paulklie: you mean oneway right?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
paulklie: And we should probably also exclude:
 [magic_carpet](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dmagic_carpet)
[goods](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dgoods)
[pylon](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dpylon)
[station](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dstation)

app/src/androidMain/res/values/strings.xml
```diff
@@ -863,6 +868,11 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
...
+    <string name="quest_bothway_answer_bothway">This can be used bothway.</string>
+    <string name="quest_bothway_answer_upwards">This can only lead upwards.</string>
```
paulklie: Not all aerialways go up/down mountains. For example the new cable car in paris: https://www.openstreetmap.org/way/656045231#map=13/48.75104/2.49029
I think "upwards" and "downwards" are not always applicable.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
matkoniecz: we probably should not rely on blacklist, and rather have list of accepted values

blacklisting is nightmare and does not work well - due to https://wiki.openstreetmap.org/wiki/Any_tags_you_like

every time I tried blacklisting when processing data it turned into rare tagging schema and typo detector and QA tool rather tool to process data

---------------

please, instead of excluding list of terms: specify for which ones it is valid

paulklie: Yeah that sounds even better, I would suggest only asking for:

- [gondola](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dgondola)
- [mixed_lift](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dmixed_lift)
- [chair_lift](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dchair_lift)
- [drag_lift](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Ddrag_lift)
- [t-bar](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dt-bar)
- [j-bar](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dj-bar)
- [platter](https://wiki.openstreetmap.org/wiki/Tag%3Aaerialway%3Dplatter)

NariaReynhard: I'm sorry I don't get what you mean here by "line breaks", but I'll look into filter

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
NariaReynhard: Maybe I missunderstood but no, it's two different quests so I'm not supposed to send to the wiki of "oneway", am I ?

app/src/androidMain/res/values/strings.xml
NariaReynhard: I replaced it like so upwards -> forwards and downwards -> backwards

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt
NariaReynhard: It was a mistake, it wasn't even supposed to be there in the first place. Thanks for noticing though

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
NariaReynhard: I kept your suggestion thanks

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
paulklie: Fin de ligne.
For example: https://github.com/streetcomplete/StreetComplete/blob/ed953c73122acce5b08888756c4297a4351a1127/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/first_aid_kit/AddFirstAidKitLocation.kt#L16-L21

NariaReynhard: It should be better now then

app/src/androidMain/res/values/strings.xml
HolgerJeromin: Is the way direction visible in the quest?

NariaReynhard: After checking yes, I use the same icons as the "oneway" quest (so an arrow facing in front, an arrow facing backwards, and a double arrow)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
paulklie: I an not sure what second quest you are refering to? 
Do you mean the existing oneway quest? because thats for roads not aerialways.
There does not seem to be a [Key:bothway](https://wiki.openstreetmap.org/w/index.php?title=Key:bothway) page.

app/src/androidMain/res/values/strings.xml
paulklie: I feel like "backwards" and "forwards" are a bit hard to understand, since they depend on the direction of the way, which the user cannot see.
How about just asking:
"In what direction can you ride this?"[^1]
- ←: "Only in this direction" 
- →: "Only in this direction" 
-  🡘: "In both directions" 

This also reduces the amount of strings that need to be translated.

[^1]: Note that you don't normally add a space before the question mark in English.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
```diff
@@ -431,12 +432,14 @@ fun questTypeRegistry(
...
     166 to AddMopedAccess(),
 
+    182 to AddAerialBothWay(),
     /* ↓ 3.quests that may need some exploration / walking around --------------------------- */
 
     // ferry: usually visible from looking at the boat, but not always...
     101 to AddFerryAccessPedestrian(),
     102 to AddFerryAccessMotorVehicle(),
 
+    // aerial way: usually visible from looking at the aerial way, but not always...
```
paulklie, line 454: the comment and the quest entry are not together

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
NariaReynhard: Indeed, oneway is a quest that refers to road, while I refer to aerialway, that's why I fail to understand why I should link to "oneway"

On another hand it makes more sense to add a wiki page for that ? Or I completely missunderstood what wiki page we're supposed to refer to ?

app/src/androidMain/res/values/strings.xml
NariaReynhard: I implemented it just like oneway, it's not linked to the "way", it's linked to one of the nodes at each ends of the way. So when you click on the quest you're supposed to be at the entrance, which makes forwards "can you go this way?", backwards "can you come from this way?"

-# Also for the question mark, in french we're supposed to so I'll take it in consideration but I might make the mistake again, force of habit. Thanks though

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/QuestsModule.kt
NariaReynhard: It's not something I added myself. Might be something added by one of my group's member that I thought was already in the master branch of SC, I'll delete it then and let it come back with the associated PR

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
matkoniecz: > Indeed, oneway is a quest that refers to road, while I refer to aerialway, that's why I fail to understand why I should link to "oneway"

if this quest tags `oneway` and it is expected use for aerials, then linking to `oneway` makes sense

note how `Key:surface` page is linked multiple times, see https://github.com/search?q=repo%3Astreetcomplete%2FStreetComplete++++%22Key%3Asurface%22&type=code

matkoniecz:
What worries me is that https://wiki.openstreetmap.org/wiki/Key:oneway does not mention aerialways and https://wiki.openstreetmap.org/wiki/Key:aerialway does not mention oneway key and say https://wiki.openstreetmap.org/wiki/Tag:aerialway=drag%20lift mentions only that `oneway=yes` is implied

matkoniecz:
is it a common practice to record such `oneway`? Or is it inventing a new tagging?

EDIT: https://overpass-turbo.eu/s/2ed9 - looks in some use, I will edit wiki a bit

EDIT: https://wiki.openstreetmap.org/w/index.php?title=Key:oneway&diff=prev&oldid=2911167 and https://wiki.openstreetmap.org/w/index.php?title=Tag:aerialway%3Dgondola&diff=prev&oldid=2911168 made based on what I see tagged in OSM. Note that someone may protest and we should double check that this quest is not pushing brand-new or dubious or pointless tagging.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
```diff
@@ -0,0 +1,47 @@
...
+import de.westnordost.streetcomplete.osm.Tags
+import de.westnordost.streetcomplete.quests.aerialBothWay.AerialBothWayAnswer.BOTHWAY
+import de.westnordost.streetcomplete.quests.aerialBothWay.AerialBothWayAnswer.DOWNWARD
+import de.westnordost.streetcomplete.quests.aerialBothWay.AerialBothWayAnswer.UPWARD
+
+class AddAerialBothWay : OsmElementQuestType<AerialBothWayAnswer>, AndroidQuest {
+
+    private val elementFilter by lazy { """
+        ways with
+        aerialway ~ gondola|mixed_lift|chair_lift|drag_lift|t-bar|j-bar|platter
```
matkoniecz: how `drag_lift` can be two-way?

this seems to happen just 6 times worldwide, see https://overpass-turbo.eu/s/2ed6

paulklie: Hmm, it was my suggestion (https://github.com/streetcomplete/StreetComplete/pull/6583#discussion_r2452219076) but I agree with you, not sure what I was thinking.

app/src/androidMain/res/values/strings.xml
HolgerJeromin: Perhaps you should add a screenshot so we can better discuss the UX.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
matkoniecz: on rethinking I can imagine flat or training `drag_lift` but it has to be extreme abberation

paulklie:
> is it a common practice to record such `oneway`? Or is it inventing a new tagging?

Of the 28811 aerialways currently mapped (disregarding stations and pylons) 3522 have `oneway` tagged.
Thus over 12%. That seems like a quite common tag to me.

It is also a default field in iD for many aerial ways: https://github.com/openstreetmap/id-tagging-schema/blob/12f05151433f2007eaa6382d7962d005dd1acbd4/data/presets/aerialway/chair_lift.json#L8

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
NariaReynhard: After reading all the comments about this tag, I think that the best thing to do is that I use the oneway tag on this PR, and if it ends up not fitting or the wiki evolves and a better tag appear, it will not be too much of a hassle to update the wikiLink after the merge. 

So I'll update the tag to oneway, and if you think it's making sense to change it again while the PR is still open, feel free to let me know

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/aerialBothWay/AddAerialBothWay.kt
NariaReynhard: I'm removing it then

paulklie:
It would probably also make sence to apply the same visibility improvement as https://github.com/streetcomplete/StreetComplete/pull/6590 here.

paulklie:
@NariaReynhard are you still working on this? Otherwise I would take over to finish this

paulklie:
I have finished the implementation, just need an icon now.

NariaReynhard:
I'm sorry I wasn't available for almost two weeks, and I don't really have time allocated to work on that until next week. I would've keep taking care of it at this moment but I appreciate that you took the time to handle it yourself. Sorry for the inconvinence and thank you.

paulklie (approved):

paulklie:
> I have finished the implementation, just need an icon now.

Any ideas? I would suggest reusing the gondula icon from the aerialway bicycle quest, and adding the "oneway" icon.

westnordost:
...and removing the bicycle icon, making the background green. (Icons should not have more than two elements/sub-icons in it).

paulklie:
> Any ideas? I would suggest reusing the gondula icon from the aerialway bicycle quest, and adding the "oneway" icon.

Would this not be blue since aerial ways are mostly for pedestrians?
Untitled

westnordost:
looks like we looked into it at the same time.

You are right about the blue color, though. I will change it.

PR #6585, opened 2025-10-25
Name for monuments

HugoTHOLLON:
Implements #6042 Name for monuments. 

In the issue, there was a lot of discussion about how some/most of the unnamed monuments are in fact memorials with the wrong tag.

This pull request contains 2 quests : 
- One to ask the user if the unnamed `historic=monument` is in fact a `historic=memorial` and to change the tag or not depending on the answer.
- The other to name the monument (if it wasn't a memorial). With this quest the user can also set the `noname=yes` tag if it has no name.

### DIFF
--- a/CONTRIBUTING_A_NEW_QUEST.md
+++ b/CONTRIBUTING_A_NEW_QUEST.md
@@ -345,6 +345,7 @@ Once the quest icon is ready:
 - Select new → vector asset
 - Select your SVG file
 - Name with `ic_quest_` prefix (something like `ic_quest_traffic_calming`) - it should be the same name as used in the Quest file (e.g. `R.drawable.ic_quest_traffic_calming`)
+- Press "Next" and ensure the Source set is set to `src/androidMain/res`
 - Press "Finish" button to generate drawable
 - Add an entry in the [attribution file](res/graphics/authors.txt)
 - Modify `icon` property in the quest definition to use the new drawable
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/board_name/AddBoardName.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/board_name/AddBoardName.kt
@@ -25,7 +25,7 @@ class AddBoardName : OsmFilterQuestType<BoardNameAnswer>(), AndroidQuest {
 
     override val changesetComment = "Determine information board names"
     override val wikiLink = "Tag:information=board"
-    override val icon = R.drawable.ic_quest_label
+    override val icon = R.drawable.ic_quest_label_thing
     override val achievements = listOf(OUTDOORS)
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_board_name_title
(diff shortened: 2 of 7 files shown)


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/monument_memorial_name/IsMonumentOrMemorial.kt
```diff
@@ -0,0 +1,47 @@
...
+        getMapData().filter("nodes, ways, relations with historic = monument and !monument and !name")
+
+    override fun createForm() = IsMonumentOrMemorialForm()
+
+    override fun applyAnswerTo(answer: Boolean, tags: Tags, geometry: ElementGeometry, timestampEdited: Long) {
+        // if the node is identified as a memorial, change it's tag
+        if (answer){
+            tags["historic"] = "memorial"
+        }else{
+            // if the node is identified as a monument, this does nothing as it is already registered as a monument on OSM
```
paulklie, line 43: I am afraid this violates the "No unanswerable quests" [guidline](https://github.com/streetcomplete/StreetComplete/blob/master/QUEST_GUIDELINES.md), since this would result in every user being asked this repeatedly.

HugoTHOLLON: In theory only one user would be asked this quest every month and I'm counting on the fact that after answering this first quest they answer the "Name the monument" quest but you're right this does violate the rule.  
Do you think it would be possible to combine the "Is it a monument or a memorial" and "Name the monument" quests into one ?

westnordost: > Do you think it would be possible to combine the "Is it a monument or a memorial" and "Name the monument" quests into one ?

I haven't been following the discussion, but what is certainly possible would be an other-answer "This is just a memorial" when being asked about the name of a monument.

HugoTHOLLON: Yes indeed but I'm not sure a user really understand the difference between the 2 and he might just write a name even if it's a memorial but it might be the best solution. You tell me.

westnordost: But if a user cannot be supposed to really understand the difference between the two, a dedicate quest that asks whether something is a monument or memorial is even more so not eligible, right?

HugoTHOLLON: I think not because a dedicated quest explicitly explains what is the difference (using the hint) while in the "Name the monument" quest would only have an option "This is just a memorial" without further explaination. 

Maybe if there was the same hint in "Name the monument" quest then a dedicated quest would be less intresting.

westnordost:
superseded by linked commit

PR #6588, opened 2025-10-30
Selecting multiple maxweight signs

### LINKED ISSUE #5803: Ability to select multiple weight limit signs
**Use case**
At least here in Czech Republic it's common to have 2 signs describing weight limit for different types of vehicles.

**Proposed Solution**
Add option to add more than one weight limit sign

paulklie:
fixes #5803


paulklie:
git does not like me, will try again with another branch

PR #6589, opened 2025-10-30
Allow selecting of multiple maxweight signs

### LINKED ISSUE #5803: Ability to select multiple weight limit signs

paulklie:
fixes #5803

### DIFF
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/max_weight/MaxWeightForm.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/max_weight/MaxWeightForm.kt
@@ -21,6 +21,7 @@ import de.westnordost.streetcomplete.data.meta.WeightMeasurementUnit
 import de.westnordost.streetcomplete.resources.Res
 import de.westnordost.streetcomplete.resources.ic_delete_24
 import de.westnordost.streetcomplete.resources.quest_maxweight_add_sign
+import de.westnordost.streetcomplete.resources.quest_maxweight_remove_sign
 import de.westnordost.streetcomplete.ui.common.Button2
 import de.westnordost.streetcomplete.ui.common.dialogs.SimpleItemSelectDialog
 import org.jetbrains.compose.resources.painterResource
@@ -63,7 +64,10 @@ fun MaxWeightForm(
                     modifier = Modifier.weight(1f)
                 )
                 IconButton(onClick = { onSignRemoved(index) }) {
-                    Icon(painterResource(Res.drawable.ic_delete_24), null)
+                    Icon(
+                        painter = painterResource(Res.drawable.ic_delete_24),
+                        contentDescription = stringResource(Res.string.quest_maxweight_remove_sign)
+                    )
                 }
             }
         }


westnordost:
Could you add a video/screenshot? I think a openening-hours-form-like UI would make sense, with an 🗑️ icon button to the right, no? (After all, the sign is so big.)

Maybe, the signs themselves could also be made smaller (again). The current size is due to in Material Design / Compose, the content paddings of all text inputs are so huge. Not sure if that default can be changed (to zero content padding, after all, the sign is already content padding enough).

westnordost:
First steps in Compose, very nice! 👍

Note that Compose suffers somewhat from that it is very easy to over-indent the code. To mitigate this, one needs to deliberately think about what composables should go into a new composable function. For specific composables that shouldn't be public, well, they can be private in that file.

Anyway, the structure should be the following:
- `MaxWeightForm` should be the topmost composable, containing a column of `MaxWeightSignForm`s each with a delete-button in its row and an add-new-sign-button at the bottom(?)
- the (old) `AddMaxWeightForm` should do little more than to call `MaxWeightForm`, like it is now. Remember that `AddMaxWeightForm` is going away in the mid term

paulklie:
Very much wip  for now 
image

paulklie:
> Could you add a video/screenshot? I think a openening-hours-form-like UI would make sense, with an 🗑️ icon button to the right, no? (After all, the sign is so big.)

I feel like having a list of potentially five, full sized signs would be very long. Though I guess such a case would be very rare.

paulklie:
Some further work:
**
image
**

paulklie:
US: 
image

westnordost:
Screenshots look good! Although, maybe the signs really need to be made smaller again (by removing the content padding of the text fields, if possible).

Also, note [my previous comment](https://github.com/streetcomplete/StreetComplete/pull/6589#issuecomment-3473398994). I see you haven't changed anything in this regard.

paulklie:
> Also, note [my previous comment](https://github.com/streetcomplete/StreetComplete/pull/6589#issuecomment-3473398994). I see you haven't changed anything in this regard.

My bad, missread the similar file names, should be good now?

paulklie:
> (by removing the content padding of the text fields, if possible).

From my understanding that seems to be defined globaly (for all text fields) so this would modify many other forms.

However I also think that more than 2 signs is exceptionally rare, so its probably ok if this rare case would use a bit more space right? Might be better if the more common case (single sign) stays easily readable.

westnordost:
I can see you struggled with this one. On the other hand, it's your first dip into compose code. Anyway, there were too many things wrong with it, so I basically rewrote the whole thing. Would you kindly review the state the PR is in now?

westnordost:
> From my understanding that seems to be defined globaly (for all text fields) so this would modify many other forms.

It's true, this is one of the reasons I created the `TextField2` composable.

westnordost:
Anyway, it's good to merge from my point of view, but please have a look at the PR before that.

westnordost (approved):

paulklie:
LGTM, cannot approve my own PR

PR #6590, opened 2025-11-01
highlight other aerialways in AddAerialwayBicycleAccess

paulklie:
Aerialways are not rendered ([and cannot be rendered](https://github.com/streetcomplete/maplibre-streetcomplete-style/issues/5)) on the map. Since at times multiple aerialways may be very close to eachother, with few other landmarks nearby that one can use to differentiate it can be possible to confuse them.
Before:
image
After:
image


westnordost:
Yeah, makes sense, thank you!

PR #6592, opened 2025-11-01
Ferry unification

### LINKED ISSUE #6541: Move ferry route detection
In https://github.com/streetcomplete/StreetComplete/pull/6375 we implemented functionality in the ferry access quests to ensure that we only ask this quest for route=ferry relations and route=ferry ways that are not part of a route=ferry relation.

This is quite a complex solution, as seen in https://github.com/streetcomplete/StreetComplete/pull/6428/files where we simply expanded the MaxWeight quest to ferries, since we require the same implementation for every quest on ferries.

However a lot of quests could be expanded to ferries:

- "operating times" style quest
- Maxheight
- Wheelchair accessibility
- Name
- Are showers available?
- Are toilets available?
- If food is available on board: is it vegan/vegetarian/kosher/halal?
- Is there internet access on board?
- Is card/cash payment accepted?

Re-implementing this logic for every quest is a lot of effort and adds a lot of complexity, and a huge amout of duplicate code.

Instead I suggest preprocessing the map data, similarly to how we remove certain, more complex relations (https://github.com/streetcomplete/StreetComplete/blob/e717cd80b2c502a366a5a067c59d224113727e98/app/src/commonMain/kotlin/de/westnordost/streetcomplete/ApplicationConstants.kt#L91), by removing any route=ferry ways that are part of a route=ferry relation before passing the map data to the quest filters.

This way we can simply filter for relations/ways with route=ferry and avoid any complex filtering in each quest.

paulklie:
fixes #6541 

Todo:

- [ ] Add new tests
- [ ] Convert Quests back to OsmElementQuestType

### DIFF
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/data/osm/mapdata/MapDataCache.kt
@@ -455,7 +455,9 @@ class MapDataCache(
             trim((maxTiles * 2) / 3)
         }
 
-        // Remove all ferry ways that are part of a ferry relation
+        // Remove all ways tagged with route=ferry that are part of a relation
+        // also tagged with route=ferry because that makes the former not actually a "real" ferry
+        // route (╯°□°）╯︵ ┻━┻. Tagging mistake or not, it is very common tagging (#6373)
         for (relation in result.relations) {
             if (relation.tags["route"] != "ferry") continue
             for (member in relation.members) {


paulklie:
@westnordost What do you think about this idea in theory?
This is mostly just a proof of concept at the moment and I am fully ok with closing this (and the related issue) if you think it is not worth it. Was just hoping to reduce the complexity of the ferry  fee quests.

westnordost:
Ah, sorry, IMO this goes in the wrong direction. Such very specific "ignore elements if..." logic has no place in MapDataCache. We do such filtering directly during download to drop some relation types, for performance reasons (see ApplicationConstants.kt `ignoreRelation`).

So, an option would be to put that logic there. But to be honest, I'd really not like to add (more) such specific mapping quirks at that location in the code unless absolutely necessary. It might have unintended side-effects. Dropping route relations (etc) is absolutely necessary for performance reasons.

If we want to make it shorter, I could imagine to add a function in `FerryUtils` like (pseudo-code, didn't test)

```kt
/** ways tagged with route=ferry that are part of a relation are not considered
    "real" ferry routes but basically just ferry lines (#6373). (╯°□°）╯︵ ┻━┻.
    Tagging mistake or not, it is very common tagging */
fun MapData.filterRealFerryRoutes() : Sequence<Element> {
  val wayIdsInFerryRoutes = wayIdsInFerryRoutes(mapData.relations)
  return this
    .filter("ways, relations with route = ferry")
    .filter { it !is Way || it.id !in wayIdsInFerryRoutes }
}
```

PR #6595, opened 2025-11-03
Bicycle charging quests

westnordost:
implement quests that 1. ask whether bicycles can be charged, too, then 2. ask how many at the same time

supersedes #6536, #6531

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
@@ -18,6 +18,7 @@ class AddChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
         nodes, ways with
           amenity = charging_station
           and !capacity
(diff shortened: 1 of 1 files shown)


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charging_station_capacity/AddChargingStationCapacity.kt
```diff
@@ -11,21 +11,20 @@ import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.
 import de.westnordost.streetcomplete.osm.Tags
 import de.westnordost.streetcomplete.osm.updateWithCheckDate
 
+// capacity for cars, actually
 class AddChargingStationCapacity : OsmFilterQuestType<Int>(), AndroidQuest {
 
     override val elementFilter = """
         nodes, ways with
           amenity = charging_station
           and !capacity
```
paulklie, line 20: would need to check capactiy:motorcars to

westnordost: Oh, right, thanks!

PR #6498, opened 2025-09-15
Exclude lifeguard and fire lookout towers from tower quests.

paulklie:
See discussion here: https://community.openstreetmap.org/t/reporting-streetcomplete-issues/135672/9

PR #6597, opened 2025-11-04
restricted observation towers to access=yes

paulklie:
This serves as a improved solution to https://github.com/streetcomplete/StreetComplete/pull/6498

Since we now ask for tower access we can restrict the fee quest to towers with `access=yes`


westnordost:
And customers? People who are *in* the e.g. amusement park and could answer that `access=customers` will also be able to answer these other quests.

paulklie:
> And customers? People who are _in_ the e.g. amusement park and could answer that `access=customers` will also be able to answer these other quests.

Good point. Not the fee one though right?

PR #6598, opened 2025-11-04
Extended quests to hitchhiking spots

### LINKED ISSUE #6527: "Signed hitchhiking spot" quests.
Once https://github.com/openstreetmap/id-tagging-schema/pull/1705 gets merged to id, and a new release created, I would like to extend the following quests to [highway=hitchhiking](https://wiki.openstreetmap.org/wiki/Tag%3Ahighway%3Dhitchhiking):

| Quest Name            | Question                                 | Coverage |
|-------------------|------------------------------------------|----------|
| AddBusStopShelter | Is there a shelter at this stop?         | 26%      |
| AddBusStopBin     | Is there a waste basket at this stop?    | 7.5%     |
| AddBusStopBench   | Is there a bench at this stop?           | 58%      |

Do you think the current quest titles are sufficient, or would additional ones need to be added for this case?

Additionally I would add it to the "things" section, so users can more easily add it to the map.

paulklie:
fixes #6527


paulklie:
Somehow missread and though the hitchhiking station was already merged in id. Will reeopen once that happens

westnordost:
It is not necessary for the iD preset to be merged, we can already add support before.

westnordost:
Ah, I get it, without iD preset, the thing cannot be named. Well, ...anyway, not the end of the world.

riQQ:
For reference
- openstreetmap/id-tagging-schema#1675
- openstreetmap/id-tagging-schema#1705

PR #6601, opened 2025-11-05
Extended quests to Watchtowers

paulklie:
Currently we only ask for observation towers, however [watchtowers](https://wiki.openstreetmap.org/wiki/Tag:tower:type%3Dwatchtower) can also be interesting and "visitable". This is most commonly the case if they are `historical=yes`.

Examples:

- Roman watch towers in the [limes](https://en.wikipedia.org/wiki/Limes_(Roman_Empire))
- Former border guard towers (for example along the iron curtain).

I did not extend the AddGeneralFee quest, since there are very few such towers with `fee=yes`, and most of those seem like incorrectly tagged observation towers.


westnordost:
Watchtowers should be considered by default private.

paulklie:
Wheelchair and step count is only asked for `access=yes` or `customer`. Or are you saying we should not ask about the access tag in general?

PR #6603, opened 2025-11-06
implement auto complete text field in compose

westnordost:
for the quest forms that ask for the operator of something.

However, the bottom sheet seems to confuse Jetpack Compose: The suggestions dropdown/popup is often not shown at the correct position, overlapping the actual text field.

Should only merge this after when the bottom sheet has been reimplemented in compose.


westnordost:
I merged this into #6842

PR #6605, opened 2025-11-07
Artwork Type

### LINKED ISSUE #2125: New Quest: Artwork Type
### General
Affected tag(s) to be modified/added: [artwork_type=*](https://wiki.openstreetmap.org/wiki/Key:artwork_type)
Question asked: **What type of artwork is this?**

I have been adding artwork information to the map because I thought my region was starved for public art. It turns out there's a lot, but a lot of information is missing from OSM. Many pieces (not just in my area) are tagged solely with `tourism=artwork` and no more. At the bare minimum, I think adding what type of art it is would be nice, although adding name and artist tags could be good too (the one could be added to the name quest?).

### Checklist
This tag can be derived from street-level imagery, but almost certainly not satellite.
Any one region should not encounter this tag very often. In my small city center, there are 7 untagged artworks.
There are 147,213 total `tourism=artwork`, 100,129 are tagged with `artwork_type`, leaving around 47,000 to survey.

### Ideas for implementation

**Element selection:** 

```
[out:json][timeout:25];
(
  node["tourism"="artwork"]["artwork_type"!~".*"]({{bbox}});
  way["tourism"="artwork"]["artwork_type"!~".*"]({{bbox}});
  relation["tourism"="artwork"]["artwork_type"!~".*"]({{bbox}});
);
out body;
>;
out skel qt;
```

It might not be necessary to ask about relations, but nodes and ways are clear enough to answer.

**Metadata needed:** 
None? Some artwork types from the wiki are probably very regional, but I would think limiting to the most-used options would eliminate this concern (e.g the Spanish/Portuguese `artwork_type=azulejo` is only used 8 times). 

**Proposed GUI:**

I think it should be along the lines of the surface quest, with square images with a short label on top of them. Most of the images from the wiki are decent representations, but some might be vague, such as Installation or Mural vs. Painting.

**What type of artwork is this?**
| Select the best option | | |
| :---: | :---: | :---: |
| <br>Sculpture | <br>Statue (person or animal) | <br>Mural (wall/ceiling artwork) |
| <br>Graffiti | <br>Installation | <br>Bust |
| <br>Stone | <br>Painting | <br>Architecture |
| <br>Mosaic/Tilework | <br>Relief | Land Art? |

Note: Re: Mosaic/Tilework, `=mosaic` is far more common, but the wiki mentions `=tilework`...

paulklie:
Closes #2125

I copied most of the implementation from SCEE (https://github.com/Helium314/SCEE/pull/796) by @vfosnar, with some modifications:

- [x] 1. Migrated to compose
- [x] 2. Resorted elements based on current usage numbers (most popular at the beginning)
- [x] 3. Replaced some images that were not in public domain, or had licensing conflicts.
- [x] 4. Added license information for all pictures
- [x] 5. Added other landart
- [x] 6. Removed [landart](https://wiki.openstreetmap.org/wiki/Tag:artwork_type%3Dland_art).

I removed [landart](https://wiki.openstreetmap.org/wiki/Tag:artwork_type%3Dland_art) for the following reasons:
1. It is quite rare, with less than 0.3% of artwork_type uses.
2. Since landarts are often visible from space, I suspect there is a higher likelihood that they are already mapped by editors using satellite images. Thus an even smaller portion of the remaining unclassified artwork, would be landarts.
3. It has the possibility of being mis-tagged instead of the more popular [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[geoglyph](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dgeoglyph). Explaining the difference between these two seems beyond scope.
4. A, comparatively large portion of landarts are relations, and are thus not included anyway: While landarts only count as 0.3% of artworks, they count as almost 14% of artwork relations.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/artwork/AddArtworkType.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/artwork/AddArtworkType.kt
@@ -16,7 +16,7 @@ class AddArtworkType : OsmFilterQuestType<ArtworkType>(), AndroidQuest {
 
     override val changesetComment = "Survey artwork type"
     override val wikiLink = "Key:artwork_type"
-    override val icon = R.drawable.quest_memorial
+    override val icon = R.drawable.quest_artwork
     override val achievements = listOf(CITIZEN)
 
     override fun getTitle(tags: Map<String, String>) = R.string.quest_artwork_title


matkoniecz:
I see that it is a draft - what remains to be done? Do you need some help?

paulklie:
Pretty much just testing, or giving feedback on the selected images.

mnalis:
Can we also add support for [artwork_type=street_art](https://wiki.openstreetmap.org/wiki/Tag:artwork_type=street_art) ?

paulklie:
> Can we also add support for [artwork_type=street_art](https://wiki.openstreetmap.org/wiki/Tag:artwork_type=street_art) ?

I think it could work, I think we would need to title it as "other street art".
Have an idea for picture? Since this category is quite wide

mnalis:
> I think it could work, I think we would need to title it as "other street art". Have an idea for picture? Since this category is quite wide

Yeah, _"other street art"_ sounds good. Perhaps this picture is somewhat representative (can't really be put in other categories):
https://wiki.openstreetmap.org/wiki/File:Painted_hand_water_pump_in_Zagreb,_Most_mladosti.jpg

paulklie:
> https://wiki.openstreetmap.org/wiki/File:Painted_hand_water_pump_in_Zagreb,_Most_mladosti.jpg

I have the feeling this might be hard to recognized when scaled down for smaller screen sizes.

mnalis:
> I have the feeling this might be hard to recognized when scaled down for smaller screen sizes.

Perhaps crop the prominent part of it? Other examples (also on that wiki) that I've had is likely even harder to recognize.


On the other hand, most of street art matches on wikimedia commons are mostly  `graffiti`, `mural` (with few `installation` and `mosaic`), so I've given up searching in that way after few dozen pages... If I know what to search for, there are few more matches similar to that image on commons (see e.g. [pimpmypump](https://commons.wikimedia.org/w/index.php?search=pimp+my+pump&title=Special%3AMediaSearch&type=image)).

E.g. maybe [this one](https://commons.wikimedia.org/wiki/File:Painted_hand_water_pump_in_Zagreb,_Stubi%C4%8Dka.jpg) has better contrast?

Or perhaps looking if something else already mapped in OSM as `street_art` which has picture would suit it better? https://overpass-turbo.eu/s/2fjc

app/src/commonMain/composeResources/values/strings.xml
```diff
@@ -730,6 +730,22 @@ Before uploading your changes, the app checks with a &lt;a href="https://www.wes
...
+    <string name="quest_artwork_installation">Installation</string>
+    <string name="quest_artwork_bust">Bust</string>
+    <string name="quest_artwork_stone">Stone</string>
+    <string name="quest_artwork_painting">Painting</string>
+    <string name="quest_artwork_architecture">Architecture</string>
+    <string name="quest_artwork_mosaic">Mosaic</string>
+    <string name="quest_artwork_relief">Relief</string>
+    <string name="quest_artwork_fountain">Fountain</string>
+    <string name="quest_artwork_land_art">Land art</string>
+
```
westnordost: unused string

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/artwork/ArtworkType.kt
```diff
@@ -0,0 +1,25 @@
...
+    RELIEF("relief"),
+    ARCHITECTURE("architecture"),
+    FOUNTAIN("fountain")
+}
+
+fun ArtworkType.applyTo(tags: Tags) {
+    tags["artwork_type"] = this.osmValue
+    if (this.osmArtworkTypeValue != null) {
+        tags["artwork_type"] = this.osmArtworkTypeValue
+    }
```
westnordost: ???

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/artwork/ArtworkTypeItem.kt
```diff
@@ -0,0 +1,48 @@
...
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.FOUNTAIN
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.GRAFFITI
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.INSTALLATION
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.MOSAIC
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.MURAL
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.PAINTING
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.RELIEF
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.SCULPTURE
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.STATUE
+import de.westnordost.streetcomplete.quests.artwork.ArtworkType.STONE
```
westnordost:
```suggestion
import de.westnordost.streetcomplete.quests.artwork.ArtworkType.*
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/artwork/ArtworkTypeItem.kt
```diff
@@ -0,0 +1,48 @@
...
+import de.westnordost.streetcomplete.resources.*
```
westnordost, line 4: No way, this is possible? Ouch! Gotta replace a lot of long `import de.westnordost.streetcomplete.resources.………` lists

paulklie: ehh, should I replace it on this branch then to?

westnordost: replace what with what?

paulklie: with individual resource imports

westnordost: no, I want it the other way around! :-D

paulklie:
> E.g. maybe [this one](https://commons.wikimedia.org/wiki/File:Painted_hand_water_pump_in_Zagreb,_Stubi%C4%8Dka.jpg) has better contrast?

I tried this one, and I think it works quite well:photo_2025-12-28_02-18-15

paulklie:
As for the icon, how about an icon like this: https://www.svgrepo.com/svg/359716/art-gallery ?

westnordost:
Well, like this, I guess. That one is black and white of course.

Or maybe take the icon of some open source (desktop?) paint app? It would be a nice gimmick. GIMP icon although arguably the biggest one is maybe not recognized well enough? Maybe MyPaint: https://github.com/mypaint/mypaint ?

westnordost:
The other street art icon is very good, because it shows two different kinds of street art.

westnordost:
This looks to be the current (flat) gimp icon:

https://commons.wikimedia.org/wiki/File:The_GIMP_icon_-_v3.0.svg

I am kind of intrigued....

westnordost:
It looks like this is done minus the quest icon. Good work!

westnordost:
(By the way, tried out MyPaint for the first time. It looks like a simpler Krita, i.e. geared towards "real" painting (with an actual canvas pattern as background). Must be pretty fun when you have a pen tablet. But boy is it slow (on my computer).)

Anyway, what do you think about trying the GIMP icon? At least among (FOSS) nerds, it should be well known and would be a nice gimmick. I expect there is a considerable intersection between OSM nerds and FOSS nerds.

mnalis:
> This looks to be the current (flat) gimp icon:
> https://commons.wikimedia.org/wiki/File:The_GIMP_icon_-_v3.0.svg

I believe that one might be (in a process of being) trademarked: https://uspto.report/TM/99400787
So perhaps best to contact GNOME foundation about allowable use, if it is to be chosen?

westnordost:
What does that mean? IANAL. In the linked filing, it says that it is a

> Mark For: This trademark registration is intended to cover the categories of downloadable computer software for creating, editing, manipulating, converting, and enhancing digital images.

mnalis:
**TL;DR:** it might, or might not, be problematic. Best to clarify with the [GNOME foundation](https://foundation.gnome.org/legal-and-trademarks/).

---

> What does that mean?

Trademark means that the usage of  logo enjoys certain protections under the law. Anyone may use it freely[^2] to refer exactly to **GIMP**, but to use it to refer to **different** things  (inside protected class) without having permission from trademark owner  would be considered violation, AFAIK.

>   Mark For: This trademark registration is intended to cover the categories of downloadable computer software for creating, editing, manipulating, converting, and enhancing digital images.

Yeah, that quote is why (instead of straight-out claiming StreetComplete may _not_ use it) I merely _suggested_ it would be best to _verify_ with GNOME foundation if we can use it for that purpose.

Because, its registration **also** says that its Primary International Class includes _"**surveying** [apparatus and instruments]"_ and _"data processing equipment and computers"_ which StreetComplete app on mobile phone _might_ be matching. 

But, again, it might not, as the text you quoted might be taking precedence[^1] which is why I recommended seeking clarification/permission from trademark owner might be a good idea.

[^2]:  one may need to indicate it is a trademark of the GNOME foundation somewhere, though
[^1]: while I might have some little more knowledge of few "intellectual property" laws then average netizen, it is more focused on copyright than trademarks, and definitely IANAL either.

FloEdelmann:
For reference: icon added in https://github.com/streetcomplete/StreetComplete/commit/4303311ec93eee052a7aeb409d97bc6644dbeaa7

westnordost:
Note the strings were added to the wrong strings.xml file.

Right now, we have two identical `strings.xml` files, one in `androidMain/res/values` and one in `commonMain/composeResources/values`. The one in `androidMain` currently remains the **source**, the other is just a copy.
I'll soon change that though (switch it around).

PR #6615, opened 2025-11-12
Recycling quest UI change

### LINKED ISSUE #6608: "PET only" image is misleading
In the recycling type quest, when you select "Plastic" SC shows a few detailed options.

Image

Selecting `Plastic bottles only` correctly adds `recycling:plastic_bottles=yes`.

Selecting `PET only` correctly adds `recycling:PET=yes` but the image is misleading. It appears as if only PET bottles are allowed, but according to the wiki `recyling:PET` allows anything made of PET.

Quote from [recycling:PET](https://wiki.openstreetmap.org/wiki/Key:recycling:PET) wiki page:
> The key recycling:PET indicates whether all kind of material made of polyethylene terephtalate (PET/PETE) are accepted for recycing or not.
> 
> Please use [recycling:pet_drink_bottles](https://wiki.openstreetmap.org/wiki/Key:recycling:pet_drink_bottles)=yes if only beverage bottles made of polyethylene terephtalate are accepted for recycling.

"PET only" should be renamed to "PET plastic bottles only" and its tag should be changed to `recycling:pet_drink_bottles`, The image is fine for this case.

If actual `PET only` containers are so common, a separate option can be added, but it should feature only the PET logo.

I will also note that such bins are explicitly labeled as accepting only PET drink bottles. I haven't seen bins which are explicitly labeled as accepting anything made from PET.

Image

Green text: "New start for each PET bottle."
White text below: "Dump your plastic drink bottles here"
Image with green tick: water, tea/juice and beer plastic bottles
Images with red circles: packaging of cosmetics, cleaning products, detergents, cooking oil, engine oil, domestic waste.

Photo taken from https://www.mezdra.bg/?p=63342

### LINKED ISSUE #6613: [v62.0-beta1] "any" plastic for recycling is now unclear
**How to Reproduce**
- Solve the recycling quest
- click on plastic
- see the first option in "plastic" submenu - is now just generically called "Plastic" (well, they are _all_ "Plastic", so it is confusing)

**Expected Behavior**
see it clearly described as "Any plastic" (as it was described before)

Pics for comparison:

Image Image

**Versions affected**
v62.0-beta1, Android 15

westnordost:
Make the list flat, but recognize which item is a subcategory of which other.

Fixes #6613

Also, fixes #6608

https://github.com/user-attachments/assets/ab3a897a-ee63-4dfc-b220-c53c0230d15c

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterialsForm.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/recycling_material/AddRecyclingContainerMaterialsForm.kt
@@ -6,10 +6,14 @@ import androidx.appcompat.app.AlertDialog
 import androidx.compose.material.Surface
 import androidx.compose.runtime.mutableStateOf
 import de.westnordost.streetcomplete.R
+import de.westnordost.streetcomplete.data.preferences.Preferences
 import de.westnordost.streetcomplete.databinding.ComposeViewBinding
 import de.westnordost.streetcomplete.quests.AbstractOsmQuestForm
 import de.westnordost.streetcomplete.quests.AnswerItem
 import de.westnordost.streetcomplete.ui.util.content
+import de.westnordost.streetcomplete.util.takeFavorites
+import org.koin.android.ext.android.inject
+import kotlin.getValue
 
 class AddRecyclingContainerMaterialsForm : AbstractOsmQuestForm<RecyclingContainerMaterialsAnswer>() {
 
@@ -17,17 +21,26 @@ class AddRecyclingContainerMaterialsForm : AbstractOsmQuestForm<RecyclingContain
     private val binding by contentViewBinding(ComposeViewBinding::bind)
     override val defaultExpanded = false
 
+    private val prefs: Preferences by inject()
+
     override val otherAnswers = listOf(
         AnswerItem(R.string.quest_recycling_materials_answer_waste) { confirmJustTrash() }
     )
 
+    private lateinit var reorderedItems: List<RecyclingMaterial>
     private val selectedItems = mutableStateOf(emptySet<RecyclingMaterial>())
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        reorderedItems = moveFavouritesToFront(RecyclingMaterial.entries)
+    }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
         binding.composeViewBase.content { Surface {
             RecyclingContainerMaterialsForm(
+                items = reorderedItems,
                 tree = RecyclingMaterial.tree,
                 selectedItems = selectedItems.value,
                 onSelectedItems = {
@@ -39,6 +52,7 @@ class AddRecyclingContainerMaterialsForm : AbstractOsmQuestForm<RecyclingContain
     }
 
     override fun onClickOk() {
+        prefs.addLastPicked(this::class.simpleName!!, selectedItems.value.toList())
         applyAnswer(RecyclingMaterials(selectedItems.value))
     }
 
@@ -51,5 +65,12 @@ class AddRecyclingContainerMaterialsForm : AbstractOsmQuestForm<RecyclingContain
             .setNegativeButton(R.string.quest_generic_confirmation_no, null)
             .show()
         }
+    }
+
+    private fun moveFavouritesToFront(originalList: List<RecyclingMaterial>): List<RecyclingMaterial> {
+        val favourites = prefs
+            .getLastPicked<RecyclingMaterial>(this::class.simpleName!!)
+            .takeFavorites(4)
+        return (favourites + originalList).distinct()
     }
 }
--- a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/recycling_material/RecyclingContainerMaterialsForm.kt
+++ b/app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/recycling_material/RecyclingContainerMaterialsForm.kt
@@ -10,7 +10,6 @@ import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import com.cheonjaeung.compose.grid.SimpleGridCells
@@ -27,13 +26,12 @@ import org.jetbrains.compose.resources.stringResource
  *  certain values are actually sub-categories of other values. */
 @Composable
 fun RecyclingContainerMaterialsForm(
+    items: List<RecyclingMaterial>,
     tree: Node<RecyclingMaterial>,
     selectedItems: Set<RecyclingMaterial>,
     onSelectedItems: (Set<RecyclingMaterial>) -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    val flattenedItems = remember(tree) { tree.yieldAll().toList() }
-
     Column(
         modifier = modifier,
         verticalArrangement = Arrangement.spacedBy(8.dp)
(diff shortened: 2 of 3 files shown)


app/src/androidMain/res/values-en/strings.xml
```diff
@@ -1036,6 +1036,7 @@ A level counts as a roof level when its windows are in the roof. Subsequently, r
     <string name="quest_carWashType_service">Staff cleans car</string>
 
     <string name="quest_charging_station_bicycles_title">Can bicycles be charged here, too?</string>
+    <string name="quest_charging_station_bicycles_answer_only">It’s only for bicycles</string>
```
paulklie, line 1039: Did this slip in accidentally?

westnordost: kind of. This file is automatically generated, so missed this on master.

mnalis (commented):
While code itself is somewhat above my pay grade 😅 , the video solves https://github.com/streetcomplete/StreetComplete/issues/6613 and looks good.

**Question:** is the list going to autosorted by most-recently-used? 

- If it will auto-sort (hopefully), then it additionally improves usability a lot (I really disliked having to additionally click _"any plastic"_ in 99% of the cases in old SC versions; it just wouldn't enter my muscle memory even if I entered it a lot - probably because all other recycling answers were single-click-for-answer)

- if not, then, uh, need to scroll down all the time would **not** be nice sideeffect 😿 

To me the new flat UI seems nicer (provided auto-sort mentioned above) and more efficient, and my wife also seemed to understand the concept of selecting some values (e.g. _"PET bottles"_) deselecting its "parent" category (e.g. _"any plastic"_) . While only anecdotal evidence, I hope other newbies will not be too confused either.

Other then that, only other potential confusion I see that _"beverage cartons"_ are considered _"plastic"_. Technically, they are [multi-layer](https://en.wikipedia.org/wiki/Tetra_Pak#/media/File:TBA_packaging_components.svg) (e.g. plastic+paper+aluminium); but hopefully that is not confusing (it would not be a problem in Croatia, as we nowadays put them in same container as _"any plastic"_ but I don't know if that might be confusing in other countries - e.g. if they are treated/collected differently then other plastic 🤷 )

westnordost:
it doesn't auto-sort, no. I think this would be confusing, as the sub-categories are all next to each other (behind each other)

paulklie:
Tbh, I think this iteration is quite confusing, took me a good 3 min to find out how to use it. My main gripe:

It is not clear that the "only" elements only apply to plastic stuff. A bin that accepts, for example, plastic packaging and paper might cause a user to not select "plastic packaging only" since they misunderstand what it is asking.

paulklie:
How about separating the plastic into:
- PET bottles
- Plastic bottles
- Drink cartons
- Plastic packaging
- Other, non packaging, plastic

And allowing multiple to be selected.

westnordost:
> It is not clear that the "only" elements only apply to plastic stuff. A bin that accepts, for example, plastic packaging and paper might cause a user to not select "plastic packaging only" since they misunderstand what it is asking.

It doesn't say "plastic packaging only", it says "any plastic packaging"

westnordost:
> Other, non packaging, plastic

It is not really possible to tag that. It would complicate things a lot.

paulklie:
> It is not really possible to tag that. It would complicate things a lot.

Would that not be 
[recycling:plastic](https://wiki.openstreetmap.org/wiki/Key:recycling:plastic)=yes
[recycling:plastic_packaging](https://wiki.openstreetmap.org/wiki/Key:recycling:plastic_packaging)=no

westnordost:
Yeah, probably. (But my statement that it complicates things stands.)

mnalis:
**TL;DR:** I'd prefer either new UI with auto-sort, or (if that is not possible) then retaining old UI instead.

---

> it doesn't auto-sort, no.

Uh, that is quite unfortunate. :cry:  Because, many of the shown features either:
- do not exist in Croatia at all (e.g. plastic packaging only, plastic bottles only, beverage cartons only,  shoes, food waste, ...), or 
- are extremely rare (i.e. mostly only in recycling centers, not regular containers) e.g. any glass, scrap metal, batteries, electric appliances, cooking oil, engine oil, ...
   And for recycling centers, the quest is unusable and I always need to leave a note anyway - as they take a lot of other unselectable things too (like paints, rubber, wood planks, medicines, rubble, asbestos etc. )

E.g. over here in 98% of the cases are just combination of 4 things: _"glass bottles and jars"_, _"paper"_, _"any plastic"_+_"cans"_ (we put them in same container, i.e. [recycling:pmd=yes](https://wiki.openstreetmap.org/wiki/Key:recycling:pmd)[^1]), _"compostable waste"_. 

Having those at **top would be extremely helpful**; while having them interspersed with things that do not exist here (and having to scroll down and search for things that are extremely common) is not a very pleasurable UX :man_shrugging: 

>  I think this would be confusing, as the sub-categories are all next to each other (behind each other)


I get the reasoning, but it doesn't work all that well when those things do not exist, IMHO.
 
And it is not perfect even when they are one after another as shown in video, due to splitting into multiple rows, so the confusion is still there IMHO (as noted by @paulklie ).

Anyway, if auto-sort is no-go, then I think UI used in previous versions of SC (even with its shortcomings) would be preferable to the new UI (both for yours [i.e. it would be more understandable for newbies], and for mine [i.e. easier for regular users to use] use cases).

[^1]: however, as PMD is not offered, we have to remember to select _"any plastic"_ + _"cans"_ answers separately, which is often forgotten (even my myself :crying_cat_face:), as we call them just "container for plastic", so when you mentally check off that one as marked, it is likely you'll forget about -- i.e. you've marked 3 items, and there are 3 containers, so that is "done" -- except it isn't :cry: .
  Having all of the common values one after another sorted at the top would make it much less likely to miss; and much easier to select appropriate containers.

westnordost:
Jesus, the time you have me invest in *one* quest...

westnordost:
I find pmd problematic because it is a tag that is essentially three different things at once: plastic packaging, drink cartons and metal (cans) packaging. So, one could say that pmd is a supercategory of these three things. But then, what is `plastic`? *Also* a supercategory of plastic packaging, but orthogonal to that! Argh!

mnalis:
**TL;DR:** I'd prefer if SC's recycling quest would allow for easy tagging in Croatia (e.g. by sorting and/or `:pmd`). 
But it is not the end of the world if it doesn't.

---

> I find pmd problematic because it is a tag that is essentially three different things at once

Yeah, I concur, but I didn't invent that tag just to complicate your life :sweat_smile:  It just happens to describe how those containers operate in Croatia (even if we don't call it "PMD" but just colloquially _"Plastic recycling containers"_) :shrug: 

The tagging situation here is _very_ inconsistent (probably due to many tags being documented at different times, and not being offered by editors) - e.g. in Zagreb, where it is almost always one container for plastic+cans+tetrapak-alike-multilayered:
- great majority of tagging uses just `:plastic`,
- about 13% uses somewhat more correct combination `:plastic`+`:cans`,
- significantly less that 0.1% use the ***actually correct***  `:plastic` + `:cans` + `:beverage_cartons` (or alternatively just `:pmd`)

---

Now, I would like if StreetComplete would help improve that tagging situation here. E.g. by: 
- Having a single click describe all three (regardless if it would tag 3 separate tags, or one PMD tag) is one helpful option, and/or
- Having all commonly used values sorted at top is another 

But having **3** different values that all have to be selected to mark **one** container, and especially having those values interspersed all around the scrollable list full of options that don't exist in Croatia, would IMHO be quite unlikely to improve the situation (to say the least) :man_shrugging: 

> Jesus, the time you have me invest in _one_ quest...

Well, yeah, sorry, but you specifically _asked_ for my opinion :wink: 

As the saying goes _"People don't want to hear your opinion. They want to hear their opinion, coming out of your mouth"_ :smiley: 
Just joking! But do rest assured, that we all do hugely appreciate your effort, even if don't always get our way :heart:

westnordost:
You are right, it is not your job to consider effort (vs impact). I'm at the peculiar position to be both the developer - striving for code cleanliness, consistent UI design, - and product manager, having to look at effort vs impact.

Your comment about you preferring the previous solution unless I added that extra feature caught me on the wrong foot because it felt like an ultimatum (and I know, it wasn't) for both personas who were screaming "no" here: The solution in this PR is more consistent with UI elsewhere in the app while also cleaner on the technical side even though I had to invest more than one day's worth of work in that.

Anyway, I've looked into reordering and it wasn't that much extra work, so I just implemented that.

westnordost:
> PMD

This is somewhat similar to the German system, by the way, known as "dual system", "green point" or "yellow sack"/"yellow container". However, there may be local differences. For example in Hamburg, the "yellow containers" also accept scrap metal and *any* plastic, not just packaging.

PR #6620, opened 2025-11-15
expanded maxheight to covered bridges

paulklie:
Also asks maxheight for bridge=[covered](https://wiki.openstreetmap.org/wiki/Tag:bridge%3Dcovered).

Of the 4583 such bridges 266 currently have maxheight set (5.8%).


westnordost:
Sounds reasonable. But does the wording fit?

paulklie:
"What height limit is indicated here?" seems to be generic enough to me.

PR #6621, opened 2025-11-15
expand general fee to device charging station

paulklie:
Currently device_charging_station (supported as a "thing" since [tagging 6.13.2](https://github.com/openstreetmap/id-tagging-schema/releases/tag/v6.13.2), recommends fee on the [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddevice_charging_station).

Currently the data leans towards `no` with 510 elements, with only 39 tagged `fee=yes`, however since charging stations are quite rare I think this would not be spammy.


westnordost:
But does the wording fit? After all, the quest is usually for *places* that require fee.

paulklie:
We already have a special title for amenities: "Do you need to pay to use this?".

Seems like it fits?

Issue #6398, opened 2025-07-13
Resurvey amenity=bicycle_repair_station tools


paulklie:
As discussed in https://github.com/streetcomplete/StreetComplete/pull/6397, we can resurvey the availability of tools at at a [amenity](https://wiki.openstreetmap.org/wiki/Key:amenity)=[bicycle_repair_station](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_repair_station). This will especially help with older stations that have had tools broken or damaged.

Currently the tools quest asks for any repair stations where at least one tool is not specified: https://github.com/streetcomplete/StreetComplete/blob/6eaf52d3eb7027b8d292f907b2b0b02366120523/app/src/main/java/de/westnordost/streetcomplete/quests/bicycle_repair_station/AddBicycleRepairStationServices.kt#L18 However the status of this tool is not preselected, to maintain similarity to the existing recycling quest as discussed here: https://github.com/streetcomplete/StreetComplete/issues/5910#issuecomment-2359649562 . The recycling quest however is already resurveyed.

I suggest we filter as following:
If all tools are set and age > 4 years:  
    Title: "Is the bicycle equipment available here still correct?".
If atleast one tool is not set: 
    Title: "What bicycle equipment is available here?" (same as currently)

For better usability i suggest we preselect any tools which are tagged as "yes" every case.

westnordost:
I'd rather keep implementation simple. I.e. do it as the recycling quest. No pre-selected stuff. After migration to Compose we might reconsider implementing pre-selection on re-survey consistently for all re-survey quests.

westnordost:
Will do this tmrw

PR #6622, opened 2025-11-15
implemented preselect for resurvey of bike repair station

paulklie:
I tried implementing preselecting elements on resurvey of selection quests, now that the migration to compose is done for these (see https://github.com/streetcomplete/StreetComplete/issues/6398). This was easier than expected.

I have completed bike repair stations for now. If this solution is accepted I would also implement this for the remaining quests.

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/AbstractQuestForm.kt
@@ -168,11 +168,14 @@ abstract class AbstractQuestForm :
 
     protected fun setTitle(text: CharSequence?) {
         binding.titleLabel.text = text
+        binding.titleLabelResurvey.text = "For: $text"
     }
 
     protected fun setResurvey(resurvey: Boolean) {
         binding.resurveyLabel.text = "Is this still correct?"
-        binding.resurveySpeechBubbleTitleContainer.isVisible = resurvey
+        binding.resurveyLabel.isVisible = resurvey
+        binding.titleLabelResurvey.isVisible = resurvey
+        binding.titleLabel.isVisible =!resurvey
     }
 
     protected fun setTitleHintLabel(text: CharSequence?) {
--- a/app/src/androidMain/res/layout/fragment_quest_answer.xml
+++ b/app/src/androidMain/res/layout/fragment_quest_answer.xml
@@ -30,42 +30,6 @@
             app:behavior_peekHeight="@dimen/quest_form_peekHeight">
 
             <LinearLayout
-                android:id="@+id/resurveySpeechBubbleTitleContainer"
-                android:background="@drawable/speech_bubble_none"
-                android:layout_width="match_parent"
-                android:layout_height="wrap_content"
-                android:orientation="vertical"
-                android:layout_marginStart="@dimen/horizontal_speech_bubble_margin"
-                android:layout_marginEnd="@dimen/horizontal_speech_bubble_margin"
-                android:layout_marginBottom="4dp"
-                android:showDividers="middle"
-                android:paddingStart="22dp"
-                android:paddingEnd="22dp"
-                android:paddingTop="16dp"
-                android:paddingBottom="16dp"
-                android:divider="@drawable/space_16dp"
-                android:elevation="@dimen/speech_bubble_elevation">
-                <LinearLayout
-                    android:layout_width="match_parent"
-                    android:layout_height="wrap_content"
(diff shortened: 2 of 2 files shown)


westnordost:
Well, for opening hours, the question is also adapted:

Is this ... still correct? The user can then just answer "yes".

If the stuff is just pre-selected without comment, users might rather think this is a bug or something. Also, to immediately make the (OK) button available is not fine, in my opinion. The user didn't give an answer, yet.

paulklie:
How about now? 
I added an additional string, and a yes/no confirmation.

I also noticed that I was mistaken: this quest form is not used with any other quests that have resurvey functionality. However I still made sure we can reuse this functionality if we ever decide to add resurveying to, for example, AddSport or AddBoatRental.

westnordost:
Yes, that is how it should be done. However, I am not quite convinced that it is worth the effort to
- add an extra string for every such re-survey-enabled quest
- add parsing to determine whether (and which) items should be pre-selected - at ***two*** locations, to boot

For what? The user still needs to check whether all that is pre-selected is still available (including scrolling down to see if any are selected further below, if applicable), the time saved of not having to click the things that are still available seems negligible.

matkoniecz:
> For what?

if user is lazy/has worse day then risk of damage to OSM data is lower

also (based on bicycle parking count preview, I kept that old buggy version for own use)..
Sometimes you think you noted/counted things, it mismatches prefilled data and it makes you look at things again - and then it turns out that you are wrong

does not happen very often to me, but far more often than I expected, and I like to think I am more careful than average mapper, especially newbie mapper

westnordost:
Your example is from bicycle parking count, though. This is about the image select.

And it works the other way around, too. E.g. recycling containers: You see paper and glass containers. Pre-selected seems to be paper and glass, but when you (would) scroll down, you see that also clothes is selected. This one is just an example off the top of my mind, my point is that a mistake could be introduced or a previous mistake not be corrected just as well.

matkoniecz:
I expect that similar may apply, though that is guessing

and I probably treat "not fixing data" as lesser problem than "breaking valid data"

also, if say 0.3% of edits is bad than breaking valid data in 0.3% of all resurveys is much worse than not fixing bad data in 0.3% of resurveys where data should be changed (as data needs to be changed only in some cases)

paulklie:
> add an extra string for every such re-survey-enabled quest

We might be able to avoid this by asking "Is this still correct?" and adding the original question in a seperate row. That way we only need to translate the "Is this still correct?" to all languages.

westnordost:
Hmm, interesting idea. So that title-logic would be implemented in the *Form class, and in case it is a re-survey quest, the two questions would be appended separated by a `\n` I presume? This also solves the issue that parsing to determine whether (and which) items should be pre-selected had to be done at two locations, right?

westnordost:
(or with an additional element in the layout that makes the question look visually different, e.g. slightly greyed out.)

> Is this still correct?
> for _"What can be recycled here?"_
>
> [ paper ] [ glass ] [ plastic ]
> [ shoes ] [ clothes ] [ compost ]
> ...
> Uh... |             No  |  Yes

paulklie:
> (or with an additional element in the layout that makes the question look visually different, e.g. slightly greyed out.)

I like this idea, would also allow removing a lot of resuvying strings.

Maybe could add a overridable function, something like `IsResurvey() : boolean`, with a default implementation that always returns false? And quests with resurvey capabilities can override it with their logic.

Maybe in future this could also be used to display a different icon for re-survey quests. For example, sometimes I am more likely to answer resurvey opening hours, since these require less time than adding the hours from scratch.

westnordost:
true, this could be possible, but I don't see the sense in showing a different icon for resurvey

paulklie:
photo_2026-01-01_17-22-13

What do you think about this?

westnordost:
Hmm, I guess the view should either just show what the user selected, or make it clear that the list cannot be interacted with (disabled). Also, for longer lists, this approach hardly works, because the currently selected things should be visible without scrolling. Also, the hint that "select all that apply" would be misleading in this case.

paulklie:
These quest forms have not been converted to compose yet (and will need to be at some point in the future) right? I guess it might make more sence for me to wait until then I guess? Since otherwise converting will just be more work

westnordost:
Which ones?

paulklie:
The "speech bubbles" with the quest title

westnordost:
Correct, they haven't. That comes after all quest forms gave been converted. (Migration happens bottom-up)

paulklie:
Yeah that makes sence, I will wait with this then

PR #6623, opened 2025-11-15
expanded shower quest to highway=services

paulklie:
Expanded the "AddCampShower" quest to highway=[services](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dservices). Really stretching the "camp" here, but the strings seem to fit well.

Currently 227 have `showers=yes` and only 8 have `showers=no`. Though I strongly suspect this is not reflective of the reality since stations without showers are less likely to have this information added.

I limited this quest to service stations with toilets=yes, since I find it unlikely that places without toilets would have showers, and we already ask for toilets.


westnordost:
IMO it makes sense, thank you!

Issue #6626, opened 2025-11-18
ApiClientException - Answer not uploaded


xrammit:
**How to Reproduce**
1. Answer an open question on the map.
2. A dialog appears saying "Error uploading the answer. Try again later."
3. The previously given answer to the open question was not uploaded to the server.

**Expected Behavior**

1. Answer an open question on the map.
2. The given answer was successfully uploaded to the server.

**Versions affected**

62.0

**Stacktrace**

(Last four digits of the node ID in the request URL were masked)

```
App version: 62.0
Device: google panther, Android 16
Locale: de-DE

Stack trace:
[de.westnordost.streetcomplete.data](https://de.westnordost.streetcomplete.data/).ApiClientException: Client request(GET https://api.openstreetmap.org/api/0.6/node/950767XXXX) invalid: 406 Not Acceptable. Text: "No acceptable content encoding found. Only identity and gzip are supported."
at [de.westnordost.streetcomplete.data](https://de.westnordost.streetcomplete.data/).osm.mapdata.MapDataApiClient.getMapDataOrNull(SourceFile:42)
at [de.westnordost.streetcomplete.data](https://de.westnordost.streetcomplete.data/).osm.mapdata.MapDataApiClient.access$getMapDataOrNull(SourceFile:23)
at [de.westnordost.streetcomplete.data](https://de.westnordost.streetcomplete.data/).osm.mapdata.MapDataApiClient$getMapDataOrNull$1.invokeSuspend(Unknown Source:12)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(SourceFile:34)
at [kotlinx.coroutines.DispatchedTask.run](https://kotlinx.coroutines.dispatchedtask.run/)(SourceFile:100)
at kotlinx.coroutines.EventLoopImplBase.processNextEvent(SourceFile:263)
at kotlinx.coroutines.BlockingCoroutine.joinBlocking(SourceFile:94)
at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(SourceFile:70)
at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source:1)
at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(SourceFile:48)
at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source:1)
[...]

PR #6628, opened 2025-11-19
Remove deflate compression from ContentEncoding

Firefishy:
Removed deflate compression as server can respond with zlib-wrapped DEFLATE format (RFC 1950) which is not supported by the io.ktor.client which I believe only supports raw DEFLATE.

Fixes: #6626

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/ApplicationModule.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/ApplicationModule.kt
@@ -30,6 +30,8 @@ val appModule = module {
         }
         install(ContentEncoding) {
             gzip()
+            // deflate is broken in KTOR, see https://youtrack.jetbrains.com/issue/KTOR-6999/Deflate-ContentEncoder-incorrectly-uses-raw-DEFLATE
+            // deflate()
             identity()
         }
     } }


matkoniecz:
Thanks!

Have you tested the new app version?

Sadly I am not able to test it right now - and not sure how to test it to avoid triggering rewrite rule, maybe change also user agent? 
Is that rewrite rule applied also to dev server?

Firefishy:
No, I've not tested it.

The dev server does not have the workarounds applied.

westnordost:
No need to test, this will work. Thank you @Firefishy !

I wonder if this should be reported upstream (and that issue report be mentioned in a comment at that point in the code). I have to admit that I don't quite understand the cause of the issue yet. Just scanned the [wikipedia article](https://en.wikipedia.org/wiki/HTTP_compression), which just has this to say:

> deflate – compression based on the [deflate](https://en.wikipedia.org/wiki/DEFLATE) algorithm (described in RFC [1951](https://www.rfc-editor.org/rfc/rfc1951)), a combination of the [LZ77](https://en.wikipedia.org/wiki/LZ77_and_LZ78#LZ77) algorithm and Huffman coding, wrapped inside the [zlib](https://en.wikipedia.org/wiki/Zlib) data format (RFC [1950](https://www.rfc-editor.org/rfc/rfc1950));

So, `deflate` doesn't work at all, or `deflate` doesn't work under certain circumstances? 
There's a sample application for using content encoding with ktor [here](https://github.com/ktorio/ktor-documentation/blob/3.3.2/codeSnippets/snippets/client-content-encoding/src/main/kotlin/com/example/Application.kt), which uses `deflate`. I seems kind of hard to believe that a general bug in `deflate` would have slipped under their radar.
When reporting an issue, ktor-people would probably want a reproduction, e.g. an API endpoint that offers this version of deflate that is incompatible with whatever implementation they have. Any OSM API call will do?

westnordost:
(By the way, digged a bit into their code. On JVM, I end up here. 

https://github.com/ktorio/ktor/blob/c68f4ee0d5c37ed516498d705562528ee4c33be5/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt#L51-L60

Didn't dig deeper.)

westnordost:
Turns out there is already an issue report. I added my reproduction. https://youtrack.jetbrains.com/issue/KTOR-6999/Deflate-ContentEncoder-incorrectly-uses-raw-DEFLATE

PR #6631, opened 2025-11-21
fixed form compleate after selecting road name

### LINKED ISSUE #6630: What's the name of this road? - unexpected behaviour off app.
When answering the quest "What's the name of this road?" If i tap an already named road to apply the name to the selected way, the "tick box" for applying the answer doesn't appear until I edit the text. (personal opinion, i preferred the drop down menu, which is now gone) 

While this doesn't add to much time to editing, it is strange behaviour to select the correct name from another road, but still have to edit the text to confirm it.

I would have thought that text populating the box, would make the tick appear, and makes confirming quite "clunky"

Street Complete app version is V62.0
Android version is Android 12, on Samsung One UI version 4.1 (galaxy S10e)

I have also tried this on a second android device (galaxy Zflip5), with the same results

I have a YouTube Video link of the "issue": https://youtube.com/shorts/kFFkBjSFIAE?feature=share

p.s. please bring back the drop down box, it was my love 🥺, and i had hoped it would get expanded to things like adding a road to a building address

paulklie:
fixes #6630

### DIFF
--- a/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/road_name/AddRoadNameForm.kt
+++ b/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/road_name/AddRoadNameForm.kt
@@ -29,9 +29,10 @@ class AddRoadNameForm : AAddLocalizedNameForm<RoadNameAnswer>() {
     override fun onClickMapAt(position: LatLon, clickAreaSizeInMeters: Double): Boolean {
         nameSuggestionsSource.getNames(position, clickAreaSizeInMeters, roadsWithNamesFilter)
             .firstOrNull()
-            ?.let { localizedNames.value = it }
-
-        checkIsFormComplete()
+            ?.let {
+                localizedNames.value = it
+                checkIsFormComplete()
+            }
         return true
     }
 


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/road_name/AddRoadNameForm.kt
```diff
@@ -31,6 +31,7 @@ class AddRoadNameForm : AAddLocalizedNameForm<RoadNameAnswer>() {
             .firstOrNull()
             ?.let { localizedNames.value = it }
 
+        checkIsFormComplete()
```
westnordost:
```suggestion
            ?.let {
                localizedNames.value = it
                checkIsFormComplete()
            }
```

The completeness only needs to be changed when localizedNAmes is set, not on every map click.

westnordost:
I didn't test my suggestion, I just wrote it in the browser, did you?

paulklie:
> I didn't test my suggestion, I just wrote it in the browser, did you?

yep! worked

westnordost:
Thanks!

PR #6642, opened 2025-12-07
expanded parking access to motorcycle parking

paulklie:
The motorcycle fee quest added in https://github.com/streetcomplete/StreetComplete/pull/6532, requires `access=yes`. 

I was planning to suggest a "who may park their motorcycle here?" quest, however I believe simply extending the existing access quest to [motorcycle_parking](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dmotorcycle_parking) is sufficient.

The string "Who is allowed to park here? Parking may be free or paid?" fits well to.

Tested.


westnordost:
LGTM

PR #6653, opened 2025-12-18
Quest: How much do you have to pay to use this parking spot?

marekkrug:
This PR is a new Quest with the name "How much do you have to pay to use this parking spot?" and was described in the Issue #6651.


app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
```diff
@@ -0,0 +1,196 @@
...
+        val timeUnit = when (timeUnitState.value) {
+            TimeUnit.HOUR -> "hour"
+            TimeUnit.DAY -> "day"
+        }
+
+        applyAnswer(SimpleCharge(amount, currency, timeUnit))
+    }
+
+    /** This function should probably be improved in the Future because it only matches some countries. */
+    private fun getCurrencyForCountry(): String = when (countryInfo.countryCode) {
```
paulklie: I wonder if we should move this to [country metadata](https://github.com/streetcomplete/countrymetadata)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
```diff
@@ -0,0 +1,42 @@
...
+import de.westnordost.streetcomplete.data.osm.geometry.ElementGeometry
+import de.westnordost.streetcomplete.data.osm.osmquests.OsmFilterQuestType
+import de.westnordost.streetcomplete.data.quest.AndroidQuest
+import de.westnordost.streetcomplete.data.user.achievements.EditTypeAchievement.CAR
+import de.westnordost.streetcomplete.osm.Tags
+
+class AddParkingCharge : OsmFilterQuestType<ParkingChargeAnswer>(), AndroidQuest {
+
+    override val elementFilter = """
+        nodes, ways, relations with amenity = parking
```
paulklie, line 23: I would also add motorcycle parking. Maybe also bicycle parking with a fee?

app/src/androidMain/res/values/strings.xml
```diff
@@ -1376,6 +1376,17 @@ If there are no signs along the whole street which apply for the highlighted sec
     <string name="quest_access_customers">Any person visiting a specific place (e.g. customers only)</string>
     <string name="quest_access_private">Only individual(s) with permission</string>
 
+    <!-- Parking Charge Quest -->
+    <string name="quest_parking_charge_amount_label">Amount in </string>
+    <string name="quest_parking_charge_varies_title">Variable pricing</string>
+    <string name="quest_parking_charge_title">How much do you have to pay to use this parking spot?</string>
```
paulklie: Hmm, we are not really asking for an individual spot right? It might be an entire garage

Maybe instead "How much do you need to pay to park here?"

paulklie (commented):
great work so far, here are some ideas / topics for discussion

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
```diff
@@ -0,0 +1,196 @@
...
+            "BG" -> "BGN"
+            "HR" -> "HRK"
+            "JP" -> "JPY"
+            "CN" -> "CNY"
+            "IN" -> "INR"
+            "AU" -> "AUD"
+            "CA" -> "CAD"
+            else -> "EUR"
+        }
+}
```
westnordost: We actually collect such meta-information in a separate repo:

https://github.com/streetcomplete/countrymetadata

In StreetComplete, we then parse this information via the `CountryInfo` class.

Also, this is apparently https://en.wikipedia.org/wiki/ISO_4217#Active_codes , so maybe add a comment this.

So, it would be good if you created a PR that adds a file like `data/currency.yml` in that repo.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
paulklie, line 1: If this is merged I suspect the form might be reused for similar quests which ask about the charge of, for example, toilets.
For this case I think it would make sence to place it in a folder named just "charge" instead, so similar quests can be grouped together here.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
```diff
@@ -0,0 +1,196 @@
...
+                    // Choose time Unit
+                    Text(
+                        text = getString(R.string.quest_parking_charge_time_unit_label),
+                        style = MaterialTheme.typography.body1
+                    )
+
+                    Row(
+                        modifier = Modifier.fillMaxWidth(),
+                        horizontalArrangement = Arrangement.spacedBy(8.dp)
+                    ) {
```
westnordost: To select a time unit, I would very much prefer a dropdown button (aka pop-up button). I already implemented a generic composable like this, have a look in `de.westnordost.streetcomplete.ui.common.DropdownButton`. See the `MaxHeightForm` for an example how it is used.

westnordost (commented):
Regarding UI design, I'd somewhat prefer a UI like this

```
[        ] EUR / [ hour ▼ ]
```
 
The currency simply being a text (unless there are *actually* several currency in use in a country??).
(Bonus: It would be nice if not the ISO currency id was used in display, but the actual currency symbol, e.g. €, $, £, ...)

The amount, being a decimal input field, you can use the generic composable I created `de.westnordost.streetcomplete.ui.common.input.DecimalInput` which handles decimal points locale-aware. I guess allowed would be at most 2 decimal digits?

Finally, in the issue it came up that actual pricing may vary depending on a number of factors which may confuse the user what to answer. An example from the discussion on the wiki page:
- 1st hour: 2$ / hour
- subsequent hours: 1$ / hour

Is there a reasonable hint text that could be added to explain what the user should answer in this case?

westnordost:
Also, I supposed specifically for this quest, it would make sense to *always* add `check_date:charge=<date>` because this information tends to be so volatile. If I remember correctly, we do the same for opening hours.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_charge/AddParkingChargeForm.kt
```diff
@@ -0,0 +1,196 @@
...
+            "CH" -> "CHF"
+            "DK" -> "DKK"
+            "SE" -> "SEK"
+            "NO" -> "NOK"
+            "PL" -> "PLN"
+            "CZ" -> "CZK"
+            "HU" -> "HUF"
+            "RO" -> "RON"
+            "BG" -> "BGN"
+            "HR" -> "HRK"
```
mnalis: Actually, Croatia ("HR") switched to "EUR" too [back in 2023](https://en.wikipedia.org/wiki/Croatia_and_the_euro)

mnalis (changes requested):

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_charge/AddParkingChargeForm.kt
marekkrug: We will probably refactor this code anyways, as you can see [here](https://github.com/streetcomplete/StreetComplete/pull/6653#discussion_r2632317333), but i will consider it when I'm doing that 👍

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
westnordost: The wikipedia page is pretty interesting. Did you know that in 2027 all the states in West Africa will get a common currency ([the eco](https://en.wikipedia.org/wiki/Eco_(currency))), many of which already have a common one (the West African CFA franc)?

Also, many states in Central Africa also have a common currency, the [Central African CFA franc](https://en.wikipedia.org/wiki/Central_African_CFA_franc).

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/parking_charge/AddParkingChargeForm.kt
marekkrug: i've fixed it now :)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
marekkrug: Do you like this one better? I'm still not sure how to seperate it properly and if I should rather do a "/" or a "per" inbetween.
Screenshot_20251220-193119
Screenshot_20251220-193305

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
marekkrug: I've created a [Pull Request](https://github.com/streetcomplete/countrymetadata/pull/32) for the two new yaml files.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
marekkrug: That's a good idea. I also think that adding certain vending machines would also be a good idea. In many cases, particularly street side parking, it makes more sense to add a [vending=parking_tickets](https://wiki.openstreetmap.org/wiki/Tag:vending%3Dparking_tickets) and adding the price tag there rather than putting it into the way object of the street it's on. Just an example of how it would be done:
https://www.openstreetmap.org/node/13240506652

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
westnordost: Very much, yes. Although,
- why the duplicate title question? Doesn't make sense
- I think a label for the amount input is unnecessary, the context is already clear by the question.
- the items in the `Row` should be centered vertically
- on code quality, it would be ideal if the whole "charge amount input" (amount + time unit) was outsourced into an own composable, just as the `CountInput` has been outsourced into an own composable. Then, the `AddParkingChargeForm` would just use that `ChargeInput` just like the `AAddCountForm` uses the `CountInput`. (Also because someone else mentioned that the charge input stuff could/should be reused later for other quests.)
- the text field for the amount can be much shorter, people are not going to enter numbers like 9999999. On that note, since there are just two input fields, I think the whole form may be larger. You can wrap the whole form into `ProvideTextStyle(MaterialTheme.typography.largeInput) { ... }` or even `extraLargeInput` (see the doc comments for these two properties).
- the time unit button is very far to the right, that's where the OK button appears. To avoid that these two overlap, the `Row` could be centered horizontally and not take the whole space.
- a bit inconsistent to use an outlined text field for the amount and a filled one for the time unit. Intuitively, I'd have done it the other way round. In doubt, better do it like in other forms, for consistency. The `CountInput` for example uses a normal (i.e. filled) text input, and generally for dropdowns I have used an outlined select button  as long as the dropdown is not the primary action.

westnordost:
Also, regarding localization:

1.
Did you test this in a right-to-left language? (e.g. Persian, Arabic)?

2.
I think there are variations where the currency symbol goes in different languages/regions, as well as for example with how many decimals a currency are allowed. In Germany, € is behind the sum and we have two decimals, but I think in some languages, it is written in front and there is a different number of decimals. It would be too much effort I think to research all that and have a huge table where it should go, but perhaps Java already provides a way to format a currency locale aware? Indeed, yes it has. Check [`NumberFormat.getCurrencyInstance`](https://docs.oracle.com/javase/8/docs/api/java/text/NumberFormat.html#getCurrencyInstance-java.util.Locale-). So, how to implement this in multiplatform? (I.e. we don't want to use Java directly in the composables). Check out `NumberFormatter.kt` to learn how to do it. Basically, we define an `expect` class/interface and have to implement it for each supported platform. (`NumberFormatter.android.kt`, ``NumberFormatter.ios.kt``). How the interface should look like depends on what we need. It seems to me we want to know three things:

1. the currency symbol. (so, we don't need to collect and maintain this information in countrymetadata, hooray!)
2. where the currency symbol goes
3. to how many decimals the number is formatted

If this information is not somehow available directly in Java and iOS, I guess the most straightforward way would be to (once) format an example value like "9.00" as currency and then parse it with a Regex (or something) to find this data in order to configure the `ChargeInput` correctly.

Point two is a bit of work while also not that important. But it would be nice to have this done "properly".

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
westnordost: Also, the dropdown that just presents 30min and 15min is a bit problematic. Already now I see in taginfo a plethora of other minutes defined. Maybe we'd really need another input field in which the user can define x day(s), x hour(s) or x minute(s).

(This **may** again make the input form too wide to conveniently fit into one line)

westnordost: Aha, we definitely need a composable that allows the user to input a time + time unit. Because the `AddParkingFeeForm` needs it. See `DurationInputViewController`. This is still the old Android-views-code, but will soon be recreated by me in compose. (in the opening_hours branch)

So, best you do nothing in this regard for now, and later when I am done with the opening_hours branch, you use the new `DurationInput` composable instead of that dropdown of yours.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
paulklie: Hmm, not all payed parking areas have vending machines though. Here in sweden alot just have a sign/qr code so you can pay with app / sms.

app/src/androidMain/res/values-de/strings.xml
```diff
@@ -912,6 +912,18 @@ Falls du von der Flut der Aufgaben überwältigt bist, kannst du in den Einstell
     <string name="quest_access_yes">"Alle Personen der Öffentlichkeit"</string>
     <string name="quest_access_customers">"Jede Person, die einen bestimmten Ort besucht (z. B. nur Kunden)"</string>
     <string name="quest_access_private">"Nur Einzelne mit Erlaubnis"</string>
+
+    <!-- Parking Charge Quest -->
+    <string name="quest_parking_charge_title">Wie viel muss man bezahlen, um hier zu parken?</string>
```
paulklie: this file is automatically generated. thus any changes would be overwritten.
Translations need to be added here https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING.md#translating-the-app

marekkrug:
@westnordost Today, I've implemented the CurrencyFormatter for both iOS and Android (I didn't get to test it on iOS so far because I don't use any apple products but this is how far i've got, but on Android it works). 
The quest now correctly shows the currency either in front or behind of the amount input field and chooses the currency via the `countryInfo.countryCode` of the quest the user is currently working on.

screenshots-sc-charge-quest

I've also increased the font size and added a quest hint that is just a proposal. Maybe you have an opinion about how to improve it.

I'm still unsure about the right-to-left language compatability, I will try to have a look into them in the next few days :)

app/src/androidMain/res/values-de/strings.xml
marekkrug: Okay. I've primarily used this to test the different locales. So I guess I'm going to remove it before the PR gets through and then put it in via the translator 👍

westnordost (commented):
Regarding the currency formatter, it looks good on first view, I prefer an API like this, however:

1.

have an expect/actual 
```kt
expect class CurrencyFormatter(val locale: Locale) {
    fun format(amount: Double): String
    val currencyCode: String
}
```

2.
have a data class like this, which is in common code(!)
```kt
data class CurrencyFormatElements(
  val symbol: String,
  val isSymbolBeforeAmount: Boolean,
  val decimalDigits: Int,
) {
    companion object {
        fun of(locale: Locale?): CurrencyFormatElements {
          // TODO do the format of the test string, then regex on the result
        }
    }
}
```

Basically, similar structure/method as used here:

https://github.com/streetcomplete/StreetComplete/pull/6691/changes#diff-9f3a21c0597c3ee1c3d82960f150b9c6f0f876b19fbe67becf1a59789cfd5e69 (if the link does not correctly point to the file, the file is TimeFormatElements.kt)

This is mainly 
- consistent with the approach linked
- no need to implement that regex stuff on every platform

On whether this `CurrencyFormatElements.of(locale)` is the nicest pattern/approach, I don't have a strong opinion. It is just consistent to my approach with TimeFormatElements, but I could be swayed if there are reasons for having a different interface.

---

Also, don't forget to add unit tests for that!

westnordost:
Also, the dropdown for duration unit, I added 

https://github.com/streetcomplete/StreetComplete/pull/6691/changes#diff-5fcb8673a43b9e77b5d7bb3a8868616fb209fbf57d361a9b5bb818759b316520

Maybe you should base your PR on top of that PR (or wait until that is merged) so that you can just reuse that widget.

app/src/androidMain/res/values/strings.xml
```diff
@@ -1376,6 +1376,18 @@ If there are no signs along the whole street which apply for the highlighted sec
...
+    <string name="quest_parking_charge_varies_title">Variable pricing</string>
+    <string name="quest_parking_charge_title">How much do you need to pay to park here?</string>
+    <string name="quest_parking_charge_varies">It varies…</string>
+    <string name="quest_parking_charge_time_unit_label">per</string>
+    <string name="quest_parking_charge_per_hour">Hour</string>
+    <string name="quest_parking_charge_per_day">Day</string>
+    <string name="quest_parking_charge_varies_description">Describe how the parking charges vary (e.g. "1 EUR/h weekdays, 2 EUR/h weekends")</string>
+    <string name="quest_parking_charge_per_30min">30min</string>
+    <string name="quest_parking_charge_per_15min">15min</string>
+    <string name="quest_parking_charge_hint">"For simplicity's sake, it is recommended that you enter the standard amount for different amounts in this quest. For example, if a parking space costs €3 per hour during the day and €1 per hour at night, enter €3 per hour. This makes it easier to compare prices from the data later on."</string>
```
paulklie:
```suggestion
    <string name="quest_parking_charge_hint">"For simplicity's sake, it is recommended that you enter the standard amount for different amounts in this quest. For example, if it costs €3 during the day and €1 at night, enter €3. This makes it easier to compare prices from the data later on."</string>
```
I suggest generalizing this, so the hint can be reused for other `charge` quests (for example for toilets) without needed additional translations.

app/src/androidMain/res/values/strings.xml
```diff
@@ -1376,6 +1376,18 @@ If there are no signs along the whole street which apply for the highlighted sec
...
+    <string name="quest_parking_charge_varies_title">Variable pricing</string>
+    <string name="quest_parking_charge_title">How much do you need to pay to park here?</string>
+    <string name="quest_parking_charge_varies">It varies…</string>
+    <string name="quest_parking_charge_time_unit_label">per</string>
+    <string name="quest_parking_charge_per_hour">Hour</string>
+    <string name="quest_parking_charge_per_day">Day</string>
+    <string name="quest_parking_charge_varies_description">Describe how the parking charges vary (e.g. "1 EUR/h weekdays, 2 EUR/h weekends")</string>
+    <string name="quest_parking_charge_per_30min">30min</string>
+    <string name="quest_parking_charge_per_15min">15min</string>
+    <string name="quest_parking_charge_hint">"For simplicity's sake, it is recommended that you enter the standard amount for different amounts in this quest. For example, if it costs €3 during the day and €1 at night, enter €3. This makes it easier to compare prices from the data later on."</string>
```
FloEdelmann: Please use typographic quotation marks and apostrophes:

```suggestion
    <string name="quest_parking_charge_varies_description">Describe how the parking charges vary (e.g. “1 EUR/h weekdays, 2 EUR/h weekends”)</string>
    <string name="quest_parking_charge_per_30min">30min</string>
    <string name="quest_parking_charge_per_15min">15min</string>
    <string name="quest_parking_charge_hint">"For simplicity’s sake, it is recommended that you enter the standard amount for different amounts in this quest. For example, if it costs €3 during the day and €1 at night, enter €3. This makes it easier to compare prices from the data later on."</string>
```

marekkrug:
> Also, the dropdown for duration unit, I added
> 
> https://github.com/streetcomplete/StreetComplete/pull/6691/changes#diff-5fcb8673a43b9e77b5d7bb3a8868616fb209fbf57d361a9b5bb818759b316520
> 
> Maybe you should base your PR on top of that PR (or wait until that is merged) so that you can just reuse that widget.

I've rebased the project to that branch and tried to use the DurationUnitDropdown. The problem is that all the values there are in plural ("hours", "minutes", "days"), but in OSM you need "hour", "minute", "day" and probably "night" and "day" in the Future (https://wiki.openstreetmap.org/wiki/Key:charge#Usage) 🤔

westnordost:
whats the problem with plural?

marekkrug:
writing `2.00 USD/hours` would be a bit weird in my opinion, in the wiki, the correct way to write it is `2.00 USD/hour`...

westnordost:
ah, right

FloEdelmann:
@marekkrug Sorry, I didn't mean to close your PR. I cleaned up and deleted the `opening_hours` branch which this PR is based on, which accidentally also closed this PR, and now it's not possible to change the target branch back to `master` here (not even after restoring the `opening_hours` branch).

I expected GitHub to behave like GitLab, which would have just changed this PR's target branch to `master` when merging #6691.

Could you please open another PR? Sorry for the extra work!

westnordost:
I just created a new branch `opening_hours` from master, now it can be reopened. I prefer this, so we don't forget about this PR and also this PR contains a lot of discussion.
In the end, we just need to remember to merge that branch into master again.

FloEdelmann:
Thanks! I could change the target branch back to `master` now, so I deleted the `opening_hours` branch once again. I'll now hide the comments as resolved, so back to parking spot fee stuff :slightly_smiling_face:

marekkrug:
The quest is now running and working as far as I can see.

Are there any more problems that need to be fixed in this quest? Otherwise, it would be fine for me if it was merged in this state and to later add quests like "How much do you have to pay to use this toilet?" or similar in a seperate merge request :)

westnordost:
I haven't been following this PR lately. Are all the previous comments resolved? As a PR author, you can tap that button yourself.

app/src/androidMain/res/values-de/strings.xml
marekkrug: I've checked RTL compatability. I don't speak any languages that are RTL so I don't really know if there are any edge cases I'm not covering, but it looks good in arab as far as i can see.

So I'm marking this conversation as resolved :)

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
marekkrug: I think everything that you mentioned was implemented. So i'm marking this conversation as resolved.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
marekkrug: i did that a while ago, so I'm marking this conversation as resolved :)

marekkrug:
I read through them and resolved a lot of them. Looks like there's nothing really remaining other than potential other quests.

westnordost:
Alright, I will do a final review soon

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/ParkingChargeAnswer.kt
```diff
@@ -0,0 +1,16 @@
...
+
+sealed interface ParkingChargeAnswer
+
+data class SimpleCharge(
+    // e.g. "1.50"
+    val amount: String,
+    // e.g. "EUR"
+    val currency: String,
+    // either "day", "hour" or "minute"
+    val timeUnit: String
```
westnordost: Unless there is a good reason not to do that, at least `amount` should be a `Double, `timeUnit` should be a `TimeUnit`.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
```diff
@@ -0,0 +1,132 @@
...
+        if (currencyFormatInfo.symbolBeforeAmount) {
+            Text(
+                text = currencyFormatInfo.symbol,
+                style = MaterialTheme.typography.h5,
+                color = MaterialTheme.colors.onSurface.copy(alpha = 0.6f),
+                modifier = Modifier.padding(end = 8.dp)
+            )
+        }
+
+        TextField(
```
westnordost: Use `DecimalInput`. This handles the locale aware decimal separator for you, as well as lets you set the maximum number of decimal places, plus the keyboard configuration.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    override fun onClickOk() {
+        val amount = amountState.value.replace(',', '.')
+        val currency = getCurrencyForCountry()
+        val timeUnit = when (durationUnitState.value) { // TODO: This could be removed if DurationUnit implements toOSMValue().
+            DurationUnit.HOURS -> "hour"
+            DurationUnit.DAYS -> "day"
+            DurationUnit.MINUTES -> "minute"
+        }
```
westnordost: create an extension function or add a method to `DurationUnit`, for example either parametrize the existing `toOsmValue` function to e.g. `toOsmValue(alwaysUseSingular: Boolean = false)` or add `toOsmValueSingular` function

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
```diff
@@ -0,0 +1,121 @@
...
+    } catch (_: Exception) {
+        "EUR"
+    }
+}
+
+fun DurationUnit.getDisplayName(form: AddParkingChargeForm): String = when (this) {
+    DurationUnit.HOURS -> form.getString(R.string.quest_parking_charge_hour)
+    DurationUnit.DAYS -> form.getString(R.string.quest_parking_charge_day)
+    DurationUnit.MINUTES -> form.getString(R.string.quest_parking_charge_minute_short)
+}
```
westnordost: Have a look at how the string resource is created for other quests / composable UI. Could also use the `DurationUnitDropdown` but add a parameter that decides whether the unit should be written always in singular or not (analogous to my comment above)

westnordost (changes requested):
Phew, I am sorry to say this, but the currency formatter stuff is quite bonkers.

Each only in parts:

- the formatting is all over the place and there is commented out stuff plus open TODO comments left
- the doc comments make no sense
- the iOS implementation at least again in parts can not work. E.g. `getCurrencyCodeFromLocale` just completely ignores the passed parameter and looks like you copy-pasted the code from the other method
- the API of `CurrencyFormatter` (again, as for every point here in parts) make no sense (why is `getCurrencyCodeFromLocale` a method of an instance of `CurrencyFormatter`, what does the class have to do with a currency code?) 
- the tests for `CurrencyFormatter` only test half of the interface
- the implementation of `CurrencyFormatter` on both platforms is over-complex and also wrong. The parameter is a country code and you search through all available locales to find the first locale that uses the currency of the country with the given code in order to format it (because you need a locale to format). Now, imagine the Euro is used in a country where one formats currency like this "1.500,00 €" (France)  and in another country like this "€ 1 500.00" (Ireland). Oopsie! So, always the `Locale` (with region) must be passed to `CurrencyFormatter`.
- the second static function of `CurrencyFormatElements` looks like it is completely copy-paste from the other one, plus, only one of these is used (outside tests) anyway! Funnily enough, the one that is *not tested* is the one that is actually used, while the one that is tested is *not* used in production 🤦
- well, maybe more things, I lost the overview

I reckon that at least parts of this confusion might have been introduced as misunderstandings due to my requests to change your initial implementation of this, of course.
And since I looked into this code so deeply already, it makes sense if I rewrite that part myself rather than play PR ping pong. So, that's what I will do, you don't need to worry about this part.

---

The other parts I only skimmed through, but nevertheless I found some stuff. First and foremost, please make sure that you review your own code, too. I notice there are some TODOs and similar comments as well as sections of leftover commented out code.

For the other stuff, see below. Note however that this was just from skimming through it, there might be more in the details.

marekkrug:
I finally took the time and fixed a lot of the remaining remarks. In the meantime, the getTitle function described at https://github.com/streetcomplete/StreetComplete/blob/master/CONTRIBUTING_A_NEW_QUEST.md#gettitle seems to have been updated, so I don't know how to fix the title. In order to let it compile again, I temporarily took the title from another quest and added a todo until I figure it out :)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/Duration.kt
```diff
@@ -4,15 +4,21 @@ import de.westnordost.streetcomplete.osm.duration.DurationUnit.*
...
-/** A duration, as used in OSM for e.g. maxstay */
+/** A duration, as used in OSM for e.g. maxstay or charge */
 @Serializable
 data class Duration(val value: Double, val unit: DurationUnit) {
-    fun toOsmValue(): String =
-        value.toShortString() + " " + when (unit) {
-            MINUTES -> if (value != 1.0) "minutes" else "minute"
-            HOURS -> if (value != 1.0) "hours" else "hour"
-            DAYS -> if (value != 1.0) "days" else "day"
+    fun toOsmValue(alwaysSingular: Boolean = false, ignoreValue: Boolean = false): String {
```
westnordost: Always singular? Ignore value? Looks like what you really want is to extend `DurationUnit` like this

```kt
enum class DurationUnit {
    MINUTES,
    HOURS,
    DAYS;

    fun toOsmValue(usePlural: Boolean = false): String = when (this) {
        MINUTES -> if (usePlural) "minutes" else "minute"
        HOURS -> if (usePlural) "hours" else "hour"
        DAYS -> if (usePlural) "days" else "day"
    } 
}
```

and use that accordingly, ditch `alwaysSingular: Boolean = false, ignoreValue: Boolean = false`

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
westnordost: 👆

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/ParkingChargeAnswer.kt
```diff
@@ -0,0 +1,18 @@
...
+    val amount: Double,
+    // e.g. "EUR"
+    val currency: String,
+    // either "day", "hour" or "minute"
+    val timeUnit: DurationUnit
+) : ParkingChargeAnswer
+
+data class ItVaries(
+    val conditional: String
+) : ParkingChargeAnswer
```
westnordost: Letting the user just input plain text for `charge:conditional`, e.g. `charge:conditional=on weekdays` is not acceptable.

This option must be removed or real alternative be implemented. Looking at current values for `charge:conditional`, I'd think this is not worth it, though. Better to leave a note.

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
```diff
@@ -0,0 +1,132 @@
...
+            onValueChange = onAmountChange,
+            placeholder = {
+                // Generate placeholder based on decimal places
+                val placeholderValue = when (currencyFormatInfo.decimalDigits) {
+                    0 -> "150"
+                    1 -> "15.0"
+                    else -> "1.50"
+                }
+                Text(placeholderValue)
+            },
```
westnordost: But why? Also, did you notice that this is not really localized, as the decimal point differs per language. Better just remove it.

westnordost (changes requested):
There are still some things commented out, and several TODOs. Do you need input for something?

app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingChargeForm.kt
marekkrug: Is this better now with the commit [fbb04b8](https://github.com/streetcomplete/StreetComplete/pull/6653/commits/fbb04b8792132aef6681427f9977d0e9b93bd894)? I tried to change the ChargeInput so that it now uses the DurationUnitDropdown but I'm not sure if I understood you correctly 🤔

westnordost: Yes

westnordost:
Anyone interested in finishing this?

marekkrug:
I think I've resolved most of the Issues you mentioned on 20.02.26, @westnordost. I'm also interested in finishing it. What's holding me back at the moment is that it has been quite a while since I wrote most of the code and that I lost the overview a little bit of what of your notes I have already fixed and what is still missing. Still, I tried to fix it all a few months ago.
Also, since this is the first time I'm contributing to StreetComplete, I was mainly struggling with finding the right places of resources and all the other stuff to use, so that's also something I am still not sure about.

It would be really helpful if you or someone else could review the current Iteration of the code once again and tell me what needs to be done :)

paulklie:
> It would be really helpful if you or someone else could review the current Iteration of the code once again and tell me what needs to be done :)

I think the most pressing issue, is fixing the merge conflicts. For this you would need to move the form and new ui components into the shared location. You can have a look at https://github.com/streetcomplete/StreetComplete/pull/6842 which already applied this for all other quests.

westnordost:
Yeah, it's easier for me to look at the PR again after the conflicts have been resolved / it has been rebased on current master. If previous PR comments have already been solved, then the final review will be somewhat of a no-brainer.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
```diff
@@ -0,0 +1,50 @@
...
+    override val wikiLink = "Key:charge"
+    override val icon = R.drawable.quest_parking_charge
+    override val achievements = listOf(CAR)
+
+    override val hint = Res.string.quest_parking_charge_hint
+
+    override val title = Res.string.quest_parking_charge_title
+
+    override fun createForm() = AddParkingChargeForm()
+
```
paulklie, line 44: I would suggest also highlighting other parking areas. This could help distinguish which area is meant in places with a lot of parking lots.

paulklie: example: https://github.com/streetcomplete/StreetComplete/blob/89e38a7ff08c0087b81a0d422df0ba0e99b40a6a/app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/aerialway/AddAerialwayBicycleAccess.kt#L60-L64

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddChargeForm.kt
paulklie, line 1: Nitpicky, but I think I would rename this to AddChargeForm, because I suspect we will reuse it for other, similar quests.

paulklie:
Great work!
I will try to test it tomorrow, only looked over the code for now

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/Duration.kt
paulklie: I am not sure this is worth it. Many languages use [more complex forms of plurals](https://www.unicode.org/cldr/cldr-aux/charts/22/supplemental/language_plural_rules.html) than the germanic (n=1 vs everything else). Might be better to always just have it in plural?

marekkrug:
The only thing that is bothering me is that it is possible to input `10.5€` and it doesn't tag it as `10.50 EUR/hour` but rather as `10.5 EUR/hour`. I will fix that bug tomorrow.

marekkrug:
@paulklie ready for review :)

marekkrug:
@paulklie ready for review :)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/DurationUnitDropdown.kt
```diff
@@ -22,12 +23,11 @@ fun DurationUnitDropdown(
...
-        itemContent = { Text(stringResource(it.text)) }
+        itemContent = { Text(stringResource(it.text(alwaysSingular))) }
     )
 }
-
-private val DurationUnit.text: StringResource get() = when (this) {
-    DurationUnit.MINUTES -> Res.string.unit_minutes
-    DurationUnit.HOURS -> Res.string.unit_hours
-    DurationUnit.DAYS -> Res.string.unit_days
+private fun DurationUnit.text(alwaysSingular: Boolean): StringResource = when (this) {
```
westnordost:
```suggestion
private fun DurationUnit.getText(useSingular: Boolean): StringResource = when (this) {
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddChargeForm.kt
```diff
@@ -0,0 +1,64 @@
...
+    val currencyFormatInfo = remember(countryInfo) {
+        CurrencyFormatElements.of(countryInfo.userPreferredLocale)
+    }
+
+    QuestForm(
+        on = on,
+        isComplete = amount.let { it != null && (it.units != 0L || it.subunits != 0L) },
+        onClickOk = {
+            val amount = amount!!
+            val currency = CurrencyFormatter(countryInfo.userPreferredLocale).currencyCode ?: "???"
```
westnordost, line 44:
```suggestion
            val currency = CurrencyFormatter(countryInfo.userPreferredLocale).currencyCode ?: "¤"
```
(This is the generic/placeholder currency symbol)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddChargeForm.kt
```diff
@@ -0,0 +1,64 @@
...
+    var amount by rememberSerializable { mutableStateOf<CurrencyAmount?>(null) }
+    var durationUnit by rememberSerializable { mutableStateOf(DurationUnit.HOURS) }
+
+    val currencyFormatInfo = remember(countryInfo) {
+        CurrencyFormatElements.of(countryInfo.userPreferredLocale)
+    }
+
+    QuestForm(
+        on = on,
+        isComplete = amount.let { it != null && (it.units != 0L || it.subunits != 0L) },
```
westnordost:
```suggestion
        isComplete = amount != null && amount.units != 0 && amount.subunits != 0,
```

I just think this reads better, no functional change.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/Duration.kt
```diff
@@ -18,5 +18,11 @@ data class Duration(val value: Double, val unit: DurationUnit) {
 enum class DurationUnit {
```
westnordost, line 19:
```suggestion
@Serializable
enum class DurationUnit {
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeAnswer.kt
```diff
@@ -0,0 +1,18 @@
+package de.westnordost.streetcomplete.quests.charge
+
+import de.westnordost.streetcomplete.osm.duration.DurationUnit
+
+sealed interface ChargeAnswer
```
westnordost: This interface is not necessary when there is only one class that implements it.

Rename `SimpleCharge` to `Charge` and use that as the class for `AddParkingCharge`.

Don't forget to rename the file.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/Duration.kt
marekkrug: But we never have the situation that the quest is with more than one time unit (`2 EUR/4 hours` would never happen). The quest is designed so that only one time unit is possible (`0.5 EUR/hour`).
Or am I missing something?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/locale/CurrencyAmount.kt
westnordost, line 1: This class (and the accompanying test) doesn't make sense to me.

Why not use a `Float` (or `Double` 🤷)? `Float` has enough precision to represent 2 decimals.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
```diff
@@ -0,0 +1,96 @@
...
+                modifier = Modifier.padding(start = 8.dp)
+            )
+        }
+        Text(
+            text = perLabel,
+            style = MaterialTheme.typography.body1
+        )
+        Column(
+            verticalArrangement = Arrangement.spacedBy(4.dp)
+        ) {
```
westnordost: what is this column for?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
```diff
@@ -0,0 +1,96 @@
...
+        horizontalArrangement = Arrangement.spacedBy(8.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        if (currencyFormatInfo.isSymbolBeforeAmount) {
+            Text(
+                text = currencyFormatInfo.symbol,
+                style = MaterialTheme.typography.h5,
+                color = MaterialTheme.colors.onSurface.copy(alpha = 0.6f),
+                modifier = Modifier.padding(end = 8.dp)
+            )
```
westnordost, line 41: no need to add padding for each element, that is what `horizontalArrangement = Arrangement.spacedBy(8.dp)` in the `Row` is for.

Also, this is not a heading, so it shouldn't be styled as a heading. Also, usually only hints and similar get an alpha.

app/src/commonMain/composeResources/values/strings.xml
```diff
@@ -1395,10 +1395,17 @@
...
+    <string name="unit_hour">hour</string>
+    <string name="unit_minute">minute</string>
     <string name="at_any_time">at any time</string>
     <string name="only_at_hours">only at following times:</string>
     <string name="except_at_hours">except at following times:</string>
 
+    <string name="quest_parking_charge_title">How much do you need to pay to park here?</string>
+    <string name="quest_parking_charge_time_unit_label">per</string>
+    <string name="quest_parking_charge_hint">"For simplicity’s sake, it is recommended that you enter the standard amount for different amounts in this quest. For example, if it costs €3 during the day and €1 at night, enter €3. This makes it easier to compare prices from the data later on."</string>
+
```
westnordost:
```suggestion
```
This string is unused as far as I can see. (For good reason, IMO it is confusing.)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        val amountAsDouble = amount?.toDouble(currencyFormatInfo.decimalDigits)
+        AutoFitTextFieldFontSize(
+            value = amountAsDouble?.toString().orEmpty(),
+            modifier = Modifier.weight(1f, fill = false)
+        ) {
```
westnordost, line 47: Why this?

westnordost (changes requested):
Can you post a screenshot/video on how the interface looks at the end?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/Duration.kt
westnordost: I don't think you are missing something, it's fine.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
marekkrug: was not needed any more

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
marekkrug: I also thought The combination of AutoFit and the fillMaxWidth Modifier was redundant. But when I removed the fillMaxWidth, it was still filling up the whole area. You did use the same AutoFit in e.g. LengthMetersInput 🤔
```kotlin
@Composable
fun LengthMetersInput(
    length: Length.Meters?,
    onChange: (Length.Meters?) -> Unit,
    maxMeterDigits: Pair<Int, Int>,
    modifier: Modifier = Modifier,
    locale: Locale = Locale.current,
    style: TextFieldStyle = TextFieldStyle.Filled,
) {
    Row(
        modifier = modifier,
        verticalAlignment = Alignment.CenterVertically,
        horizontalArrangement = Arrangement.spacedBy(8.dp)
    ) {
        AutoFitTextFieldFontSize(
            value = length?.meters?.toShortString().orEmpty(),
            modifier = Modifier.weight(1f)
        ) {
            DecimalInput(
                value = length?.meters,
                onValueChange = { meters ->
                    onChange(meters?.let { Length.Meters(it) })
                },
                modifier = Modifier.fillMaxWidth(),
                maxIntegerDigits = maxMeterDigits.first,
                maxFractionDigits = maxMeterDigits.second,
                locale = locale,
                isUnsigned = true,
                style = style,
            )
        }
        Text("m")
    }
}
```

marekkrug: Screenshot_20260815_163049_StreetComplete Dev
without AutoFitTextFieldFontSize
Screenshot_20260815_162125_StreetComplete Dev
with AutoFitTextFieldFontSize

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
marekkrug: What style should I do instead of the heading h5 for the currency code? I did this styling [a while ago](https://github.com/streetcomplete/StreetComplete/pull/6653#discussion_r2643063719) and haven't changed it since.

app/src/commonMain/composeResources/values/strings.xml
marekkrug: It is being used as a hint. I discussed the hint with @paulklie [a while ago](https://github.com/streetcomplete/StreetComplete/pull/6653#discussion_r2702884841) and that was our result, but we can also remove it if you think that's better.

marekkrug:
bitmap
Here are some Screenshots from the current state of the quest :)

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
marekkrug: Should I just keep it that way or do you have a better option?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
westnordost: I'd say, apply no style.

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
westnordost: Oh, I see. Well then, the form looks good this way, so let's keep it as is.

app/src/commonMain/composeResources/values/strings.xml
westnordost: Hm well, I find it quite confusing. Also, as an advice to correctly specify it, somewhat questionable. The wiki page doesn't mention that it would be fine to only specify the "standard charge" (whatever that may be) if the situation is more complex and it would be very un-OSM-like to suggest something like this.

https://wiki.openstreetmap.org/wiki/Key:charge

€3 during day and €1 during night would also be taggable (just not in the app, because no interface for that has been built). E.g.

`charge=3 EUR/hour`
`charge:conditional=1 EUR/hour @ (20:00-06:00)`

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+class AddParkingCharge : OsmFilterQuestType<Charge>() {
+    override val elementFilter = """
+        nodes, ways, relations with amenity = parking
+        and access ~ yes|customers|public
+        and fee = yes
+        and (
+            !charge and !charge:conditional
+            or charge older today -18 months
+        )
```
westnordost: This means it would be asked also on `charge:conditional=...` when it is old enough. But since the app doesn't offer a way to specify more complicated charge conditions, it should **never** show again if `charge:conditional is specified. 

I.e.


```suggestion
        and !charge:conditional
        and (!charge or charge older today -18 months)
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/locale/CurrencyAmount.kt
marekkrug: It was a mix of

- It is nice to have a toString() method that does all the formatting
- It's nice to have it in a half-parsed form for converting it to a string later and to do logic with the units/subunits like this:
`isComplete = amount != null && (amount!!.units != 0L || amount!!.subunits != 0L),`
- I heard somewhere that, as a principle, you should never store money as a floating point number 
 
But I see the argument that it could be done a lot simpler without any intermediate structure. Should I refractor it to only use `Double`?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
marekkrug: Screenshot_20260815_181046_StreetComplete Dev
👍

app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/locale/CurrencyAmount.kt
westnordost: Yes please. As you want to format the string to have only X decimals and exactly this number of decimals, you could add a function like

```kt
/** Formats this double to exactly the given number of [decimals]. 
  * E.g. both `12.305` and `12.3` become `"12.30"` when [decimals] is 2. */
fun Double.formatPadded(decimals: Int): String {
    val int = toInt()
    val fraction = ((this - toInt()) * 10.0.pow(decimals)).toInt()
    val fractionPadded = fraction.toString().padStart(decimals, '0')
    return "$int.$fractionPadded"
}
```

to `de.westnordost.streetcomplete.util.ktx.Double.kt`

app/src/commonMain/composeResources/values/strings.xml
marekkrug: Just an idea:
`<string name="quest_parking_charge_hint">"Enter the charge and the time unit it applies to, e.g. per hour or per day. If the charge varies, e.g. by time of day or vehicle type, this cannot be fully expressed here yet. Either enter the amount that most commonly applies or create an Issue if you want to."</string>`

app/src/commonMain/kotlin/de/westnordost/streetcomplete/util/locale/CurrencyAmount.kt
marekkrug: done

marekkrug: oh, I just found an error: if you input `2.4`, it tags `2.39`, the same with `2.8` -> `2.79`. Seems to be a rounding error...

marekkrug: replacing `.toInt()` with `.roundToInt()` seems to fix it for now

app/src/commonMain/composeResources/values/strings.xml
marekkrug: @westnordost I'm also okay with just removing the hint. Should I remove it?

westnordost: yes, please do

marekkrug:
What are the next steps to finishing this PR? Do you still want to test the quest, @paulklie ?

app/src/commonMain/kotlin/de/westnordost/streetcomplete/osm/duration/DurationUnitDropdown.kt
```diff
@@ -14,6 +14,7 @@ import org.jetbrains.compose.resources.stringResource
 fun DurationUnitDropdown(
     selectedDuration: DurationUnit,
     onSelectedDuration: (DurationUnit) -> Unit,
+    alwaysSingular: Boolean = false,
     modifier: Modifier = Modifier
```
westnordost:
```suggestion
    modifier: Modifier = Modifier,
    alwaysSingular: Boolean = false,
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddParkingCharge.kt
```diff
@@ -0,0 +1,51 @@
...
+    """
+    override val changesetComment = "Add parking charges"
+    override val wikiLink = "Key:charge"
+    override val icon = Res.drawable.quest_parking_charge
+    override val achievements = listOf(CAR)
+    override val title = Res.string.quest_parking_charge_title
+
+    override fun getHighlightedElements(element: Element, mapData: MapDataWithGeometry) =
+        mapData.filter("""
+             nodes, ways with amenity = parking
```
westnordost:
```suggestion
             nodes, ways, relations with amenity = parking
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/ChargeInput.kt
```diff
@@ -0,0 +1,80 @@
...
+                        KeyboardType.Number
+                    }
+                )
+            )
+        }
+
+        if (!currencyFormatInfo.isSymbolBeforeAmount) {
+            Text(
+                text = currencyFormatInfo.symbol,
+                style = MaterialTheme.typography.h5,
```
westnordost:
```suggestion
```

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddChargeForm.kt
```diff
@@ -0,0 +1,64 @@
...
+    val currencyFormatInfo = remember(countryInfo) {
+        CurrencyFormatElements.of(countryInfo.userPreferredLocale)
+    }
+
+    QuestForm(
+        on = on,
+        isComplete = amount != null && amount != 0.0,
+        onClickOk = {
+            val amount = amount!!
+            val currency = CurrencyFormatter(countryInfo.userPreferredLocale).currencyCode ?: "¤"
```
westnordost:
```suggestion
            val currency = CurrencyFormatter(countryInfo.userPreferredLocale).currencyCode ?: "???"
```

Ah, it's the currency **code**, not the currency symbol. Then, actually, "???" was better

app/src/commonMain/kotlin/de/westnordost/streetcomplete/quests/charge/AddChargeForm.kt
```diff
@@ -0,0 +1,64 @@
...
+    ) {
+        ProvideTextStyle(MaterialTheme.typography.extraLargeInput) {
+            ChargeInput(
+                amount = amount,
+                onAmountChange = { amount = it },
+                currencyFormatInfo = currencyFormatInfo,
+                durationUnit = durationUnit,
+                onDurationUnitChange = { durationUnit = it },
+                perLabel = stringResource(Res.string.quest_parking_charge_time_unit_label),
+                modifier = Modifier.padding(16.dp)
```
westnordost:
```suggestion
```

westnordost (approved):
Looks good to me now!

